PR #6918, opened 2023-02-01
NIFI-11129 - adding PutMongoBulk processor to use the bulkWrite API

sebastianrothbucher:
NIFI-11129 - adding PutMongoBulk processor to use the bulkWrite API for way more efficient mass updates or inserts

# Summary

[NIFI-11129](https://issues.apache.org/jira/browse/NIFI-11129)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

- [x]  Pull Request refers to a feature branch with one commit containing changes

# Verification

### Build

  - [x] JDK 8 (nifi-mongodb-processors)
  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

- [ ] n/a New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [ ] n/a New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [ ] n/a (but tested additionalDetails.html) Documentation formatting appears as expected in rendered files

### DIFF
--- a/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
+++ b/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
@@ -17,7 +17,6 @@
 package org.apache.nifi.processors.mongodb;
 
 import com.mongodb.WriteConcern;
-import com.mongodb.client.ClientSession;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.model.BulkWriteOptions;
 import com.mongodb.client.model.Collation;
@@ -61,10 +60,10 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 @Tags({ "mongodb", "insert", "update", "write", "put", "bulk" })
 @InputRequirement(Requirement.INPUT_REQUIRED)
@@ -86,16 +85,6 @@ public class PutMongoBulkOperations extends AbstractMongoProcessor {
             .defaultValue("true")
             .build();
 
-    static final PropertyDescriptor TRANSACTIONS_ENABLED = new PropertyDescriptor.Builder()
-            .name("Transactions Enabled")
-            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
-            .description("Run all actions in one MongoDB transaction")
-            .required(false)
-            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
-            .allowableValues("true", "false")
-            .defaultValue("false")
-            .build();
-
     static final PropertyDescriptor CHARACTER_SET = new PropertyDescriptor.Builder()
         .name("Character Set")
         .description("The Character Set in which the data is encoded")
@@ -104,22 +93,10 @@ public class PutMongoBulkOperations extends AbstractMongoProcessor {
         .defaultValue("UTF-8")
         .build();
 
-    private final static Set<Relationship> relationships;
-    private final static List<PropertyDescriptor> propertyDescriptors;
-
-    static {
-        List<PropertyDescriptor> _propertyDescriptors = new ArrayList<>();
-        _propertyDescriptors.addAll(descriptors);
-        _propertyDescriptors.add(ORDERED);
-        _propertyDescriptors.add(TRANSACTIONS_ENABLED);
-        _propertyDescriptors.add(CHARACTER_SET);
-        propertyDescriptors = Collections.unmodifiableList(_propertyDescriptors);
-
-        final Set<Relationship> _relationships = new HashSet<>();
-        _relationships.add(REL_SUCCESS);
-        _relationships.add(REL_FAILURE);
-        relationships = Collections.unmodifiableSet(_relationships);
-    }
+    private final static Set<Relationship> relationships =
+            Stream.of(REL_SUCCESS, REL_FAILURE).collect(Collectors.toUnmodifiableSet());
+    private final static List<PropertyDescriptor> propertyDescriptors =
+            Stream.concat(descriptors.stream(), Stream.of(ORDERED, CHARACTER_SET)).collect(Collectors.toUnmodifiableList());
 
     @Override
     public Set<Relationship> getRelationships() {
@@ -141,7 +118,6 @@ public class PutMongoBulkOperations extends AbstractMongoProcessor {
         final Charset charset = Charset.forName(context.getProperty(CHARACTER_SET).getValue());
         final WriteConcern writeConcern = clientService.getWriteConcern();
 
-        ClientSession clientSession = null;
         try {
             final MongoCollection<Document> collection = getCollection(context, flowFile).withWriteConcern(writeConcern);
 
@@ -172,39 +148,15 @@ public class PutMongoBulkOperations extends AbstractMongoProcessor {
                 updateModels.add(writeModel);
             }
 
-            if (context.getProperty(TRANSACTIONS_ENABLED).asBoolean()) {
-                clientSession = clientService.startSession();
-                clientSession.startTransaction();
-                // now run this w/in a transaction
-                collection.bulkWrite(clientSession, updateModels, (new BulkWriteOptions().ordered(context.getProperty(ORDERED).asBoolean())));
-            } else {
-                collection.bulkWrite(updateModels, (new BulkWriteOptions().ordered(context.getProperty(ORDERED).asBoolean())));
-            }
+            collection.bulkWrite(updateModels, (new BulkWriteOptions().ordered(context.getProperty(ORDERED).asBoolean())));
             getLogger().info("bulk-updated {} into MongoDB", flowFile);
 
             session.getProvenanceReporter().send(flowFile, getURI(context));
             session.transfer(flowFile, REL_SUCCESS);
-
-            if (clientSession != null) {
-                if (clientSession.hasActiveTransaction()) {
-                    clientSession.commitTransaction();
-                }
-                clientSession.close();
-            }
         } catch (Exception e) {
             getLogger().error("Failed to bulk-update {} into MongoDB", flowFile, e);
             session.transfer(flowFile, REL_FAILURE);
             context.yield();
-            if (clientSession != null) {
-                try {
-                    if (clientSession.hasActiveTransaction()) {
-                        clientSession.abortTransaction();
-                    }
-                    clientSession.close();
-                } catch (Exception ee) {
-                    getLogger().warn("Cannot rollback client session", ee); // (but no further action)
-                }
-            }
         }
     }
 
--- a/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
+++ b/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
@@ -57,7 +57,6 @@ import org.bson.codecs.DecoderContext;
 import org.bson.conversions.Bson;
 import org.bson.json.JsonReader;
 
-import java.io.ByteArrayOutputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.nio.charset.Charset;
@@ -80,7 +79,7 @@ public class PutMongoBulkOperations extends AbstractMongoProcessor {
     static final PropertyDescriptor ORDERED = new PropertyDescriptor.Builder()
             .name("Ordered")
             .expressionLanguageSupported(ExpressionLanguageScope.NONE)
-            .description("Ordered execution of bulk-writes & break on error - otherwise arbitrary order & continue on error")
+            .description("Ordered execution of bulk-writes and break on error - otherwise arbitrary order and continue on error")
             .required(false)
             .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
             .allowableValues("true", "false")
@@ -145,15 +144,12 @@ public class PutMongoBulkOperations extends AbstractMongoProcessor {
         ClientSession clientSession = null;
         try {
             final MongoCollection<Document> collection = getCollection(context, flowFile).withWriteConcern(writeConcern);
-            // Read the contents of the FlowFile into a byte array
-            ByteArrayOutputStream os = new ByteArrayOutputStream();
-            session.exportTo(flowFile, os);
 
             // parse
             final BsonArrayCodec arrayCodec = new BsonArrayCodec();
             final DecoderContext decoderContext = DecoderContext.builder().build();
             final BsonArray updateItems;
-            try (final Reader reader = new InputStreamReader(session.read(flowFile))) {
+            try (final Reader reader = new InputStreamReader(session.read(flowFile), charset)) {
                 updateItems = arrayCodec.decode(new JsonReader(reader), decoderContext);
             }
 
@@ -199,15 +195,15 @@ public class PutMongoBulkOperations extends AbstractMongoProcessor {
             getLogger().error("Failed to bulk-update {} into MongoDB", flowFile, e);
             session.transfer(flowFile, REL_FAILURE);
             context.yield();
-            try {
-                if (clientSession != null) {
+            if (clientSession != null) {
+                try {
                     if (clientSession.hasActiveTransaction()) {
                         clientSession.abortTransaction();
                     }
                     clientSession.close();
+                } catch (Exception ee) {
+                    getLogger().warn("Cannot rollback client session", ee); // (but no further action)
                 }
-            } catch (Exception ee) {
-                getLogger().warn("Cannot rollback client session", ee); // (but no further action)
             }
         }
     }
--- a/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
+++ b/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
@@ -1,0 +1,320 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.mongodb;
+
+import com.mongodb.WriteConcern;
+import com.mongodb.client.ClientSession;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.model.BulkWriteOptions;
+import com.mongodb.client.model.Collation;
+import com.mongodb.client.model.CollationAlternate;
+import com.mongodb.client.model.CollationCaseFirst;
+import com.mongodb.client.model.CollationMaxVariable;
+import com.mongodb.client.model.CollationStrength;
+import com.mongodb.client.model.DeleteManyModel;
+import com.mongodb.client.model.DeleteOneModel;
+import com.mongodb.client.model.DeleteOptions;
+import com.mongodb.client.model.InsertOneModel;
+import com.mongodb.client.model.ReplaceOneModel;
+import com.mongodb.client.model.ReplaceOptions;
+import com.mongodb.client.model.UpdateManyModel;
+import com.mongodb.client.model.UpdateOneModel;
+import com.mongodb.client.model.UpdateOptions;
+import com.mongodb.client.model.WriteModel;
+import org.apache.nifi.annotation.behavior.InputRequirement;
+import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
+import org.apache.nifi.annotation.behavior.SystemResource;
+import org.apache.nifi.annotation.behavior.SystemResourceConsideration;
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.expression.ExpressionLanguageScope;
+import org.apache.nifi.flowfile.FlowFile;
+import org.apache.nifi.processor.ProcessContext;
+import org.apache.nifi.processor.ProcessSession;
+import org.apache.nifi.processor.Relationship;
+import org.apache.nifi.processor.exception.ProcessException;
+import org.apache.nifi.processor.util.StandardValidators;
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+import org.bson.Document;
+import org.bson.codecs.BsonArrayCodec;
+import org.bson.codecs.DecoderContext;
+import org.bson.conversions.Bson;
+import org.bson.json.JsonReader;
+
+import java.io.ByteArrayOutputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+@Tags({ "mongodb", "insert", "update", "write", "put", "bulk" })
+@InputRequirement(Requirement.INPUT_REQUIRED)
+@CapabilityDescription("Writes the contents of a FlowFile to MongoDB as bulk-update")
+@SystemResourceConsideration(resource = SystemResource.MEMORY)
+public class PutMongoBulkOperations extends AbstractMongoProcessor {
+    static final Relationship REL_SUCCESS = new Relationship.Builder().name("success")
+            .description("All FlowFiles that are written to MongoDB are routed to this relationship").build();
+    static final Relationship REL_FAILURE = new Relationship.Builder().name("failure")
+            .description("All FlowFiles that cannot be written to MongoDB are routed to this relationship").build();
+
+    static final PropertyDescriptor ORDERED = new PropertyDescriptor.Builder()
+            .name("Ordered")
+            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+            .description("Ordered execution of bulk-writes & break on error - otherwise arbitrary order & continue on error")
+            .required(false)
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .allowableValues("true", "false")
+            .defaultValue("true")
+            .build();
+
+    static final PropertyDescriptor TRANSACTIONS_ENABLED = new PropertyDescriptor.Builder()
+            .name("Transactions Enabled")
+            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+            .description("Run all actions in one MongoDB transaction")
+            .required(false)
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .build();
+
+    static final PropertyDescriptor CHARACTER_SET = new PropertyDescriptor.Builder()
+        .name("Character Set")
+        .description("The Character Set in which the data is encoded")
+        .required(true)
+        .addValidator(StandardValidators.CHARACTER_SET_VALIDATOR)
+        .defaultValue("UTF-8")
+        .build();
+
+    private final static Set<Relationship> relationships;
+    private final static List<PropertyDescriptor> propertyDescriptors;
+
+    static {
+        List<PropertyDescriptor> _propertyDescriptors = new ArrayList<>();
+        _propertyDescriptors.addAll(descriptors);
+        _propertyDescriptors.add(ORDERED);
+        _propertyDescriptors.add(TRANSACTIONS_ENABLED);
+        _propertyDescriptors.add(CHARACTER_SET);
+        propertyDescriptors = Collections.unmodifiableList(_propertyDescriptors);
+
+        final Set<Relationship> _relationships = new HashSet<>();
+        _relationships.add(REL_SUCCESS);
+        _relationships.add(REL_FAILURE);
+        relationships = Collections.unmodifiableSet(_relationships);
+    }
+
+    @Override
+    public Set<Relationship> getRelationships() {
+        return relationships;
+    }
+
+    @Override
+    public List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+        return propertyDescriptors;
+    }
+
+    @Override
+    public void onTrigger(final ProcessContext context, final ProcessSession session) throws ProcessException {
+        final FlowFile flowFile = session.get();
+        if (null == flowFile) {
+            return;
+        }
+
+        final Charset charset = Charset.forName(context.getProperty(CHARACTER_SET).getValue());
+        final WriteConcern writeConcern = clientService.getWriteConcern();
+
+        ClientSession clientSession = null;
+        try {
+            final MongoCollection<Document> collection = getCollection(context, flowFile).withWriteConcern(writeConcern);
+            // Read the contents of the FlowFile into a byte array
+            ByteArrayOutputStream os = new ByteArrayOutputStream();
+            session.exportTo(flowFile, os);
+
+            // parse
+            final BsonArrayCodec arrayCodec = new BsonArrayCodec();
+            final DecoderContext decoderContext = DecoderContext.builder().build();
+            final BsonArray updateItems;
+            try (final Reader reader = new InputStreamReader(session.read(flowFile))) {
+                updateItems = arrayCodec.decode(new JsonReader(reader), decoderContext);
+            }
+
+            List<WriteModel<Document>> updateModels = new ArrayList<>();
+            for (Object item : updateItems) {
+                final BsonDocument updateItem = (BsonDocument) item;
+                if (updateItem.keySet().size() != 1) {
+                    getLogger().error("Invalid bulk-update in {}: more than one type given {}", flowFile, String.join(", ", updateItem.keySet()));
+                    session.transfer(flowFile, REL_FAILURE);
+                    context.yield();
+                    return;
+                }
+                final WriteModel<Document> writeModel = getWriteModel(context, session, flowFile, updateItem);
+                if (null == writeModel) {
+                    getLogger().error("Invalid bulk-update in {}: invalid update type {}", flowFile, getUpdateType(updateItem));
+                    session.transfer(flowFile, REL_FAILURE);
+                    context.yield();
+                    return;
+                }
+                updateModels.add(writeModel);
+            }
+
+            if (context.getProperty(TRANSACTIONS_ENABLED).asBoolean()) {
+                clientSession = clientService.startSession();
+                clientSession.startTransaction();
+                // now run this w/in a transaction
+                collection.bulkWrite(clientSession, updateModels, (new BulkWriteOptions().ordered(context.getProperty(ORDERED).asBoolean())));
+            } else {
+                collection.bulkWrite(updateModels, (new BulkWriteOptions().ordered(context.getProperty(ORDERED).asBoolean())));
+            }
+            getLogger().info("bulk-updated {} into MongoDB", flowFile);
+
+            session.getProvenanceReporter().send(flowFile, getURI(context));
+            session.transfer(flowFile, REL_SUCCESS);
+
+            if (clientSession != null) {
+                if (clientSession.hasActiveTransaction()) {
+                    clientSession.commitTransaction();
+                }
+                clientSession.close();
+            }
+        } catch (Exception e) {
+            getLogger().error("Failed to bulk-update {} into MongoDB", flowFile, e);
+            session.transfer(flowFile, REL_FAILURE);
+            context.yield();
+            try {
+                if (clientSession != null) {
+                    if (clientSession.hasActiveTransaction()) {
+                        clientSession.abortTransaction();
+                    }
+                    clientSession.close();
+                }
+            } catch (Exception ee) {
+                getLogger().warn("Cannot rollback client session", ee); // (but no further action)
+            }
+        }
+    }
+
+    private WriteModel<Document> getWriteModel(ProcessContext context, ProcessSession session, FlowFile flowFile, BsonDocument updateItem) {
+        final String updateType = getUpdateType(updateItem);
+        final BsonDocument updateSpec = (BsonDocument) updateItem.get(updateType);
+        final WriteModel<Document> writeModel;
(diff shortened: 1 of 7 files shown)


nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulk.java
```diff
@@ -0,0 +1,353 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+@EventDriven
+@Tags({ "mongodb", "insert", "update", "write", "put", "bulk" })
+@InputRequirement(Requirement.INPUT_REQUIRED)
+@CapabilityDescription("Writes the contents of a FlowFile to MongoDB as bulk-update")
+@SystemResourceConsideration(resource = SystemResource.MEMORY)
+public class PutMongoBulk extends AbstractMongoProcessor {
```
MikeThomsen, line 73: Two questions:

1) Do we need a minimum driver version to use this?
2) Have you considered breaking this functionality up so it can efficiently use the bulkwrite api with the record api for bulk ingestion as well as a generic bulk write operation processor?

sebastianrothbucher: Hi Mike,
thx for reviewing!
re 1.: I tested with the Mongo driver bundled with 1.13 at the time and it worked; I just had to do some adjustments for the driver bundled with 1.20-SNAPSHOT. Concerning operations: all Mongo 4.2+ (i.e. all active versions) support bulkWrite. I'd say: we should be fine
re 2.: I deliberately did not use records to allow freeform updates with $set, $inc, etc. as operations. Of course, you could define an avro schema for this, but I think it's a stretch. Original intention was to use it when you need more fine-grained control of updates vis-a-vis PutMongoRecord. Does that make sense? Could of course xpand the additional description HTML...

MikeThomsen: #2 makes sense. Is the bulkwrite api supposed to be faster and more flexible, or just more flexible? Based on your response, if the former then that has implications for PutMongoRecord too.

sebastianrothbucher: putMongoRecord uses the same bulkWrite API under the hood, so the PubMongoBulk would be more flexible and equally fast compared to PutMongoRecord

sebastianrothbucher:
anything I can do to get the ball rolling again?

MikeThomsen:
@sebastianrothbucher yeah, I'm adding this to my TODO to get a review going.

exceptionfactory (changes requested):
Thanks for the contribution @sebastianrothbucher. I noticed some implementation concerns, but before proceeding to review the code details, it would be helpful to evaluate the scope of this Processor.

Although the documentation makes it clear that the input JSON needs to match MongoDB requirements, the generalized nature of the Processor seems like it could be confusing. The `Put` prefix generally signifies some kind of insert operation, but the implementation allows insert, update, replace, or delete. Although having a general Processor can be helpful, the scope seems to open-ended in this situation. One option is renaming the Processor to something like `ExecuteMongo`, along the lines of `ExecuteSQL`. On the other hand, the conditionals inside the `onTrigger` method perform significantly different operations. Is the goal to support a combination of insert, update, and delete operations in a single execution? This is generally counter to most other Processors.

The Mongo Client Service should also be the preferred configuration method.

With that background, I think it would be helpful to refine the scope and target use case for this Processor.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulk.java
```diff
@@ -0,0 +1,353 @@
...
+            .description("All FlowFiles that cannot be written to MongoDB are routed to this relationship").build();
+
+    static final PropertyDescriptor ORDERED = new PropertyDescriptor.Builder()
+            .name("Ordered")
+            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+            .description("Ordered execution of bulk-writes & break on error - otherwise arbitrary order & continue on error")
+            .required(false)
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .allowableValues("true", "false")
+            .defaultValue("false")
```
MikeThomsen, line 86: This should be true because [the documentation](https://www.mongodb.com/docs/manual/reference/method/db.collection.bulkWrite/) says that the default behavior is `true`. We don't want to invert expectations.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulk.java
```diff
@@ -0,0 +1,353 @@
...
+            .build();
+
+    static final PropertyDescriptor USE_TRANSACTION = new PropertyDescriptor.Builder()
+            .name("Use transaction")
+            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+            .description("Run all actions in one MongoDB transaction - does NOT work with a client service so far")
```
MikeThomsen, line 92: Can you elaborate on the caveat about it not working with the client service?

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulk.java
```diff
@@ -0,0 +1,353 @@
...
+        final ComponentLog logger = getLogger();
+
+        final Charset charset = Charset.forName(context.getProperty(CHARACTER_SET).getValue());
+        final WriteConcern writeConcern = getWriteConcern(context);
+
+        ClientSession clientSession = null;
+        try {
+            final MongoCollection<Document> collection = getCollection(context, flowFile).withWriteConcern(writeConcern);
+            // Read the contents of the FlowFile into a byte array
+            final byte[] content = new byte[(int) flowFile.getSize()];
```
MikeThomsen, line 151: This is a cleaner way to do it:

```
ByteArrayOutputStream os = new ByteArrayOutputStream();
session.exportTo(ff, os);
```

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulk.java
```diff
@@ -0,0 +1,353 @@
...
+                final BsonDocument updateItem = (BsonDocument) item;
+                if (updateItem.keySet().size() != 1) {
+                    logger.error("Invalid bulk-update: more than one type given {}", String.join(", ", updateItem.keySet()));
+                    session.transfer(flowFile, REL_FAILURE);
+                    context.yield();
+                    return;
+                }
+                final String updateType = updateItem.keySet().iterator().next();
+                final BsonDocument updateSpec = (BsonDocument) updateItem.get(updateType);
+                if ("insertOne".equals(updateType)) {
```
MikeThomsen, line 168: You can simplify this if/else tree by using a [WriteModel](https://mongodb.github.io/mongo-java-driver/3.7/javadoc/com/mongodb/client/model/WriteModel.html) and adding the `updateModels.add` call after the else-ifs/else.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulk.java
```diff
@@ -0,0 +1,353 @@
...
+        }
+
+        final ComponentLog logger = getLogger();
+
+        final Charset charset = Charset.forName(context.getProperty(CHARACTER_SET).getValue());
+        final WriteConcern writeConcern = getWriteConcern(context);
+
+        ClientSession clientSession = null;
+        try {
+            final MongoCollection<Document> collection = getCollection(context, flowFile).withWriteConcern(writeConcern);
```
MikeThomsen, line 149: If you're relying on the connection configuration properties here, that has been deprecated in 1.x and removed in 2.x. A client controller service will be required.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/test/java/org/apache/nifi/processors/mongodb/PutMongoBulkIT.java
```diff
@@ -0,0 +1,154 @@
...
+ */
+package org.apache.nifi.processors.mongodb;
+
+import org.apache.nifi.util.TestRunner;
+import org.bson.Document;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.Assert.assertEquals;
```
MikeThomsen, line 25: These are JUnit 4 and need to be replaced with the JUnit 5 versions.

MikeThomsen (changes requested):
Sorry it took a while to get back to you. Left you with a bunch of feedback. Good first pass at this processor.

exceptionfactory:
> @exceptionfactory
> 
> > On the other hand, the conditionals inside the onTrigger method perform significantly different operations. Is the goal to support a combination of insert, update, and delete operations in a single execution? This is generally counter to most other Processors.
> 
> We already have something similar in the Elasticsearch package where you can do a very complex set of bulk operations from a single flowfile with PutElasticsearchRecord.

Thanks for the reference @MikeThomsen, that is a helpful comparison. With that in mind, does it make sense to align this Processor to something more record-oriented, as opposed to requiring JSON?

MikeThomsen:
> Thanks for the reference @MikeThomsen, that is a helpful comparison. With that in mind, does it make sense to align this Processor to something more record-oriented, as opposed to requiring JSON?

No, I think it's good the way it is, and it doesn't bill itself as a record-aware processor. Maybe adding "Operations" to the end of it to further clarify that it is calling the BulkWrite API, but as-is I think it should fit a useful Mongo niche.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulk.java
```diff
@@ -0,0 +1,353 @@
...
+            session.transfer(flowFile, REL_SUCCESS);
+
+            if (clientSession != null) {
+                if (clientSession.hasActiveTransaction()) {
+                    clientSession.commitTransaction();
+                }
+                clientSession.close();
+            }
+        } catch (Exception e) {
+            logger.error("Failed to bulk-update {} into MongoDB due to {}", new Object[] {flowFile, e}, e);
```
exceptionfactory, line 228: The `Object[]` array wrapper is not necessary, and the `due to {}` convention is not necessary either, so it should be removed.
```suggestion
            logger.error("Failed to bulk-update {} into MongoDB", flowFile, e);
```

exceptionfactory (changes requested):
Including `Operations` in the Processor name sounds like it would be helpful.

The comments about issues with the MongoDB Client Service need to be resolved, in addition to to the JUnit 4 test references. If these issues can be addressed, we should be able to move forward with the review.

sebastianrothbucher:
thanks both of you, I'll take care of it

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulk.java
sebastianrothbucher: This needs extension to the MongoDBControllerService API. I did that now - but I can just as well live w/ not having this option. I'm using the processor daily without transactions and it's still valuable

sebastianrothbucher:
diff processors failing on otherwise identical code (I did rebase before pushing); got no windows box unfortunately to just debug in

joewitt:
@exceptionfactory Seems like the build checks out fine.  Was that the last concern?

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+@EventDriven
```
exceptionfactory, line 67: This annotation is deprecated and now removed from the main branch, so it should be removed.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+            .description("Ordered execution of bulk-writes & break on error - otherwise arbitrary order & continue on error")
+            .required(false)
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .allowableValues("true", "false")
+            .defaultValue("true")
+            .build();
+
+    static final PropertyDescriptor USE_TRANSACTION = new PropertyDescriptor.Builder()
+            .name("Use transaction")
```
exceptionfactory, line 89: Recommend renaming to `Transactions Enabled`:
```suggestion
    static final PropertyDescriptor TRANSACTIONS_ENABLED = new PropertyDescriptor.Builder()
            .name("Transactions Enabled")
```

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+@CapabilityDescription("Writes the contents of a FlowFile to MongoDB as bulk-update")
+@SystemResourceConsideration(resource = SystemResource.MEMORY)
+public class PutMongoBulkOperations extends AbstractMongoProcessor {
+    static final Relationship REL_SUCCESS = new Relationship.Builder().name("success")
+            .description("All FlowFiles that are written to MongoDB are routed to this relationship").build();
+    static final Relationship REL_FAILURE = new Relationship.Builder().name("failure")
+            .description("All FlowFiles that cannot be written to MongoDB are routed to this relationship").build();
+
+    static final PropertyDescriptor ORDERED = new PropertyDescriptor.Builder()
+            .name("Ordered")
```
exceptionfactory, line 79: It would be helpful to change the name of this property to make the purpose clearer. What do you think of `Execution Order`, with an enumeration of values, `ORDERED` and `RANDOM`?

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+
+        ClientSession clientSession = null;
+        try {
+            final MongoCollection<Document> collection = getCollection(context, flowFile).withWriteConcern(writeConcern);
+            // Read the contents of the FlowFile into a byte array
+            ByteArrayOutputStream os = new ByteArrayOutputStream();
+            session.exportTo(flowFile, os);
+
+            // parse
+            final BsonArray updateItems = BsonArray.parse(os.toString(charset));
```
exceptionfactory, line 153: Instead of exporting the FlowFile to a string, the `BsonArrayCodec`has a `decode` method that takes a BsonReader. There is a little bit more involved to set it up, but it would allow reading from a `java.io.Reader`, which could be an instance of `InputStreamReader`. Although the parsed objects will be loaded into memory, it would avoid reading the FlowFile content as a string.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+                    final ReplaceOptions options = parseReplaceOptions(updateSpec);
+                    writeModel = new ReplaceOneModel<>((BsonDocument) updateSpec.get("filter"),
+                            toBsonDocument((BsonDocument) updateSpec.get("replacement")), options);
+                } else if ("deleteOne".equals(updateType)) {
+                    final DeleteOptions options = parseDeleteOptions(updateSpec);
+                    writeModel = new DeleteOneModel<>((BsonDocument) updateSpec.get("filter"), options);
+                } else  if ("deleteMany".equals(updateType)) {
+                    final DeleteOptions options = parseDeleteOptions(updateSpec);
+                    writeModel = new DeleteManyModel<>((BsonDocument) updateSpec.get("filter"), options);
+                } else {
```
exceptionfactory, line 185: It looks like this conditional logic could be broken out to a separate method that returns a `WriteModel`, or `null` on failure. That could make the `onTrigger` method easier to maintain.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+                    logger.error("Invalid bulk-update: invalid update type {}", updateType);
```
exceptionfactory, line 186: It would be helpful to include the FlowFile reference for this error.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+            session.exportTo(flowFile, os);
+
+            // parse
+            final BsonArray updateItems = BsonArray.parse(os.toString(charset));
+
+            List<WriteModel<Document>> updateModels = new ArrayList<>();
+            for (Object item : updateItems) {
+                final BsonDocument updateItem = (BsonDocument) item;
+                if (updateItem.keySet().size() != 1) {
+                    logger.error("Invalid bulk-update: more than one type given {}", String.join(", ", updateItem.keySet()));
```
exceptionfactory, line 159: It would be helpful to include the FlowFile reference for this error.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+            if (context.getProperty(USE_TRANSACTION).asBoolean()) {
+                clientSession = clientService.startSession();
+                clientSession.startTransaction();
+                // now run this w/in a transaction
+                collection.bulkWrite(clientSession, updateModels, (new BulkWriteOptions().ordered(context.getProperty(ORDERED).asBoolean())));
+            } else {
+                collection.bulkWrite(updateModels, (new BulkWriteOptions().ordered(context.getProperty(ORDERED).asBoolean())));
+            }
+            logger.info("bulk-updated {} into MongoDB", flowFile);
+            // (could also return the result again as JSON - mostly not needed afaik)
```
exceptionfactory, line 203: This comment could be removed.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+            if (clientSession != null) {
+                if (clientSession.hasActiveTransaction()) {
+                    clientSession.commitTransaction();
+                }
+                clientSession.close();
+            }
+        } catch (Exception e) {
+            logger.error("Failed to bulk-update {} into MongoDB", flowFile, e);
+            session.transfer(flowFile, REL_FAILURE);
+            context.yield();
```
exceptionfactory, line 217: Is there a reason for this yield?

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+            try {
+                if (clientSession != null) {
+                    if (clientSession.hasActiveTransaction()) {
+                        clientSession.abortTransaction();
+                    }
+                    clientSession.close();
+                }
+            } catch (Exception ee) {
+                logger.error("Cannot rollback client session", ee); // (but no further action)
```
exceptionfactory, line 226: Recommend changing this to a warning since no further action is possible.

exceptionfactory (changes requested):
@joewitt and @sebastianrothbucher I went through the code in more detail and noted several recommendations. After addressing these issues, I think this should be ready to go forward.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: this one time, I'd stick with naming identical to the API - https://www.mongodb.com/docs/manual/reference/method/db.collection.bulkWrite/

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,304 @@
...
+            // parse
+            final BsonArray updateItems = BsonArray.parse(os.toString(charset));
+
+            List<WriteModel<Document>> updateModels = new ArrayList<>();
+            for (Object item : updateItems) {
+                final BsonDocument updateItem = (BsonDocument) item;
+                if (updateItem.keySet().size() != 1) {
+                    logger.error("Invalid bulk-update: more than one type given {}", String.join(", ", updateItem.keySet()));
+                    session.transfer(flowFile, REL_FAILURE);
+                    context.yield();
```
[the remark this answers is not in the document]
sebastianrothbucher: Same as PutMongo - and I think it makes sense to back off on error

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: Same as PutMongo - and I think it makes sense to back off on error

sebastianrothbucher:
Thanks! Did address them all (or at least leave an excuse in the comment ;-) )

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
exceptionfactory: Thanks for the reference, that's a good point, aligning with the MongoDB terminology seems reasonable in this case.

sebastianrothbucher:
Test is due to newer mongo - working on it

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    @Override
+    public void onTrigger(final ProcessContext context, final ProcessSession session) throws ProcessException {
+        final FlowFile flowFile = session.get();
+        if (null == flowFile) {
+            return;
+        }
+
+        final Charset charset = Charset.forName(context.getProperty(CHARACTER_SET).getValue());
```
exceptionfactory, line 142: The `charset` does not appear to be used. It looks like it should be passed to the `InputStreamReader` on construction.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,320 @@
...
+        final WriteConcern writeConcern = clientService.getWriteConcern();
+
+        ClientSession clientSession = null;
+        try {
+            final MongoCollection<Document> collection = getCollection(context, flowFile).withWriteConcern(writeConcern);
+            // Read the contents of the FlowFile into a byte array
+            ByteArrayOutputStream os = new ByteArrayOutputStream();
+            session.exportTo(flowFile, os);
```
exceptionfactory, line 150: This export operation is not required and should be removed.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,320 @@
...
+        _propertyDescriptors.addAll(descriptors);
+        _propertyDescriptors.add(ORDERED);
+        _propertyDescriptors.add(TRANSACTIONS_ENABLED);
+        _propertyDescriptors.add(CHARACTER_SET);
+        propertyDescriptors = Collections.unmodifiableList(_propertyDescriptors);
+
+        final Set<Relationship> _relationships = new HashSet<>();
+        _relationships.add(REL_SUCCESS);
+        _relationships.add(REL_FAILURE);
+        relationships = Collections.unmodifiableSet(_relationships);
```
exceptionfactory, line 122: These declarations can be streamlined and replaced with `List.of()` and `Set.of()`

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    @Override
+    public Set<Relationship> getRelationships() {
+        return relationships;
+    }
+
+    @Override
+    public List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+        return propertyDescriptors;
```
exceptionfactory, line 132: This appears to be missing the property descriptors from the parent class, so the Client Service is not defined.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,320 @@
...
+            .name("Ordered")
+            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+            .description("Ordered execution of bulk-writes & break on error - otherwise arbitrary order & continue on error")
+            .required(false)
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .allowableValues("true", "false")
+            .defaultValue("true")
+            .build();
+
+    static final PropertyDescriptor TRANSACTIONS_ENABLED = new PropertyDescriptor.Builder()
```
exceptionfactory, line 90: Reviewing the implementation, is there value in making this a configurable property? On the one hand, removing the property avoids changing the client service interface and introducing the new method. On the other hand, should transactions always be used?

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,320 @@
...
+public class PutMongoBulkOperations extends AbstractMongoProcessor {
+    static final Relationship REL_SUCCESS = new Relationship.Builder().name("success")
+            .description("All FlowFiles that are written to MongoDB are routed to this relationship").build();
+    static final Relationship REL_FAILURE = new Relationship.Builder().name("failure")
+            .description("All FlowFiles that cannot be written to MongoDB are routed to this relationship").build();
+
+    static final PropertyDescriptor ORDERED = new PropertyDescriptor.Builder()
+            .name("Ordered")
+            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+            .description("Ordered execution of bulk-writes & break on error - otherwise arbitrary order & continue on error")
```
exceptionfactory, line 83: Use of the ampersand character should be avoided.
```suggestion
            .description("Ordered execution of bulk-writes and break on error - otherwise arbitrary order and continue on error")
```

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
```diff
@@ -0,0 +1,320 @@
...
+            try {
+                if (clientSession != null) {
+                    if (clientSession.hasActiveTransaction()) {
+                        clientSession.abortTransaction();
+                    }
+                    clientSession.close();
+                }
+            } catch (Exception ee) {
+                getLogger().warn("Cannot rollback client session", ee); // (but no further action)
+            }
```
exceptionfactory, line 211: It looks like the try-catch block should be moved inside the conditional for when the clientSession is not null.

exceptionfactory (changes requested):
Thanks for your patience on this pull request @sebastianrothbucher. Unfortunately there are still some implementation problems that I noted in the latest set of comments. They appear to be missed in the course of refactoring, which can happen.

To make it easier to review subsequent changes, please put new changes in a new commit, even when rebasing, so that it is easier to follow the changes from one set of comments to another.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: I did test ;-) and in fact, line 11

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: I did what PutMongo also did to stay kinda in sync; for first, shorthand does not work b/c of 113; to me, it's even uglier to change style midway. Up2u, I can do it either way

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: thx & sorry!

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: thx & sorry!

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: Not really, I almost never use them tbh b/c most of the time, the document is the scope of work. Tried to stick to the driver here also

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: bmg

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
sebastianrothbucher: ok

sebastianrothbucher:
thanks for checking!

sebastianrothbucher:
2nd failure is a Heisenbug ;-) 
think we're good for this Processor

exceptionfactory:
> 2nd failure is a Heisenbug ;-) think we're good for this Processor

Recent build changes introduced automated running of integration tests, and it looks like it is failing on the new IT:

```
Error:  Failures: 
Error:    PutMongoBulkOperationsIT.testBulkWriteInsert:65 expected: <3> but was: <0>
```
A number of the existing integration tests are being skipped because they are not in a position to run in an automated workflow, or they are otherwise unreliable.

In this case, the new integration test needs to be reliable, or it should not be included. Does it work for you on a local build? GitHub runners are less powerful, which can introduce certain unexpected behaviors, but this is something that will need to be addressed.

sebastianrothbucher:
I still have to look. I ran tests before each push - and they just worked. Also from the error messages - it seems like sth gets stuck due to load. Let me get back2u

sebastianrothbucher:
I don't think it was sth I did, specifically; it's just that Mongo might not yet have committed the deletion of the DB in question. I nonetheless did two things: 1.) use different test docs in different tests (which should not be necessary but gives additoinal security). and 2.) add more deletion and an explicit close to the Mongo write tests - which hopefully got this from 0.1% to 0.001% or sth

I re-ran the full test suite - and I have some other spurious errors. One in GridFS I totally cannot explain (except somehow the CI overwrites the mongo docker image - and we use another MongoDB on Github CI; just didn't see it).

The other issue is in GetMongoIT. We seem to overwrite flowfile attrs with "environment variables". Looks to me the following change makes it a little more stable. Didn't include it in the commit; maybe still useful

```
diff --git a/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/test/java/org/apache/nifi/processors/mongodb/GetMongoIT.java b/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/test/java/org/apache/nifi/processors/mongodb/GetMongoIT.java
index 87a0ca3d56..58e9adea6f 100644
--- a/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/test/java/org/apache/nifi/processors/mongodb/GetMongoIT.java
+++ b/nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/test/java/org/apache/nifi/processors/mongodb/GetMongoIT.java
@@ -502,10 +502,9 @@ public class GetMongoIT extends AbstractMongoIT {
             db.getCollection(collections[x])
                 .insertOne(new Document().append("msg", "Hello, World"));
 
-            Map<String, String> attrs = new HashMap<>();
-            attrs.put("db", dbs[x]);
-            attrs.put("collection", collections[x]);
-            runner.enqueue(query, attrs);
+            runner.setEnvironmentVariableValue("db", dbs[x]);
+            runner.setEnvironmentVariableValue("collection", collections[x]);
+            runner.enqueue(query);
             runner.run();
 
             db.drop();
```

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
exceptionfactory: Thanks for the response. As this is a new Processor, this is a good opportunity to improve the convention, so moving to `List.of()` and `Set.of()` declarations, as opposed to the static initializer and underscored collection variables, would be helpful.

nifi-nar-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/PutMongoBulkOperations.java
exceptionfactory: Thanks for the reply. If you don't think it is necessary to support transactions for now, it seems like it would be better to remove this property, and the method from MongoDBClientService. Removing the option would simplify both the configuration and implementation. It looks like it wouldn't be difficult to add later if needed, but if you don't see a common need, then this might be a case of "less is more" for the initial version.

exceptionfactory (commented):
Thanks for the latest round of updates, I noted two more comments, and based on your response and updates, I think this should be ready to go.

sebastianrothbucher:
all right - addressed both final issues

exceptionfactory (approved):
Thanks again for your patience and perseverance on this new Processor @sebastianrothbucher. After one more review, I noticed that the `Ordered` property can be set as `required` because it should always have a value, but I will make that adjustment when merging. The latest version looks good, thanks again! +1 merging

PR #6935, opened 2023-02-08
NIFI-11151: Improving code reusability of DBCP services

Lehel44:
# Summary

[NIFI-11151](https://issues.apache.org/jira/browse/NIFI-11151)

# Tracking

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-11151) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 8
  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for the contribution @Lehel44. The separate builder for the `BasicDataSource` configuration looks useful, but taking a look at the `AbstractDBCPConnectionPool` as a whole, it does not seem to be a good candidate for further changes and extension. It has some features that should not be incorporated in any new components, such as the sensitive dynamic property support using the `SENSITIVE.` prefix. Instead, it seems better to introduce a new component hierarchy as part of any new Connection Pool services, which may provide opportunities for refactoring after that implementation.

Lehel44:
Thanks for the related reviews to this PR @exceptionfactory @turcsanyip. With a more general approach I refactored the AbstractDBCPConnectionPool to be more generic and flexible. 
- I moved the PropertyDescriptors to a different class to avoid unnecessary inheritance.
- I kept parts that is currently being reused and can be reused in the abstract super class
- Moved domain specific code to the extending subclasses
- Created DataSourceConfiguration to decouple PropertyDescriptors and DataSource configuration.

nifi-nar-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/utils/DefaultDataSourceValues.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+package org.apache.nifi.dbcp.utils;
+
+import org.apache.commons.dbcp2.BasicDataSource;
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+
+public enum DefaultDataSourceValues {
+
+
+    MAX_WAIT_MILLIS("500", "500 millis"),
```
turcsanyip, line 25:
```suggestion
    MAX_WAIT_TIME("500", "500 millis"),
```
Other constants (`MAX_CONN_LIFETIME`, `EVICTION_RUN_PERIOD`) do not have `_MILLIS` suffix which I think is correct because the time period can be defined in different time units.

nifi-nar-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/utils/DefaultDataSourceValues.java
```diff
@@ -0,0 +1,74 @@
...
+    MAX_CONN_LIFETIME("-1", "-1"),
+    /**
+     * Copied from {@link GenericObjectPoolConfig#DEFAULT_TIME_BETWEEN_EVICTION_RUNS_MILLIS} in Commons-DBCP 2.7.0
+     */
+    EVICTION_RUN_PERIOD("-1", "-1"),
+    /**
+     * Copied from {@link GenericObjectPoolConfig#DEFAULT_MIN_EVICTABLE_IDLE_TIME_MILLIS} in Commons-DBCP 2.7.0
+     * and converted from 1800000L to "1800000 millis" to "30 mins"
+     */
+    MIN_EVICTABLE_IDLE_TIME_MILLIS("1800000", "30 mins"),
```
turcsanyip, line 48:
```suggestion
    MIN_EVICTABLE_IDLE_TIME("1800000", "30 mins"),
```
Other constants (`MAX_CONN_LIFETIME`, `EVICTION_RUN_PERIOD`) do not have `_MILLIS` suffix which I think is correct because the time period can be defined in different time units.

nifi-nar-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/utils/DefaultDataSourceValues.java
turcsanyip, line 25: I think it is not necessary to define the same value in 2 formats because the millisec value could be calculated from the property value using `FormatUtils.getTimeDuration()` (the -1 special value needs to be handled too).

turcsanyip (changes requested):
Thanks @Lehel44 for this reworked versions of the PR!
The general concept looks good to me now.
I just added some initial comments but have not tried/tested the code with flows.

nifi-nar-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/AbstractDBCPConnectionPool.java
```diff
@@ -365,168 +121,151 @@ public List<ConfigVerificationResult> verify(final ConfigurationContext context,
...
-        if (validationQuery != null && !validationQuery.isEmpty()) {
+    protected DataSourceConfiguration createDataSourceConfiguraton(final ConfigurationContext context) {
+        final String url = context.getProperty(DBCPProperties.DATABASE_URL).getValue();
+        final String user = context.getProperty(DBCPProperties.DB_USER).getValue();
+        final String password = context.getProperty(DBCPProperties.DB_PASSWORD).getValue();
+        final String driverName = context.getProperty(DB_DRIVERNAME).evaluateAttributeExpressions().getValue();
+
+        return new DataSourceConfiguration.Builder(url, driverName, user, password)
+                .build();
+    }
```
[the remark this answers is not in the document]
Lehel44: These four properties are the necessary ones and the optionals are set to default values. I'd assume those won't be modified very often. What do you think about adding a copy constructor to DataSourceConfiguration, so the extended classes could do something like `new DataSourceConfiguration(extendedConfig).Builder().optionalConfigValue1(12)...build()` ?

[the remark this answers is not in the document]
Lehel44: This way we can't override the builder's contructor. Some child classes modifies the url for example which is a required builder field. I think a copy constructor would work.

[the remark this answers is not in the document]
Lehel44: I agree that it is less reusable. The problem with returning with builder is that the mandatory fields can't be changed afterwards. The url is mandatory and is overwritten by several descendant classes. You could override a getUrl method, but I think it would be an overdesign, so I would just make the method plain abstract.

Lehel44:
Thanks for the comments @turcsanyip. I tested the DBCPConnectionPool manually and it worked fine.

nifi-nar-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/AbstractDBCPConnectionPool.java
```diff
@@ -365,168 +119,143 @@ public List<ConfigVerificationResult> verify(final ConfigurationContext context,
...
+            kerberosUser = getKerberosUserByCredentials(context);
         }
         return kerberosUser;
     }
 
-    protected String getUrl(ConfigurationContext context) {
-        return context.getProperty(DATABASE_URL).evaluateAttributeExpressions().getValue();
+    protected KerberosUser getKerberosUserByCredentials(final ConfigurationContext context) {
+        return null;
     }
```
exceptionfactory, line 217: Instead of having this method implementation return `null` in this abstract class, it looks like it could be removed, and the concrete implementation could override the `getKerberosUser()` method.

exceptionfactory (approved):
Thanks for making the adjustments @Lehel44, the latest version looks good from a code perspective.

Future extensions may raise other questions, but these changes should provide a good basis for additional iteration if necessary.

I defer to @turcsanyip for any additional comments and review.

nifi-nar-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/AbstractDBCPConnectionPool.java
```diff
@@ -365,168 +119,136 @@ public List<ConfigVerificationResult> verify(final ConfigurationContext context,
...
-        dataSource.setDriver(getDriver(driverName, dburl));
-        dataSource.setMaxWaitMillis(maxWaitMillis);
-        dataSource.setMaxTotal(maxTotal);
-        dataSource.setMinIdle(minIdle);
-        dataSource.setMaxIdle(maxIdle);
-        dataSource.setMaxConnLifetimeMillis(maxConnLifetimeMillis);
-        dataSource.setTimeBetweenEvictionRunsMillis(timeBetweenEvictionRunsMillis);
-        dataSource.setMinEvictableIdleTimeMillis(minEvictableIdleTimeMillis);
-        dataSource.setSoftMinEvictableIdleTimeMillis(softMinEvictableIdleTimeMillis);
+    protected abstract DataSourceConfiguration getDataSourceConfigurationBuilder(final ConfigurationContext context);
```
turcsanyip, line 158: It does not return a builder but the configuration object itself.
```suggestion
    protected abstract DataSourceConfiguration getDataSourceConfiguration(final ConfigurationContext context);
```

turcsanyip (changes requested):
@Lehel44 Overall it looks good. Tested with PostgreSQL and Snowflake.
Could you please fix the method name I commented?

turcsanyip (approved):
Thanks @Lehel44!
Thanks @exceptionfactory for your review!

+1 LGTM
Merging to main and 1.x

PR #6937, opened 2023-02-09
NIFI-11159 Fixing connections with source having reassigned id

simonbence:
# Summary

[NIFI-11159](https://issues.apache.org/jira/browse/NIFI-11159)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 8
  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -612,17 +612,32 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
         });
     }
 
-    private void removeMissingConnections(final ProcessGroup group, final VersionedProcessGroup proposed, final Map<String, Connection> connectionsByVersionedId) {
+    private void removeMissingConnections(final ProcessGroup group, final VersionedProcessGroup proposed, final Map<String, Connection> connectionsByVersionedId) { // remove ones with same id but different source as well
         final Set<String> connectionsRemoved = new HashSet<>(connectionsByVersionedId.keySet());
+        final Set<String> connectionsRemovedDueToChangingSourceId = new HashSet<>();
 
         for (final VersionedConnection proposedConnection : proposed.getConnections()) {
             connectionsRemoved.remove(proposedConnection.getIdentifier());
+        }
+
+        for (final VersionedConnection proposedConnection : proposed.getConnections()) {
+            if (connectionsByVersionedId.containsKey(proposedConnection.getIdentifier())
+                &&
+                !proposedConnection.getSource().getId().equals(connectionsByVersionedId.get(proposedConnection.getIdentifier()).getSource().getVersionedComponentId())
+            ) {
+                connectionsRemovedDueToChangingSourceId.add(proposedConnection.getIdentifier());
+                connectionsRemoved.add(proposedConnection.getIdentifier());
+            }
         }
 
         for (final String removedVersionedId : connectionsRemoved) {
             final Connection connection = connectionsByVersionedId.get(removedVersionedId);
             LOG.info("Removing {} from {}", connection, group);
             group.removeConnection(connection);
+        }
+
+        for (final String removedVersionedId : connectionsRemovedDueToChangingSourceId) {
+            connectionsByVersionedId.remove(removedVersionedId);
         }
     }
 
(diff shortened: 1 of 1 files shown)


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -667,6 +682,8 @@ private Set<String> updateConnectionDestinations(final ProcessGroup group, final
                 newDestination == null
                 ||
                 (newDestination.getConnectableType() == ConnectableType.OUTPUT_PORT && !newDestination.getProcessGroup().equals(connection.getProcessGroup()))
+                ||
+                (newDestination.getConnectableType() == ConnectableType.INPUT_PORT && !connection.getDestination().getVersionedComponentId().get().equals(proposedConnection.getDestination().getId()) && connection.getDestination().getName().equals(proposedConnection.getDestination().getName()))
```
markap14, line 686: I could be wrong, but I'm pretty sure that this block of code is not actually necessary. We don't need to handle this case specially, as far as I can tell. Unless there's something that I'm missing.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -612,18 +612,33 @@ private void synchronizeControllerServices(final ProcessGroup group, final Versi
...
+        final Set<String> connectionsRemovedDueToChangingSourceId = new HashSet<>();
 
         for (final VersionedConnection proposedConnection : proposed.getConnections()) {
             connectionsRemoved.remove(proposedConnection.getIdentifier());
         }
 
+        for (final VersionedConnection proposedConnection : proposed.getConnections()) {
+            if (connectionsByVersionedId.containsKey(proposedConnection.getIdentifier())
+                &&
+                !proposedConnection.getSource().getId().equals(connectionsByVersionedId.get(proposedConnection.getIdentifier()).getSource().getVersionedComponentId())
```
markap14, line 626: This line appears to be incorrect. IntelliJ is flagging this because the operands for the `equals` method are not compatible - the lefthand side is a `String` while the righthand side is an `Optional<String>`. As a result, this conditional is always going to be `true` (due to the preceding `!`). This means that for every VersionedConnection that is proposed, if the connection exists, we will remove it. So if any connection in the flow has data this will fail. We definitely need to fix this.

I think we can fix this by changing the call to `getVersionedComponentId()` to `getVersionedComponentId().orElse(null)`.
But this is already a very complicated expression. Please extract out appropriate variable names to make this more readable and add some comments to explain what we're doing here. Perhaps:
```
        // Check for any case where there's an existing connection whose ID matches the proposed connection, but whose source doesn't match
        // the proposed source ID. The source of a Connection should never change from one component to another. However, there are cases
        // in which the Versioned Component ID might change, in order to avoid conflicts with sibling Process Groups. In such a case, we must remove
        // the connection and create a new one, since we cannot simply change the source in the same way that we can change the destination.
        final Set<String> connectionsRemovedDueToChangingSourceId = new HashSet<>();
        for (final VersionedConnection proposedConnection : proposed.getConnections()) {
            final Connection existingConnection = connectionsByVersionedId.get(proposedConnection.getIdentifier());
            if (existingConnection != null) {
                final String proposedSourceId = proposedConnection.getSource().getId();
                final String existingSourceId = existingConnection.getSource().getVersionedComponentId().orElse(null);

                if (!Objects.equals(proposedSourceId, existingSourceId)) {
                    connectionsRemovedDueToChangingSourceId.add(proposedConnection.getIdentifier());
                    connectionsRemoved.add(proposedConnection.getIdentifier());
                }
            }
        }
```

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -612,18 +612,33 @@ private void synchronizeControllerServices(final ProcessGroup group, final Versi
         });
     }
 
-    private void removeMissingConnections(final ProcessGroup group, final VersionedProcessGroup proposed, final Map<String, Connection> connectionsByVersionedId) {
+    private void removeMissingConnections(final ProcessGroup group, final VersionedProcessGroup proposed, final Map<String, Connection> connectionsByVersionedId) { // remove ones with same id but different source as well
```
markap14, line 615: I think the comment here was intended as an internal note to self, perhaps? If so, please remove. It exceeds the 200-character line limit, causing checkstyle failures.

markap14:
Thanks @simonbence looks good to me. +1 merged to main

PR #6949, opened 2023-02-14
NIFI-11134: added Labels to Flow Configuration History

markobean:
# Summary

Updated Flow Configuration History to track changes for Labels (added, removed and modified). No unit tests were written; most other auditors do not have unit tests either. However, it was tested/verified in a running nifi.

Request reviewer pay particular attention to the AspectJ annotations in LabelAuditor.java. I used other Auditor classes as a model, but I was unfamiliar with these annotations initially.

[NIFI-11134](https://issues.apache.org/jira/browse/NIFI-11134)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

Installed and started NiFi. Added Label. Verified new entry in the Flow Configuration History corresponding to adding a new Label.
Updated text of Label. Verified new entry in the Flow Configuration History, and confirmed the Value and Previous Value were correct. Repeated this for different updates including an empty value. 
Removed Label. Verified new entry in the Flow Configuration History corresponding to removing the Label.

### Build

  - [X] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/action/Component.java
+++ b/nifi-api/src/main/java/org/apache/nifi/action/Component.java
@@ -36,5 +36,6 @@ public enum Component {
     ParameterProvider,
     AccessPolicy,
     User,
-    UserGroup;
+    UserGroup,
+    Label;
 }
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/audit/LabelAuditor.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/audit/LabelAuditor.java
@@ -40,8 +40,6 @@ import java.util.Objects;
 public class LabelAuditor extends NiFiAuditor {
     private static final Logger logger = LoggerFactory.getLogger(LabelAuditor.class);
 
-    static final String SOURCE_NAME = "LABEL";
-
     /**
      * Audits the creation of a Label.
      *
@@ -94,10 +92,11 @@ public class LabelAuditor extends NiFiAuditor {
                 labelAction.setSourceId(label.getIdentifier());
                 labelAction.setSourceType(Component.Label);
                 labelAction.setOperation(Operation.Configure);
-                labelAction.setSourceName(SOURCE_NAME); // Source Name is a required field for the database but not applicable for a label
+                // Source Name is a required field for the database but not applicable for a label; use UUID to create a unique name
+                labelAction.setSourceName(label.getIdentifier());
 
                 final FlowChangeConfigureDetails actionDetails = new FlowChangeConfigureDetails();
-                actionDetails.setName(SOURCE_NAME);
+                actionDetails.setName(label.getIdentifier());
                 actionDetails.setValue(updatedLabelValue);
                 actionDetails.setPreviousValue(originalLabelValue);
                 labelAction.setActionDetails(actionDetails);
@@ -164,7 +163,8 @@ public class LabelAuditor extends NiFiAuditor {
             action.setOperation(operation);
             action.setTimestamp(new Date());
             action.setSourceId(label.getIdentifier());
-            action.setSourceName(SOURCE_NAME);
+            // Labels do not have a Name; use UUID to provide a unique name
+            action.setSourceName(label.getIdentifier());
             action.setSourceType(Component.Label);
 
             if (actionDetails != null) {
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/audit/LabelAuditor.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/audit/LabelAuditor.java
@@ -1,0 +1,193 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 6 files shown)


exceptionfactory (commented):
Thanks for implementing auditing for Labels @markobean!

The general approach looks good. Although there is minimal unit testing of other Auditor classes, I created and pushed a test for the `LabelAuditor` following a similar pattern to the test for the `ProcessGroupAuditor`. The Spring Configuration enables the AspectJ interception, which is a helpful confirmation that the code works as expected.

This works at runtime, but I have one concern about the Source Name. Since Labels do not have a name field, viewing the changes in the Flow Configuration History does not provide a way to differentiate between different Labels. I changed `N/A` to `LABEL`, but neither are very helpful. One option could be to use the Label ID as the Source Name. It would provide a basic way to differentiate between actions performed on different Label instances. Another option would be to generate an SHA-256 hash of the Label value, which could indicate changes. One more option could be to use the truncated Label value as the Source Name. What do you think?

markobean:
Originally, I had "Label" as the as the Name. However, on the NiFi History table, it shows the same value, i.e. "Label" for both Name and Type. I found that redundant and didn't care for it. Using the UUID might not be bad. However, there is another ticket I plan on working soon which adds the UUID column to the NiFi History table. So, once again, we're into redundancy. That is why I chose simply "N/A" because in reality there is no Name for a Label.

That being said, it's not a hill I'm willing to die on no matter what the value is. I'm ok with any of  "N/A", "LABEL" or UUID value. Still, if I had my druthers, I'd go with "N/A" since it is the most accurate.

exceptionfactory:
> Originally, I had "Label" as the as the Name. However, on the NiFi History table, it shows the same value, i.e. "Label" for both Name and Type. I found that redundant and didn't care for it. Using the UUID might not be bad. However, there is another ticket I plan on working soon which adds the UUID column to the NiFi History table. So, once again, we're into redundancy. That is why I chose simply "N/A" because in reality there is no Name for a Label.
> 
> That being said, it's not a hill I'm willing to die on no matter what the value is. I'm ok with any of "N/A", "LABEL" or UUID value. Still, if I had my druthers, I'd go with "N/A" since it is the most accurate.

Thanks for the reply @markobean!

I can see value in a separate PR that introduces the UUID as a column to help with this situation. In light of that idea, and to at least have a way to distinguished Labels for now, what do you think about `LABEL-$UUIDPrefix`, where `$UUIDPrefix` is the first eight characters of the UUID (such as `LABEL-abcdef01`)? Obviously that is not guaranteed to be unique, but it will often be different enough for general differentiation.

markobean:
> > Originally, I had "Label" as the as the Name. However, on the NiFi History table, it shows the same value, i.e. "Label" for both Name and Type. I found that redundant and didn't care for it. Using the UUID might not be bad. However, there is another ticket I plan on working soon which adds the UUID column to the NiFi History table. So, once again, we're into redundancy. That is why I chose simply "N/A" because in reality there is no Name for a Label.
> > That being said, it's not a hill I'm willing to die on no matter what the value is. I'm ok with any of "N/A", "LABEL" or UUID value. Still, if I had my druthers, I'd go with "N/A" since it is the most accurate.
> 
> Thanks for the reply @markobean!
> 
> I can see value in a separate PR that introduces the UUID as a column to help with this situation. In light of that idea, and to at least have a way to distinguished Labels for now, what do you think about `LABEL-$UUIDPrefix`, where `$UUIDPrefix` is the first eight characters of the UUID (such as `LABEL-abcdef01`)? Obviously that is not guaranteed to be unique, but it will often be different enough for general differentiation.

Actually, I'm starting to think "N/A" really is the way to go. The reason is that when you put `Label` (or presumably `Label-$UUDPrefix`) into the search field, you get no results. I can see users expecting to get results when they enter the "Name" value into the search. (I think we ought to add a "go to" link in the NiFi History table too, now that I think about it.)

exceptionfactory:
> Actually, I'm starting to think "N/A" really is the way to go. The reason is that when you put `Label` (or presumably `Label-$UUDPrefix`) into the search field, you get no results. I can see users expecting to get results when they enter the "Name" value into the search. (I think we ought to add a "go to" link in the NiFi History table too, now that I think about it.)

The search behavior is worth considering, but that is outside of the Flow Configuration History, so that seems like less of a concern than being able to distinguish items in the Flow Configuration History view itself. The Flow Configuration History can be filtered on Component ID or User, unfortunately filtering on Name is not an option right now, but that could also be improved. 

I agree that adding a link from the History row to the component would be very helpful.

markobean:
> > Actually, I'm starting to think "N/A" really is the way to go. The reason is that when you put `Label` (or presumably `Label-$UUDPrefix`) into the search field, you get no results. I can see users expecting to get results when they enter the "Name" value into the search. (I think we ought to add a "go to" link in the NiFi History table too, now that I think about it.)
> 
> The search behavior is worth considering, but that is outside of the Flow Configuration History, so that seems like less of a concern than being able to distinguish items in the Flow Configuration History view itself. The Flow Configuration History can be filtered on Component ID or User, unfortunately filtering on Name is not an option right now, but that could also be improved.
> 
> I agree that adding a link from the History row to the component would be very helpful.

Ok, now I understand the motivation for a unique "Name" better when put in the context of the NiFi History, aka Flow Configuration History view itself. So humor me and assume UUID is a column in that table. Choose your poison ("poison" meaning there will be inherent redundancy): `Lable-$UUDPrefix` or straight `UUID`. If uniqueness is the goal, the latter would be a better choice.

If you agree, I'll make that update this weekend.

And, to be clear the "go to" as well as UUID in the history table is out of scope for this particular PR; we'll cover that in separate ticket(s).

exceptionfactory:
Thanks @markobean, going with just the UUID itself sounds like the simple solution for now, so if you agree and make the changes, that sounds good.

I agree that other UI changes are out scope for this issue and could be handled separately.

markobean:
Pushed latest changes which includes using Label UUID for the Name in the configuration history.

Thanks @exceptionfactory for the review and for adding unit tests.

exceptionfactory (approved):
Thanks for making the adjustments @markobean, the latest version looks good! +1 merging

PR #6969, opened 2023-02-17
NIFI-11196 Restrict ListFile to primary node only

nandorsoma:
# Summary

[NIFI-11196](https://issues.apache.org/jira/browse/NIFI-11196)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ListFile.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ListFile.java
@@ -19,6 +19,7 @@ package org.apache.nifi.processors.standard;
 
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
+import org.apache.nifi.annotation.behavior.PrimaryNodeOnly;
 import org.apache.nifi.annotation.behavior.Stateful;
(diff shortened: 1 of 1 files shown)


mattyb149:
@markap14 This looks consistent with the docs but wanted to check with you since it could affect files landing in each node

nandorsoma:
> @markap14 This looks consistent with the docs but wanted to check with you since it could affect files landing in each node

Thanks for looking at it [@mattyb149](https://github.com/mattyb149)! Just a little more info. I noticed after opening the PR that when the processor is [listing a local directory](https://github.com/apache/nifi/blob/a63c3594c9dbbbbd1afcb0844b7ab0e1c3e6f171/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ListFile.java#L135), it may work on multiple nodes while it is guaranteed that the local directory contains different files on each node. In that situation, the state is stored locally, so it doesn't matter whether the primary node changes. If I'm right, this pr is not acceptable. Instead, we could update the description accordingly and add a validation (of possible) that prevents using the processor on all nodes when the `Input Directory Location` is set to `Local`.

mattyb149:
I think that was the original use case for ListFile, to list files local to each node, so I'm not sure we want to prevent that. Is the local state the issue?

nandorsoma:
Closing this as it became clear to me that only the description of the processor is obsolete, and it is a valid use-case to use this processor on a multi-node env with `Local` Input Directory.

PR #6984, opened 2023-02-22
NIFI-11206 Make KerberosKeyTabUserService respect NIFI_ALLOW_EXPLICIT_KEYTAB env variable

nandorsoma:
# Summary

[NIFI-11206](https://issues.apache.org/jira/browse/NIFI-11206)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


bbende:
Hi @nandorsoma , I don't think this is correct to add this to the `KerberosKeytabUserService` because the service is protected by restricted permissions, so if someone wants to protect access to the keytab they can decide who gets access to the service through the restricted permissions. You can see that the original `KeytabCredentialsService` also does not check for `NIFI_ALLOW_EXPLICIT_KEYTAB`.

The case where we need to check the explicit usage of keytabs is when it is directly in a processor property because then there is no separation between the processor and the keytab.

bbende:
Also one more reference point is the description of the env var from nifi-env.sh
```
# Set to false to force the use of Keytab controller service in processors
# that use Kerberos. If true, these processors will allow configuration of keytab
# and principal directly within the processor. If false, these processors will be
# invalid if attempting to configure these properties. This may be advantageous in
# a multi-tenant environment where management of keytabs should be performed only by
# a user with elevated permissions (i.e., users that have been granted the 'ACCESS_KEYTAB'
# restriction).
```
So the idea is if you set `NIFI_ALLOW_EXPLICIT_KEYTAB=false`, then you are forced to use a `KerberosCredentialsService` or `KerberosUserService`, so we can't make those invalid.

PR #6990, opened 2023-02-26
NIFI-11205 Adding deprecation log for redundant Kerberos properties in Kafka processors

nandorsoma:
# Summary

[NIFI-11205](https://issues.apache.org/jira/browse/NIFI-11205)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 8
  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-kafka-bundle/nifi-kafka-shared/src/main/java/org/apache/nifi/kafka/shared/validation/KafkaDeprecationValidator.java
+++ b/nifi-nar-bundles/nifi-kafka-bundle/nifi-kafka-shared/src/main/java/org/apache/nifi/kafka/shared/validation/KafkaDeprecationValidator.java
@@ -28,16 +28,16 @@ import static org.apache.nifi.kafka.shared.component.KafkaClientComponent.SELF_C
 
 public class KafkaDeprecationValidator {
 
-    private static final DeprecationLogger deprecationLogger = DeprecationLoggerFactory.getLogger(KafkaDeprecationValidator.class);
+    public static void validate(Class<?> componentClass, final String identifier, final ValidationContext validationContext) {
+        final DeprecationLogger deprecationLogger = DeprecationLoggerFactory.getLogger(componentClass);
 
-    public static void validate(String processorName, final String identifier, final ValidationContext validationContext) {
         final PropertyValue credentialsServiceProperty = validationContext.getProperty(KERBEROS_CREDENTIALS_SERVICE);
         final PropertyValue principalProperty = validationContext.getProperty(KERBEROS_PRINCIPAL).evaluateAttributeExpressions();
         final PropertyValue keyTabProperty = validationContext.getProperty(KERBEROS_KEYTAB).evaluateAttributeExpressions();
 
         if (credentialsServiceProperty.isSet()) {
             deprecationLogger.warn("{}[id={}] [{}] Property should be replaced with [{}] Property",
-                    processorName,
+                    componentClass.getSimpleName(),
                     identifier,
                     KERBEROS_CREDENTIALS_SERVICE.getDisplayName(),
                     SELF_CONTAINED_KERBEROS_USER_SERVICE.getDisplayName()
@@ -46,7 +46,7 @@ public class KafkaDeprecationValidator {
 
         if (principalProperty.isSet() || keyTabProperty.isSet()) {
             deprecationLogger.warn("{}[id={}] [{}] and [{}] Properties should be replaced with [{}] Property",
-                    processorName,
+                    componentClass.getSimpleName(),
                     identifier,
                     KERBEROS_PRINCIPAL.getDisplayName(),
                     KERBEROS_KEYTAB.getDisplayName(),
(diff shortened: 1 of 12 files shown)


nifi-nar-bundles/nifi-kafka-bundle/nifi-kafka-shared/src/main/java/org/apache/nifi/kafka/shared/validation/KafkaDeprecationValidator.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.nifi.deprecation.log.DeprecationLoggerFactory;
+
+import static org.apache.nifi.kafka.shared.component.KafkaClientComponent.KERBEROS_CREDENTIALS_SERVICE;
+import static org.apache.nifi.kafka.shared.component.KafkaClientComponent.KERBEROS_KEYTAB;
+import static org.apache.nifi.kafka.shared.component.KafkaClientComponent.KERBEROS_PRINCIPAL;
+import static org.apache.nifi.kafka.shared.component.KafkaClientComponent.SELF_CONTAINED_KERBEROS_USER_SERVICE;
+
+public class KafkaDeprecationValidator {
+
+    private static final DeprecationLogger deprecationLogger = DeprecationLoggerFactory.getLogger(KafkaDeprecationValidator.class);
```
exceptionfactory, line 31: The deprecation logs should reference the component class, instead of the `KafkaDeprecationValidator`. This could be implemented by passing the Processor `Class` reference to the `validate` method, and then getting the `DeprecationLogger` instance in the method.

nifi-nar-bundles/nifi-kafka-bundle/nifi-kafka-shared/src/main/java/org/apache/nifi/kafka/shared/validation/KafkaDeprecationValidator.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public static void validate(String processorName, final String identifier, final ValidationContext validationContext) {
```
exceptionfactory, line 33: Recommend changing the signature to use `Class<?> componentClass` instead of `String processorName` to support mapping the deprecation warnings to the component more closely.

exceptionfactory (changes requested):
Thanks for implementing the deprecation logging @nandorsoma.

The implementation works well in that it requires minimal changes to each component. I recommend adjusting the `DeprecationLoggerFactory` call to take the Component Class instance in the `validate()` method, which would ensure that the deprecation logger itself maps to the component type.

exceptionfactory (changes requested):
@nandorsoma In the course of reviewing this pull request again as compared to the one that removes the deprecated properties, I think there is additional work necessary.

I noticed the removal pull request adds the `Kerberos User Service` property in several components where it does not currently exist.

For these kinds of property deprecations, we need to have a solution that provides the alternative for NiFi 1.21.0.

With that background, there are two options. One is introducing the new property to the 2_0 components, the other is deprecating the entire set of 2_0 components for removal, leaving just the 2_6 components. That actually seems like the better approach on this pull request, which would also impact the removal pull request.

What do you think @markap14, should the Kafak_2_0 components be deprecated for removal as well?

Based on some consideration of a new design for Kafka components in general, deprecating the 2_0 components in NiFi 1.21 and removing them in NiFi 2.0.0 seems like the best option, but it would be helpful to get some additional input given the widespread usage of these components.

nandorsoma:
Thank you for the review, @exceptionfactory! I didn't know that the logger should reference the componentClass, but it is a good idea to pass it instead of the processorName.
Regarding the new properties in the 2_0 processors... Maybe I misunderstood the concept of the deprecation logs. I didn’t know we needed to provide an alternative for the deprecated component on the same main version. Is it really a requirement? The message is unlucky because it suggests that, but with a little more detail, it would be clear. If it is unacceptable, I will favor removing 2_0 processors because I think it would be wrong to create a precedence to add a non-bugfix change to this branch.

exceptionfactory:
Thanks for the reply @nandorsoma.

The deprecation logs are a tool, sometimes suggesting alternatives, but sometimes no alternatives are available. The log message indicating an alternative is what caught my attention.

Adding the DeprecationNotice to the 2_0 components would remove the need to log warnings for the properties on those components. That would make use of the current implementation for the 2_6 components.

Will look for some additional feedback before moving in one direction or the other.

markap14:
Yeah @exceptionfactory I think the 2.0 components can be deprecated at this point as well.

exceptionfactory:
Thanks for the confirmation @markap14!

With that in mind, @nandorsoma, the 2_0 components should have the `@DeprecationNotice` annotation added, referencing the 2_6 alternative, and the property deprecation logging for the 2_0 components can be removed.

exceptionfactory:
One other note @nandorsoma, in conjunction with the PR to remove the deprecated properties, please add the Kerberos User Service property to the ConsumeKafkaRecord_2_6 and KafkaRecordSink_2_6 components. That will provide a clear migration path in the NiFi 1.21.0 version.

nandorsoma:
Thanks for your review, @exceptionfactory! Please see my latest commit, which addresses your comments!

PR #6991, opened 2023-02-27
NIFI-11224 - FF attribute support in Where condition of QuerySalesforceObject

pvillard31:
# Summary

[NIFI-11224](https://issues.apache.org/jira/browse/NIFI-11224)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-salesforce-bundle/nifi-salesforce-processors/src/main/java/org/apache/nifi/processors/salesforce/QuerySalesforceObject.java
+++ b/nifi-nar-bundles/nifi-salesforce-bundle/nifi-salesforce-processors/src/main/java/org/apache/nifi/processors/salesforce/QuerySalesforceObject.java
@@ -343,7 +343,9 @@ public class QuerySalesforceObject extends AbstractProcessor {
         AtomicReference<String> nextRecordsUrl = new AtomicReference<>();
         String sObject = context.getProperty(SOBJECT_NAME).getValue();
         String fields = context.getProperty(FIELD_NAMES).getValue();
-        String customWhereClause = context.getProperty(CUSTOM_WHERE_CONDITION).evaluateAttributeExpressions(originalFlowFile).getValue();
+        String customWhereClause = originalFlowFile == null
+                ? context.getProperty(CUSTOM_WHERE_CONDITION).evaluateAttributeExpressions().getValue() :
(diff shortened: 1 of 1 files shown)


nifi-nar-bundles/nifi-salesforce-bundle/nifi-salesforce-processors/src/main/java/org/apache/nifi/processors/salesforce/QuerySalesforceObject.java
```diff
@@ -326,24 +326,24 @@ protected Collection<ValidationResult> customValidate(ValidationContext validati
     public void onTrigger(ProcessContext context, ProcessSession session) throws ProcessException {
         boolean isCustomQuery = CUSTOM_QUERY.getValue().equals(context.getProperty(QUERY_TYPE).getValue());
 
+        FlowFile flowFile = session.get();
```
mark-bathori, line 329: I think the null check for the FlowFile would be still needed here, so it would be better to move that check condition outside from the `isCustomQuery` check.

pvillard31: Thanks for the review, since we do want to evaluate EL wether the FlowFile is null or not, I added a null check in the ``processQuery`` part.

mark-bathori (approved):
Thank you for the improvements @pvillard31 
+1 LGTM

nifi-nar-bundles/nifi-salesforce-bundle/nifi-salesforce-processors/src/main/java/org/apache/nifi/processors/salesforce/QuerySalesforceObject.java
Lehel44: I recommend removing flowfile == null check before the condition. Based @markap14 answer in case of processors annotated with @TriggerSerially this check is not necessary. However if a processor can have multiple executing thread, the following can happen:

"- 1 FlowFile in queue and 2 threads start running.
-  Thread 1 calls session.get() and gets the only FlowFile.
-  Then Thread 2 calls session.get() and gets back a null."

The flowfile is optional and the null check will be handled in the `processQuery` and `processCustomQuery` methods.

PR #6993, opened 2023-02-28
NIFI-11231 Stateless NiFi sensitive parameter context support

slambrose:
# Summary

[NIFI-11231](https://issues.apache.org/jira/browse/NIFI-11231)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-11231) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 11
  - [X] JDK 17

### Licensing

- [X] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html) - N/A
- [X] New dependencies are documented in applicable `LICENSE` and `NOTICE` files - N/A

### Documentation


nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/engine/StatelessFlowManager.java
```diff
@@ -270,10 +271,9 @@ public FlowFileQueue createFlowFileQueue(final LoadBalanceStrategy loadBalanceSt
     public ReportingTaskNode createReportingTask(final String type, final String id, final BundleCoordinate bundleCoordinate, final Set<URL> additionalUrls, final boolean firstTimeAdded,
                                                  final boolean register, final String classloaderIsolationKey) {
 
-        if (type == null || id == null || bundleCoordinate == null) {
```
joewitt, line 273: With this removal will the log/exception information still be as specific as it was?  The code is clearly cleaner but we go backwards in communicating intent.  Am i reading that right?

slambrose: Valid point. I used this implementation because it was cleaner as well. The requireNonNull throws a generic NullPointerException, but doesn't have the full details like the current code:
throw new NullPointerException("Must supply type, id, and bundle coordinate in order to create Reporting Task. Provided arguments were type=" + type + ", id=" + id + ", bundle coordinate = " + bundleCoordinate);
                
I can change this back if desired, just let me know.

joewitt: Yeah I'd not change this since while the code is cleaner it is also far less effective in conveying intent to the user.  Ultimately that is our desire given this block of code.

slambrose: Done :)

nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/engine/StatelessFlowManager.java
```diff
@@ -320,9 +321,55 @@ public ReportingTaskNode createReportingTask(final String type, final String id,
...
+            taskNode = new ComponentBuilder()
+                    .identifier(id)
+                    .type(type)
+                    .bundleCoordinate(bundleCoordinate)
+                    .statelessEngine(statelessEngine)
+                    .additionalClassPathUrls(additionalUrls)
+                    .flowManager(this)
+                    .buildParameterProvider();
+        } catch (ParameterProviderInstantiationException e) {
+            throw new RuntimeException("Could not create Parameter Context Task of type " + type + " with ID " + id, e);
```
exceptionfactory, line 344: Following the pattern of other methods, this should throw an `IllegalStateException` instead of the more generic `RuntimeException`.

nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/engine/StatelessFlowManager.java
```diff
@@ -320,9 +321,55 @@ public ReportingTaskNode createReportingTask(final String type, final String id,
...
+                                                         final boolean registerLogObserver) {
+        if (type == null || id == null || bundleCoordinate == null) {
+            throw new NullPointerException("Must supply type, id, and bundle coordinate in order to create Reporting Task. Provided arguments were type=" + type + ", id=" + id
+                + ", bundle coordinate = " + bundleCoordinate);
+        }
+
+        // make sure the first reference to LogRepository happens outside of a NarCloseable so that we use the framework's ClassLoader
+        final LogRepository logRepository = LogRepositoryFactory.getRepository(id);
+
+        final ParameterProviderNode taskNode;
```
exceptionfactory, line 333: The variable name should be changed to `providerNode` instead of `taskNode` to follow the naming convention of the class.
```suggestion
        final ParameterProviderNode providerNode;
```

nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/engine/StatelessFlowManager.java
```diff
@@ -320,9 +321,55 @@ public ReportingTaskNode createReportingTask(final String type, final String id,
...
+                    .buildParameterProvider();
+        } catch (ParameterProviderInstantiationException e) {
+            throw new RuntimeException("Could not create Parameter Context Task of type " + type + " with ID " + id, e);
+        }
 
-        throw new UnsupportedOperationException("Parameter Providers are not supported in Stateless NiFi");
+        LogRepositoryFactory.getRepository(taskNode.getIdentifier()).setLogger(taskNode.getLogger());
+
+        if (firstTimeAdded) {
+            final Class<?> taskClass = taskNode.getParameterProvider().getClass();
```
exceptionfactory, line 350:
```suggestion
            final Class<?> providerClass = taskNode.getParameterProvider().getClass();
```

nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/engine/StatelessFlowManager.java
```diff
@@ -320,9 +321,55 @@ public ReportingTaskNode createReportingTask(final String type, final String id,
...
+                }
+            } catch (final Exception e) {
+                throw new ComponentLifeCycleException("Failed to invoke On-Added Lifecycle methods of " + taskNode.getParameterProvider(), e);
+            }
+        }
+
+        if (registerLogObserver) {
+            onParameterProviderAdded(taskNode);
+
+            // Register log observer to provide bulletins when reporting task logs anything at WARN level or above
```
exceptionfactory, line 367: This comment appears to be copied from Reporting Tasks, so it should be updated or removed.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/logging/ParameterContextTaskLogObserver.java
```diff
@@ -0,0 +1,50 @@
...
+package org.apache.nifi.logging;
+
+import org.apache.nifi.controller.ParameterProviderNode;
+import org.apache.nifi.events.BulletinFactory;
+import org.apache.nifi.reporting.Bulletin;
+import org.apache.nifi.reporting.BulletinRepository;
+import org.apache.nifi.reporting.ComponentType;
+import org.apache.nifi.reporting.Severity;
+
+public class ParameterContextTaskLogObserver implements LogObserver {
```
exceptionfactory, line 26: This should be renamed to `ParameterProviderLogObserver` to reflect the type of object being observed.
```suggestion
public class ParameterProviderLogObserver implements LogObserver {
```

exceptionfactory (changes requested):
Thanks for the contribution @slambrose!

I noted a handful a small naming issues, but the basic approach follows the example of other components.

Can you provide some additional background on the intended use case? The Stateless execution environment has a different approach for loading flow parameters using [Parameter Value Providers](https://github.com/apache/nifi/tree/main/nifi-stateless/nifi-stateless-assembly#passing-parameters). This is different than Parameter Providers in traditional NiFi, which is part of the reason that Parameter Providers are not supported right now.

@gresockj May also have some additional background from work on both Parameter Providers and Parameter Value Providers.

slambrose (commented):
Implemented all requested changes.

gresockj:
> Hmm.. I'm not sure Parameter Value Provider would work for us since these have to be passed in on runtime and known beforehand. We're going to be supporting a variety of flows stored in Registry. The goal is to run stateless NiFi with any given Registry url, bucket id, flow id, and flow version without any knowledge of what properties/params users have configured in their flows. So those flows that utilize sensitive parameter contexts need to be able to work under any running statless nifi pod/container. We wouldn't be changing the run command or properties files for each flow that's processed with stateless NiFi. Let me chat some more with @Dye357 to see if there's anything I'm missing.

Hi @slambrose, what you describe should be possible with the existing EnvironmentVariableParameterValueProvider.  Let me know if you have any other questions about the setup.

PR #4008, opened 2020-01-22
NIFI-7060 Replace properties in nifi.properties by environment variables

yarox:
Makes it possible to replace properties set in `nifi.properties` using environment variables when launching from Docker.

`start.sh` replaces some properties defined in `nifi.properties` with a set of environment variables.

This adds some flexibility as the users are able to modify properties without touching the `nifi.properties` file but it can lead to some frustration when users think they can replace any property by defining an environment variable with the same name. Also, when users want to provide their own `nifi.properties` file as some properties will be overwritten.

I believe we should only replace a property when there is an environment variable defined with the same name (we should use the same names that appear in the `nifi.properties` file in uppercase and replacing any `.` with `_`). Defaults should be taken from the `nifi.properties` file.

This way users can replace the properties file with a custom one, choose to define environment variables, or both.

PR #6995, opened 2023-02-28
NIFI-7060 populate NiFi and NiFi Registry properties files using env vars in Docker Containers

ChrisSamo632:
# Summary

[NIFI-7060](https://issues.apache.org/jira/browse/NIFI-7060) populate NiFi and NiFi Registry properties files using env vars in Docker Containers

Refreshes the original work by @yarox in #4008, rebased against latest `main` and extended to cover `apache/nifi-registry` along with `apache/nifi` images.

- Allows for the specification of general env vars to cover any properties to be updated or added before instance startup.
- Caters for existing env vars where they don't match the property names to be populated.
- Allows for the injection of sensitive properties using files and referencing those via an env var, e.g. for passwords, sensitive keys or secrets
- Fixes various inconsistencies between `dockermaven` and `dockerhub` scripts in NiFi and NiFi Registry (also tidies up some issues around OIDC env vars for Registry)

TODO:

- [ ] update documentation
- [ ] build and test images for different env var use cases (e.g. mixed-case properties)
- [ ] rebase branch to single commit (left as multiple currently for retaining original work)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


exceptionfactory (commented):
Thanks bringing back the pull request for discussion @ChrisSamo632.

There are a lot of good ideas and improvements in this PR, and it might be helpful to break them into separate pieces, as some of them could go forward more easily.

Related to the title of the pull request, I'm wondering if we should consider a different approach to environment-based property injection. The approach is specific to the Docker startup scripts, which is helpful in some respects, but limiting in others. Instead of making the environment variable-based property resolution depend on Docker launch scripts, it seems worth considering adjustments to `nifi-properties` loading itself, where environment variable-based resolution could be enabled through a boostrap.conf argument. That would allow non-Docker-based deployments to opt-in to environment variable resolution. That's just one idea based on approaches in other systems, and also from NiFi Stateless.

With that note, I would still like to see a number of the other incremental improvements in this PR either way.

ChrisSamo632:
@exceptionfactory I agree that there's utility in discussing some of the changes being made herein, it was the main driver for raising a Draft PR (rather than pushing through with more updates), so glad you noticed and thanks for commenting.

I agree that an instance-based loading mechanism would make more sense and I'd been considering that as I took forward the (Docker Image-focussed, but not strictly limited) current env var approach, however I didn't want to go through making such changes without some discussion as I think there needs to be agreement on a common approach of how to load properties and what order of precedence is to be used (i.e. properties file vs. env var vs. defaults in property loader code, etc.) - this seems like it's worth a ticket in its own right (whether that's NIFI-7060 that's re-shaped *or* a new ticket to be raised).

Other things that have been covered in this PR so far that we may/not want to retain:
- [ ] address `shellcheck` errors/warnings in the Docker Image start scripts
- [ ] load (sensitive) properties from files
  - do we keep this as a Docker Image-focussed approach or look to generalise this for other deployments too (I'm not sure how much sense it makes as likely any such files are just as readable as the `nifi/-registry.properties` files themselves on such systems, but in Docker/k8s the injection of a sensitive value as a file does make things a little more secure as the values can't be seen by a simple inspection of the running container's env)
- [ ] update `maven-antrun-plugin` for `dockermaven` builds
- [ ] refactor the Docker Image start scripts for `nifi-registry`
  - this fixed a few bugs/inconsistencies between the current `dockermaven` and `dockerhub` scripts currently on `main`, including:
    - missing `needClientAuth` settings
    - inability to configure the `database` for the Registry Flow Provider
    - missing LDAP `Referral Strategy` config
    - invalid OIDC properties (copied from `nifi`, so missing the `.registry` part of the property name *and* some properties are missing/present when unused)

I'm not precious about keeping this PR open (after the discussion) and/or linking it to a different Jira ticket in future to cover the points we want to keep from this existing work vs. raising/reshaping remaining changes in NIFI-7060/a new ticket

exceptionfactory:
Thanks for the thoughtful reply @ChrisSamo632.

For the simple things first, it seems like the following items could be pulled out to a simpler PR:

- address shellcheck errors/warnings in the Docker Image start scripts
- update maven-antrun-plugin for dockermaven builds

The changes for NiFi Registry also seem self-contained, but might involve a little more review.

Loading properties from files would be a helpful general improvement to properties loading. I could see that as either an incremental improvement, or part of a more generalized adjustment to property loading. The work around Sensitive Property Providers for loading values from secrets managers introduced some abstractions for NiFiProperties, which may need to be evaluated as part of the approach.

One of the short-comings of the current Docker container is the need to define environment variables and map them to properties in the script. This requires manual updates to support additional properties. One the major advantages of moving environment variable loading into the application would be supporting convention-based mapping. This is a standard feature of Spring Boot, and provides a good pattern. With bootstrap.conf, it should be possible to select the property sources, perhaps the default is the current file properties, but the Docker image would set the value use convention-based environment variables. NiFi Registry is the only application component that uses Spring Boot, so I'm not suggesting we need to move in that direction, just that we should able to follow the same basic idea of convention-based environment variable mapping.

Those are my initial thoughts. I'm sure others like @kevdoran have some ideas along these lines.

With that, if you're interested in carving out some of these things, I think that would be useful, and we should put some thought into a new Jira issue that encompasses the goals of NIFI-7060.

exceptionfactory:
@ChrisSamo632 Just following up on this discussion, what do you think about breaking out some of the simpler improvements into separate pull requests, then focusing in on the environment-variable-backed properties?

ChrisSamo632:
> @ChrisSamo632 Just following up on this discussion, what do you think about breaking out some of the simpler improvements into separate pull requests, then focusing in on the environment-variable-backed properties?

@exceptionfactory, yes it's on my list of things to try and look into as soon as I've time, although it'll likely be a couple of weeks at least - I'd propose raising one/separate PRs for points 1, 3 & 4 in my suggested list above then a Jira for point 2 and either NIFI-7060 is reshaped into a more general "environment-variable-backed properties" piece (and this PR closed) or a new Jira raised to cover that and effetively replace NIFI-7060

ChrisSamo632:
@exceptionfactory I've raised tickets:

- [NIFI-11311](https://issues.apache.org/jira/browse/NIFI-11311) to upgrade `maven-antrun-plugin`
- [NIFI-11312](https://issues.apache.org/jira/browse/NIFI-11312) to rationalise the `nifi-registry` Docker Images builds
- [NIFI-11313](https://issues.apache.org/jira/browse/NIFI-11313) to address the `shellcheck` issues
- [NIFI-11314](https://issues.apache.org/jira/browse/NIFI-11314) for loading sensitive properties from local files (referenced by environment variables)
- [NIFI-11315](https://issues.apache.org/jira/browse/NIFI-11315) for general environment variable configuration of applications

With these, I'll close this PR and (hopefully sometime soon) break out the changes for the first 3 simpler tickets (although they're open for others to do first if they are interested and beat me to them).

exceptionfactory:
Thanks @ChrisSamo632, sounds good!

kevdoran:
Hi @exceptionfactory @ChrisSamo632 - Sorry I missed this PR initially, and then I've been out sick recently. I'm going to try to ramp up on this set of changes so I can help with reviews.

Regarding 
- [NIFI-11315](https://issues.apache.org/jira/browse/NIFI-11315) for general environment variable configuration of applications

I did look into this a while back. I agree that moving this capability into NiFi application code itself would be the most robust. Some searches showed that it might be be possible to use SpringBoot's `ConfigurationProperties` library without using SpringBoot, but that would probably be too big a lift. It's an option, but maybe we could implement our own solution. I'll try to comment on the Jira.

PR #6996, opened 2023-02-28
NIFI-11232 Fix buffer handling in ContentClaimInputStream

Chrzi:
# Summary

[NIFI-11232](https://issues.apache.org/jira/browse/NIFI-11232)

Under some circumstances the buffer is not recreated correctly in ContentClaimInputStream leading to a situation where data is read from the buffer and the buffer trying to read from a closed stream. More details in the issue.

One case where this can happen is in a custom processor using Apache Tika. Passing the InputStream created of a flowfile by the processSession to Tika exposes the bug, since it uses the mark and reset feature extensively.

This PR creates a new buffer when the stream is marked to buffer the readLimit (which is the primary cause why it was introduced in NIFI-10888). It also throws the buffer away during reset, if a reset within the buffer is not possible. 

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/io/TestContentClaimInputStream.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/io/TestContentClaimInputStream.java
@@ -63,7 +63,7 @@ public class TestContentClaimInputStream {
         final byte[] buff = new byte[5];
         StreamUtils.fillBuffer(in, buff);
 
-        Mockito.verify(repo, Mockito.times(1)).read(contentClaim);
+        Mockito.verify(repo).read(contentClaim);
         Mockito.verifyNoMoreInteractions(repo);
 
         final String contentRead = new String(buff);
@@ -85,7 +85,7 @@ public class TestContentClaimInputStream {
         final byte[] buff = new byte[2];
         StreamUtils.fillBuffer(in, buff);
 
-        Mockito.verify(repo, Mockito.times(1)).read(contentClaim);
+        Mockito.verify(repo).read(contentClaim);
         Mockito.verifyNoMoreInteractions(repo);
 
         final String contentRead = new String(buff);
@@ -107,7 +107,7 @@ public class TestContentClaimInputStream {
         final byte[] buff = new byte[5];
 
         final int invocations = 10;
-        for (int i=0; i < invocations; i++) {
+        for (int i = 0; i < invocations; i++) {
             in.mark(5);
 
             StreamUtils.fillBuffer(in, buff, true);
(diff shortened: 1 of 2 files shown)


exceptionfactory (commented):
Thanks for submitting the contribution to address this issue @Chrzi!

One general question, given the important nature of the problem, have you evaluated implementing a unit test for this class to verify the intended behavior?

Chrzi:
I have confirmed the behaviour using the unit test provided in the Issue, which I also used to reproduce the bug.

At the moment I am having issues executing `mvn clean install -P contrib-check` (the contrib-check fails) and just asked on Slack for help.

If that works I want to get the unit test into a state where it is mergeable.

exceptionfactory:
> If that works I want to get the unit test into a state where it is mergeable.

Thanks, adding the unit test to the PR would be very helpful.

Chrzi:
I tested that the new unit test `testBigReadAfterSmallRereads` really tests for this bug, if ran against the current main branch it fails the large read in line 262 with an IOException.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/io/TestContentClaimInputStream.java
```diff
@@ -152,7 +153,7 @@ public void testMultipleResetCallsAfterMark() throws IOException {
             in.reset();
         }
 
-        Mockito.verify(repo, Mockito.times(invocations + 1)).read(contentClaim); // Will call reset() 'invocations' times plus the initial read
+        Mockito.verify(repo, Mockito.times( 1)).read(contentClaim);
```
exceptionfactory, line 156:
```suggestion
        Mockito.verify(repo).read(contentClaim);
```

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/io/TestContentClaimInputStream.java
```diff
@@ -183,11 +184,88 @@ public void testRereadWithOffset() throws IOException {
...
```
exceptionfactory, line 187:
```suggestion
        Mockito.verify(repo).read(contentClaim);
```

exceptionfactory (commented):
Thanks for addressing this issue and adding the unit tests @Chrzi.

I noted several style-related suggestions, but the overall approach looks sound.

Tagging @markap14 for additional review since this relates to recent changes.

markap14:
Thanks for the fix @Chrzi !

exceptionfactory:
Thanks for the confirmation @markap14!

Thanks again for the analysis and solution @Chrzi, nice work! As my suggestions were related to style and formatting, I pushed a commit to the branch implementing the changes and plan on merging.

exceptionfactory (approved):

PR #7016, opened 2023-03-06
[NIFI-10792] Fixed bug to allow for processing files larger than 10MB…

dan-s1:
…. Also refactored to reduce the cognitive complexity in the onTrigger method.

# Summary

[NIFI-10792](https://issues.apache.org/jira/browse/NIFI-10792)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/poi/ConvertExcelToCSVProcessor.java
+++ b/nifi-nar-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/poi/ConvertExcelToCSVProcessor.java
@@ -51,7 +51,9 @@ import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


dan-s1:
@exceptionfactory I did not end up including the unit test I had as it was a unit which tested with a 20MB file. I would have thought there should be a unit test to exercise the change I made.  Please advise.

mh013370:
> @exceptionfactory I did not end up including the unit test I had as it was a unit which tested with a 20MB file. I would have thought there should be a unit test to exercise the change I made. Please advise.

Anecdotally, the NiFi mock framework will read the entire FF contents into memory (I discovered this in #6369) so I think you're correct in not including unit tests testing larger FFs.

exceptionfactory:
Thanks for the comment @mh013370, this issue is similar to the size limits for Tar files resolved in #6369.

@dan-s1, Although unit tests are the preferred way to confirm expected behavior, unit tests are not optimal for testing these kinds of scenarios. For this particular situation, confirming existing functionality is good, and runtime verification is better than introducing large files or long-running tests into the repository.

exceptionfactory:
Thanks @dan-s1, I plan to take a closer look at this pull request soon.

The `excel-streaming-reader` seems acceptable, although discarding styles, graphs, and other elements does not necessarily sound like a problem because that would not necessarily translate to record-oriented data for other services.

nifi-nar-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/poi/ConvertExcelToCSVProcessor.java
```diff
@@ -299,55 +225,62 @@ public void process(InputStream inputStream) throws IOException {
...
-                }
+            ff = session.write(ff, out -> {
+                PrintStream outPrint = new PrintStream(out, false, StandardCharsets.UTF_8);
+                sheetHandler.setOutput(outPrint);
+                sheet.forEach(row -> {
+                    sheetHandler.startRow(row.getRowNum());
+                    row.forEach(sheetHandler::cell);
+                    sheetHandler.endRow();
+                });
+                sheetHandler.close();
```
emiliosetiadarma, line 279: Is `outPrint.close()` call no longer needed after `sheetHandler.close()`?

nifi-nar-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/poi/ConvertExcelToCSVProcessor.java
```diff
@@ -199,93 +166,52 @@ public final List<PropertyDescriptor> getSupportedPropertyDescriptors() {
     @Override
     public void onTrigger(final ProcessContext context, final ProcessSession session) throws ProcessException {
         final FlowFile flowFile = session.get();
-        if ( flowFile == null ) {
+        if (flowFile == null) {
             return;
         }
 
-        final String desiredSheetsDelimited = context.getProperty(DESIRED_SHEETS).evaluateAttributeExpressions(flowFile).getValue();
+        final Map<String, Boolean> desiredSheets = getDesiredSheets(context, flowFile);
```
emiliosetiadarma, line 173: Wondering if we could use the primitive `boolean` type here, based on a cursory read-through it doesn't appear we need the Object version here

emiliosetiadarma (commented):
Hey, I was able to verify that I was able to convert an excel spreadsheet > 10 MB successfully into CSV, I just have a few questions

nifi-nar-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/poi/ConvertExcelToCSVProcessor.java
dan-s1: `outPrint.close()` is no longer there and the PrintStream represented with the variable `outPrint` is closed in the sheetHandler's close method.

nifi-nar-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/poi/ConvertExcelToCSVProcessor.java
dan-s1: No a primitive cannot be used here. The Map represented with desiredSheets keeps track of which specified sheets were found (lines 189-196)

dan-s1:
@emiliosetiadarma @exceptionfactory I believe I addressed the issues brought up. Is there anything else needed?

exceptionfactory (approved):
Thanks for your work on this improvement @dan-s1, and thanks for the testing @emiliosetiadarma.

The code changes for streaming look good, existing unit tests pass and I was able to convert a 20 MB XLSX file at runtime with these changes.

I pushed a commit that adds the `log4j-to-slf4j` dependency, which was something missing after recent Apache POI upgrades, supporting log routing from Log4j 2 to the NiFi standard SLF4J. I also removed the explicit dependency on `graphics2d`, which has a holdover from earlier Apache POI versions and no longer used. Lastly, I reverted the reformatting of the properties and relationships, since it is easier to maintain the list of one per line versus wrapping lines.

With these changes, planning to merge following a successful automated build.

nifi-nar-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/poi/ConvertExcelToCSVProcessor.java
```diff
@@ -313,7 +334,7 @@ private String getSheetsNotFound(Map<String, Boolean> desiredSheets) {
      * do most of the work of parsing the contents of the Excel sheet
      * and outputs the contents as a (basic) CSV.
      */
-    private static class SheetToCSV {
+    private class SheetToCSV {
```
dan-s1, line 337: @exceptionfactory Shouldn't this class be static?

exceptionfactory: That was a good change, but there was an `e.printStackTrace()` in a catch block that I changed to call `getLogger().warn()`, requiring the class to be non-static. The other option is to pass the ComponentLog reference to the SheetToCSV class, but since the class was previously non-static, this seemed like a less impactful change. What do you think?

dan-s1: @exceptionfactory I did not realize that. The way you have it then is fine. My motto keep it simple :)

exceptionfactory: Thanks @dan-s1, I appreciate the attention to detail!

exceptionfactory (approved):
Automated builds look good. +1 merging

PR #7023, opened 2023-03-08
NIFI-11261 Add Primary Node State handling to GetAzureEventHub

exceptionfactory:
# Summary

[NIFI-11261](https://issues.apache.org/jira/browse/NIFI-11261) Adds Primary Node State Change handling to `GetAzureEventHub` to improve Processor behavior in a clustered deployment.

Although `ConsumeAzureEventHub` provides the recommend approach for receiving events, `GetAzureEventHub` can be configured to run on the cluster primary node only to reduce the potential for duplication.

Updates include checking for the Execution Node status and avoiding unnecessary Event Hub Consumer Client creation. The new handler method for Primary Node State changes closes the Consumer Client when a node has its primary status revoked, and creates a new Consumer Client when a node is elected primary.

Additional changes include upgrading the Qpid Proton J dependency from 0.34.0 to [0.34.1](https://qpid.apache.org/releases/qpid-proton-j-0.34.1/release-notes.html) and setting the Consumer Client Identifier to the Processor Identifier for improved connection tracking.

New unit test methods provide basic confirmation of Processor behavior when configured for primary node execution.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 11
  - [X] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/GetAzureEventHub.java
+++ b/nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/GetAzureEventHub.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Optional;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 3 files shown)


nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/GetAzureEventHub.java
```diff
@@ -332,7 +385,13 @@ private EventHubConsumerClient createEventHubConsumerClient(final ProcessContext
             final AzureNamedKeyCredential azureNamedKeyCredential = new AzureNamedKeyCredential(policyName, policyKey);
             eventHubClientBuilder.credential(fullyQualifiedNamespace, eventHubName, azureNamedKeyCredential);
         }
-        return eventHubClientBuilder.buildConsumerClient();
+
+        // Set Azure Event Hub Client Identifier using Processor Identifier instead of default random UUID
+        final AmqpClientOptions clientOptions = new AmqpClientOptions();
+        clientOptions.setIdentifier(getIdentifier());
```
turcsanyip, line 391: The processor's identifier is the same on all cluster nodes so all instances would use the same id.
When the processor is configured to run in "All nodes" mode, it can be confusing to have multiple clients with the same identifier. Also, in "Primary node" mode it would be useful to see which node the client belongs to.
So I suggest adding some node-specific tag in the identifier.

turcsanyip (changes requested):
@exceptionfactory Thanks for adding primary node handling to the processor!
I tested it on a cluster and works as expected. The code also looks good to me.
I have only a minor suggestion. Pls. see below.

turcsanyip (approved):
Thanks @exceptionfactory !

+1 LGTM
Merging to main and 1.x

PR #7045, opened 2023-03-15
NIFI-11288 Add missing dependencies required by AWS AssumeRoleWithWebIdentity method

zhangzhonglai:
…Identity method

# Summary

[NIFI-11288](https://issues.apache.org/jira/browse/NIFI-11288)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-service-api/pom.xml
+++ b/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-service-api/pom.xml
@@ -37,6 +37,10 @@
             <artifactId>aws-java-sdk-s3</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-sts</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-api</artifactId>
         </dependency>
--- a/nifi-registry/nifi-registry-extensions/nifi-registry-aws/nifi-registry-aws-extensions/pom.xml
+++ b/nifi-registry/nifi-registry-extensions/nifi-registry-aws/nifi-registry-aws-extensions/pom.xml
@@ -41,6 +41,10 @@
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
+            <artifactId>sts</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
             <artifactId>apache-client</artifactId>
             <exclusions>
                 <exclusion>


ChrisSamo632:
Presumably a duplicate of [NIFI-8641](https://issues.apache.org/jira/browse/NIFI-8641) also?

nandorsoma (commented):
Thanks for the contribution @zhangzhonglai! Based on @ChrisSamo632's comment, it seems like it is not the first time when this feature is requested, so it is worth putting effort into this topic. Could you write a short guide on how to test this feature with one of the processors? If I'm not mistaken we need to obtain a jwt token, which we need to pass to the processor and will be used by the client lib. In the current state how did you obtain the jwt token and how did you pass it to the processor?

zhangzhonglai:
Thanks for your reply, @nandorsoma. In our case, it's a bit complicated, we use NiFi in AWS EKS. Due to permissions issues, the ops will provide me with a service account that has the IAM role authenticated by OIDC, this service account will add an ENV: `AWS_WEB_IDENTITY_TOKEN_FILE : /var/run/secrets/eks.amazonaws.com/serviceaccount/token`  to the Pod. In the NiFi, I use the `AWSCredentialsProviderControllerService` processor and by setting `Use Default Credentials` to `true`, then I can read the temporary permissions provided by the STS to access AWS related services. Without this environment variable, the `AWSCredentialsProviderControllerService` will report an error: 
```
2023-03-14 15:56:44,700 DEBUG [Timer-Driven Process Thread-2] c.a.auth. AWSCredentialsProviderChain Unable to load credentials from WebIdentityTokenCredentialsProvider: To use assume role profiles the aws-java-sdk- sts module must be on the class path. 
```
I think the key is the AWS_WEB_IDENTITY_TOKEN_FILE environment variable, which the AWS client will use to get the information it needs.

Some related urls:

* https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
* https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

ChrisSamo632:
Looking at #5775, the main difference there is that it tried to introduce the ability to let the user define where the token file was located on the NiFi server/in the NiFi container.

This **can** be defined through the `AWS_WEB_IDENTITY_TOKEN_FILE` environment variable as you indicate @zhangzhonglai, but not everyone will want/be able to use that approach - so the use of the configurable setting within the `AWSCredentialsProviderControllerService` would be a sensible addition and more in-fitting with how NiFi tries to make things configurable.

It might be worth considering rejuvinating @nuttybrewer's PR and include your additional dependenies to allow the use of the default env vars as well

zhangzhonglai:
Yes, I think you are right @ChrisSamo632 . My requirements are the minimum dependency requirements for the default behavior of the AWS client. If a more customized solution is required, #5775 with my additional dependencies would be a good option.

nandorsoma:
@zhangzhonglai, a little addition to what @ChrisSamo632 said. In NiFi, we prefer solutions when we can do explicit configuration. It is easier to test, features are more apparent, less documentation is needed, and most of the time it involves less surprise. However, there are more and more libraries that do some kind of "automagic", but we need to integrate them into NiFi. The main issue is that they are not designed in mind that they need to be part of a framework. Though our goal is to make them feel native to NiFi. If it can be done via workaround, so be it. That was my general take on the topic.

Regarding #5775. We can reference environment variables in properties with expression language, so the PR might work for you without the extra library. For that, we need to set the `ExpressionLanguageScope` of the property to anything other than `NONE`. Preferably to `FLOWFILE_ATTRIBUTES`. Could you test that the PR mentioned solves your issue?

zhangzhonglai:
Sorry for the late reply @nandorsoma.

> Regarding https://github.com/apache/nifi/pull/5775. We can reference environment variables in properties with expression language, so the PR might work for you without the extra library. For that, we need to set the `ExpressionLanguageScope` of the property to anything other than `NONE`. Preferably to `FLOWFILE_ATTRIBUTES`. Could you test that the PR mentioned solves your issue?

As our DevOps recycled the test environment, I currently do not have the conditions in which I can reproduce and test this solution. Perhaps I will test this solution in the future when I am in a position to do so. Maybe you can close this PR first for now.

nandorsoma:
> Sorry for the late reply @nandorsoma.
> 
> > Regarding #5775. We can reference environment variables in properties with expression language, so the PR might work for you without the extra library. For that, we need to set the `ExpressionLanguageScope` of the property to anything other than `NONE`. Preferably to `FLOWFILE_ATTRIBUTES`. Could you test that the PR mentioned solves your issue?
> 
> As our DevOps recycled the test environment, I currently do not have the conditions in which I can reproduce and test this solution. Perhaps I will test this solution in the future when I am in a position to do so. Maybe you can close this PR first for now.

Thank you for the reply @zhangzhonglai. Then I'm closing this PR. If you need help in the future feel free to contact us on the mailing list or over slack. Thanks again!

juldrixx:
I'm trying to do IRSA with NiFi on EKS but it won't work when I set the controller service to use Default Credentials. I just have 403. How did you make it work?

zhangzhonglai:
> I'm trying to do IRSA with NiFi on EKS but it won't work when I set the controller service to use Default Credentials. I just have 403. How did you make it work?

Sorry for the late reply. All my changes to NiFi are in this PR.
The other thing to note is the authorization of the pod. What we use is to add a `ServiceAccount` with `eks.amazonaws.com/role-arn` and give that `ServiceAccount` to the pod by the `serviceAccountName` property.

PR #7051, opened 2023-03-16
NIFI-11255 Allowable value for 'Use s3.region Attribute'

krisztina-zsihovszki:
# Summary

[NIFI-11255](https://issues.apache.org/jira/browse/NIFI-11255)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/AbstractAWSProcessor.java
+++ b/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/AbstractAWSProcessor.java
@@ -30,7 +30,7 @@ import com.amazonaws.regions.Regions;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.conn.ssl.DefaultHostnameVerifier;
 import org.apache.nifi.annotation.lifecycle.OnScheduled;
-import org.apache.nifi.annotation.lifecycle.OnShutdown;
+import org.apache.nifi.annotation.lifecycle.OnStopped;
 import org.apache.nifi.components.AllowableValue;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.ValidationContext;
@@ -66,14 +66,14 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
- * Abstract base class for aws processors.  This class uses aws credentials for creating aws clients
+ * Abstract base class for AWS processors.  This class uses AWS credentials for creating AWS clients
  *
- * @deprecated use {@link AbstractAWSCredentialsProviderProcessor} instead which uses credentials providers or creating aws clients
+ * @deprecated use {@link AbstractAWSCredentialsProviderProcessor} instead which uses credentials providers or creating AWS clients
  * @see AbstractAWSCredentialsProviderProcessor
  *
  */
 @Deprecated
-public abstract class AbstractAWSProcessor<ClientType extends AmazonWebServiceClient> extends AbstractSessionFactoryProcessor {
+public abstract class AbstractAWSProcessor<ClientType extends AmazonWebServiceClient> extends AbstractSessionFactoryProcessor implements AwsClientProvider<ClientType> {
 
     public static final Relationship REL_SUCCESS = new Relationship.Builder().name("success")
             .description("FlowFiles are routed to success relationship").build();
@@ -153,9 +153,6 @@ public abstract class AbstractAWSProcessor<ClientType extends AmazonWebServiceCl
             .addValidator(StandardValidators.URL_VALIDATOR)
             .build();
 
-    protected volatile ClientType client;
-    protected volatile Region region;
-
     protected static final String VPCE_ENDPOINT_SUFFIX = ".vpce.amazonaws.com";
     protected static final Pattern VPCE_ENDPOINT_PATTERN = Pattern.compile("^(?:.+[vpce-][a-z0-9-]+\\.)?([a-z0-9-]+)$");
 
@@ -165,6 +162,8 @@ public abstract class AbstractAWSProcessor<ClientType extends AmazonWebServiceCl
 
     private static final ProxySpec[] PROXY_SPECS = {ProxySpec.HTTP_AUTH};
     public static final PropertyDescriptor PROXY_CONFIGURATION_SERVICE = ProxyConfiguration.createProxyConfigPropertyDescriptor(true, PROXY_SPECS);
+
+    private final AwsClientCache<ClientType> awsClientCache = new AwsClientCache<>();
 
     public static AllowableValue createAllowableValue(final Regions region) {
         return new AllowableValue(region.getName(), region.getDescription(), "AWS Region Code : " + region.getName());
@@ -226,7 +225,6 @@ public abstract class AbstractAWSProcessor<ClientType extends AmazonWebServiceCl
 
         return validationResults;
     }
-
     protected ClientConfiguration createConfiguration(final ProcessContext context) {
         return createConfiguration(context, context.getMaxConcurrentTasks());
     }
@@ -287,7 +285,7 @@ public abstract class AbstractAWSProcessor<ClientType extends AmazonWebServiceCl
 
     @OnScheduled
     public void onScheduled(final ProcessContext context) {
-        setClientAndRegion(context);
+        getConfiguration(context);
     }
 
     /*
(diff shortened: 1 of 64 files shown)


Lehel44 (changes requested):
Thanks @krisztina-zsihovszki for the contribution. I checked if overloading methods that is supposed to be overriden can be avoided but doesn't seem like an option here. I left a few comments.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/s3/AbstractS3Processor.java
```diff
@@ -198,6 +218,88 @@ protected AmazonS3Client createClient(final ProcessContext context, final AWSCre
...
+        String regionValue = context.getProperty(S3_CUSTOM_REGION).getValue();
+        return ATTRIBUTE_DRIVEN_REGION.getValue().equals(regionValue);
+    }
+
+    private static AllowableValue[] getCustomAvailableRegions() {
+        final List<AllowableValue> values = new ArrayList<>();
+        Collections.addAll(values, AbstractAWSCredentialsProviderProcessor.getAvailableRegions());
+        values.add(ATTRIBUTE_DRIVEN_REGION);
+        return values.toArray(new AllowableValue[0]);
+    }
```
Lehel44, line 301:
```suggestion
    private static AllowableValue[] getCustomAvailableRegions() {
    List<AllowableValue> values = new ArrayList<>(Arrays.asList(AbstractAWSCredentialsProviderProcessor.getAvailableRegions()));
    values.add(ATTRIBUTE_DRIVEN_REGION);
    return values.toArray(new AllowableValue[0]);
}
```

Lehel44: We can also use Apache Commons ArrayUtils::addAll

```suggestion
    private static AllowableValue[] getCustomAvailableRegions() {
    AllowableValue[] availableRegions = AbstractAWSCredentialsProviderProcessor.getAvailableRegions();
    AllowableValue[] customRegions = new AllowableValue[]{ATTRIBUTE_DRIVEN_REGION};
    return ArrayUtils.addAll(availableRegions, customRegions);
}
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/PutS3Object.java
```diff
@@ -503,11 +503,21 @@ private synchronized void ageoffLocalState(long ageCutoff) {
...
             return;
         }
 
+       try {
+            setClientAndRegion(context, flowFile.getAttributes());
+        } catch (Exception e) {
+            getLogger().error("Failed to initialize S3 client", e);
+            flowFile = session.penalize(flowFile);
+            session.transfer(flowFile, REL_FAILURE);
+            return;
```
turcsanyip, line 518: @krisztina-zsihovszki  Multiple threads can execute `onTrigger()` in parallel so it is not safe to create and store the client in the `client` field of the processor (`setClientAndRegion()` is doing that).
I suggest implementing client caching similar to [NIFI-10152](https://issues.apache.org/jira/browse/NIFI-10152) for Azure ADLS processors. It would also eliminate the recreation of the client in every onTrigger if the client can be reused (same region).

turcsanyip (changes requested):
@krisztina-zsihovszki Thanks for adding the caching!

I have a few notes regarding the cache key (`AwsClientDetails` class):
- it does not contain the `ClientType` so clients with different type but same credentials, region, etc. would be mapped to the same key
- `AwsClientDetails` contains the common config from `AbstractAWSProcessor` only but some processor implementations add some extra config (like signer in S3) which may affect the shared clients (these should be reflected in the key in order to create distinct clients)
- `AwsClientDetails` needs to be created by the caller and kept consistent with the other parameters which is not straightforward

Just an idea but it may solve all the problems above: add new method `AwsClientProvider.getClientKey()` with the same parameters as `createClient()`
- => the processor implementations are responsible for providing the right key, based on the same parameters they will build the client from, each implementation knows what key should be used
- => different key types mean different client types

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/AbstractAWSProcessor.java
```diff
@@ -425,9 +413,11 @@ protected AWSCredentials getCredentials(final PropertyContext context) {
 
     @OnShutdown
     public void onShutdown() {
-        if ( getClient() != null ) {
+        if (getClient() != null) {
             getClient().shutdown();
         }
+
+        getAwsClientFactory().getClients().forEach(AmazonWebServiceClient::shutdown);
```
turcsanyip, line 420: If we do not store the client separately, only the cached clients need to be shut down.

However, I noticed an old issue with this method: it is hooked on `@OnShutdown`. I believe the intent was `@OnStopped`. Now it runs only at NiFi shutdown which is mostly useless as the JVM will quit and client be destroyed. Also, it shuts down the lastly used client only (clients recreated at start/stop that are not shut down).
As it has not been an issue so far, I think we can simply delete this method for now.
In order to implement the client shutdown properly, we should also handle clients evicted from the cache too. It could be implemented in a separate jira.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/AwsClientProvider.java
```diff
@@ -0,0 +1,53 @@
...
+    ClientType createClient(final ProcessContext context, final AWSCredentials credentials, final ClientConfiguration config);
+
+    /**
+     * Create AWS client using credentials provider. This is the preferred method for creating AWS clients
+     * @param context process context
+     * @param credentialsProvider AWS credentials provider
+     * @param config AWS client configuration
+     * @return ClientType AWS client
+     */
+    ClientType createClient(final ProcessContext context, final AWSCredentialsProvider credentialsProvider, final ClientConfiguration config);
```
turcsanyip, line 45: With moving the cache access logic to `getClient()`, this `AwsClientProvider` interface can expose `createClient(ProcessContext, AwsClientDetails)` only, which is the "original" / "main" method for creating the client. In this case we can get rid of `AWSCredentials`, `AWSCredentialsProvider` and `ClientConfiguration` in this interface (those are rather implementation details in the abstract processors).

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/AwsClientFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import com.amazonaws.ClientConfiguration;
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.nifi.processor.ProcessContext;
+
+public class AwsClientFactory<ClientType extends AmazonWebServiceClient> {
```
turcsanyip, line 29: As `AwsClientProvider` creates the client and this class handles the cache, so I would rename it to `AwsClientCache`.

turcsanyip (changes requested):

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/AbstractAWSProcessor.java
```diff
@@ -423,17 +408,28 @@ protected AWSCredentials getCredentials(final PropertyContext context) {
...
-        final AWSConfiguration awsConfiguration = getConfiguration(context);
-        this.client = awsConfiguration.getClient();
-        this.region = awsConfiguration.getRegion();
+    protected ClientType getClient(final ProcessContext context) {
+        final AwsClientDetails awsClientDetails = new AwsClientDetails(getRegion(context));
+        return getClient(context, awsClientDetails);
+    }
+
+    protected AwsClientCache<ClientType> getAwsClientCache() {
+        return awsClientCache;
```
turcsanyip, line 432: `getAwsClientCache()` is not called or overridden in child classes. The method should be `private` or even the field could be accessed directly instead of the getter method.

turcsanyip (approved):
@krisztina-zsihovszki Thanks for adding this new feature and for all the review changes!
@Lehel44 Thanks for your review!

+1 LGTM
Merging to main and 1.x.

PR #7057, opened 2023-03-19
NIFI-11302: add Registry version to About dialog fixing bug for Java 11

markobean:
# Summary
This PR fixes a bug introduced with NIFI-8899, adding NiFi Registry version information to the About dialog. The previous implementation works only with Java 8. This implementation was confirmed to work correctly with Java 11 by installing Registry and observing the About dialog after building and running using JDK 11.

[NIFI-11302](https://issues.apache.org/jira/browse/NIFI-11302)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/RegistryAboutResource.java
```diff
@@ -56,8 +59,13 @@ public RegistryAboutResource(
...
     )
     public Response getVersion() {
-        final String implVersion = NiFiRegistryApiApplication.class.getPackage().getImplementationVersion();
-        final RegistryAbout version = new RegistryAbout(implVersion);
-        return Response.status(Response.Status.OK).entity(version).build();
+        Properties props = new Properties();
+        try (InputStream stream = getClass().getResourceAsStream(PROPERTIES_FILE)){
+            props.load(stream);
+        } catch (IOException e) {
+            return Response.status(Response.Status.OK).entity("unknown").build();
```
mh013370, line 66: Should we log that an error occurred here so it would be possible to triage why "unknown" appears as a version?

exceptionfactory (commented):
Thanks for working to correct this problem @markobean!

The current approach based on [Package.getImplementationVersion()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Package.html#getImplementationVersion()) does not seem to be inherently problematic on Java 11, but rather reflects the lack of package information in the JAR.

Did you evaluate whether the existing approach could work with updates to the Maven configuration so that the necessary manifest information is generated?

markobean:
> Thanks for working to correct this problem @markobean!
> 
> The current approach based on [Package.getImplementationVersion()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Package.html#getImplementationVersion()) does not seem to be inherently problematic on Java 11, but rather reflects the lack of package information in the JAR.
> 
> Did you evaluate whether the existing approach could work with updates to the Maven configuration so that the necessary manifest information is generated?

I did not consider updating the manifest information. I made the assumption the missing information was inherently part of default Java 11 packaging and implemented an approach that would be independent of the availability of such information.

exceptionfactory:
Thanks for the reply @markobean.

I recommend keeping the existing approach since uses standard Java package manifests as opposed to loading properties.

There are some options for where to make the changes, but after some testing, adjusting the Maven JAR plugin configuration for `nifi-registry-data-model` seems like the best location. It did require moving the the `RegistryAbout` model class to a subpackage to avoid confusion with the standard `org.apache.nifi.registry` package, but the following commit preserves the same basic approach using `getPackage().getImplementationVersion()`.

https://github.com/exceptionfactory/nifi/commit/6c4386ec654eeff8e11af8b40db946cac8cd9c8f

exceptionfactory (approved):
Thanks for incorporating the changes @markobean, the latest version looks good! +1 merging

PR #7070, opened 2023-03-22
NIFI-11319 Fixing TestStatelessBootstrap::testClassNotAllowed

simonbence:
The test was broken. My understaning is that the test supposed to ensure that classes from outside the allowed NAR directory should not be loaded expect if they are part of a spefic set of classes (java or jdk "modules"). The original configuration of the test used the "target" folder as NAR folder, which contained every product of the build, including the class "org.apache.nifi.stateless.bootstrap.RunStatelessFlow" used as test subject, thus the class was loaded and the test broke. I narrowed down the "nar folder" in order to not have the test subject included, this I would believe solves the issue.

[NIFI-11319](https://issues.apache.org/jira/browse/NIFI-11319)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


markap14:
Thanks @simonbence that does appear to address the issue. Interestingly, I didn't encounter it when running `mvn clean install` because all of the directories under `target` hadn't yet been completed when that runs. But if you run in the IDE after building or run via `mvn test` after the build has completed, it failed. +1 merged to main and support branches

PR #7076, opened 2023-03-23
NIFI-11294 - Add support for component state as checkpointing strategy

malthe:
# Summary

[NIFI-11294](https://issues.apache.org/jira/browse/NIFI-11294) ConsumeAzureEventHub should default to processor state for checkpointing

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/ConsumeAzureEventHub.java
+++ b/nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/ConsumeAzureEventHub.java
@@ -338,6 +338,7 @@ public class ConsumeAzureEventHub extends AbstractSessionFactoryProcessor {
     private volatile boolean isRecordReaderSet = false;
     private volatile boolean isRecordWriterSet = false;
     private volatile String serviceBusEndpoint;
+    private volatile CheckpointStrategy checkpointStrategy;
 
     @Override
     protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
@@ -402,6 +403,8 @@ public class ConsumeAzureEventHub extends AbstractSessionFactoryProcessor {
             isRecordWriterSet = StringUtils.isNotEmpty(newValue);
         } else if (SERVICE_BUS_ENDPOINT.equals(descriptor)) {
             serviceBusEndpoint = newValue;
+        } else if (CHECKPOINT_STRATEGY.equals(descriptor)) {
+            checkpointStrategy = CheckpointStrategy.valueOf(newValue);
         }
     }
 
@@ -441,10 +444,6 @@ public class ConsumeAzureEventHub extends AbstractSessionFactoryProcessor {
         final String eventHubName = context.getProperty(EVENT_HUB_NAME).evaluateAttributeExpressions().getValue();
         final String consumerGroup = context.getProperty(CONSUMER_GROUP).evaluateAttributeExpressions().getValue();
 
-        final CheckpointStrategy checkpointStrategy = CheckpointStrategy.valueOf(
-                context.getProperty(CHECKPOINT_STRATEGY).getValue()
-        );
-
         CheckpointStore checkpointStore;
         final Map<String, EventPosition> legacyPartitionEventPosition;
 
@@ -458,7 +457,18 @@ public class ConsumeAzureEventHub extends AbstractSessionFactoryProcessor {
             checkpointStore = new BlobCheckpointStore(blobContainerAsyncClient);
             legacyPartitionEventPosition = getLegacyPartitionEventPosition(blobContainerAsyncClient, consumerGroup);
         } else {
-            checkpointStore = new ComponentStateCheckpointStore(processSessionFactory);
+            checkpointStore = new ComponentStateCheckpointStore(
+                    new ComponentStateCheckpointStore.State() {
+                        @Override
+                        public Map<String, String> getState() throws IOException {
+                            return processSessionFactory.createSession().getState(Scope.CLUSTER).toMap();
+                        }
+
+                        @Override
+                        public void setState(Map<String, String> map) throws IOException {
+                            processSessionFactory.createSession().setState(map, Scope.CLUSTER);
+                        }
+                    });
             legacyPartitionEventPosition = Collections.emptyMap();
         }
 
@@ -528,8 +538,34 @@ public class ConsumeAzureEventHub extends AbstractSessionFactoryProcessor {
                 writeRecords(eventBatchContext, session, stopWatch);
             }
 
-            // Commit ProcessSession and then update Azure Event Hubs checkpoint status
-            session.commitAsync(eventBatchContext::updateCheckpoint);
+            // As a special case, when the checkpoint strategy is component state,
+            // we reuse the current session.
+            if (checkpointStrategy == CheckpointStrategy.COMPONENT_STATE) {
+                eventBatchContext = new EventBatchContext(
+                        eventBatchContext.getPartitionContext(),
+                        eventBatchContext.getEvents(),
+                        new ComponentStateCheckpointStore(
+                                new ComponentStateCheckpointStore.State() {
+                                    @Override
+                                    public Map<String, String> getState() throws IOException {
+                                        return session.getState(Scope.CLUSTER).toMap();
+                                    }
+
+                                    @Override
+                                    public void setState(Map<String, String> map) throws IOException {
+                                        session.setState(map, Scope.CLUSTER);
+                                    }
+                                }
+                        ),
+                        eventBatchContext.getLastEnqueuedEventProperties()
+                );
+                // Update checkpoint and then commit changes, sharing the session.
+                eventBatchContext.updateCheckpoint();
+                session.commitAsync();
+            } else {
+                // Commit ProcessSession and then update Azure Event Hubs checkpoint status
+                session.commitAsync(eventBatchContext::updateCheckpoint);
+            }
         } catch (final Exception e) {
             final PartitionContext partitionContext = eventBatchContext.getPartitionContext();
             getLogger().error("Event Batch processing failed Namespace [{}] Event Hub [{}] Consumer Group [{}] Partition [{}]",
--- a/nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/ConsumeAzureEventHub.java
+++ b/nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/ConsumeAzureEventHub.java
@@ -16,6 +16,7 @@
  */
 package org.apache.nifi.processors.azure.eventhub;
 
+import com.azure.core.amqp.AmqpClientOptions;
 import com.azure.core.credential.AzureNamedKeyCredential;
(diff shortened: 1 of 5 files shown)


malthe:
Note that the existing tests (which I have adapted, somewhat unsuccessfully) are not very good – they do not actually exercise the processor's logic end-to-end, but only the "meat". I think the whole test setup for this processor really needs to be reworked entirely.

turcsanyip:
Will review...

exceptionfactory (commented):
Thanks for implementing this feature @malthe! Just noting that the automated builds flagged a couple Checkstyle violations, and the unit tests failed due to unnecessary stubbing. That usually means there are unnecessary mock calls that need to be removed or adjusted so the `when` expectations are only called when needed.

malthe:
@exceptionfactory thanks – these issues have been fixed now.

The good is that the actual code seems pretty well-functioning and reasonable well-crafted as well. However, the tests are bad/wrong and I fear I have only made them worse.

malthe:
@turcsanyip I have now implemented proper load-balancing in a9bd297b50515918092ec01a0730ee795e917932 – and to fix the issues with stopping the processor, I have removed most of the processor state fields, shifting them into a event batch processing closure instead.

nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/ConsumeAzureEventHub.java
```diff
@@ -486,11 +563,36 @@ protected String getTransitUri(final PartitionContext partitionContext) {
...
+                                new ComponentStateCheckpointStore.State() {
+                                    public StateMap getState() throws IOException {
+                                        return session.getState(Scope.CLUSTER);
+                                    }
+
+                                    public boolean replaceState(StateMap oldValue, Map<String, String> newValue) {
+                                        return false;
+                                    }
+                                }
+                        ),
```
exceptionfactory, line 586: The inner anonymous class results in multiple levels of nesting that make this hard to follow. Recommend breaking it out to an explicit class.

nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/ConsumeAzureEventHub.java
```diff
@@ -403,36 +420,76 @@ public void stopClient() {
...
+        if (checkpointStrategy == CheckpointStrategy.AZURE_BLOB_STORAGE) {
+            final String containerName = defaultIfBlank(context.getProperty(STORAGE_CONTAINER_NAME).evaluateAttributeExpressions().getValue(), eventHubName);
+            final String storageConnectionString = createStorageConnectionString(context);
+            final BlobContainerAsyncClient blobContainerAsyncClient = getBlobContainerAsyncClient(containerName, storageConnectionString);
+            checkpointStore = getCheckpointStoreFromBlobContainer(blobContainerAsyncClient);
+            legacyPartitionEventPosition = getLegacyPartitionEventPosition(blobContainerAsyncClient, consumerGroup);
+        } else {
+            checkpointStore = new ComponentStateCheckpointStore(
+                    identifier,
+                    new ComponentStateCheckpointStore.State() {
```
exceptionfactory, line 447: Having two different anonymous inner class implementations make this hard to follow. Recommend breaking this out to a distinct class.

exceptionfactory (changes requested):
Thanks for the updates on this new feature @malthe! I noted several test implementation concerns, and a couple recommendations on the implementation itself.

nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/utils/ComponentStateCheckpointStore.java
```diff
@@ -0,0 +1,228 @@
...
+                            KEY_OWNERSHIP,
+                            partitionOwnership.getFullyQualifiedNamespace(),
+                            partitionOwnership.getEventHubName(),
+                            partitionOwnership.getConsumerGroup(),
+                            partitionOwnership.getPartitionId()
+                    );
+
+                    final Map<String, String> newState = new HashMap<>(oldState.toMap());
+                    long timestamp = System.currentTimeMillis();
+                    String eTag = identifier + "/" + timestamp;
```
exceptionfactory, line 92: Should the `eTag` be created based on the state version as opposed the current timestamp? That seems like it would align better with the purpose of the tag for tracking ownership and ensuring consistent version control across nodes.

nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/utils/ComponentStateCheckpointStore.java
```diff
@@ -0,0 +1,228 @@
...
+            final String fullyQualifiedNamespace = parts[1];
+            final String eventHubName = parts[2];
+            final String consumerGroup = parts[3];
+            final String partitionId = parts[4];
+            PartitionContext partitionContext = new PartitionContext(
+                    fullyQualifiedNamespace,
+                    eventHubName,
+                    consumerGroup,
+                    partitionId
+            );
```
exceptionfactory, line 211: Recommend pulling this logic out to a separate method so the formatting approach is clear.

nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/ConsumeAzureEventHub.java
```diff
@@ -403,36 +420,76 @@ public void stopClient() {
...
+                        public StateMap getState() throws IOException {
+                            final ProcessSession session = processSessionFactory.createSession();
+                            return session.getState(Scope.CLUSTER);
+                        }
+
+                        public boolean replaceState(StateMap oldValue, Map<String, String> newValue) throws IOException {
+                            final ProcessSession session = processSessionFactory.createSession();
+                            if (!session.replaceState(oldValue, newValue, Scope.CLUSTER)) {
```
turcsanyip, line 455: `ProcessSession.replaceState()` does not provide fully proper optimistic locking because the state can be changed by another session between `replaceState()` and `commit()`. So even if `replaceState()` returns `true`, the state change may be omitted at commit time due to a concurrent update.

I suggest using `ProcessContext.replace()` instead which has the right optimistic locking semantics which is required from a checkpoint store implementation.

Please note `ProcessContext.replace()` cannot initialize the state currently (see also [NIFI-11595](https://issues.apache.org/jira/browse/NIFI-11595)) so it needs to be created with `setState()` before using `replace()`. E.g. in `@OnScheduled` like this:
```
    @OnScheduled
    public void onScheduled(ProcessContext context) throws IOException {
        if (getNodeTypeProvider().isPrimary()) {
            final StateManager stateManager = context.getStateManager();
            final StateMap state = stateManager.getState(Scope.CLUSTER);

            if (!state.getStateVersion().isPresent()) {
                stateManager.setState(new HashMap<>(), Scope.CLUSTER);
            }
        }
    }
```

turcsanyip (changes requested):
@malthe I tried the processor but it seems to process all messages again and again after restarts. I cannot see the checkpoint data saved in the state (only ownership) so I think this is the root cause. Could you please check it?

nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/test/java/org/apache/nifi/processors/azure/eventhub/TestConsumeAzureEventHub.java
```diff
@@ -117,13 +131,22 @@ public class TestConsumeAzureEventHub {
...
+    private ArgumentCaptor<Consumer<EventBatchContext>> eventBatchProcessorCapture;
+
     private MockConsumeAzureEventHub processor;
 
     private TestRunner testRunner;
 
     @BeforeEach
     public void setupProcessor() {
-        processor = new MockConsumeAzureEventHub();
+        processor = spy(new MockConsumeAzureEventHub());
```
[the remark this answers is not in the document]
malthe: Previously, the processor wasn't tested through the `onTrigger` but through some internal methods.

I changed the tests so that all testing happens through the actually exposed interface. However, this requires some additional mocking in order to implement the checkpoint strategy for example.

(The reason being, we don't have the luxury of testing against a real event hub here.)

But instead of spying on the processor, I suppose I could implement some alternative methods on the mock class. I don't remember if I tried that and faced some issues perhaps.

nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/test/java/org/apache/nifi/processors/azure/eventhub/utils/TestComponentStateCheckpointStore.java
```diff
@@ -0,0 +1,154 @@
...
+    public void testClaimOwnershipSuccess() {
+        var claimed = checkpointStore.claimOwnership(List.of(
+            partitionOwnership
+        )).blockFirst();
+        var listed = checkpointStore.listOwnership(
+                EVENT_HUB_NAMESPACE,
+                EVENT_HUB_NAME,
+                CONSUMER_GROUP
+        ).blockFirst();
+        assert listed.getETag().equals(claimed.getETag());
```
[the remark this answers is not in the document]
malthe: Fixed in 4aa8240c0ed8dc52dedcd7e4463e1672d5abf34d.

nifi-nar-bundles/nifi-azure-bundle/nifi-azure-processors/src/main/java/org/apache/nifi/processors/azure/eventhub/utils/ComponentStateCheckpointStore.java
malthe: I'm not sure – but that logic was copied from the blob-based checkpoint store.

In fact, whenever I could, I have used the exact same logic between them.

turcsanyip:
@malthe Just a heads-up that [NIFI-11595](https://issues.apache.org/jira/browse/NIFI-11595) has been merged fixing the `replaceState()` logic I mentioned in [my previous comment](https://github.com/apache/nifi/pull/7076#discussion_r1206003566). It makes the state handling of the checkpoints simpler, though it was not a blocker.

Considering that there was no activity on the PR in the past months, I would like to ask if you are still interested in this feature? Do you intend to finish the PR? I'm happy to jump in and continue the implementation based on your PR if you have other interests now.

Thanks

malthe:
@turcsanyip please do – I am currently not actively using NiFi so just following along from the sideline.

turcsanyip:
Thanks @malthe! I'll go ahead and try to finish the implementation.

exceptionfactory:
Thanks for your work on this @malthe, and thanks for planning to take this on @turcsanyip. I am closing the pull request for now given the state of several conflicts, but feel free to open a new one when it is ready for review.

PR #7092, opened 2023-03-28
NIFI-11327 Add Export/Import All - NiFi CLI - NiFi Registry

timeabarna:
# Summary
Have a command registry export-all-flows that does the following:
List all the buckets, for each bucket, list all flows, for each flow, list all versions and export each version. All files should be landing in a target directory provided as an argument of the function.

Have a command registry import-all-flows that does the following:
It takes a directory as input (the one created by the export-all-flows command), and goes through the files to create the corresponding buckets, flows and flows versions.
The original author, bucket id and flow id need to be kept.

Use cases:
use case 1: NiFi 1 -> connecting to NiFi Registry 1, "re-initialising" an existing NiFi Registry, the NiFi Registry does not change, only its configuration.
export everything
change the NR flow definition repo backend from local FS to database (for example), or from git to database, etc
import everything
the existing NiFi instance should not see any change
use case 2: NiFi 1 -> Registry 1, NiFi 2 -> Registry 2, Disaster Recovery kind of thing between two different sites.
export everything from NR1
import everything into NR2
If the import into NR2 is adding new versions, then NiFi 2 should be able to update an existing PG to a newer version of the flow.

[NIFI-11327](https://issues.apache.org/jira/browse/NIFI-11327)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -295,6 +295,48 @@ public Response createFlowVersion(
         return Response.status(Response.Status.OK).entity(createdSnapshot).build();
     }
 
+    @POST
+    @Path("{flowId}/versions/migrate")
```
exceptionfactory, line 299: Having a verb `migrate` in the path does not follow general REST API conventions, where the path should be a noun and the HTTP method is the verb. One option could be to make `migrate` and `Boolean` query parameter on the `/versions` resource method. However, the word `migrate` itself does not clearly convey what is happening, so perhaps another parameter name would be better. As the method name mentions keeping the author, it seems like a better approach would be to adding a query parameter named something like `preserveAuthor`. If there are other properties to be preserved, perhaps it could be named `preserveSourceProperties`.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/CommandOption.java
```diff
@@ -164,7 +164,9 @@ public enum CommandOption {
     OUTPUT_TYPE("ot", "outputType", "The type of output to produce (json or simple)", true),
     VERBOSE("verbose", "verbose", "Indicates that verbose output should be provided", false),
     RECURSIVE("r", "recursive", "Indicates the command should perform the action recursively", false),
-    HELP("h", "help", "Help", false)
+    HELP("h", "help", "Help", false),
+    SKIP("skip", "skip", "Indicates to skip an operation if target object exists", true),
+    KEEP("keep", "keep", "Indicates to keep the original format or type of an object", true)
```
exceptionfactory, line 169: Similar to `skip`, the word `keep` by itself does not indicate the meaning. Perhaps `preserve-format` or `preserve-type` could be better.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ExportAllFlows.java
```diff
@@ -0,0 +1,182 @@
...
+        final Map<String, Bucket> bucketMap = getBucketMap(client, isInteractive);
+
+        // Gather all flows and create a map for quick access by flow id
+        final Map<String, VersionedFlow> flowMap = getFlowMap(client, bucketMap, isInteractive);
+
+        // Gather all versions for all the flows
+        final List<VersionedFlowSnapshotMetadata> versionedFlowSnapshotMetadataList = getVersionedFlowSnapshotMetadataList(client, flowMap, isInteractive);
+
+        // Prepare flow version exports
+        final List<VersionedFlowSnapshot> versionedFlowSnapshotList = getVersionedFlowSnapshotResults(client, outputDirectory, bucketMap, flowMap, versionedFlowSnapshotMetadataList, isInteractive);
```
exceptionfactory, line 90: Are there potential memory concerns with this approach? What happens if there are hundreds of large flows?

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ImportAllFlows.java
```diff
@@ -0,0 +1,298 @@
...
+                    .collect(Collectors.toList());
+        } catch (Exception e) {
+            throw new NiFiRegistryException("File listing not possible due to ", e);
+        }
+        return files;
+    }
+
+    private List<ImportedSnapshot> deserializeSnapshots(final List<String> files) throws IOException {
+        final List<ImportedSnapshot> importedSnapshots = new ArrayList<>();
+        final ObjectMapper mapper = JacksonUtils.getObjectMapper();
```
exceptionfactory, line 230: `ObjectMapper` is thread safe, so this can be declared as a member variable on the class.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ImportAllFlows.java
```diff
@@ -0,0 +1,298 @@
...
+        final List<String> files;
+
+        try (final Stream<Path> paths = Files.list(Paths.get(directory))) {
+            files = paths
+                    .filter(Files::isRegularFile)
+                    .filter(path -> path.getFileName().toString().startsWith(FILE_NAME_PREFIX))
+                    .map(Path::toString)
+                    .collect(Collectors.toList());
+        } catch (Exception e) {
+            throw new NiFiRegistryException("File listing not possible due to ", e);
```
exceptionfactory, line 223: This approach to error messages needs to be changed. The message should stand by itself, it will not be concatenated with the exception. Recommend the following:
```suggestion
            throw new NiFiRegistryException("File listing failed", e);
```

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ImportAllFlows.java
```diff
@@ -0,0 +1,298 @@
...
+        final Map<Pair<String, String>, String> flowCreated = new HashMap<>();
+
+        // Gather all flow versions and create a map for easier search by flow id
+        final Map<String, List<Integer>> versionMap = getVersionMap(client, flowMap, isInteractive);
+
+        // Create file path list
+        final List<String> files = getFilePathList(properties);
+
+        // Deserialize file content
+        final List<ImportedSnapshot> importedSnapshots = deserializeSnapshots(files);
```
exceptionfactory, line 114: Instead of serializing all snapshots, is it possible to iterate through each one? It looks like that could create problems for sorting, so perhaps there is no other way to avoid loading everything.

exceptionfactory (commented):
Thanks for working on this feature @timeabarna. I noted several general questions regarding the naming of the new REST Resource and potential concerns related to reading everything into memory. See the individual comments for details.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
bbende: We had discussed this offline and landed on the `/migrate` approach to clearly separate the usage of the API. The idea was that no one should really be calling the `/migrate` end-point, except for a specific utility that wants to move data between two registries. 

However, I agree that it generally does not follow REST standards, so I'm not that opposed to using a query parameter on the existing end-points. If we were to go with a query parameter then it should probably be `preserveSourceProperties` because it is more general than just the `author`. 

We have 3 different entities that would be migrated:

- `Buckets` - need to preserve id, rather than generating new
- `Flows` - need to preserve id, rather than generating new
- `Flow Versions` - need to preserve id and author, rather than generating new and using authenticated user

exceptionfactory: Thanks for the reply and background @bbende.

Unfortunately JAX-RS does not appear to support method-based matching using Request query parameters, but it seems like introducing a `preserveSourceProperties` parameter in the existing method would provide a clear separation of behavior.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ImportAllFlows.java
bbende: I'm wondering if we can change the filenames that are produced by export to contain the 3 fields below that we are sorting on, then we can sort the `List<String> files` in the same order?

So exported filenames would be:
```
<Bucket-Name>-<Flow-Name>-<Version>.json
```
This would also have the added benefit of being human readable for anyone looking at the exported files.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ExportAllFlows.java
bbende: I think the way the result classes worked kind of led it to be implemented this way, but I agree with David's concern that we should see if we can write out the snapshots one-by-one. Maybe we can pass some kind iterator/helper object to the result class that allow it to keep calling next() and writing out the result?

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ImportAllFlows.java
```diff
@@ -0,0 +1,298 @@
...
+                } else {
+                    printMessage(isInteractive, flowName + SKIPPING_FLOW_CREATION);
+                }
+            } else if (!flowCreated.containsKey(new ImmutablePair<>(bucketId, flowName))) {
+                createFlow(client, flowCreated, flowId, flowName, flowDescription, bucketId);
+            }
+
+            // Create missing flow versions
+            if (!versionMap.getOrDefault(flowId, Collections.emptyList()).contains(flowVersion)) {
+                createFlowVersion(client, inputSource, flowId);
```
bbende, line 156: What is the expectation for handling nested versioning during a migration?

For example, `Outer PG` under version control, and inside it there is `Child PG` also under version control. In registry, the snapshot of `Outer PG` will have a process group for `Child PG` with coordinates pointing to the registry it came from:
```
"versionedFlowCoordinates" : {
        "storageLocation" : "http://localhost:18080/nifi-registry-api/buckets/876e0978-ffe5-43c5-ab2c-69dd0a8d3c52/flows/412512c1-c95f-4a2a-9df3-8a988e326df7/versions/1",
        "registryUrl" : "http://localhost:18080",
        "bucketId" : "876e0978-ffe5-43c5-ab2c-69dd0a8d3c52",
        "flowId" : "412512c1-c95f-4a2a-9df3-8a988e326df7",
        "version" : 1
      }
```

If these commands were being used to replicate/move between two different registries at different locations, then when moving `Outer PG` to the new registry, it would need to update `storageLocation` and `registryUrl` to the location of the new registry. Otherwise when a NiFi instance pulls `Outer PG` from the new registry, it will then come to 'Child PG' and try to pull it from the old registry, or fail because no registry client exists for the URL of the old registry.

timeabarna:
Hello @bbende and @exceptionfactory,

Thanks you very much for your review. I've updated the code based on your recommendations, improved serialisation and added storage location update.

Can you please check again?

Thanks in advance.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -295,6 +295,48 @@ public Response createFlowVersion(
         return Response.status(Response.Status.OK).entity(createdSnapshot).build();
     }
 
+    @POST
+    @Path("{flowId}/versions/preserveSourceProperties")
```
bbende, line 299: I think David's suggestion was that we should add an optional query param to the existing create method, so we wouldn't need this whole new method and instead a client could specify `?preserveSourceProperties=true` 

@exceptionfactory is that what you were thinking?

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketResource.java
```diff
@@ -92,6 +92,32 @@ public Response createBucket(
         return Response.status(Response.Status.OK).entity(createdBucket).build();
     }
 
+    @POST
+    @Path("migrate")
```
bbende, line 96: Same as previous comment about using a query param on the existing create bucket and being consistent with calling it `preserveSourceProperties`

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/service/ServiceFacade.java
```diff
@@ -70,6 +70,8 @@ public interface ServiceFacade {
 
     Bucket deleteBucket(String bucketIdentifier, RevisionInfo revisionInfo);
 
+    Bucket migrateBucket(Bucket bucket);
```
bbende, line 73: This could go away, or could become `createBucket(Bucket bucket, boolean preserveSourceProperties)`

nifi-registry/nifi-registry-core/nifi-registry-provider-api/src/main/java/org/apache/nifi/registry/hook/EventType.java
```diff
@@ -41,6 +41,14 @@ public enum EventType {
             EventFieldName.VERSION,
             EventFieldName.USER,
             EventFieldName.COMMENT),
+
+    MIGRATE_FLOW_VERSION(
```
bbende, line 45: See later comments about query param approach... If we use the query param approach on the existing end-point then I think this probably goes away

nifi-registry/nifi-registry-core/nifi-registry-client/src/main/java/org/apache/nifi/registry/client/BucketClient.java
```diff
@@ -36,6 +36,16 @@ public interface BucketClient {
...
 
+    /**
+     * Creates the given bucket. If migration parameter is true the bucket will be created with the provided identifier.
+     * False will set a generated identifier.
+     *
+     * @param bucket the bucket to create
+     * @param migration whether the operation is migration related
+     * @return the created bucket with the populated identifier
+     */
+    Bucket create(Bucket bucket, boolean migration) throws NiFiRegistryException, IOException;
```
bbende, line 47: Should call this `preserveSourceProperties` to be consistent

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/service/StandardServiceFacade.java
```diff
@@ -202,6 +202,22 @@ public Bucket deleteBucket(final String bucketIdentifier, final RevisionInfo rev
                 () -> registryService.deleteBucket(bucketIdentifier));
     }
 
+    @Override
+    public Bucket migrateBucket(final Bucket bucket) {
```
bbende, line 206: I think this method can go away right? It would just use the regular create

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ImportAllFlows.java
```diff
@@ -0,0 +1,330 @@
...
+
+            // Create missing flow versions
+            if (!versionMap.getOrDefault(flowId, Collections.emptyList()).contains(flowVersion)) {
+                //update storage location
+                final String registryUrl = getRequiredArg(properties, CommandOption.URL);
+
+                updateStorageLocation(snapshot.getFlowContents().getVersionedFlowCoordinates(), registryUrl);
+                for (VersionedProcessGroup processGroup : snapshot.getFlowContents().getProcessGroups()) {
+                    updateStorageLocation(processGroup.getVersionedFlowCoordinates(), registryUrl);
+                }
```
bbende, line 166: Right now I think this will only handle the root group and first level groups, but there could be a child group under version control at any arbitrary level. 
```
Main PG
  - PG Level 1 (not under VC)
    - PG Level 2 (not under VC) 
      - PG Level 3 (Version Controlled) 
```

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
exceptionfactory: Yes, that's what I had in mind, not a separate path.

timeabarna:
Thanks @bbende and @exceptionfactory for your help, I've updated the PR.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/result/registry/VersionedFlowSnapshotsResult.java
```diff
@@ -0,0 +1,77 @@
...
+        while (versionedFlowSnapshots.hasNext()) {
+            final VersionedFlowSnapshot versionedFlowSnapshot = versionedFlowSnapshots.next();
+            if (exportDirectoryName != null) {
+                final String bucketName = versionedFlowSnapshot.getBucket().getName().replaceAll(SEPARATOR, REPLACEMENT);
+                final String flowName = versionedFlowSnapshot.getFlow().getName().replaceAll(SEPARATOR, REPLACEMENT);
+                final int version = versionedFlowSnapshot.getSnapshotMetadata().getVersion();
+                final String exportFileName = String.format(EXPORT_FILE_NAME, exportDirectoryName, FILE_NAME_PREFIX, bucketName, flowName, version);
+                try (final OutputStream resultOut = Files.newOutputStream(Paths.get(exportFileName))) {
+                    JacksonUtils.write(versionedFlowSnapshot, resultOut);
+                }
```
bbende, line 71: If the output directory doesn't exist, currently the export fails, but it isn't very clear why...
```
#> registry export-all-flows --outputDirectory /tmp/registry-export -verbose

All buckets collected...


All flows collected...


All flow versions collected...


ERROR: /tmp/registry-export/toolkit_registry_export_all_B1_Resuable_1
```

With adding `-verbose` we can see:
```
java.nio.file.NoSuchFileException: /tmp/registry-export/toolkit_registry_export_all_B1_Resuable_1
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:219)
	at java.base/java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:478)
	at java.base/java.nio.file.Files.newOutputStream(Files.java:220)
	at org.apache.nifi.toolkit.cli.impl.result.registry.VersionedFlowSnapshotsResult.write(VersionedFlowSnapshotsResult.java:69)
	at org.apache.nifi.toolkit.cli.impl.command.CommandProcessor.processCommand(CommandProcessor.java:256)
	at org.apache.nifi.toolkit.cli.impl.command.CommandProcessor.processGroupCommand(CommandProcessor.java:233)
	at org.apache.nifi.toolkit.cli.impl.command.CommandProcessor.process(CommandProcessor.java:188)
	at org.apache.nifi.toolkit.cli.CLIMain.runInteractiveCLI(CLIMain.java:124)
	at org.apache.nifi.toolkit.cli.CLIMain.main(CLIMain.java:68)
```

Match we should add something here like:
```
catch(Exception e) {
  throw new RuntimeException("Unable to write flow snapshot to: " + exportFileName, e);
}
```
?

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ImportAllFlows.java
```diff
@@ -0,0 +1,331 @@
...
+    @Override
+    public String getDescription() {
+        return "From a provided directory as input, the directory content must be generated by the export-all-flows command, " +
+                "based on the file contents, the corresponding buckets, flows and flow versions will be created." +
+                "If not configured otherwise, already existing objects will be skipped.";
+    }
+
+    @Override
+    public StringResult doExecute(final NiFiRegistryClient client, final Properties properties) throws IOException, NiFiRegistryException, ParseException, CommandException {
+        final boolean skip = Boolean.parseBoolean(getRequiredArg(properties, CommandOption.SKIP_EXISTING));
```
bbende, line 99: It seems like `SKIP_EXISTING` really only relates to skipping the import of versions for a flow when the flow already exists. So say my registry already has `Flow 1` with `v1` and `v2` and now I am trying to do an import that has the same flow, but has one additional version of `v3`. If I specify `--skipExisting true` then `v3` does not get imported, and if specify `--skipExisting false` then it does get imported.

Is there a situation in one of the uses cases where you would want to use `--skipExisting true` and not get `v3` imported?

If we do keep `--skipExisting`, I am thinking it might be clearer if the way it was used was to make it a non-required argument, and if it is not specified then the behavior is as if it was `--skipExisting false`. If they do specify it, then it can be an argument with no value and specifying `--skipExisting` means `--skipExisting true`.

Default usage would just be:
```
registry import-all-flows -i /path/to/export
```

Thoughts?

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/registry/flow/ImportAllFlows.java
```diff
@@ -0,0 +1,331 @@
...
+            final VersionedFlowSnapshot snapshot = MAPPER.readValue(fileContent, VersionedFlowSnapshot.class);
+
+            final String bucketName = snapshot.getBucket().getName();
+            final String bucketDescription = snapshot.getBucket().getDescription();
+            final String flowName = snapshot.getFlow().getName();
+            final String flowDescription = snapshot.getFlow().getDescription();
+            final int flowVersion = snapshot.getSnapshotMetadata().getVersion();
+            // The original bucket and flow ids must be kept otherwise NiFi won't be able to synchronize with the NiFi Registry
+            final String flowId = snapshot.getFlow().getIdentifier();
+            final String bucketId = snapshot.getBucket().getIdentifier();
```
bbende, line 135: Could we print a message at the beginning and end of this loop like?
```
Importing <flow Name> - <flowVersion> to <bucketName>
...
Successfully imported <flow Name> - <flowVersion> to <bucketName>
```
I think this would help keep track of the current messages to know where it is during the import.
I guess the successful message would only print if not skipping.

bbende (commented):
Latest updates look good, this is very close.

nifi-docs/src/main/asciidoc/toolkit-guide.adoc
```diff
@@ -768,6 +770,325 @@ This command migrates nifi.properties back from AWS_KMS to AES_GCM protection sc
...
+
+1. Export versions:
+
+ ./bin/cli.sh registry export-all-flows -u http://nifi-registry-1:18080 --outputDirectory "/my_dir/flow_exports"
+
+2. Import versions:
+
+ ./bin/cli.sh registry import-all-flows -u http://nifi-registry-2:18080 --input "/my_dir/flow_exports" --skipExisting false
+
+
```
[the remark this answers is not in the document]
timeabarna: Thanks for catching this, incorrectly resolved merge conflict.

bbende (approved):
Thanks for all the updates, this looks good to me. Everything is working as described, going to merge.

PR #7116, opened 2023-04-03
NIFI-11380: Refactor CaptureChangeMySQL with improvements

mattyb149:
# Summary

[NIFI-11380](https://issues.apache.org/jira/browse/NIFI-11380) This PR makes a number of improvements to the readability, maintainability, and capability of the CaptureChangeMySQL processor. See the Jira for a detailed list

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-cdc/nifi-cdc-mysql-bundle/nifi-cdc-mysql-processors/src/main/java/org/apache/nifi/cdc/mysql/event/handler/BinlogEventHandler.java
```diff
@@ -0,0 +1,39 @@
...
+
+public interface BinlogEventHandler<T extends EventData, S extends BinlogEventInfo> {
+
+    void handleEvent(T eventData,
+                     final boolean writeEvent,
+                     DataCaptureState dataCaptureState,
+                     CaptureChangeMySQL.BinlogResourceInfo binlogResourceInfo,
+                     CaptureChangeMySQL.BinlogEventState binlogEventState,
+                     final String sql,
+                     AbstractBinlogEventWriter<S> eventWriter,
```
nandorsoma, line 35: It seems like there is a 1:1 connection between EventHandlers and EventWriters. So InsertEventHandler will always use InsertEventWriter. Is there a specific reason then that the eventWriter is coming from "outside"? Am I missing something?

nifi-nar-bundles/nifi-cdc/nifi-cdc-mysql-bundle/nifi-cdc-mysql-processors/src/main/java/org/apache/nifi/cdc/mysql/processors/CaptureChangeMySQL.java
```diff
@@ -974,139 +938,92 @@ public void outputEvents(ProcessSession session, ComponentLog log) throws IOExce
...
-                                        : new DDLEventInfo(ddlTableInfo, timestamp, currentBinlogFile, currentBinlogPosition, sql);
-                                currentEventInfo = ddlEvent;
-                                currentEventWriter = ddlEventWriter;
-                                currentSequenceId.set(ddlEventWriter.writeEvent(currentSession, transitUri, ddlEvent, currentSequenceId.get(), REL_SUCCESS, eventWriterConfiguration));
-                            }
-                            // Remove all the keys from the cache that this processor added
-                            if (cacheClient != null) {
-                                cacheClient.removeByPattern(this.getIdentifier() + ".*");
+                        if (isQueryDDL(normalizedQuery)) {
+                            if (!(databaseNamePattern != null && !databaseNamePattern.matcher(binlogResourceInfo.getCurrentDatabase()).matches())) {
```
nandorsoma, line 1006: This can be simplified.
```suggestion
                            if (databaseNamePattern == null || databaseNamePattern.matcher(binlogResourceInfo.getCurrentDatabase()).matches())) {
```

nifi-nar-bundles/nifi-cdc/nifi-cdc-mysql-bundle/nifi-cdc-mysql-processors/src/main/java/org/apache/nifi/cdc/mysql/processors/CaptureChangeMySQL.java
```diff
@@ -839,19 +802,18 @@ protected void connect(List<InetSocketAddress> hosts, String username, String pa
...
-                // Ensure driverLocation and driverName are correct before establishing binlog connection
-                // to avoid failing after binlog messages are received.
-                // Actual JDBC connection is created after binlog client gets started, because we need
-                // the connect-able host same as the binlog client.
-                registerDriver(driverLocation, driverName);
-            } catch (InitializationException e) {
-                throw new RuntimeException("Failed to register JDBC driver. Ensure MySQL Driver Location(s)" +
-                        " and MySQL Driver Class Name are configured correctly. " + e, e);
-            }
+        try {
```
nandorsoma, line 805: Can you explain why `createEnrichmentConnection` check is no longer needed? I don't follow.

nifi-nar-bundles/nifi-cdc/nifi-cdc-mysql-bundle/nifi-cdc-mysql-processors/src/main/java/org/apache/nifi/cdc/mysql/processors/CaptureChangeMySQL.java
```diff
@@ -464,58 +460,44 @@ public class CaptureChangeMySQL extends AbstractSessionFactoryProcessor {
...
+    private final BinlogEventState binlogEventState = new BinlogEventState();
 
     private final BeginTransactionEventWriter beginEventWriter = new BeginTransactionEventWriter();
+    private final BeginEventHandler beginEventHandler = new BeginEventHandler();
     private final CommitTransactionEventWriter commitEventWriter = new CommitTransactionEventWriter();
+    private final CommitEventHandler commitEventHandler = new CommitEventHandler();
     private final DDLEventWriter ddlEventWriter = new DDLEventWriter();
+    private final DDLEventHandler ddlEventHandler = new DDLEventHandler();
     private final InsertRowsWriter insertRowsWriter = new InsertRowsWriter();
+    private final InsertEventHandler insertEventHandler = new InsertEventHandler();
```
nandorsoma, line 493: I'm wondering, is there an option to extract these handlers to a different class, that is able to retreive the required handler? This could greatly simplify the code, but I'm not sure that we can shrink the required parameters to a simple enough interface.

nifi-nar-bundles/nifi-cdc/nifi-cdc-mysql-bundle/nifi-cdc-mysql-processors/src/main/java/org/apache/nifi/cdc/mysql/processors/CaptureChangeMySQL.java
```diff
@@ -1391,8 +1296,12 @@ private class JDBCConnectionHolder {
         private JDBCConnectionHolder(InetSocketAddress host, String username, String password, Map<String, String> customProperties, long connectionTimeoutMillis) {
             this.connectionUrl = "jdbc:mysql://" + host.getHostString() + ":" + host.getPort();
             connectionProps.putAll(customProperties);
-            connectionProps.put("user", username);
-            connectionProps.put("password", password);
+            if (username != null) {
+                connectionProps.put("user", username);
+            }
+            if (password != null) {
```
nandorsoma, line 1302: Does it make sense to set password when username is null?

nifi-nar-bundles/nifi-cdc/nifi-cdc-mysql-bundle/nifi-cdc-mysql-processors/src/main/java/org/apache/nifi/cdc/mysql/event/handler/InsertEventHandler.java
```diff
@@ -0,0 +1,45 @@
...
+public class InsertEventHandler implements BinlogEventHandler<WriteRowsEventData, InsertRowsEventInfo> {
+    @Override
+    public void handleEvent(final WriteRowsEventData eventData, final boolean writeEvent, DataCaptureState dataCaptureState,
+                            CaptureChangeMySQL.BinlogResourceInfo binlogResourceInfo, CaptureChangeMySQL.BinlogEventState binlogEventState,
+                            final String sql, AbstractBinlogEventWriter<InsertRowsEventInfo> eventWriter,
+                            EventWriterConfiguration eventWriterConfiguration, ProcessSession session, final long timestamp) {
+        InsertRowsEventInfo eventInfo = dataCaptureState.isUseGtid()
+                ? new InsertRowsEventInfo(binlogResourceInfo.getCurrentTable(), timestamp, dataCaptureState.getGtidSet(), eventData)
+                : new InsertRowsEventInfo(binlogResourceInfo.getCurrentTable(), timestamp, dataCaptureState.getBinlogFile(), dataCaptureState.getBinlogPosition(), eventData);
+        if (writeEvent) {
```
nandorsoma, line 38: There are multiple cases when the handler does nothing when it is not a write event. If the handler would be able to decide alone whether it is a write event or not, it would be ok. But in this case, I think we should even call the event handler. What do you think?

nandorsoma (changes requested):
Thank you for your contribution @mattyb149! I've found few things, could you check them?
Also during testing, I've noticed that with this version I get AccessDeniedException when I try to use a user that has hostname in its name. Like `'nifi'@'172.17.0.1'`. The same flow with the same db configuration works with the current upstream version. I used docker for testing. Could you verify that?

```
docker run --name mysql-test-container -p 3306:3306 -e MYSQL_ROOT_PASSWORD=my-secret-pw -d mysql:8.0.33

docker exec -it mysql-test-container sh

mysql -u root -p
-- type in pw: my-secret-pw

CREATE DATABASE nifitest;
USE nifitest;

CREATE USER 'tmp'@'localhost' IDENTIFIED BY 'password';
-- It is intentionally localhost, this way you will get an AccessDeniedException with the resolved ip. From that exception you will know which ip you should use instead of localhost.
-- Another option is to get the container ip, but maybe it is important how the ip is resolved.

-- Create test data
CREATE TABLE PERSON (
    ID int NOT NULL AUTO_INCREMENT,
    FIRST_NAME varchar(255) NOT NULL,
    LAST_NAME varchar(255) NOT NULL,
    AGE int,
    PRIMARY KEY (ID)
);

INSERT INTO PERSON (FIRST_NAME, LAST_NAME) VALUES ('Nandor Soma', 'Abonyi');
-- Test processor. Hopefully you will also receive AccessDeniedException containing the resolved ip address.

-- create a new user with the correct ip:
CREATE USER 'nifi'@'172.17.0.1' IDENTIFIED BY 'password';

GRANT ALL PRIVILEGES ON *.* TO 'nifi'@'172.17.0.1' WITH GRANT OPTION;
FLUSH PRIVILEGES;

-- Test that you still receive AccessDeniedException with the new nifi user, but you don't receive anything with the upstream version.
```

pvillard31:
Hey @mattyb149 - while you're doing this, can we upgrade mysql-binlog-connector to 0.28.0. The upgrade to this version is breaking a few tests in the Groovy test class though:

````
[ERROR] Failures: 
[ERROR]   CaptureChangeMySQLTest.testInitialGtidIgnoredWhenStatePresent:1352 java.lang.NumberFormatException: Error at index 0 in: "xxxxxxxx"
[ERROR]   CaptureChangeMySQLTest.testInitialGtidNoStatePresent:1395 java.lang.NumberFormatException: Error at index 0 in: "xxxxxxxx"
[ERROR]   CaptureChangeMySQLTest.testUpdateStateUseGtid:1207 expected: <> but was: <null>
[INFO] 
[ERROR] Tests run: 34, Failures: 3, Errors: 0, Skipped: 0
````

mattyb149:
@pvillard I had tried to upgrade but noticed the breaking tests so went back to the last version that didn't involve a regression or change in behavior. Is this the time to upgrade and change the logic? I suppose as it doesn't manifest itself to the user we just have to change our logic to support any changes in behavior, so I will take another look. If it will affect the user (such as giving different output from before) then we might need to keep this version or at least add it to the Migration Guide.

pvillard31:
Yep, definitely agree with you. I guess it's worth looking into it as they mention performance improvements.

nifi-nar-bundles/nifi-cdc/nifi-cdc-mysql-bundle/nifi-cdc-mysql-processors/src/main/java/org/apache/nifi/cdc/mysql/processors/CaptureChangeMySQL.java
mattyb149: We stopped using the DistributedMapCache and are always using an internal Map now, so we always create an enrichment connection, so we don't need to see if we need to.

nifi-nar-bundles/nifi-cdc/nifi-cdc-mysql-bundle/nifi-cdc-mysql-processors/src/main/java/org/apache/nifi/cdc/mysql/processors/CaptureChangeMySQL.java
mattyb149: That was what I was hoping to do, but I think the required parameters made the interface method pretty awkward. I can take another look

nandorsoma (approved):
Thanks for the followup @mattyb149! The current version looks good, merging to `main` and `nifi-1.x`.

PR #7117, opened 2023-04-04
NIFI-11287: detect dependent properties when the property it depends on references a parameter (UI)

mtien-apache:
!-- Licensed to the Apache Software Foundation (ASF) under one or more -->

# Summary

[NIFI-11287](https://issues.apache.org/jira/browse/NIFI-11287) (UI)
To test:
- add the Compress Content processor to the canvas
- for property "Mode", select between "compress" and "decompress". You should see the "Compress Level" dependent property show and hide
- create and add a parameter context with a parameter that has a value of "decompress" 
- back in the processor configurations, configure the "Mode" property with the new parameter. The "Compress Level" property should hide. 

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
@@ -33,7 +33,6 @@
                 'nf.ProcessGroupConfiguration',
                 'nf.Settings',
                 'nf.ParameterContexts',
-                'nf.CanvasUtils',
                 'lodash'],
             function ($,
                       Slick,
@@ -47,7 +46,6 @@
                       nfProcessGroupConfiguration,
                       nfSettings,
                       nfParameterContexts,
-                      nfCanvasUtils,
                       _) {
                 factory($,
                     Slick,
@@ -61,7 +59,6 @@
                     nfProcessGroupConfiguration,
                     nfSettings,
                     nfParameterContexts,
-                    nfCanvasUtils,
                     _);
             });
     } else if (typeof exports === 'object' && typeof module === 'object') {
@@ -77,7 +74,6 @@
             require('nf.ProcessGroupConfiguration'),
             require('nf.Settings'),
             require('nf.ParameterContexts'),
-            require('nf.CanvasUtils'),
             require('lodash'));
     } else {
         factory(root.$,
@@ -92,7 +88,6 @@
             root.nf.ProcessGroupConfiguration,
             root.nf.Settings,
             root.nf.ParameterContexts,
-            root.nf.CanvasUtils,
             root._);
     }
 }(this, function ($,
@@ -107,13 +102,12 @@
                   nfProcessGroupConfiguration,
                   nfSettings,
                   nfParameterContexts,
-                  nfCanvasUtils,
                   _) {
 
     var groupId = null;
     var supportsSensitiveDynamicProperties = false;
     var propertyVerificationCallback = null;
-    var currentParameterContext = null;
+    var currentParameters = null;
     var COMBO_MIN_WIDTH = 212;
     var EDITOR_MIN_WIDTH = 212;
     var EDITOR_MIN_HEIGHT = 100;
@@ -1473,7 +1467,8 @@
 
             // sets the available parameters for the specified property descriptor
             var loadParameters = function (propertyDescriptor, parameterDeferred, setParameters) {
-                parameterDeferred(propertyDescriptor, groupId).done(function (parameters) {
+                parameterDeferred(nfCommon.isSensitiveProperty(propertyDescriptor), groupId).done(function (parameters) {
+                    currentParameters = parameters;
                     setParameters(parameters);
                 });
             };
@@ -1731,10 +1726,10 @@
                                                 referencingParameter = null;
 
                                                 // check if the property references a parameter
-                                                if (!_.isEmpty(currentParameterContext)) {
+                                                if (!_.isEmpty(currentParameters)) {
                                                     const paramReference = getExistingParametersReferenced(propertyValue);
                                                     if (paramReference.length > 0) {
-                                                        referencingParameter = paramReference[0].parameter.value;
+                                                        referencingParameter = paramReference[0].value;
                                                     }
                                                 }
 
@@ -1883,14 +1878,26 @@
      * @param {type} properties
      * @param {type} descriptors
      * @param {type} history
+     * @param {type} options
      */
-    var loadProperties = function (table, properties, descriptors, history) {
+    var loadProperties = function (table, properties, descriptors, history, options) {
         // save the original descriptors and history
         table.data({
             'descriptors': descriptors,
             'history': history
         });
 
+        if (_.isFunction(options.parameterDeferred)) {
+            options.parameterDeferred(false, groupId).done(function (parameters) {
+                currentParameters = parameters;
+                processProperties(table, properties, descriptors);
+            });
+        } else {
+            processProperties(table, properties, descriptors);
+        }
+    };
+
+    var processProperties = function (table, properties, descriptors) {
         // get the grid
         var propertyGrid = table.data('gridInstance');
         var propertyData = propertyGrid.getData();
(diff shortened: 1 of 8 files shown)


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
```diff
@@ -2001,6 +2021,43 @@
...
+     * @param {boolean} sensitive
+     * @returns {ParameterEntity[]}
+     */
+    var getExistingParametersReferenced = function (parameterReference, parameterContext, sensitive) {
+        var parameters = _.get(parameterContext.component, 'parameters', []);
+        var existingParametersReferenced = [];
+
+        if (!_.isNil(parameterReference) && !_.isEmpty(parameters)) {
+            // can't use from common/constants because we are modifying the lastIndex below
+            var paramRefsRegex = /#{(')?([a-zA-Z0-9-_. ]+)\1}/gm;
```
exceptionfactory, line 2038: Is it necessary to define a capturing group for the `'` character? It seems like the pattern could be adjusted to use a single capturing group:
```suggestion
            var paramRefsRegex = /#{'?([a-zA-Z0-9-_. ]+)'?}/gm;
```

mtien-apache: @exceptionfactory It's not necessary to define a separate capturing group for `'`. I updated this block of code.

mtien-apache:
Thanks for reviewing, @exceptionfactory! I fixed the syntax errors and addressed your comments. Would you look again?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
```diff
@@ -2001,6 +2020,45 @@
...
 
+    /**
+     * Gets all the referenced parameters from the {parameterContext} based on the value of {parameterReference} with matching {sensitive} property
+     *
+     * @param {string} parameterReference
+     * @param {ParameterContextEntity} parameterContext
+     * @param {boolean} sensitive
+     * @returns {ParameterEntity[]}
+     */
+    var getExistingParametersReferenced = function (parameterReference, parameterContext, sensitive) {
```
exceptionfactory, line 2031: Is there a reason for passing `parameterContext` and `sensitive` as arguments? It looks like `sensitive` is always `false` and `parameterContext` always references `currentParameterContext`. Unless there is some other reason to keep them, it seems like the implementation could be simplified.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
```diff
@@ -2261,9 +2319,32 @@
...
+         * @argument {object} options
          */
-        loadProperties: function (properties, descriptors, history) {
-            return this.each(function () {
+        loadProperties: function (properties, descriptors, history, options) {
+            var self = this;
+            var groupId = null;
+
+            var loadParameterContext = function (options) {
+                if (typeof options.getFullParameterContextDeferred === 'function') {
```
exceptionfactory, line 2329: Is there a particular reason for passing `getFullParameterContextDeferred` and `getParameterContext` as properties of the `options` argument? Could they be passed as specific arguments instead of properties of `options`? If there is some other reason, recommend at least providing additional comments for the `options` argument.

exceptionfactory (commented):
Thanks for correcting the syntax issues @mtien-apache. I noted a couple additional questions on the implementation itself.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
mtien-apache: @exceptionfactory I've simplified the function by removing the `sensitive` and `parameterContext` parameters.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
mtien-apache: @exceptionfactory I added additional comments for the `options` argument. The `options` would only be necessary when `loadProperties` is called from processor configurations because it requires access to a parameter context in order to reference parameters. `loadProperties` can be called from other parts of NiFi, such as the Controller Service or Reporting Task that don't require additional options. In order to know this was called from processor configs, we can check for the parameter context options.

mtien-apache:
I changed the way `jqeury.propertytable` retrieves the parameter context when calling `loadProperties`. Instead of a deferred, `loadProperties` will now check for a new parameter, `type`, and if it's a `processor-configuration` or `controller-service`, then it will make an ajax request to get the parameter context. 

@exceptionfactory can you check again? Thanks!

mcgilman (changes requested):
Thanks for the PR @mtien-apache and thanks for the review @exceptionfactory. I think we can simplify the code within the Property Table because we should already be able to fetch the Parameter Context.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
```diff
@@ -2001,6 +2026,44 @@
         propertyData.setItems([]);
     };
 
+    /**
+     * Gets all the referenced parameters from the {parameterContext} based on the value of {parameterReference}
+     *
+     * @param {string} parameterReference
+     * @returns {ParameterEntity[]}
+     */
+    var getExistingParametersReferenced = function (parameterReference) {
```
mcgilman, line 2035: It also looks like the Property Table already has a `containsParameterReference` function. Can it be used instead? I see that one is simply detecting while the other is extracting the Parameter Reference. If there's no opportunity for reuse that's ok. Just wanted to raise awareness here in case.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
```diff
@@ -1931,13 +1948,20 @@
                                 if (property.hidden === false) {
                                     // Get the property value by propertyName
                                     var propertyValue = properties[dependency.propertyName];
+                                    referencingParameter = null;
+
+                                    // check if the property references a parameter
+                                    if (!_.isEmpty(currentParameterContext)) {
```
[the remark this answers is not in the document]
mtien-apache: @mcgilman Thanks for reviewing! I removed the `type` parameters and updated to use the provided callbacks from the client when calling `loadProperties`.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/propertytable/jquery.propertytable.js
mtien-apache: @mcgilman Good point. I changed the pattern for these cases.

exceptionfactory: Thanks for noting that background @mcgilman, that makes sense, and thanks for making the changes @mtien-apache.

exceptionfactory (commented):
Thanks for making the additional changes @mtien-apache, I will defer to @mcgilman for final review.

mcgilman (approved):
Thanks @mtien-apache!

PR #7142, opened 2023-04-08
NIFI-11411 - Upgrade to Bouncycastle 1.73

pvillard31:
# Summary

[NIFI-11411 ](https://issues.apache.org/jira/browse/NIFI-11411)
https://www.bouncycastle.org/releasenotes.html

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,7 +116,7 @@
         <org.apache.commons.text.version>1.10.0</org.apache.commons.text.version>
         <org.apache.httpcomponents.httpclient.version>4.5.14</org.apache.httpcomponents.httpclient.version>
         <org.apache.httpcomponents.httpcore.version>4.4.16</org.apache.httpcomponents.httpcore.version>
-        <org.bouncycastle.version>1.71</org.bouncycastle.version>
+        <org.bouncycastle.version>1.73</org.bouncycastle.version>
         <testcontainers.version>1.18.0</testcontainers.version>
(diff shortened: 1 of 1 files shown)


pvillard31:
Sounds like the 1.73 bits are not yet available everywhere on Maven.
Release went out today so we may need to wait a bit.
I can also downgrade to 1.72 if preferred.

exceptionfactory (changes requested):
Thanks for the upgrade @pvillard31, unfortunately it looks like the artifacts are not yet available in Maven Central.

More concerning, however, it looks like the new Bouncy Castle Provider is still 11 MB, which is almost twice the size of version 1.71.

We decided to skip 1.72 because it included experimental and broken algorithms that greatly expanded the size of the binary. Unfortunately, it would appear that Bouncy Castle has not removed those files in version 1.73. The number of references to Bouncy Castle in NiFi is significant, so it looks like we need to evaluate options. One solution may be to produce a shaded version for NiFi that trims the JAR to a reasonable size. Another option that we probably need to evaluate is placing Bouncy Castle in the root library directory, thus avoiding the proliferation of copies.

With that background, it looks like this upgrade needs further evaluation.

pvillard31:
Oh yeah, I completely forgot our discussion about the size. Good point!

PR #7184, opened 2023-04-20
NIFI-11472 Make PutFTP processor more multithread friendly

MormonJesus69420:
Add an extra check during directory creation to see if directory wasn't already created in another thread.

From Issue:

Problem happens when a PutFTP is set to run several concurrent tasks and two (or more ) FlowFiles come in and both need to create the same directory. One of them will create directory and succeed immediately while the other will try to create directory, but fail since it already exist, throw an error, the FlowFile will then be penalized and on second run will succeed.

While it is not the biggest error, as files are getting transferred in the end, but the bulletins and errors are annoying, especially in production environment where you don't want to get unnecessary errors.

We found that the solution involves a simple change to the FTPTransfer.java class in:
nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/FTPTransfer.java
On line 398 and ensureDirectoryExists method you can simply add another if check which double checks that the directory exists when it fails to create one.
```java
final boolean cdSuccessful = setWorkingDirectory(remoteDirectory);

if (!cdSuccessful) {    
  if (client.makeDirectory(remoteDirectory)) {        
    logger.debug("Remote Directory not found: created directory [{}]", remoteDirectory);    
  } else if (!setWorkingDirectory(remoteDirectory)) { 
         // Double check that the dir exists as it might have been created in another thread        
    throw new IOException("Failed to create remote directory " + remoteDirectory);    
  }
}
```

# Summary

[NIFI-11472](https://issues.apache.org/jira/browse/NIFI-11472)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/FTPTransfer.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/FTPTransfer.java
@@ -395,7 +395,7 @@ public class FTPTransfer implements FileTransfer {
         if (!cdSuccessful) {
             if (client.makeDirectory(remoteDirectory)) {
                 logger.debug("Remote Directory not found: created directory [{}]", remoteDirectory);
-            } else {
+            } else if (!setWorkingDirectory(remoteDirectory)) { // Double check that the dir exists as it might have been created in another thread
                 throw new IOException("Failed to create remote directory " + remoteDirectory);
             }
         }


arpadboda:
Hello @MormonJesus69420 , thanks for your contribution!
I'm not sure if using multiple threads to transfer to the same ftp makes any sense, this operation is limited by bandwith anyway. 
So I would prefer to restrict it to single threaded usage, but the current implementation allows uploading to multiple hosts based on flowfile attributes, in which case multiple threads might make sense, so I'm not against this change.
Weak +1, I'm ok to merge with a 2nd approval.

MormonJesus69420:
I am sorry, I managed to make the most basic mistake in such a small change, I forgot to add a `!` to the `setWorkingDirectory(remoteDirectory)` method call.

MormonJesus69420:
Strange, I don't understand why it failed on the Windows action. I don't have the ability to test it on Windows either, since we use Linux for development at work. Also my branch is based off of the latest nifi/main branch.

exceptionfactory (commented):
Thanks for the contribution @MormonJesus69420, and thanks for the review @arpadboda!

The Windows build often takes longer than the others, so it timed out, I restarted it to try again.

The use case of multiple concurrent tasks for PutFTP makes sense, the introducing the additional call to `setWorkingDirectory` seems reasonable under the circumstances. Within the context of a single NiFi server, a more robust solution might be to add `synchronized` to this method, but that would not help when running PutFTP across multiple NiFi nodes in a cluster.

With that background, attempting to change to the directory if the `makeDirectory` command fails seems like the best approach. I will monitor the Windows build status for verification.

exceptionfactory (approved):
Thanks again @MormonJesus69420! +1 merging

MormonJesus69420:
Hi, I see that the PR is already closed, but thought I might add the example where we tested running PutFTP processor using 1 and 2 concurrent tasks. As long as we were able to provide the data to send, you can see that upload speed was almost doubled. As such the bottleneck was how fast we were able to provide data to send, rather than sending of data. 
Comparison in MB/s transfer speeds between 1 and 2 concurrent tasks in PutFTP

PR #7185, opened 2023-04-20
NIFI-11473 Flow version change in NiFi should not stop a component wh…

timeabarna:
…en only position is changed

# Summary
When going from one flow version to another and the position of a component is changing, but not its configuration, the component should not be stopped.

[NIFI-11473](https://issues.apache.org/jira/browse/NIFI-11473)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -217,6 +217,10 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
                 }
             }
 
+            if (diff.getDifferenceType() == DifferenceType.POSITION_CHANGED) {
+                continue;
+            }
+
             final VersionedComponent component = diff.getComponentA() == null ? diff.getComponentB() : diff.getComponentA();
             updatedVersionedComponentIds.add(component.getIdentifier());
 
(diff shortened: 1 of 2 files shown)


mattyb149:
Reviewing...

mattyb149:
Is there a reproduction path for this? Using the registry I started the processor then saved the version. Then I moved the processor and committed the changes. When I change version back, it appears to set the desired version but the processor is not stopped (although a dialog pops up saying . This is without the fix in this PR, so I'm thinking there's a different way to reproduce this behavior?

timeabarna:
Hello @mattyb149,

I've created a GenerateFlowFile processor which is running.  Created a version where only position changed.
Added debug point at FlowUpdateResource updateFlow method and a debug point to StandardVersionedFlowSynchronizer updateProcessor method.

At version change, without patch:
FlowUpdateResource updateFlow runningComponents set contains GenerateFlowFile processor.
StandardVersionedFlowSynchronizer updateProcessor goes to the updateProcessor if branch
nifi-app.log shows FlowUpdateResource Stopping 1 Processors and FlowUpdateResource Restarting 1 Processors

With patch:
FlowUpdateResource updateFlow runningComponents set empty.
StandardVersionedFlowSynchronizer updateProcessor goes to the setPosition if branch
nifi-app.log shows FlowUpdateResource Stopping 0 Processors and FlowUpdateResource Restarting 0 Processors
If my memory serves well only Updating Flow can be seen on UI.

pvillard31:
What I usually do to check this:
- have a GenerateFlowFile to InvokeHTTP
- have a ListenHTTP
Version this is as v1.
Move the ListenHTTP processor. Version this is as v2.
With processors running, without the fix, v1 to v2 would generate errors because ListenHTTP is restarted and InvokeHTTP cannot send data during the restart. With the fix, there should be no error.

mattyb149:
+1 LGTM, reproduced without the PR and verified the behavior with the PR is as expected. Thanks for the fix! Merging to main and 1.x

PR #7194, opened 2023-04-24
NIFI-11167 - Add Excel Record Reader

dan-s1:
# Summary

[NIFI-11167](https://issues.apache.org/jira/browse/NIFI-11167)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


MikeThomsen:
@dan-s1 you have some unit test failures in your commit.

MikeThomsen:
@exceptionfactory @dan-s1 This is a really interesting PR. I've had a number of clients that would benefit from this sort of capability. My one big question/concern is whether or not this new reader will be tolerant of poorly formatted or inconsistent documents. As an example, one of the things I've seen over the years is people doing stuff like inserting arbitrary values at various points of the spreadsheet. Comments, stuff like that. Will this reader be flexible enough to ignore junk and badly formatted stuff or will it just fail on contact with such data?

dan-s1:
@MikeThomsen @exceptionfactory The test failures I see are

> Error:  Failures: 
> Error:    TestExcelSchemaInference.testInferenceIncludesAllRecords:62 expected: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, TIME:HH:mm:ss a, STRING]> but was: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, STRING]>
> Error:    TestExcelSchemaInference.testInferenceIncludesAllRecordsWithEL:94 expected: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, TIME:HH:mm:ss a, STRING]> but was: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, STRING]>

The problem is on my box those tests are passing. I am not sure how to go about fixing these short of removing those assertions. Please advise.

exceptionfactory:
> @MikeThomsen @exceptionfactory The test failures I see are
> 
> > Error:  Failures:
> > Error:    TestExcelSchemaInference.testInferenceIncludesAllRecords:62 expected: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, TIME:HH:mm:ss a, STRING]> but was: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, STRING]>
> > Error:    TestExcelSchemaInference.testInferenceIncludesAllRecordsWithEL:94 expected: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, TIME:HH:mm:ss a, STRING]> but was: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, STRING]>
> 
> The problem is on my box those tests are passing. I am not sure how to go about fixing these short of removing those assertions. Please advise.

The problem is due to the fact that the timestamp is implicitly local, so the expected value needs to account for that fact. I will take a closer look at the test to determine the best solution.

exceptionfactory:
> @MikeThomsen @exceptionfactory The test failures I see are
> 
> > Error:  Failures:
> > Error:    TestExcelSchemaInference.testInferenceIncludesAllRecords:62 expected: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, TIME:HH:mm:ss a, STRING]> but was: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, STRING]>
> > Error:    TestExcelSchemaInference.testInferenceIncludesAllRecordsWithEL:94 expected: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, TIME:HH:mm:ss a, STRING]> but was: <CHOICE[TIMESTAMP:yyyy/MM/dd/ HH:mm, STRING]>
> 
> The problem is on my box those tests are passing. I am not sure how to go about fixing these short of removing those assertions. Please advise.

After taking a closer look, the problem does not appear to related to the time zone necessarily. The test is failing across all platforms in different time zones.

@dan-s1 Does the test succeed when running a Maven build from the command line? Can you provide the version of Java and Maven you are using?

exceptionfactory:
@MikeThomsen That's a good question about strict versus lenient parsing. I have not reviewed the pull request in detail yet either, but I plan to review it soon. One primary goal is to replace the existing ConvertExcelToCSVProcessor.

dan-s1:
@exceptionfactory When running mvn -v I have

```
Apache Maven 3.9.0 (9b58d2bad23a66be161c4664ef21ce219c2c8584)
Maven home: /opt/apache-maven-3.9.0
Java version: 11.0.18, vendor: Red Hat, Inc., runtime: /usr/lib/jvm/java-11-openjdk-11.0.18.0.10-1.el7_9.x86_64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.4.232-1.el7.elrepo.x86_64", arch: "amd64", family: "unix"
```

dan-s1:
> @exceptionfactory @dan-s1 This is a really interesting PR. I've had a number of clients that would benefit from this sort of capability. My one big question/concern is whether or not this new reader will be tolerant of poorly formatted or inconsistent documents. As an example, one of the things I've seen over the years is people doing stuff like inserting arbitrary values at various points of the spreadsheet. Comments, stuff like that. Will this reader be flexible enough to ignore junk and badly formatted stuff or will it just fail on contact with such data? IMHO, to be genuinely useful it would need to have a flexible mode and a strict mode so that people could make at least a "best effort" at extracting data.
> 
> (I haven't actually dug into the PR, so just asking some high level questions. Going to try to start reviewing today or tomorrow)

@MikeThomsen  I believe the Inference schema strategy will work for you customers as it is will infer all the different data types a column can have.  Using that will allow for the tolerance while a more specific schema will enforce the strict mode.

dan-s1:
@exceptionfactory Are the locales where these CI builds English speaking? I suspect if not, then PM is not displayed in the Excel spreadsheet as it would be in an English speaking locale. Hence there would be no column to match the 'TIME:HH:mm:ss a' format as there is on my box which is an English speaking locale. I did some quick research and it appears by default Excel displays dates based on regional settings.

exceptionfactory:
> @exceptionfactory Are the locales where these CI builds English speaking?

@dan-s1, no the automated builds use different locales (`EN`, `HI`, `JP`, and `FR`) to surface potential issues.

It sounds like changing the tests to avoid using language-based indicators (such as AM and PM) would be the best way forward.

dan-s1:
@exceptionfactory So my changes did improve things though now on the Windows box it is failing due to it interpreting a number as a string. What would you like me to do for this?

exceptionfactory:
Thanks for improving the tests @dan-s1. The Windows build failure needs to be corrected, can you narrow down why it is failing?

dan-s1:
@exceptionfactory The only thing I can think of right now is perhaps there are differences between how the Excel document is interpreted on a native Windows environment and on Linux environments.  The file being tested against was created with LibreOffice (I believe it was one of the test files from ConvertExcelToCsv and I added a column and saved with LibreOffice). I will look into this further.

dan-s1:
@exceptionfactory I was able to find the discrepancy and it is not what I thought. It is actually dependent on the java.util.Locale and not the operating system. In order to use the org.apache.nifi.util.SchemaInferenceUtil.getDataType method for inferring the correct data type, I needed to get the cell value as a string. To do that I used [ apache.poi.ss.usermodel.DataFormatter](apache.poi.ss.usermodel.DataFormatter) which I hoped would take care of any formatting and give the best inference possible especially if the cell value was a timestamp or perhaps a big integer. It is only now I realize that class is heavily reliant on java.util.Locale which determine how numbers and dates are formatted in a given locale. Hence the difference across the CI builds which are running in different locales.  The particular  field which has the differences is a field with a number in scientific notation. In FR it is represented for example as 9,8765E+08 while the other locales represent it as 9.8765E+08. The underlying NIFI code which parses numbers does not recognize commas hence it is inferring the field's data type as a string and not a float. I still believe the choice I made for the inference logic is correct as the POI handles all of the necessary formatting. Though it seems  I probably should use simpler data which would work across all locales. Please let me know how you would like me to proceed.

dan-s1:
@exceptionfactory I simplified the unit tests to avoid any  locale issues (much like org.apache.nifi.csv.TestCSVSchemaInference). I also added unit tests to highlight the difference between locales when the schema is inferred.

exceptionfactory:
Thanks @dan-s1! For subsequent changes on this pull request, please add new commits so it we can track the changes necessary. I started the automated build to evaluate the changes.

dan-s1:
@exceptionfactory I realize you are very busy as I see a lot of activity with your name but I was wondering if this PR can please be considered?  If there are any issues I would like to answer them and/or fix them while the code is somewhat fresher in my mind. Thank you!

exceptionfactory:
Thanks @dan-s1, this is on my list to review, I plan to take a closer look soon.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
```diff
@@ -0,0 +1,193 @@
...
+            .displayName("Sheets to Extract")
+            .description("Comma separated list of Excel document sheet names whose rows should be extracted from the excel document. If this property" +
+                    " is left blank then all the rows from all the sheets will be extracted from the Excel document. The list of names is case in-sensitive. Any sheets not" +
+                    " specified in this value will be ignored. A bulletin will be generated if a specified sheet(s) are not found.")
+            .required(false)
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor FIRST_ROW_NUM = new PropertyDescriptor
```
exceptionfactory, line 78:
```suggestion
    public static final PropertyDescriptor STARTING_ROW = new PropertyDescriptor
```

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
```diff
@@ -0,0 +1,193 @@
...
+
+    private int getFirstRow(final PropertyContext context) {
+        int rawFirstRow = context.getProperty(FIRST_ROW_NUM).asInteger();
+        return getZeroBasedIndex(rawFirstRow);
+    }
+
+    static int getZeroBasedIndex(int rawFirstRow) {
+        return rawFirstRow > 0 ? rawFirstRow - 1 : 0;
+    }
+    private String[] getRawDesiredSheets(final PropertyContext context) {
```
exceptionfactory, line 115: Methods should have a newline separating them from previous methods.

```suggestion

    private String[] getRawDesiredSheets(final PropertyContext context) {
```

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
```diff
@@ -0,0 +1,193 @@
...
+        properties.add(DateTimeUtils.DATE_FORMAT);
+        properties.add(DateTimeUtils.TIME_FORMAT);
+        properties.add(DateTimeUtils.TIMESTAMP_FORMAT);
+
+        return properties;
+    }
+
+    @Override
+    protected SchemaAccessStrategy getSchemaAccessStrategy(final String allowableValue, final SchemaRegistry schemaRegistry, final PropertyContext context) {
+        if (allowableValue.equalsIgnoreCase(HEADER_DERIVED.getValue())) {
```
exceptionfactory, line 149: It is better to use `equals()` for property value comparisons, since the case should always match.
```suggestion
        if (HEADER_DERIVED.getValue().equals(allowableValue)) {
```

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
```diff
@@ -0,0 +1,221 @@
...
+        } else {
+            this.timestampFormat = args.getTimestampFormat();
+            LAZY_TIMESTAMP_FORMAT = () -> DataTypeUtils.getDateFormat(timestampFormat);
+        }
+
+        try {
+            this.rowIterator = new RowIterator(args.getInputStream(), desiredSheets, args.getFirstRow(), logger);
+        } catch (RuntimeException e) {
+            String msg = "Error occurred while processing record file";
+            logger.error(msg, e);
```
exceptionfactory, line 92: It is not necessary to log and error and throw the exception in this case. Recommend removing the logger.error call and allowing the framework to log the exception.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
```diff
@@ -0,0 +1,221 @@
...
+                                    : convertSimpleIfPossible(cellValue, dataType, fieldName);
+                            currentRowValues.put(fieldName, value);
+                        }
+                    });
+        }
+
+        return currentRowValues;
+    }
+
+    public static Object getCellValue(Cell cell) {
```
exceptionfactory, line 140: Is there a reason this method is public? If it is public for testing, it probably should be moved to a utility class.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
```diff
@@ -0,0 +1,221 @@
...
+    public RecordSchema getSchema() throws MalformedRecordException {
+        return schema;
+    }
+
+    @Override
+    public void close() throws IOException {
+        this.rowIterator.close();
+    }
+
+    void setLogger(ComponentLog logger) {
```
exceptionfactory, line 217: Is there a reason for this setter method as opposed to passing the logger in the constructor?

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReaderArgs.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+package org.apache.nifi.excel;
+
+import org.apache.nifi.logging.ComponentLog;
+import org.apache.nifi.serialization.record.RecordSchema;
+
+import java.io.InputStream;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+public class ExcelRecordReaderArgs {
```
exceptionfactory, line 25: Recommend renaming this to `ExcelRecordReaderConfiguration`, although it does include the `InputStream`. In that case, it may be better to pass the InputStream separately.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordSource.java
```diff
@@ -0,0 +1,54 @@
...
+    public ExcelRecordSource(final InputStream in, final PropertyContext context, final Map<String, String> variables) {
+        try {
+            String desiredSheetsDelimited = context.getProperty(ExcelReader.DESIRED_SHEETS).evaluateAttributeExpressions(variables).getValue();
+            String [] rawDesiredSheets = ExcelReader.getRawDesiredSheets(desiredSheetsDelimited, null);
+            Integer rawFirstRow = context.getProperty(ExcelReader.FIRST_ROW_NUM).evaluateAttributeExpressions(variables).asInteger();
+            int firstRow = rawFirstRow != null ? rawFirstRow : NumberUtils.toInt(ExcelReader.FIRST_ROW_NUM.getDefaultValue());
+            firstRow = ExcelReader.getZeroBasedIndex(firstRow);
+            this.rowIterator = new RowIterator(in, getDesiredSheets(rawDesiredSheets), firstRow);
+        } catch (RuntimeException e) {
+            throw new ProcessException(e);
```
exceptionfactory, line 43: ProcessException should include a message, but as it does not seem to be adding any value, it probably makes sense to remove the try-catch and allow RuntimeExceptions to bubble up.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelUtils.java
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.excel;
+
+import org.apache.poi.ss.usermodel.Row;
+
+public class ExcelUtils {
```
exceptionfactory, line 21: Unless there are other utility methods, it is probably cleaner to localize `hasCells` instead of creating this one-method class.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Override
+    public boolean hasNext() {
+        setCurrent();
+        boolean next = currentRow != null;
+        if(!next) {
+            String sheetsNotFound = getSheetsNotFound(desiredSheets);
+            if (!sheetsNotFound.isEmpty() && log) {
+                logger.warn("Excel sheet(s) not found: {}", sheetsNotFound);
+            }
```
exceptionfactory, line 69: This check and warning seems unnecessary in general, perhaps there are other places this could be logged if needed.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -0,0 +1,155 @@
...
+    public Row next() {
+        return currentRow;
+    }
+
+    @Override
+    public void close() throws IOException {
+        this.workbook.close();
+    }
+
+    void setLogger(ComponentLog logger) {
```
exceptionfactory, line 151: As elsewhere, having a setter for the ComponentLog does not seem like the best approach, if the ComponentLog is needed, it should be passed in the constructor.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
```diff
@@ -0,0 +1,246 @@
...
+public class TestExcelRecordReader {
+
+    private static final String DATA_FORMATTING_FILE = "dataformatting.xlsx";
+    private static final String MULTI_SHEET_FILE = "twoSheets.xlsx";
+
+    @Test
+    public void testNonExcelFile() throws IOException {
+        ExcelRecordReaderArgs args = new ExcelRecordReaderArgs.Builder()
+                .withInputStream(getInputStream("notExcel.txt"))
+                .withLogger(Mockito.mock(ComponentLog.class))
```
exceptionfactory, line 57: Instead of calling `Mockito.mock()` `ComponentLog` can be declared as a member variable with the `@Mock` annotation, and the class can use `@ExtendWith(MockitoExtension.class)` to instantiate the mock.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
```diff
@@ -0,0 +1,246 @@
...
+        return recordReader;
+    }
+
+    private List<Record> getRecords(ExcelRecordReader recordReader, boolean coerceTypes, boolean dropUnknownFields, boolean print) throws IOException, MalformedRecordException {
+        Record record;
+        List<Record> records = new ArrayList<>();
+        while ((record = recordReader.nextRecord(coerceTypes, dropUnknownFields)) != null) {
+            if (print) {
+                System.out.println(record.toMap());
+            }
```
exceptionfactory, line 117: This should be removed, `System.out.println()` should not be included in tests.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
exceptionfactory, line 115: Private methods should be defined after public and protected methods in general class ordering, so it would be helpful to move this down in the class.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
```diff
@@ -0,0 +1,193 @@
...
+                .withDesiredSheets(desiredSheets)
+                .withFirstRow(firstRow)
+                .withInputStream(in)
+                .withLogger(logger)
+                .withSchema(schema)
+                .withTimeFormat(timeFormat)
+                .withTimestampFormat(timestampFormat)
+                .build();
+
+        return new ExcelRecordReader(args);
```
exceptionfactory, line 186: Instead of passing the `InputStream` with the configuration, recommend keeping it as a separate argument to the constructor, the same applies to `ComponentLog`.

exceptionfactory (commented):
Thanks for putting together the new Excel Record Reader @dan-s1! It looks like this covers a lot of capabilities. I plan on taking a closer look at several things, but I highlighted several code questions on an initial review.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
dan-s1: I have this specifically for unit testing so I could instantiate a MockComponentLog using the instantiated `ExcelRecordReader` See line 106 in TestExcelRecordReader

exceptionfactory: Thanks, that is understandable, but optional logging or checking for log statements is generally not the best approach. However, it should be possible to use the NiFi TestRunner to configure and enable the Controller Service, which would provide a ComponentLog that could be checked without wiring the log using this approach.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelUtils.java
dan-s1: This method is used in 3 different classes ExcelHeaderSchemaStrategy (line 79), ExcelRecordReader (line 113) and ExcelSchemaInference (line 64). I did not want to have to duplicate the code.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
dan-s1: @exceptionfactory What do you mean by

> the Controller Service ?
ExcelRecordReader is not a a controller service

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: This the bulletin generated which is remnant from the ConvertExcelToCsv processor. I thought this is the only place to log it as it is the only data structure that iterates the sheets of the Excel document.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
exceptionfactory: I meant to say, the `ExcelReader` Controller Service can be configured and enabled through the TestRunner.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
exceptionfactory: Some of the other Record Readers add a record.count attribute, which is useful for tracking. That's probably a better option than logging a warning.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
dan-s1: Sorry I do not see how. In order to instantiate a TestRunner don't you need a class which implements 
`org.apache.nifi.processor.Processor` ? `ExcelReader` does not implement it.

exceptionfactory: You can use the `NoOpProcessor.class`. See [StandardPrivateKeyServiceTest](https://github.com/apache/nifi/blob/607b53504b57d5d7fde84a721ce75056c92a1f70/nifi-nar-bundles/nifi-standard-services/nifi-key-service-bundle/nifi-key-service/src/test/java/org/apache/nifi/key/service/StandardPrivateKeyServiceTest.java) as an example.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: I thought it may be useful for operators to know explicitly when a sheet did not exist. I personally think that is clearer than a record count which could be misleading.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: Instead of passing a ComponentLog, could I just use a static final instance of `org.sl4j.Logger` to handle all logging?

exceptionfactory: `ComponentLog` is necessary to associate log messages with particular instances of the component, so it is required. It should be possible to pass through the constructor.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelUtils.java
dan-s1: @exceptionfactory What would you like me to do for this? Leave as is, declare the method static in one class and have the others use it or duplicate it for all 3 classes that use it?

exceptionfactory: Thanks, it is fine for now, leaving it as is sounds good.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: I guess my problem here is `RowIterator` is used by both `ExcelRecordReader` and `ExcelRecordSource`. Only `ExcelRecordReader` has an instance of `ComponentLog` while  `ExcelRecordSource` does not. That is why I have conditional logging depending on which class instantiates an instance of  `RowIterator`. I did not see how `ExcelRecordSource` could get an instance of `ComponentLog`. Please advise.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
dan-s1: I will defer to what you want, but I personally order my methods in the order they are called as that follows a logical progression (a tip I picked up years ago)

exceptionfactory: Thanks, although the code base is not absolutely consistent, the general pattern is static, then public, protected, and private methods.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
exceptionfactory: `ExcelRecordReader` can pass its `ComponentLog` when instantiating `ExcelRecordSource`. Are there other places where `ComponentLog` is not available?

exceptionfactory: Sounds good!

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: @exceptionfactory I did not see other record readers with a record.count attribute. Which one(s) did  you see this in? I did a search for `record,.count`

dan-s1: @exceptionfactory Does the exception have to be specifically an `IOException`? The problem I have is if that or any other checked exception is thrown then the `hasNext`  method signature has to change and that clashes with the` java.util.Iterator` interface. Would a `RuntimeException` be okay to throw?

exceptionfactory: It could be something else, probably `ProcessException` would be a good option as a more specific subclass of `RuntimeException`.

dan-s1: Perfect, thanks!

dan-s1:
@exceptionfactory I believe I covered all the points your brought up. Please let me know how this looks and whether there is anything else to change. 
Can I mark the conversations resolved or is that something the reviewer does? I have seen it done both ways on projects and I am not sure how Apache NIFI does it.
Thanks!

dan-s1:
@exceptionfactory Just checking if there is anything else I need to do for this PR. Thanks!

dan-s1:
@exceptionfactory I went through the code again and I corrected things based on feedback you had given on other PRs.

exceptionfactory:
Thanks @dan-s1, I will review the latest version soon.

dan-s1:
@exceptionfactory Seems like the failure in the build was unrelated to the code I checked in.

dan-s1:
@exceptionfactory Can you please restart the Windows Zulu JDK 11 FR ? It had timed out yesterday. Thanks!

dan-s1:
@exceptionfactory 👍Finally that Windows build succeeded.

exceptionfactory (changes requested):
Thanks for the recent updates @dan-s1. In the course of testing, it looks like the Use Field Names strategy does not work as expected. Instead of the cell numbers, the Use Fields from Header strategy should reflect the cell field values.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            String dateFormat = context.getProperty(DateTimeUtils.DATE_FORMAT).getValue();
+            String timeFormat = context.getProperty(DateTimeUtils.TIME_FORMAT).getValue();
+            String timestampFormat = context.getProperty(DateTimeUtils.TIMESTAMP_FORMAT).getValue();
+            final TimeValueInference timeValueInference = new TimeValueInference(dateFormat, timeFormat, timestampFormat);
+            final Map<String, FieldTypeInference> typeMap = new LinkedHashMap<>();
+            IntStream.range(0, headerRow.getLastCellNum())
+                    .forEach(index -> {
+                        final Cell cell = headerRow.getCell(index);
+                        final String fieldName = Integer.toString(index);
```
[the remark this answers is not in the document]
dan-s1: @exceptionfactory Actually that is the expected behavior as this is not CSV hence the values in the header column do not necessarily have to be strings. I used the field index since it is predictable and unchanging. Personally I am really not sure why this strategy is necessary once there is the inference strategy which is more encompassing than the header strategy..

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/test/java/org/apache/nifi/excel/TestExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,123 @@
...
+        final Map<PropertyDescriptor, String> properties = new HashMap<>();
+        new ExcelReader().getSupportedPropertyDescriptors().forEach(prop -> properties.put(prop, prop.getDefaultValue()));
+        final PropertyContext context = new MockConfigurationContext(properties, null);
+        properties.put(ExcelReader.STARTING_ROW, "2");
+        ExcelHeaderSchemaStrategy headerSchemaStrategy = new ExcelHeaderSchemaStrategy(context, logger);
+
+        RecordSchema schema = headerSchemaStrategy.getSchema(null, getInputStream("simpleDataFormatting.xlsx"), null);
+
+        final List<String> fieldNames = schema.getFieldNames();
+        assertEquals(Arrays.asList("0", "1", "2", "3"), fieldNames);
```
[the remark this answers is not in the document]
dan-s1: Same response as above

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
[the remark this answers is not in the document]
dan-s1: @exceptionfactory Also the other problem  or "feature" with that is that all the field types will be strings.
I am open to whatever you prefer. Let me know whether we should remove it or modify it.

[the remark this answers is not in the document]
exceptionfactory: The string typing is also a problem as you noted. Theoretically that could also be changed to follow an approach similar to the Infer Schema strategy, reading all rows to determine potential values. In the interest of completing a first version, this strategy could be removed for now and evaluated more closely in a follow-on Jira issue.

mkalavala:
I verified this PR by executing a flow to convert xlsx to json . I tested with both ConvertRecord and QueryRecord processors. 
Note: Since the columns are denoted with numbers , i have to quote the QueryRecord select sqls like `select * from FLOWFILE where "0" <> ''`  .

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelRecordReaderConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+package org.apache.nifi.excel;
+
+import org.apache.nifi.serialization.record.RecordSchema;
+
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+public class ExcelRecordReaderConfiguration {
+    private RecordSchema schema;
+    private AtomicReferenceArray<String> requiredSheets;
```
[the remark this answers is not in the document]
dan-s1: @exceptionfactory You are right. It also simplifies a lot of code. I was under the impression there may be a threading issue.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,116 @@
...
+                        typeInference.addPossibleDataType(dataType);
+                    });
+
+            final List<RecordField> fields = typeMap.entrySet().stream()
+                    .map(entry -> new RecordField(entry.getKey(), entry.getValue().toDataType(), true))
+                    .collect(Collectors.toList());
+
+            return new SimpleRecordSchema(fields);
+        } catch (Exception e) {
+            throw new SchemaNotFoundException(errMsg, e);
```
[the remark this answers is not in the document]
dan-s1: Done

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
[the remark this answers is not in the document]
dan-s1: Done

dan-s1:
@exceptionfactory Can you please restart the Windows Zulu JDK 11 FR build? It timed out. Thanks!

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -0,0 +1,147 @@
...
+        this.sheets = this.workbook.iterator();
+        this.requiredSheets = requiredSheets != null ? requiredSheets.stream()
+                .collect(Collectors.toMap(key -> key, value -> Boolean.FALSE)) : new HashMap<>();
+        this.firstRow = firstRow;
+        this.logger = logger;
+    }
+
+    @Override
+    public boolean hasNext() {
+        setCurrent();
```
exceptionfactory, line 59: This approach requires calling `hasNext()` before calling `next()`, which may be conventional, but should not be required. Instead, the next row should be set after calling `next()` because that method should move the iterator forward.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelSchemaInference.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return createSchema(typeMap);
+    }
+
+    private void inferSchema(final Row row, final Map<String, FieldTypeInference> typeMap) {
+        if (ExcelUtils.hasCells(row)) {
+            IntStream.range(0, row.getLastCellNum())
+                    .forEach(index -> {
+                        final Cell cell = row.getCell(index);
+                        final String fieldName = Integer.toString(index);
```
exceptionfactory, line 68: What do you think of prefixing this field name with `column_`? So the first field name would be `column_0`? That would be easier to work with in absence of a defined schema.

exceptionfactory (commented):
Thanks for the adjustments @dan-s1, two additional comments and I think that should cover things.

I noticed that the `RowIterator` sets the current row in `hasNext()` instead of `next()`, which does not seem to be the way to contract should be implemented.

I also noted one more suggestion regarding field names for the Infer Schema strategy, to make it a bit easier to work with in this initial version.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/ExcelSchemaInference.java
dan-s1: Sounds good to me. Per the comment on the query earlier this would certainly be an improvement.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: Just trying to understand. With that refactor then one could never call hasNext since the instant variable currentRow is null.

exceptionfactory: It should be possible to call `hasNext()` in any order. Taking a closer look, it seems like the current implementation of `setCurrent` needs further refinement because it sets `currentSheet` to `null` on every invocation. One general approach could be to set the initial sheet and row in the constructor.

In general, it should be possible to call `hasNext()` and unlimited number of times without advancing the iterator, and it calling `next()` should advance the iterator forward. When all sheets and rows have been read, `next()` should throw an `IllegalStateException` if called again.

dan-s1: It looks like I could take out lines
```
currentSheet = null;
currentRows = null;
```
from setCurrent without any issues. Since the RowIterator is for internal use, could we just implement the next method and not the hasNext method? It seem there would be considerable amount of change needed just to have the hasNext method work.

exceptionfactory: Although it may not be the most intuitive approach, having `hasNext()` throw an `UnsupportedOperationException` and changing callers to just use `next()` could work, although it raises a question of logic handling for the calling class. Another option could be folding the functionality into the calling class.

I think it would be cleaner in the end to rework the RowIterator, in addition to removing the `public` visibility from the class. I could take a look at reworking RowIterator and pushing the changes to this branch if you think the changes are too involved.

dan-s1: I am reworking it

dan-s1: I simplified the RowIterator by doing some up front processing in the constructor to determine which sheets need to be processed. That allowed for simplifying later logic to be uniform just to act on the the sheets needed. I hope its okay that I have a call to `setCurrent()` in the constructor in order to initialize the `currentRow` to allow hasNext to be called initially.

dan-s1:
@exceptionfactory I just added the check to the `RowIterator` `next()` method to see whether to throw a `NoSuchElementException` or not.

dan-s1:
@exceptionfactory  Please let me know if there is anything else. Thanks!

exceptionfactory (commented):
Thanks for the persistence in working through the feedback @dan-s1!

The latest version looks good in terms of code and functionality. The basic capabilities of schema inference and configurable schemas appear to work as designed.

I should have noticed this earlier, but with the last change necessary is moving this Processor to the nifi-poi-bundle. The `excel-streaming-reader` dependency brings in at least 10 MB of additional dependencies, which are already present in `nifi-poi-nar`. I recommend creating a new module named `nifi-poi-services` and adding that to `nifi-poi-nar`. It will require moving all of the existing code to the new directory, but otherwise it should be a fairly direct refactor. I can assist with this if needed.

That move will avoid introducing additional copies of the dependency in the binary distribution, and then this should be ready to go.

dan-s1:
@exceptionfactory In order to do this I do need many of the dependencies that are already in `nifi-schema-registry-service-api` as well as classes that are in that module e.g. (`org.apache.nifi.schema.inference.InferSchemaAccessStrategy` and `org.apache.nifi.schema.inference.SchemaInferenceUtil`). Do I just include that module as a dependency which whereby I would get the other dependencies as transient dependencies?

dan-s1:
@exceptionfactory I found that I had to cut and paste files

1. org/apache/nifi/schema/inference/CachedSchemaAccessStrategy.java
2. org/apache/nifi/schema/inference/InferSchemaAccessStrategy.java
3. org/apache/nifi/schema/inference/SchemaInferenceUtil.java
4. org/apache/nifi/util/SchemaInferenceUtil.java

Wouldn't it make sense to move these files under module `nifi-schema-inference-utils` so both `nifi-record-serialization-services ` and `nifi-poi-services`  could use them ?

exceptionfactory:
> @exceptionfactory I found that I had to cut and paste files
> 
>     1. org/apache/nifi/schema/inference/CachedSchemaAccessStrategy.java
> 
>     2. org/apache/nifi/schema/inference/InferSchemaAccessStrategy.java
> 
>     3. org/apache/nifi/schema/inference/SchemaInferenceUtil.java
> 
>     4. org/apache/nifi/util/SchemaInferenceUtil.java
> 
> 
> Wouldn't it make sense to move these files under module `nifi-schema-inference-utils` so both `nifi-record-serialization-services ` and `nifi-poi-services` could use them ?

Yes, given the usage pattern, it moving the `schema.inference` classes to a shared module makes sense. It looks like `nifi-schema-inference-utils` does not depend on `nifi-record-serialization-services-api` right now, which is the reason for their current location. It actually seems like a `nifi-record-serialization-services-shared` module under `nifi-record-serialization-services-bundle` would be a reasonable approach. That would keep the classes closer to their current usage, but also support reuse in `nifi-poi-services` as planned.

dan-s1:
@exceptionfactory I added the new modules as you requested. Please let me know if these are what you envisioned.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/pom.xml
```diff
@@ -75,6 +75,11 @@
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-csv</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.pjfanning</groupId>
+            <artifactId>excel-streaming-reader</artifactId>
```
exceptionfactory, line 80: This dependency can be removed now that the ExcelReader is in a different module.

exceptionfactory (changes requested):
Thanks for making the adjustments @dan-s1, this is close to what I had in mind, but there is an issue with the `VolatileSchemaCache`. As a Controller Service, that needs to stay in the `nifi-record-serialization-services` module. Moving that module a service configuration to a separate JAR means that the Controller Service would be registered through multiple NAR files, which creates runtime problems. If that Service, test, and Controller Service reference change can be moved back, I think that should work as designed.

dan-s1:
@exceptionfactory Done

exceptionfactory:
Thanks for moving the `VolatileSchemaCache` back to the record services module. The latest code changes look good.

In the course of review, I noticed that `nifi-poi-services` was not included in `nifi-poi-nar`, so it wasn't being picked up at runtime. I also noticed that the `Required Sheets` property was not being evaluated with the attributes passed to the `createRecordReader` method. I pushed a commit correcting those two issues, along with adjusting spacing and adding the `final` keyword in a few places.

With those changes, this should be ready to go, and I will plan on merging pending successful build completion.

exceptionfactory (approved):
Thanks again for working through the feedback, addressing the issues, and contributing this new feature @dan-s1! +1 merging

dan-s1:
@exceptionfactory Thank you! It sure was a learning experience.

PR #7197, opened 2023-04-26
NIFI-11485 Add nifi toolkit to registry image

nathluu:
# Summary

[NIFI-11485](https://issues.apache.org/jira/browse/NIFI-11485)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


nifi-registry/nifi-registry-docker-maven/dockermaven/pom.xml
```diff
@@ -61,6 +61,22 @@
                                     <goal>run</goal>
                                 </goals>
                             </execution>
+                            <execution>
+                                <id>copy-toolkit-for-docker</id>
+                                <phase>process-sources</phase>
+                                <configuration>
+                                    <target name="copy toolkit assembly to nifi-registry-docker for image build">
```
MikeThomsen, line 68: I think it would better to use the approach used in the main docker image where the NiFi binary is downloaded rather than pulled from the nifi-assembly package.

https://github.com/apache/nifi/blob/3d60e4a4f602b88686806ee84d1da816c9c398a2/nifi-docker/dockerhub/Dockerfile#L58

MikeThomsen (changes requested):

nifi-registry/nifi-registry-docker-maven/dockermaven/pom.xml
nathluu: This change is to build a registry container image all from the source code, if there is a change in nifi registry code it will also be update the image with that change. By that way you should refer Dockerfile at [nifi-docker/dockermaven/Dockerfile](https://github.com/apache/nifi/blob/3d60e4a4f602b88686806ee84d1da816c9c398a2/nifi-docker/dockermaven/Dockerfile#L43) but not [nifi/nifi-docker/dockerhub/Dockerfile](https://github.com/apache/nifi/blob/3d60e4a4f602b88686806ee84d1da816c9c398a2/nifi-docker/dockerhub/Dockerfile#L58)

nathluu: I also updated the code to include changes to [nifi-registry/nifi-registry-core/nifi-registry-docker/dockerhub/Dockerfile](https://github.com/apache/nifi/blob/main/nifi-registry/nifi-registry-core/nifi-registry-docker/dockerhub/Dockerfile) which is used to publish registry official image to hub.docker.com. Please help to review.

ferencerdei:
The ticket description mentioned that you need the NiFi toolkit to encrypt the nifi-registry properties. In my opinion, you should be able to do this with an init container in Kubernetes. With your approach, you'd bundle a different application into nifi-registry which we shouldn't do.

nathluu:
Hi @ferencerdei,
nifi-toolkit is a utility tool released alongside every NiFi and NiFi registry release so I think it totally can be integrated into the NiFi Registry container image just as NiFi container image. The important reason is that the encryption of nifi-registry.properties file cannot happen at the pod startup time as it needs the nifi-registry.properties and bootstrap.conf files to exist beforehand. Please refer to these helm charts for details how a nifi and nifi registry pod is started
https://github.com/cetic/helm-nifi/blob/master/templates/statefulset.yaml
https://github.com/dysnix/charts/blob/main/dysnix/nifi-registry/templates/statefulset.yaml

exceptionfactory (approved):
Thanks for the addition @nathluu, and thanks for the feedback @MikeThomsen and @ferencerdei.

Although there are a variety of integration approaches when it comes to the NiFi Toolkit, including the Toolkit with Registry follows the same approach as the standard NiFi image. From that perspective, it seems reasonable to bundle the toolkit with Registry for now.

+1 merging

PR #7211, opened 2023-04-30
NIFI-8287: Updating SQS processors to use AWS SDK version 2.x

gresockj:
# Summary

[NIFI-8287](https://issues.apache.org/jira/browse/NIFI-8287)

A flow definition is available on the JIRA issue to facilitate testing.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/pom.xml
```diff
@@ -88,6 +98,11 @@
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-security-utils</artifactId>
```
exceptionfactory, line 103: The `nifi-security-utils` module includes a number of unnecessary items. Instead of using `KeyStoreUtils`, the builders in `nifi-security-ssl` should be used, avoiding this dependency.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/v2/AbstractAwsProcessor.java
```diff
@@ -0,0 +1,451 @@
...
+            .description("FlowFiles are routed to success relationship").build();
+    public static final Relationship REL_FAILURE = new Relationship.Builder().name("failure")
+            .description("FlowFiles are routed to failure relationship").build();
+
+    public static final Set<Relationship> relationships = Collections.unmodifiableSet(
+            new HashSet<>(Arrays.asList(REL_SUCCESS, REL_FAILURE)));
+
+    public static final PropertyDescriptor CREDENTIALS_FILE = CredentialPropertyDescriptors.CREDENTIALS_FILE;
+    public static final PropertyDescriptor ACCESS_KEY = CredentialPropertyDescriptors.ACCESS_KEY;
+    public static final PropertyDescriptor SECRET_KEY = CredentialPropertyDescriptors.SECRET_KEY;
```
exceptionfactory, line 97: Is it necessary to re-declare these property variables? Does that limit the changes required in subclasses?

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/v2/AbstractAwsProcessor.java
```diff
@@ -0,0 +1,451 @@
...
+        builder.socketTimeout(Duration.ofMillis(communicationsTimeout));
+        builder.maxConnections(context.getMaxConcurrentTasks());
+
+        if (this.getSupportedPropertyDescriptors().contains(SSL_CONTEXT_SERVICE)) {
+            final SSLContextService sslContextService = context.getProperty(SSL_CONTEXT_SERVICE).asControllerService(SSLContextService.class);
+            if (sslContextService != null) {
+                final TrustManager[] trustManagers = new TrustManager[] { sslContextService.createTrustManager() };
+                final KeyManagerFactory keyManagerFactory;
+                try {
+                    keyManagerFactory = KeyStoreUtils.loadKeyManagerFactory(sslContextService.createTlsConfiguration());
```
exceptionfactory, line 334: It looks like AWS has a `FileSToreTlsKeyManagersProvider` that could be used. That should avoid the need for KeyStoreUtils and `nifi-security-utils`.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/v2/AbstractAwsProcessor.java
```diff
@@ -0,0 +1,451 @@
...
+                    builder.tlsTrustManagersProvider(() -> trustManagers);
+                    builder.tlsKeyManagersProvider(() -> keyManagerFactory.getKeyManagers());
+                } catch (final TlsException e) {
+                    throw new RuntimeException("Failed to configure TLS Key Manager", e);
```
exceptionfactory, line 338: This should probably be something more specific like `IllegalArgumentException`.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/v2/AbstractAwsProcessor.java
```diff
@@ -0,0 +1,451 @@
...
+        return new AllowableValue(region.id(), description, "AWS Region Code : " + region.id());
+    }
+
+    public static AllowableValue[] getAvailableRegions() {
+        final List<AllowableValue> values = new ArrayList<>();
+        for (final Region region : Region.regions()) {
+            values.add(createAllowableValue(region));
+        }
+        return values.toArray(new AllowableValue[0]);
+    }
```
exceptionfactory, line 449: It probably makes more sense to move these methods to an enum or utility class.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/v2/AwsClientCache.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.nifi.processors.aws.v2;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import org.apache.nifi.processor.ProcessContext;
+import software.amazon.awssdk.core.SdkClient;
+
+public class AwsClientCache<T extends SdkClient> {
+
+    private static final int MAXIMUM_CACHE_SIZE = 10;
```
exceptionfactory, line 26: Any particular reason for the cache size of 10? It would be worth adding a comment of some kind.

exceptionfactory (changes requested):
Thanks for working on this refactoring @gresockj! Most of the changes appear straightforward. I noted one particular concern and alternative when it comes to handling the TLS Key Managers to avoid the dependency on `nifi-security-utils`, and a couple other minor comments.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/v2/AbstractAwsProcessor.java
gresockj: Yes, for example, there are 95 places `CREDENTIALS_FILE` is referenced by subclasses in the code.  Not strictly necessary to re-declare here, but would limit the required changes.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/v2/AwsClientCache.java
gresockj: This was copied from the v1 version of the AwsClientCache, but looking at it now I don't see a strong reason to have a cache size limit.  Since the cache key boils down to the AWS region, multiple clients would be cached if a subclass of AbstractAwsProcessor parsed the region from the ProcessContext (as in the case of S3).  However, since there are currently only 38 regions, and likely not to be more than a few dozen regions in the future, it doesn't seem like a big risk to remove the limit.

exceptionfactory (approved):
Thanks for making the adjustments @gresockj, the latest version looks good, all SQS Processors work as expected with this upgrade. +1 merging

PR #7228, opened 2023-05-04
NIFI-11524 Documentation and config experience improvement for ExecuteStreamCommand

nandorsoma:
# Summary

[NIFI-11524](https://issues.apache.org/jira/browse/NIFI-11524)

This PR indends to add documentation for the ExecuteSteamCommand processor and improves the configuration experience by adding dependency between properties.

I've noticed that the processor adds every Dynamic Property as environment variable even when it is a command argument. I didn't change this behavior because it might introduce a breaking change, but I'd change this behavior in a separate 2.0 only PR. Simply adding another regex check there would introduce the 4. regex check which is quite ugly so it would be a good opportunity for refactor. Another option would be to add validation for environment variables to ensure they don't contain `.` character, because most shells don't support it. But I'm not sure about that because there are commands that can utilize such env variables even though they are not common.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ExecuteStreamCommand.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ExecuteStreamCommand.java
@@ -65,6 +65,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.ProcessBuilder.Redirect;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -148,14 +149,15 @@ import java.util.regex.Pattern;
 @SupportsBatching
 @InputRequirement(Requirement.INPUT_REQUIRED)
 @Tags({"command execution", "command", "stream", "execute"})
-@CapabilityDescription("Executes an external command on the contents of a flow file, and creates a new flow file with the results of the command.")
+@CapabilityDescription("The ExecuteStreamCommand processor provides a flexible way to integrate external commands and scripts into NiFi data flows."
+        + " ExecuteStreamCommand can pass the incoming FlowFile's content to the command that it executes similarly how piping works.")
 @SupportsSensitiveDynamicProperties
 @DynamicProperties({
         @DynamicProperty(name = "An environment variable name", value = "An environment variable value",
                 description = "These environment variables are passed to the process spawned by this Processor"),
-        @DynamicProperty(name = "command.argument.<NUMBER>", value = "Argument to be supplied to the command",
+        @DynamicProperty(name = "command.argument.<commandIndex>", value = "Argument to be supplied to the command",
                 description = "These arguments are supplied to the process spawned by this Processor when using the "
-                        + "Command Arguments Strategy : Dynamic Property Arguments. The NUMBER will determine the order.")
+                        + "Command Arguments Strategy : Dynamic Property Arguments. <commandIndex> is a number and it will determine the order.")
 })
 @WritesAttributes({
         @WritesAttribute(attribute = "execution.command", description = "The name of the command executed"),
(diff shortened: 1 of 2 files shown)


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ExecuteStreamCommand.java
```diff
@@ -363,8 +348,37 @@ protected PropertyDescriptor getSupportedDynamicPropertyDescriptor(final String
...
+        if (DYNAMIC_PROPERTY_ARGUMENTS_STRATEGY.getValue() != argumentStrategy) {
+            for (final PropertyDescriptor propertyDescriptor : validationContext.getProperties().keySet()) {
+                if (!propertyDescriptor.isDynamic()) {
+                    continue;
+                }
+
+                final String propertyName = propertyDescriptor.getName();
+                final Matcher matcher = COMMAND_ARGUMENT_PATTERN.matcher(propertyName);
+                if (matcher.matches()) {
+                    logger.warn(String.format("'%s' should be set to '%s' when command arguments are supplied as Dynamic Properties. The property '%s' will be ignored.",
```
nandorsoma, line 375: In a separate 2.0 only PR I'd add an invalid validation result here. Is it ok?

exceptionfactory: That sounds like a good approach. For this pull request, `String.format()` should not be used for the log message, instead `{}` placeholders should be used.
```suggestion
                    logger.warn("[{}] should be set to [{}] when command arguments are supplied as Dynamic Properties. The property [{}] will be ignored.",
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/resources/docs/org.apache.nifi.processors.standard.ExecuteStreamCommand/additionalDetails.html
```diff
@@ -0,0 +1,166 @@
...
+                <td>/path/to/dir</td>
+            </tr>
+        </table>
+
+        <h3>Configuring environment variables</h3>
+        <p>In addition to specifying command arguments using the Command Argument field or Dynamic Properties, users can also use environment variables with
+        the ExecuteStreamCommand processor. Environment variables are a set of key-value pairs that can be accessed by processes running on the system.
+        ExecuteStreamCommand will treat every Dynamic Property as an environment variable that doesn't match the pattern 'command.arguments.&lt;commandIndex&gt;'.</p>
+
+        <p>Consider that we want to execute a maven command with the processor. If there are multiple Java versions installed on the system, you can specify
```
exceptionfactory, line 81: `Maven` should be capitalized.
```suggestion
        <p>Consider that we want to execute a Maven command with the processor. If there are multiple Java versions installed on the system, you can specify
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/resources/docs/org.apache.nifi.processors.standard.ExecuteStreamCommand/additionalDetails.html
```diff
@@ -0,0 +1,166 @@
...
+Java version: 11.0.18, vendor: Eclipse Adoptium, runtime: /path/to/another/java/home
+Default locale: en_US, platform encoding: UTF-8
+OS name: "mac os x", version: "13.1", arch: "x86_64", family: "mac"
+            </pre>
+        </code>
+
+        <h3>Streaming input to the command</h3>
+        <p>ExecuteStreamCommand passes the incoming FlowFile's content to the command that it executes similarly how piping works.
+        It is possible to disable this behavior with the Ignore STDIN property. In the above examples we didn't use the
+        incoming FlowFile's content, so in this cases we could leverage this property for additional performance gain.</p>
```
exceptionfactory, line 120:
```suggestion
        incoming FlowFile's content, so in this case we could leverage this property for additional performance gain.</p>
```

exceptionfactory (commented):
Thanks for the additional documentation and improvements @nandorsoma! I just noted a couple very minor things, otherwise this should be ready to go.

nandorsoma:
Thank you for your review @exceptionfactory! Please see my latest commit where I applied your suggestions.

PR #7231, opened 2023-05-05
[NIFI-2964] Added ability for AttributeToJSON to handle nested JSON when either outputting to a flow file or an attribute.

dan-s1:
# Summary

[NIFI-2964](https://issues.apache.org/jira/browse/NIFI-2964)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
@@ -65,8 +65,8 @@ import java.util.stream.Collectors;
 @WritesAttribute(attribute = "JSONAttributes", description = "JSON representation of Attributes")
 public class AttributesToJSON extends AbstractProcessor {
     public enum JsonHandlingStrategy implements DescribedValue {
-        ESCAPED_STRING("Escaped String", "Escapes JSON attribute values to strings"),
-        NESTED_OBJECT("Nested Object", "Handles JSON attribute values as nested structured objects");
+        ESCAPED("Escaped", "Escapes JSON attribute values to strings"),
+        NESTED("Nested", "Handles JSON attribute values as nested structured objects or arrays");
 
         JsonHandlingStrategy(String displayName, String description) {
             this.displayName = displayName;
@@ -158,7 +158,7 @@ public class AttributesToJSON extends AbstractProcessor {
             .required(true)
             .expressionLanguageSupported(ExpressionLanguageScope.NONE)
             .allowableValues(JsonHandlingStrategy.class)
-            .defaultValue(AttributesToJSON.JsonHandlingStrategy.ESCAPED_STRING.getValue())
+            .defaultValue(AttributesToJSON.JsonHandlingStrategy.ESCAPED.getValue())
             .build();
 
     public static final Relationship REL_SUCCESS = new Relationship.Builder().name("success")
@@ -304,7 +304,7 @@ public class AttributesToJSON extends AbstractProcessor {
     }
 
     private Map<String, Object> getFormattedAttributes(Map<String, Object> flowFileAttributes) throws JsonProcessingException {
-        if (JsonHandlingStrategy.ESCAPED_STRING == jsonHandlingStrategy) {
+        if (JsonHandlingStrategy.ESCAPED == jsonHandlingStrategy) {
             return flowFileAttributes;
         }
 
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
@@ -21,9 +21,9 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.annotation.behavior.EventDriven;
+import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.SideEffectFree;
 import org.apache.nifi.annotation.behavior.SupportsBatching;
-import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.WritesAttribute;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.Tags;
@@ -33,24 +33,24 @@ import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.flowfile.attributes.CoreAttributes;
 import org.apache.nifi.processor.AbstractProcessor;
-import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.ProcessorInitializationContext;
+import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
 
 import java.io.BufferedOutputStream;
 import java.io.OutputStream;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Collections;
 import java.util.stream.Collectors;
 
 @EventDriven
@@ -122,6 +122,16 @@ public class AttributesToJSON extends AbstractProcessor {
             .defaultValue("false")
             .build();
 
+    public static final PropertyDescriptor RETAIN_JSON_REGEX = new PropertyDescriptor.Builder()
+            .name("retain-json-regex")
+            .displayName("Retain Json Regular Expression")
+            .description("Regular expression that will be evaluated against the flow file attributes to select "
+                    + "the matching attributes whose value are already JSON and should remain as JSON and not be escaped.")
+            .required(false)
+            .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
+            .addValidator(StandardValidators.createRegexValidator(0, Integer.MAX_VALUE, true))
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .build();
 
     public static final Relationship REL_SUCCESS = new Relationship.Builder().name("success")
             .description("Successfully converted attributes to JSON").build();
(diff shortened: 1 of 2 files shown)


dan-s1:
@exceptionfactory I looked over my code since last Friday's initial commit and I realized there were other things that needed to be added so I have added 2 other commits. I believe for my PR for NIFI-11167 I squashed too much when I had to make changes. For this PR would you like me to squash these or should I leave this as is?

exceptionfactory:
Thanks @dan-s1, it's fine for now, if you end up making additional changes before further review, that squashing commits sounds good, otherwise it looks good for now.

dan-s1:
@exceptionfactory It was brought to my attention with my previous changes I was not keeping the current functionality of escaping JSON. I therefore pushed another commit which adds a property to allow for filtering which attributes should be retained as JSON or when left blank all attributes are escaped as the current functionality is. I also squashed the other commits.

exceptionfactory (commented):
Thanks for noting the compatibility issue @dan-s1. Unfortunately, I think the fact of compatibility issues highlights some of the potential complexity that this introduces.

If we go forward with this change, it seems better to have a simple property like `JSON Handling Strategy` with values of `Escaped String` or `Nested Object`. That way, anything detected as `JSON` would be treated the same way, without the potential complexity of pattern-matching on FlowFile attribute names. What do you think of that approach?

dan-s1:
@exceptionfactory Sounds good to me.

dan-s1:
@exceptionfactory I agree with you that is counter intuitive so much so my first iteration of this PR actually always converted to JSON. I was told though that converting all attributes to JSON would cause a problem in a particular flow. So per your suggestion could we have 3 allowable values, `Escaped String`, `Nested Object` and `Pick&Choose`? If  `Pick&Choose` is chosen then there is another property which depends on it which would be the regex for picking which attributes should be converted to JSON.

exceptionfactory:
Thanks @dan-s1, having three allowable values sounds like it could be the best solution in terms of both flexibility and usability. For the third option, it could be something like `Nested Object for Matched Attributes`, and then the dependent property could be `JSON Attribute Names Pattern`.

I am still wondering about the details of the flow in terms of why selective conversion might be necessary. It adds code complexity, which impacts future maintenance. Can you provide any additional details on the use case, and why it would work in conjunction with other Processors?

dan-s1:
@exceptionfactory After much deliberation we realized you were right that NIFI has other processors to handle the JSON transformations needed. We would still like though the ability to handle attributes as nested objects as you suggested:

> If we go forward with this change, it seems better to have a simple property like JSON Handling Strategy with values of Escaped String or Nested Object. That way, anything detected as JSON would be treated the same way, without the potential complexity of pattern-matching on FlowFile attribute names.

This would clearly highlight the fact that the processor can handle nested objects. Is that still okay?

exceptionfactory:
> @exceptionfactory After much deliberation we realized you were right that NIFI has other processors to handle the JSON transformations needed. We would still like though the ability to handle attributes as nested objects as you suggested:
> 
> > If we go forward with this change, it seems better to have a simple property like JSON Handling Strategy with values of Escaped String or Nested Object. That way, anything detected as JSON would be treated the same way, without the potential complexity of pattern-matching on FlowFile attribute names.
> 
> This would clearly highlight the fact that the processor can handle nested objects. Is that still okay?

Thanks for evaluating the options. Yes, I think the Handling Strategy approach with the two options provides clarity on both the implementation side and the usability side. If you can make those adjustments to the PR, that would be great.

dan-s1:
@exceptionfactory Added the handling strategy as we discussed.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
```diff
@@ -240,24 +281,51 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
...
             session.transfer(original, REL_FAILURE);
         }
     }
+
+    private Map<String, Object> getAsJson(Map<String, Object> atrList) throws JsonProcessingException {
+        if(JsonHandlingStrategy.ESCAPED_STRING.equals(jsonHandlingStrategy)) {
+            return atrList;
+        }
+
+        Map<String, Object> asJson = new HashMap<>();
```
exceptionfactory, line 311: Recommend using `LinkedHashMap` to preserve key ordering.
```suggestion
        Map<String, Object> asJson = new LinkedHashMap<>();
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
```diff
@@ -240,24 +281,51 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
...
                 session.transfer(atFlowfile, REL_SUCCESS);
             }
         } catch (JsonProcessingException e) {
             getLogger().error(e.getMessage());
             session.transfer(original, REL_FAILURE);
         }
     }
+
+    private Map<String, Object> getAsJson(Map<String, Object> atrList) throws JsonProcessingException {
+        if(JsonHandlingStrategy.ESCAPED_STRING.equals(jsonHandlingStrategy)) {
```
exceptionfactory, line 307: This can be simplified:
```suggestion
        if (JsonHandlingStrategy.ESCAPED_STRING == jsonHandlingStrategy) {
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
```diff
@@ -240,24 +281,51 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
...
+                FlowFile atFlowfile = session.putAttribute(original, JSON_ATTRIBUTE_NAME, objectMapper.writeValueAsString(asJson));
                 session.transfer(atFlowfile, REL_SUCCESS);
             }
         } catch (JsonProcessingException e) {
             getLogger().error(e.getMessage());
             session.transfer(original, REL_FAILURE);
         }
     }
+
+    private Map<String, Object> getAsJson(Map<String, Object> atrList) throws JsonProcessingException {
```
exceptionfactory, line 306: Recommend adjusting the argument name. The method name should also be adjusted to more accurately reflect the fact that the input Map may not be changed.
```suggestion
    private Map<String, Object> getFormattedAttributes(final Map<String, Object> flowFileAttributes) throws JsonProcessingException {
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
```diff
@@ -59,9 +60,37 @@
...
 @WritesAttribute(attribute = "JSONAttributes", description = "JSON representation of Attributes")
 public class AttributesToJSON extends AbstractProcessor {
+    public enum JsonHandlingStrategy implements DescribedValue {
+        ESCAPED_STRING("Escapes any nested JSON as a string", "Escaped String"),
+        NESTED_OBJECT("Handles nested JSON as JSON", "Nested Object");
+
+        private final String description;
+        private final String displayName;
+
+        JsonHandlingStrategy(String description, String displayName) {
```
exceptionfactory, line 74: As a general convention, recommend placing the display name before the description.
```suggestion
        ESCAPED_STRING("Escaped String", "Escapes JSON attribute values to strings"),
        NESTED_OBJECT("Nested Object", "Handles JSON attribute values as nested structured objects");

        private final String description;
        private final String displayName;

        JsonHandlingStrategy(String displayName, String description) {
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
```diff
@@ -122,6 +151,15 @@ public class AttributesToJSON extends AbstractProcessor {
             .defaultValue("false")
             .build();
 
+    public static final PropertyDescriptor JSON_HANDLING_STRATEGY = new PropertyDescriptor.Builder()
+            .name("Json Handling Strategy")
+            .displayName("Json Handling Strategy")
```
exceptionfactory, line 156: `JSON` should be all-caps:
```suggestion
            .name("JSON Handling Strategy")
            .displayName("JSON Handling Strategy")
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
```diff
@@ -122,6 +151,15 @@ public class AttributesToJSON extends AbstractProcessor {
...
+            .description("Strategy to use for handling attributes which contain nested JSON.")
+            .required(true)
+            .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
```
exceptionfactory, line 159: Recommend leaving this as `NONE` since it can also be parameterized using Parameter Contexts, instead of the Variable Registry or environment variables, which are not preferred.

exceptionfactory (changes requested):
Thanks for implementing the new property changes @dan-s1. This looks close to completion, just noted a few additional recommendations.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AttributesToJSON.java
dan-s1: Should I add that there or leave it as is? I just pushed the changes you had requested.

exceptionfactory: Thanks, if you could push that change as well, that would be helpful.

dan-s1: Done. Let me know if there is anything else.

exceptionfactory (commented):
Thanks for making the latest updates @dan-s1, the functionality looks good.

In the course of final review, one final question: what do you think about shortening the strategy values to just `Escaped` and `Nested`? This came to mind since the nested structure could be an array. Other than that, this looks ready to go.

dan-s1:
No problem. Should the display name for `Nested` also be changed to `Nested`?
In addition should the description for `Nested` be changed from 

> Handles JSON attribute values as nested structured objects

to 
"Handles JSON attribute values as nested structured objects or arrays."

exceptionfactory:
> No problem. Should the display name for `Nested` also be changed to `Nested`? In addition should the description for `Nested` be changed from
> 
> > Handles JSON attribute values as nested structured objects
> 
> to "Handles JSON attribute values as nested structured objects or arrays."

Thanks, yes, shortening both the enum and the display name to would be good. Updating the description would also be helpful.

The net result should be `NESTED("Nested", ...)` and `ESCAPED("Escaped", ...)` where `...` is the description.

dan-s1:
@exceptionfactory  Done. Let me know if there is anything else.

exceptionfactory (approved):
Thanks again for working through the feedback on this new feature @dan-s1, the latest version looks good! Planning to merge soon.

PR #7240, opened 2023-05-12
NIFI-11178: Improve ListHDFS performance, incremental loading refactor.

Lehel44:
# Summary

[NIFI-11178](https://issues.apache.org/jira/browse/NIFI-11178)

- Removed FileStatus collections to improve memory usage. The processor sends out every FileStatus to a record or a flowfile if it's eligibe for listing
- Introduced HdfsObjectWriter (similar to S3ObjectWriter) to handle record and flowfile writing with a common interface.
- Refactored the recursive file listing method to an iterator, moved the branching conditions to the createPathFilter logic.
- Completely refactored incremental loading. The last entries are not held back anymore.
- Added logic for legacy state transitioning.
- Fixed unit tests
- Updated processor documentation.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/ListHDFS.java
```diff
@@ -452,243 +305,201 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-        return new MapRecord(getRecordSchema(), values);
-    }
+    private FileCountRemoteIterator<FileStatus> getFileStatusIterator(final Path path, final boolean recursive, final FileSystem hdfs, final PathFilter filter) {
+        final Deque<Path> pathStack = new ArrayDeque<>();
+        pathStack.push(path);
 
-    private RecordSchema getRecordSchema() {
-        return RECORD_SCHEMA;
-    }
+        return new FileCountRemoteIterator<>() {
```
tpalfy, line 401: This iterator could be simplified if the stack was handled within and the filtering is not done here.
(`ListHDFS` already does some of in the `determineListable` so all the filtering could be combined in one place.)
Also probably better for it to implement `Iterable` instead of `RemoteIterator`.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/HdfsObjectWriter.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.nifi.schema.access.SchemaNotFoundException;
+
+import java.io.IOException;
+
+/**
+ * Interface for common management of writing to records and to flowfiles.
+ */
+public interface HdfsObjectWriter {
```
tpalfy, line 27: This interface has many methods that are only implemented by one of its two implementing classes. That is usually a sign that the separation of concerns is not adequate.
In this case a single `write` method or something similar would be probably better.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/ListHDFS.java
```diff
@@ -452,243 +305,201 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-            final long fileModTime = status.getModificationTime();
-            if (fileModTime > latestTimestampEmitted) {
-                latestTimestampEmitted = fileModTime;
-            }
+            long totalFileCount = fileStatusIterator.getFileCount();
+            getLogger().debug("Found a total of {} files in HDFS, {} are listed", totalFileCount, listedFileCount);
+        } catch (final IOException | IllegalArgumentException | SchemaNotFoundException e) {
+            getLogger().error("Failed to perform listing of HDFS", e);
+            writer.finishListingExceptionally(e);
+            return;
```
tpalfy, line 341: This relates to my previous comment about `HdfsObjectWriter` being too granular.
When we have a `FlowFileObjectWriter` we don't remove the created flowfiles from the session  and neither do we rollback the session. So the flowfiles will be emitted.
However the status won't be saved so we will emit the flowfiles for the same listed files again when the `onTrigger` runs next time. This leads to duplication of flowfiles.

Consolidating the `HdfsObjectWriter` interface will probably solve this problem as well.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/FileStatusIterable.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class FileStatusIterable implements Iterable<FileStatus> {
+
+    private final Path path;
+    private final boolean recursive;
+    private final FileSystem hdfs;
+    private final UserGroupInformation userGroupInformation;
+    private final AtomicLong totalFileCount = new AtomicLong();
+
+    public FileStatusIterable(final Path path, final boolean recursive, final FileSystem hdfs, final UserGroupInformation userGroupInformation) {
```
tpalfy, line 35: Minor: Just wondering if we could get rid of either the `FileStatusIterable` or the `FileStatusIterator`. Specifically the `Iterable` is really only used so that we can have a `foreach` instead of a `for` calling `hasNext` and `next` explicitly.
This is more declarative though so just a thought.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/FileStatusIterator.java
```diff
@@ -0,0 +1,96 @@
...
+import java.security.PrivilegedExceptionAction;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class FileStatusIterator implements Iterator<FileStatus> {
+
+    private static final String IO_ERROR_MESSAGE = "IO error occured while iterating HFDS";
```
tpalfy, line 35:
```suggestion
    private static final String IO_ERROR_MESSAGE = "IO error occurred while iterating HFDS";

```

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/writer/RecordObjectWriter.java
```diff
@@ -0,0 +1,187 @@
...
+                session.remove(flowFile);
+            } else {
+                final Map<String, String> attributes = new HashMap<>(writeResult.getAttributes());
+                attributes.put("record.count", String.valueOf(writeResult.getRecordCount()));
+                flowFile = session.putAllAttributes(flowFile, attributes);
+
+                session.transfer(flowFile, REL_SUCCESS);
+            }
+        } catch (Exception e) {
+            throw new ProcessException("An error occured while writing results", e);
```
tpalfy, line 141:
```suggestion
            throw new ProcessException("An error occurred while writing results", e);

```

tpalfy (changes requested):

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/writer/FlowFileObjectWriter.java
```diff
@@ -0,0 +1,117 @@
...
+                                final long latestModificationTime, final List<String> latestModifiedStatuses) {
+        this.session = session;
+        this.fileStatuses = fileStatuses;
+        this.minimumAge = minimumAge;
+        this.maximumAge = maximumAge;
+        this.pathFilter = pathFilter;
+        this.fileStatusManager = fileStatusManager;
+        this.latestModificationTime = latestModificationTime;
+        this.latestModifiedStatuses = latestModifiedStatuses;
+        listedFileCount = 0;
```
turcsanyip, line 61: The common fields should be extracted into an abstract class.
`determineListable()` should be moved there too and in that case most of the method parameters are not needed.
There are also other common methods too: `getAbsolutePath()`, `getPerms()`, `getPermissionsString()`, `getListedFileCount()`

turcsanyip (changes requested):

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/ListHDFS.java
```diff
@@ -319,131 +240,36 @@ public void onPropertyModified(final PropertyDescriptor descriptor, final String
...
+            final String legacyLatestListingTimestampString = stateMap.get(LEGACY_LISTING_TIMESTAMP_KEY);
+            final String legacyLatestEmittedTimestampString = stateMap.get(LEGACY_EMITTED_TIMESTAMP_KEY);
+
+            if (legacyLatestListingTimestampString != null) {
+                final long legacyLatestListingTimestamp = Long.parseLong(legacyLatestListingTimestampString);
+                final long legacyLatestEmittedTimestamp = Long.parseLong(legacyLatestEmittedTimestampString);
+                latestModificationTime = legacyLatestListingTimestamp == legacyLatestEmittedTimestamp ? legacyLatestListingTimestamp + 1 : legacyLatestListingTimestamp;
+            } else if (latestListedTimestampString != null) {
+                latestModificationTime = Long.parseLong(latestListedTimestampString);
+                latestModifiedStatuses = new ArrayList<>(Arrays.asList(latestFiles.split("\\s")));
```
turcsanyip, line 270: It looks minor but it is hard to follow why `latestListedTimestampString` becomes `latestModificationTime` after a simple parsing.
Could not we just call there variables: `latestTimestampString`, `latestTimestamp`, `latestFilesString`, `latestFiles`?
It would be more consistent with the state property names too.
I would apply these names in `FileStatusManager` too.

turcsanyip, line 270: Are we sure that file and directory names cannot contain spaces within the path?
Space separated paths may not work, I'm afraid.
Just an idea but other processors add the paths in separate entries in the state like `latest.file.1`, `latest.file.2`, etc.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/FilterMode.java
```diff
@@ -0,0 +1,84 @@
...
+    public String getDescription() {
+        return description;
+    }
+
+    public static FilterMode forName(String filterMode) {
+        return Stream.of(values())
+                .filter(fm -> fm.getValue().equalsIgnoreCase(filterMode))
+                .findFirst()
+                .orElseThrow(
+                        () -> new IllegalArgumentException("Invalid SnowflakeInternalStageType: " + filterMode));
```
markap14, line 82: Seems to be a copy/paste error - SNowflakeInternalStageType??

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/FilterMode.java
```diff
@@ -0,0 +1,84 @@
...
+    public String getDisplayName() {
+        return displayName;
+    }
+
+    @Override
+    public String getDescription() {
+        return description;
+    }
+
+    public static FilterMode forName(String filterMode) {
```
markap14, line 77: Is there a reason we need this `forName` instead of just using `valueOf`?

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/ListHDFS.java
```diff
@@ -286,392 +200,141 @@ protected Collection<ValidationResult> customValidate(ValidationContext context)
...
-        final boolean recursive = context.getProperty(RECURSE_SUBDIRS).asBoolean();
-        String fileFilterMode = context.getProperty(FILE_FILTER_MODE).getValue();
+        try (final FileSystem hdfs = getFileSystem()) {
+            final boolean recursive = context.getProperty(RECURSE_SUBDIRS).asBoolean();
+            final PathFilter pathFilter = createPathFilter(context);
+            final RecordSetWriterFactory writerFactory = context.getProperty(RECORD_WRITER).asControllerService(RecordSetWriterFactory.class);
 
-        final Set<FileStatus> statuses;
-        try {
+            final FileStatusManager fileStatusManager = new FileStatusManager();
```
turcsanyip, line 260: It may happen that a new file arrives with the same timestamp as the latest one(s) from the previous run and there are no other newer files in this round (so no new timestamp). In this case, the file should be appended to the others already in the state having the same latest timestamp.
So I believe `FileStatusManager` should be initialized with `latestTimestamp` and `latestFiles` from `onTrigger()` instead of 0 / empty list in its default constructor.
In this case, it would also be enough to pass `FileStatusManager` to the object writers and they can retrieve the previous latest values from it in the constructor.

turcsanyip (changes requested):

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/ListHDFS.java
tpalfy: I agree with @turcsanyip except the last part. Conceptually I would advise against initializing a final field from a non-final running field, even if it's hasn't changed at that point yet. I would keep those fields in the the object writer constructor.

tpalfy: Also a heads up when passing the file list. In the non-legacy case it is created like this:
```
latestFiles = stateMap.toMap().entrySet().stream()
                        .filter(entry -> entry.getKey().startsWith("latest.file"))
                        .map(Map.Entry::getValue)
                        .collect(Collectors.toList());
```
Although this is usually a mutable list the specification doesn't guarantee it. So the `FileStatusManger` constructor should create an instance of it's own, like this for example:
```
    public FileStatusManager(final long initialLatestTimestamp, final List<String> initialLatestFiles) {
        this.latestTimestamp = initialLatestTimestamp;
        this.latestFiles = new ArrayList<>(initialLatestFiles);
    }
```

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/FileStatusIterable.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    class FileStatusIterator implements Iterator<FileStatus> {
+
+        private static final String IO_ERROR_MESSAGE = "IO error occurred while iterating HFDS";
+
+        private final Deque<Path> dirStatuses;
+
+        private FileStatus nextFileStatus;
+        private RemoteIterator<FileStatus> hdfsIterator;
```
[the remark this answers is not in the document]
Lehel44: Thanks for the suggestion. The outer custom iterator class is also called FileStatusIterator, I'd avoid the "hiding field" code smell, remoteIterator works well.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/writer/HdfsObjectWriter.java
```diff
@@ -0,0 +1,108 @@
...
+        fileCount = 0L;
+    }
+
+    public abstract void write();
+
+    public long getListedFileCount() {
+        return fileCount;
+    }
+
+    boolean determineListable(final FileStatus status, final long minimumAge, final long maximumAge, final PathFilter filter,
```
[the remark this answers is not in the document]
Lehel44: Indeed, there is no need for them.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/writer/HdfsObjectWriter.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.nifi.processor.ProcessSession;
+import org.apache.nifi.processors.hadoop.util.FileStatusIterable;
+import org.apache.nifi.processors.hadoop.util.FileStatusManager;
+
+import java.util.List;
+
+/**
+ * Interface for common management of writing to records and to flowfiles.
+ */
+public abstract class HdfsObjectWriter {
```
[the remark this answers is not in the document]
Lehel44: Yes I thought about simply FileStatusWriter but HadoopFileStatusWriter is more describing.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/ListHDFS.java
```diff
@@ -286,392 +200,141 @@ protected Collection<ValidationResult> customValidate(ValidationContext context)
...
-                    final String listingTimestmapString = stateMap.get(LISTING_TIMESTAMP_KEY);
-                    if (listingTimestmapString != null) {
-                        latestTimestampListed = Long.parseLong(listingTimestmapString);
-                    }
-
-                    getLogger().debug("Found new-style state stored, latesting timestamp emitted = {}, latest listed = {}",
-                        new Object[] {latestTimestampEmitted, latestTimestampListed});
-                }
+            latestTimestamp = 0L;
+            latestFiles = new ArrayList<>();
```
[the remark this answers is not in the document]
Lehel44: Thanks for the suggestion, I'll change it.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/ListHDFS.java
```diff
@@ -36,642 +33,306 @@
...
+                return path -> fileFilterRegexPattern.matcher(path.toString()).matches()
                         || fileFilterRegexPattern.matcher(Path.getPathWithoutSchemeAndAuthority(path).toString()).matches();
-            } else {
-                accepted =  fileFilterRegexPattern.matcher(path.getName()).matches();
-            }
-            return accepted;
-        };
+            // FILTER_DIRECTORIES_AND_FILES
+            default:
+                return path -> Stream.of(path.toString().split("/"))
```
tpalfy, line 318: This case is not handling the path properly. It is working on the entire url. This can lead to dropping files that should be listed.

tpalfy (changes requested):

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/writer/RecordObjectWriter.java
```diff
@@ -0,0 +1,154 @@
...
+            } else {
+                final Map<String, String> attributes = new HashMap<>(writeResult.getAttributes());
+                attributes.put("record.count", String.valueOf(writeResult.getRecordCount()));
+                flowFile = session.putAllAttributes(flowFile, attributes);
+
+                session.transfer(flowFile, REL_SUCCESS);
+            }
+        } catch (Exception e) {
+            throw new ProcessException("An error occurred while writing results", e);
+        }
```
tpalfy, line 129: The `session.putAllAttributes` throws an exception while the `OutputStream out = session.write` is still open.
We can fix it like this:
```suggestion
        final WriteResult writeResult;
        try (
            final OutputStream out = session.write(flowFile);
            final RecordSetWriter recordWriter = writerFactory.createWriter(logger, RECORD_SCHEMA, out, flowFile)
        ) {
            recordWriter.beginRecordSet();

            for (FileStatus status : fileStatusIterable) {
                if (determineListable(status)) {
                    recordWriter.write(createRecordForListing(status));
                    fileStatusManager.update(status);
                }
            }

            writeResult = recordWriter.finishRecordSet();
        } catch (Exception e) {
            throw new ProcessException("An error occurred while writing results", e);
        }
        
        fileCount = writeResult.getRecordCount();
        if (fileCount == 0) {
            session.remove(flowFile);
        } else {
            final Map<String, String> attributes = new HashMap<>(writeResult.getAttributes());
            attributes.put("record.count", String.valueOf(writeResult.getRecordCount()));
            flowFile = session.putAllAttributes(flowFile, attributes);

            session.transfer(flowFile, REL_SUCCESS);
        }
```

tpalfy (changes requested):

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/pom.xml
```diff
@@ -154,6 +154,11 @@
             <groupId>org.glassfish.jaxb</groupId>
             <artifactId>jaxb-runtime</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-mock-record-utils</artifactId>
+            <scope>compile</scope>
```
tpalfy, line 160:
```suggestion
        <dependency>
            <groupId>org.apache.nifi</groupId>
            <artifactId>nifi-mock-record-utils</artifactId>
            <scope>test</scope>
```

PR #7242, opened 2023-05-12
NIFI-11471: Define new stateless configuration points

ohnoitsyou:
NIFI-11471:
Add two new properties:

  nifi.stateless.component.enableTimeout
  nifi.stateless.processor.startTimeout

to allow configuring the StatelessEngine and ProcessScheduler.

This allows an operator to configure what kind of startup time the flow can tolerate.

Previously these values were hard coded.

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


exceptionfactory:
Thanks for the contribution @ohnoitsyou. All of the automated checks failed, so please run the local `contrib-check` profile with a full build to ensure all existing tests pass. As a framework-level change, this will require some careful review.

ohnoitsyou:
I think I've got it figured out, just dealing with the intermittent test failures

nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/config/PropertiesFileEngineConfigurationParser.java
```diff
@@ -53,6 +54,9 @@ public class PropertiesFileEngineConfigurationParser {
     private static final String CONTENT_REPO_DIRECTORY = PREFIX + "content.repository.directory";
     private static final String STATUS_TASK_INTERVAL = PREFIX + "status.task.interval";
 
+    private static final String COMPONENT_ENABLE_TIMEOUT = PREFIX + "component.enableTimeout";
+    private static final String PROCESSOR_START_TIMEOUT = PREFIX + "processor.startTimeout";
```
markap14, line 58: We should use dotted names instead of camel-case - `component.enable.timeout` and `processor.start.timeout` rather than `enableTimeout` and `startTimeout`

nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/config/PropertiesFileEngineConfigurationParser.java
```diff
@@ -111,6 +115,9 @@ public StatelessEngineConfiguration parseEngineConfiguration(final File properti
 
         final String statusTaskInterval = properties.getProperty(STATUS_TASK_INTERVAL, "1 min");
 
+        final long processorStartTimeout = TimeUnit.SECONDS.toMillis(Long.parseLong(properties.getProperty(PROCESSOR_START_TIMEOUT, "10")));
+        final long componentEnableTimeout = TimeUnit.SECONDS.toMillis(Long.parseLong(properties.getProperty(COMPONENT_ENABLE_TIMEOUT, "10")));
```
markap14, line 119: Timeout properties should never be integer values - they need to always be time periods. So "10 secs" is a reasonable default, but we need to parse the property value using `FormatUtils.getTimeDuration`

nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/config/PropertiesFileEngineConfigurationParser.java
```diff
@@ -161,6 +168,16 @@ public List<ExtensionClientDefinition> getExtensionClients() {
             public String getStatusTaskInterval() {
                 return statusTaskInterval;
             }
+
+            @Override
+            public long getProcessorStartTimeout() {
```
markap14, line 173: Should take an argument for the `TimeUnit` to use

nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/engine/StatelessEngineConfiguration.java
```diff
@@ -89,4 +90,20 @@ default boolean isLogExtensionDiscovery() {
      * A <code>null</code> value indicates that no status tasks are scheduled.
      */
     String getStatusTaskInterval();
+
+    /**
+     * @return a String representing the number of milliseconds that the process scheduler should wait for a process to start
+     * Defaults to 10 seconds or 10_000 milliseconds
+     */
+    default long getProcessorStartTimeout() {
```
markap14, line 98: Should take an argument for the `TimeUnit` to use

nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/engine/StatelessEngineConfiguration.java
```diff
@@ -89,4 +90,20 @@ default boolean isLogExtensionDiscovery() {
...
+       return TimeUnit.SECONDS.toMillis(10);
+    }
+
+    /**
+     * @return a String representing the number of milliseconds that the StatelessEngine should wait for a component to enable
+     * Defaults to 10 seconds or 10_000 milliseconds
+     */
+    default long getComponentEnableTimeout() {
```
markap14, line 106: Should take an argument for the `TimeUnit` to use

nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/engine/StandardStatelessEngine.java
```diff
@@ -733,6 +742,11 @@ public Builder statusTaskInterval(final String statusTaskInterval) {
             return this;
         }
 
+        public Builder statelessEngineConfiguration(final StatelessEngineConfiguration statelessEngineConfiguration) {
+            this.statelessEngineConfiguration = statelessEngineConfiguration;
+            return this;
+        }
```
markap14, line 748: The only thing that we need to pass in here is the timeout - we should avoid passing in the entire StatelessEngineConfiguration and just take in the timeout

nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/flow/StatelessFlowConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nifi.stateless.flow;
+
+public class StatelessFlowConfiguration {
```
markap14, line 20: I don't think this class is necessary. It's simply a wrapper around a long. But a `StatelessFlowConfiguration` would imply a great deal more configuration. We should instead just provide the timeout to the Stateless Flow

nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/engine/StatelessEngine.java
```diff
@@ -71,4 +71,6 @@ public interface StatelessEngine {
     CounterRepository getCounterRepository();
 
     Duration getStatusTaskInterval();
+
+    StatelessEngineConfiguration getStatelessEngineConfiguration();
```
markap14, line 75: This is unnecessary. It's not used anywhere.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/controller/ProcessSchedulerConfig.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nifi.controller;
+
+public class ProcessSchedulerConfig {
```
markap14, line 20: Not sure that this class really provides any benefit over providing the timeout directly to the scheduler.

markap14 (changes requested):
Thanks for the update @ohnoitsyou. I do think that it introduces a good bit of complexity that is unnecessary, though, and does not adhere to standard practices such as using time periods like `10 secs` for properties. See comments inline. Thanks!

nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/engine/StatelessEngineConfiguration.java
exceptionfactory: Using `java.time.Duration` could provide another more flexible approach that avoids confusion related to time units.

nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/config/PropertiesFileEngineConfigurationParser.java
markap14: Fair enough.

nifi-stateless/nifi-stateless-bundle/nifi-stateless-engine/src/main/java/org/apache/nifi/stateless/flow/StatelessFlowConfiguration.java
markap14: Additional config elements can be passed in the constructor as well. Or we could use a Builder. Typically I'd recommend a Builder pattern after about 4-5 constructor args. We could also potentially provide a Configuration type of class like this in the future, if necessary. If this were a public API I think a "config" type of property might make more sense - but we would either want to rename it, or truly encapsulate all of the configuration elements for a Stateless Flow in it. But in this case, it's not a public API, so we can change it any time.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/controller/ProcessSchedulerConfig.java
markap14: See comment above.

nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/config/PropertiesFileEngineConfigurationParser.java
ohnoitsyou: @markap14 After a quick chat with David H. I brought up that FormatUtils is located within `nifi-utils` which is something stateless doesn't already import. What would be your opinion on including this? If the answer is "don't", how would you recommend converting from strings to `Duration`?

markap14: So it's okay to bring `nifi-utils` into `nifi-stateless-engine`. Definitely not ok to bring it into `nifi-stateless-api` though. So we should probably have the `StatelessEngineConfiguration` return a `String` representation instead of a `Duration`. Then, do the conversion in the `nifi-stateless-engine` module where it'll be used.

It probably actually makes sense to move the FormatUtils functionality into nifi-api because there have been other places where it'd be nice to have it available, such as directly in NiFiProperties. But that's beyond the scope of this Jira :) But something to consider for NiFi 2.0

exceptionfactory: Thanks for the summary @markap14, the FormatUtils has some deprecated methods that need to be removed for NiFi 2.0, and I could see value in pulling it out of `nifi-utils` and into `nifi-api`. In the meantime, it sounds like having the Configuration method return a String value and doing the conversion elsewhere is the best option.

ohnoitsyou:
@markap14 @exceptionfactory I have rebased this branch on the current main and made what I believe are the requested fixes. It appears to build on my end. If you could approve the workflow, I think it'll build successfully here in CI.

LMK if you think I've captured what you've asked for.

markap14:
Thanks @ohnoitsyou changes look good to me. Was able to verify behavior and added and additional system test to verify. +1 merged to both `main` and `support/nifi-1.x` branches.

PR #7248, opened 2023-05-15
[NIFI-11528] Refactored Groovy test in nifi-commons/nifi-utils to Jav…

dan-s1:
…a (and JUnit 5)

# Summary

[NIFI-11528](https://issues.apache.org/jira/browse/NIFI-11528)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


nifi-commons/nifi-utils/src/test/java/org/apache/nifi/util/TestFormatUtils.java
```diff
@@ -0,0 +1,368 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class TestFormatUtils {
+    @SuppressWarnings("deprecation")
+    @Test
+    public void testGetTimeDurationShouldConvertWeeks() {
+        final List<String> WEEKS = List.of("1 week", "1 wk", "1 w", "1 wks", "1 weeks");
+        final long EXPECTED_DAYS = 7L;
```
exceptionfactory, line 38: Understanding that this was copied over from the Groovy implementation, can you review the code and update the method variables to follow the standard camel case convention?

nifi-commons/nifi-utils/src/test/java/org/apache/nifi/util/TestFormatUtils.java
```diff
@@ -0,0 +1,368 @@
...
+    /**
+     * Positive flow test for decimal inputs that are extremely small
+     */
+    @Test
+    public void testGetPreciseTimeDurationShouldHandleSmallDecimalValues() {
+        final Map<String, Map<String, Object>> SCENARIOS = Map.of(
+                "decimalNanos", Map.of("originalUnits", TimeUnit.NANOSECONDS, "expectedUnits", TimeUnit.NANOSECONDS, "originalValue", 123.4, "expectedValue", 123.0),
+                "lessThanOneNano", Map.of("originalUnits", TimeUnit.NANOSECONDS, "expectedUnits", TimeUnit.NANOSECONDS, "originalValue", 0.9, "expectedValue", 1.0),
+                "lessThanOneNanoToMillis", Map.of("originalUnits", TimeUnit.NANOSECONDS, "expectedUnits", TimeUnit.MILLISECONDS, "originalValue", 0.9, "expectedValue", 0.0),
+                "decimalMillisToNanos", Map.of("originalUnits", TimeUnit.MILLISECONDS, "expectedUnits", TimeUnit.NANOSECONDS, "originalValue", 123.4, "expectedValue", 123_400_000.0)
```
exceptionfactory, line 189: The nested `Map.of()` calls are a bit difficult to read. Also, it would be helpful to backport this change to the 1.x support branch, where Java 8 is still required. With that in mind, it might make more sense to re-evaluate the test method approach to make it more intuitive, as opposed to using a Map of Maps. Using JUnit 5 parameterized tests would be much better.

nifi-commons/nifi-utils/src/test/java/org/apache/nifi/util/TestFormatUtils.java
```diff
@@ -0,0 +1,368 @@
...
+     */
+    @Test
+    public void testMakeWholeNumberTimeShouldHandleMetricConversions() {
+        final Map<String, Map<String, Object>> SCENARIOS = Map.of(
+                "secondsToMillis", Map.of("originalUnits", TimeUnit.SECONDS, "expectedUnits", TimeUnit.MILLISECONDS, "originalValue", 123.4, "expectedValue", 123_400L),
+                "secondsToMicros", Map.of("originalUnits", TimeUnit.SECONDS, "expectedUnits", TimeUnit.MICROSECONDS, "originalValue", 1.000_345, "expectedValue", 1_000_345L),
+                "millisToMicros", Map.of("originalUnits", TimeUnit.MILLISECONDS, "expectedUnits", TimeUnit.MICROSECONDS, "originalValue", 0.75, "expectedValue", 750L),
+                "nanosToNanosGE1", Map.of("originalUnits", TimeUnit.NANOSECONDS, "expectedUnits", TimeUnit.NANOSECONDS, "originalValue", 123.4, "expectedValue", 123L),
+                "nanosToNanosLE1", Map.of("originalUnits", TimeUnit.NANOSECONDS, "expectedUnits", TimeUnit.NANOSECONDS, "originalValue", 0.123, "expectedValue", 1L)
+        );
```
exceptionfactory, line 227: Changing the approach to use JUnit 5  parameterized tests would be much better.

exceptionfactory (changes requested):
Thanks for working on this test conversion @dan-s1. Unfortunately the original implementation shows its historical roots. With JUnit 5 supporting parameterized tests, this is a good opportunity to refactor the test methods to use that approach instead of nested Maps.

dan-s1:
@exceptionfactory In my latest commit I broke up 3 of the original tests into multiple unit tests as each of them were testing too many things. To the left of the arrow is the original unit test name and to the right are what they were converted to

```
testGetTimeDurationShouldHandleDecimalValues -> testGetTimeDurationWithWholeNumbers, testGetTimeDurationWithDecimalNumbers
testGetPreciseTimeDurationShouldHandleDecimalValues -> testGetPreciseTimeDurationWithWholeNumbers, testGetPreciseTimeDurationWithDecimalNumbers
testShouldGetSmallerTimeUnit -> testGetSmallerTimeUnitWithNull, testGetSmallerTimeUnitWithNanoseconds, testShouldGetSmallerTimeUnit
```

exceptionfactory (approved):
Thanks for refactoring and making the parameterized test changes @dan-s1, this is a helpful example of using JUnit 5! +1 merging

dan-s1:
@exceptionfactory Thank you! What about the other issue I brought up yesterday? Quoting the message:

> I am about to check in the changes as requested but I noticed that there is a test by the same name in a different package nifi-commons/nifi-utils/src/test/java/org/apache/nifi/processor. It is also exercising nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java. I think though my unit test class is in the right package while that one is not. Shouldn't these two test classes be be combined into to one class and placed in the correct package? If so, should I also make this part of this PR? Also it looks like the unit tests in that other class could benefit with the use of JUnit 5 parameterized tests.

exceptionfactory:
Thanks for following up on that question @dan-s1. You're right, I think this new `TestFormatUtils` is in the correct location and the other `TestFormatUtils` should be merged with this one. If you are interested in merging those tests and refactoring them to be parameterized, that would be a helpful follow-on PR.

PR #7313, opened 2023-05-30
NIFI-11614 Improve Validation for JndiJmsConnectionFactoryProvider

exceptionfactory:
# Summary

[NIFI-11614](https://issues.apache.org/jira/browse/NIFI-11614) Improves property validation for the `JndiJmsConnectionFactoryProvider` Controller Service with a set of allowed URL schemes and disallowed Context Factory classes.

The current implementation only checks for a configured value with standard properties, but the new validation narrows the scope to provide better validation prior to enabling the provider.

The initial set of allowed URL schemes is based on the [ActiveMQ Using JMS](https://activemq.apache.org/components/artemis/documentation/latest/using-jms.html) documentation. The new URL validation narrows the scope of potential values, which should be noted in migration guidance, providing a clearer set of supported values

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProperties.java
+++ b/nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProperties.java
@@ -39,8 +39,6 @@ import static org.apache.nifi.processor.util.StandardValidators.NON_EMPTY_VALIDA
 
 public class JndiJmsConnectionFactoryProperties {
 
-    private static final String CONTEXT_FACTORY_DISALLOWED_PROPERTY = "org.apache.nifi.jms.cf.jndi.context.factory.disallowed";
-
     public static final String URL_SCHEMES_ALLOWED_PROPERTY = "org.apache.nifi.jms.cf.jndi.provider.url.schemes.allowed";
 
     public static final PropertyDescriptor JNDI_INITIAL_CONTEXT_FACTORY = new Builder()
@@ -48,7 +46,7 @@ public class JndiJmsConnectionFactoryProperties {
             .displayName("JNDI Initial Context Factory Class")
             .description("The fully qualified class name of the JNDI Initial Context Factory Class. See additional details documentation for disallowed classes.")
             .required(true)
-            .addValidator(new JndiJmsContextFactoryValidator())
+            .addValidator(NON_EMPTY_VALIDATOR)
             .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
             .build();
 
@@ -126,35 +124,7 @@ public class JndiJmsConnectionFactoryProperties {
                 .build();
     }
 
-    private static class JndiJmsContextFactoryValidator implements Validator {
-        private static final String DEFAULT_DISALLOWED_CONTEXT_FACTORY = "LdapCtxFactory";
-
-        private static final String DISALLOWED_CONTEXT_FACTORY;
-
-        static {
-            DISALLOWED_CONTEXT_FACTORY = System.getProperty(CONTEXT_FACTORY_DISALLOWED_PROPERTY, DEFAULT_DISALLOWED_CONTEXT_FACTORY);
-        }
-
-        @Override
-        public ValidationResult validate(final String subject, final String input, final ValidationContext context) {
-            final ValidationResult.Builder builder = new ValidationResult.Builder().subject(subject).input(input);
-
-            if (input == null || input.isEmpty()) {
-                builder.valid(false);
-                builder.explanation("Context Factory is required");
-            } else if (input.endsWith(DISALLOWED_CONTEXT_FACTORY)) {
-                builder.valid(false);
-                builder.explanation(String.format("Context Factory [%s] not allowed", DISALLOWED_CONTEXT_FACTORY));
-            } else {
-                builder.valid(true);
-                builder.explanation("Context Factory allowed");
-            }
-
-            return builder.build();
-        }
-    }
-
-    private static class JndiJmsProviderUrlValidator implements Validator {
+    static class JndiJmsProviderUrlValidator implements Validator {
 
         private static final Pattern URL_SCHEME_PATTERN = Pattern.compile("^([^:]+)://.+$");
 
@@ -173,14 +143,14 @@ public class JndiJmsConnectionFactoryProperties {
                 "vm"
         )));
 
-        private static final Set<String> ALLOWED_SCHEMES;
+        private final Set<String> allowedSchemes;
 
-        static {
+        JndiJmsProviderUrlValidator() {
             final String allowed = System.getProperty(URL_SCHEMES_ALLOWED_PROPERTY);
             if (allowed == null || allowed.isEmpty()) {
-                ALLOWED_SCHEMES = DEFAULT_ALLOWED_SCHEMES;
+                allowedSchemes = DEFAULT_ALLOWED_SCHEMES;
             } else {
-                ALLOWED_SCHEMES = Arrays.stream(allowed.split(SPACE_SEPARATOR)).collect(Collectors.toSet());
+                allowedSchemes = Arrays.stream(allowed.split(SPACE_SEPARATOR)).collect(Collectors.toSet());
             }
         }
 
@@ -196,7 +166,7 @@ public class JndiJmsConnectionFactoryProperties {
                 builder.explanation("URL scheme allowed");
             } else {
                 builder.valid(false);
-                final String explanation = String.format("URL scheme not allowed. Allowed URL schemes include %s", ALLOWED_SCHEMES);
+                final String explanation = String.format("URL scheme not allowed. Allowed URL schemes include %s", allowedSchemes);
                 builder.explanation(explanation);
             }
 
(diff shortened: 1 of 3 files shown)


turcsanyip:
Will review...

nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProperties.java
```diff
@@ -114,4 +119,67 @@ public static PropertyDescriptor getDynamicPropertyDescriptor(final String prope
...
+    }
+
+    private static class JndiJmsProviderUrlValidator implements Validator {
+        /** JNDI JMS URL Allowed Schemes based on ActiveMQ Connection Factory */
+        private static final Set<String> ALLOWED_SCHEMES = Collections.unmodifiableSet(new LinkedHashSet<>(Arrays.asList(
+                "jgroups",
+                "tcp",
+                "udp",
+                "vm"
+        )));
```
turcsanyip, line 151: @exceptionfactory I have concerns that we can enumerate all supported schemes because they are vendor specific. E.g. WebLogic uses `t3(s)://`. Basically, it can be an arbitrary name.

I also understand that `ldap://` is problematic but WebSphere MQ relies on it when using JNDI (or the local `file://` can be used).

Do you have any idea how to handle these shortcomings while keeping the validation logic?

exceptionfactory: Thanks for highlighting these concerns @turcsanyip.

The restrictive approach to URL schemes does present some potential limitations.

I expanded the default list of allowed schemes, and I also introduced a Java System property that supports overriding the default values. I updated the additional details documentation to list the property name and describe how it can be configured. This approach will allow an administrator to customize validation for specific environments and custom drivers while providing default validation for general deployments.

nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProperties.java
```diff
@@ -114,4 +126,108 @@ public static PropertyDescriptor getDynamicPropertyDescriptor(final String prope
...
+        }
+
+        @Override
+        public ValidationResult validate(final String subject, final String input, final ValidationContext context) {
+            final ValidationResult.Builder builder = new ValidationResult.Builder().subject(subject).input(input);
+
+            if (input == null || input.isEmpty()) {
+                builder.valid(false);
+                builder.explanation("Context Factory is required");
+            } else if (input.endsWith(DISALLOWED_CONTEXT_FACTORY)) {
```
turcsanyip, line 145: Do we expect to disallow only one context factory in the future?
If the main target is `LdapCtxFactory`, then we could use a more explicit property like "disallow.ldap" with true/false.
Or we can make it generic via accepting multiple values. I don't think someone would want to disable another context factory but enable ldap at the same time.

nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProperties.java
```diff
@@ -114,4 +126,108 @@ public static PropertyDescriptor getDynamicPropertyDescriptor(final String prope
                 .build();
     }
 
+    private static class JndiJmsContextFactoryValidator implements Validator {
+        private static final String DEFAULT_DISALLOWED_CONTEXT_FACTORY = "LdapCtxFactory";
+
+        private static final String DISALLOWED_CONTEXT_FACTORY;
+
+        static {
+            DISALLOWED_CONTEXT_FACTORY = System.getProperty(CONTEXT_FACTORY_DISALLOWED_PROPERTY, DEFAULT_DISALLOWED_CONTEXT_FACTORY);
```
turcsanyip, line 135: It seems `LdapCtxFactory` can be enabled via the system property only if some dummy value is set. Empty string value should be handled as "allow anything".

nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/test/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProviderTest.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Test
+    void testPropertiesUrlSchemeSystemProperty() {
+        setFactoryProperties();
+
+        runner.setProperty(provider, JndiJmsConnectionFactoryProperties.JNDI_PROVIDER_URL, TEST_PROVIDER_URL);
+
+        runner.assertValid(provider);
+    }
```
turcsanyip, line 110: This test case tests the same as `testPropertiesValid()` because both `tcp` and `test` are set in the system property for all tests. It would be better to apply the system property for this test only (as its name would imply).

turcsanyip (changes requested):
Thanks for the changes @exceptionfactory! The approach of customization via system properties look good.

I added a few improvement comments. The most important one is the single vs. multiple disallowed context factories. Pls. see them below.

nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProperties.java
exceptionfactory: Thanks for highlighting this concern. On further evaluation, after implementing the ability to override the default validation using Java System properties, it seems better to limit the validation to the URL schemes. The latest commit removes this validation.

nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProperties.java
exceptionfactory: Removed this validator as described.

nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/resources/docs/org.apache.nifi.jms.cf.JndiJmsConnectionFactoryProvider/additionalDetails.html
```diff
@@ -91,5 +91,77 @@ <h2>Example:</h2>
...
+</p>
+<p>
+  The following Java System property can be configured to override the default disallowed Context Factory Class:
+</p>
+
+<ul>
+  <li>
+    <code>org.apache.nifi.jms.cf.jndi.context.factory.disallowed</code>
+  </li>
+</ul>
```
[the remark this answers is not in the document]
exceptionfactory: Removed this validator and associated documentation section.

nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/test/java/org/apache/nifi/jms/cf/JndiJmsConnectionFactoryProviderTest.java
exceptionfactory: Thanks for pointing this out. I adjusted the System property evaluation and changed the test approach to support setting and clearing the System property in the context of a single test method.

exceptionfactory:
Thanks for the additional feedback @turcsanyip!

On further consideration, the URL Scheme validation seems sufficient, so I removed the Context Factory class validation and associated documentation. I also adjusted the System property evaluation approach so that the unit tests could evaluate both scenarios.

turcsanyip (approved):
Thanks @exceptionfactory!

+1 LGTM
Merging to main and support/1.x.

schnells:
Thanks for implementing a validator. Because these are very useful.

Unfortunately, this currently prevents the use of salace and nifi. the solace messaging protocol (SMF) uses a url format like `smfs://<solace>:<port>`

Unfortunately the validator fails with `smfs` urls. Is it possible to include smfs in the validator list?
I would also open the necessary pull request. Because currently we cannot use nifi 1.22 in combination with solace.

exceptionfactory:
Thanks for the feedback @schnells. As this pull request is closed, it would be helpful to create an Apache NiFi Jira issue for further discussion.

Adding `smfs` to the default list sounds like a reasonable improvement.

As an alternative for now, the Provider supports a Java System property that can be used to set the allowed URLs.

https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-jms-processors-nar/1.22.0/org.apache.nifi.jms.cf.JndiJmsConnectionFactoryProvider/additionalDetails.html

In this scenario, the following line can be added to `bootstrap.conf` to allow `smfs`:

```
java.arg.jndiJmsUrlSchemesAllowed=-Dorg.apache.nifi.jms.cf.jndi.provider.url.schemes.allowed=smfs
```

schnells:
Thanks @exceptionfactory for the quick response! I will absolutely have a look into the java system property. 
I will create a Jira issue in the near future.

PR #7329, opened 2023-06-02
NIFI-11630 Deprecate ECMAScript Script Engine

exceptionfactory:
# Summary

[NIFI-11630](https://issues.apache.org/jira/browse/NIFI-11630) Deprecates the ECMAScript Script Engine option for scripted components.

The [JEP 335](https://openjdk.org/jeps/335) deprecated the Nashorn engine for Java 11 and Java 17 does not include Nashorn.

The deprecation includes adding DEPRECATED to the Script Engine property display name and logging a deprecation warning when starting a component that uses one of the deprecated Script Engines.

Subsequent efforts can remove the ECMAScript Script Engine option from the main branch while this pull request target the support branch for version 1 releases.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-scripting-bundle/nifi-scripting-processors/pom.xml
+++ b/nifi-nar-bundles/nifi-scripting-bundle/nifi-scripting-processors/pom.xml
@@ -45,6 +45,10 @@
         </dependency>
         <dependency>
             <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-deprecation-log</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-record-serialization-service-api</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


mattyb149:
Should we apply this PR to `main` as well, even if it will be removed before the first 2.0 release, just to make sure they're in sync at this moment? Also should we append to the property's description some generic "this is deprecated and is subject to being removed in a later release" to go along with the `DEPRECATED` addition to the display name?

exceptionfactory:
> Should we apply this PR to `main` as well, even if it will be removed before the first 2.0 release, just to make sure they're in sync at this moment? Also should we append to the property's description some generic "this is deprecated and is subject to being removed in a later release" to go along with the `DEPRECATED` addition to the display name?

Thanks for the feedback @mattyb149. I was thinking that this doesn't need to be applied to the `main` branch, but as a sort of TODO marker, it could be applied to the `main` branch, and then a follow up PR can implement the removal for 2.0.

I am not opposed to adding more details to the description, but as this also includes a warning for the nifi-deprecation.log, I think that will actually be a better pointer to indicate the need for making changes.

mattyb149:
I'm good with it as-is, just wanted to make it known *why* something was being deprecated.

PR #7338, opened 2023-06-05
NIFI-11631 Add OracleDB support for Nifi Registry

dam4rus:
Tested with Oracle Database version 19. Migrating, bucket creation, version controlling flows all worked fine for me. Will probably need a more thorough testing.

# Summary

[NIFI-11631](https://issues.apache.org/jira/browse/NIFI-11631)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [x] JDK 17

### Licensing

### Documentation


mattyb149:
Thanks for this! I tried this at one point ([oracle_registry](https://github.com/mattyb149/nifi/commit/e912b9155cce0b919f0a3d4d9edfc55c7038a65b)) but had trouble setting up integration tests using testcontainers. I think we really need integration tests here like we have with the other implementation(s). Feel free to compare your PR against my aforementioned commit to make sure we are on the same page. I tested mine manually and everything worked fine, but I never got the integration test stuff working.

dam4rus:
Thanks @mattyb149 for bringing your change to my attention. Will definitely check it out. I was not aware that this was already WIP. I was also not aware of existing ITs related to DB in Nifi Registry. I'll try to spin up some tests for OracleDB

mattyb149:
Looking good! Needs to be rebased against the latest main to resolve merge conflicts though

nifi-registry/nifi-registry-core/nifi-registry-framework/pom.xml
```diff
@@ -420,6 +420,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.oracle.database.jdbc</groupId>
+            <artifactId>ojdbc8</artifactId>
+            <version>23.2.0.0</version>
```
exceptionfactory, line 426: This dependency should have the `test` scope.

nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/DatabaseMetadataService.java
```diff
@@ -16,7 +16,21 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import oracle.jdbc.datasource.OracleCommonDataSource;
```
exceptionfactory, line 29: Referencing this specific class is a problem for compatibility with other vendors, so a different approach is necessary to check for Oracle as the version.

nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/DatabaseMetadataService.java
```diff
@@ -391,8 +391,12 @@ public FlowSnapshotEntity getFlowSnapshot(final String flowIdentifier, final Int
 
     @Override
     public FlowSnapshotEntity getLatestSnapshot(final String flowIdentifier) {
-        final String sql = "SELECT * FROM FLOW_SNAPSHOT WHERE flow_id = ? ORDER BY version DESC LIMIT 1";
-
+        final String sql;
+        if (jdbcTemplate.getDataSource() instanceof OracleCommonDataSource) {
```
exceptionfactory, line 395: A different approach is needed to check for the use of Oracle.

exceptionfactory (changes requested):
On a quick review, there is a problem with the explicit reference to the Oracle Data Source class. The Oracle JDBC driver should not be a compilation dependency, and the `instanceof` checks are not suitable to determine Oracle usage at runtime, so a different approach is necessary before this can go forward.

nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/DatabaseMetadataService.java
dam4rus: I was pretty sure this is not the right way to do it and I have seen throughout the code that Flyway's `DatabaseTypeRegister.getDatabaseTypeForConnection(connection)` was used to get the database type. However, opening then closing the connection just to determine the database type, then have `jdbcTemplate` open another connection to execute the query have not felt right either. I guess it's not that big of an issue, especially if the data source is a connection pool, and go with the Flyway approach?

bbende: I think the bigger issue here is we really don't want to be checking database types through out the code and having to issue different SQL statements. So even if we did not directly reference the Oracle class here, I would still be against needing to check DB type.

We have a similar problem with this MSSQL PR:
https://github.com/apache/nifi/pull/7245

If there is absolutely no way to make Oracle and MSSQL confirm to the standard SQL that is already used, then I think a more holistic solution may be needed, possible ideas:
- Consider an ORM, we can't use Hibernate because of licensing, could consider EclipseLink
- Create our own ORM like layer for producing SQL statements where we detect the DB type once during start up and based on that we create some kind SQL statement factory based on DB type.

exceptionfactory: @dam4rus I think we need to take a step back, as we cannot go forward with either Oracle or MSSQL as they stand. This pull request would introduce a direct dependency on Oracle, which is not an option.

It sounds like the best approach is to take a step back as @bbende suggested and consider an abstraction approach, then come back to consider Oracle and MSSQL using that abstraction.

bbende: Yes, sorry I did not mean that this PR should introduce an ORM or abstraction layer. I meant that maybe we should take a step back and consider what options we have for producing DB specific SQL, and try to put that in place before we move forward with Oracle or MSSQL.

mattyb149: Did you have to add the Oracle-specific stuff just for the tests? IIRC my branch was working with Oracle and I just couldn't get the integration tests working. Of course we want tests in there, but if that's the only issue I would think we could get we 3 reviewers to test it manually and go back to adding tests later if/when it's better supported?

dam4rus: @mattyb149 No, they are not only for the tests. These sql queries are executed via some REST API calls and that's why they were eluded during my initial manual testing.

@bbende No, I'm sorry because I've just misinterpreted your comment. Sounds good to me.

@exceptionfactory I've actually got rid of directly depending on Oracle. Even though this PR won't be merged as it is I'll push my recent commit to fix that and be more consistent with #7245 by using Flyway instead.

exceptionfactory:
Thanks again for the work on this feature @dam4rus. Based on the discussion, I created a new Jira issue [NIFI-12059](https://issues.apache.org/jira/browse/NIFI-12059) describing a general need for Registry framework adjustments, along with a potential solution. Once such a solution is in place, then we should be in a better position in implement support for additional database vendors, so I'm closing this pull request for now.

PR #7352, opened 2023-06-07
NIFI-11609 Support Request-Response pattern in MQTT5 Processors

nandorsoma:
# Summary

[NIFI-11609](https://issues.apache.org/jira/browse/NIFI-11609)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/PublishMQTT.java
+++ b/nifi-nar-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/PublishMQTT.java
@@ -28,6 +28,8 @@ import org.apache.nifi.annotation.documentation.Tags;
 import org.apache.nifi.annotation.lifecycle.OnScheduled;
 import org.apache.nifi.annotation.lifecycle.OnStopped;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.ValidationContext;
+import org.apache.nifi.components.ValidationResult;
 import org.apache.nifi.expression.AttributeExpression;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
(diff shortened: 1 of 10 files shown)


nifi-nar-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/PublishMQTT.java
```diff
@@ -328,21 +439,57 @@ public void process(ProcessContext context, FlowFile flowfile, InputStream in, S
...
                     processedRecords.getAndIncrement();
                 }
             } catch (SchemaNotFoundException | MalformedRecordException e) {
                 throw new ProcessException("An error happened during creating components for serialization.", e);
             }
         }
 
+        private String extractRecordValue(Record record, RecordPath recordPath) {
+            final Optional<FieldValue> fv = recordPath.evaluate(record).getSelectedFields().findFirst();
+            if (fv.isPresent()) {
```
nandorsoma, line 473: It is a bit misleading, because FieldValue is present even when there is no result with a given record path. This makes me wonder that is it a valid scenario when fv is empty.

nifi-nar-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/PublishMQTT.java
```diff
@@ -139,11 +202,15 @@ public class PublishMQTT extends AbstractMQTTProcessor {
...
             PROP_TOPIC,
+            PROP_TOPIC_IS_RECORD_PATH,
             PROP_RETAIN,
             PROP_QOS,
             RECORD_READER,
             RECORD_WRITER,
             MESSAGE_DEMARCATOR,
+            PROP_RESPONSE_TOPIC,
+            PROP_CORRELATION_DATA,
+            PROP_CORRELATION_DATA_IS_RECORD_PATH,
```
turcsanyip, line 213: @nandorsoma Thanks for adding this new feature in the MQTT processors!

The functionality looks good, however the configuration of the new properties is not really straightforward. E.g. when sending the response to a request, the response topic (contained in the original request) needs to be set in  `Topic`, not `Response Topic` (`Response Topic` is relevant for request messages, not responses). Also, the new properties may be confusing in the "normal" (non request/response) mode, so in most of the cases.

That's why I would suggest to document the request/response scenarios in the additional details page.
Also, introducing `Messaging Type` (or `Messaging Pattern`) property in PublishMQTT seems useful for me with the following allowable values:
- **Standalone**: default/current logic (no `Correlation Data` or `Response Topic` displayed)
- **Request**: sending a request message and expecting a response to it (`Correlation Data` and `Response Topic` displayed)
- **Response**: sending a response message to a received request (`Correlation Data` is displayed, but `Response Topic` is not)

The role/origin of `Correlation Data` is a bit different in the last 2 cases. For requests, it is the message sender's responsibility to generate a unique correlation data. For responses, the message sender has to echo the received correlation data back. It can be mentioned in the property description or the additional details page.

nandorsoma:
Closing as the proposed solution with the extra fields is inconvenient. We need to find a generic way to specify raw value or record path in the same field.

PR #7353, opened 2023-06-07
NIFI-11658 Streamline using single parameter context for nested PGs

timeabarna:
# Summary

[NIFI-11658](https://issues.apache.org/jira/browse/NIFI-11658)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/nf-process-group-configuration.js
```diff
@@ -513,6 +552,20 @@
 
             // initialize the parameter context combo
             $('#process-group-parameter-context-combo').combo('destroy').combo(comboOptions);
+
+            // initialize parameter context recursive checkbox
+            getAllDescendantProcessGroup(groupId).then(function (response) {
```
mcgilman, line 557: Making this call when the user does not have permissions to the current Process Group, results in an error in the UI. Currently, we should be handling this without a user-facing error because the user may lack permissions for the current PG but have explicit permissions for a Controller Service in the listing.

mcgilman (changes requested):
Thanks for the PR @timeabarna! I've left some comments below. Let me know if you have any questions or if additional background on any of the files in the changeset would be helpful.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/process-group-configuration.jsp
```diff
@@ -109,6 +100,34 @@
             <div id="process-group-controller-services-tab-content" class="configuration-tab">
                 <div id="process-group-controller-services-table" class="settings-table"></div>
             </div>
+            <div id="process-group-parameter-contexts-tab-content" class="configuration-tab">
```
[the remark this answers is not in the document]
timeabarna: Hello @mcgilman ,
Thank you for your review.
We are expecting to add more configuration elements to Parameter Context, moving Parameter Context modifications to a separate tab makes Process Group Configuration window less overloaded.

mtien-apache:
Reviewing.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/process-group-configuration.jsp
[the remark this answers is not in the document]
mcgilman: What additional configuration elements for the current Process Group's referenced Parameter Context are expected?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-client-dto/src/main/java/org/apache/nifi/web/api/entity/ProcessGroupUpdateStrategy.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.web.api.entity;
+
+public enum ProcessGroupUpdateStrategy {
+    UPDATE_PROCESS_GROUP_ONLY,
+    UPDATE_PROCESS_GROUP_WITH_DESCENDANTS
```
exceptionfactory, line 21: Instead of repeating the `UPDATE_PROCESS_GROUP` prefix, recommend changing the naming to be more concise. What about `CURRENT_GROUP` and `CURRENT_GROUP_WITH_CHILDREN`? Much of User Guide documentation mentions "Child Process Groups".

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/NiFiServiceFacade.java
```diff
@@ -1209,9 +1209,10 @@ Set<DocumentedTypeDTO> getControllerServiceTypes(final String serviceType, final
      * Gets all process groups in the specified parent group.
      *
      * @param parentGroupId The id of the parent group
-     * @return process group
+     * @param includeDescendants if process groups from descendant groups should be included
+     * @return List of process groups
      */
-    Set<ProcessGroupEntity> getProcessGroups(String parentGroupId);
+    Set<ProcessGroupEntity> getProcessGroups(String parentGroupId, boolean includeDescendants);
```
exceptionfactory, line 1215: Instead of using a `boolean` here, why not pass the `ProcessGroupUpdateStrategy`?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/ProcessGroupDAO.java
```diff
@@ -59,9 +59,10 @@ public interface ProcessGroupDAO {
      * Gets all of the process groups.
      *
      * @param parentGroupId The parent group id
+     * @param includeDescendants if process groups from descendant groups should be included
      * @return The process groups
      */
-    Set<ProcessGroup> getProcessGroups(String parentGroupId);
+    Set<ProcessGroup> getProcessGroups(String parentGroupId, boolean includeDescendants);
```
exceptionfactory, line 65: See above note on using the `ProcessGroupUpdateStrategy` enum.

exceptionfactory (commented):
Thanks for working on this feature @timeabarna! I noted a couple suggestions about parameter naming and passing around the update strategy.

timeabarna:
Hello @exceptionfactory ,

Thank you ver much for your review, I've updated the PR based on your recommendation.

exceptionfactory (changes requested):
@timeabarna On further review, I noticed several additional implementation issues.

The first issue is that when NiFi is configured with Registry integration, the Add Process Group dialog has a large scroll bar with the Import from Registry link at the end. The sizing changes need to accommodate optional display of the Import from Registry link without resizing or introducing scroll bars.

The second issue relates to uploading a Process Group. An uploaded Process Group may or may not include a Parameter Context, and as a result, the Parameter Context selection field will be confusing. The upload Process Group Resource method does not have the Parameter Context Identifier right now, so from a functionality and usability point of view, it seems like it would be better to disable the Parameter Context selection field, with a message stating something along the lines of `Process Uploaded Parameter Contexts`. With that approach, the Parameter Context field only applies when creating a new Process Group, as opposed to when uploading a Process Group Flow Definition.

timeabarna:
Thanks @exceptionfactory and @mcgilman for your review, I've updated the PR.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/process-group-configuration.jsp
```diff
@@ -40,6 +40,12 @@
                         <div class="setting-name">Process group parameter context</div>
                         <div class="editable setting-field">
                             <div id="process-group-parameter-context-combo"></div>
+                            <div id="parameter-contexts-recursive-container">
+                                <div id="parameter-contexts-recursive" class="nf-checkbox checkbox-unchecked"></div>
+                                <div class="nf-checkbox-label">Apply recursively</div>
```
exceptionfactory, line 45: The positioning of this checkbox seems out of place in comparison to the other form elements. One option could be to place it directly under the `Process Group Parameter Context` select field.

However, another complicating factor is that unlike the other settings, this setting is not persistent. In other words, if a user checks the box and clicks the `Apply` button, the box will be unchecked when returning to the screen. This seems like it could create some confusion for users, especially if the UI gives the impression that Parameter Context selection will always be applied to any nested Process Groups.

With that background, another option could be to move this checkbox field right above, or next to the `Apply` button. That would probably warrant changing the labeling, because then it would appear to apply to all settings, instead of just the Parameter Context.

Yet one more would be to introduce a new tab in the Process Group configuration dialog specific to Parameter Contexts. That would also help differentiate permanent settings from temporary options like this option to apply Parameter Contexts recursively.

exceptionfactory (changes requested):
Thanks for the latest round of updates @timeabarna. The REST Resource implementation looks good, and the adjustments to the Process Group creation dialog also appear functional and clear in terms of Parameter Context selection.

The only remaining concern is the implementation of the `Apply recursively` checkbox on the Process Group configuration screen. I proposed some implementation options, let me know what you think, but it seems like we need an alternative approach to differentiate permanent settings from temporary options.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/process-group-configuration.jsp
timeabarna: Thanks @exceptionfactory for your insights.

The original idea was to have a dedicated tab for Parameter Context however after a discussion this was dropped as the current UI implementation is not applicable for this. I think the main concern is that having Apply buttons at two different contexts (Process Group Configuration and Parameter Context) require some extra validations and checks which would probably introduce some regressions.

The agreement wast to add the Apply recursive button next to the Parameter Context selection field and as there are some more new additions to the Process Group Configuration screen, such as Per Process Group Logging and some stateless related configurations, have a separate PR for rearranging the screen.

The new screen planed to be similar to the Parameter Provider screen with multiple columns where configuration fields grouped together based on some logic which needs to be discussed as part of a different activity.

@mcgilman , @mtien-apache and @sardell please feel free to chime in and share your thoughts.

timeabarna:
Bumping to prevent stale status

exceptionfactory (approved):
Thanks for your work on this feature @timeabarna, and thanks for the review and help on the development @mcgilman.

Following an offline discussion, it sounds like the best path forward is to merge this pull request into the main branch as part of the work for NiFi 2.0. With this feature in place, additional work can be done to refine the Process Group configuration screen before we release the final version.

With that background, I will plan on merging this soon.

PR #7357, opened 2023-06-08
NIFI-11663: Bump flyway-core version to 9.19.3

r-sidd:
# Summary

[NIFI-11663](https://issues.apache.org/jira/browse/NIFI-11663)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/pom.xml
+++ b/nifi-registry/pom.xml
@@ -37,7 +37,7 @@
     <properties>
         <jax.rs.api.version>2.1</jax.rs.api.version>
         <spring.boot.version>2.7.12</spring.boot.version>
-        <flyway.version>8.5.13</flyway.version>
+        <flyway.version>9.19.3</flyway.version>
         <flyway.tests.version>7.0.0</flyway.tests.version>
         <swagger.ui.version>3.12.0</swagger.ui.version>
         <groovy.eclipse.compiler.version>3.7.0</groovy.eclipse.compiler.version>


pvillard31:
I'd be very careful about this. Upgrading flyway to another major release could impact the range of database versions we support for NiFi Registry and how we deal with the internal H2 database. @bbende - do you have thoughts on this?

exceptionfactory:
> I'd be very careful about this. Upgrading flyway to another major release could impact the range of database versions we support for NiFi Registry and how we deal with the internal H2 database. @bbende - do you have thoughts on this?

I agree, this is a major version upgrade, and the Flyway change log highlights several database version adjustments. It does add support for more recent versions in some cases. This needs to be evaluated manually with the Registry integration tests to ensure those work as expected.

bbende:
Yea this is tricky. I wonder if we should consider migrating to Liquibase for 2.0.0 since it doesn't seem to be as restrictive with dropping versions, but I'm not exactly sure how an upgrade would work where there is an existing DB that Liquibase needs to take control, although I think you can setup the migrations to skip stuff that is already there.

exceptionfactory:
[Flyway 9.0.0](https://flywaydb.org/documentation/learnmore/releaseNotes#9.0.0) dropped support for MariaDB 10.2, which results in failing database integration for NiFi Registry.

[MariaDB 10.2.44](https://mariadb.com/kb/en/mariadb-10244-release-notes/) was the last version planned for the version 10.2 series, so that version is officially end of life. From that perspective, we should consider raising the minimum version of MariaDB to a more recent version, which could allow this upgrade to go forward.

However, until we update the minimum supported MariaDB versions and associated tests, the upgrade to Flyway 9 cannot go forward.

The two vulnerabilities mentioned in the NiFi Jira issue impact PostgreSQL and do not apply to Flyway, so this is not a direct security concern.

A separate NiFi Jira issue should be used to review supported minimum database versions, and then the Flyway upgrade can be re-evaluated.

PR #7370, opened 2023-06-12
NIFI-11680 Correct Buffer Size Calculation for Connection Balancing

exceptionfactory:
# Summary

[NIFI-11680](https://issues.apache.org/jira/browse/NIFI-11680) Corrects `PeerChannel` methods used for Cluster Connection Load Balancing to avoid potential `BufferOverflowException` errors with the Bouncy Castle Security Provider for TLS.

The Bouncy Castle [ProvSSLSessionBase](https://github.com/bcgit/bc-java/blob/master/tls/src/main/java/org/bouncycastle/jsse/provider/ProvSSLSessionBase.java) uses a more coarse-grained calculation for Application Buffer Size and Packet Buffer Size than the standard JDK [SSLSessionImpl](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/sun/security/ssl/SSLSessionImpl.java). As a result of this difference in behavior, the `PeerChannel.encrypt()` method can expand the encrypted destination buffer greater than the SSLSession Application Buffer Size. The current approach works with the standard JDK implementation, but results in failures with files greater than 500 KB when using the Bouncy Castle Provider.

The revised approach calculates the expanded size of the intermediate prepared buffer based on the size of the destination buffer, as opposed to using the SSLSession Application Buffer Size.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/queue/clustered/client/async/nio/PeerChannel.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/queue/clustered/client/async/nio/PeerChannel.java
@@ -162,12 +162,13 @@ public class PeerChannel implements Closeable {
         while (plaintext.hasRemaining()) {
             encrypt(plaintext);
 
-            final int bytesRemaining = prepared.capacity() - prepared.position();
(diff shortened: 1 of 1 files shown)


markap14:
Thanks @exceptionfactory changes look good to me. Was able to do some testing and all appears to work as expected. +1 will merge.

PR #7380, opened 2023-06-14
NIFI-11691: Support VARBINARY and LONGVARBINARY types in PutDatabaseRecord

mattyb149:
# Summary

[NIFI-11691](https://issues.apache.org/jira/browse/NIFI-11691) This PR adds support for the java.sql.Types VARBINARY and LONGVARBINARY in PutDatabaseRecord. It converts the record field value to `byte[]` and then calls `setBytes()` when setting the parameter value in the PreparedStatement

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 11
  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
@@ -886,13 +886,15 @@ public class PutDatabaseRecord extends AbstractProcessor {
                 try {
                     ps.setBytes(index, byteArray);
                 } catch (SQLException e) {
-                    throw new IOException("Unable to parse binary data " + value, e.getCause());
+                    throw new IOException("Unable to parse binary data with size" + byteArray.length, e.getCause());
                 }
             } else {
+                byte[] byteArray = new byte[0];
                 try {
-                    ps.setBytes(index, value.toString().getBytes(StandardCharsets.UTF_8));
+                    byteArray = value.toString().getBytes(StandardCharsets.UTF_8);
+                    ps.setBytes(index, byteArray);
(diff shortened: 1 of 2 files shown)


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
```diff
@@ -868,6 +868,33 @@ private void setParameter(PreparedStatement ps, int index, Object value, int fie
...
+                }
+                byte[] byteArray = (byte[]) value;
+                try {
+                    ps.setBytes(index, byteArray);
+                } catch (SQLException e) {
+                    throw new IOException("Unable to parse binary data " + value, e.getCause());
+                }
+            } else {
+                try {
+                    ps.setBytes(index, value.toString().getBytes(StandardCharsets.UTF_8));
```
exceptionfactory, line 893: It might be worth adding a comment on this line, since `value.toString()` may not provide an expected representation of the underlying object type. Perhaps there should be some additional type checking in this conditional to determine if `value` is an array type or collection type, otherwise it seems like this could produce unexpected results.

exceptionfactory (commented):
Thanks for addressing this issue @mattyb149, the general approach looks good, but I noted a couple questions about the alternative type handling.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
mattyb149: This is just in case the data came in as a String

exceptionfactory (approved):
Thanks for making the adjustments @mattyb149! +1 merging

PR #7422, opened 2023-06-22
NIFI-11740: Added SnowflakeConfigurationService

turcsanyip:
# Summary

[NIFI-11740](https://issues.apache.org/jira/browse/NIFI-11740)

This PR adds SnowflakeConfigurationService that can be used by multiple Snowflake processors and controller services.
The solution is optimised for NiFi 2.0 where the legacy properties can be removed from StandardSnowflakeIngestManagerProviderService and only "Snowflake Configuration Service" would survive.
For the same reason, some of the properties and code sections are duplicated. I don't think it may worth extracting them in common classes because they will simply be removed from StandardSnowflakeIngestManagerProviderService in 2.0.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/util/SnowflakeConstants.java
+++ b/nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/util/SnowflakeConstants.java
@@ -14,24 +14,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.nifi.snowflake.service.util;
 
-public final class ConnectionUrlFormatParameters extends SnowflakeCommonParameters {
+public final class SnowflakeConstants {
 
-    private final String jdbcUrl;
+    public static final String SNOWFLAKE_HOST_SUFFIX = ".snowflakecomputing.com";
 
-    public ConnectionUrlFormatParameters(final String jdbcUrl,
-            final String organizationName,
-            final String accountName,
-            final String accountLocator,
-            final String cloudRegion,
-            final String cloudType) {
-        super(organizationName, accountName, accountLocator, cloudRegion, cloudType);
-        this.jdbcUrl = jdbcUrl;
-    }
+    public static final String DEFAULT_SCHEME = "https";
 
-    public String getJdbcUrl() {
-        return jdbcUrl;
+    public static final int DEFAULT_PORT = 443;
+
+    private SnowflakeConstants() {
     }
 }
--- a/nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-processors/src/main/java/org/apache/nifi/processors/snowflake/PutSnowflakeInternalStage.java
+++ b/nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-processors/src/main/java/org/apache/nifi/processors/snowflake/PutSnowflakeInternalStage.java
@@ -77,11 +77,13 @@ public class PutSnowflakeInternalStage extends AbstractProcessor {
 
     public static final PropertyDescriptor DATABASE = new PropertyDescriptor.Builder()
             .fromPropertyDescriptor(SnowflakeProperties.DATABASE)
+            .description("The database to use. If not specified, the default database will be used if it is configured for the connection or the user.")
             .dependsOn(INTERNAL_STAGE_TYPE, SnowflakeInternalStageType.NAMED, SnowflakeInternalStageType.TABLE)
             .build();
 
     public static final PropertyDescriptor SCHEMA = new PropertyDescriptor.Builder()
             .fromPropertyDescriptor(SnowflakeProperties.SCHEMA)
+            .description("The schema to use. If not specified, the default schema will be used if it is configured for the connection.")
             .dependsOn(INTERNAL_STAGE_TYPE, SnowflakeInternalStageType.NAMED, SnowflakeInternalStageType.TABLE)
             .build();
 
--- a/nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services-api/src/main/java/org/apache/nifi/processors/snowflake/SnowflakeConfiguration.java
+++ b/nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services-api/src/main/java/org/apache/nifi/processors/snowflake/SnowflakeConfiguration.java
@@ -14,24 +14,43 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+package org.apache.nifi.processors.snowflake;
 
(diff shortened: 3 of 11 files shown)


nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/util/SnowflakeConstants.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.snowflake.service.util;
+
+public final class SnowflakeConstants {
```
exceptionfactory, line 19: I recommend using an `enum` opposed to a class with static member variables for constant values.

nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
```diff
@@ -161,17 +188,44 @@ public void onEnabled(final ConfigurationContext context) throws InitializationE
...
+            final AccountIdentifierFormatParameters parameters = getAccountIdentifierFormatParameters(context);
+            account = accountIdentifierFormat.getAccount(parameters);
+            accountUrl = accountIdentifierFormat.getAccountUrl(parameters);
+        }
+
+        URL url = null;
+        try {
+            url = new URL(accountUrl);
+        } catch (MalformedURLException e) {
+            // url parse failure => handle accountUrl as hostname only
```
exceptionfactory, line 220: Perhaps this should be logged at the debug level for potential troubleshooting?

exceptionfactory (commented):
Thanks for the contribution @turcsanyip, I added a few comments, the approach looks good in general, but I will defer further feedback so that perhaps another reviewer with more Snowflake experience can evaluate the details.

nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
```diff
@@ -196,7 +250,7 @@ public SimpleIngestManager getIngestManager() {
     }
 
     private AccountIdentifierFormatParameters getAccountIdentifierFormatParameters(ConfigurationContext context) {
-        final String hostUrl = context.getProperty(HOST_URL)
+        final String accountUrl = context.getProperty(ACCOUNT_URL)
```
nandorsoma, line 253: This code block is duplicated in `StandardSnowflakeConfigurationService` and might worth extracting it.

nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/util/AccountIdentifierFormatParameters.java
```diff
@@ -19,19 +19,19 @@
 
 public final class AccountIdentifierFormatParameters extends SnowflakeCommonParameters {
 
-    private final String hostUrl;
+    private final String accountUrl;
 
-    public AccountIdentifierFormatParameters(final String hostUrl,
+    public AccountIdentifierFormatParameters(final String accountUrl,
```
nandorsoma, line 24: When I tried to understand the code, the parts related to this class were hard to read. In `AccountIdentifierFormatTest`, I already pointed out that it is a bit weird that 5/6 parameters of this class can be null. I think the root issue is that this class is an unnecessary intermediate step, at least in this form. We need to parse the properties from the context into this POJO, then in the `AccountIdentifierFormat`, we need to declare again which properties we need. The clean way would be to create `*FormatParameter` POJOs for each `AccountIdentifierFormat`. But this could be an overkill. The other option would be removing this class and extracting the required parameters from the context in the `AccountIdentifierFormat` class. Either option is good for me; the latter is probably better because the extra complexity wouldn't bring additional value.

nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/util/ConnectionUrlFormatParameters.java
```diff
@@ -19,19 +19,19 @@
 
 public final class ConnectionUrlFormatParameters extends SnowflakeCommonParameters {
 
-    private final String snowflakeUrl;
+    private final String jdbcUrl;
 
-    public ConnectionUrlFormatParameters(final String snowflakeUrl,
+    public ConnectionUrlFormatParameters(final String jdbcUrl,
```
nandorsoma, line 24: I have the same feeling as at `AccountIdentifierFormat`.

nifi-nar-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/util/SnowflakeConstants.java
exceptionfactory: That's a good point about the multiple types, in that case, `DEFAULT_PORT` would need to return the String `443`.

Either way, I recommend renaming this to something like `SnowflakePropertyValue`. On the other hand, as these values all relate to URI construction, it may be better to refactor usage to something more focused as opposed to creating an unconstrained location for global variables.

exceptionfactory (commented):
@turcsanyip On further review of this PR, can you provide some additional background as to the value added?

If the purpose of the service is simply shared configuration properties, it seems like that could already be accomplished using a Parameter Context. It might be different if there were different a number of different authentication methods for Snowflake services, analogous to the cloud providers, but the options appear more limited. It seems like this adds a layer of indirection for configuration, as opposed to making use of a shared Parameter Context. The Private Key Service can already be shared between components.

Perhaps there are other features this service could provide that might not be readily apparent on a general review?

exceptionfactory (changes requested):
@turcsanyip Given the lack of momentum on this pull request, and the broader concern over whether it makes the most sense from a design perspective, I am going to close the pull request for now. We can discuss it further on the Jira issue, or feel free to open it again when you are ready. Thanks!

PR #7423, opened 2023-06-22
NIFI-11741: Add guava lib to nifi-ranger-resource project, in order t…

taz1988:
…o include it into the ext/ranger/install/lib/ directory during assembly the artifact

# Summary

[NIFI-11741](https://issues.apache.org/jira/browse/NIFI-11741)

Fix No class def found error. It happens if nifi build with include-ranger option and try to use credentialsbuilder

# Tracking

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-11741) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

After the fix, if I executed ```java -cp "nifi-assembly/target/nifi-2.0.0-SNAPSHOT-bin/nifi-2.0.0-SNAPSHOT/ext/ranger/install/lib/*" org.apache.ranger.credentialapi.buildks create sslTrustStore -value "test" -provider "jceks://file/<<absolute_path>>/test2.jceks"``` the no class def found error doesn't happened anymore.

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-assembly/src/main/assembly/ranger.xml
+++ b/nifi-assembly/src/main/assembly/ranger.xml
@@ -58,6 +58,7 @@
                 <include>org.slf4j:slf4j-api</include>
                 <include>org.slf4j:jcl-over-slf4j</include>
                 <include>org.apache.commons:commons-lang3</include>
+                <include>com.google.guava:guava</include>
             </includes>
         </dependencySet>
         <!-- Write out scripts from nifi-ranger-resources to ext/ranger/scripts -->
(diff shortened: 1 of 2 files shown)


taz1988:
I tried to investigate what is changed (The ranger version is not changed between 1.21.0 and 1.22). Try to spend more time tomorrow to figure out what is happening

exceptionfactory (commented):
@taz1988 After closer evaluation, with Guava already being a transitive dependency of the Ranger `credentialsbuilder` library, the better solution was to add `com.google.guava:guava` to the dependency include set for the `ext/ranger/install/lib` directory. This places the Guava dependency in that directory as expected.

Will plan on merging pending a successful build.

exceptionfactory (approved):

taz1988:
@exceptionfactory thanks for doing the change! I thinked about that solution too. In the meantime I was able to track it down, that the problem exist since guava version update - https://issues.apache.org/jira/browse/NIFI-11645 If you build code before that everything is good, but after it it is happen this missing jar error. Not 100% sure why this cause that issue

PR #7431, opened 2023-06-26
NIFI-11743 Adding restrictions on return values of FlowRegistryClient

simonbence:
# Summary

[NIFI-11743](https://issues.apache.org/jira/browse/NIFI-11743)

The current definition of FlowRegistryClient allows implementations to return with a null value in cases such as list of buckets, flows or versions. In practice, this will not happen, but future implementations of changes has this option now.

This is something we want to eliminate both because it is considered as a bad pattern and because NiFi expect a valid (but maybe empty) return value. From code perspective this will not cause any change, but the document will be extended accordingly and will communicate that null is not a valid return value.

Even if this is not a code change and the current registry client behaves in this manner, it is considered as a breaking change thus this will be part of the 2.x line only.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation


exceptionfactory (approved):
Thanks for the background and updates @simonbence, this is a helpful clarification going forward. +1 merging

PR #7440, opened 2023-06-26
NIFI-11758: Added local file upload option in PutAzure*Storage proces…

turcsanyip:
…sors

# Summary

[NIFI-11758](https://issues.apache.org/jira/browse/NIFI-11758)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/src/main/java/org/apache/nifi/processors/dataupload/DataUploadSource.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/src/main/java/org/apache/nifi/processors/dataupload/DataUploadSource.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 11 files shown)


exceptionfactory (commented):
Thanks for the contribution @turcsanyip.

Although the Jira issue provides some background on potential use cases, this raises some larger questions about integration patterns for processing local files. The Local File use case presumes some other process is responsible for landing files in a particular location.

One important note, the Local File strategy has security implications, and should require applying the `READ_FILESYSTEM` permission to the associated Processors. That also impacts current behavior.

nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/src/main/java/org/apache/nifi/processors/dataupload/DataUploadSource.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.dataupload;
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum DataUploadSource implements DescribedValue {
+
+    FLOWFILE_CONTENT("FlowFile's Content", "The content of the incoming FlowFile will be uploaded."),
```
exceptionfactory, line 23: Recommend avoiding the possessive and just naming this `FlowFile Content` if this goes forward.

exceptionfactory (changes requested):
@turcsanyip On further consider, the concern regarding filesystem permissions seems to warrant a different approach. Assigning new permissions to these Processors would impact existing deployments, potentially restricting access more than necessary for standard use cases.

As an alternative approach, a new Controller Service interface could be introduced that provides File System access. The general contract could return an `InputStream` from a `java.net.URI` object. The implementation could easily return a `FileInputStream` for a URI-derived Path, and the buffered upload approach could work. This could also support a similar approach for other Processors. The new interface could be named something along the lines of `FileResourceProvider` or perhaps `StreamResourceProvider`, depending on how generic this should be scoped.

The advantage of the new Controller Service interface is that the implementation can be annotated with the Read Filesystem Permission. This avoids impacting existing flows, and makes it clear that this approach requires additional permissions.

If this sounds like a good approach, it may be helpful to close the current PR and open a new one along the lines of the proposed implementation.

turcsanyip:
Thanks @exceptionfactory for sharing your ideas! Closing this PR in favour of a new one based on the suggested concept.

PR #7441, opened 2023-06-26
NIFI-11480 add option to group PutElasticsearchRecord errors by Elasticsearch _bulk error type

ChrisSamo632:
# Summary

[NIFI-11480](https://issues.apache.org/jira/browse/NIFI-11480) - add option to group PutElasticsearchRecord errors by Elasticsearch _bulk error type

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecord.java
+++ b/nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecord.java
@@ -36,6 +36,7 @@ import org.apache.nifi.elasticsearch.IndexOperationRequest;
 import org.apache.nifi.elasticsearch.IndexOperationResponse;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
+import org.apache.nifi.logging.ComponentLog;
 import org.apache.nifi.processor.ProcessContext;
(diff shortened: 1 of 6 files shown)


mattyb149:
How do you test different errors in a NiFi instance (not a unit test)? Given the Operation I can only generate one kind of error at a time. This looks good though, just want to verify I can get groups of errors then I'll merge

ChrisSamo632:
> How do you test different errors in a NiFi instance (not a unit test)? Given the Operation I can only generate one kind of error at a time. This looks good though, just want to verify I can get groups of errors then I'll merge

@mattyb149 my suggestion (although I'll not be in a position to try it for a while), would be to create an index with a numeric field mapping, index a document with a known id, then run the processor in update mode with 2 records - 1 with a mismatched id (should result in a not_found error), the 2nd with the correct id but a non-numeric value for the numeric field (should give a "format exception")

From memory, this is the kind of thing I've tried to show in the unit/integration tests

This could possibly all be done with a single FlowFile and run off the processor - the index operation can be set by the value of a field within a record. So something like:

1. Op=index, id=1, fields: foo="bar", baz=1
2. Op=update, id=unknown, fields: foo="zap"
3. Op=update, id=1, fields: baz="not a number"

Index=bulk_test, Index Op record path=/Op, id record path=/id

That's a chance Elasticsearch will set the "baz" fields mapping to string by default, but hopefully it'll choose Integer. If so, we'll need to separately define the index mappings and pre-create the index with an InvokeHTTP call (for example). If you don't get to it first, I'll try to put a Flow Definition together and attach it to the jira ticket as an example

mattyb149:
+1 LGTM, thanks for the improvement and the reproduction flow! Merging to support/nifi-1.x and main

PR #7446, opened 2023-06-28
NIFI-11536: implemented AbstractStoreScanner and KeyStoreScanner that…

emiliosetiadarma:
… reloads SSL context, changed TrustStoreScanner to extend the AbstractStoreScanner, and implemented unit tests

# Summary

[NIFI-11536](https://issues.apache.org/jira/browse/NIFI-11536)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/util/AbstractStoreScanner.java
```diff
@@ -0,0 +1,175 @@
...
+        this._scanner.setScanInterval(scanInterval);
+    }
+
+    protected SslContextFactory getSslContextFactory() {
+        return sslContextFactory;
+    }
+
+    protected abstract String getFileType();
+    protected abstract Resource getResource();
+    protected abstract Logger getLogger();
```
exceptionfactory, line 166: Instead of making `getLogger()` abstract, the Logger can be defined as an instance variable with `getClass()` so that concrete classes will have distinct loggers.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/util/AbstractStoreScanner.java
```diff
@@ -0,0 +1,175 @@
...
+
+    public void setScanInterval(int scanInterval) {
+        this._scanner.setScanInterval(scanInterval);
+    }
+
+    protected SslContextFactory getSslContextFactory() {
+        return sslContextFactory;
+    }
+
+    protected abstract String getFileType();
```
exceptionfactory, line 164: Recommend using a constructor argument to provide the File Type, instead of an abstract method.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/util/AbstractStoreScanner.java
```diff
@@ -0,0 +1,175 @@
...
+    protected abstract Resource getResource();
```
exceptionfactory, line 165: The `Resource` could also be provided as a constructor argument, eliminating the need for abstract methods. With those changes, the same class can be used without the need for subclasses.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/util/AbstractStoreScanner.java
```diff
@@ -0,0 +1,175 @@
...
+import java.io.IOException;
+import java.util.Collections;
+
+import static org.apache.nifi.security.util.SslContextFactory.createSslContext;
+
+public abstract class AbstractStoreScanner extends ContainerLifeCycle implements Scanner.DiscreteListener {
+    private final SslContextFactory sslContextFactory;
+    private final TlsConfiguration tlsConfiguration;
+    private final File file;
+    private final Scanner _scanner;
```
exceptionfactory, line 41: Although Jetty uses the `_scanner` name, recommend removing the `_` character to follow Apache NiFi project conventions.

exceptionfactory (changes requested):
Thanks for addressing the issue @emiliosetiadarma! The general implementation approach looks good.

Reviewing the differences between the Scanner implementations, it looks like it comes down to configuration. Recommend making one concrete class named `StoreScanner` and using constructor arguments to provide the appropriate settings.

emiliosetiadarma:
Thanks @exceptionfactory for the quick review! Will address the comments

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/util/AbstractStoreScanner.java
emiliosetiadarma: Got it. Will change it to use `getClass()`

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/util/AbstractStoreScanner.java
emiliosetiadarma: Made the constructor argument take in a fileType argument

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/util/AbstractStoreScanner.java
emiliosetiadarma: Making the changes

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/util/AbstractStoreScanner.java
emiliosetiadarma: Making the changes

emiliosetiadarma:
I made the changes while still preserving `KeyStoreScanner` and `TrustStoreScanner` as the derived classes for two reasons. 

1. So the logger is class-specific
2. So the `server` would have beans with distinct types. Having all have the same type would cause some existing tests to fail since it uses `server.getBean`.

Some thoughts: I was wondering if we were to have just one class whether it's a good idea to have both a keystore scanner and truststore scanner in the same class. One advantage this would have is if both keystore and truststore changed, then the `SSLContext` would be reloaded only once, as opposed to having two scanners that reloaded it twice.

exceptionfactory:
> I made the changes while still preserving `KeyStoreScanner` and `TrustStoreScanner` as the derived classes for two reasons.
> 
>     1. So the logger is class-specific

It is not necessary to define a class-specific logger, although maintaining two separate class implementations might be useful based on the second concern.

> 
>     2. So the `server` would have beans with distinct types. Having all have the same type would cause some existing tests to fail since it uses `server.getBean`.

That is a good point regarding `addBean` and `getBean`. With that in mind, it does seem reasonable to maintain two separate classes.

> 
> 
> Some thoughts: I was wondering if we were to have just one class whether it's a good idea to have both a keystore scanner and truststore scanner in the same class. One advantage this would have is if both keystore and truststore changed, then the `SSLContext` would be reloaded only once, as opposed to having two scanners that reloaded it twice.

Although having one class might be worth considering, in general the keystore and truststore should change independently, so keeping two separate classes sounds good.

Thanks for considering the options.

exceptionfactory (changes requested):
Thanks for the updates @emiliosetiadarma, this looks closer to completion. Can you consolidate the two test classes into a single `StoreScannerTest` now that there is a single implementation class?

exceptionfactory (approved):
Thanks for working through the feedback @emiliosetiadarma, the latest version works as expected in success and failure conditions. +1 merging

PR #7448, opened 2023-06-29
NIFI-11761 Fixed MiNiFi restart issue when graceful shutdown period expires. MiNiFi restart sends bootstrap to background

briansolo1985:
# Summary

[NIFI-11761](https://issues.apache.org/jira/browse/NIFI-11761)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/command/StopRunner.java
+++ b/minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/command/StopRunner.java
@@ -45,7 +45,7 @@ public class StopRunner implements CommandRunner {
     private final ProcessUtils processUtils;
 
     public StopRunner(BootstrapFileProvider bootstrapFileProvider, MiNiFiParameters miNiFiParameters, MiNiFiCommandSender miNiFiCommandSender,
-        CurrentPortProvider currentPortProvider, GracefulShutdownParameterProvider gracefulShutdownParameterProvider, ProcessUtils processUtils) {
+                      CurrentPortProvider currentPortProvider, GracefulShutdownParameterProvider gracefulShutdownParameterProvider, ProcessUtils processUtils) {
         this.bootstrapFileProvider = bootstrapFileProvider;
         this.miNiFiParameters = miNiFiParameters;
         this.miNiFiCommandSender = miNiFiCommandSender;
@@ -56,6 +56,7 @@ public class StopRunner implements CommandRunner {
 
     /**
      * Shutdown the MiNiFi and the managing bootstrap process as well.
+     *
      * @param args the input arguments
      * @return status code
      */
@@ -94,6 +95,7 @@ public class StopRunner implements CommandRunner {
                 status = ERROR.getStatusCode();
             }
         } catch (IOException e) {
+            CMD_LOGGER.warn("An error has occurred while stopping MiNiFi. Force killing process with pid=" + minifiPid, e);
             killProcessTree(minifiPid);
         } finally {
             if (lockFile.exists() && !lockFile.delete()) {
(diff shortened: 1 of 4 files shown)


minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/command/StopRunner.java
```diff
@@ -112,6 +114,19 @@ private void gracefulShutDownMiNiFiProcess(long minifiPid) throws IOException {
         if (minifiPid != UNINITIALIZED) {
             processUtils.shutdownProcess(minifiPid, "MiNiFi has not finished shutting down after {} seconds. Killing process.",
                 gracefulShutdownParameterProvider.getGracefulShutdownSeconds());
+            int maxRetry = 5;
+            while (processUtils.isProcessRunning(minifiPid)) {
```
ferencerdei, line 118: What do you think about putting this into the processUtils' killProcessTree method? it is called from multiple places (from the stopRunner directly, and from the shutdownProcess method as well)

minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/command/StopRunner.java
```diff
@@ -112,6 +114,19 @@ private void gracefulShutDownMiNiFiProcess(long minifiPid) throws IOException {
...
+                if (maxRetry == 0) {
+                    throw new IOException("Failed to stop MiNiFi process. MiNiFi process is still running after graceful shutdown has completed");
```
bejancsaba, line 120: We could explicitly say here "...after graceful shutdown completed and killing attempted afterwards" or something along those lines I think we can't get here only if graceful time expired kill was attempted but still the process didn't stop. Right?

minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/command/StopRunner.java
```diff
@@ -112,6 +114,19 @@ private void gracefulShutDownMiNiFiProcess(long minifiPid) throws IOException {
...
+                }
+                CMD_LOGGER.debug("MiNiFi process is still running after shutdown has completed");
```
bejancsaba, line 122: I think this could be WARN as if we are here that is already not good

minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/command/StopRunner.java
```diff
@@ -112,6 +114,19 @@ private void gracefulShutDownMiNiFiProcess(long minifiPid) throws IOException {
         if (minifiPid != UNINITIALIZED) {
             processUtils.shutdownProcess(minifiPid, "MiNiFi has not finished shutting down after {} seconds. Killing process.",
                 gracefulShutdownParameterProvider.getGracefulShutdownSeconds());
+            int maxRetry = 5;
```
bejancsaba, line 117: I'm not sure whether it would be useful or not to externalise at least this one (maybe the sleep time is not justified but it can be argued that this makes sense to be configurable) what do you think?

bejancsaba (commented):
Thanks for the change had only minor comments please take a look.

briansolo1985:
Thanks for your comments. I tried addressed all of them. I had to change the log message differently because with moving the related code to UnixProcessUtils the context has changed as well.

ferencerdei:
Thanks for the update +1 from my side

PR #7449, opened 2023-06-29
NIFI-11334: PutIceberg processor instance interference due same class loader usage

mark-bathori:
# Summary

[NIFI-11334](https://issues.apache.org/jira/browse/NIFI-11334)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/AbstractIcebergProcessor.java
```diff
@@ -117,6 +122,15 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
         }
     }
 
+    @Override
+    public String getClassloaderIsolationKey(PropertyContext context) {
+        final KerberosUserService kerberosUserService = context.getProperty(KERBEROS_USER_SERVICE).asControllerService(KerberosUserService.class);
+        if (kerberosUserService != null) {
+            return kerberosUserService.getIdentifier();
```
turcsanyip, line 129: I would suggest using the same classloader isolation key as the one in the other hadoop related components: kerberos principal.
```
                final KerberosUser kerberosUser = kerberosUserService.createKerberosUser();
                return kerberosUser.getPrincipal();
```
The controller service identifier also works but it may be too restrictive. It creates separate classloaders for controller service having the same principal but they could share the classloader. The user can create a process group with the kerberos service and the iceberg processor in it and then copy it multiple times. Not the best design because the kerberos service should be extracted in the parent process group in this case but I can imagine it happening.

turcsanyip (changes requested):
Thanks @mark-bathori for fixing this bug in the end!
I tested it with an unsecured and a secured Hive and it works as expected.
Added some comments regarding naming and code design but the functionality is already correct.

nifi-nar-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/catalog/IcebergCatalogFactory.java
```diff
@@ -0,0 +1,82 @@
...
+public class IcebergCatalogFactory {
+
+    private final IcebergCatalogService catalogService;
+
+    public IcebergCatalogFactory(IcebergCatalogService catalogService) {
+        this.catalogService = catalogService;
+    }
+
+    public Catalog create() {
+        return switch (catalogService.getCatalogServiceType()) {
```
krisztina-zsihovszki, line 42: Using arrow operator is not supported in Java 8, since this processor is also targeted for 1.x, please use Java 8 compatible syntax.

nifi-nar-bundles/nifi-iceberg-bundle/nifi-iceberg-services/src/main/java/org/apache/nifi/services/iceberg/HiveCatalogService.java
```diff
@@ -69,28 +65,43 @@ protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
...
-    private HiveCatalog catalog;
-
     @Override
     protected Collection<ValidationResult> customValidate(ValidationContext validationContext) {
 
         final List<ValidationResult> problems = new ArrayList<>();
-        String configMetastoreUri = null;
-        String configWarehouseLocation = null;
+        boolean configMetastoreUriPresent = false;
+        boolean configWarehouseLocationPresent = false;
```
krisztina-zsihovszki, line 73: Not related to the actual change just noticed now that the expression language support related config of the properties can be enchanced.

In my view both METASTORE_URI and WAREHOUSE_LOCATION can use

    .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
    
  since flow files are not used when they are evaluated.
  (The same comment applies for WAREHOUSE_PATH in HadoopCatalogService.)

nifi-nar-bundles/nifi-iceberg-bundle/nifi-iceberg-services/src/main/java/org/apache/nifi/services/iceberg/AbstractCatalogService.java
```diff
@@ -44,24 +55,30 @@ public abstract class AbstractCatalogService extends AbstractControllerService i
...
+        if (configFiles != null && !configFiles.trim().isEmpty()) {
             for (final String configFile : configFiles.split(",")) {
-                conf.addResource(new Path(configFile.trim()));
+                File file = new File(configFile.trim());
+                try (final InputStream fis = new FileInputStream(file);
+                     final InputStream in = new BufferedInputStream(fis)) {
+                    final StandardDocumentProvider documentProvider = new StandardDocumentProvider();
+                    documentList.add(documentProvider.parse(in));
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
```
krisztina-zsihovszki, line 68: I'd rather use ProcessException and add some error message as well.

nifi-nar-bundles/nifi-iceberg-bundle/nifi-iceberg-services-api/src/main/java/org/apache/nifi/services/iceberg/IcebergCatalogProperties.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.services.iceberg;
+
+public class IcebergCatalogProperties {
+
+    public static final String METASTORE_URI = "metastoreUri";
+    public static final String WAREHOUSE_LOCATION = "warehouseLocation";
```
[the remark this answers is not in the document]
mark-bathori: Yes the Enum is more suitable for this. Since the enum value will be used as the catalogProperty key the "icebergPropertyName" is not needed anymore.

nifi-nar-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/AbstractIcebergProcessor.java
mark-bathori: Unfortunately the kerberos principal can't be used here since the controller service is not in enabled state when the isolation key is being retrieved so it will return with null.
As I can see the HDFS processors are using the kerberos principal. It works properly there due the `dynamicallymodifiesclasspath` which is used on the `hadoopConfigPath` property, so the isolation key is going to be evaluated also at processor start. This behaviour is missing here.

turcsanyip (approved):
Thanks @mark-bathori!

+1 LGTM

Ready to merge if @krisztina-zsihovszki approves too and when the CI build is ready.

krisztina-zsihovszki (approved):
lgtm, thank you @mark-bathori  for the changes

mark-bathori:
Thanks for the reviews @turcsanyip @krisztina-zsihovszki.

PR #7458, opened 2023-07-03
NIFI-11758: Added FileResourceService and used it in PutAzure*Storage…

turcsanyip:
… processors for local file upload

# Summary

[NIFI-11758](https://issues.apache.org/jira/browse/NIFI-11758)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/pom.xml
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/pom.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.nifi</groupId>
+        <artifactId>nifi-extension-utils</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>nifi-data-upload</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-utils</artifactId>
+            <version>2.0.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-file-resource-service-api</artifactId>
+        </dependency>
+    </dependencies>
+
+</project>
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/src/main/java/org/apache/nifi/processors/dataupload/DataUploadProperties.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/src/main/java/org/apache/nifi/processors/dataupload/DataUploadProperties.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.dataupload;
+
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.expression.ExpressionLanguageScope;
+import org.apache.nifi.fileresource.service.api.FileResourceService;
+
+import static org.apache.nifi.processors.dataupload.DataUploadSource.FLOWFILE_CONTENT;
+
+public class DataUploadProperties {
+
+    public static final PropertyDescriptor DATA_TO_UPLOAD = new PropertyDescriptor.Builder()
+            .name("data-to-upload")
+            .displayName("Data to Upload")
+            .description("The source of the content to be uploaded.")
+            .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+            .required(true)
+            .allowableValues(DataUploadSource.class)
+            .defaultValue(FLOWFILE_CONTENT.getValue())
+            .build();
+
+    public static final PropertyDescriptor FILE_RESOURCE_SERVICE = new PropertyDescriptor.Builder()
+            .name("file-resource-service")
+            .displayName("File Resource Service")
+            .description("File Resource Service providing a local file to be uploaded.")
+            .identifiesControllerService(FileResourceService.class)
+            .required(true)
+            .dependsOn(DATA_TO_UPLOAD, DataUploadSource.LOCAL_FILE)
+            .build();
+}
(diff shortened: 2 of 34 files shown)


nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/src/main/java/org/apache/nifi/processors/dataupload/DataUploadProperties.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.nifi.expression.ExpressionLanguageScope;
+import org.apache.nifi.fileresource.service.api.FileResourceService;
+
+import static org.apache.nifi.processors.dataupload.DataUploadSource.FLOWFILE_CONTENT;
+
+public class DataUploadProperties {
+
+    public static final PropertyDescriptor DATA_TO_UPLOAD = new PropertyDescriptor.Builder()
+            .name("data-to-upload")
+            .displayName("Data to Upload")
```
exceptionfactory, line 29: This property name seems a bit too generic to be clear. Aligning with the name of the allowable values enumeration, it seems like `Data Upload Source` or just `Data Source` would be clearer.

nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.nifi</groupId>
+        <artifactId>nifi-extension-utils</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>nifi-data-upload</artifactId>
```
exceptionfactory, line 26: Thinking about the name of this module and related elements, this seems like it should be named for more for the source than the intended destination. What do you think about `nifi-data-input` or `nifi-data-transfer`. Both of those also seem a bit generic, perhaps `nifi-resource-transfer` and then naming the `enum` as `ResourceTransferSource`?

nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/src/main/java/org/apache/nifi/processors/dataupload/DataUploadUtil.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * Returns the size of the FileResource if it is provided (not null). Otherwise, returns the size of the FlowFile.
+     *
+     * @param flowFile the FlowFile which is used when no FileResource is provided
+     * @param fileResource the FileResource
+     * @return size of the FileResource or the FlowFile
+     */
+    public static long getUploadSize(final FlowFile flowFile, final FileResource fileResource) {
```
exceptionfactory, line 76: As mentioned elsewhere, this has more to do with the source. Recommend naming this method something like `getResourceSize()` or `getTransferSize()`.

nifi-nar-bundles/nifi-standard-services/nifi-file-resource-service-bundle/nifi-file-resource-service/src/main/java/org/apache/nifi/fileresource/service/StandardFileResourceService.java
```diff
@@ -0,0 +1,113 @@
...
+                @Restriction(
+                        requiredPermission = RequiredPermission.READ_FILESYSTEM,
+                        explanation = "Provides operator the ability to read from any file that NiFi has access to.")
+        }
+)
+public class StandardFileResourceService extends AbstractControllerService implements FileResourceService {
+
+    public static final PropertyDescriptor FILE_PATH = new PropertyDescriptor.Builder()
+            .name("file-path")
+            .displayName("File Path")
```
exceptionfactory, line 60: Although this property value can be resolved from attributes, it otherwise seems limiting and potentially confusing to have a service that only returns a single FileResource. One option could be to make this property something like `Base Directory` and then change the signature of the `FileResourceService` to take a `URI` or `Path`. In other words, making the Service contract explicitly support return different files depending on input parameters, as opposed scoping to the bounds of the an attributes Map, which could be empty.

exceptionfactory (changes requested):
Thanks for reworking this capability @turcsanyip. The general approach looks good, but I made some comments regarding module and property naming.

The main question is the FileResourceService method contract. Limiting the input to a Map of attributes seems too generic, and effectively requires the calling Processor to know what Service implementation is being used. Instead, the calling Processor should explicitly send either a relative path or full path. A full path expressed as a `URI` seems like the most flexible option.

exceptionfactory (commented):
On further evaluation and offline discussion with @markap14, the `FileResourceService` method contract looks good. On the positive side, the scope of access is the responsibility of the implementation, which aligns with the File System Access permission on the `StandardFileResourceService` implementation.

I can make the other naming adjustments and then this should be ready to go.

nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/src/main/java/org/apache/nifi/processors/dataupload/DataUploadProperties.java
exceptionfactory: Renamed to `Resource Transfer Source` in updated commit.

nifi-nar-bundles/nifi-extension-utils/nifi-data-upload/pom.xml
exceptionfactory: Renamed to `nifi-resource-transfer` in updated commit.

nifi-nar-bundles/nifi-standard-services/nifi-file-resource-service-bundle/nifi-file-resource-service/src/main/java/org/apache/nifi/fileresource/service/StandardFileResourceService.java
exceptionfactory: Resolved based on further review retaining the current approach.

nandorsoma (commented):
Thanks for the improvement @turcsanyip and @exceptionfactory, for the final modifications.
- I've tested the processors, and they are working as expected
- Code-wise, I've found a few things. Please see them inline.
- During testing, I found it a bit cumbersome that the `File Name` and `Blob Name` name properties are on the processor while the source path is in the service. In my mind, they are logically connected, especially in this case. Purely from the user's point of view, I think it would be better to put these fields into the FileResourceService, though I understand it is not possible as it resides in a separate module. Tbh, I don't have a good idea to solve it atm, but I wanted to point it out.
- Another thing I noticed - not strictly related to the PR -, is that blob storage only allows `[a-z0-9\-]` characters as resource names. The issue is that the error message is not too helpful:
```
Status code 400, "﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidResourceName</Code><Message>The specifed resource name contains invalid characters.
RequestId:4d255e46-801e-0013-484b-b6a312000000
Time:2023-07-14T12:03:59.4382821Z</Message></Error>"
```
I changed the filename accordingly, but it still threw an error. It took a while to realize that the issue was the container name, and this restriction applies to it. I wonder whether we could add extra validation to improve the user experience.

nifi-nar-bundles/nifi-extension-utils/nifi-resource-transfer/src/main/java/org/apache/nifi/processors/transfer/ResourceTransferUtils.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Get File Resource from File Resource Service based on provided Source otherwise return empty
+     *
+     * @param resourceTransferSource type of the data upload
+     * @param context process context with properties
+     * @param flowFile FlowFile with attributes to use in expression language
+     * @return Optional FileResource retrieved from FileResourceService if Source is File Resource Service, otherwise empty
+     * @throws ProcessException Thrown if Source is File Resource but FileResourceService is not provided in the context
+     */
+    public static Optional<FileResource> getFileResource(final ResourceTransferSource resourceTransferSource, final ProcessContext context, final FlowFile flowFile) {
```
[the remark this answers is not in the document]
exceptionfactory: With this as a public utility method, I wanted to make it clear that it may not return a `FileResource`, so I changed it to use the `Optional` wrapper. I considered adjusting the calling code, and I will take another look at that in light of your comments.

nifi-nar-bundles/nifi-extension-utils/nifi-resource-transfer/src/main/java/org/apache/nifi/processors/transfer/ResourceTransferUtils.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Returns the input stream of the FileResource if it is provided (not null). Otherwise, returns the input stream of the FlowFile.
+     *
+     * @param session the session to read the FlowFile
+     * @param flowFile the FlowFile which is read when no FileResource is provided
+     * @param fileResource the FileResource
+     * @return input stream of the FileResource or the FlowFile
+     */
+    public static InputStream getTransferInputStream(final ProcessSession session, final FlowFile flowFile, final FileResource fileResource) {
```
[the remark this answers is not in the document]
exceptionfactory: I attempted to minimize changes, but with the adjustment to return `Optional<FileResource>`, this particular method probably needs to be adjusted.

exceptionfactory:
> During testing, I found it a bit cumbersome that the `File Name` and `Blob Name` name properties are on the processor while the source path is in the service. In my mind, they are logically connected, especially in this case. Purely from the user's point of view, I think it would be better to put these fields into the FileResourceService, though I understand it is not possible as it resides in a separate module. Tbh, I don't have a good idea to solve it atm, but I wanted to point it out. 

Thanks for the feedback and testing @nandorsoma! I pushed an update that removes two utility methods and instead uses the functional `Optional.map().orElse()` approach. This seems more natural under the circumstances.

This optional approach to file transfer is somewhat outside the bounds of normal processing, given that the `StandardFileResourceService` allows reading directly from the file system, so I understand how the property organization may not be the most intuitive. The correct configuration does require aligning the Service `File Path` with the expected input values on a given FlowFile, but that seems reasonable given that standard usage patterns will not make use of this feature. There could be room for improvement in subsequent iterations, but at least the current approach avoids introducing new permission restrictions on the existing Processors.

nandorsoma (approved):
Thank you, @exceptionfactory. The current version looks good. I will merge it once the checks have been passed.

nandorsoma:
Merging to `main` and `support/nifi-1.x`.

PR #7462, opened 2023-07-06
NIFI-11434 added support for optionally allowing zip entries with dat…

joewitt:
…a descriptors

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


nandorsoma (commented):
Thanks for this improvement, @joewitt! Functionally looks good, though I've found one thing that you can simplify.

nandorsoma:
I noticed one more thing. In the case of protected zip, we use `EncryptedZipInputStreamCallback`. Can we support this feature when zip is protected? Do we need to at all? If not, I'd suggest disabling the new property when the password field has a value. I also started to think that adding a few more tests with a zip file actually containing data descriptors would make sense. What do you think?

markap14:
Thanks @joewitt. Code changes make sense. Updated tests look good. I think all of the comments have been addressed. +1 will merge.

PR #7465, opened 2023-07-09
NIFI-11788 Modified logback configuration to have better cle…

MikeThomsen:
…anup

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
+++ b/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
@@ -30,11 +30,11 @@
             -->
             <fileNamePattern>${org.apache.nifi.minifi.bootstrap.config.log.dir}/${org.apache.nifi.minifi.bootstrap.config.log.app.file.name}_%d{yyyy-MM-dd_HH}.%i.${org.apache.nifi.minifi.bootstrap.config.log.app.file.extension}.gz</fileNamePattern>
             <!-- Control the maximum size of each log file before rolling over -->
-            <maxFileSize>1MB</maxFileSize>
+            <maxFileSize>10MB</maxFileSize>
             <!-- Control the maximum number of log archive files kept and asynchronously delete older files -->
             <maxHistory>10</maxHistory>
             <!-- Control the total size of all log archive files for this appender -->
-            <totalSizeCap>10MB</totalSizeCap>
+            <totalSizeCap>100MB</totalSizeCap>
             <!-- Log files exceeding maximum settings will be rolled on startup -->
             <cleanHistoryOnStart>true</cleanHistoryOnStart>
         </rollingPolicy>
@@ -55,11 +55,11 @@
             -->
             <fileNamePattern>${org.apache.nifi.minifi.bootstrap.config.log.dir}/${org.apache.nifi.minifi.bootstrap.config.log.bootstrap.file.name}_%d.${org.apache.nifi.minifi.bootstrap.config.log.bootstrap.file.extension}.gz</fileNamePattern>
             <!-- Control the maximum size of each log file before rolling over -->
-            <maxFileSize>1MB</maxFileSize>
+            <maxFileSize>10MB</maxFileSize>
             <!-- Control the maximum number of log archive files kept and asynchronously delete older files -->
             <maxHistory>10</maxHistory>
             <!-- Control the total size of all log archive files for this appender -->
-            <totalSizeCap>10MB</totalSizeCap>
+            <totalSizeCap>100MB</totalSizeCap>
             <!-- Log files exceeding maximum settings will be rolled on startup -->
             <cleanHistoryOnStart>true</cleanHistoryOnStart>
         </rollingPolicy>
--- a/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
+++ b/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
@@ -35,6 +35,8 @@
             <maxFileSize>1MB</maxFileSize>
             <!-- Provide a cap of 10 MB across all archive files -->
             <totalSizeCap>10MB</totalSizeCap>
+            <!-- This is supposed to force log cleanup when the service starts/restarts -->
+            <cleanHistoryOnStart>true</cleanHistoryOnStart>
         </rollingPolicy>
         <immediateFlush>true</immediateFlush>
         <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
@@ -54,6 +56,9 @@
             <fileNamePattern>${org.apache.nifi.minifi.bootstrap.config.log.dir}/${org.apache.nifi.minifi.bootstrap.config.log.bootstrap.file.name}_%d.${org.apache.nifi.minifi.bootstrap.config.log.bootstrap.file.extension}.gz</fileNamePattern>
             <!-- Keep 5 rolling periods worth of logs-->
             <maxHistory>5</maxHistory>
+            <totalSizeCap>100MB</totalSizeCap>
+            <!-- This is supposed to force log cleanup when the service starts/restarts -->
+            <cleanHistoryOnStart>true</cleanHistoryOnStart>
         </rollingPolicy>
         <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
             <pattern>%date %level [%thread] %logger{40} %msg%n</pattern>
--- a/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
+++ b/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
@@ -29,13 +29,13 @@
               To ZIP rolled files, replace '.log' with '.log.zip'.
             -->
             <fileNamePattern>${org.apache.nifi.minifi.bootstrap.config.log.dir}/${org.apache.nifi.minifi.bootstrap.config.log.app.file.name}_%d{yyyy-MM-dd_HH}.%i.${org.apache.nifi.minifi.bootstrap.config.log.app.file.extension}.gz</fileNamePattern>
-            <!-- Keep 10 rolling periods worth of log files-->
-            <maxHistory>10</maxHistory>
-            <!-- Max size each log file will be-->
+            <!-- Control the maximum size of each log file before rolling over -->
             <maxFileSize>1MB</maxFileSize>
-            <!-- Provide a cap of 10 MB across all archive files -->
+            <!-- Control the maximum number of log archive files kept and asynchronously delete older files -->
+            <maxHistory>10</maxHistory>
+            <!-- Control the total size of all log archive files for this appender -->
             <totalSizeCap>10MB</totalSizeCap>
-            <!-- This is supposed to force log cleanup when the service starts/restarts -->
+            <!-- Log files exceeding maximum settings will be rolled on startup -->
             <cleanHistoryOnStart>true</cleanHistoryOnStart>
         </rollingPolicy>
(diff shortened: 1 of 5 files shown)


minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
```diff
@@ -35,6 +35,8 @@
             <maxFileSize>1MB</maxFileSize>
             <!-- Provide a cap of 10 MB across all archive files -->
             <totalSizeCap>10MB</totalSizeCap>
+            <!-- This is supposed to force log cleanup when the service starts/restarts -->
+            <cleanHistoryOnStart>true</cleanHistoryOnStart>
```
joewitt, line 39: what does this mean?  it cleans everything or just to what the limit was?   If i keep 30 files or 30 days does it clean all that or does it get anything older?  Why is that needed if that is what it is does?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
```diff
@@ -30,10 +30,13 @@
               To GZIP rolled files, replace '.log' with '.log.gz'.
               To ZIP rolled files, replace '.log' with '.log.zip'.
             -->
-            <fileNamePattern>${org.apache.nifi.bootstrap.config.log.dir}/nifi-app_%d{yyyy-MM-dd_HH}.%i.log</fileNamePattern>
+            <fileNamePattern>${org.apache.nifi.bootstrap.config.log.dir}/nifi-app_%d{yyyy-MM-dd_HH}.%i.log.gz</fileNamePattern>
             <maxFileSize>100MB</maxFileSize>
             <!-- keep 30 log files worth of history -->
             <maxHistory>30</maxHistory>
+            <totalSizeCap>100GB</totalSizeCap>
```
joewitt, line 37: if the default is 100MB per file and keeping 30 files then 3GB seems like a reasonable default cap size.  Am I reading that right?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
```diff
@@ -53,6 +56,9 @@
             <fileNamePattern>${org.apache.nifi.bootstrap.config.log.dir}/nifi-user_%d.log</fileNamePattern>
             <!-- keep 30 log files worth of history -->
             <maxHistory>30</maxHistory>
+            <totalSizeCap>5GB</totalSizeCap>
```
joewitt, line 59: perhaps we should add max file size on this to be consistent with others?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
```diff
@@ -64,6 +70,9 @@
         <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
             <fileNamePattern>${org.apache.nifi.bootstrap.config.log.dir}/nifi-request_%d.log</fileNamePattern>
             <maxHistory>30</maxHistory>
+            <totalSizeCap>5GB</totalSizeCap>
+            <!-- This is supposed to force log cleanup when the service starts/restarts -->
```
joewitt, line 74: max file size here too?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
```diff
@@ -82,6 +91,9 @@
             <fileNamePattern>${org.apache.nifi.bootstrap.config.log.dir}/nifi-bootstrap_%d.log</fileNamePattern>
             <!-- keep 5 log files worth of history -->
             <maxHistory>5</maxHistory>
+            <totalSizeCap>100MB</totalSizeCap>
```
joewitt, line 94: max file size?

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
MikeThomsen: When I tested it locally, I set `maxHistory` to 1 so I could easily experiment with logback's behavior. I found the cleanup was really inconsistent, but if I set `cleanHistoryOnStart` to true it would at least be (or seem to be) very strict at enforcing cleanup if the node got bounced. The workflow I did was:

1. Set maxHistory to 1 (1 day)
2. Copy the current timestamped file to TS - 1 day
3. Restart node

On restart, logback cleaned up the old files.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
MikeThomsen: `3GB` is fine here. I'm not married to any of these values, I was just erring on the side of caution.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
MikeThomsen: Probably a good idea

MikeThomsen:
@joewitt my looking into this was prompted by our devops team noticing that log cleanup was simply not happening on our clusters even after the 30 day window. I'm not done investigating on our end because there's some funky stuff happening on those boxes where .log files are getting rotated and then gzipped, and the logback gzip hint (adding ".gz" to the name in the rolling strategy) was not present in those configs. It might be something weird with monitoring on those boxes, but I'm hoping to at least help us tighten up our config to help mitigate anything logback might be doing to contribute.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
```diff
@@ -30,10 +30,13 @@
               To GZIP rolled files, replace '.log' with '.log.gz'.
               To ZIP rolled files, replace '.log' with '.log.zip'.
             -->
-            <fileNamePattern>${org.apache.nifi.bootstrap.config.log.dir}/nifi-app_%d{yyyy-MM-dd_HH}.%i.log</fileNamePattern>
+            <fileNamePattern>${org.apache.nifi.bootstrap.config.log.dir}/nifi-app_%d{yyyy-MM-dd_HH}.%i.log.gz</fileNamePattern>
```
exceptionfactory, line 33: I think this particular change adding `.gz` should be reverted. Although it is more efficient in terms of space usage, it should either be something done across the board, or left as it is for now.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
```diff
@@ -95,6 +107,9 @@
             <maxFileSize>10MB</maxFileSize>
             <maxHistory>10</maxHistory>
             <totalSizeCap>100MB</totalSizeCap>
+            <totalSizeCap>5GB</totalSizeCap>
```
exceptionfactory, line 110: This `5GB` setting duplicates the existing value of `100MB`. In this particular case, the setting for `nifi-deprecation.log` is intentionally smaller to avoid excessive repetition.

exceptionfactory (changes requested):
Thanks for putting together these adjustments @MikeThomsen.

The general approach of adding the `totalSizeCap` and `cleanHistoryOnStart` makes sense and seems like it should provide better expected behavior.

The `nifi-deprecation.log` already has a `totalSizeCap` of `100MB`, which is intentionally small. Setting a reasonable larger default for other logs would be useful.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/logback.xml
mosermw: Actually "keep 30 log files worth of history" is a bit misleading.  maxHistory ignores the %i part of the fileNamePattern.  If your pattern is "nifi-app_%d{**yyyy-MM-dd_HH**}.%i.log" then it's maxHistory hours, and a pattern of "nifi-app_%d{**yyyy-MM-dd**}.%i.log" is maxHistory days.  You can rollover with the %i pattern hundreds of times and maxHistory doesn't care.
Regardless, I'm in favor of a smaller default totalSizeCap instead of 100GB.  nifi-registry and stateless appear to use 10GB here.  I have no idea if totalSizeCap respects the %i pattern.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
mosermw: I tested this briefly, and cleanHistoryOnStart did respect the existing maxHistory when it cleaned up much older files that were outside that window.

exceptionfactory (commented):
@MikeThomsen Are you planning to revisit this soon? If you are able to incorporate the feedback and rebase, I think this should be close to completion.

MikeThomsen:
@exceptionfactory yeah. Going to try to finish it up this weekend.

joewitt:
@MikeThomsen Wanted to ping and see if you're able to get this one wrapped up?  Seems good to get it done.  Probably makes sense as 'main/2.x only'.  As we close in on 2x i am less inclined to make changes to defaults like this on the 1x/support line but others views may differ.

exceptionfactory (commented):
@MikeThomsen, @joewitt, and @mosermw, I rebased and pushed a commit that sets standard values across components and logs. MiNiFi retains a smaller set of default values given the intended deployment approach, and NiFi deprecation logging has smaller settings, but the rest of the files following the same approach.

Here is a summary:

- Set MiNiFi maxHistory to 10 and totalSizeCap to 10MB
- Set NiFi maxHistory to 30 and totalSizeCap to 3GB except for deprecation log
- Set Stateless maxHistory to 30 and totalSizeCap to 3GB
- Set Registry maxHistory to 30 and totalSizeCap to 3GB

I don't think this should be applied to the support branch, but seems like a reasonable introduction for NiFi 2.0.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
```diff
@@ -29,12 +29,14 @@
...
-            <!-- Max size each log file will be-->
+            <!-- Control the maximum size of each log file before rolling over -->
             <maxFileSize>1MB</maxFileSize>
-            <!-- Provide a cap of 10 MB across all archive files -->
+            <!-- Control the maximum number of log archive files kept and asynchronously delete older files -->
+            <maxHistory>10</maxHistory>
+            <!-- Control the total size of all log archive files for this appender -->
             <totalSizeCap>10MB</totalSizeCap>
+            <!-- Log files exceeding maximum settings will be rolled on startup -->
+            <cleanHistoryOnStart>true</cleanHistoryOnStart>
```
pvillard31, line 39: While we're making changes in this file, I'd be in favor for increasing 1MB to 10MB and 10MB to 100MB as the current values seem a bit small to me.

exceptionfactory: Thanks @pvillard31, I was also thinking those defaults were on the small side. I pushed a change for this MiNiFi configuration to use `maxFileSize` of 10MB and `totalSizeCap` of 100MB.

pvillard31 (approved):
+1

PR #7473, opened 2023-07-12
NIFI-11794 - Fix NPE + configure max attempts for Redis State Provider

pvillard31:
# Summary

[NIFI-11794](https://issues.apache.org/jira/browse/NIFI-11794) - Fix NPE + configure max attempts for Redis State Provider

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/state-management.xml
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/conf/state-management.xml
@@ -105,6 +105,9 @@
                     nifi.properties.  This means that a TLS-enabled Redis connection is only possible if the Apache NiFi instance is running in secure mode.
                     If this property is false, an insecure Redis connection will be used even if the Apache NiFi instance is secure (default false).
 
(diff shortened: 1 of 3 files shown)


ottobackwards:
This looks good, is there a way to test this specifically that we should cover?

pvillard31:
I tried adding a unit test showing the issue but didn't find a way to reproduce it. You have to simulate two nodes at the same time trying to access the same key.

ottobackwards:
would a failure to connect over max attempts ( to get the `null` and new check) work?  without the nodes at all? or would that fail a different way?

ottobackwards:
then again, a null check is a pretty obvious fix there, maybe we don't need to....

pvillard31:
That would not cause the null error. EXEC returns null in case the transaction has been aborted. So you'd need to simulate one node/thread to have a transaction starting with a WATCH on the key, then another node/thread completing a transaction on the same key, to have the first one receiving a null. The logs I shared in the JIRA comment are coming from a cluster that I patched with this fix.

pvillard31:
Thanks @ottobackwards

PR #7481, opened 2023-07-14
NIFI-8128: add support for specifying a password for Sentinel

EndzeitBegins:
# Summary

[NIFI-8128](https://issues.apache.org/jira/browse/NIFI-8128)

In order to reproduce the issue in a test case the Redis Sentinel setting `sentinel auth-pass` is required, which is only available in a newer Redis server binary as the one bundled into the `com.github.kstyrc:embedded-redis` dependency.
Thus I bundled a more recent binary (for Linux only).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
+++ b/nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
@@ -82,6 +82,11 @@
             <artifactId>nifi-record-path</artifactId>
             <version>2.0.0-SNAPSHOT</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-ssl-context-service-api</artifactId>
+            <scope>compile</scope>
+        </dependency>
         <!-- Test dependencies -->
         <dependency>
             <groupId>org.apache.nifi</groupId>
@@ -96,15 +101,15 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>com.github.kstyrc</groupId>
-            <artifactId>embedded-redis</artifactId>
-            <version>0.6</version>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>${testcontainers.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.apache.nifi</groupId>
-            <artifactId>nifi-ssl-context-service-api</artifactId>
-            <scope>compile</scope>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 </project>
--- a/nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
+++ b/nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
@@ -85,7 +85,7 @@
         <dependency>
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-ssl-context-service-api</artifactId>
-            <scope>compile</scope>
+            <scope>provided</scope>
         </dependency>
         <!-- Test dependencies -->
         <dependency>
--- a/nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/test/java/org/apache/nifi/redis/service/ITRedisDistributedMapCacheClientService.java
+++ b/nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/test/java/org/apache/nifi/redis/service/ITRedisDistributedMapCacheClientService.java
@@ -169,7 +169,7 @@ public class ITRedisDistributedMapCacheClientService {
     private RedisContainer setupStandaloneRedis(final @Nullable String redisPassword) throws IOException {
         int redisPort = getAvailablePort();
 
-        RedisContainer redisContainer = new RedisContainer("redis:7.0.12-alpine");
+        RedisContainer redisContainer = new RedisContainer(CONTAINER_IMAGE_TAG);
         redisContainer.mountConfigurationFrom(testDirectory);
         redisContainer.setPort(redisPort);
         redisContainer.addPortBinding(redisPort, redisPort);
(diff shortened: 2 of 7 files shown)


exceptionfactory (changes requested):
Thanks for the contribution @EndzeitBegins.

Including the Redis Server binary in the source repository is not a good strategy for testing due to size and version considerations. One option could be to evaluate an integration test using Testcontainers, but in absence of something else, it would be best to remove the server binary and associated integration test methods.

EndzeitBegins:
@exceptionfactory Thanks for the quick review.

I understand. I was thinking the same, but assumed it might be okay-ish due to the fact, that the test library used `embedded-redis` does the same. 

I'll take a look at replacing `embedded-redis`, which hasn't seen maintenance for a long time by the way, with Testcontainers.

If that proves to time consuming I'd propose to remove the binary and `@Disable` the one test that requires a newer version completely, with an hint that in order to execute it, one must provide a newer binary locally.

I think it's a good thing to at least have a test case that reproduces an undesired behaviour, before fixing it.

Would you be fine with both approaches?

EndzeitBegins:
I removed both the binary file and the dependency on the unmaintained [embedded-redis](https://github.com/kstyrc/embedded-redis) test library. 

It proved a little more complicated than anticipated to get the tests working with Testcontainers, due to the auto-discovery mechanisms of Redis (Sentinel), but it seems to work now. 

I would appreciate if you'd take another look at the PR, @exceptionfactory.

nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
```diff
@@ -82,6 +82,11 @@
             <artifactId>nifi-record-path</artifactId>
             <version>2.0.0-SNAPSHOT</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-ssl-context-service-api</artifactId>
+            <scope>compile</scope>
```
exceptionfactory, line 88: This dependency should be marked as provided.
```suggestion
            <scope>provided</scope>
```

nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/test/java/org/apache/nifi/redis/service/ITRedisDistributedMapCacheClientService.java
```diff
@@ -64,22 +73,154 @@
...
+            redisConnectionPool.onDisabled();
+        }
+
+        redisContainers.forEach(RedisContainer::stop);
+    }
+
+    private RedisContainer setupStandaloneRedis(final @Nullable String redisPassword) throws IOException {
+        int redisPort = getAvailablePort();
+
+        RedisContainer redisContainer = new RedisContainer("redis:7.0.12-alpine");
```
exceptionfactory, line 172: Recommend setting a static member variable for the image name and version, and reusing it in all locations.
```suggestion
        RedisContainer redisContainer = new RedisContainer(CONTAINER_IMAGE_TAG);
```

nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/test/java/org/apache/nifi/redis/testcontainers/RedisContainer.java
```diff
@@ -0,0 +1,113 @@
...
+            if (mountDirectory == null) {
+                mountDirectory = Files.createTempDirectory("redis-container-configuration");
+            }
+
+            Path configFile = mountDirectory.resolve("redis-" + UUID.randomUUID() + ".conf");
+            Files.write(configFile, configurationOptions, StandardCharsets.UTF_8);
+
+            return configFile;
+        } catch (IOException ioException) {
+            throw new IllegalStateException("Cannot start container because configuration could not be written!", ioException);
```
exceptionfactory, line 110: Exclamation marks should be avoided in error messages.
```suggestion
            throw new IllegalStateException("Cannot start container because configuration could not be written", ioException);
```

nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/test/java/org/apache/nifi/redis/testcontainers/RedisSentinelContainer.java
```diff
@@ -0,0 +1,115 @@
...
+            addConfigurationOption("sentinel sentinel-pass " + sentinelPassword);
+        }
+    }
+
+    @Override
+    protected void configure() {
+        super.configure();
+
+        System.err.println("SENTINEL AT PORT " + port);
+        System.err.println(configurationOptions);
```
exceptionfactory, line 109: Recommend removing System print statements. Loggers could be used if absolutely necessary.

exceptionfactory (commented):
Thanks for making the changes @EndzeitBegins! Using Testcontainers look much better than the previous approach. I noted a couple very minor recommendations, but this looks like a good way to go.

nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
EndzeitBegins: Sure, I just moved this declaration up, as it was listed under the "test dependencies" comment.

nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/test/java/org/apache/nifi/redis/testcontainers/RedisSentinelContainer.java
EndzeitBegins:
```suggestion
```

EndzeitBegins: Whoops, this was a leftover from debugging. Good catch!

nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
exceptionfactory: Thanks!

EndzeitBegins:
I applied the recommended changes, @exceptionfactory. Thank you for the code review.

exceptionfactory (approved):
Thanks for working through the feedback @EndzeitBegins, the latest changes look good and the updated integration tests with Testcontainers work as expected. +1 merging

PR #7484, opened 2023-07-17
NIFI-11817: Improve ListHDFS extensibility

Lehel44:
# Summary

[NIFI-11817](https://issues.apache.org/jira/browse/NIFI-11817)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/ListHDFS.java
```diff
@@ -263,24 +267,35 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
+                    .minimumAge(minimumAge)
+                    .maximumAge(maximumAge)
+                    .previousLatestTimestamp(latestTimestamp)
+                    .previousLatestFiles(latestFiles)
+                    .writerFactory(writerFactory)
+                    .hdfsPrefix(getAttributePrefix())
+                    .logger(getLogger())
+                    .build();
+
+            final HadoopFileStatusWriter writer = (writerFactory == null)
```
simonbence, line 292: Adding a builder looks to be a great step forward! It could be even more beneficial for us if the builder would not only build the context but the writer and the actual writer instance creation would be hidden behind the `build` method. I think, adding an extra method which can help the builder decide what kind of output we expect (maybe a small enum) would help with this and this way also the context and the writer implementations might remain unexposed.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/util/writer/HadoopWriterContext.java
```diff
@@ -0,0 +1,202 @@
...
+            return this;
+        }
+
+        public Builder hdfsPrefix(final String hdfsPrefix) {
+            this.hdfsPrefix = hdfsPrefix;
+            return this;
+        }
+
+        public HadoopWriterContext build() {
+            return new HadoopWriterContext(session, successRelationship, fileStatusIterable, fileStatusManager, pathFilter, minimumAge, maximumAge,
```
simonbence, line 150: I would suggest to add some validation here which could check if all the mandatory attributes are set. Like for example: checking if `session` is not null.

simonbence (commented):
Other than two minor improvement points, the PR looks good to me. Thanks for iterating on this!

PR #7489, opened 2023-07-17
NIFI-11823 - fix NUMERIC support in PutBigQuery

pvillard31:
# Summary

[NIFI-11823](https://issues.apache.org/jira/browse/NIFI-11823) - fix NUMERIC support in PutBigQuery

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
```diff
@@ -17,50 +17,99 @@
...
+                   } else if (tableSchema.getFields(field.getIndex()).getType().equals(Type.NUMERIC)) {
+                       value = BigDecimalByteStringEncoder.encodeToNumericByteString((BigDecimal) value);
+                   }
+               } else if (value instanceof Long) {
+                   value = Int64Value.of((long) value).toByteString();
+               } else if (value instanceof Float) {
+                   value = FloatValue.of((float) value).toByteString();
+               } else if (value instanceof Double) {
+                   value = DoubleValue.of((double) value).toByteString();
+               }
```
turcsanyip, line 93: `Integer` should be added too.
Avro `long` type is accepted for `NUMERIC` field but `int` is not.

nifi-nar-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
```diff
@@ -17,50 +17,99 @@
...
+                   value = ((Float) value).doubleValue();
+               }
+               setField(value, field, builder);
+               break;
+
+           // matches NUMERIC and BIGNUMERIC types in BigQuery
+           // BQTableSchemaToProtoDescriptor.class
+           case BYTES:
+               if(value instanceof BigDecimal) {
+                   if(tableSchema.getFields(field.getIndex()).getType().equals(Type.BIGNUMERIC)) {
```
turcsanyip, line 82: Minor formatting:
```suggestion
               if (value instanceof BigDecimal) {
                   if (tableSchema.getFields(field.getIndex()).getType().equals(Type.BIGNUMERIC)) {
```

turcsanyip (changes requested):
@pvillard31 Thanks for fixing the decimal handling in the BigQuery processor!

I tested it with different numeric types and it works properly except Avro `int` (pls. see my comment below).

I can also see regression in the `testStreamingNoErrorWithDateFormat` IT test. Could you please check it? For me it constantly succeeds on main but fails on the PR branch.

nifi-nar-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
pvillard31: Fixed, thanks

nifi-nar-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
pvillard31: So I tried to consider the case where we send an Avro ``int`` into a ``NUMERIC`` column for BigQuery but this is completely messing things up. I can add something like:
````java
 } else if (value instanceof Integer) {
    value = Int64Value.of((int) value).toByteString();
 }
````
But then, in the test when checking the returned value:
````java
assertEquals(firstElt.get(0).getDoubleValue(), 1.0);
````
It would tell me:
````
org.opentest4j.AssertionFailedError: expected: <2.64E-7> but was: <1.0>
````

If sending the Avro ``int`` into a ``INT64`` column for BigQuery, then it works as expected.

I'm going to push a commit containing the additional else if so that it does not completely fail in NiFi but I believe this would push wrong data into the BQ table.

**Comment by pvillard31** [comment]
> I can also see regression in the `testStreamingNoErrorWithDateFormat` IT test. Could you please check it? For me it constantly succeeds on main but fails on the PR branch.

Good catch, I guess I focused on the tests I added and forgot to re-run this one... after investigation it feels like trying to send a ``long`` into a NUMERIC column is not possible with the current code. We may need to convert int/long into a decimal type before switching to the ByteString. For now, I just fixed the test by switching this specific column to INT64 instead of NUMERIC but curious to hear your thoughts about this.

**Comment by turcsanyip** [comment]
Thanks @pvillard31 for your detailed explanation!

I did not check the result in the BigQuery table, just saw that `long` was successful, and that's why I thought `int` should work too. Now I understand that none of `int`, `long`, `float` or `double` can be inserted into a `NUMERIC` field with the current code because it creates a different binary representation.

What we can do (if we want to support cross-type conversions, e.g. `int` to `numeric`) is to convert the Java `int`, `long`, etc. values into `BigDecimal` first, and then use `BigDecimalByteStringEncoder` (as for the values originally in `BigDecimal`).

Something like this:
```
            case BYTES:
                if (value instanceof Integer) {
                    value = new BigDecimal((int) value);
                } else if (value instanceof Long) {
                    value = new BigDecimal((long) value);
                } else if (value instanceof Float || value instanceof Double) {
                    value = new BigDecimal(value.toString());
                }

                if (value instanceof BigDecimal) {
                    if (tableSchema.getFields(field.getIndex()).getType().equals(Type.BIGNUMERIC)) {
                        value = BigDecimalByteStringEncoder.encodeToBigNumericByteString((BigDecimal) value);
                    } else if (tableSchema.getFields(field.getIndex()).getType().equals(Type.NUMERIC)) {
                        value = BigDecimalByteStringEncoder.encodeToNumericByteString((BigDecimal) value);
                    }
                }
```

File: nifi-nar-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java [hunk 3]
```diff
@@ -17,50 +17,101 @@
...
+                   value = Long.valueOf((Integer) value);
+               }
+
+               setField(value, field, builder);
+               break;
+
+           // FLOAT64
+           case DOUBLE:
+               if (value instanceof Float) {
+                   value = ((Float) value).doubleValue();
```
turcsanyip, line 73: `Float.doubleValue()` can lead to weird numbers due to different precisions of `float` and `double`.
E.g. `21.34` in `float` would be converted to `21.34000015258789` in `double`.

I suggest using `Double.valueOf(value.toString())` instead.

pvillard31:
Thanks @turcsanyip - I went ahead with the changes your suggested

turcsanyip (approved):
Thanks @pvillard31!

+1 LGTM
Merging to main and 1.x.

PR #7492, opened 2023-07-17
NIFI-11827: Create Schema Registry Controller Service for AWS Glue

Lehel44:
# Summary

[NIFI-11827](https://issues.apache.org/jira/browse/NIFI-11827)

After I looked into the schema-registry-serde:SchemaRegistryClient I dropped the idea to use it. I examined the underlying Glue client which is much more customizable. I figured out every single way to get a schema from Glue.

```
SchemaVersionNumber schemaVersionNumber = SchemaVersionNumber.builder()
        //.versionNumber(1L)
        .latestVersion(true)
        .build();

SchemaId schemaId = SchemaId.builder()
        .registryName(registryName)
        .schemaName(schemaName)
        //.schemaArn(arn)
        .build();

GetSchemaVersionRequest request = GetSchemaVersionRequest.builder()
        .schemaVersionNumber(schemaVersionNumber)
        .schemaId(schemaId)
        //.schemaVersionId(schemaVersionId)
        .build();

GetSchemaVersionResponse schemaVersionResponse = client.getSchemaVersion(request);
```

- SchemaVersionNumber: versionNumber OR latestVersion, only one can be specified. I can think of a version property with a default value. In case of the default value we just set the latestVersion(true).

- SchemaId: (registryName and schemaName) OR schemaArn, only one can be specified.
- GetSchemaVersionRequest: (schemaVersionId) OR (schemaVersionNumber and schemaId), only once can be specified.

So  we have a few options to retrieve a schema. Currently without modifying lots of code we can only support this option:
schemaVersionNumber, registryName, schemaName (Use 'Schema Name' property)
in case we would like to add the option to retrieve a schema by
ARN - we either have to create a new SchemaIdentifier implementation or new SchemaStrategy for AWS Glue
schemaVersionId - we have to change the StandardSchemaIdentifier interface's schemaVersionId type from OptionalLong to Optional<String> or we need to create a new SchemaStrategy for AW Glue

Please note that I removed the builder.maxConnections(context.getMaxConcurrentTasks()) when creating the HTTP Client as it's not applicable in ConfigurationContext.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-aws-bundle/nifi-aws-schema-registry-service/src/main/java/org/apache/nifi/aws/schemaregistry/client/GlueSchemaRegistryClient.java
```diff
@@ -0,0 +1,109 @@
...
+
+        final GetSchemaVersionResponse schemaVersionResponse = getSchemaVersionResponse(schemaName, schemaVersionNumber);
+
+        return createRecordSchema(schemaVersionResponse);
+    }
+
+    @Override
+    public RecordSchema getSchema(String schemaName, int version) throws IOException, SchemaNotFoundException {
+        final SchemaVersionNumber schemaVersionNumber = SchemaVersionNumber.builder()
+                .versionNumber((long) version)
```
exceptionfactory, line 63: As `SchemaRegistryClient` is an interface internal to this implementation, it seems like the `version` parameter should be a `long` to avoid this hidden casting in the implementation.

exceptionfactory (changes requested):
Thanks for working on this new feature @Lehel44! The general approach looks good, I noted some implementation recommendations.

nifi-nar-bundles/nifi-aws-bundle/pom.xml
```diff
@@ -38,6 +38,8 @@
         <module>nifi-aws-abstract-processors</module>
         <module>nifi-aws-parameter-value-providers</module>
         <module>nifi-aws-parameter-providers</module>
+        <module>nifi-aws-schema-registry-service</module>
+        <module>nifi-aws-schema-registry-service-nar</module>
```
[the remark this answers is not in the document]
exceptionfactory: It looks like the `nifi-aws-schema-registry-service-nar` is 21 MB, which very large by comparison to most others. For this reason, I recommend bundling the new service JAR in the existing `nifi-aws-nar`, which should avoid additional copies of many transitive dependencies.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-schema-registry-service/src/main/java/org/apache/nifi/aws/schemaregistry/AmazonGlueSchemaRegistry.java
```diff
@@ -0,0 +1,248 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.aws.schemaregistry;
+
+import com.amazonaws.regions.Regions;
```
exceptionfactory, line 19: This should be changed to use the version 2 Region class `software.amazon.awssdk.regions.Region`. See the `RegionUtil` class for creating the allowable values.

exceptionfactory (changes requested):
Thanks for making the adjustments @Lehel44, this looks close to completion. The service works as expected at runtime, but I noticed one reference to the version 1 `Region` class instead of version 2. This should be changed so that the service uses AWS SDK version 2 classes for all operations.

Lehel44:
Thanks for the review and testing @exceptionfactory!

exceptionfactory (approved):
Thanks for making the adjustments @Lehel44, the latest version looks good! +1 merging

PR #7495, opened 2023-07-18
NIFI-11828 - Confluent Schema Encoding Access Strategy - Schema ID versus Schema Version ID

pvillard31:
# Summary

[NIFI-11828](https://issues.apache.org/jira/browse/NIFI-11828)

When we're using the Confluent Schema Encoding access strategy, we're currently retrieving the schema ID from the bytes at the beginning of the message and we're creating a Schema Identifier that is set with the Schema ID when it should actually use the Schema Version ID.

This bug may cause issues when trying to interoperate the Confluent Schema Encoding with different Schema Registries.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-standard-record-utils/src/main/java/org/apache/nifi/schema/access/ConfluentSchemaRegistryStrategy.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-standard-record-utils/src/main/java/org/apache/nifi/schema/access/ConfluentSchemaRegistryStrategy.java
@@ -54,7 +54,7 @@ public class ConfluentSchemaRegistryStrategy implements SchemaAccessStrategy {
 
         // This encoding follows the pattern that is provided for serializing data by the Confluent Schema Registry serializer
         // as it is provided at:
-        // http://docs.confluent.io/current/schema-registry/docs/serializer-formatter.html#wire-format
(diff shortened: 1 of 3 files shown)


nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-standard-record-utils/src/main/java/org/apache/nifi/schema/access/ConfluentSchemaRegistryStrategy.java
```diff
@@ -67,7 +67,7 @@ public RecordSchema getSchema(final Map<String, String> variables, final InputSt
         final int schemaId = bb.getInt();
 
         final SchemaIdentifier schemaIdentifier = SchemaIdentifier.builder()
-                .id(Long.valueOf(schemaId))
+                .schemaVersionId(Long.valueOf(schemaId))
                 .version(1)
```
bbende, line 71: I'm wondering if we can also remove this hard-coded version 1, I don't think it was even used for anything before, but now that we are switching to `schemaVersionId`, it should definitely not be used.

pvillard31: yep, that's fair, I don't see this version 1 used anywhere - I pushed a commit to remove it

bbende (approved):
Looks good!

PR #7502, opened 2023-07-19
NIFI-10904 Changed the font color for drop down menus to avoid the app…

Freedom9339:
…earance of menu items being disabled

# Summary

[NIFI-00000](https://issues.apache.org/jira/browse/NIFI-00000) - Changed the font color for drop down menus in controller and processor properties. The current color gives the appearance that the drop down option is disabled.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/common-ui.css
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/common-ui.css
@@ -162,9 +162,14 @@ ul.links span.header-link-over {
     display: none;
 }
 
-.blank, .unset, .sensitive {
+.blank, .sensitive {
     font-weight: normal !important;
     color: #a8a8a8 !important;
+}
+
+.unset {
+    font-weight: normal !important;
+    color: #262626 !important;
 }
 
 .required {
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/common-ui.css
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/common-ui.css
@@ -162,14 +162,9 @@ ul.links span.header-link-over {
     display: none;
 }
 
-.blank, .sensitive {
+.blank, .unset, .sensitive {
     font-weight: normal !important;
     color: #a8a8a8 !important;
-}
-
-.unset {
-    font-weight: normal !important;
-    color: #262626 !important;
 }
 
 .required {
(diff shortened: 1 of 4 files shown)


exceptionfactory (changes requested):
Thanks for the contribution @Freedom9339!

The user interface has a large number of references to the `unset` CSS class, so this change would have a broader impact than just the drop-down menu items. For that reason, it may be necessary to introduce a new CSS class or use a different class to implement a more targeted adjustment.

Freedom9339:
Thank you for your feedback @exceptionfactory. I've changed it to use .combo-option-text instead of .unset to set the color. This should only affect the color of drop down menu texts.

exceptionfactory (changes requested):
Thanks for the update @Freedom9339. Unfortunately this approach is also too broad because it changes the styles of all disabled elements. The best approach will involve a new style class that applies specifically to selectable options that are not disabled.

exceptionfactory (changes requested):
@Freedom9339 Sorry for the delay in reply. Unfortunately this change does not work either because the color property is not getting applied when rendering in the browser.

Freedom9339:
@markobean Yes, that's what I expected it to be displayed. Thank You.

exceptionfactory (commented):
@Freedom9339 and @markobean, I tested the change on Firefox 116 and confirmed that the new CSS is present, but it does not change the color. According to Web Developer Tools, the `color` style is not applied because the `unset` class takes precedence with its associated color. It appears that adding the `!important` flag to the color declaration could work.

On that note, however, it seems like it would still be helpful to distinguished the `Reference parameter...` and `Create Controller Service...` options from other options. Perhaps italicizing would be helpful as a way to differentiate these choices from other available options.

mtien-apache:
I checked in Chrome and the `unset` class still takes precedence.

Screenshot 2023-08-25 at 6 38 37 PM

I wonder if we just need to remove the `unset` class. It seems like this class is used for disabled, unavailable, or sensitive items. Also not a bad idea to italicize `Reference parameter...` and `Create Controller Service...` options to help differentiate from other options.

Freedom9339:
@mtien-apache Yea, so for those 2 cases, Reference parameter and Create Controller Service, the unset tag does take precedence, However, I thought this was intended for these 2 options, as well as any disabled options. If there are any other options, they will get the new tag and be colored black. I can remove the unset tag for these 2 options and italicize them. That way they will be colored black, but still be distinguishable from other non disabled options.

mtien-apache:
@Freedom9339 Looks like the `unset` tag is initially added when the the option item is created, but should have probably been removed after verifying the option is valid. I don't think the new tag is necessary. If the `unset` tag is not present, then the text color will default to black.

mtien-apache (approved):
LGTM! Thanks for all the fixes. There are some CI build failures but it doesn't look like they're related to your work. Will merge soon.

PR #7529, opened 2023-07-27
NIFI-8531: Upgrading Kinesis processors to AWS SDK 2.x

gresockj:
# Summary

[NIFI-8531](https://issues.apache.org/jira/browse/NIFI-8531)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


ChrisSamo632:
@gresockj FYI you have a checkstyle failure with the latest commit

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.3.0:check (check-style) on project nifi-aws-abstract-processors: You have 1 Checkstyle violation. -> [Help 1]
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRaw.java
```diff
@@ -38,12 +39,16 @@ public KinesisRecordProcessorRaw(final ProcessSessionFactory sessionFactory, fin
...
-        final String partitionKey = kinesisRecord.getPartitionKey();
-        final String sequenceNumber = kinesisRecord.getSequenceNumber();
-        final Date approximateArrivalTimestamp = kinesisRecord.getApproximateArrivalTimestamp();
-        final byte[] data = kinesisRecord.getData() != null ? kinesisRecord.getData().array() : new byte[0];
+        final String partitionKey = kinesisRecord.partitionKey();
+        final String sequenceNumber = kinesisRecord.sequenceNumber();
+        final Instant approximateArrivalTimestamp = kinesisRecord.approximateArrivalTimestamp();
+        final ByteBuffer dataBuffer = kinesisRecord.data();
+        byte[] data = dataBuffer != null ? new byte[dataBuffer.remaining()] : new byte[0];
+        if (data != null) {
```
ChrisSamo632, line 49:
```suggestion
        if (dataBuffer != null) {
```

`data` is never `null` here, but `dataBuffer` might be

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -73,11 +74,15 @@ void startProcessingRecords() {
...
     @Override
-    void processRecord(final List<FlowFile> flowFiles, final Record kinesisRecord, final boolean lastRecord,
+    void processRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kinesisRecord, final boolean lastRecord,
                        final ProcessSession session, final StopWatch stopWatch) {
         boolean firstOutputRecord = true;
         int recordCount = 0;
-        final byte[] data = kinesisRecord.getData() != null ? kinesisRecord.getData().array() : new byte[0];
+        final ByteBuffer dataBuffer = kinesisRecord.data();
+        byte[] data = dataBuffer != null ? new byte[dataBuffer.remaining()] : new byte[0];
+        if (data != null) {
```
ChrisSamo632, line 83:
```suggestion
        if (dataBuffer != null) {
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -371,9 +391,9 @@ protected PropertyDescriptor getSupportedDynamicPropertyDescriptor(final String
     private ValidationResult validateDynamicKCLConfigProperty(final String subject, final String input, final ValidationContext context) {
         final ValidationResult.Builder validationResult = new ValidationResult.Builder().subject(subject).input(input);
 
-        if (!subject.matches("^(?!with)[a-zA-Z]\\w*$")) {
+        if (!subject.matches("^(?!with)[a-zA-Z][a-zA-Z0-9_]*(\\.[a-zA-Z][a-zA-Z0-9_]*)?$")) {
```
ChrisSamo632, line 394:
```suggestion
        if (!subject.matches("^(?!with)[a-zA-Z]\\w*(\\.[a-zA-Z]\\w*)?$")) {
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -383,22 +403,46 @@ private ValidationResult validateDynamicKCLConfigProperty(final String subject,
                     .valid(false).build();
         }
 
-        final KinesisClientLibConfiguration kclTemp = new KinesisClientLibConfiguration("validate", "validate", null, "validate");
+        final ConfigsBuilder configsBuilderTemp = new ConfigsBuilder(
+                getStreamName(context),
+                getApplicationName(context),
+                KinesisAsyncClient.builder().region(Region.US_WEST_2).build(),
```
ChrisSamo632, line 409: Should we use `Region.of(context.getProperty(REGION).getValue())` if the property is set (presumably should be)?

At least add a comment here explaining that these builders aren't actually going ot be used beyond validation (to avoid future maintenance confusion)

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
ChrisSamo632: Is the !`with` prefix restriction still relevant with the updated Kinesis Client?

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -292,27 +316,23 @@ public class ConsumeKinesisStream extends AbstractKinesisStreamProcessor {
...
     private static final Set<Relationship> RELATIONSHIPS = Collections.singleton(REL_SUCCESS);
     private static final Set<Relationship> RECORD_RELATIONSHIPS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(REL_SUCCESS, REL_PARSE_FAILURE)));
 
     private static final PropertyUtilsBean PROPERTY_UTILS_BEAN;
     private static final BeanUtilsBean BEAN_UTILS_BEAN;
+
     static {
         PROPERTY_UTILS_BEAN = new PropertyUtilsBean();
-        PROPERTY_UTILS_BEAN.addBeanIntrospector(new FluentPropertyBeanIntrospector("with"));
+        PROPERTY_UTILS_BEAN.addBeanIntrospector(new FluentPropertyBeanIntrospector(""));
```
ChrisSamo632, line 335: If we no longer expet/have a `with` prefix, do we need the `BeanUtils` at all, or will straight up reflection work for the direct field names?

ChrisSamo632:
The default `logback.xml` should be updated for the `<logger name="com.amazonaws.services.kinesis" level="WARN" />` entry - does the new Kinesis client still kick out lots of noise by default? At the last, the package name has been changed so either this entry needs updating or the logging level entry should be removed

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -481,25 +525,25 @@ public void onScheduled(ProcessContext context) {
 
     @Override
     public void onTrigger(final ProcessContext context, final ProcessSessionFactory sessionFactory) {
-        if (worker == null) {
+        if (scheduler == null) {
             synchronized (WORKER_LOCK) {
```
ChrisSamo632, line 529: Should we change all `worker` references to `scheduler` for consistency?

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -530,106 +574,147 @@ public void stopConsuming(final ProcessContext context) {
...
+                }
+            } catch (final IllegalAccessException | InvocationTargetException e) {
+                throw new ProcessException(String.format("Unable to set Kinesis Client Configuration Builder property %s with value %s", StringUtils.capitalize(name), value), e);
+            }
+        });
+
+        getLogger().info("Kinesis Scheduler prepared for application {} to process stream {} as scheduler ID {}...",
+                getApplicationName(context), streamName, schedulerId);
+
+        final Scheduler scheduler = new Scheduler(
```
ChrisSamo632, line 645:
```suggestion
        return new Scheduler(
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -638,14 +723,14 @@ private boolean shutdownWorker(final ProcessContext context) {
                     getLogger().info("Kinesis Worker shutdown");
                 }
             }
-        } catch (InterruptedException | TimeoutException | ExecutionException e) {
+        } catch (final InterruptedException | TimeoutException | ExecutionException e) {
             getLogger().warn("Problem while shutting down Kinesis Worker: {}", e.getLocalizedMessage(), e);
             success = false;
         }
         return success;
     }
```
ChrisSamo632, line 732: `getDynamoDBOverride` is never used, can be deleted

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -530,106 +574,147 @@ public void stopConsuming(final ProcessContext context) {
...
      *  Developer note: if setting KCL configuration explicitly from processor properties, be sure to add them to the
      *  DISALLOWED_DYNAMIC_KCL_PROPERTIES list so Dynamic Properties can't be used to override the static properties
      */
-    KinesisClientLibConfiguration prepareKinesisClientLibConfiguration(final ProcessContext context, final String workerId) {
+    @VisibleForTesting
+    ConfigsBuilder prepareConfigsBuilder(final ProcessContext context, final String workerId, final ProcessSessionFactory sessionFactory) {
         @SuppressWarnings("deprecated")
-        final KinesisClientLibConfiguration kinesisClientLibConfiguration = new KinesisClientLibConfiguration(
-                getApplicationName(context),
+        final ConfigsBuilder configsBuilder = new ConfigsBuilder(
```
ChrisSamo632, line 687:
```suggestion
        return new ConfigsBuilder(
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -530,106 +574,147 @@ public void stopConsuming(final ProcessContext context) {
...
+    }
+
     /*
      *  Developer note: if setting KCL configuration explicitly from processor properties, be sure to add them to the
      *  DISALLOWED_DYNAMIC_KCL_PROPERTIES list so Dynamic Properties can't be used to override the static properties
      */
-    KinesisClientLibConfiguration prepareKinesisClientLibConfiguration(final ProcessContext context, final String workerId) {
+    @VisibleForTesting
+    ConfigsBuilder prepareConfigsBuilder(final ProcessContext context, final String workerId, final ProcessSessionFactory sessionFactory) {
         @SuppressWarnings("deprecated")
```
ChrisSamo632, line 686: Appears to no longer be deprecated in the new Kinesis Client

ChrisSamo632 (changes requested):
Thanks for the update @gresockj, things look generally good, but a few comments from me around code format and a few possible improvements/questions to consider

Stuff I've not done but would like to:
- [ ] run the integration tests
- [ ] build & test the changes (particularly `Consume`) against Kinesis - do you have an example test Flow to hand by any chance?

ChrisSamo632 (changes requested):
Thanks for the update @gresockj, things look generally good, but a few comments from me around code format and a few possible improvements/questions to consider

Stuff I've not done but would like to:
- [ ] run the integration tests
- [ ] build & test the changes (particularly `Consume`) against Kinesis - do you have an example test Flow to hand by any chance?

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/ITConsumeKinesisStreamEndpointOverride.java
```diff
@@ -49,16 +61,20 @@ public class ITConsumeKinesisStreamEndpointOverride extends ITConsumeKinesisStre
...
-                .withEndpointConfiguration(dynamoDBEndpointConfig)
-                .withCredentials(awsCredentialsProvider)
+        dynamoDB = DynamoDbClient.builder()
+                .credentialsProvider(awsCredentialsProvider)
+                .endpointOverride(URI.create(LOCAL_STACK_DYNAMODB_ENDPOINT_OVERRIDE))
+                .region(Region.of(REGION))
+                .httpClient(ApacheHttpClient.create())
                 .build();
 
         waitForKinesisToInitialize();
```
ChrisSamo632, line 80: `ACCESS_KEY` and `SECRET_KEY` properties are no longer valid for the processor, they need to be set via the `AWSCredentialsProviderControllerService` instead, e.g.

```java
        final AWSCredentialsProviderControllerService credentialsService = new AWSCredentialsProviderControllerService();
        runner.addControllerService("credentials-service", credentialsService);
        runner.setProperty(credentialsService, CredentialPropertyDescriptors.ACCESS_KEY, ACCESS_KEY);
        runner.setProperty(credentialsService, CredentialPropertyDescriptors.SECRET_KEY, SECRET_KEY);
        runner.enableControllerService(credentialsService);
        runner.setProperty(ConsumeKinesisStream.AWS_CREDENTIALS_PROVIDER_SERVICE, "credentials-service");
```

ChrisSamo632:
Trying to run the `Override` integration tests against LocalStack (having simply `localstack start -d`) and seeing the following in the logs, not sure whether this is a NiFi code, test config or Localstack issue (not yet been able to diagnose):

```java
[pool-3-thread-1] INFO org.apache.nifi.processors.aws.kinesis.stream.ConsumeKinesisStream - ConsumeKinesisStream[id=6a1123f6-69e0-455d-978c-19ae3b382626] Requesting Kinesis Worker shutdown
[ForkJoinPool.commonPool-worker-1] WARN software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable - Caught exception while requesting initial worker shutdown.
java.lang.NullPointerException: Cannot invoke "java.util.concurrent.ScheduledFuture.cancel(boolean)" because "this.takerFuture" is null
	at software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseCoordinator.stopLeaseTaker(DynamoDBLeaseCoordinator.java:332)
	at software.amazon.kinesis.coordinator.Scheduler.lambda$createWorkerShutdownCallable$14(Scheduler.java:765)
	at software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable.call(GracefulShutdownCoordinator.java:153)
	at software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable.call(GracefulShutdownCoordinator.java:39)
	at software.amazon.kinesis.coordinator.GracefulShutdownCoordinator.lambda$startGracefulShutdown$0(GracefulShutdownCoordinator.java:29)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1796)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
[pool-3-thread-1] WARN org.apache.nifi.processors.aws.kinesis.stream.ConsumeKinesisStream - ConsumeKinesisStream[id=6a1123f6-69e0-455d-978c-19ae3b382626] Problem while shutting down Kinesis Worker: java.lang.NullPointerException: Cannot invoke "java.util.concurrent.ScheduledFuture.cancel(boolean)" because "this.takerFuture" is null: java.util.concurrent.ExecutionException: java.lang.NullPointerException: Cannot invoke "java.util.concurrent.ScheduledFuture.cancel(boolean)" because "this.takerFuture" is null
[pool-3-thread-1] WARN org.apache.nifi.processors.aws.kinesis.stream.ConsumeKinesisStream - ConsumeKinesisStream[id=6a1123f6-69e0-455d-978c-19ae3b382626] One or more problems while shutting down Kinesis Worker, see logs for details
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/ITConsumeKinesisStream.java
```diff
@@ -102,7 +121,7 @@ private void startKCL(TestRunner runner, InitialPositionInStream initialPosition
     }
 
     @AfterEach
-    public void tearDown() throws InterruptedException {
+    public void tearDown() throws InterruptedException, ExecutionException {
```
ChrisSamo632, line 124: The `CBOR` env var [has changed](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-client-changes.html) and should be updated in all `ConsumeKinesis` ITs (although I can't recall why this was needed)

ChrisSamo632:
> I will take another look at the integration tests. Since they were already failing prior to this PR, I didn't put much effort into updating them, however it would probably still be valuable to get them working again.

Probably worth our looking at using Test Containers to run LocalStack for these integraiton tests, make them simpler for everyone

But that could be a follow-on ticket if too involved for this PR

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/ITConsumeKinesisStreamEndpointOverride.java
```diff
@@ -36,11 +41,18 @@ public class ITConsumeKinesisStreamEndpointOverride extends ITConsumeKinesisStre
...
 
-    private final AwsClientBuilder.EndpointConfiguration kinesisEndpointConfig =
-            new AwsClientBuilder.EndpointConfiguration(LOCAL_STACK_KINESIS_ENDPOINT_OVERRIDE, REGION);
+                @Override
+                public String secretAccessKey() {
+                    return SECRET_KEY;
+                }
+            });
 
     private final AwsClientBuilder.EndpointConfiguration dynamoDBEndpointConfig =
```
ChrisSamo632, line 57: No longer used/required with the updated `AmazonDynamoDBClient.builder`

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/firehose/AbstractKinesisFirehoseProcessor.java
```diff
@@ -20,17 +20,14 @@
...
-import com.amazonaws.auth.AWSCredentials;
-import com.amazonaws.auth.AWSCredentialsProvider;
-import com.amazonaws.services.kinesisfirehose.AmazonKinesisFirehoseClient;
+import org.apache.nifi.processors.aws.v2.AbstractAwsSyncProcessor;
+import software.amazon.awssdk.services.firehose.FirehoseClient;
+import software.amazon.awssdk.services.firehose.FirehoseClientBuilder;
 
 /**
- * This class provides processor the base class for kinesis firehose
+ * This class is the base class for kinesis  processors
```
turcsanyip, line 28: It is still true that `AbstractKinesisFirehoseProcessor` is the base class for the kinesis firehose processors (not all kinesis processors).
```suggestion
 * This class is the base class for Kinesis Firehose processors
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/AbstractKinesisAsyncStreamProcessor.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.nifi.processor.ProcessContext;
+import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.processors.aws.v2.AbstractAwsAsyncProcessor;
+import software.amazon.awssdk.services.kinesis.KinesisAsyncClient;
+import software.amazon.awssdk.services.kinesis.KinesisAsyncClientBuilder;
+
+/**
+ * This class is the base class for kinesis stream processors that use the async KinesisClient
+ */
+public abstract class AbstractKinesisAsyncStreamProcessor extends AbstractAwsAsyncProcessor<KinesisAsyncClient, KinesisAsyncClientBuilder> {
```
turcsanyip, line 29: We have "Kinesis Firehose" and "Kinesis Stream" processors. Putting (A)Sync between Kinesis and Stream sounds a bit strange to me (not the stream is sync/async). I would suggest `AbstractKinesisStreamAsyncProcessor` and `AbstractKinesisStreamSyncProcessor` instead.

turcsanyip (changes requested):
@gresockj Thanks for migrating the Kinesis processors to v2 client lib!
The change generally looks good to me, I also tested it with a simple flow.

We had issues with stopping worker threads using the v1 lib and v2 seems to have similar problems. The consumer client creates 20 `aws-java-sdk-NettyEventLoop-1-*` threads. Not sure we need that many, furthermore these threads remain alive when the processor is stopped. Could you please check if the thread count can be configurable and how to stop these threads?

When I stop the consume processor I get the following error:
```
2023-07-31 23:04:40,056 ERROR [prefetch-cache-shardId-000000000000-0000] s.a.k.r.polling.PrefetchRecordsPublisher turcsanyi-stream:shardId-000000000000 :  Unexpected exception was thrown. This could probably be an issue or a bug. Please search for the exception/error online to check what is going on. If the issue persists or is a recurring problem, feel free to open an issue on, https://github.com/awslabs/amazon-kinesis-client.
java.lang.IllegalStateException: Shutdown has been called on the cache, can't accept new requests.
```
Do you see the same?

Not related to this change: `AbstractAwsProcessor` has `client` and `region` fields but the `awsClientCache` is used instead. I think `client` / `region` should be removed. We got rid of those fields in the v1 abstract processor. I guess the client cache solution was ported from there.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-abstract-processors/src/main/java/org/apache/nifi/processors/aws/v2/AbstractAwsProcessor.java
```diff
@@ -192,6 +181,28 @@ public abstract class AbstractAwsProcessor<T extends SdkClient, U extends AwsSyn
...
+     *
+     * @see AbstractProcessor
+     */
+    @Override
+    public void onTrigger(final ProcessContext context, final ProcessSessionFactory sessionFactory) throws ProcessException {
+        final ProcessSession session = sessionFactory.createSession();
+        try {
+            onTrigger(context, session);
+            session.commitAsync();
+        } catch (final Throwable t) {
```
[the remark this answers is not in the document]
gresockj: I see your point, but this was derived from `AbstractProcessor#onTrigger`, which also catches `Throwable`, so I'm inclined to leave it as is.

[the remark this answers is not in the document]
ChrisSamo632: Don't like it (the `Throwable` handling in `AbstractProcessor`), but fair enough, makes sense to leave this as-is in that case

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
gresockj: No, but if someone is still trying to use the old methods, this could still help them catch the error.  None of the new methods will begin with "with" either.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
gresockj: BeanUtils still gives us the benefit of setting the value properly based on the type.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
gresockj: You mean like `workerId` and `workerState`?  I left these because the library still refers to them by the same name.

ChrisSamo632: Think I noted the same after further reviewing the code/changes/docs, so happy to not confuse things further than already may be the case with the AWS SDK itself

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
ChrisSamo632: That's fair, but is the `FluentPropertyBeanIntrospector` still required, or can we just use the `PROPERTY_UTILS_BEAN` instance without a `BeanIntrospector` being added (or is there a better alternative)?

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
ChrisSamo632: I guess that's a fair point, no object to the `(?!with)` prefix check staying in the RegEx

gresockj:
> @gresockj Thanks for migrating the Kinesis processors to v2 client lib! The change generally looks good to me, I also tested it with a simple flow.
> 
> We had issues with stopping worker threads using the v1 lib and v2 seems to have similar problems. The consumer client creates 20 `aws-java-sdk-NettyEventLoop-1-*` threads. Not sure we need that many, furthermore these threads remain alive when the processor is stopped. Could you please check if the thread count can be configurable and how to stop these threads?

Looking into this -- 
> 
> When I stop the consume processor I get the following error:
> 
> ```
> 2023-07-31 23:04:40,056 ERROR [prefetch-cache-shardId-000000000000-0000] s.a.k.r.polling.PrefetchRecordsPublisher turcsanyi-stream:shardId-000000000000 :  Unexpected exception was thrown. This could probably be an issue or a bug. Please search for the exception/error online to check what is going on. If the issue persists or is a recurring problem, feel free to open an issue on, https://github.com/awslabs/amazon-kinesis-client.
> java.lang.IllegalStateException: Shutdown has been called on the cache, can't accept new requests.
> ```
> 
> Do you see the same?



> @gresockj Thanks for migrating the Kinesis processors to v2 client lib! The change generally looks good to me, I also tested it with a simple flow.
> 
> We had issues with stopping worker threads using the v1 lib and v2 seems to have similar problems. The consumer client creates 20 `aws-java-sdk-NettyEventLoop-1-*` threads. Not sure we need that many, furthermore these threads remain alive when the processor is stopped. Could you please check if the thread count can be configurable and how to stop these threads?

It looks like the thread count is configurable, via the following code:

```
NettyNioAsyncHttpClient.builder()
            .eventLoopGroupBuilder(SdkEventLoopGroup.builder()
            .numberOfThreads(MAX_EVENT_LOOP_GROUP_THREADS))
```

However, when I do this, it seems to spawn a new thread loop group every time I stop/start a processor, and also for each new ConsumeKinesisStream processor I create, regardless of whether I cache the `SdkClient`.  On the other hand, when I leave the default `NettyNioAsyncHttpClient` builder in place, I get a constant 24 threads regardless of starting/stopping and adding new processors.  As it is, I'm inclined to leave the configuration as is, and perhaps this could be a future improvement if we can determine how to properly set the event loop group thread count.

> 
> When I stop the consume processor I get the following error:
> 
> ```
> 2023-07-31 23:04:40,056 ERROR [prefetch-cache-shardId-000000000000-0000] s.a.k.r.polling.PrefetchRecordsPublisher turcsanyi-stream:shardId-000000000000 :  Unexpected exception was thrown. This could probably be an issue or a bug. Please search for the exception/error online to check what is going on. If the issue persists or is a recurring problem, feel free to open an issue on, https://github.com/awslabs/amazon-kinesis-client.
> java.lang.IllegalStateException: Shutdown has been called on the cache, can't accept new requests.
> ```
> 
> Do you see the same?

Yes, unfortunately this appears to be an open issue in the client: https://github.com/awslabs/amazon-kinesis-client/issues/914

> 
> Not related to this change: `AbstractAwsProcessor` has `client` and `region` fields but the `awsClientCache` is used instead. I think `client` / `region` should be removed. We got rid of those fields in the v1 abstract processor. I guess the client cache solution was ported from there.

Done, good catch.

gresockj:
> I will take another look at the integration tests. Since they were already failing prior to this PR, I didn't put much effort into updating them, however it would probably still be valuable to get them working again.

I am inclined to leave the integration tests as they are for now, since they did not work prior to this PR.  Perhaps a follow-on issue could be to implement working integration tests.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/PutKinesisStream.java
```diff
@@ -126,72 +127,60 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
 
-        List<FlowFile> flowFiles = filterMessagesByMaxSize(session, batchSize, maxBufferSizeBytes, AWS_KINESIS_ERROR_MESSAGE);
+        final List<FlowFile> flowFiles = KinesisProcessorUtils.filterMessagesByMaxSize(session, batchSize, maxBufferSizeBytes, AWS_KINESIS_ERROR_MESSAGE, getLogger());
 
-        HashMap<String, List<FlowFile>> hashFlowFiles = new HashMap<>();
-        HashMap<String, List<PutRecordsRequestEntry>> recordHash = new HashMap<String, List<PutRecordsRequestEntry>>();
-
-        final AmazonKinesisClient client = getClient(context);
+        final HashMap<String, List<FlowFile>> hashFlowFiles = new HashMap<>();
+        final HashMap<String, List<PutRecordsRequestEntry>> recordHash = new HashMap<>();
```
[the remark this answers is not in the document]
ChrisSamo632: Missed this one, it's not a big deal to fix if there are no other changes though

ChrisSamo632:
> That warning will apparently be resolved in the next amazon-kinesis-client version: [awslabs/amazon-kinesis-client#745](https://github.com/awslabs/amazon-kinesis-client/issues/745)

Fix included in KCL 2.5.1, we're upgrading to 2.5.2 here 👍

ChrisSamo632:
> > I will take another look at the integration tests. Since they were already failing prior to this PR, I didn't put much effort into updating them, however it would probably still be valuable to get them working again.
> 
> I am inclined to leave the integration tests as they are for now, since they did not work prior to this PR. Perhaps a follow-on issue could be to implement working integration tests.

I agree it's not worth delaying this PR for the sake of the already flakey ITs. If one doesn't already exist, could you raise a Jira to investigate further/fix, maybe along with using TestContainers to launch LocalStack and run everything in a consistent environment?

ChrisSamo632 (approved):
LGTM @gresockj, thanks for persisting with the changes - just one very minor comment remaining for me, which I don't think is a blocker. Plus 1 follow-on ticket to raise for the integration tests.

I'd be keen to get @turcsanyip's re-review for this as well, he's been good at spotting issues with Kinesis for the processors in the past.

joewitt:
Seems like this is ready based on commentary and recent activity.  Anyone else have further input?

turcsanyip:
I would have a look at it and check my comments. Will do later today.

turcsanyip (approved):
Thanks @gresockj for your answers and review changes!

+1 LGTM

PR #7537, opened 2023-07-28
NIFI-11778 refactored Groovy tests in nifi-elasticsearch-restapi-processors to Java (and JUnit 5)

dan-s1:
# Summary

[NIFI-11778](https://issues.apache.org/jira/browse/NIFI-11778) / [NIFI-9290](https://issues.apache.org/jira/browse/NIFI-9290)
Refactored all the Groovy unit tests in nifi-elasticsearch-restapi-processors to Java.
Along with the refactor, the building of the individual JSON documents were removed and replaced with files which are now read in thereby reducing the size of the unit tests and increasing the readability of the unit tests.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation


dan-s1:
@exceptionfactory I see 2 of the builds failed but the errors are not related to the code I checked in.

ChrisSamo632:
@dan-s1 I re-triggered the Checks, definitely looks like ES related assertion failures now (originally you were right that the tests for for some unrelated transient issue)

```
Error:  Failures: 
Error:    PutElasticsearchRecordTest.testDateTimeFormatting:600 org.opentest4j.AssertionFailedError: expected: <2> but was: <0>
Error:    PutElasticsearchRecordTest.testRecordPathFeatures:337 org.opentest4j.AssertionFailedError: expected: <1> but was: <0>
Error:  Tests run: 124, Failures: 2, Errors: 0, Skipped: 0
Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0:test (default-test) on project nifi-elasticsearch-restapi-processors: There are test failures.
Error:  
Error:  Please refer to /home/runner/work/nifi/nifi/nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/target/surefire-reports for the individual test results.
```

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/AbstractPaginatedJsonQueryElasticsearchTest.java
```diff
@@ -0,0 +1,350 @@
...
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public abstract class AbstractPaginatedJsonQueryElasticsearchTest extends AbstractJsonQueryElasticsearchTest<AbstractPaginatedJsonQueryElasticsearch> {
+   private static final String TEST_DIR = "src/test/resources/AbstractPaginatedJsonQueryElasticsearchTest";
+    private static final String MATCH_ALL;
```
ChrisSamo632, line 42:
```suggestion
   private static final String MATCH_ALL;
```

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/AbstractPaginatedJsonQueryElasticsearchTest.java
```diff
@@ -0,0 +1,350 @@
...
+
+   static {
+       try {
+           MATCH_ALL = Files.readString(Paths.get("src/test/resources/common/matchAllQuery.json"));
+           MATCH_ALL_WITH_SORT_BY_MESSAGE = Files.readString(Paths.get(TEST_DIR, "matchAllWithSortByMessageQuery.json"));
+           MATCH_ALL_WITH_SORT_BY_MSG = Files.readString(Paths.get(TEST_DIR,"matchAllWithSortByMsgQueryWithoutSize.json"));
+       } catch (IOException e) {
+           throw new RuntimeException(e);
+       }
+   }
```
ChrisSamo632, line 54: Probably better to put these into a `@BeforeAll` method like they're loaded elsewhere, simply call the parent's `super.setUpBeforeClass()` to load common files too

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,782 @@
...
+
+        runner.run();
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
```
ChrisSamo632, line 345: It's probably worth taking this opportunity to split up these tests rather than keeping the really large `Test` scenarios that run the processor multiple times, with transfer state clearning in between - it would be easit to understand where failures are happening if we had everything in separate tests

Hopefully most of these can use a common `private` method that does the common setup/assertions, and simply set the test-specific properties/assertion logic

ChrisSamo632: N.B. there are several test methods in this class (and probably others) that should be considered for such refactoring - suggest searching for any instance of `clearTransferState` and considering whether the test(s) should be refactored

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
```diff
@@ -0,0 +1,243 @@
...
+
+        // third page - no hits
+        runOnce(runner);
+        AbstractJsonQueryElasticsearchTest.testCounts(runner, 0, 0, 0, 0);
+        assertTrue(runner.getStateManager().getState(Scope.LOCAL).toMap().isEmpty());
+        reset(runner);
+
+
+        // test hits splitting
+        runner.setProperty(AbstractPaginatedJsonQueryElasticsearch.SEARCH_RESULTS_SPLIT, ResultOutputStrategy.PER_HIT.getValue());
```
ChrisSamo632, line 175: These tests could probably be refactored - but be careful to retain the `clearTransferState` use between the "pages" where `SEARCH_RESULTS_SPLIT` remains the same (e.g. `PER_RESPONSE` above, `PER_HIT` here, and `PER_QUERY` later)

ChrisSamo632 (changes requested):
Thanks for the big refactor @dan-s1 - couple of code style/suggestions from me, plus worth considering breaking up some of the larger test cases that have grown to be somewhat unmaintainable over time - now seems a sensible opportunity to refactor those

Plus you've some tests failing in the Checks/when run locally that need fixing

Also, this is a duplicate of [NIFI-9290](https://issues.apache.org/jira/browse/NIFI-9290)

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/AbstractPaginatedJsonQueryElasticsearchTest.java
dan-s1: @ChrisSamo632 Just would like to clarify this. You want me to pull all of these up into `AbstractJsonQueryElasticsearchTest` `setUpBeforeClass` method and not just MATCH_ALL which seems to be already loaded by `AbstractJsonQueryElasticsearchTest`?

ChrisSamo632: Yes, exactly that - you're correct that `super` isn't a thing (my bad), but you should be able to refer to `AbstractJsonQueryElasticsearchTest. setUpBeforeClass()`

dan-s1:
> @dan-s1 I re-triggered the Checks, definitely looks like ES related assertion failures now (originally you were right that the tests for for some unrelated transient issue)
> 
> ```
> Error:  Failures: 
> Error:    PutElasticsearchRecordTest.testDateTimeFormatting:600 org.opentest4j.AssertionFailedError: expected: <2> but was: <0>
> Error:    PutElasticsearchRecordTest.testRecordPathFeatures:337 org.opentest4j.AssertionFailedError: expected: <1> but was: <0>
> Error:  Tests run: 124, Failures: 2, Errors: 0, Skipped: 0
> Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0:test (default-test) on project nifi-elasticsearch-restapi-processors: There are test failures.
> Error:  
> Error:  Please refer to /home/runner/work/nifi/nifi/nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/target/surefire-reports for the individual test results.
> ```

@ChrisSamo632 Do you have any suggestions how to reproduce these failures? On my local box all unit tests are passing. I thought it might be a locale issue and I tried setting the locale to "FR" but the unit test still passed.

dan-s1:
> 

NM I got it to fail by using this

`TimeZone.setDefault(TimeZone.getTimeZone("Europe/Paris"));`

dan-s1:
@ChrisSamo632 Would it be okay to fix the test errors by placing in `setUpBeforeClass`

`TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
`
?

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
dan-s1: @ChrisSamo632 So in test `testRecordPathFeatures` there are  8 `clearTransferState` calls. I understand you would like these split up into 9 separate unit tests. Any suggestions as to what these other tests should be called? It is not apparent to me what each of these scenarios are testing.

ChrisSamo632:
> @ChrisSamo632 Sorry by accident closed this PR and the reopened. Would it be okay to fix the test errors by placing in `setUpBeforeClass`
> 
> `TimeZone.setDefault(TimeZone.getTimeZone("UTC"));` ?

I'm not sure this sounds like the right way to fix this - the Checks run against multiple timezones for good reason, we want to be sure that processors can handle changes in locale. Having to set the tests to use a specific timezone to pass, seems like we might be hiding a potential issue either in the tests or the processor code itself.

Unfortunately, handling multiple timezones isn't my strong suite, although I'll try to take a look. I wonder whether @exceptionfactory or @markap14 might be able to provide some guidance on this one?

exceptionfactory:
> > @ChrisSamo632 Sorry by accident closed this PR and the reopened. Would it be okay to fix the test errors by placing in `setUpBeforeClass`
> > `TimeZone.setDefault(TimeZone.getTimeZone("UTC"));` ?
> 
> I'm not sure this sounds like the right way to fix this - the Checks run against multiple timezones for good reason, we want to be sure that processors can handle changes in locale. Having to set the tests to use a specific timezone to pass, seems like we might be hiding a potential issue either in the tests or the processor code itself.
> 
> Unfortunately, handling multiple timezones isn't my strong suite, although I'll try to take a look. I wonder whether @exceptionfactory or @markap14 might be able to provide some guidance on this one?

I will try to take a closer look at the tests soon, but I agree, changing the default timezone in test setup code is not the right approach.

All of the automated builds run with different timezones to ensure expected behavior in various locations, so all tests must pass without changing the system timezone.

In general, the issue means that there is some timezone conversion happening, which could point to a runtime issue, or simply an issue with the sample data used for testing. I'm not sure exactly why this has not surfaced earlier, so it points to an issue with the test-related changes.

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+
+    @Override
+    public Class<? extends AbstractPutElasticsearch> getTestProcessor() {
+        return PutElasticsearchRecord.class;
+    }
+
+    @BeforeAll
+    public static void setUpBeforeClass() throws Exception {
+        simpleSchema = Files.readString(Paths.get(TEST_DIR, "simpleSchema.json"));
+        TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
```
ChrisSamo632, line 85: Needs a closer look - this doesn't seem right

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
+
+       clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
```
ChrisSamo632, line 349: `testTimestampDateFormatAndScriptRecordPath`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
+
+        clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
```
ChrisSamo632, line 398: `testNullRecordPaths`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
+
+         clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
```
ChrisSamo632, line 427: `testIndexOperationRecordPath`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
+
+        clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
```
ChrisSamo632, line 456: `testIncompatibleTimestampRecordPath`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
+
+       clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
```
ChrisSamo632, line 473: `testNotFoundELRecordPaths`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
+
+        clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
```
ChrisSamo632, line 494: `testEmptyELRecordPaths`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
+
+        runner.enqueue(Paths.get(TEST_DIR, "8_flowFileContents.json"), Collections.singletonMap("schema.name", "recordPathTest"));
```
ChrisSamo632, line 517: `testInvalidDynamicTemplatesRecordPath`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+        MockFlowFile failure = runner.getFlowFilesForRelationship(PutElasticsearchRecord.REL_FAILURE).get(0);
+        failure.assertAttributeEquals("elasticsearch.put.error", String.format("Field referenced by %s must be Map-type compatible or a String parsable into a JSON Object", "/dynamic_templates"));
+
+        runner.clearTransferState();
+
+        clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
```
ChrisSamo632, line 530: `testRecordPathFieldDefaults`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+    }
+
+    @Test
+    public void testDateTimeFormatting() throws Exception{
```
ChrisSamo632, line 567: `testDefaultDateTimeFormatting`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+
+        runner.clearTransferState();
+
+         clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
```
ChrisSamo632, line 612: `testCustomDateTimeFormatting`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+        assertEquals(1,
+                runner.getProvenanceEvents().stream().filter(
+                    e -> ProvenanceEventType.SEND.equals(e.getEventType()) && "1 Elasticsearch _bulk operation batch(es) [2 error(s), 3 success(es)]".equals(e.getDetails())
+                ).count());
+
+        runner.clearTransferState();
+        runner.clearProvenanceEvents();
+
+        // errors still counted/logged even if not outputting to the error relationship
+        runner.removeProperty(PutElasticsearchRecord.RESULT_RECORD_WRITER);
```
ChrisSamo632, line 753: `testErrorsLoggedWithoutErrorRelationship`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,784 @@
...
+
+        assertEquals(1,
+                runner.getProvenanceEvents().stream().filter(e -> ProvenanceEventType.SEND.equals(e.getEventType())
+                        && "1 Elasticsearch _bulk operation batch(es) [1 error(s), 4 success(es)]".equals(e.getDetails())
+                ).count());
+
+        runner.clearTransferState();
+        runner.clearProvenanceEvents();
+
+        runner.setProperty(PutElasticsearchRecord.NOT_FOUND_IS_SUCCESSFUL, "false");
```
ChrisSamo632, line 729: `testDocumentNotFoundErrorHandling`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
ChrisSamo632: I've added some suggestions to break up this Test Case, but also for others in this class (i.e. `testDateTimeFormatting` and `testErrorRelationship`)

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
dan-s1: @exceptionfactory I do not know why this did not surface earlier. I did not make any changes to assertion code. I cut and pasted it from the previous Groovy closure to a Java Consumer.

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJsonTest.java
```diff
@@ -0,0 +1,409 @@
...
+            assertEquals(1L, scriptCount);
+            assertEquals(1L, falseScriptedUpsertCount);
+            assertEquals(1L, dynamicTemplatesCount);
+        };
+        basicTest(0, 0, 1, consumer, null);
+
+        runner.clearTransferState();
+        runner.clearProvenanceEvents();
+
+        runner.setProperty(PutElasticsearchJson.INDEX_OP, IndexOperationRequest.Operation.Upsert.getValue().toLowerCase());
```
ChrisSamo632, line 244: `simpleTestWithScriptedUpsert`

ChrisSamo632 (changes requested):
Have suggested some new test case names as requested @dan-s1

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
ChrisSamo632: I wonder whether this is due to the Groovy tests having converted, for example, a `LocalDateTime` into a `Timestamp`, to an `Instant`, and finally the `long` representation of the "epoch millis" - this would (I think) cater for the Locale/TimeZone change in the JVM, e.g. a `Europe/Paris` TZ instead of `UTC`, allowing the tests to use a consistent basis for creating, using, and then asserting the timestamp/data/time

With the change to a JSON-based structure for the FlowFile contents, are the timestamps being read from the file in one TZ then the tests and trying to use/assert them in another?

I recall having to spend some time (no pun intended) on getting these tests to work originally, but it was a while ago now, so can't recall the specifics (think I maybe tried to start with a hardcoded timestamp-millis, but couldn't figure out how to get that to work)

dan-s1: @ChrisSamo632 @exceptionfactory This information helped a lot. I tweeked my code to modify the read in JSON files with the same values you had used with 
`Time.valueOf(LOCAL_TIME).getTime()`
`Timestamp.valueOf(LOCAL_DATE_TIME).toInstant().toEpochMilli())`
`Date.valueOf(LOCAL_DATE).getTime())`

Now even with changing the timezone with
`TimeZone.setDefault(TimeZone.getTimeZone("Europe/Paris"));`
the tests pass.

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
dan-s1: @ChrisSamo632 Trying to clarify is this the name for the second or the first test ?

ChrisSamo632: A suggested name for everything between here and the next `runner.clearTransferState()`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
dan-s1: `testPagination` is not an actual unit test. It is an overridden method. It is called in the parent's `AbstractPaginatedJsonQueryElasticsearchTest` unit tests

1. `testScroll`
2. `testPit`
3. `testSearchAfter`

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
dan-s1: I got it. I was able to refactor all the unit tests as you had suggested.

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJsonTest.java
dan-s1: Should line 260-266 be its own unit test? If so what should it be named? Perhaps simpleTestWithNonJsonScriptedUpsert.

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
ChrisSamo632: You're right, I wonder whether these assertions could be re-written in a way similar to those in `PutElasticsearchJsonTest`, which also overrides the same method?

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJsonTest.java
ChrisSamo632: Sounds like a sensible refactor, maybe `testInvalidScript` or `testNonJsonScript`?

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
dan-s1: I do not see a method named `testPagination` in `PutElasticsearchJsonTest`

ChrisSamo632: `SearchElssticsearch` (use your IDE's functionality to check for any other users of the method)

dan-s1: `testPagination` is only implemented by `PaginatedJsonQueryElasticsearchTest` and `SearchElasticsearchTest` and it is called by `AbstractPaginatedJsonQueryElasticsearchTest` three times. Can you please elaborate on what you want? I do not fully understand the code and exactly how each of these test classes are exercising their features. Thanks!

dan-s1: I added a `validatePagination` method which greatly simplified the `testPagination` method. Please let me know if that is what you had in mind.

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,850 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 1);
+
+        final String errorResponses = runner.getFlowFilesForRelationship(PutElasticsearchJson.REL_ERROR_RESPONSES).get(0).getContent();
+        //assertThat(errorResponses, containsString("not_found"))
+        assertTrue(errorResponses.contains("not_found"));
+        //assertThat(errorResponses, containsString("For input string: 20abc"))
```
ChrisSamo632, line 818:
```suggestion
        assertTrue(errorResponses.contains("not_found"));
```

Remove unnecessary commented out code

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,850 @@
...
+        runner.getFlowFilesForRelationship(PutElasticsearchRecord.REL_FAILED_RECORDS).get(0).assertAttributeEquals(PutElasticsearchRecord.ATTR_RECORD_COUNT, "1");
+        runner.getFlowFilesForRelationship(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS).get(0).assertAttributeEquals(PutElasticsearchRecord.ATTR_RECORD_COUNT, "4");
+
+        assertEquals(1,
+                runner.getProvenanceEvents().stream().filter(e -> ProvenanceEventType.SEND.equals(e.getEventType())
+                        && "1 Elasticsearch _bulk operation batch(es) [1 error(s), 4 success(es)]".equals(e.getDetails())).count());
+    }
+
+    @Test
+    public void testErrorsLoggedWithoutErrorRelationship() throws Exception {
```
ChrisSamo632, line 769: It would be better to rationalise these test cases and just have the differences present in individual test cases, e.g.

```java
private void testErrors(final int failedRecordCount, final int successfulRecordCount, final int errorResponseCount) throws Exception {
        final JsonRecordSetWriter writer = new JsonRecordSetWriter();
        runner.addControllerService("writer", writer);
        runner.setProperty(writer, SchemaAccessUtils.SCHEMA_ACCESS_STRATEGY, SchemaAccessUtils.SCHEMA_NAME_PROPERTY);
        runner.setProperty(writer, SchemaAccessUtils.SCHEMA_REGISTRY, "registry");
        runner.enableControllerService(writer);
        clientService.setResponse(IndexOperationResponse.fromJsonResponse(Files.readString(Paths.get(TEST_COMMON_DIR, "sampleErrorResponse.json"))));
        registry.addSchema(ERROR_TEST_SCHEMA, errorTestSchema);
        runner.setProperty(PutElasticsearchRecord.RESULT_RECORD_WRITER, "writer");

        runner.assertValid();

        runner.enqueue(VALUES, Collections.singletonMap("schema.name", ERROR_TEST_SCHEMA));
        runner.run();

        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, failedRecordCount > 0 ? 1 : 0);
        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, successfulRecordCount > 0 ? 1 : 0);
        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, errorResponseCount > 0 ? 1 : 0);

        if (failedRecordCount > 0) {
            runner.getFlowFilesForRelationship(PutElasticsearchRecord.REL_FAILED_RECORDS).get(0).assertAttributeEquals(PutElasticsearchRecord.ATTR_RECORD_COUNT, String.valueOf(failedRecordCount));
        }
        if (successfulRecordCount > 0) {
            runner.getFlowFilesForRelationship(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS).get(0).assertAttributeEquals(PutElasticsearchRecord.ATTR_RECORD_COUNT, String.valueOf(successfulRecordCount));
        }
        if (errorResponseCount > 0) {
            final String errorResponses = runner.getFlowFilesForRelationship(PutElasticsearchJson.REL_ERROR_RESPONSES).get(0).getContent();
            assertTrue(errorResponses.contains("not_found"));
            assertTrue(errorResponses.contains("For input string: 20abc"));
        }

        assertEquals(1,
                runner.getProvenanceEvents().stream().filter(e -> ProvenanceEventType.SEND.equals(e.getEventType())
                        && "1 Elasticsearch _bulk operation batch(es) [%d error(s), %d success(es)]".formatted(failedRecordCount, successfulRecordCount).equals(e.getDetails())).count());
    }

    @Test
    public void testErrorRelationship() throws Exception {
        runner.setProperty(PutElasticsearchRecord.NOT_FOUND_IS_SUCCESSFUL, "true");
        runner.setProperty(PutElasticsearchRecord.LOG_ERROR_RESPONSES, "true");
        testErrors(1, 4, 0);
    }

    @Test
    public void testErrorsLoggedWithoutErrorRelationship() throws Exception {
        runner.setProperty(PutElasticsearchRecord.NOT_FOUND_IS_SUCCESSFUL, "false");
        testErrors(2, 3, 0);
    }

    @Test
    public void testDocumentNotFoundErrorHandling() throws Exception {
        runner.setProperty(PutElasticsearchRecord.NOT_FOUND_IS_SUCCESSFUL, "false");
        runner.setProperty(PutElasticsearchRecord.OUTPUT_ERROR_RESPONSES, "true");
        testErrors(2, 3, 1);
    }
```

This will reduce duplicated code across a few of the test cases

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,850 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESS, 1);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILURE, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+    }
+
+    @Test
+    public void testEmptyELRecordPaths() throws Exception {
```
ChrisSamo632, line 530: Lot of duplicated code with `testNotFoundELRecordPaths`, should we combine these into a common `testELRecordPaths` that each of these test cases calls, setting any specfic properties, etc.?

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,850 @@
...
+            assertEquals(2, headerFieldsCount);
+            assertEquals(1, routingCount);
+            assertEquals(1, versionCount);
+        };
+
+        basicTest(0, 0, 1, consumer);
+    }
+
+    @Test
+    public void simpleTestWithRequestParametersAndBulkHeadersFlowFileEL() {
```
ChrisSamo632, line 229: Pretty much the same as `simpleTestWithRequestParametersAndBulkHeaders`, consider creating a common `testWithRequestParametersAndBulkHeaders`, then have these test cases set test specific properties, etc. (think the main difference is that one passes a `Map` of EL values)

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
ChrisSamo632: `SearchElasticsearchTest` uses `null` as one of the `ResultOutputStrategy` values, which defaults to `PER_RESPONSE` in `AbstractJsonQueryElasticsearch`, i.e. all `ResponseOutputStrategy` values are exercised in both test classes - this is probably something we should rationalise though, either use the `ResultOutputStrategy.values()` array as you've done in `PaginatedJsonQueryElasticsearchTest`, or create the array like in `SearchElasticsearchTest` with `null` - the danger with the latter is that it assumes the default will never change (which would be a breaking change, so likely wouldn't, but really it's a dangerous assumtion to have been made in the test code, so using the `.values()` approach is probably better really, so let's be consistent)

What you've done is certainly along the right lines though @dan-s1 - taking another look, I think most of the test setup & validation that's in `SearchElasticsearchTest` should probably be the same in `PaginatedJsonQueryElasticsearchTest` (where it's not already), how about:

* change the `AbstractPaginatedJsonQueryElasticsearchTest#testPagination` to **not** be `abstract` but instead provide the combined implementation from `SearchElasticsearchTest` & `PaginatedJsonQueryElasticsearchTest`
  * create runner, set pagination type (N.B. you've missed this completely from the new `PaginatedJsonQueryElasticsearchTest#testPagination` implementation), set query, iterate through all `ResultOutputStrategy.values()`
  * iterate over 3 pages in a loop, alernately setting `OUTPUT_NO_HITS` to `true` or `false` (or only a single iteration if it's ` PER_QUERY` test)
* call `validatePagination`, which looks like it should be the same/very similar to that already in `SearchElasticsearchTest`
* make `assertState` and `abstract` method, then implement in each test class
  * for `SearchElasticsearchTest`, this probably stays about the same, although we'll want to retain the `assertTrue(runner.getStateManager().getState(Scope.LOCAL).toMap().isEmpty());` entries where appropriate from `validatePagination` (and add in a call to `assertState` in their place)
  * for `PaginatedJsonQueryElasticsearchTest`, this is probably just always an `assertTrue(runner.getStateManager().assertStateNotSet();` check or such

Hope that makes sense

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
```diff
@@ -0,0 +1,850 @@
...
+        runner.assertTransferCount(PutElasticsearchRecord.REL_RETRY, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_FAILED_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_SUCCESSFUL_RECORDS, 0);
+        runner.assertTransferCount(PutElasticsearchRecord.REL_ERROR_RESPONSES, 0);
+    }
+
+    @Test
+    public void testCustomDateTimeFormatting() throws Exception {
+        clientService.setEvalConsumer((List<IndexOperationRequest> items) -> {
+            String timestampOutput = LOCAL_DATE_TIME.format(DateTimeFormatter.ofPattern("yy MMM D H"));
```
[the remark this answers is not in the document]
dan-s1: With that change I get an assertion error
```
java.lang.AssertionError: org.opentest4j.AssertionFailedError: Did not get expected timestamp count ==> 
Expected :2
Actual   :0
```

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java
dan-s1: I do not see the following block currently in `testDocumentNotFoundErrorHandling`

 ```
final JsonRecordSetWriter writer = new JsonRecordSetWriter();
        runner.addControllerService("writer", writer);
        runner.setProperty(writer, SchemaAccessUtils.SCHEMA_ACCESS_STRATEGY, SchemaAccessUtils.SCHEMA_NAME_PROPERTY);
        runner.setProperty(writer, SchemaAccessUtils.SCHEMA_REGISTRY, "registry");
        runner.enableControllerService(writer);
        clientService.setResponse(IndexOperationResponse.fromJsonResponse(Files.readString(Paths.get(TEST_COMMON_DIR, "sampleErrorResponse.json"))));
        registry.addSchema(ERROR_TEST_SCHEMA, errorTestSchema);
        runner.setProperty(PutElasticsearchRecord.RESULT_RECORD_WRITER, "writer");
```

can it still be used for it?

Comment (dan-s1) [reply nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java:769]: NM I see that it does not make a difference

File: nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java (hunk 21 above)
Comment (dan-s1) [reply nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecordTest.java:530]: Done

File: nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java (hunk 4 above)
Comment (dan-s1) [reply nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java]: Sort of. I pulled up `testPagination`, `validatePagination` and `assertState` as is from `SearchElasticsearchTest` to `AbstractPaginatedJsonQueryElasticsearchTest` and that works for both `SearchElasticsearchTest` and `PaginatedJsonQueryElasticsearchTest`. Not sure what you mean with the last bullet with making assertState abstract as the method in `SearchElasticsearchTest` is already being used in multiple methods. Will changing it adversely effect the other methods which use it?

Comment (dan-s1) [reply nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java]: One second it did not work as I said, I forgot to remove `testPagination` from `PaginatedJsonQueryElasticsearchTest`. When I do then I have 3 failures for `testPit`, `testScroll` and `testSearchAfter`. They all seem to error in `testCounts` on
 `runner.assertTransferCount(AbstractJsonQueryElasticsearch.REL_ORIGINAL, original);`

where expected was 0 but actual was 1

Comment (dan-s1) [reply nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java]: `validatePagination` from `SearchElasticsearchTest` as is will not work for `PaginatedJsonQueryElasticsearchTest`
as there are differences in the  call to `AbstractJsonQueryElasticsearchTest.testCounts` in the `validatePagination` of `SearchElasticsearchTest` and `PaginatedJsonQueryElasticsearchTest`

**Comment by dan-s1** [comment]
@ChrisSamo632 Checking in to see if my latest changes are okay. Thanks!

**Comment by ChrisSamo632** [comment]
> @ChrisSamo632 Checking in to see if my latest changes are okay. Thanks!

@dan-s1 I've been away, but hoping to take a look at your latest changes soon/this week.

Note that you've now got some conflicts after #7441 was recently merged - couple of new tests to migrate over for the updated processors

**Comment by dan-s1** [comment]
@ChrisSamo632 Do I manually cut and paste those changes as in my branch those Groovy files have been deleted?

**Comment by ChrisSamo632** [comment]
> @ChrisSamo632 Do I manually cut and paste those changes to my branch since those Groovy files have been deleted in my branch?

I'd guess so, and make them work for the Java setup rather than Groovy

**Comment by dan-s1** [comment]
@exceptionfactory Just double checking before I proceed that the way to resolve the conflicts is for me to cut and paste the changes from [7441](https://github.com/apache/nifi/pull/7441) to my code. Thanks!

File: nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/AbstractPaginatedJsonQueryElasticsearchTest.java [hunk 24]
```diff
@@ -0,0 +1,348 @@
...
+    public void testPit() throws Exception {
+        testPagination(PaginationType.POINT_IN_TIME);
+    }
+
+    @Test
+    public void testSearchAfter() throws Exception {
+        testPagination(PaginationType.SEARCH_AFTER);
+    }
+
+    public abstract void testPagination(final PaginationType paginationType) throws Exception;
```
ChrisSamo632, line 301:
```suggestion
    private void testPagination(final PaginationType paginationType) throws Exception {
        final TestRunner runner = createRunner(false);
        final TestElasticsearchClientService service = AbstractJsonQueryElasticsearchTest.getService(runner);
        service.setMaxPages(2);
        runner.setProperty(AbstractPaginatedJsonQueryElasticsearch.PAGINATION_TYPE, paginationType.getValue());
        runner.setProperty(AbstractJsonQueryElasticsearch.QUERY, matchAllWithSortByMsgWithSizeQuery);

        // Tests flowfile per page, hits splitting and hits combined
        for (final ResultOutputStrategy resultOutputStrategy : ResultOutputStrategy.values()) {
            runner.setProperty(AbstractPaginatedJsonQueryElasticsearch.SEARCH_RESULTS_SPLIT, resultOutputStrategy.getValue());

            for (int iteration = 1; iteration < 4; iteration++) {
                // Check that changing OUTPUT_NO_HITS doesn't have any adverse effects on pagination
                runner.setProperty(AbstractJsonQueryElasticsearch.OUTPUT_NO_HITS, String.valueOf(iteration % 2 > 0).toLowerCase());

                runOnce(runner);
                validatePagination(runner, resultOutputStrategy, paginationType, iteration);

                if (ResultOutputStrategy.PER_QUERY.equals(resultOutputStrategy)) {
                    break;
                }

                runner.clearTransferState();
                if (!isStateUsed()) {
                    // reset PaginatedJsonQueryElasticsearch to re-run the query with different OUTPUT_NO_HITS setting
                    reset(runner);
                }
            }
            reset(runner);
        }
    }

    abstract void validatePagination(final TestRunner runner, final ResultOutputStrategy resultOutputStrategy, final PaginationType paginationType, int iteration) throws IOException;
```

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
```diff
@@ -0,0 +1,217 @@
...
+        testPaginationExpiration(PaginationType.SCROLL);
+    }
+
+    @Test
+    public void testPitExpiration() throws Exception {
+        testPaginationExpiration(PaginationType.POINT_IN_TIME);
+    }
+
+    @Test
+    public void testSearchAfterExpiration() throws Exception {
```
ChrisSamo632, line 85:
```suggestion
   void testScrollExpiration() throws Exception {
        testPaginationExpiration(PaginationType.SCROLL);
    }

    @Test
   void testPitExpiration() throws Exception {
        testPaginationExpiration(PaginationType.POINT_IN_TIME);
    }

    @Test
    void testSearchAfterExpiration() throws Exception {
```

No need for the `public` modifier on this test cases when using JUnit 5

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
```diff
@@ -0,0 +1,217 @@
...
+            AbstractJsonQueryElasticsearchTest.testCounts(runner, 0, 1, 0, 0);
+            runner.getFlowFilesForRelationship(AbstractJsonQueryElasticsearch.REL_HITS).get(0).assertAttributeEquals("hit.count", "20");
+            // the "last" page.number is used, so 2 here because there were 2 pages of hits
+            runner.getFlowFilesForRelationship(AbstractJsonQueryElasticsearch.REL_HITS).get(0).assertAttributeEquals("page.number", "2");
+            assertEquals(20, runner.getFlowFilesForRelationship(AbstractJsonQueryElasticsearch.REL_HITS).get(0).getContent().split("\n").length);
+            assertTrue(runner.getStateManager().getState(Scope.LOCAL).toMap().isEmpty());
+        }
+    }
+
+    private static void assertState(final MockStateManager stateManager, final PaginationType paginationType, final int hitCount, final int pageCount) throws IOException {
```
ChrisSamo632, line 190:
```suggestion
    private void assertState(final MockStateManager stateManager, final PaginationType paginationType, final int hitCount, final int pageCount) throws IOException {
```

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PaginatedJsonQueryElasticsearchTest.java
```diff
@@ -0,0 +1,104 @@
...
+            reset(runner);
+            if(ResultOutputStrategy.PER_QUERY.equals(resultOutputStrategy)) {
+                break;
+            }
+        }
+    }
+
+    private static void validatePagination(final TestRunner runner, final ResultOutputStrategy resultOutputStrategy) {
+        runner.getStateManager().assertStateNotSet();
+        switch (resultOutputStrategy) {
```
ChrisSamo632, line 73:
```suggestion
    @Override
    void validatePagination(final TestRunner runner, final ResultOutputStrategy resultOutputStrategy, final PaginationType paginationType, int iteration) {
        runner.getStateManager().assertStateNotSet();

        switch (resultOutputStrategy) {
```

ChrisSamo632 (commented):
@dan-s1 had a go at more completely suggesting the changes I was thinking about previously to rationalise the `testPagination` and related methods between the `AbstractPaginatedJsonQueryElasticsearchTest` and child classes

This, hopefully, shows that these two classes should behave much the same and make it easier to maintain these tests in the future

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
dan-s1: I realize there is no need for the `public`  modifier but I seemed to have this conversation with @exceptionfactory  in an earlier ticket (cannot find right now) whether to drop `public` or not and he was in favor of keeping it and having a follow on ticket which would handle removing the `public` modifier from all the unit tests. 
@exceptionfactory  please let me know whether you are still in favor of keeping the `public` modifier or can we remove it? Thanks!

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
```diff
@@ -0,0 +1,217 @@
...
+                runner.clearTransferState();
+                if(ResultOutputStrategy.PER_QUERY.equals(resultOutputStrategy)) {
+                    break;
+                }
+            }
+            reset(runner);
+        }
+    }
+
+    private static void validatePagination(final TestRunner runner, final ResultOutputStrategy resultOutputStrategy, final PaginationType paginationType, int iteration) throws IOException {
```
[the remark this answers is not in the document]
dan-s1: @ChrisSamo632 I am trying to implement your suggestions. I pulled up `validatePagination` to `AbstractPaginatedJsonQueryElasticsearchTest` and overrode it in `SearchElasticsearchTest` and `PaginatedJsonQueryElasticsearchTest`. THe problem I am having is with `testPagination`.  I pulled the one up from `SearchElasticsearchTest` to `AbstractPaginatedJsonQueryElasticsearchTest`.  All of the tests in `SearchElasticsearchTest`pass but tests `testScroll` and `testSearchAfter` in `PaginatedJsonQueryElasticsearchTest` both fail at `AbstractJsonQueryElasticsearchTest#testCounts` line 387 

```
org.opentest4j.AssertionFailedError: 
Expected :20
Actual   :0
```
Please advise.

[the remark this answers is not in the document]
ChrisSamo632: The code I pasted in my suggestion (https://github.com/apache/nifi/pull/7537#discussion_r1304789339) worked for me when running locally last night, which is why I suggested it (don't copy `testPagination` from either of the classes as they are, neither will work fully for the other class - I've combined the two to get them working using the same logic (including some things that had been missed from one or the other test class previously)

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
ChrisSamo632: Personally I'd suggest just removing these here as you're already modifying the entire file/class anyway (and the warnings in IntelliJ are annoying, but one can live with those if there's a plan to wholesale change all `public` accessors in the future)

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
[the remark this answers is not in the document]
dan-s1: @ChrisSamo632 Thank you! I am sorry for completely overlooking that.

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/SearchElasticsearchTest.java
dan-s1: @ChrisSamo632 Should I then remove public the unit tests in all the classes that I have refactored?

ChrisSamo632: @dan-s1 that's probably the approach I'd take, although I think these were the only methods I saw the warning for last night (but I didn't click through all the migrated classes, just the three where `testPagination` is being looked at)

but again, I'm happy to accept this staying as-is for now if there's a plan to remove them in the near future

dan-s1:
@ChrisSamo632 I am trying to implement the changes from [#7441](https://github.com/apache/nifi/pull/7441) For `testBatchingAndErrorRelationship` in `PutElasticsearchJsonTest.java`  for the equivalent line of  `runner.assertTransferCount(PutElasticsearchJson.REL_SUCCESS, 4)`  is failing as the transfer count is 6 and not 4. Please advise as I do not see any changes in `PutElasticsearchJson`.

dan-s1:
@ChrisSamo632 NM I forgot to account for the changes to `src/test/resources/common/sampleErrorResponse.json`

dan-s1:
@ChrisSamo632 I have made the changes you had requested. I also incorporated the changes from https://github.com/apache/nifi/pull/744. I broke up `PutElasticsearchRecordTest#testErrorRelationship` into five separate tests 

1. testFailedRecordsOutput
2. testFailedRecordsOutputGroupedByErrorType
3. testNotFoundResponsesTreatedAsFailedRecords
4. testNotFoundFailedRecordsGroupedAsErrorType
5. testErrorsLoggedWithoutErrorRelationship

and I refactored out the common code into the method `testErrorRelationship`

ChrisSamo632:
@dan-s1 I'll take a look shortly hopefully, but note that you've still got some conflicts with `main`, you'll need to rebase against (or merge in) `main`

dan-s1:
@ChrisSamo632 I am not sure I can do anything regarding these conflicts. I already incorporated the changes of https://github.com/apache/nifi/pull/7441 for `PutElasticsearchRecord` and other related changes.  As for the groovy files, naturally there will be a conflict as my branch deleted them and replaced them with Java files.

exceptionfactory:
> @ChrisSamo632 I am not sure I can do anything regarding these conflicts. I already incorporated the changes of #7441 for `PutElasticsearchRecord` and other related changes. As for the groovy files, naturally there will be a conflict as my branch deleted them and replaced them with Java files.

@dan-s1 You will need to rebase your PR and then force-push to this feature branch to resolve the conflicts. If you have already incorporated the changes, then during the rebase process, you should be able to just mark the `groovy` files as deleted, but the change to `PutElasticsearchRecord` needs to be incorporated.

dan-s1:
@exceptionfactory Thanks for the clarification. I have made those changes and pushed.

ChrisSamo632 (approved):
LGTM, thanks for continuing with the updates @dan-s1; provided the checks pass (these tests all work for me locally at least), I'm happy to merge

PR #7584, opened 2023-08-08
NIFI-11921: populate system load average in HB

KalmanJantner:
# Summary

[NIFI-11921](https://issues.apache.org/jira/browse/NIFI-11921)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/c2/c2-client-bundle/c2-client-service/src/main/java/org/apache/nifi/c2/client/service/C2HeartbeatFactory.java
+++ b/c2/c2-client-bundle/c2-client-service/src/main/java/org/apache/nifi/c2/client/service/C2HeartbeatFactory.java
@@ -228,7 +228,7 @@ public class C2HeartbeatFactory {
 
         // getSystemLoadAverage is not available in Windows, so we need to prevent to send invalid data
         if (systemLoadAverage >= 0) {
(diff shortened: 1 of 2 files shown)


exceptionfactory (commented):
Thanks for the contribution @KalmanJantner! Does this change effectively deprecate the `cpuUtilization` property, or is that set somewhere else? If it is not longer set, then that should be noted in the API property annotation details.

KalmanJantner:
> Thanks for the contribution @KalmanJantner! Does this change effectively deprecate the `cpuUtilization` property, or is that set somewhere else? If it is not longer set, then that should be noted in the API property annotation details.

@exceptionfactory thank you for the quick feedback. We will keep the `cpuUtilization` as cpp agent able to provide such information. Only JVM not fully support getting CPU Utilization (https://www.oracle.com/java/technologies/faq-sun-packages.html) only `SystemLoadAverage` is supported. So this change is about to extend what kind of CPU related information we can provide via HB and populate as much as we can in a given agent on the given platform.

exceptionfactory (approved):
Thanks for the background explanation @KalmanJantner. Retaining the existing property and populating the new load average property makes sense under those circumstances. +1 merging

PR #7591, opened 2023-08-10
NIFI-11874 update Process Group configuration layout in the UI

markobean:
# Summary

[NIFI-11874](https://issues.apache.org/jira/browse/NIFI-11874)

UI modification for the layout of Process Group configuration. It separates the increasing number of configuration options into two columns. Also, to assist in splitting configuration input into logically related pieces as well as be consistent with the configuration of other components, the Comments section was moved to its own tab. This presents better in the UI and keeps the Apply button easily accessible without the need to scroll (for most display resolutions.)

The PR includes additional details when mouse-over/hover the new "?" icon for several of the Process Group properties to explicitly clarify that the default connections settings apply only to new connections - not existing connections. 

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

Installed NiFi on several different platforms each having a different display size and resolution. The results presented clearly in all cases tested.

### Build

  - [X] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
@@ -81,7 +81,7 @@
 }
 
 #general-process-group-configuration div.settings-columns .settings-column {
-    margin-right: 20px;
+    margin-right: 50px;
 }
 
 #general-process-group-configuration input, #general-process-group-configuration textarea {
@@ -95,7 +95,6 @@
 }
 
 #parameter-contexts-recursive-container {
-    float: left;
     margin-left: 10px;
 }
 
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
@@ -132,7 +132,8 @@
 }
 
 #process-group-comments {
-    height: 300px;
+    height: 100px;
+    width: 750px;
 }
 
 #read-only-process-group-comments {
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
@@ -91,7 +91,6 @@
 
 #process-group-parameter-context-combo {
     width: 328px;
-    float: left;
 }
 
 #parameter-contexts-recursive-container {
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
@@ -76,6 +76,14 @@
 
 /* general */
 
+#general-process-group-configuration div.settings-columns {
+    display: flex;
+}
+
+#general-process-group-configuration div.settings-columns .settings-column {
+    margin-right: 20px;
+}
+
 #general-process-group-configuration input, #general-process-group-configuration textarea {
     width: 350px;
     vertical-align: middle;
@@ -125,7 +133,7 @@
 }
 
 #process-group-comments {
-    height: 100px;
+    height: 300px;
 }
 
 #read-only-process-group-comments {
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/process-group-configuration.jsp
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/process-group-configuration.jsp
@@ -104,7 +104,7 @@
                         <div class="settings-column">
                             <div class="setting">
                                 <div class="setting-name">Default FlowFile Expiration
-                                    <div class="fa fa-question-circle" alt="Info" title="This setting applies as the default for any new connections. Configuration of existing connections are unaffected."></div>
+                                    <div class="fa fa-question-circle" alt="Info" title="This setting applies as the default FLowFile expiration for any new connection. Configuration of existing connections is unaffected."></div>
                                 </div>
                                 <div class="editable setting-field">
                                     <input type="text" id="process-group-default-flowfile-expiration" class="setting-input"/>
@@ -115,7 +115,7 @@
                             </div>
                             <div class="setting">
                                 <div class="setting-name">Default Back Pressure Object Threshold
-                                    <div class="fa fa-question-circle" alt="Info" title="This setting applies as the default for any new connections. Configuration of existing connections are unaffected."></div>
+                                    <div class="fa fa-question-circle" alt="Info" title="This setting applies as the default back pressure object count for any new connection. Configuration of existing connections is unaffected."></div>
                                 </div>
                                 <div class="editable setting-field">
                                     <input type="text" id="process-group-default-back-pressure-object-threshold" class="setting-input"/>
@@ -126,7 +126,7 @@
                             </div>
                             <div class="setting">
                                 <div class="setting-name">Default Back Pressure Data Size Threshold
-                                    <div class="fa fa-question-circle" alt="Info" title="This setting applies as the default for any new connections. Configuration of existing connections are unaffected."></div>
+                                    <div class="fa fa-question-circle" alt="Info" title="This setting applies as the default back pressure data size for any new connection. Configuration of existing connections is unaffected."></div>
                                 </div>
                                 <div class="editable setting-field">
                                     <input type="text" id="process-group-default-back-pressure-data-size-threshold" class="setting-input"/>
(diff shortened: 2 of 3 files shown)


mtien-apache (commented):
@markobean Thanks for the PR! Utilizing the extra real estate of the Configuration dialog is a great idea. I have a couple of suggestions:

Screenshot 2023-09-13 at 2 59 50 PM

- Now that we have more than one column, the `Apply Recursively` checkbox looks out of place. Furthermore, if the browser width reduces, then the checkbox and label get pushed into the Parameter Context menu. I suggest moving the checkbox and label to the right of the `Apply` button and change the label to: `Apply Parameter Context recursively`, to be more clear of the action.

Screenshot 2023-09-13 at 3 00 04 PM

- Since we have still have a good amount of empty space remaining in the General tab, I suggest to remove the Comments tab and add the Comments `textarea` as a third column in the General tab. This way we can also keep the `Apply` click action in one place.

Thoughts?

markobean:
> @markobean Thanks for the PR! Utilizing the extra real estate of the Configuration dialog is a great idea. I have a couple of suggestions:
> 
> Screenshot 2023-09-13 at 2 59 50 PM
> * Now that we have more than one column, the `Apply Recursively` checkbox looks out of place. Furthermore, if the browser width reduces, then the checkbox and label get pushed into the Parameter Context menu. I suggest moving the checkbox and label to the right of the `Apply` button and change the label to: `Apply Parameter Context recursively`, to be more clear of the action.
> 
> Screenshot 2023-09-13 at 3 00 04 PM
> * Since we have still have a good amount of empty space remaining in the General tab, I suggest to remove the Comments tab and add the Comments `textarea` as a third column in the General tab. This way we can also keep the `Apply` click action in one place.
> 
> Thoughts?

Thanks for the feedback @mtien-apache! I am away for the next 10 days, but will get to this when I return.

As far as the comments tab, I like it being separate to be consistent with other components. It also leaves room for growth for the main General tab. It has seen the addition of numerous properties, and I feel this may continue over time.

Thanks,
Mark

exceptionfactory:
> As far as the comments tab, I like it being separate to be consistent with other components. It also leaves room for growth for the main General tab. It has seen the addition of numerous properties, and I feel this may continue over time.
> 
@markobean and @mtien-apache One note about the Comments tab: although it aligns with the approach for the Processor configuration dialog, the important difference is the location of the Apply button. On the Processor configuration dialog, the Apply button is always visible, making it clear that it applies to all tabs. With the current approach, the button is not always visible. I could see separating the Comments field into its own tab, but only if we also relocate the Apply button.

markobean:
@mtien-apache I have updated the UI elements so that the "Apply recursively" checkbox appears immediately underneath the Process Group Parameter Context combo. I think placing it next to the "Apply" button is too disjointed; it makes more sense to keep the checkbox next to the item to which it applies.

@exceptionfactory There is an "Apply" button on both the General and Comments tab. Did your comment of "always visible" mean you wanted to see it on the Controller Services tab as well. I think it is not necessary on that tab. Sure, maybe it's a little inconsistent, but there are no changes that need to be applied on the Controller Services tab; it would serve no functional purpose.

Part of the problem may be that the "Apply" button on a Process Group configuration dialog does not dismiss the dialog. It remains present. I have always found that behavior a little odd. I prefer to keep any changes to that as a separate ticket though; I'm not sure if there is a driving reason the dialog should not be dismissed and such a change should be commented on separately, IMO. If you think a change is needed and falls within the scope of this ticket, let me know.

Lastly, sorry for the slow reply. I was on vacation and it took a week on top of the time away to get caught up.

markobean:
Branch has been rebased to main as well.

exceptionfactory (changes requested):
Thanks for following up on the feedback @markobean.

> @exceptionfactory There is an "Apply" button on both the General and Comments tab. Did your comment of "always visible" mean you wanted to see it on the Controller Services tab as well. I think it is not necessary on that tab. Sure, maybe it's a little inconsistent, but there are no changes that need to be applied on the Controller Services tab; it would serve no functional purpose.

To clarify, the reason for having the Apply button always visible is to indicate that saves settings from all tabs, not just the selected tab. For example, entering something in the Comments field, then switching to the Process Group configuration, then pressing Apply would save all of those changes together. The Processor configuration reflects this with the presence of the Apply button regardless of the tab. If we separate Comments to its own tab, we should follow the same approach with the Apply button.

> Part of the problem may be that the "Apply" button on a Process Group configuration dialog does not dismiss the dialog. It remains present. I have always found that behavior a little odd. I prefer to keep any changes to that as a separate ticket though; I'm not sure if there is a driving reason the dialog should not be dismissed and such a change should be commented on separately, IMO. If you think a change is needed and falls within the scope of this ticket, let me know.

Thanks for raising that point, it is worth noting. Although it would expand the scope of this issue, it seems best to couple separate tabs with changing the dialog and apply button behavior. Alternatively, if this pull request kept the Comments field on the same tab, and simply adjusted the layout to multiple columns, that would provide a better incremental step to a separate issue. The subsequent issue could cover moving Comments and implementing the Apply button behavior changes.

markobean:
It's a slight scope creep to the original ticket, but I think it's reasonable. I'm going to refactor the UI a bit more than originally intended such that there will be a Cancel and Apply button on all tabs of the PG configuration. The "X" in the upper right will also go away since it will no longer be needed. Selection of either button will dismiss the dialog. This will be consistent with the configuration of other components such as processors.

exceptionfactory:
> It's a slight scope creep to the original ticket, but I think it's reasonable. I'm going to refactor the UI a bit more than originally intended such that there will be a Cancel and Apply button on all tabs of the PG configuration. The "X" in the upper right will also go away since it will no longer be needed. Selection of either button will dismiss the dialog. This will be consistent with the configuration of other components such as processors.

Thanks for the reply, that sounds like a good direction.

mcgilman:
@markobean Thanks for the PR!

@exceptionfactory @mtien-apache Thanks for the reviews!

I just wanted to chime in here to provide some additional background that may help the direction we proceed here. The Process Group dialog used to be similar to other dialogs in terms of size and input field placement. However, when we introduced fine grain access controls we also added support for Process Group scoped Controller Services. Previous to that there was only a single collection of Controller Services that were available for all components. Because we introduced Process Group scoped Controller Services, we duplicated this table in the Process Group dialog. The previous size of the dialog didn't accommodate the full table and we moved this to a full-screen dialog. 

This Controller Service table is something that I believe we should probably consider now when it comes to the Comments tab and the Apply button. The Apply button was only on the General tab because changes to the Controller Services are not saved when the user presses Apply. They are saved as the user Adds, Edits, Enables, Disables, and Deletes the Services directly in the Controller Service tab. This is also why the dialog doesn't close automatically after Applying changes to the General tab as the user may still want/need to make changes on the Controller Service tab. Because of this, I would avoid placing the Apply or Cancel buttons on the Controller Service tab.

Personally, I would vote for not having a separate Comments tab. While it's nice that there is some alignment with other component configurations, I worry it would leave the Process Group configuration a little confusing. We'd have 3 tabs where only two have an Apply button and pressing the Apply button would save changes made on both but not the other. Leaving the basic behavior in place where there are two tabs, General (with the Apply button) and Controller Services (where changes are saved automatically), would be the least surprising to users and would also eliminate a lot of the scope creep mentioned earlier.

exceptionfactory:
Thanks for the additional background and perspective @mcgilman!

I tend to agree that having a separate tab for the Comments creates more complexity, as I highlighted earlier with the potential confusion related to button visibility and location. A multiple-column view still provides plenty of room for Comments. If we go down the road of separate tabs, then the button handling is a concern. If not, the scope remains focused on streamlining the arrangement on the current General tab.

Open to perspectives, but I would defer to @mcgilman on the optimal way forward.

markobean:
Ok, you've convinced me to return the Comments to the General tab rather than a dedicated tab. I'll proceed with this.

Thanks @mcgilman for the explanation for why the Apply doesn't close the dialog.

mtien-apache:
Thanks @exceptionfactory and @mcgilman for weighing in and agree with your reasonings. 

@markobean Thanks for all the changes! It seems there are many layout options to go with, so a few iterations are probably necessary until we get the right one (which we will). As far as the "Apply Recursively" checkbox, I don't mind it directly underneath the Parameter Content combo. I just want to point out it may throw off the visual rhythm of the rows of menu items across the dialog.

markobean:
@mtien-apache @exceptionfactory @mcgilman 
I've updated this PR to remove the Comments tab. Now the comments field is on the General tab and spans both columns of properties giving plenty of space to work with. It remains resizeable, if needed.

The placement of the Apply Recursively does throw off alignment between the two columns. But I don't find it too distracting. I would find it more distracting to intentionally add a spacer in the second column just to align with the first. The next best alternative would be to move the Process Group Parameter Context to the bottom of column 1 where the additional checkbox would not interfere with horizontal alignment with column 2. 

Note: I have also submitted a PR for the this Jira issue to merge into the `support/nifi-1.x` branch (PR #7844)

markobean:
image

markobean:
@mtien-apache Here's an option which would keep the 2 columns in better alignment. 
Thoughts?

image

markobean:
@mtien-apache, @exceptionfactory, @mcgilman 

Ping. Any additional comments on this PR? Also, note that I opened a [separate PR](https://github.com/apache/nifi/pull/7844) for merging to nifi-1.x.

Thanks.

mcgilman:
Thanks @markobean. I'll have a look today.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
```diff
@@ -87,7 +95,6 @@
 }
 
 #parameter-contexts-recursive-container {
-    float: left;
```
mcgilman, line 90: Removal of the float here is good. With its removal though, we no longer need to float `process-group-parameter-context-combo` (the element before it) so we can remove that style too and we can remove the `div.clear` that comes after it.

mcgilman (approved):
The changes look good. Thanks @markobean! I left one minor comment below that we can further clean up with the changes you've made here.

@exceptionfactory or @mtien-apache did you want to check out the updates or should I merge once Mark makes the final changes I suggested below?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
markobean: Thanks @mcgilman. I implemented your recommendation, but the next line "Execution Engine" appeared in-line with "Apply recursively" when I removed `<div class="clear"></div>`.  I believe you were referring to removing this at Line 50, right? So, I left this in place, but I removed the float for `process-group-parameter-context-combo`

markobean:
@mcgilman @mtien-apache @exceptionfactory 
Hey all.. thank you for the time you've put into this PR so far. I'd like to get it over the line - and more importantly merge the corresponding nifi-1.x PR - before we get to the 1.24.0 release.

https://github.com/apache/nifi/pull/7844

Appreciate your time!

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
mcgilman: I'll take a look. But if nothing is floating (which should be the case now) then there should be no need to clear. But maybe I missed something else.

mcgilman: I see. It is because the elements inside `parameter-contexts-recursive-container` are floated left. It would probably be clearer if the `div.clear` were the last element in that element but it's ok. I'll merge this one. Thanks again for the PR!

mcgilman:
@exceptionfactory CI was apparently not run because you had requested changes. I just manually triggered it. I plan to merge this one once CI completes unless you had any other concerns. Thanks!

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/css/process-group-configuration.css
markobean: Thanks @mcgilman !

Would you mind taking a look/merging the PR for nifi-1.x as well so this can get into the 1.24.0 release?
#7844

exceptionfactory (approved):
Thanks for the work @markobean, and thanks for the review @mcgilman! No other comments, looks good!

PR #7604, opened 2023-08-14
NIFI-11946 Added preserveSourceProperties query parameter to create flow API

ravinarayansingh:
# Summary

[NIFI-11946](https://issues.apache.org/jira/browse/NIFI-11946)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
@@ -339,7 +339,7 @@ public class RegistryService {
         if (versionedFlow.getBucketIdentifier() == null) {
             versionedFlow.setBucketIdentifier(bucketIdentifier);
         }
-        if (!preserveSourceProperties || (versionedFlow.getCreatedTimestamp() <= 0) || (versionedFlow.getModifiedTimestamp() <= 0)) {
+        if (versionedFlow.getCreatedTimestamp()) {
             final long timestamp = System.currentTimeMillis();
             versionedFlow.setCreatedTimestamp(timestamp);
             versionedFlow.setModifiedTimestamp(timestamp);
--- a/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
@@ -106,7 +106,7 @@ public class RegistryService {
         this.registryUrlAliasService = Validate.notNull(registryUrlAliasService);
     }
 
-    private <T>  void validate(T t, String invalidMessage) {
+    private <T> void validate(T t, String invalidMessage) {
         final Set<ConstraintViolation<T>> violations = validator.validate(t);
         if (violations.size() > 0) {
             throw new ConstraintViolationException(invalidMessage, violations);
@@ -218,7 +218,7 @@ public class RegistryService {
             final List<BucketEntity> bucketsWithSameName = metadataService.getBucketsByName(bucket.getName());
             if (bucketsWithSameName != null) {
                 for (final BucketEntity bucketWithSameName : bucketsWithSameName) {
-                    if (!bucketWithSameName.getId().equals(existingBucketById.getId())){
+                    if (!bucketWithSameName.getId().equals(existingBucketById.getId())) {
                         throw new IllegalStateException("A bucket with the same name already exists - " + bucket.getName());
                     }
                 }
@@ -323,7 +323,7 @@ public class RegistryService {
 
     // ---------------------- VersionedFlow methods ---------------------------------------------
 
-    public VersionedFlow createFlow(final String bucketIdentifier, final VersionedFlow versionedFlow) {
+    public VersionedFlow createFlow(final String bucketIdentifier, final VersionedFlow versionedFlow, final boolean preserveSourceProperties) {
         if (StringUtils.isBlank(bucketIdentifier)) {
             throw new IllegalArgumentException("Bucket identifier cannot be null or blank");
         }
@@ -339,10 +339,12 @@ public class RegistryService {
         if (versionedFlow.getBucketIdentifier() == null) {
             versionedFlow.setBucketIdentifier(bucketIdentifier);
         }
+        if (!preserveSourceProperties) {
+            final long timestamp = System.currentTimeMillis();
+            versionedFlow.setCreatedTimestamp(timestamp);
+            versionedFlow.setModifiedTimestamp(timestamp);
+        }
 
-        final long timestamp = System.currentTimeMillis();
-        versionedFlow.setCreatedTimestamp(timestamp);
-        versionedFlow.setModifiedTimestamp(timestamp);
 
         validate(versionedFlow, "Cannot create versioned flow");
 
@@ -480,7 +482,7 @@ public class RegistryService {
             final List<FlowEntity> flowsWithSameName = metadataService.getFlowsByName(existingBucket.getId(), versionedFlow.getName());
             if (flowsWithSameName != null) {
                 for (final FlowEntity flowWithSameName : flowsWithSameName) {
-                     if(!flowWithSameName.getId().equals(existingFlow.getId())) {
+                    if (!flowWithSameName.getId().equals(existingFlow.getId())) {
                         throw new IllegalStateException("A versioned flow with the same name already exists in the selected bucket");
                     }
                 }
@@ -729,7 +731,7 @@ public class RegistryService {
      * Returns all versions of a flow, sorted newest to oldest.
      *
      * @param bucketIdentifier the id of the bucket to search for the flowIdentifier
-     * @param flowIdentifier the id of the flow to retrieve from the specified bucket
+     * @param flowIdentifier   the id of the flow to retrieve from the specified bucket
      * @return all versions of the specified flow, sorted newest to oldest
      */
     public SortedSet<VersionedFlowSnapshotMetadata> getFlowSnapshots(final String bucketIdentifier, final String flowIdentifier) {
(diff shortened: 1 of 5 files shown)


exceptionfactory (changes requested):
Thanks for the contribution @ravinarayansingh.

On initial review, supporting this feature raises some concerns. The created and modified timestamps are intended to reflect what actually occurred. With this optional flag, however, there is no indication that the timestamps do not reflect when a flow was actually created or modified. For this reason, supporting this option seems like it could raise potential usability and tracking issues.

Do you have any additional background for the motivation behind this change?

ravinarayansingh:
> Thanks for the contribution @ravinarayansingh.
> 
> On initial review, supporting this feature raises some concerns. The created and modified timestamps are intended to reflect what actually occurred. With this optional flag, however, there is no indication that the timestamps do not reflect when a flow was actually created or modified. For this reason, supporting this option seems like it could raise potential usability and tracking issues.
> 
> Do you have any additional background for the motivation behind this change?

Hi @exceptionfactory 

I'm looking to upgrade a NiFi Registry that currently uses h2 and the Filesystem as a persistence provider. My goal is to enhance the registry by migrating to Postgres as the provider. During this transition, I need to ensure that the flows from the old registry are transferred to the new one while retaining their original authors and timestamps.

The issue I've encountered is that the current behavior of the "create flow version" API only preserves the author information but not the timestamp. As a solution, I've submitted a pull request (PR) that addresses this concern.

Thanks

ravinarayansingh:
> > > Thanks for the contribution @ravinarayansingh.
> > > On initial review, supporting this feature raises some concerns. The created and modified timestamps are intended to reflect what actually occurred. With this optional flag, however, there is no indication that the timestamps do not reflect when a flow was actually created or modified. For this reason, supporting this option seems like it could raise potential usability and tracking issues.
> > > Do you have any additional background for the motivation behind this change?
> > 
> > 
> > Hi @exceptionfactory
> > I'm looking to upgrade a NiFi Registry that currently uses h2 and the Filesystem as a persistence provider. My goal is to enhance the registry by migrating to Postgres as the provider. During this transition, I need to ensure that the flows from the old registry are transferred to the new one while retaining their original authors and timestamps.
> > The issue I've encountered is that the current behavior of the "create flow version" API only preserves the author information but not the timestamp. As a solution, I've submitted a pull request (PR) that addresses this concern.
> > Thanks
> 
> Thanks for the additional background @ravinarayansingh, that is very helpful.
> 
> In that scenario, the best approach would be to export the information from H2 and import it to PostgreSQL. The [H2 Tutorial](http://www.h2database.com/html/tutorial.html) provides some basic guidance on loading a H2 file, which could then be used for exporting. There will be some differences between H2 and PostgreSQL syntax, but this would be a better way to handle such a migration.
> 
> The create flow REST methods are not necessarily intended for bulk migration, so from that perspective, introducing this new parameter does not seem like the best solution.


As we know  create bucket , create flow version both API's  already have _preserveSourceProperties_  parameter  and for this create Flow i have added .

While I acknowledge that the current approach might not be optimal, I'm curious about the initial rationale for including this parameter. Additionally, could you clarify the reasoning behind exclusively maintaining the author's information?

exceptionfactory:
> Hi @exceptionfactory thanks for response
> 
> As we know create bucket , create flow version both API's already have _preserveSourceProperties_ parameter and for this one _create Flow_ i have added .
> 
> While I acknowledge that the current approach might not be optimal, I'm curious about the initial rationale for including this parameter. Additionally, could you clarify the reasoning behind exclusively maintaining the author's information?

Thanks for the reply @ravinarayansingh, that is a good point and question.

[NIFI-11327](https://issues.apache.org/jira/browse/NIFI-11327) introduced the `preserveSourceProperties` parameter as part of supporting exporting and import operations. From that perspective, there are several create methods that support these operations, so my previous statement was not accurate.

Given that the basic purpose of this parameter is to support some types of migration, the question whether to extend the implementation to timestamps, instead of keeping it limited to the author. Preserving the created timestamp might make sense to maintain historical tracking, but it seems like the modified timestamp should reflect the time when the migration operation occurred.

Is there some particular reason for not wanting to reflect the current timestamp when importing flows?

ravinarayansingh:
> Is there some particular reason for not wanting to reflect the current timestamp when importing flows?

Thanks @exceptionfactory for quick response 

No, there is no specific reason and you are right I should set ModifiedTimestamp with current timestamp.
if  you are agree then I can update the PR with required changes

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/service/StandardServiceFacade.java
```diff
@@ -242,7 +242,7 @@ public List<BucketItem> getBucketItems() {
     // ---------------------- Flow methods ----------------------------------------------
 
     @Override
-    public VersionedFlow createFlow(final String bucketIdentifier, final VersionedFlow versionedFlow) {
+    public VersionedFlow createFlow(final String bucketIdentifier, final VersionedFlow versionedFlow,final boolean preserveSourceProperties) {
```
exceptionfactory, line 245:
```suggestion
    public VersionedFlow createFlow(final String bucketIdentifier, final VersionedFlow versionedFlow, final boolean preserveSourceProperties) {
```

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/service/StandardServiceFacade.java
```diff
@@ -253,7 +253,7 @@ public VersionedFlow createFlow(final String bucketIdentifier, final VersionedFl
         }
 
         final VersionedFlow createdFlow = createRevisableEntity(versionedFlow, VERSIONED_FLOW_ENTITY_TYPE, currentUserIdentity(),
-                () -> registryService.createFlow(bucketIdentifier, versionedFlow));
+                () -> registryService.createFlow(bucketIdentifier, versionedFlow,preserveSourceProperties));
```
exceptionfactory, line 256:
```suggestion
                () -> registryService.createFlow(bucketIdentifier, versionedFlow, preserveSourceProperties));
```

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -99,11 +99,15 @@ public Response createFlow(
...
+                final VersionedFlow flow,
+            @ApiParam(
+                    value = "Whether source properties like identifier,name createdTimestamp and modifiedTimestamp should be kept")
+            @QueryParam("preserveSourceProperties")
+            final boolean preserveSourceProperties) {
 
         verifyPathParamsMatchBody(bucketId, flow);
 
-        final VersionedFlow createdFlow = serviceFacade.createFlow(bucketId, flow);
+        final VersionedFlow createdFlow = serviceFacade.createFlow(bucketId, flow,preserveSourceProperties);
```
exceptionfactory, line 110:
```suggestion
        final VersionedFlow createdFlow = serviceFacade.createFlow(bucketId, flow, preserveSourceProperties);
```

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -99,11 +99,15 @@ public Response createFlow(
             @ApiParam("The bucket identifier")
                 final String bucketId,
             @ApiParam(value = "The details of the flow to create.", required = true)
-                final VersionedFlow flow) {
+                final VersionedFlow flow,
+            @ApiParam(
+                    value = "Whether source properties like identifier,name createdTimestamp and modifiedTimestamp should be kept")
```
exceptionfactory, line 104:
```suggestion
                    value = "Whether source properties like identifier, name, and created should be retained")
```

nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
```diff
@@ -339,10 +339,12 @@ public VersionedFlow createFlow(final String bucketIdentifier, final VersionedFl
         if (versionedFlow.getBucketIdentifier() == null) {
             versionedFlow.setBucketIdentifier(bucketIdentifier);
         }
+        if (!preserveSourceProperties || (versionedFlow.getCreatedTimestamp() <= 0) || (versionedFlow.getModifiedTimestamp() <= 0)) {
```
exceptionfactory, line 342: Actually this approach could be changed so that if the caller sets the timestamps, then they will be preserved, instead of passing the `preserveSourceProperties` argument to this service level. The checks should also be limited to the created timestamp so that the modified timestamp is always updated.
```suggestion
        if (versionedFlow.getCreatedTimestamp()) {
```

exceptionfactory (commented):
Thanks @ravinarayansingh, making adjustments to only retain the created timestamp, but always update the modified timestamp, seems like the best way forward. I also noted a couple minor spacing adjustments, and a recommendation for how to handle checking for and passing down the existing created timestamp.

ravinarayansingh:
> Thanks @ravinarayansingh, making adjustments to only retain the created timestamp, but always update the modified timestamp, seems like the best way forward. I also noted a couple minor spacing adjustments, and a recommendation for how to handle checking for and passing down the existing created timestamp.

hi @exceptionfactory  I have made the changes please have a look 
thanks

ravinarayansingh:
Hi @exceptionfactory 
let me know if I have to do any changes from my side

exceptionfactory (approved):
Thanks for making the adjustments and working through the feedback @ravinarayansingh, the latest version looks good. +1 merging

PR #7608, opened 2023-08-14
NIFI-11789: Support setting fragment.count on last FlowFile when Output Batch Size is set in ExecuteSQL processors

mattyb149:
# Summary

[NIFI-11789](https://issues.apache.org/jira/browse/NIFI-11789) This PR changes AbstractExecuteSQL in order to attempt to try to set the `fragment.count` attribute on the "last" FlowFile to be committed to the session in `onTrigger`.  It introduces some extra concepts that aren't consistent with the intended design of ProcessSession, and currently doesn't work when the query returns multiple result sets. That would require the DB driver to support a handful of methods and configurations that not all of them do (notably Derby). 

Perhaps it is too much outside the Process Session design principles to try and make this work. I a user NEEDS Output Batch Size it's because they have too much data to send downstream at once, which makes a downstream Merge seem like a moot point as the bottleneck would presumably show up there. Perhaps someone will do some performance metrics on how the two types of flows perform, one with Output Batch Size set and one without.

Until more discussion and pathfinding is done, this PR is in draft status.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation


joewitt:
@mattyb149 Can you elaborate on how this doesn't align/work with ProcessSession concepts?

PR #7610, opened 2023-08-15
NIFI-11947 Updated Create Flow Version API to  preserve timestamp  when preserveS…

ravinarayansingh:
…ourceProperties is set to true

# Summary

[NIFI-11947](https://issues.apache.org/jira/browse/NIFI-11947)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
@@ -547,9 +547,10 @@ public class RegistryService {
         }
 
         // validation will ensure that the metadata and contents are not null
-        if (flowSnapshot.getSnapshotMetadata() != null & !preserveSourceProperties) {
+        if (flowSnapshot.getSnapshotMetadata() != null && !preserveSourceProperties ) {
             flowSnapshot.getSnapshotMetadata().setTimestamp(System.currentTimeMillis());
         }
+
         // these fields aren't used for creation
         flowSnapshot.setFlow(null);
         flowSnapshot.setBucket(null);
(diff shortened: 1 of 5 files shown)


exceptionfactory (changes requested):
@ravinarayansingh Similar to the comments on #7604 for NIFI-11946, this approach raises some tracking concerns regarding the standard meaning for the created and modified timestamps. Feel free to respond on that pull request and the answer there should also determine whether this pull request should go forward.

nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
```diff
@@ -539,19 +539,23 @@ public VersionedFlow deleteFlow(final String bucketIdentifier, final String flow
...
+        return createFlowSnapshot(flowSnapshot,false);
+    }
+    public VersionedFlowSnapshot createFlowSnapshot(final VersionedFlowSnapshot flowSnapshot, final boolean preserveSourceProperties) {
         if (flowSnapshot == null) {
             throw new IllegalArgumentException("Versioned flow snapshot cannot be null");
         }
 
         // validation will ensure that the metadata and contents are not null
-        if (flowSnapshot.getSnapshotMetadata() != null) {
+        if (flowSnapshot.getSnapshotMetadata() != null && !preserveSourceProperties ) {
```
exceptionfactory, line 550: Similar to the approach in #7604, it looks like this method could be modified to check for a timestamp value already specified in the calling method. That would avoid having to pass down the `preserveSourceProperties` argument to this level.

exceptionfactory (changes requested):
@ravinarayansingh Revisiting this pull request after merging #7604, it seems like the same approach could work with setting the timestamp in the calling `BucketFlowResource.createFlowVersion()` method.

nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
ravinarayansingh: > Similar to the approach in #7604, it looks like this method could be modified to check for a timestamp value already specified in the calling method. That would avoid having to pass down the `preserveSourceProperties` argument to this level.

Hi @exceptionfactory  
I have made the requested change please have a look

exceptionfactory (approved):
Thanks for making the adjustments @ravinarayansingh, the latest version looks good. +1 merging

PR #7634, opened 2023-08-21
NIFI-11873 Refactored PeerSelectorTest from Groovy to Java.

dan-s1:
Refactored PeerSelectorTest from Groovy to Java. Replaced the Groovy mock objects with Mockito mocks and changed scope of certain variables and methods in PeerSelector from private to package private in order to allow access by PeerSelectorTest.

# Summary

[NIFI-11873](https://issues.apache.org/jira/browse/NIFI-11873)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


dan-s1:
@exceptionfactory 21/22 of pass. There is one test testShouldFetchRemotePeerStatusesInFailureScenario which  fails but the reasons vary each time its run. It appears it is due to some random feature which is being used.  I am not sure though how this all works and I would appreciate another set of eyes why this is happening.
What I am seeing is the failure is either at attempt 2 and sometimes at attempt 4
 Each time actual value is 
       ` PeerStatus[hostname=node1.nifi,port=-1,secure=false,flowFileCount=0]`
        instead of
        `PeerStatus[hostname=node2.nifi,port=-1,secure=false,flowFileCount=0]` or `null`

exceptionfactory (changes requested):
Thanks for working on this refactoring @dan-s1. Unfortunately the Groovy-based test included a number of access violations for calling private methods. It would be much better to avoid the method visibility changes solely for the purpose of maintaining these tests. That may mean some tests need to be removed or rewritten completely. This is one of the more complicated classes, so that may also point to the need for other refactoring, as a first step, I recommend making adjustment to avoid visibility changes. If that presents too much difficulty, it may be better to close and revisit later.

exceptionfactory:
I realize that I recommended changing the visibility of one method when commenting on the Jira, but I did not expect it to be necessary for some many methods.

dan-s1:
I did not realize either. I worked off your original comment thinking it would be okay for the others. Aside from variables I am not sure how I would not reduce the scope of the methods whose scope had to change. I am not sure I understand everything going on in `PeerSelector` so I do not feel I would be the right one to refactor it. Is there anyway what I have done in this PR be used? I did get most of the tests to work by using Mockito which I thought be worthwhile to preserve.

exceptionfactory:
Thanks for your work on it, it may be a helpful background for someone else to take it up. If there are any tests that can be migrated without method visibility changes, that would might be a helpful starting point for a new Java-based test class. Then subsequent work could handle refactoring the remaining tests.

dan-s1:
Should I close the ticket or just have the ticket no longer assigned to me? Also should I add you comments regarding the visibility issues to the ticket so others will know what to be aware of?

exceptionfactory:
> Should I close the ticket or just have the ticket no longer assigned to me? Also should I add you comments regarding the visibility issues to the ticket so others will know what to be aware of?

The ticket should be kept open, but you can unassign yourself. The pull request comments are automatically mirrored to the Jira issue.

exceptionfactory:
Thanks again for your work on this pull request @dan-s1. I am closing it for now, but it will be linked in the Jira issue for future reference.

PR #7661, opened 2023-08-30
NIFI-11772 Removing flow.xml support

simonbence:
This is a draft pull request in order to finalize the approach. Things not aimed in this PR:
- Finalize documentation
- Toolkit changes

# Summary

[NIFI-11772](https://issues.apache.org/jira/browse/NIFI-11772)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


exceptionfactory (commented):
Thanks for working through this extensive set of files @simonbence! With the removal of Templates now merged under [NIFI-12006](https://issues.apache.org/jira/browse/NIFI-12006) can you rebase this PR and indicate when it is ready for review?

minifi/minifi-integration-tests/src/test/resources/conf/nifi.properties
```diff
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 # Core Properties #
-nifi.flow.configuration.file=./target/flow.xml.gz
```
exceptionfactory, line 17: Should this be flow.json.gz?

nifi-commons/nifi-flow-encryptor/src/main/java/org/apache/nifi/flow/encryptor/JsonFlowEncryptor.java
```diff
@@ -33,9 +34,13 @@ public class JsonFlowEncryptor extends AbstractFlowEncryptor {
     @Override
     public void processFlow(final InputStream inputStream, final OutputStream outputStream,
                             final PropertyEncryptor inputEncryptor, final PropertyEncryptor outputEncryptor) {
+
+        final BufferedInputStream bufferedInputStream = new BufferedInputStream(inputStream);
+        bufferedInputStream.mark(1);
```
exceptionfactory, line 39: This addition does not appear necessary. The only reason for it in the standard class was to support content type detection.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/resources/FlowConfiguration.xsd
```diff
@@ -410,8 +410,6 @@
             <xs:extension base="PortType">
                 <xs:sequence>
                     <xs:element name="maxConcurrentTasks" type="xs:positiveInteger" minOccurs="0" ></xs:element>
```
exceptionfactory, line 412: This entire file can be deleted.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/resources/conf/nifi.properties
```diff
@@ -69,6 +69,10 @@ nifi.web.https.host=
 nifi.web.https.port=
 nifi.web.jetty.working.directory=./target/work/jetty
 
+# security properties #
+nifi.sensitive.props.key=key
+nifi.sensitive.props.algorithm=PBEWITHMD5AND256BITAES-CBC-OPENSSL
```
exceptionfactory, line 74: This algorithm is deprecated and no longer supported on the main branch, so these two properties can be removed.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-properties-loader/src/test/resources/properties/conf/duplicates.nifi.properties
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-nifi.flow.configuration.file=./conf/flow.xml.gz
 nifi.flow.configuration.file=./conf/flow.json.gz
+nifi.flow.configuration.file=./conf/flow2.json.gz
```
exceptionfactory, line 17: Is this change correct?

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/FlowSynchronizationIT.java
```diff
@@ -546,7 +546,7 @@ private void restartWithOnlySingleFlowPersistenceFile(final String filenameToDel
         node2.stop();
 
         final File confDir = new File(node2.getInstanceDirectory(), "conf");
-        assertEquals(1, confDir.listFiles(file -> file.getName().equals("flow.xml.gz")).length);
+        //assertEquals(1, confDir.listFiles(file -> file.getName().equals("flow.xml.gz")).length);
```
exceptionfactory, line 549: This commented line should be removed.

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
```diff
@@ -106,16 +97,16 @@ public void testStartupWithDifferentFlow() throws IOException, NiFiClientExcepti
...
-        Files.copy(Paths.get("src/test/resources/flows/mismatched-flows/flow2.xml.gz"), flowXmlFile.toPath());
+        Files.copy(Paths.get("src/test/resources/flows/mismatched-flows/flow2.json.gz"), flowXmlFile.toPath());
 
         node2.start(true);
 
         final File backupFile = getBackupFile(node2ConfDir);
         final NodeDTO node2Dto = getNodeDtoByApiPort(5672);
 
-        verifyFlowContentsOnDisk(backupFile);
+//        verifyFlowContentsOnDisk(backupFile);
```
exceptionfactory, line 109: Is this verify method not working?

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
```diff
@@ -124,61 +115,62 @@ public void testStartupWithDifferentFlow() throws IOException, NiFiClientExcepti
...
-        final FlowSnippetDTO childContents = childGroup.getContents();
-
-        final Set<ProcessorDTO> childProcessors = childContents.getProcessors();
-        assertEquals(1, childProcessors.size());
-
-        final ProcessorDTO procDto = childProcessors.iterator().next();
-        assertFalse(procDto.getId().endsWith("00"));
-        assertFalse(procDto.getName().endsWith("00"));
-    }
+    // TODO we need a JSON analogue for this
```
exceptionfactory, line 137: We should remove this method and create a new Jira issue instead of commenting out the method.

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
```diff
@@ -248,8 +228,8 @@ class ConfigEncryptionTool {
...
         options.addOption(Option.builder("u").longOpt(OUTPUT_AUTHORIZERS_ARG).hasArg(true).argName("file").desc("The destination authorizers.xml file containing protected config values (will not modify input authorizers.xml)").build())
-        options.addOption(Option.builder("f").longOpt(FLOW_XML_ARG).hasArg(true).argName("file").desc("The flow.xml.gz file currently protected with old password (will be overwritten unless -g is specified)").build())
-        options.addOption(Option.builder("g").longOpt(OUTPUT_FLOW_XML_ARG).hasArg(true).argName("file").desc("The destination flow.xml.gz file containing protected config values (will not modify input flow.xml.gz)").build())
+//        options.addOption(Option.builder("f").longOpt(FLOW_XML_ARG).hasArg(true).argName("file").desc("The flow.xml.gz file currently protected with old password (will be overwritten unless -g is specified)").build())
+//        options.addOption(Option.builder("g").longOpt(OUTPUT_FLOW_XML_ARG).hasArg(true).argName("file").desc("The destination flow.xml.gz file containing protected config values (will not modify input flow.xml.gz)").build())
```
exceptionfactory, line 232: These commented lines should be removed.

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
```diff
@@ -715,20 +695,20 @@ class ConfigEncryptionTool {
...
-        final PropertyEncryptor outputEncryptor = new PropertyEncryptorBuilder(newFlowPassword).setAlgorithm(newAlgorithm).build()
-
-        final FlowEncryptor flowEncryptor = new StandardFlowEncryptor()
-        flowEncryptor.processFlow(flowXmlContent, flowOutputStream, inputEncryptor, outputEncryptor)
-
-        // Overwrite the original flow file with the migrated flow file
-        Files.move(tempFlowXmlFile.toPath(), Paths.get(outputFlowXmlPath), StandardCopyOption.ATOMIC_MOVE)
-        loadFlowXml(outputFlowXmlPath)
-    }
+//    private InputStream migrateFlowXmlContent(InputStream flowXmlContent, String existingFlowPassword, String newFlowPassword, String existingAlgorithm = DEFAULT_FLOW_ALGORITHM, String newAlgorithm = DEFAULT_FLOW_ALGORITHM) {
```
exceptionfactory, line 698: This needs to be uncommented and working before going forward.

exceptionfactory (changes requested):
Thanks for working on this change @simonbence! I noted a few details to address.

In particular, commented out methods need to be addressed, either corrected or removed. In the case of the encrypt-config command, it should already support flow.json.gz, but methods need to be renamed.

As far as the framework integration tests in nifi-framework-cluster, I believe it would be better to just remove all of them since they are not actively maintained or run. Perhaps @markap14 can weigh in on this question, but the system-test-suite should provide all the necessary coverage, and those older tests, along with accompanying flow configurations, can be removed.

The removal of the legacy code related to port authorization looks good.

One general question is whether we should switch the configuration property or keep the json.file property. Relying on the json.file property name would simplify upgrading without having to change property values. On the other hand, it is a simple manual change that seems reasonable. Thoughts?

markap14:
Thanks for thoroughly reviewing @exceptionfactory . I'll plan on reviewing this in more depth next week as well. In the meantime, I do agree with your comment above re: deleting the antiquated integration tests.

exceptionfactory:
> Thanks for thoroughly reviewing @exceptionfactory . I'll plan on reviewing this in more depth next week as well. In the meantime, I do agree with your comment above re: deleting the antiquated integration tests.

Thanks for the confirmation on the integration tests @markap14!

exceptionfactory (commented):
@simonbence I removed the obsolete framework integration tests in #7784, so if you can rebase, that would streamline the review.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-properties-loader/src/test/resources/properties/conf/duplicates.nifi.properties
simonbence: Yes. According to the test uses this config, it is intentional to have two values for the same key

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
simonbence: I fixed the test within the story. Please take a look. There are some things you might want to ask about, especially the removeal of the disconnection. As far as I understood it 1. did not work as expected even before the PR 2. was unnecessary for the given test scenario. I might miss something so please mention if you disagree with any of the changes

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
simonbence: I am not sure if this is the right way to move forward. This is a method specific to the XML flow definition and needs the already deleted XML based flow encryptor. I would find it better to remove it alltogether. As a tentative step I did add back the XmlFlowEncryptor but not to it's original place. Right not it is situated within this module until it's future is not decided. I did not yet put much effort to test this as it is now, because 1. I would like to hear your opinion before 2. originally the changes on the toolkit was planned for a followup story. But based on your answer I will adjust my actions on the matter

simonbence:
Hi @exceptionfactory ! Thank you very much for the effort you did put on the review! I added some changes. Also answering your question:

`
One general question is whether we should switch the configuration property or keep the json.file property. Relying on the json.file property name would simplify upgrading without having to change property values. On the other hand, it is a simple manual change that seems reasonable. Thoughts?
`
I was thinking on this myself as well, but I deciced to use the original property name without remarking the file format because from now on, this should be the default format thus it looks both unnecessary and confusing to highlight this.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/audit/PortAuditor.java
```diff
@@ -93,14 +91,8 @@ public Port updatePortAdvice(ProceedingJoinPoint proceedingJoinPoint, PortDTO po
         final int maxConcurrentTasks = port.getMaxConcurrentTasks();
 
         final Set<String> existingUsers = new HashSet<>();
-        final Set<String> existingGroups = new HashSet<>();
         boolean isPublicPort = false;
```
markap14, line 94: It looks like `isPublicPort` will always be false here, as you've removed the `if (port instanceof PublicPort) {... }`. I think we need to instead set `boolean isPublicPort = port instanceof PublicPort;`

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
```diff
@@ -106,100 +95,57 @@ public void testStartupWithDifferentFlow() throws IOException, NiFiClientExcepti
...
+        Files.deleteIfExists(flowJsonFile.toPath());
+        Files.copy(Paths.get("src/test/resources/flows/mismatched-flows/flow2.json.gz"), flowJsonFile.toPath());
 
         node2.start(true);
 
-        final File backupFile = getBackupFile(node2ConfDir);
-        final NodeDTO node2Dto = getNodeDtoByApiPort(5672);
+        while (getNifiClient().getControllerClient().getNodes().getCluster().getNodes().stream().filter(n -> n.getStatus().equals("CONNECTED")).count() != 2) {
+            Thread.sleep(3000);
+        }
```
markap14, line 106: This should all be replaced with `waitForAllNodesConnected();`

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
```diff
@@ -106,100 +95,57 @@ public void testStartupWithDifferentFlow() throws IOException, NiFiClientExcepti
...
     private File getBackupFile(final File confDir) throws InterruptedException {
-        waitFor(() -> getFlowXmlFiles(confDir).size() == 2);
-
-        final List<File> flowXmlFiles = getFlowXmlFiles(confDir);
-        assertEquals(2, flowXmlFiles.size());
-
-        flowXmlFiles.removeIf(file -> file.getName().equals("flow.xml.gz"));
+        waitFor(() -> getFlowJsonFiles(confDir).size() == 1);
 
+        final List<File> flowXmlFiles = getFlowJsonFiles(confDir);
```
markap14, line 122: Variable name should have been `flowJsonFiles`

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
```diff
@@ -106,100 +95,57 @@ public void testStartupWithDifferentFlow() throws IOException, NiFiClientExcepti
...
+            Thread.sleep(3000);
+        }
 
+        final File backupFile = getBackupFile(node2ConfDir);
         verifyFlowContentsOnDisk(backupFile);
-        disconnectNode(node2Dto);
         verifyInMemoryFlowContents();
-
-        // Reconnect the node so that we can properly shutdown
-        reconnectNode(node2Dto);
```
markap14, line 123: Is there a reason that we're changing the logic here?
Previously, the idea was to disconnect Node 2 so that any request made to the node would be a response directly from that node rather than a cluster response; then issue the request to Node 2 and verify that its flow is correct. Then, it would reconnect the node so that the shutdown logic would complete successfully.

The logic here has now changed to "verify that the cluster's flow is correct", which is not the same as "verify that Node 2's flow is correct."

That said, there is now a better way to handle this, by doing something like:
```
switchClientToNode(2);
getNifiClient().getFlowClient(DO_NOT_REPLICATE)...
```
This way, we don't have to disconnect the node, but we can issue requests directly to that node. But I think we need to either restore the logic or use the DO_NOT_REPLICATE to ensure that we're verifying Node 2's flow.

markap14 (changes requested):
Thanks for all the work you've put into this @simonbence. Not gonna lie, I was pretty scared when I saw 11,000 lines removed! But most of it was flow.xml files, etc. Made me feel much better :)

Overall I think it looks good. I commented inline a couple of concerns, but I think all are trivial to address. Otherwise, I'm a +1.

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
simonbence: Thanks! I would believe the previous logic did not work properly as it was before remoeval (it did a REST call against the endpoint `cluster/nodes/${nodeId}` which behaves based on the accompanying `NodeEntity` / `NodeDTO`, but the status was not set to be DISCONNECTING, so in practice I think it did not disconnect), but thanks for the context! I will look how it behaves this way

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
```diff
@@ -722,7 +707,7 @@ class ConfigEncryptionTool {
         final PropertyEncryptor inputEncryptor = new PropertyEncryptorBuilder(existingFlowPassword).setAlgorithm(existingAlgorithm).build()
         final PropertyEncryptor outputEncryptor = new PropertyEncryptorBuilder(newFlowPassword).setAlgorithm(newAlgorithm).build()
 
-        final FlowEncryptor flowEncryptor = new StandardFlowEncryptor()
+        final FlowEncryptor flowEncryptor = new XmlFlowEncryptor()
```
exceptionfactory, line 710: I may not have communicated clearly in the previous comment, but `XmlFlowEncryptor` can be removed and `StandardFlowEncryptor` should be used in this location.

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
```diff
@@ -385,26 +370,26 @@ class ConfigEncryptionTool {
...
+//
+//                if (flowXmlPath == outputFlowXmlPath) {
+//                    // TODO: Add confirmation pause and provide -y flag to offer no-interaction mode?
+//                    logger.warn("The source flow.xml.gz and destination flow.xml.gz are identical [${outputFlowXmlPath}] so the original will be overwritten")
+//                }
+//
+//                if (!commandLine.hasOption(NIFI_PROPERTIES_ARG)) {
+//                    printUsageAndThrow("In order to migrate a flow.xml.gz, a nifi.properties file must also be specified via '-n'/'--${NIFI_PROPERTIES_ARG}'.", ExitCode.INVALID_ARGS)
+//                }
+//            }
```
exceptionfactory, line 392: This block of code needs to be uncommented and working. Although the code says `flow.xml.gz`, it also works with `flow.json.gz`, so names should be adjusted accordingly.

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
```diff
@@ -248,8 +235,6 @@ class ConfigEncryptionTool {
         options.addOption(Option.builder("i").longOpt(OUTPUT_LOGIN_IDENTITY_PROVIDERS_ARG).hasArg(true).argName("file").desc("The destination login-identity-providers.xml file containing protected config values (will not modify input login-identity-providers.xml)").build())
         options.addOption(Option.builder("a").longOpt(AUTHORIZERS_ARG).hasArg(true).argName("file").desc("The authorizers.xml file containing unprotected config values (will be overwritten unless -u is specified)").build())
         options.addOption(Option.builder("u").longOpt(OUTPUT_AUTHORIZERS_ARG).hasArg(true).argName("file").desc("The destination authorizers.xml file containing protected config values (will not modify input authorizers.xml)").build())
-        options.addOption(Option.builder("f").longOpt(FLOW_XML_ARG).hasArg(true).argName("file").desc("The flow.xml.gz file currently protected with old password (will be overwritten unless -g is specified)").build())
-        options.addOption(Option.builder("g").longOpt(OUTPUT_FLOW_XML_ARG).hasArg(true).argName("file").desc("The destination flow.xml.gz file containing protected config values (will not modify input flow.xml.gz)").build())
```
exceptionfactory, line 252: These options need to be retained and changed to indicate flow.json.gz. We will need separate work on the toolkit in general, but at least for this PR, it should continue to function with flow.json.gz.

exceptionfactory (changes requested):
Thanks for making the updates @simonbence. This looks closer to completion, pending responses to feedback from @markap14. I also noted a couple more changes necessary in the toolkit class.

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
simonbence: In this case I am not 100% sure if I get your intent. `StandardFlowEncryptor` is (actually: was) a just a composite implementation which helped NiFi decide is the `XmlFlowEncryptor` or the `JsonFlowEncryptor` needs to be used. As `XmlFlowEncryptor` is not there anymore `StandardFlowEncryptor` was deleted as well. In the other hand this toolkit method needs an encryptor capable to work with XML, let it be `XmlFlowEncryptor` directly or `StandardFlowEncryptor` (which will call `XmlFlowEncryptor` in the background). So if we are willing to keep providing this particular service (`migrateFlowXmlContent`) we need the `XmlFlowEncryptor` in place. If this is your aim, I would limit it's scope to the toolkit (I added back this way) and would not add back the composite (`StandardFlowEncryptor`) as it provides no feature during the standard run of NiFi

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
simonbence: Yes, as mentioned in the description, originally I wanted to do the toolkit changes in a follow up story after clarifying the cleanup and general changes within this PR. Depending on your answer for my comment above hovere I will change this part accordingly. For further toolkit related changes I suggest to move them in a followup PR.

exceptionfactory: Thanks for the reply @simonbence. Handling the toolkit changes in a separate PR sounds good, but in that case, I recommend reverting this particular removal and keeping the toolkit changes as minimal as possible. As these argument names do not impact actual flow.xml.gz usage, leaving them in place seems fine for now, and more substantive changes can be handled separately. So from that perspective, I recommend limiting the toolkit changes to only those that are required for the removal of actual flow.xml.gz handling code.

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
exceptionfactory: As noted in relation to the arguments, it looks like this can be uncommented for this PR, and then other changes can be handled separately.

nifi-toolkit/nifi-toolkit-encrypt-config/src/main/groovy/org/apache/nifi/properties/ConfigEncryptionTool.groovy
exceptionfactory: Thanks for the reply and clarifying the implementations. In this case, it looks like the implementation should be changed to the JsonFlowEncryptor. The `migrateFlowXmlContent()` method can be renamed to `migrateFlowContent()` and that seems sufficient for this PR. How does that sound?

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
```diff
@@ -106,124 +94,72 @@ public void testStartupWithDifferentFlow() throws IOException, NiFiClientExcepti
...
-        // gzipped bytes and because if the two flows do differ, we want to have the String representation so that we can compare to see how they are different.
-        final String flowXml = readFlow(backupFile);
-        final String expectedFlow = readFlow(new File("src/test/resources/flows/mismatched-flows/flow1.xml.gz"));
-
-        assertEquals(expectedFlow, flowXml);
-
         // Verify some of the values that were persisted to disk
         final File confDir = backupFile.getParentFile();
-        final String loadedFlow = readFlow(new File(confDir, "flow.xml.gz"));
+        final String loadedFlow = readFlow(new File(confDir, "flow.json.gz"));
```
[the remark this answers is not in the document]
simonbence: https://issues.apache.org/jira/browse/NIFI-12176

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/JoinClusterWithDifferentFlow.java
```diff
@@ -236,20 +172,9 @@ private void verifyInMemoryFlowContents() throws NiFiClientException, IOExceptio
...
-
-    private PropertyEncryptor createEncryptorFromProperties(Properties properties) {
-        final NiFiProperties niFiProperties = NiFiProperties.createBasicNiFiProperties(null, properties);
-
-        final String propertiesKey = niFiProperties.getProperty(NiFiProperties.SENSITIVE_PROPS_KEY);
-        final String propertiesAlgorithm = niFiProperties.getProperty(NiFiProperties.SENSITIVE_PROPS_ALGORITHM);
-        return new PropertyEncryptorBuilder(propertiesKey).setAlgorithm(propertiesAlgorithm).build();
+        final ControllerServicesEntity controllerLevelServices = getNifiClient().getFlowClient(DO_NOT_REPLICATE).getControllerServices();
+        final Set<ControllerServiceEntity> controllerServices = controllerLevelServices.getControllerServices();
+        assertEquals(2, controllerServices.size());
```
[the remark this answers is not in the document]
simonbence: I think that is correct. The original test had assertion to check for 1 Controller Service. As I mentioned there are a lot of moving parts (prod code, test code, test data) but based on what I did see, there is a behaviour in difference compared to the XML. I created a ticket for this: https://issues.apache.org/jira/browse/NIFI-12203

I will push a commit soon in order to adjust the test.

exceptionfactory (commented):
Thanks for working through the feedback @simonbence, the latest version appears to address the remaining feedback.

I pushed one more commit to remove the FlowConfiguration.xsd, which is no longer needed as noted earlier.

Do you have any additional feedback @markap14? If not, I can proceed with merging soon.

exceptionfactory (approved):
Confirmed the latest changes with @markap14 in offline conversation. Thanks again for the work on this @simonbence! +1 merging to main!

PR #7664, opened 2023-09-01
NIFI-12019 bugfix for SynchronousFileWatcher time check interval

mosermw:
# Summary

[NIFI-12019](https://issues.apache.org/jira/browse/NIFI-12019)  This PR corrects SynchronousFileWatcher so that it resets its last checked timestamp whenever it uses the configured UpdateMonitor to check on the file being watched.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

- [n/a] Documentation formatting appears as expected in rendered files

### DIFF
--- a/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/file/monitor/SynchronousFileWatcher.java
+++ b/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/file/monitor/SynchronousFileWatcher.java
@@ -59,7 +59,7 @@ public class SynchronousFileWatcher {
     /**
      * Checks if the file has been updated according to the configured {@link UpdateMonitor} and resets the state
      *
-     * @return true if updated; false otherwise
+     * @return true if updated; false if not updated or if not yet time to check
      * @throws IOException if failure occurs checking for changes
      */
     public boolean checkAndReset() throws IOException {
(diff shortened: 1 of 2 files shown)


nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/file/monitor/SynchronousFileWatcher.java
```diff
@@ -77,21 +77,16 @@ public boolean checkAndReset() throws IOException {
...
-                final StateWrapper wrapper = lastState.get();
+                final Object oldState = lastState.get().getState();
                 final Object newState = monitor.getCurrentState(path);
-                if (newState == null && wrapper.getState() == null) {
+                if (newState == null && oldState == null) {
                     return false;
                 }
-                if (newState == null || wrapper.getState() == null) {
-                    lastState.set(new StateWrapper(newState));
+                lastState.set(new StateWrapper(newState));
```
mattyb149, line 85: If newState is `null` does this cause issues creating a wrapper around it?

mosermw: I don't think so?  StateWrapper is a simple pojo with getters/setters and no real logic, so I don't think it cares.  This method handles checking for nulls on newState and oldState, so I was happy that NPE wasn't possible.

mattyb149: Why the null checks then? Just to be defensive?

mosermw: We don't know what UpdateMonitor implementation is currently in use, and that interface doesn't guarantee that `monitor.getCurrentState(path)` returns NotNull.  So we have to check for nulls before using `newState.equals()` on line 89.

mattyb149:
+1 LGTM, thanks for the fix! Merging to support/nifi-1.x and main

mosermw:
I appreciate you looking at this, @mattyb149.

joewitt:
@mattyb149 @mosermw I had never even heard of this class.  But today I noticed this test failure https://github.com/joewitt/nifi/actions/runs/6252186486/job/16974874215

Can you please look into this?  I had never noticed this class failing before.

mosermw:
@joewitt I'm looking at this now.  SynchronousFileWatcher starts a background thread to check if a file has changed, so the unit test timing tolerances are likely too tight.  Not sure if there's a good pattern for testing things like this that remove timing issues from the test.

PR #7665, opened 2023-09-01
NIFI-11197 Initial check in for Yaml record reader

dan-s1:
# Summary

[NIFI-11197](https://issues.apache.org/jira/browse/NIFI-11197)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParserFactory.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParserFactory.java
@@ -23,27 +23,20 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Objects;
 
-public class JsonParserFactory implements TokenParserFactory{
-    private static final JsonFactory JSON_FACTORY = new JsonFactory();
-    private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
-
+public class JsonParserFactory implements TokenParserFactory {
     @Override
-    public JsonParser getJsonParser(InputStream in) throws IOException {
-        JsonParser jsonParser = JSON_FACTORY.createParser(in);
-        jsonParser.setCodec(JSON_MAPPER);
+    public JsonParser getJsonParser(final InputStream in, final StreamReadConstraints streamReadConstraints, final boolean allowComments) throws IOException {
+        Objects.requireNonNull(in, "Input Stream required");
+        Objects.requireNonNull(streamReadConstraints, "Stream Read Constraints required");
 
-        return jsonParser;
-    }
-
-    @Override
-    public ObjectMapper createCodec(boolean allowComments, StreamReadConstraints streamReadConstraints) {
-        ObjectMapper codec = new ObjectMapper();
-        if(allowComments) {
-            codec.enable(JsonParser.Feature.ALLOW_COMMENTS);
+        final ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.getFactory().setStreamReadConstraints(streamReadConstraints);
+        if (allowComments) {
+            objectMapper.enable(JsonParser.Feature.ALLOW_COMMENTS);
         }
-        codec.getFactory().setStreamReadConstraints(streamReadConstraints);
-
-        return codec;
+        final JsonFactory jsonFactory = objectMapper.getFactory();
+        return jsonFactory.createParser(in);
     }
 }
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParserFactory.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParserFactory.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.json;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.StreamReadConstraints;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public class JsonParserFactory implements TokenParserFactory{
+    private static final JsonFactory JSON_FACTORY = new JsonFactory();
+    private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
+
+    @Override
+    public JsonParser getJsonParser(InputStream in) throws IOException {
+        JsonParser jsonParser = JSON_FACTORY.createParser(in);
+        jsonParser.setCodec(JSON_MAPPER);
+
+        return jsonParser;
+    }
+
+    @Override
+    public ObjectMapper createCodec(boolean allowComments, StreamReadConstraints streamReadConstraints) {
+        ObjectMapper codec = new ObjectMapper();
+        if(allowComments) {
+            codec.enable(JsonParser.Feature.ALLOW_COMMENTS);
+        }
+        codec.getFactory().setStreamReadConstraints(streamReadConstraints);
+
+        return codec;
+    }
+}
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonRecordSource.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonRecordSource.java
@@ -18,6 +18,7 @@ package org.apache.nifi.json;
 
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.StreamReadConstraints;
 import com.fasterxml.jackson.core.io.SerializedString;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.apache.nifi.schema.inference.RecordSource;
@@ -29,6 +30,11 @@ import java.io.InputStream;
 
 public class JsonRecordSource implements RecordSource<JsonNode> {
     private static final Logger logger = LoggerFactory.getLogger(JsonRecordSource.class);
+
+    private static final StreamReadConstraints DEFAULT_STREAM_READ_CONSTRAINTS = StreamReadConstraints.defaults();
+
+    private static final boolean ALLOW_COMMENTS_ENABLED = true;
+
     private final JsonParser jsonParser;
     private final StartingFieldStrategy strategy;
 
@@ -41,7 +47,7 @@ public class JsonRecordSource implements RecordSource<JsonNode> {
     }
 
     public JsonRecordSource(final InputStream in, final StartingFieldStrategy strategy, final String startingFieldName, TokenParserFactory tokenParserFactory) throws IOException {
-        jsonParser = tokenParserFactory.getJsonParser(in);
+        jsonParser = tokenParserFactory.getJsonParser(in, DEFAULT_STREAM_READ_CONSTRAINTS, ALLOW_COMMENTS_ENABLED);
         this.strategy = strategy;
 
         if (strategy == StartingFieldStrategy.NESTED_FIELD) {
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonRecordSource.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonRecordSource.java
@@ -16,12 +16,10 @@
  */
 package org.apache.nifi.json;
 
-import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.io.SerializedString;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.nifi.schema.inference.RecordSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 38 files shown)


dan-s1:
@exceptionfactory Is there anything else you would like me to do for this PR?

exceptionfactory:
@dan-s1 Nothing yet, I have been focusing on several other things, but plan to review this soon.

nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonRecordSource.java
```diff
@@ -31,29 +29,23 @@
...
+    }
+
+    public JsonRecordSource(final InputStream in, final StartingFieldStrategy strategy, final String startingFieldName, TokenParserFactory tokenParserFactory) throws IOException {
+        jsonParser = tokenParserFactory.getJsonParser(in);
         this.strategy = strategy;
-        this.startingFieldName = startingFieldName;
 
-        if (strategy == StartingFieldStrategy.NESTED_FIELD) {
-            final SerializedString serializedNestedField = new SerializedString(this.startingFieldName);
+        if (StartingFieldStrategy.NESTED_FIELD.equals(strategy)) {
```
exceptionfactory, line 47: The comparison of `enum` values should use `==` instead of `equals()`, although both will work.
```suggestion
        if (StartingFieldStrategy.NESTED_FIELD == strategy) {
```

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/test/java/org/apache/nifi/yaml/TestYamlTreeRowRecordReader.java
```diff
@@ -0,0 +1,1362 @@
...
+        final String msg = mre.getCause().getMessage();
+        assertTrue(msg.contains("account.balance"));
+        assertTrue(msg.contains("true"));
+        assertTrue(msg.contains("Double"));
+        assertTrue(msg.contains("Boolean"));
+    }
+
+    @Test
+    void testMergeOfSimilarRecords() throws Exception {
+        // GIVEN
```
exceptionfactory, line 692: The `GIVEN` and `WHEN/THEN` comments can be removed.

exceptionfactory (commented):
Thanks for working on this new feature @dan-s1. The changes for reusing the JSON-based components look straightforward. I plan on doing some additional testing soon, but the general approach looks good.

nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonRecordSource.java
dan-s1: @exceptionfactory Sorry I thought I needed to do that to avoid an NPE since `strategy` could be null.

exceptionfactory: No problem, reversing the order avoided the NPE, so that was a good change, but it will also work with `==` as suggested.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/test/java/org/apache/nifi/yaml/TestYamlTreeRowRecordReader.java
```diff
@@ -0,0 +1,1347 @@
...
+            //final Object[] thirdRecordValues = reader.nextRecord().getValues();
+            //assertArrayEquals(new Object[] {3, "Maria Doe", 4750.89, "123 My Street", "My City", "ME", "11111", "USA"}, thirdRecordValues);
+
+            final Object[] fourthRecordValues = reader.nextRecord().getValues();
+            assertArrayEquals(new Object[] {4, "Xi Doe", 4820.09, "321 Your Street", "Your City", "NV", "33333", "USA"}, fourthRecordValues);
+
+
+            assertNull(reader.nextRecord());
+        }
+    }
```
exceptionfactory, line 334: These test methods should be removed, along with the supporting files.
```suggestion
```

exceptionfactory (changes requested):
Thanks for your patience on this PR @dan-s1. In the course of review, I noticed several disabled test methods. Unless you are planning to adjust them for this PR, they should be removed. As noted to the comments for those tests, those particular styles may not be applicable to JSON, so it is better to leave them out for the initial version.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/test/java/org/apache/nifi/yaml/TestYamlTreeRowRecordReader.java
dan-s1: @exceptionfactory I removed all the disabled tests. It turns out the yaml file mentioned in one of the tests did not actually exist so there was nothing to remove. I assume I must have made the change in the path of the test but never committed the actual yaml file.

exceptionfactory:
@dan-s1 The failure appears to be an intermittent issue based on heap space for the build. We may need to address that separately, but it is not a concern for this PR.

However, I just merged a separate PR adding `StreamConstraints` handling to the JSON Readers, resulting in merge conflicts. Can you rebase on the latest main branch?

dan-s1:
@exceptionfactory There are conflicts (I guess obviously) I will try to resolve and then push.

dan-s1:
@exceptionfactory I am trying to rework my code to handle the changes made in #7823 for NIFI-12153. What I am discovering is that YAML supports comments and there is no way to turn off parsing of comments.  In addition I am trying to mimic `TestJsonTreeRowRecordReader#testReadJSONStringTooLong` and I am trying to use `StreamReadConstraints.builder().maxStringLength(1).build()` yet no exception is thrown even though I have strings longer than one character long. It would seem that aspect only works for Json parsing and not Yaml parsing. I did find the following article [How to parse large YAML file in Java or Kotlin](https://stackoverflow.com/questions/74805240/how-to-parse-large-yaml-file-in-java-or-kotlin) which would seem to handle larger Yaml files but I am not sure if that is the same setting as `StreamReadConstraints.builder().maxStringLength(1).build()`. Please advise.

exceptionfactory:
> @exceptionfactory I am trying to rework my code to handle the changes made in #7823 for NIFI-12153. What I am discovering is that YAML supports comments and there is no way to turn off parsing of comments. In addition I am trying to mimic `TestJsonTreeRowRecordReader#testReadJSONStringTooLong` and I am trying to use `StreamReadConstraints.builder().maxStringLength(1).build()` yet no exception is thrown even though I have strings longer than one character long. It would seem that aspect only works for Json parsing and not Yaml parsing. I did find the following article [How to parse large YAML file in Java or Kotlin](https://stackoverflow.com/questions/74805240/how-to-parse-large-yaml-file-in-java-or-kotlin) which would seem to handle larger Yaml files but I am not sure if that is the same setting as `StreamReadConstraints.builder().maxStringLength(1).build()`. Please advise.

Thanks for the details @dan-s1. If the YAML implementation does not support the same constraints, then I recommend not including those property descriptors on the YAML Record Reader implementation.

MikeThomsen:
@dan-s1 Just out of (genuine) curiosity, what prompted this feature? I have never seen anyone use YAML as anything other than a configuration file format. What horror story/interesting use case prompted it? :-D

dan-s1:
> > @exceptionfactory I am trying to rework my code to handle the changes made in #7823 for NIFI-12153. What I am discovering is that YAML supports comments and there is no way to turn off parsing of comments. In addition I am trying to mimic `TestJsonTreeRowRecordReader#testReadJSONStringTooLong` and I am trying to use `StreamReadConstraints.builder().maxStringLength(1).build()` yet no exception is thrown even though I have strings longer than one character long. It would seem that aspect only works for Json parsing and not Yaml parsing. I did find the following article [How to parse large YAML file in Java or Kotlin](https://stackoverflow.com/questions/74805240/how-to-parse-large-yaml-file-in-java-or-kotlin) which would seem to handle larger Yaml files but I am not sure if that is the same setting as `StreamReadConstraints.builder().maxStringLength(1).build()`. Please advise.
> 
> Thanks for the details @dan-s1. If the YAML implementation does not support the same constraints, then I recommend not including those property descriptors on the YAML Record Reader implementation.

@exceptionfactory Then would it be okay to have in `YamlTreeReader` code like:

```
@Override
        final List<PropertyDescriptor> properties = new ArrayList<>(super.getSupportedPropertyDescriptors());
        //NOTE: Remove those properties which are not applicable for Yaml.
        properties.remove(AbstractJsonRowRecordReader.MAX_STRING_LENGTH);
        properties.remove(AbstractJsonRowRecordReader.ALLOW_COMMENTS);
        
        return properties;
    }
```

dan-s1:
> @dan-s1 Just out of (genuine) curiosity, what prompted this feature? I have never seen anyone use YAML as anything other than a configuration file format. What horror story/interesting use case prompted it? :-D

@MikeThomsen Please refer to the ticket for more details but the submitter's use case was to have ability to convert from Yaml to Json.

exceptionfactory:
> @exceptionfactory Then would it be okay to have in `YamlTreeReader` code like:
> 
> ```
> @Override
>     protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
>         final List<PropertyDescriptor> properties = new ArrayList<>(super.getSupportedPropertyDescriptors());
>         //NOTE: Remove those properties which are not applicable for Yaml.
>         properties.remove(AbstractJsonRowRecordReader.MAX_STRING_LENGTH);
>         properties.remove(AbstractJsonRowRecordReader.ALLOW_COMMENTS);
>         
>         return properties;
>     }
> ```

Thanks for the example @dan-s1, yes, if those properties do not appear to be honored with the YAML implementation, then that is one approach. However, given the number of times that this method will be called, it would be better to define a local static list of supported properties and return that reference in `getSupportedPropertyDescriptors()`. That does require manual maintenance of the list for the YAML Reader, but that is probably better in this case as it should force a review of whether any new property would apply to both the JSON and YAML implementations.

dan-s1:
@exceptionfactory The problem with static is I need many of the properties which are defined in instance methods in `SchemaRegistryService`. I basically want to mimic `getSupportedPropertyDescriptors` from `JsonTreeReader` which is not using any static methods minus the 2 properties which are not applicable for Yaml.

dan-s1:
@exceptionfactory I resolved the conflicts but mistakenly rebased and squashed the commits. I believe all the changes you had requested are all in there. Is that okay?

exceptionfactory:
Thanks for the updates @dan-s1, I will take a closer look soon.

dan-s1:
@exceptionfactory The failed build seems to not have to do with any of my changes.

dan-s1:
@exceptionfactory Any chance this can make 1.24?

nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/TokenParserFactory.java
```diff
@@ -18,13 +18,19 @@
...
-
-    ObjectMapper createCodec(boolean allowComments, StreamReadConstraints streamReadConstraints);
+    /**
+     * Get JSON Parser implementation for provided Input Stream with configured settings
+     *
+     * @param in Input Stream to be parsed
+     * @param streamReadConstraints Stream Read Constraints applied
+     * @param allowComments Whether to allow comments when parsing
+     * @return JSON Parser
+     * @throws IOException Thrown on failures ot read the Input Stream
```
dan-s1, line 33: Instead of
`@throws IOException Thrown on failures ot read the Input Stream`
this should be
`@throws IOException Thrown on failures to read the Input Stream`

nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParserFactory.java
```diff
@@ -23,27 +23,20 @@
...
+        final ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.getFactory().setStreamReadConstraints(streamReadConstraints);
+        if (allowComments) {
+            objectMapper.enable(JsonParser.Feature.ALLOW_COMMENTS);
         }
-        codec.getFactory().setStreamReadConstraints(streamReadConstraints);
-
-        return codec;
+        final JsonFactory jsonFactory = objectMapper.getFactory();
+        return jsonFactory.createParser(in);
```
dan-s1, line 40: The previous code in `AbstractJsonRowRecordReader.java` formally on line 146 when creating the jsonParser also called `setCodec`.  Shouldn't you have something similar here?
Instead of
` return jsonFactory.createParser(in);`

change to
```
final JsonParser jsonParser = jsonFactory.createParser(in);
jsonParser.setCodec(objectMapper);
return jsonParser;
```

nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonRecordSource.java
```diff
@@ -29,6 +30,11 @@
 
 public class JsonRecordSource implements RecordSource<JsonNode> {
     private static final Logger logger = LoggerFactory.getLogger(JsonRecordSource.class);
```
dan-s1, line 32: If the added static final private variables below are uppercase, this should also be for consistency.

dan-s1 (commented):
@exceptionfactory  Great changes! Just had some minor comments.

nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParserFactory.java
exceptionfactory: Setting the codec is not necessary because the `JsonFactory` is coming from the `ObjectMapper`, as opposed to the other way around.

dan-s1: In the previous code also the `JsonFactory` was coming from the `ObjectMapper`and yet `setCodec` was still called. See the previous line 145:
 `jsonParser = codec.getFactory().createParser(in);`

exceptionfactory: Thanks for noting the previous approach. The line calling `setCodec()` was unchanged in the most recent updates for NIFI-12153, but it actually should have been removed as part of the changes to incorporate the constraint settings. So in this case, it is an opportunity to correct the implementation.

exceptionfactory:
n@dan-s1 In the course of runtime testing, I found an issue with the Max String Length property value being `null` in `storePropertyValues` in `JsonTreeReader`. This is understandable since the YamlTreeReader removes that property. I rebased the branch and added one more commit with a `buildStreamReadContraints()` method that `YamlTreeReader` overrides. This approach will be clearer than null-checking the property value. I plan to run a few more tests soon.

exceptionfactory (approved):
Thanks again for the contribution @dan-s1, after making one more adjustment to handle the `Allow Comments` property, the latest version passes all tests and works as expected during runtime verification. +1 merging

dan-s1:
@exceptionfactory Thank you! I am not sure if there is a delay or not but in Jira I still see the status as Patch Available.

exceptionfactory:
You're welcome @dan-s1. I am verifying the changes for the support branch, I updated the Jira to resolved for 2.0.0, and should be able to include 1.24.0 shortly.

PR #7667, opened 2023-09-04
NIFI-12022 Extract verification logic from JMSConnectionFactoryProvider

nandorsoma:
# Summary

[NIFI-12022](https://issues.apache.org/jira/browse/NIFI-12022)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-jms-bundle/nifi-jms-cf-service/src/main/java/org/apache/nifi/jms/cf/AbstractJMSConnectionFactoryProvider.java
+++ b/nifi-nar-bundles/nifi-jms-bundle/nifi-jms-cf-service/src/main/java/org/apache/nifi/jms/cf/AbstractJMSConnectionFactoryProvider.java
@@ -1,0 +1,156 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


nifi-nar-bundles/nifi-jms-bundle/nifi-jms-cf-service/src/main/java/org/apache/nifi/jms/cf/AbstractJMSConnectionFactoryProvider.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Base JMS controller service implementation that provides verification logic.
+ */
+public abstract class AbstractJMSConnectionFactoryProvider extends AbstractControllerService implements JMSConnectionFactoryProviderDefinition, VerifiableControllerService {
+    private static final String ESTABLISH_CONNECTION = "Establish Connection";
+    private static final String VERIFY_JMS_INTERACTION = "Verify JMS Interaction";
+
+    protected volatile IJMSConnectionFactoryProvider delegate;
```
turcsanyip, line 45: @nandorsoma Thanks for extracting the connection verification logic to an abstract parent class which can be used by other JMS controller services too!

However, I don't think that `IJMSConnectionFactoryProvider delegate` can be useful for other implementors because it was created for a specific use case (for sharing the same handler class among `JMSConnectionFactoryProvider` and the JMS processors). So I would leave it as it is in `JMSConnectionFactoryProvider`.

nandorsoma: As we agreed, I will leave the delegate logic as is because it is true that it makes the code unnecessarily complex for simple Providers, but this is the only way to create a parent class for all use cases. Having this "umbrella" is a more significant advantage than what the less complex code could bring.

nandorsoma:
@turcsanyip as a followup we could try to refactor the tests, because I really don't like passing null to the `setConnectionFactoryProperties`.

turcsanyip (approved):
Thanks @nandorsoma!

+1 LGTM
Merging to main and 1.x.

PR #7668, opened 2023-09-04
NIFI-11940 Add SSLContextService to GetSplunk processor

nandorsoma:
# Summary

[NIFI-11940](https://issues.apache.org/jira/browse/NIFI-11940)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-splunk-bundle/nifi-splunk-processors/src/main/java/org/apache/nifi/processors/splunk/GetSplunk.java
+++ b/nifi-nar-bundles/nifi-splunk-bundle/nifi-splunk-processors/src/main/java/org/apache/nifi/processors/splunk/GetSplunk.java
@@ -49,6 +49,8 @@ import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.io.OutputStreamCallback;
 import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.scheduling.SchedulingStrategy;
+import org.apache.nifi.ssl.SSLContextService;
 
 import java.io.BufferedOutputStream;
 import java.io.IOException;
@@ -68,7 +70,6 @@ import java.util.Set;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import org.apache.nifi.scheduling.SchedulingStrategy;
 
 @TriggerSerially
 @InputRequirement(InputRequirement.Requirement.INPUT_FORBIDDEN)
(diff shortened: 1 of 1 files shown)


exceptionfactory (changes requested):
Thanks for the improvement @nandorsoma.

Unfortunately, the design of the Splunk SDK `Service` class is not suited for a configurable `SSLSocketFactory` based on the static configuration approach. Calling the `Service.setSSLSocketFactory` method changes the Socket Factory for all instances of the Splunk Service, which would impact all instances of the Processor, and other Processors that use the Splunk SDK.

One option might be to use instance class-loading, as used in several Hadoop-based Processors. Another more involved option would be the replace the Splunk SDK client with custom network client libraries that support configurable TLS properties.

nandorsoma:
> Thanks for the improvement @nandorsoma.
> 
> Unfortunately, the design of the Splunk SDK `Service` class is not suited for a configurable `SSLSocketFactory` based on the static configuration approach. Calling the `Service.setSSLSocketFactory` method changes the Socket Factory for all instances of the Splunk Service, which would impact all instances of the Processor, and other Processors that use the Splunk SDK.
> 
> One option might be to use instance class-loading, as used in several Hadoop-based Processors. Another more involved option would be the replace the Splunk SDK client with custom network client libraries that support configurable TLS properties.

Thanks for the quick review @exceptionfactory! You are right about the issue regarding static configuration. Thank you for bringing that to my attention!

Lehel44:
Thanks @nandorsoma for the contribution. I tested the SSLContextService with the GetSplunk processor and it works well.
It's important to note that as there isn't an explicit option for configuring SSL modes, using the "https" scheme requires the obligatory use of an SSLContextService. Consequently, the necessity for the new property hinges on the specified scheme value.
Regarding the proposed solution by @exceptionfactory can we move forward with @RequiredInstanceClassloading? I am open to testing it should an update be made.

exceptionfactory:
@Lehel44 I don't think we should move forward with the `RequiresInstanceClassloading` change here unless we do it all for the Splunk Processors, and that raises some other questions. For now, it is still possible to work around this issue by updating the JVM cacerts trust store instead.

joewitt:
@exceptionfactory What is the hesitation to use the instance classloader solution here?  It seems that is what it was made for.  It probably makes sense to ensure that is done on each splunk processor as you note but why avoid it?

exceptionfactory:
> @exceptionfactory What is the hesitation to use the instance classloader solution here? It seems that is what it was made for. It probably makes sense to ensure that is done on each splunk processor as you note but why avoid it?

The optimal solution would involve replacing the Splunk SDK with lower-level HTTP request and response handling. However, that would be a lot more work, so going forward with instance ClassLoader approach is probably the best solution under the circumstances. I withdraw my comment on avoiding that approach.

joewitt:
@nandorsoma @Lehel44 Do you plan to complete the change then to move to instance classloading so we can work past the poor usage of statics for security settings in the splunk libs?

nandorsoma:
> @nandorsoma @Lehel44 Do you plan to complete the change then to move to instance classloading so we can work past the problematic usage of statics for security settings in the splunk libs?

Thanks for bumping it, @joewitt! We were about to close this PR, but as @exceptionfactory withdrew his previous comment, we will push the PR over the finish line.

joewitt:
@nandorsoma do you still plan to complete this?

nandorsoma:
> @nandorsoma do you still plan to complete this?

Yep, @joewitt, I intended to do this on the week starting tomorrow.

nandorsoma:
Thanks @joewitt and @exceptionfactory for the patience. I've added the isolation functionalty and rebased the branch as it was a bit old. Tested it with splunk enterprise edition, works as intended.

exceptionfactory (approved):
Thanks for making the adjustments @nandorsoma, the instance ClassLoading approach looks good. +1 merging

pvillard31:
Do we have plans to do the same for the other Splunk processors (sinks)?

PR #7672, opened 2023-09-08
NIFI-12028 - Add s3.region attribute to ListS3 generated flow files

pvillard31:
# Summary

[NIFI-12028](https://issues.apache.org/jira/browse/NIFI-12028) - Add s3.region attribute to ListS3 generated flow files

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


markap14:
Thanks for putting this up @pvillard31. I did some quick testing, and all looks good with default settings. However, whenever I specify a Record Writer, it looks like the region is not output in the created Record. Can you update the Record-oriented case to output the region also, so that the two stay in-sync?

pvillard31:
Hey @markap14 - yeah so I decided that even if using a record writer, I'd add the region as an attribute instead of adding it to each record since the region would be the same for each record already and I'm not sure to see any value adding it into a record. If you think differently, I can easily add it to the records as well.

MikeThomsen (approved):

PR #7673, opened 2023-09-08
NIFI-11938: Created ConsumeSlack and ListenSlack Processors to consume message events and commands from Slack

markap14:
# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 17

### Licensing

### Documentation


nifi-nar-bundles/nifi-slack-bundle/nifi-slack-processors/src/main/java/org/apache/nifi/processors/slack/consume/ConsumeChannel.java
```diff
@@ -0,0 +1,861 @@
...
+        String minTsValue = stateMap.get(stateKeys.REPLY_MIN_TS);
+        if (minTsValue == null) {
+            minTsValue = latestTs;
+        }
+        final String maxTsValue = stateMap.get(stateKeys.REPLY_MAX_TS);
+        final SlackTimestamp minTs = new SlackTimestamp(minTsValue);
+        final SlackTimestamp maxTs = maxTsValue == null ? new SlackTimestamp() : new SlackTimestamp(maxTsValue);
+        final SlackTimestamp maxParentTs = new SlackTimestamp(latestTs);
+
+        // TODO: Test this heavily
```
pvillard31, line 224: Should it be removed?

joewitt:
Build is good but we need L&N.

For the slack lib added it is at least: https://github.com/SlackAPI/java-slack-sdk/blob/main/LICENSE

That will need to be in nifi-assembly/LICENSE as it is MIT.

Looking at transitive deps and will provide more

joewitt:
I also think these libs pull in some rather dated libraries.  We should likely override them to their newer counterparts as the CVE nature of these is often considerable.  Looking in depth now

exceptionfactory:
> I also think these libs pull in some rather dated libraries. We should likely override them to their newer counterparts as the CVE nature of these is often considerable. Looking in depth now

Thanks for looking into this @joewitt, we should definitely override to the latest versions of items whenever we can, especially when introducing new libraries.

joewitt:
should bump to slack lib 1.32.1

PR #4204, opened 2020-04-12
NIFI-7355 Added Gremlin bytecode client service.

MikeThomsen:
Thank you for submitting a contribution to Apache NiFi.

Please provide a short description of the PR here:

#### Description of PR

_Enables X functionality; fixes bug NIFI-YYYY._

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced 
     in the commit message?

- [ ] Does your PR title start with **NIFI-XXXX** where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically `master`)?

- [ ] Is your initial contribution a single, squashed commit? _Additional commits in response to PR reviewer feedback should be made on this branch and pushed to allow change tracking. Do not `squash` or use `--force` when pushing to allow for clean monitoring of changes._

### For code changes:
- [ ] Have you ensured that the full suite of tests is executed via `mvn -Pcontrib-check clean install` at the root `nifi` folder?
- [ ] Have you written or updated unit tests to verify your changes?
- [ ] Have you verified that the full build is successful on both JDK 8 and JDK 11?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [ ] If applicable, have you updated the `LICENSE` file, including the main `LICENSE` file under `nifi-assembly`?
- [ ] If applicable, have you updated the `NOTICE` file, including the main `NOTICE` file found under `nifi-assembly`?
- [ ] If adding new Properties, have you added `.displayName` in addition to .name (programmatic access) for each of the new properties?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #7677, opened 2023-09-10
NIFI-7355: add bytecode submission to tinkerpop service

levilentz:
# Summary

This brings back https://github.com/apache/nifi/pull/4204 which went stale. In my changes I have been able to address: 

* https://github.com/apache/nifi/pull/4204#discussion_r534548392
* https://github.com/apache/nifi/pull/4204#discussion_r534549148

The new code consolidates the script submissions and gremlin bytecode submission functionality into one service. I am still working to make a unit test to test this with in-memory janusgraph, but I have tested the following extensively in the nifi UI: 

* Script Submission with service settings
* Bytecode Submission with yaml file settings

Using a localized janusgarph server. I am in the process of testing the other two options (script submission with yaml, and bytecode with service settings), but expect them to work as expected. 

I was unable to replicate: 

* https://github.com/apache/nifi/pull/4204#issuecomment-737444733 and would appreciate if @mattyb149  could detail exactly how he was creating that error so I can address

This code is ready for review, and I do not anticipate additional functionality at this time save for the addition of unit tests. 

[NIFI-7355](https://issues.apache.org/jira/browse/NIFI-7355)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

Can you please advise if you want me to squash this commit. I wasnt sure if that was necessary given the fact that I am building off a stale PR. 

# Verification

### Build

- [N/A] Build completed using `mvn clean install -P contrib-check` 
  - [X] JDK 17

I am unable to complete the contrib-check because the neo4j bundle is complaining about in-complete licenses. Should I update those? 

### Licensing

### Documentation


levilentz:
@MikeThomsen  -- bought back an OLD pr from the grave to complete out the tinkerpop client service to allow interaction with tinkerpop utilizing janusgraph. Appears to work on a my personal machine, and will be testing it on our corporate project shortly. 

@mattyb149  -- you have one comment on the previous PR about GStrings. Can you elaborate on how you are using GStrings within your script submission?

MikeThomsen:
@levilentz no problem. I'll hit you up this week if I have any concerns.

mattyb149:
I'll have to find the flow I used to test this, I assume if I was using GStrings it was because I had an expression I wanted to evaluate like `"${fieldName}"` or something, I'll try to take a look soon.

levilentz:
> Thanks for the contribution @levilentz. Without going into the code details for now, there are two items worth highlighting for adjustment:
> 
>     1. There are several Checkstyle warnings shown in the Static Analysis job that need to be corrected:
> 
> 
> ```
> Warning:  src/main/java/org/apache/nifi/graph/TinkerpopClientService.java:[21,43] (imports) AvoidStarImport: Using the '.*' form of import should be avoided - org.apache.nifi.annotation.behavior.*.
> Warning:  src/main/java/org/apache/nifi/graph/TinkerpopClientService.java:[26,34] (imports) AvoidStarImport: Using the '.*' form of import should be avoided - org.apache.nifi.components.*.
> Warning:  src/main/java/org/apache/nifi/graph/TinkerpopClientService.java:[45,20] (imports) AvoidStarImport: Using the '.*' form of import should be avoided - javax.script.*.
> Warning:  src/main/java/org/apache/nifi/graph/TinkerpopClientService.java:[47,17] (imports) AvoidStarImport: Using the '.*' form of import should be avoided - java.util.*.
> Warning:  src/test/java/org/apache/nifi/graph/TestableGremlinClientService.java:[20,8] (imports) UnusedImports: Unused import - org.apache.tinkerpop.gremlin.driver.Client.
> ```
> 
>     2. New unit tests should use Java instead of Groovy, so the `GremlinByteCodeClientIT` class should be refactored.

Thank you for the quick turn @exceptionfactory  -- I have just simply removed the groovy IT tests in the most recent commit and have fixed the check style.

levilentz:
One other thing I have done is I am not re-using the traversal source per this documentation: 

https://docs.aws.amazon.com/neptune/latest/userguide/best-practices-gremlin-java-reuse.html

In testing, if we recreate the traversal source for each query, it is time-intensive and uses a lot of overhead inside of the JVM. I will be testing this again locally here shortly, but this has worked on a custom build of nifi for quite some time. Let me know your thoughts!

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/pom.xml
```diff
@@ -110,6 +110,28 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-jsr223</artifactId>
+            <version>3.0.17</version>
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-dateutil</artifactId>
+            <version>3.0.17</version>
```
exceptionfactory, line 127: These version references should use the project property named `nifi.groovy.version`:

```suggestion
            <version>${nifi.groovy.version}</version>
        </dependency>
        <dependency>
            <groupId>org.codehaus.groovy</groupId>
            <artifactId>groovy-jsr223</artifactId>
            <version>${nifi.groovy.version}</version>
        </dependency>
        <dependency>
            <groupId>org.codehaus.groovy</groupId>
            <artifactId>groovy-dateutil</artifactId>
            <version>${nifi.groovy.version}</version>
```

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/pom.xml
```diff
@@ -110,6 +110,28 @@
...
+        </dependency>
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+            <version>1.15</version>
```
exceptionfactory, line 132: This version can be removed since it is specified in the parent Maven configuration.
```suggestion
```

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,538 @@
...
+            .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
+            .build();
+
+    public static final PropertyDescriptor PASSWORD = new PropertyDescriptor.Builder()
+            .name("password")
+            .displayName("Password")
+            .description("The password used to authenticate with the gremlin server." +
+                    " Note: when using a remote.yaml file, this password (if set) will override any password set in the yaml file")
+            .required(false)
+            .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
```
exceptionfactory, line 175: Sensitive properties should not support expression language, instead, they can be supplied through Parameter Contexts if necessary.
```suggestion
```

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,538 @@
...
+                    "This configuration property has no effect unless a value for the Extension JAR field is " +
+                    "also provided.")
+            .dependsOn(EXTRA_RESOURCE)
+            .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
+            .dependsOn(CONNECTION_SETTINGS, YAML_SETTINGS)
+            .required(false)
+            .build();
+
+    public static final PropertyDescriptor SSL_CONTEXT_SERVICE = new PropertyDescriptor.Builder()
+            .name("tinkerpop-ssl-context-service")
```
exceptionfactory, line 207: Recommend removing the prefix for the property name.
```suggestion
            .name("ssl-context-service")
```

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,538 @@
...
+    protected Cluster.Builder setupSSL(ConfigurationContext context, Cluster.Builder builder) {
+        if (context.getProperty(SSL_CONTEXT_SERVICE).isSet()) {
+            SSLContextService service = context.getProperty(SSL_CONTEXT_SERVICE).asControllerService(SSLContextService.class);
+            builder
+                    .enableSsl(true)
+                    .keyStore(service.getKeyStoreFile())
+                    .keyStorePassword(service.getKeyStorePassword())
+                    .keyStoreType(service.getKeyStoreType())
+                    .trustStore(service.getTrustStoreFile())
+                    .trustStorePassword(service.getTrustStorePassword());
```
exceptionfactory, line 354: Instead of passing individual properties, the Builder supports an `sslContext()` method that accepts a Netty SslContext. The Netty JdkSslContext wraps the Java SSLContext returned from the `SSLContextService.createContext()` method.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,538 @@
...
+
+            return resultAttributes;
+
+        } catch (Exception ex) {
+            throw new ProcessException(ex);
+        }
+    }
+
+    protected Map<String, String> bytecodeSubmission(String s, Map<String, Object> map, GraphQueryResultCallback graphQueryResultCallback) {
+        String hash = DigestUtils.md5Hex(s);
```
exceptionfactory, line 448: MD5 is not a secure hashing function. Although this is not a cryptographic function, recommend using SHA-256.

exceptionfactory (commented):
Thanks for making the adjustments @levilentz. I noted a few other general recommendations regarding dependency versioning and other issues.

The custom JAR and class loading warrants some further evaluation to make sure it behaves as expected, but I see the `RequiresInstanceClassLoading` annotation, so that is good.

I will defer to others for more detailed implementation evaluation.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,538 @@
...
+
+    @Override
+    public Collection<ValidationResult> customValidate(ValidationContext context) {
+        Collection<ValidationResult> results = new ArrayList<>();
+        boolean jarsIsSet = !StringUtils.isEmpty(context.getProperty(EXTRA_RESOURCE).getValue());
+        boolean clzIsSet = !StringUtils.isEmpty(context.getProperty(EXTENSION_CLASSES).getValue());
+
+        if (jarsIsSet && clzIsSet) {
+            try {
+                final ClassLoader loader = this.getClass().getClassLoader();
```
MikeThomsen, line 306: I don't think this works on 2.x and Java 17. These are the changes I had to make to another project.

image

@exceptionfactory thoughts?

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,538 @@
...
+        return builder;
+    }
+
+
+    public void loadClasses(ConfigurationContext context) {
+        String path = context.getProperty(EXTRA_RESOURCE).getValue();
+        String classList = context.getProperty(EXTENSION_CLASSES).getValue();
+        if (path != null && classList != null && !path.isEmpty() && !classList.isEmpty()) {
+            try {
+                ClassLoader loader = Thread.currentThread().getContextClassLoader();
```
MikeThomsen, line 367: See above. Code that worked for me with nifi-mock didn't work in the live framework.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/resources/docs/org.apache.nifi.graph.GremlinBytecodeClientService/additionalDetails.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
```
MikeThomsen, line 1: This needs to be redone to match the TinkerpopClientService.

MikeThomsen (changes requested):

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/test/java/org/apache/nifi/graph/GremlinClientServiceIT.java
```diff
@@ -42,20 +42,20 @@ public void setup() throws Exception {
         clientService = new TestableGremlinClientService();
         runner = TestRunners.newTestRunner(NoOpProcessor.class);
         runner.addControllerService("gremlinService", clientService);
-        runner.setProperty(clientService, AbstractTinkerpopClientService.CONTACT_POINTS, "localhost");
```
MikeThomsen, line 45: This class should be renamed and the tests expanded a bit to flex the functionality.

MikeThomsen (changes requested):
Another minor change needed.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
MikeThomsen: I think MD5 is fine for this because the goal is to just make it quick and easy to uniquely identify compiled scripts.

joewitt:
@MikeThomsen @exceptionfactory I have been messing with this very module for the past 24 hours or so because I'm trying to move all of our Groovy work to be more Java 21 compliant and also our Groovy 3x stuff is past support/progress.  Nothing in the Groovy 3x line appears to work with Java 21.  So on main i'm trying for now to move all things to Groovy 4.  The Tinkerpop/graph stuff has an even older relationship with Groovy and seems to be on entirely unsupported lines.  I am finding information gathering on Groovy super tough frankly.

Anyway, I suggest we remove the groovy support from the graph module entirely.  We should in my view move to a mode whereby the only Groovy bits in our codebase are to allow people to write scripted components in groovy.  No unit tests should be in Groovy and we should eliminate those that are.  I did not realize what a complicated story this gave our build process.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/pom.xml
```diff
@@ -45,7 +46,7 @@
         <dependency>
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-utils</artifactId>
-            <version>2.0.0-SNAPSHOT</version>
+            <version>${project.version}</version>
```
joewitt, line 49: I do not recommend we change from 2.0.0-SNAPSHOT to ${project.version}.  In my experience this breaks during the release process.  The explicit versions get automatically handled.

joewitt:
@levilentz please take a look here https://github.com/apache/nifi/pull/7677#issuecomment-1719643027.    cc @mattyb149

joewitt:
@levilentz @mattyb149 @exceptionfactory @MikeThomsen Alternatively I propose I put up this PR which moves all the things to Groovy 4.x in a few mins here.  And dumps Groovy support from the graph processors.  Then you rebase to that once merged and attempt to restore Groovy support but on the ${nifi.groovy.version} basis which will be Groovy 4.0.15.  I was getting an error about Java 17 which itself is a problem since we're at the very least Java 17 (though trying to get to Java 21). 

How does that sound?

levilentz:
> @levilentz @mattyb149 @exceptionfactory @MikeThomsen Alternatively I propose I put up this PR which moves all the things to Groovy 4.x in a few mins here. And dumps Groovy support from the graph processors. Then you rebase to that once merged and attempt to restore Groovy support but on the ${nifi.groovy.version} basis which will be Groovy 4.0.15. I was getting an error about Java 17 which itself is a problem since we're at the very least Java 17 (though trying to get to Java 21).
> 
> How does that sound?

@joewitt obviously I have not been as plugged into this ongoing conversation, but a couple of thoughts for this specific component: 

* This new functionality really needs some sort of JVM injection support, as that is where the performance gain is from a gremlin-perspective. We have used groovy because it was the easiest way to interact with: https://github.com/apache/nifi/blob/main/nifi-nar-bundles/nifi-graph-bundle/nifi-graph-processors/src/main/java/org/apache/nifi/processors/graph/ExecuteGraphQueryRecord.java 
* From a gremlin perspective, pinning the groovy version is really important as all the tinkerpop tooling are built on top of specific groovy versions (as it is all groovy based), so it is causing headaches making specific builds work when it is pinned to the groovy version in tinkerpop. Therefore, flexibility to use a specific groovy version in this component would be really neat. 
* As for groovy tests, I am already working on refactoring these and removing all groovy tests

So to get to your proposal, I dont have any issues with that as long as I have some guidance on how the ExecuteGraphQueryRecordProcessor will work after that refactor. If you have a PR for me to review that would be immensely helpful.

joewitt:
@levilentz Thanks that is helpful.  My goal is getting NiFi to a good build basis for Java 21.  The Groovy stuff in its various forms is causing problems in doing that as these old Groovy versions neither build nor run on Java 21 (even Java 17 seems dubious).  We can avoid pinning them as it has been however where things get problematic is when a service depends on very old versions.

What is the newest/most recent version of Groovy that Tinkerpop/Gremlin can work with?

joewitt:
@levilentz here is the PR i just put up.  https://github.com/apache/nifi/pull/7692/commits

Assuming this works we can merge my PR.  Then you could rebase and try to do your changes on top of that.  You can put Groovy back in and do your update.  The problem is going to be that we can't really stick with Groovy versions that are too old.  Even when trying to run this stuff on Java 17 it failed to build due to not being able to understand Java class version 61 (which is 17) so something is going to have to give/break here.

levilentz:
> @levilentz Thanks that is helpful. My goal is getting NiFi to a good build basis for Java 21. The Groovy stuff in its various forms is causing problems in doing that as these old Groovy versions neither build nor run on Java 21 (even Java 17 seems dubious). We can avoid pinning them as it has been however where things get problematic is when a service depends on very old versions.
> 
> What is the newest/most recent version of Groovy that Tinkerpop/Gremlin can work with?

@joewitt  -- oh boy, that does not sound fun! We might need to think on that a bit as tinkerpop's development cycle is much slower. 

Per their release notes, gremlin 3.7.0 is on Groovy 4.0.9 -- https://github.com/apache/tinkerpop/blob/master/CHANGELOG.asciidoc

joewitt:
@levilentz Ok that is a great new thing then.  If my PR gets merged can you then rebase and start again adding the needed Groovy libs back in?  Tests that use groovy will be a no go I think as the compiler for it appears non functional but you could see.   But since they're on 4.0.9 we'll be on 4.0.15 that should be fine....   You agree?

levilentz:
@joewitt -- just reviewed that PR, and it does not contain anything that is giving me heartburn. I will need to address your same concern as @exceptionfactory  is keying off the same here: 

https://github.com/apache/nifi/pull/7677#discussion_r1323108590

So yeah, lets plan on me working on this PR (need to address comments, documentation, and then unit tests), and will keep an eye out on https://github.com/apache/nifi/pull/7692, rebasing after it is merged.

joewitt:
awesome thank you @levilentz

mattyb149:
+1 for going to Groovy 4, I'll review Joe's PR too. Are you going to switch from gremlin-groovy to gremlin-java? What is the impact for things like [Janusgraph](https://docs.janusgraph.org/) that only appear to support Gremlin as the query language?

levilentz:
> +1 for going to Groovy 4, I'll review Joe's PR too. Are you going to switch from gremlin-groovy to gremlin-java? What is the impact for things like [Janusgraph](https://docs.janusgraph.org/) that only appear to support Gremlin as the query language?

@mattyb149 to answer this it really depends on what is going to happen to the other scripted processors that use groovy. Do you have any guidance on what that will look like for nifi as a whole? 

In my opinion, depending on the guidance, moving away from groovy will have minimal impact on this. For the scripted version of this processor, there is no groovy dependency. For the bytecode version, I have been running a custom build of this processor for ~ 3 years, which leverages java-gremlin to load data into janusgraph. As long as I have a way to inject the graph traversal source, it shouldent really matter if the underlying code is compiled groovy or a set of standardized methods on a custom jar. 

How we do that will just depend on what nifi is doing with the other groovy-injected processors such as: https://github.com/apache/nifi/blob/main/nifi-nar-bundles/nifi-groovyx-bundle/nifi-groovyx-processors/src/main/java/org/apache/nifi/processors/groovyx/ExecuteGroovyScript.java

joewitt:
The move to Groovy 4.0.15 is in main now so you can probably rebase and try again

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/pom.xml
levilentz: Implemented.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/pom.xml
levilentz: Removed.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: Removed.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: Removed.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/pom.xml
exceptionfactory: The latest version still shows Groovy 3.0.17

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: I had tested this code using JDK17, and it loaded the custom classes just fine, however if this is the more acceptable use case, I can switch it over and test.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: I think moving from md5 will help with security findings when deployed within a secure environment. So moving to SHA-256.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services-nar/src/main/resources/META-INF/NOTICE
```diff
@@ -108,13 +108,11 @@ The following binary components are provided under the Apache Software License v
   (ASLv2) Exp4j
     Copyright 2017
 
-  (ASLv2) Groovy 2.4.16 (http://www.groovy-lang.org)
-        groovy-2.4.16-indy
-        groovy-json-2.4.16-indy
-        groovy-sql-2.4.16-indy
+  (ASLv2) Groovy 2.5.10 (http://www.groovy-lang.org)
```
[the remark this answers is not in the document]
levilentz: @MikeThomsen  -- L&N check?

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: @exceptionfactory  -- I have updated this to utilize the `.createContext()` functionality. I am not super familiar with netty, so let me know if you see any problems with `JdkSslContext` that I have created.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/test/java/org/apache/nifi/graph/GremlinClientServiceIT.java
levilentz: Tracking, I am planning on building out integration tests utilizing test containers, will be doing that shortly after I verify the most recent changes work with my local test env.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: Tested and this works as expected.

mattyb149:
There's a Checkstyle error failing the tests

levilentz:
> There's a Checkstyle error failing the tests

@mattyb149  -- sorry still working on cleaning this up, should have something in by thursday addressing all PR comments.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services-nar/src/main/resources/META-INF/NOTICE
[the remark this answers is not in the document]
MikeThomsen: License and notice.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: So one odd thing, in all my testing, it doesnt matter if I set EXTENSION_CLASSES to be any specific classes I need, they seem to be available as long as I have the classes available inside of the Jar(s) present in EXTENSION_CLASSES. I have included unit test that proves this. Any thoughts on just removing the EXTENSION_CLASSES? I could imaging keeping it as we would need functionality such as this if we are to move away from Groovy entirely within NiFi.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services-nar/src/main/resources/META-INF/NOTICE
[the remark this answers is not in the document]
levilentz: @MikeThomsen do you have any guidance on how to do that? I looked at other processors/controllers that used groovy and they seem similarly out of compliance with your comment. So just not sure how best to generate this.

levilentz:
@mattyb149  @joewitt  @exceptionfactory @MikeThomsen  -- ready for another review. I have done the following: 

1. Updated documentation to better show how to use this component
2. Addressed all outstanding PR concerns on the code (need some guidance on the licensing stance) 
3. Build out a unit test showing the gremlin bytecode functionality

I was not able to get test containers working with JG, so I might still tinker with that to see if I can get the whole test suite working. 

For that integration test utilizing gremlin bytecode that I have built out, it needs to load a gremlin serializer, which requires setting the custom jar. I have one that does that, but it is ~15MB in size, so I dont think it is wise to add it to the git history. Do you have any guidance on how best to create that jar on the fly for the test? Or should I just put some documentation on how to download the jar to run the test? It just simply is the standard serializers available in the gremlin driver package, but they are not loaded by default in the controller.

joewitt:
@levilentz I'd just clean up the test you mention to be an Integration test that someone can run with instructions or just remove it.  Can you please squash these down to a single commit to make merging easier.  I prefer to avoid using 'squash and merge' as I'd like to validate in a local build . I started grabbing a commit at a time and it was creating conflicts.  THanks!

levilentz:
@joewitt  -- I have cleanup the integration test. There is technically a way to use it with janusgraph-core in memory and convert it to an actual unit test, but it will require this to have a lot of O&M to continue working as new version of groovy/gremlin/java come out. I would recommend leaving it as a simple IT. I have added documentation on how to run it, and someone knowledgeable about this process would be able to run it. 

I have also squashed all the commits. Let me know if you have any other changes you would like to add to this!

joewitt:
Ok.  The workflow needed approval to run again so just clicked that.  Meanwhile if we can get @mattyb149 @MikeThomsen to weight in or merge that would be great.  Given all the discussion it seems all are good so with that understanding I'll plan to merge tomorrow if things build fine and nobody has alternative input.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,545 @@
...
+            .description("A selection that toggles for between script submission or as bytecode submission")
+            .allowableValues(SCRIPT_SUBMISSION, BYTECODE_SUBMISSION)
+            .defaultValue("script-submission")
+            .required(true)
+            .build();
+
+    public static final PropertyDescriptor CONNECTION_SETTINGS = new PropertyDescriptor.Builder()
+            .name("connection-settings")
+            .displayName("Settings Specification")
+            .description("Selecting \"Service-Defined Settings\" connects using the setting on this service. Selecting \"Yaml Settings\" uses the specified yaml file for connection settings. ")
```
mattyb149, line 109: References to YAML should be `YAML` in all uppercase except for file extensions. Also should the display name be 
"Connection Settings Specification" for more clarity?

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,545 @@
...
+            .allowableValues(SERVICE_SETTINGS, YAML_SETTINGS)
+            .defaultValue("service-settings")
+            .required(true)
+            .build();
+
+    public static final PropertyDescriptor CONTACT_POINTS = new PropertyDescriptor.Builder()
+            .name("tinkerpop-contact-points")
+            .displayName("Contact Points")
+            .description("A comma-separated list of hostnames or IP addresses where an OpenCypher-enabled server can be found.")
```
mattyb149, line 118: References to OpenCypher should be Tinkerpop or Gremlin

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,545 @@
...
+                results.add(new ValidationResult.Builder().subject(EXTENSION_CLASSES.getDisplayName()).valid(false).explanation(ex.toString()).build());
+            }
+        }
+
+        boolean standardConfigIsSet = context.getProperty(CONTACT_POINTS).isSet()
+                && context.getProperty(PATH).isSet()
+                && context.getProperty(PORT).isSet();
+        boolean fileIsSet = context.getProperty(REMOTE_OBJECTS_FILE).isSet();
+
+        if (standardConfigIsSet && fileIsSet) {
```
mattyb149, line 330: Do we need all of these if the properties are dependent?

mattyb149 (commented):
Looks good, mostly cosmetic/doc stuff but was curious if we can get rid of groovy-jsr223 and use the Grovy-specific classes for evaluating and compiling scripts.

joewitt:
@levilentz Do you think you'll be able to knock out the rebase and @mattyb149 commentary?  This thing looks close and want to ensure we can get it across the finish line.  You've definitely hung in there quite well through a lot on this one!  Thanks

joewitt:
Awesome thanks!  Dont mean to rush you - just dont want to let this linger given how much effort you put in plus how much harder i made it with the java 21/groovy updates :). Thanks

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,545 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+
+@Tags({"graph", "gremlin"})
+@CapabilityDescription("This service interacts with a tinkerpop-compliant graph service, providing both script submission and bytecode submission capabilities. " +
+        "Script submission is the default, with the script command being sent to the gremlin server as text. This should only be used for simple interactions with a tinkerpop-compliant server " +
+        "such as counts or other operations that do not require the injection of custom classed. " +
+        "Bytecode submission allows much more flexibility. When providing a jar, custom serializers can be used and pre-compiled graph logic can be utilized by groovy scripts" +
+        "provided by processors such as the ExecuteGraphQueryRecordProcessor.")
+@SeeAlso(classNames = {"ExecuteGraphQueryRecordProcessor"})
```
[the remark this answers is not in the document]
levilentz: Updated.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: Updated throughout.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: Updated to gremlin.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: This was erroneous. Removed several of the validation checks.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/pom.xml
```diff
@@ -90,6 +90,39 @@
...
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.groovy</groupId>
+            <artifactId>groovy</artifactId>
+            <version>${nifi.groovy.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.groovy</groupId>
+            <artifactId>groovy-jsr223</artifactId>
```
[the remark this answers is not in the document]
levilentz: @mattyb149 I have tried removing it, but it fails with a NPE, because it uses `javax.script.ScriptEngineManager.getEngineByName("groovy")` to compile the groovy script. See:

https://github.com/apache/nifi/pull/7677/files#diff-bade439cf2e90169d5771ee9bf9a3df690c07e80384309141e66c3aad42d422aR253

And: 

https://github.com/apache/nifi/pull/7677/files#diff-bade439cf2e90169d5771ee9bf9a3df690c07e80384309141e66c3aad42d422aR458

Do you have a recommended way to do this?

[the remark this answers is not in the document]
levilentz: @mattyb149  -- was able to refactor the code to remove this! Sorry if you got a notification of a previous message, some googling around got me around the issue.

levilentz:
@joewitt was able to get to all of @mattyb149, and my integration tests work as expected. However, because I have had to remove the JSR223 per: 

https://github.com/apache/nifi/pull/7677#discussion_r1335286584

I would like an additional review if possible. A lot of this groovy compilation and injection is new to me. In the mean time, I will be doing a complete test a running nifi instance.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,525 @@
...
+    private GraphTraversalSource traversalSource;
+    private boolean scriptSubmission = true;
+    boolean usesSSL;
+    protected String transitUrl;
+
+    @OnEnabled
+    public void onEnabled(final ConfigurationContext context) throws InitializationException {
+        loadClasses(context);
+        GroovyClassLoader loader = new GroovyClassLoader(this.getClass().getClassLoader());
+        groovyShell = new GroovyShell(loader, new Binding());
```
MikeThomsen, line 252: @mattyb149 I think [this](https://docs.groovy-lang.org/latest/html/api/groovy/lang/GroovyShell.html#%3Cinit%3E(java.lang.ClassLoader)) is the safer constructor.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -0,0 +1,525 @@
...
+        if (getLogger().isDebugEnabled()) {
+            getLogger().debug(map.toString());
+        }
+
+        Binding bindings = new Binding();
+        map.forEach(bindings::setProperty);
+        bindings.setProperty("g", traversalSource);
+        bindings.setProperty("log", getLogger());
+        try {
+            Object result = groovyShell.parse(s, bindings).run();
```
MikeThomsen, line 469: You don't appear to be using the pre-compiled code to do the submission. Am I missing something in how `groovyShell` works here?

MikeThomsen (changes requested):
Got a few questions.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: Updated to this constructor.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: @MikeThomsen  -- good catch, I updated it to use the compiled code.

levilentz:
@joewitt @MikeThomsen @mattyb149 I have updated the code with your suggested changes, and they are working great! Let me know if you have any other changes.

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -140,7 +139,7 @@ public class TinkerpopClientService extends AbstractControllerService implements
             .required(true)
             .defaultValue("/gremlin")
             .addValidator(StandardValidators.NON_EMPTY_EL_VALIDATOR)
```
mattyb149, line 141: Is it possible to configure a Gremlin server such that the path is empty? If so we should replace this validator with either Validators.TRUE or something like that

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
```diff
@@ -235,9 +234,10 @@ public class TinkerpopClientService extends AbstractControllerService implements
             SSL_CONTEXT_SERVICE
     ));
 
-    private ScriptEngineManager MANAGER = new ScriptEngineManager();
-    private ScriptEngine engine;
-    private Map<String, CompiledScript> compiledCode;
+//    private ScriptEngineManager MANAGER = new ScriptEngineManager();
```
mattyb149, line 237: Please delete dead code

mattyb149 (changes requested):
Looking good, just want to get rid of that dependency on the graph processors. Thanks for sticking with this! I'm looking forward to the capability

nifi-nar-bundles/nifi-graph-bundle/nifi-other-graph-services/src/main/java/org/apache/nifi/graph/TinkerpopClientService.java
levilentz: @mattyb149  -- https://tinkerpop.apache.org/javadocs/current/core/org/apache/tinkerpop/gremlin/driver/Cluster.Builder.html#path(java.lang.String) looks like it defaults to "/gremlin" if not provided, so I think we can safely change this validator.

joewitt:
Seems like the various rounds of commentary are sorted and the build checks out.  Went ahead and merged to main.  Thanks!  Anything to follow can happen in a new JIRA/PR

PR #7760, opened 2023-09-19
NIFI-12038 introduce PackageFlowFile processor and fix an NPE in FlowFilePackagerV3

mosermw:
# Summary

[NIFI-12038](https://issues.apache.org/jira/browse/NIFI-12038)
New PackageFlowFile processor, which will package FlowFile attributes and content into an output FlowFile that can be exported from NiFi and imported back into NiFi, preserving the original attributes and content.

While testing this, I found and fixed an issue with FlowFilePackagerV3 where an attribute with a null value causes NullPointerException.  See NIFI-4298 for how to get attributes with null value.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFilePackagerV3.java
+++ b/nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFilePackagerV3.java
@@ -36,7 +36,7 @@ public class FlowFilePackagerV3 implements FlowFilePackager {
         } else {
             writeFieldLength(out, attributes.size()); //write out the number of attributes
             for (final Map.Entry<String, String> entry : attributes.entrySet()) { //write out each attribute key/value pair
-                writeString(entry.getKey() == null ? "" : entry.getKey(), out);
+                writeString(entry.getKey(), out);
                 writeString(entry.getValue() == null ? "" : entry.getValue(), out);
             }
         }
--- a/nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFilePackagerV3.java
+++ b/nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFilePackagerV3.java
@@ -36,8 +36,8 @@ public class FlowFilePackagerV3 implements FlowFilePackager {
         } else {
             writeFieldLength(out, attributes.size()); //write out the number of attributes
             for (final Map.Entry<String, String> entry : attributes.entrySet()) { //write out each attribute key/value pair
-                writeString(entry.getKey(), out);
-                writeString(entry.getValue(), out);
+                writeString(entry.getKey() == null ? "" : entry.getKey(), out);
+                writeString(entry.getValue() == null ? "" : entry.getValue(), out);
             }
         }
 
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PackageFlowFile.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PackageFlowFile.java
@@ -1,0 +1,146 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.standard;
+
+import org.apache.nifi.annotation.behavior.InputRequirement;
+import org.apache.nifi.annotation.behavior.SideEffectFree;
+import org.apache.nifi.annotation.behavior.SupportsBatching;
+import org.apache.nifi.annotation.behavior.WritesAttribute;
+import org.apache.nifi.annotation.behavior.WritesAttributes;
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.SeeAlso;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.components.PropertyDescriptor;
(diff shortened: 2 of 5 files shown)


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PackageFlowFile.java
```diff
@@ -0,0 +1,146 @@
...
+@CapabilityDescription("This processor will package FlowFile attributes and content into an output FlowFile that can be exported from NiFi"
+        + " and imported back into NiFi, preserving the original attributes and content.")
+@WritesAttributes({
+        @WritesAttribute(attribute = "mime.type", description = "The mime.type will be changed to application/flowfile-v3"),
+        @WritesAttribute(attribute = "filename", description = "The filename will be appended with '.pkg'")
+})
+@SeeAlso({UnpackContent.class, MergeContent.class})
+public class PackageFlowFile extends AbstractProcessor {
+
+    static final String PACKAGE_FILENAME_EXTENSION = ".pkg";
```
mh013370, line 66: Is there any concern about conflicts with existing uses of the `.pkg` file extension (namely for macOS)?

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PackageFlowFile.java
```diff
@@ -0,0 +1,146 @@
...
+                }
+            }
+        });
+
+        packagedFlowFile = session.putAllAttributes(packagedFlowFile, attributes);
+        session.transfer(packagedFlowFile, REL_SUCCESS);
+        session.transfer(flowFiles, REL_ORIGINAL);
+    }
+
+    private String createFilename(final List<FlowFile> flowFiles) {
```
mh013370, line 139: Do you think users would prefer more positive control over the name of the output filename, perhaps via an attribute?

exceptionfactory: For keeping the functionality simple, it may be better to leave the filename unchanged. A subsequent UpdateAttribute Processor could handle changes if needed.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PackageFlowFile.java
mosermw: Thank you both for looking at this!  For changing the filename, I followed what MergeContent does when creating a FlowFile-V3 package [1].  If you feel that PackageFlowFile should operate differently, I can change it, but MergeContent consistency was my motivation.

[1] - https://github.com/apache/nifi/blob/main/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/MergeContent.java#L884

exceptionfactory: Thanks for the background reference @mosermw. Given that this new Processor only supports FlowFile V3, I think that is a reason to avoid the filename adjustment in the initial version.

mosermw: Sounds good, I will make the changes so it does not modify the filename attribute.

nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFilePackagerV3.java
```diff
@@ -36,8 +36,8 @@ public void packageFlowFile(final InputStream in, final OutputStream out, final
         } else {
             writeFieldLength(out, attributes.size()); //write out the number of attributes
             for (final Map.Entry<String, String> entry : attributes.entrySet()) { //write out each attribute key/value pair
-                writeString(entry.getKey(), out);
-                writeString(entry.getValue(), out);
+                writeString(entry.getKey() == null ? "" : entry.getKey(), out);
```
joewitt, line 39: @mosermw how could a flowfile attribute key be null?  Do we allow that somehow?

mosermw: I don't think the NiFi framework allows null for an attribute key.  However, the Java HashMap allows a null key.  I felt that FlowFilePackagerV3 is generic enough that it could be used in unit tests where the framework isn't present to enforce that restriction.  Essentially, I changed this line to be safe in "you never know" situations.

I thought of not outputting an attribute with a null key at all, but that would throw off the number of attributes written by `writeFieldLength` on line 37.

joewitt: I see.  Yeah if that were just test code then fair enough.  But it being in the actual packager and the way it handles it makes it seem like that would be ok/handled.  We dont want to give that impression.  If you detect a null key you should reject the whole thing (the given flowfile/attrs are not valid and should not be packaged).  I recommend simply not doing this null check and fixing anything that tried to pass nulls.

mosermw: OK that makes sense.  Successfully handling a null key allows bad behavior.  Developers need to be placed in timeout if they cause a NullPointerException!

joewitt: Thanks - otherwise it seemed ready to roll.  Happy to merge to both lines.  I'll keep an eye for a commit that dumps the null thing

nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFilePackagerV3.java
```diff
@@ -37,7 +37,7 @@ public void packageFlowFile(final InputStream in, final OutputStream out, final
             writeFieldLength(out, attributes.size()); //write out the number of attributes
             for (final Map.Entry<String, String> entry : attributes.entrySet()) { //write out each attribute key/value pair
                 writeString(entry.getKey(), out);
-                writeString(entry.getValue(), out);
+                writeString(entry.getValue() == null ? "" : entry.getValue(), out);
```
joewitt, line 40: @mosermw we gotta dump this line based on last convo.  i'll do it locally prior to merge.

joewitt: err this is the actual value.  i dont recall whether we enforce that or not

PR #7768, opened 2023-09-21
NIFI-12101: Fixed bugs that caused StatelessBasicsIT.testChangeFlowVe…

markap14:
…rsion to fail. There were a couple of related issues. Firstly, for Stateless Groups we were not waiting for the group to fully stop before attempting to update the flow but were waiting only for child/descendant groups to stop. This was due to misleading naming in the ProcessGroupUpdateStrategy enum. Renamed enum values to more clearly convey what they mean. Additionally, renamed the enum to more clearly convey what how it is used, since it is used not only for updating groups but also for retrieving groups. Additionally, when waiting for stateless groups to stop, we inadvertently checked the Set of affectedComponents' IDs against Process Group IDs so addressed the logic there and also ensured that we were consistent in setting the AffectedComponentEntity's ProcessGroup (i.e., its ProcessGroupNameDTO) as we were inconsistently setting it across the codebase.

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


markap14:
Of note, in order to verify the fix, you can update the AbstractPort class such that the shutdown method sleeps before making the port as stopped:
```
    @Override
    public void shutdown() {
        try { Thread.sleep(5000L); } catch (final Exception e) {}
        
        scheduledState.set(ScheduledState.STOPPED);
        logger.info("{} shutdown", this);
    }

```
This will cause the system test to fail consistently with the patch applied, as NiFi will attempt to update the group before the ports have stopped. Once applying the patch, NiFi will appropriately wait for the ports to stop before attempting to update the flow.

exceptionfactory (commented):
Thanks for working to address this issue @markap14.

Unfortunately it looks like the integration test is still failing intermittently, and other tests also failed in the initial system test builds:

```
Error:    ParameterContextIT.testParamChangeWhileReferencingControllerServiceDisabling » Timeout testParamChangeWhileReferencingControllerServiceDisabling() timed out after 5 minutes
Error:    ParameterContextIT.testParamChangeWhileReferencingProcessorStartingButInvalid » Timeout testParamChangeWhileReferencingProcessorStartingButInvalid() timed out after 5 minutes
Error:    ParameterContextIT.testProcessorRestartedWhenParameterChanged » Timeout testProcessorRestartedWhenParameterChanged() timed out after 5 minutes
Error:    ParameterContextIT.testProcessorRestartedWhenParameterChangedWhenReferencedThroughEL » Timeout testProcessorRestartedWhenParameterChangedWhenReferencedThroughEL() timed out after 5 minutes
Error:    ParameterContextIT.testProcessorStartedAfterLongValidationPeriod » Timeout testProcessorStartedAfterLongValidationPeriod() timed out after 5 minutes
Error:    RegistryClientIT.testChangeVersionOnParentThatCascadesToChild » Timeout testChangeVersionOnParentThatCascadesToChild() timed out after 5 minutes
Error:    RegistryClientIT.testControllerServiceUpdateWhileRunning » Timeout testControllerServiceUpdateWhileRunning() timed out after 5 minutes
Error:    RegistryClientIT.testStartVersionControlThenModifyAndRevert » Timeout testStartVersionControlThenModifyAndRevert() timed out after 5 minutes
Error:    StatelessBasicsIT.testChangeFlowVersion » Timeout testChangeFlowVersion() timed out after 5 minutes
```

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-client-dto/src/main/java/org/apache/nifi/web/api/entity/ProcessGroupRecursivity.java
```diff
@@ -16,7 +16,7 @@
  */
 package org.apache.nifi.web.api.entity;
 
-public enum ProcessGroupUpdateStrategy {
-    CURRENT_GROUP,
-    CURRENT_GROUP_WITH_CHILDREN
+public enum ProcessGroupRecursivity {
```
exceptionfactory, line 19: This name change impacts the REST API parameters, which also impacts UI JavaScript code. It would be better to separate these changes into a distinct PR specific to the main branch so that the resolution for the integration test issue could be backported to the support branch.

markap14:
Yeah, sorry about that @exceptionfactory I did push without running through all of the system tests. I ran only the ones that I thought were affected, but I've updated the PR and ensured that all pass now. 
Also - good call on the support line. I had overlooked that. But there are a few things different between the two, so I think it's best to leave this PR as-is and generate a new PR for the support line. I'll push up a new PR for that one in a bit.

markap14:
Actually @exceptionfactory nevermind about the support branch - this fix does not need to go to the support branch, as it affects 2.0 only. The issue is in the StatelessBasicsIT, which tests running a Process Group using the Stateless Execution Engine. So it does not apply to the 1.x line.

joewitt:
@exceptionfactory if you're close to merging go ahead but i'm doing a local build and will also run sys tests. Github CI looks great so at this point we're clear to fire.  I'll take it unless you're already done

exceptionfactory (changes requested):
Thanks for addressing the issues with the other system tests @markap14, that looks good.

Changing the value names in the `ProcessGroupRecursivity` enum still requires some adjustments to the JavaScript code, so either those changes need to be incorporated in this pull request, or the name changes should be reverted for now to avoid having the REST API out of sync with the UI.

exceptionfactory:
> @exceptionfactory if you're close to merging go ahead but i'm doing a local build and will also run sys tests. Github CI looks great so at this point we're clear to fire. I'll take it unless you're already done

Thanks @joewitt, the system test changes themselves look good, but I just noted that JavaScript UI code changes are also necessary to align with renaming the enum values.

markap14:
Thanks @exceptionfactory just realized that I didn't push that change. Will push another commit in one moment

exceptionfactory:
> Thanks @exceptionfactory just realized that I didn't push that change. Will push another commit in one moment

Thanks for the quick turnaround @markap14, that looks like that should cover the necessary changes. I will wait for one more system test run and then plan to merge!

joewitt:
@exceptionfactory sorry i was already well into this.  merged to main!

exceptionfactory:
> @exceptionfactory sorry i was already well into this. merged to main!

No problem, latest run looks good, thanks @joewitt and @markap14!

PR #7777, opened 2023-09-22
NIFI-11177: Add defensive code for null values for Iceberg

mattyb149:
# Summary
[NIFI-11177](https://issues.apache.org/jira/browse/NIFI-11177) This PR adds defensive checks to return null values when any conversions occur

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
@@ -238,7 +238,7 @@ public class DataTypeUtils {
 
     public static UUID toUUID(Object value) {
         if (value == null) {
-            throw new IllegalTypeConversionException("Null values cannot be converted to a UUID");
+            return null;
         }
 
         if (value instanceof UUID) {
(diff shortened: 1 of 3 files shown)


mark-bathori (commented):
Thanks @mattyb149 for fixing this null handling issue. Since every converter needs a null check in the `convert(Object data)` method I think it would be a better solution to add this null check in a new method in the `DataConverter<S, T>` abstract class and then this would call to the `convert` method. With this change we don't need to add the same null check to every implementation.

mattyb149:
This should work since the top-level `convert()` is for each RecordField in the Record, but shouldn't we have the null checks in the individual `convert()s` too just in case `convert()` is called directly from somewhere else? Or do we trust in having the Record/Map/Array converters call the new method (I'm calling it `convertWithNullCheck()`)?

mark-bathori (approved):
It shouldn't be called from other places but of course it could happen. I'm fine with the current solution also.
LGTM +1

mattyb149:
I'll push a commit with the new method and change the complex types to call that, just for safety :)

mark-bathori:
I think the last commit is not needed since the original solution contains null check in every `convert` implementation, so we have 2 options here:

1. Add null check in every `convert` method implementation, so the `convertWithNullCheck` is not needed.
2. Use the `convertWithNullCheck`, make the original `convert` method private and remove the null checks from the implementations.

mark-bathori:
I think the last commit is not needed since the original solution contains null check in every `convert` implementation. The only issue could be if a new implementation would be added without a null check but it is very unlikely.

pvillard31 (approved):

PR #7782, opened 2023-09-22
NIFI-12116 Configured ObjectMapper to not allow multiline JSON.

dan-s1:
# Summary

[NIFI-12116](https://issues.apache.org/jira/browse/NIFI-12116)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
@@ -726,6 +726,7 @@
                         <exclude>src/test/resources/TestUpdateRecord/schema/person-with-name-string-fields.avsc</exclude>
                         <exclude>src/test/resources/TestUpdateRecord/schema/person-with-name-string.avsc</exclude>
                         <exclude>src/test/resources/TestUpdateRecord/schema/person-with-stringified-name.avsc</exclude>
+                        <exclude>src/test/resources/TestValidateJson/multiline.json</exclude>
                         <exclude>src/test/resources/TestValidateJson/schema-simple-example-missing-required.json</exclude>
                         <exclude>src/test/resources/TestValidateJson/schema-simple-example-unmatched-pattern.json</exclude>
                         <exclude>src/test/resources/TestValidateJson/schema-simple-example.json</exclude>
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
@@ -726,7 +726,6 @@
                         <exclude>src/test/resources/TestUpdateRecord/schema/person-with-name-string-fields.avsc</exclude>
                         <exclude>src/test/resources/TestUpdateRecord/schema/person-with-name-string.avsc</exclude>
                         <exclude>src/test/resources/TestUpdateRecord/schema/person-with-stringified-name.avsc</exclude>
-                        <exclude>src/test/resources/TestValidateJson/multiline.json</exclude>
                         <exclude>src/test/resources/TestValidateJson/schema-simple-example-missing-required.json</exclude>
                         <exclude>src/test/resources/TestValidateJson/schema-simple-example-unmatched-pattern.json</exclude>
                         <exclude>src/test/resources/TestValidateJson/schema-simple-example.json</exclude>
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateJson.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateJson.java
@@ -17,6 +17,7 @@
 package org.apache.nifi.processors.standard;
 
 import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.networknt.schema.JsonSchema;
@@ -167,6 +168,7 @@ public class ValidateJson extends AbstractProcessor {
     static {
         MAPPER = new ObjectMapper();
         MAPPER.configure(JsonParser.Feature.ALLOW_COMMENTS, true);
+        MAPPER.configure(DeserializationFeature.FAIL_ON_TRAILING_TOKENS, true);
     }
 
     private JsonSchema schema;
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateJson.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateJson.java
@@ -46,6 +46,7 @@ import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.processor.AbstractProcessor;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
+import org.apache.nifi.processor.ProcessorInitializationContext;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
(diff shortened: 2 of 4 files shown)


markap14:
-1
@dan-s1 this is a super common format that we don't want to consider invalid. While I do think it's reasonable to have the option to consider that invalid, we should not be changing the behavior like this. I'd be fine with adding a property to the processor to make it optional, though.

joewitt:
@dan-s1 it is basically this https://jsonlines.org/ and it is super weirdly common in the wild so at this point while it isn't valid JSON is valid 'JSON Lines'.  We need to keep supporting it.

dan-s1:
The problem though without my change the ObjectMapper only reads the first object in the incoming file and validates against that.  I wouldn't think that is the correct behavior either.

ChrisSamo632:
> The problem though without my change the ObjectMapper only reads the first object in the incoming file and validates against that. I wouldn't think that is the correct behavior either.

So what we could do with is some sort of `ValidateJsonRecord` then, i.e. use a `JsonTreeReader` to iterate through all JSON Objects within a FlowFile, then validate each of those and output the results. `ValidateRecord` would be a pattern to consider following here, just that we'll be applying the JSON Schema validation rather than the Avro Schema validation currently used by `ValidateRecord`

r-vandenbos:
@ChrisSamo632 making this processor record based is something I have been working on and have mostly finished, I have some validation message handling to finish and unit tests.

r-vandenbos:
@dan-s1 apologies I didn't create one as I did the work a couple of hours after finding the issue. I will create one today / tomorrow or I can just use yours?

dan-s1:
@r-vandenbos My use case has to incorporate the ability to view the document as one document of JSON. I was very interested to have @markap14 suggestion incorporated whether to treat the flow file as a single JSON document or as JSON lines. Also how are you handling if one line (record) fails but all the other pass?

dan-s1:
@ChrisSamo632 Did you mean to actually make these records? I do not think JsonTreeReader iis needed here, Jackson has  `JsonMapper` which could allow for iteration of all the JSON objects thereby allowing for validating each object.

ChrisSamo632:
> @ChrisSamo632 Did you mean to actually make these records? I do not think JsonTreeReader iis needed here, Jackson has `JsonMapper` which could allow for iteration of all the JSON objects thereby allowing for validating each object.

Whether a RecordReader is used/needed here is a question for the implementation - what I'm getting at is the need to iterate through all Objects within the FlowFile - typically, that's done with a Reader in most Processors, but it doesn't *have* to be done that way here.

One option would be to use a Reader (which means we could have a file format other than JSON as input) then convert to JSON as needed within the processor in order to validate with a JSON Schema (instead of an Avro Schema as currently done in `ValidateRecord`). But that's maybe going too far/not the use case we're needing to address here. Yes, another option is to use a `JsonMapper` (which is basically what the JsonTreeReader is doing, right?) to iterate through the Objects to validate each in turn.

If `ValidateJson` is updated, then an option will indeed be needed to `Validate Whole` or `Validate Each` (whatever you want to call the options). Again, my suggesiton would be to look at `ValidateRecord` for how it separates the Record-based output to `valid` and `invalid` relationships depending upon whether they meet the defined schema.

dan-s1:
@ChrisSamo632 ValidateRecord allows for splitting up the flowfile and sending records to both the valid and invalid relationships. That is now what we wanted necessarily in our use case but I am not sure how NIFI would like to implement this.
Also this processor should only take JSON as JSON schema is meant to validate JSON and has the ability to confirm more than just field and type checking as the Avro schema does.

ChrisSamo632:
> JSON schema is meant to validate JSON and has the ability to confirm more than just field and type checking as the Avro schema does.

That's be an argument (from my perspective) to use JSON Schema to validate other types of data as well. JSON Schema is obviousl written with JSON validation being the main aim, but there's no real reason (that I'm aware of) that it couldn't be used to do the same thing for other data formats. NiFi users could then use JSON Schemas (and/.or Avro Schemas) to validate FlowFile content rather than having to use the `ScriptedValidateRecord` processor, for example

But, again, that's probably more of a stretch than what's needed here/now, and one to consider for the future. Splitting the output to `valid`/`invalid` would follow the same pattern as for `ValidateRecord`, so would seem to make sense for consistency. There might be other patterns the community wich to adopt instead though

dan-s1:
@joewitt @markap14 @exceptionfactory Can you please share what you feel this processor should do? I wanted to use @markap14 original idea to have a property whether to support JSON lines or not. After that should this processor be patterned after ValidateRecord as @ChrisSamo632 suggested or not i.e. allowing for splitting up the flowfile? Thank you!

ChrisSamo632:
@dan-s1 more that in my original comment here, I actually suggested a new processor, e.g. `ValidateJsonRecord`, which would be an option to provide alternative functionality

But I'd also be mindful to not introduce more processors/functionality unnecessarily, as it could lead to more confusion for users

The use cases to address would apart to be:
1. Validate an entire FlowFile against a JSON Schema
2. Validate every JSON Object (or Record in standard NiFi parlance) from within a FlowFile against a JSON Schema - the suggestion is to do this for all Objects and provide a much information as possible as the result, e.g. split the FlowFile content (as appropriate) between `valid` & `invalid` relationships

r-vandenbos:
@ChrisSamo632 @dan-s1 I have followed the ValidateRecord pattern and used a JsonTreeReader to inspect each record regardless if it is formatted JSON or NDJSON. Each record is then validated against the schema and based on the result it is appended to a valid or invalid flowfile which in the end are sent to their relevant relationships. I set record.count attributes for both flow files and the invalid flow file has an error message attribute (which I need to finish).

@dan-s1 what is your use case for needing the whole file routed to invalid? As I write a record count attribute for valid records could you not use a RouteOnAttribute processor to validate if the record.count on the valid flow file is the same as the total record count (which I am presuming you have or can get)? If they are not the same then handle that in your flow?

N.B Attribute names are up for change.

dan-s1:
@ChrisSamo632 Your suggestion of a new processor e.g. `ValidateJsonRecord` seems to be similar to what NIFI has already for Jolt, `JoltTransformJSON` which handles single documents and `JoltTransformRecord` which can handle multiple  documents.  Though I would have to wonder if moving forward in 2.x does NIFI want to support both `JoltTransformJSON` and `JoltTransformRecord` or only `JoltTransformRecord`?

dan-s1:
@r-vandenbos I would like to go with @ChrisSamo632 suggestion of two processors. Can you please submit a ticket for your work but creating a new processor that can handle records and I will fix ValidateJson with @markap14 original suggestion of a property. As I mentioned before, this seems to align with what NIFI already has for Jolt `JoltTransformJSON` which handles single documents and `JoltTransformRecord` which can handle multiple documents. Hopefully now there will be ValidateJson and ValidateJsonRecord.

markap14:
@dan-s1 @ChrisSamo632 @r-vandenbos I do think we want to be careful here. Breaking it up into two processors that really do essentially the same thing is probably a mistake in my opinion. I think it makes much more sense to have a single processor and let a property determine whether or not it's expecting multiple JSON documents. We'd also want to avoid any mention of a "JSON Record" - the entire point of Records is that it doesn't matter whether the bytes are in JSON, XML, Logs/plaintext, Avro, etc. They are records.

It would definitely make sense to me to take the approach that Chris pointed out - to have two relationships - valid and invalid, where each individual JSON element is routed to one or the other. So each incoming FlowFile would be split into, at most, 2 FlowFiles. This would be consistent with how ValidateRecord works.

Jolt is broken into two different processors because JoltTransformJSON was created before NiFi had a notion of records. So then we created JoltTransformRecord in order to allow the same JOLT language to be used for Avro, XML, etc.

exceptionfactory (changes requested):
I agree with your comments and concerns @markap14. I don't think we should add a ValidateJsonRecord Processor.

In its current form, this ValidateJson Processor makes sense as expecting a well-formed JSON node, whether an object or array. From that perspective, formats like newline-delimited JSON are not valid, because that is in fact multiple JSON objects.

This discussion and others highlight a general interest in using JSON Schema, which has a number of benefits.

As an incremental step to more involved changes, I think there is value in clarifying the current behavior, which is that NDJSON is not supported. If we don't want to make that change incrementally, then I think a larger PR that implements cohesive support for FlowFile or Record-level validation is the best way forward.

joewitt:
since i was asked i'll just state I don't have any special input here worth sharing.  There are more than enough thoughtful perspectives in this so I'll stay out of it at this point.

dan-s1:
@exceptionfactory Is then the only thing needed for this PR some documentation indicating that NDJSON is not supported?

exceptionfactory (commented):
@dan-s1 Thanks for following up on this.

Yes, my recommendation is to simply update the description with a sentence stating that "This Processor does not support input containing multiple JSON objects, such as newline-delimited JSON." All other changes in this PR would be reverted. That would set a baseline for where things stand today.

exceptionfactory:
> @exceptionfactory Can I also add that if the flow file does contain multi line json that only the first object will be validated? The full text I would like to add is
> 
> > This Processor does not support input containing multiple JSON objects, such as newline-delimited JSON. If the flow file contains newline-delimited JSON only the first line will be validated.

That sounds good!

exceptionfactory (approved):
Thanks for updating the documentation based on the re-scoping of this pull request @dan-s1, the update looks good. +1 merging

dan-s1:
@exceptionfactory Should @r-vandenbos now create a new ticket and PR for his work to make ValidateJson support records?

exceptionfactory:
@dan-s1 Yes, a separate Jira issue would be the next step.

Based on the discussion, there are a couple ways forward.

1. Update ValidateJson to support NDJSON, validating each object using the configured JSON Schema settings
2. Update ValidateRecord to support validating Records using JSON Schema

The second option would have larger benefits, with the ability to validate any kind of Record using JSON Schema. It also has greater complexity, as it requires some translation approach from NiFi Records to objects that the JSON Schema and understand.

It would be helpful to outline the general strategy and discuss further details on a new Jira issue before opening a new pull request, as there are some important details to consider.

PR #7785, opened 2023-09-24
NIFI-12074 updated images to base on Java 21 and added Python3

joewitt:
# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-docker/dockerhub/DockerRun.sh
+++ b/nifi-docker/dockerhub/DockerRun.sh
@@ -22,4 +22,4 @@ DOCKER_IMAGE="$(grep -Ev '(^#|^\s*$|^\s*\t*#)' DockerImage.txt)"
 NIFI_IMAGE_VERSION="$(echo "${DOCKER_IMAGE}" | cut -d : -f 2)"
 
 echo "Running Docker Image: ${DOCKER_IMAGE}"
-docker run -d --name "nifi-${NIFI_IMAGE_VERSION}" -p 8080:8080 -p 8443:8443 -p 10000:10000 -p 8000:8000 -p 8181:8181 "${DOCKER_IMAGE}"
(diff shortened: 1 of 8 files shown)


joewitt:
The resulting image grew from 1.7GB to 2.2GB.  Most of this is from the inclusion of Python3 which we'll want so users can leverage the new Python capabilities in NiFi 2.x.  We will need to find ways to reduce the overall size more as we go.

exceptionfactory (commented):
Thanks for putting together these changes @joewitt! Switching to the BellSoft [Liberica JDK](https://bell-sw.com/libericajdk/) image looks good, which aligns with the default JDK that Spring Framework builds use for container images.

I also agree with disabling plain HTTP access given the inherent security issues. We should evaluate removing plain HTTP access completely in a separate issue, but these changes look good.

I would like to do some testing, but the changes look straightforward on initial review. If others get to reviewing this before I have the opportunity to test, no need to wait for further review, but I will try to follow up soon.

markap14:
Thanks @joewitt and @exceptionfactory . I did some light testing. I was able to verify that it started up as expected running Java 21 and that Python is installed. Unfortunately, though, it does not uncomment the line in nifi.properties needed in order to enable Python processors. Probably should get that line uncommented in the docker image. Otherwise, all looks good to me.

markap14:
Thanks for the update @joewitt . Looks good to me. Was able to bring it up, pointing to my local branch's `nifi-nar-bundles/nifi-py4j-bundle/nifi-python-test-extensions/src/main/resources/extensions` directory as a mount and it successfully started up and allowed me to start using Python processors. +1 will merge to main.

PR #7796, opened 2023-09-26
NIFI-12104 Separate a non-atomic Redis DMC implementation from the ex…

bbende:
…isting one for use in Put/Fetch DMC when Redis is clustered

# Summary

[NIFI-12104](https://issues.apache.org/jira/browse/NIFI-12104)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/main/java/org/apache/nifi/redis/service/RedisDistributedMapCacheClientService.java
+++ b/nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/main/java/org/apache/nifi/redis/service/RedisDistributedMapCacheClientService.java
@@ -18,63 +18,30 @@ package org.apache.nifi.redis.service;
 
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.Tags;
-import org.apache.nifi.annotation.lifecycle.OnDisabled;
-import org.apache.nifi.annotation.lifecycle.OnEnabled;
-import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.ValidationContext;
 import org.apache.nifi.components.ValidationResult;
-import org.apache.nifi.controller.AbstractControllerService;
-import org.apache.nifi.controller.ConfigurationContext;
 import org.apache.nifi.distributed.cache.client.AtomicCacheEntry;
 import org.apache.nifi.distributed.cache.client.AtomicDistributedMapCacheClient;
 import org.apache.nifi.distributed.cache.client.Deserializer;
 import org.apache.nifi.distributed.cache.client.Serializer;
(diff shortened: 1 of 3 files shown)


nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/main/java/org/apache/nifi/redis/service/RedisDistributedMapCacheClientService.java
```diff
@@ -332,43 +127,4 @@ public <K, V> boolean replace(final AtomicCacheEntry<K, V, byte[]> entry, final
...
-    }
-
-    private <K> byte[] serialize(final K key, final Serializer<K> keySerializer) throws IOException {
-        final ByteArrayOutputStream out = new ByteArrayOutputStream();
-
-        keySerializer.serialize(key, out);
-        return out.toByteArray();
-    }
-
-    private <T> T withConnection(final RedisAction<T> action) throws IOException {
```
MikeThomsen, line 358: The connection handling looks the same to me. Am I missing something? I am not seeing where one use is transactional, but the other isn't.

MikeThomsen (commented):
LGTM overall, but need a little clarification on how you used the Redis API.

nifi-nar-bundles/nifi-redis-bundle/nifi-redis-extensions/src/main/java/org/apache/nifi/redis/service/RedisDistributedMapCacheClientService.java
bbende: The transactional code is in the methods from `AtomicDistributedMapCache` that are implemented in `RedisDistributedMapCacheClientService`. Since we can't rename the existing service (it would be better to name that one `AtomicRedisDistributedMapCacheClientService`), we can leave that one alone and move the methods that are from only `DistributedMapCacheClientService` down to `SimpleRedisDistributedMapCacheClientService`. This if a processor declares that it only needs `DistributedMapCacheClient` (not atomic), then it can use the Simple Redis one even when Redis is clustered.

MikeThomsen (approved):
Change seems sound to me. Not a Redis expert by any means, but LGTM.

MikeThomsen:
The cherry-pick to 1.x was kinda messy. Looks like `CONNECTION_POOL` was moved to `RedisUtils` in another commit. I'll try to chase that down later.

gresockj (commented):
Code looks good, and I tested out that the `SimpleRedisDistributedMapCacheClientService` works with a clustered Redis connection pool service.

bbende:
Thanks @MikeThomsen and @gresockj 

I was able to resolve the cherry-pick to `support/nifi-1.x`, it was because there is one change for a new redis processor that was not back-ported there, so worked around that.

PR #7809, opened 2023-09-28
NIFI-12131 Fix potential NPE when FlowAnalyzer and RuleViolationsManager are not set.

tpalfy:
Fix NPE: Revamped RuleViolationsManager and FlowAnalyzer handling to make sure no issue occurs when these are not set. Fix ResourceNotFoundException: When determining the subject permissions for a rule violation the type of the subject is now known, so we try to lookup all possible types. Non-matching tpyes throw a ResourceNotFoundException exception though. Going to ignore those.

# Summary

[NIFI-12131](https://issues.apache.org/jira/browse/NIFI-12131)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/flow/AbstractFlowManager.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/flow/AbstractFlowManager.java
@@ -59,6 +59,7 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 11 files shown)


exceptionfactory (commented):
Thanks for working on this issue @tpalfy.

Can you clarify under what conditions the FlowAnalyzer and RuleViolationsManager would be null? Now that these are core framework components, it seems like it might be better to have alternative implementations, as opposed to making them optional at multiple points in the framework-core-api.

tpalfy:
@exceptionfactory I'm not entirely familiar with all possible NiFi runtimes but in general only the regular NiFi would have FlowAnalyzer and RuleViolationsManager set. At least that was the intent.
All other runtimes (MiNiFi, Stateless NiFi) would have these as null.

I think Optional servers exactly this kind of purpose.
Setting any kind of dummy implementation would require us to identify all the injection and/or explicit instantiation points and we would need to maintain the dummy classes. It would lead to a more convoluted design and implementation in my opinion.

exceptionfactory (approved):
Thanks for the reply @tpalfy. On evaluation of other methods and implementations like `StatelessFlowManager`, this looks like a helpful way forward. +1 merging

PR #7811, opened 2023-09-29
NIFI-12099 Repurpose Template icon in top left bar

elcsiga:
# Summary

[NIFI-12099](https://issues.apache.org/jira/browse/NIFI-12099)

https://github.com/apache/nifi/assets/1295926/364ef977-26b8-4eaf-ae19-29ed32ef2dc7

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


sardell:
It looks like the draggable icons are no longer centered in their container, which is more obvious when hovering over an icon.

Before your changes:

before

After:
after

rfellows:
reviewing...

rfellows:
This is good and works pretty well. The only improvement I could recommend would be to refresh the `current-user` after configuring a registry client. As it stands now, we are relying on the background polling of the current-user endpoint to update the current user which is where it is determined if `canVersionFlows` is true or false. It is this property that determines if we show the `No Registry Client available` dialog or the `Import Version` dialog.
If you delete the registry client and then try to drag on a new import before the polling happens, you can be shown import dialog with no registries:
Screenshot 2023-10-18 at 3 35 34 PM

Similarly, if you do the same too soon after configuring a registry client, you can get a stale `No Registry Client available` dialog even though you have one configured.

Maybe @scottyaslan can provide some more insight into if this is consistent with other areas of the application.

elcsiga:
@rfellows good point, now adding or removing a registry client also reloads `/current-user`

rfellows (approved):
Changes look good to me 👍 
Thanks @elcsiga

PR #7816, opened 2023-09-29
NIFI-12149 Create nifi-redis-utils and minor improvements to util methods

bbende:
This is the 1.x version of #7812


gresockj (approved):
Thanks for this update, looks good and still works.

gresockj:
Cherry-picked manually

PR #7817, opened 2023-09-29
NIFI-12151 Fixed StandardPrivateKeyService fails due to missing Bounc…

turcsanyip:
…yCastleProvider

# Summary

[NIFI-12151](https://issues.apache.org/jira/browse/NIFI-12151)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-services/nifi-key-service-bundle/nifi-key-service/src/test/java/org/apache/nifi/key/service/StandardPrivateKeyServiceTest.java
+++ b/nifi-nar-bundles/nifi-standard-services/nifi-key-service-bundle/nifi-key-service/src/test/java/org/apache/nifi/key/service/StandardPrivateKeyServiceTest.java
@@ -21,7 +21,8 @@ import org.apache.nifi.util.NoOpProcessor;
 import org.apache.nifi.util.TestRunner;
 import org.apache.nifi.util.TestRunners;
 
-import org.bouncycastle.openssl.PKCS8Generator;
+import org.bouncycastle.asn1.ASN1ObjectIdentifier;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.bouncycastle.openssl.jcajce.JcaPEMWriter;
 import org.bouncycastle.openssl.jcajce.JcaPKCS8Generator;
 import org.bouncycastle.openssl.jcajce.JceOpenSSLPKCS8EncryptorBuilder;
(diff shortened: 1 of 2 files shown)


nifi-nar-bundles/nifi-standard-services/nifi-key-service-bundle/nifi-key-service/src/test/java/org/apache/nifi/key/service/StandardPrivateKeyServiceTest.java
```diff
@@ -118,6 +134,21 @@ void testGetPrivateKeyEncryptedKey() throws Exception {
...
+                AES_128_CBC,
+                AES_192_CBC,
+                AES_256_CBC,
+                DES3_CBC,
+                PBE_SHA1_RC4_128,
+                PBE_SHA1_RC4_40,
+                PBE_SHA1_3DES,
+                PBE_SHA1_2DES,
+                PBE_SHA1_RC2_128,
+                PBE_SHA1_RC2_40
```
exceptionfactory, line 148: It is not necessary to test all of these algorithms, particularly those algorithms where the only difference is the key size. I recommend a much smaller list:
```suggestion
                AES_256_CBC,
                DES3_CBC
```

exceptionfactory (changes requested):
Thanks for tracking down this issue and providing a clean resolution @turcsanyip. The code changes look good, the only recommendation is scoping down the list of algorithms tested.

nifi-nar-bundles/nifi-standard-services/nifi-key-service-bundle/nifi-key-service/src/test/java/org/apache/nifi/key/service/StandardPrivateKeyServiceTest.java
turcsanyip: Thanks @exceptionfactory for your review!

Yes, it makes sense to narrow down the encryption algorithms to be tested.
However, I would keep one more: `PBE_SHA1_3DES`
This was the original test input and it ends up in the [PKCS12](https://github.com/bcgit/bc-java/blob/7349980f39aa26541f69d99e66416e84b9f7c005/pkix/src/main/java/org/bouncycastle/openssl/jcajce/JceOpenSSLPKCS8DecryptorProviderBuilder.java#L101) branch, so it is handled differently than `AES_256_CBC` and `DES3_CBC`.

exceptionfactory (approved):
Thanks for adjusting the set of algorithms @turcsanyip, I agree with the list included. +1 merging

PR #7826, opened 2023-10-01
NIFI-12157 Remove nifi-logging-utils and NiFiLog

exceptionfactory:
# Summary

[NIFI-12157](https://issues.apache.org/jira/browse/NIFI-12157) Removes the `nifi-logging-utils` module and replaces `NiFiLog` wrapper references with direct references to the standard SLF4J `LoggerFactory` used throughout the repository.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/util/ClientResponseUtils.java
```diff
@@ -41,7 +40,7 @@ public static void drainClientResponse(final Response response) {
                 bis = new BufferedInputStream(response.readEntity(InputStream.class));
                 IOUtils.copy(bis, new NullOutputStream());
             } catch (final IOException ioe) {
-                logger.info("Failed clearing out non-client response buffer due to: " + ioe, ioe);
+                logger.info("Failed clearing out non-client response buffer", ioe);
```
joewitt, line 43: We should avoid logging an exception at info level.  In this case the output should be warn as it happening would be interesting to know why we weren't able to drain the response.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/firewall/impl/FileBasedClusterNodeFirewall.java
```diff
@@ -56,7 +55,7 @@ public class FileBasedClusterNodeFirewall implements ClusterNodeFirewall {
```
joewitt, line 55: On Line 105 for unknown host we should avoid including the exception.  We know precisely why it was unknown and we're logging unknown host itself.  In previous behavior the stack would not print without debug so this will preserve that plus intent anyway.

exceptionfactory:
Thanks for the detailed notes on those methods @joewitt, I made the proposed adjustments and pushed the changes.

PR #7829, opened 2023-10-02
NIFI-12161: While waiting for node to startup, time out after 5 minut…

markap14:
…es and throw an Exception

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/scheduling/StandardProcessScheduler.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/scheduling/StandardProcessScheduler.java
@@ -73,6 +73,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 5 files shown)


joewitt:
Should change the title of the PR and JIRA now given what this is finding...

joewitt:
system tests ran successfully on linux.  running again for good measure as they never pass two times in a row :)

exceptionfactory:
@markap14 Are you planning to keep the additional logging added, or remove some of it once the tests pass consistently?

joewitt:
i closed my laptop before they finished so it timed out/closed.  but it was going well.  this seems much more stable

markap14:
@joewitt thanks. I agree, I think the stability is significantly improved. @exceptionfactory I think the logging is helpful. I considered changing much of it to DEBUG. However, what we find is that when something does time out, it's often very often intermittent so changing the log level to DEBUG and then triggering it to happen again is generally not easy to accomplish and can take a huge amount of time. While most of the log levels would not be appropriate for a running application, I feel like INFO level log messages are fair for system tests to denote what it's waiting on. So I think it makes sense to keep the log levels as they are.

exceptionfactory:
> @exceptionfactory I think the logging is helpful. I considered changing much of it to DEBUG. However, what we find is that when something does time out, it's often very often intermittent so changing the log level to DEBUG and then triggering it to happen again is generally not easy to accomplish and can take a huge amount of time. While most of the log levels would not be appropriate for a running application, I feel like INFO level log messages are fair for system tests to denote what it's waiting on. So I think it makes sense to keep the log levels as they are.

Thanks for the reply @markap14, on further review, I see the the logging statements are primarily in the test suite class, so more logging there makes sense for debugging.

PR #7852, opened 2023-10-06
NIFI-12185 Add nifi-bom Bill of Materials

exceptionfactory:
# Summary

[NIFI-12185](https://issues.apache.org/jira/browse/NIFI-12185) Adds a `nifi-bom` Bill of Materials module at the root level containing a Maven configuration with shared dependencies.

The `nifi-bom` defines a limited set of dependencies with specific scopes designed for extension components. The purpose of `nifi-bom` is to support both internal and external extensions with consistent versioning for shared interfaces.

This pull request applies the `nifi-bom` to the `nifi-commons` and `nifi-nar-bundles` parent projects. This approach uncovered a number of implicit transitive dependencies on `nifi-property-utils`, containing the `StringUtils` class, and `nifi-utils`, which provides common validators and other utilities. The `nifi-property-utils` module is available to all components based on its presence in the `lib` directory, and `nifi-standard-services-api-nar` already bundles `nifi-utils`. Inheriting the defined scope and version from `nifi-bom` provided the opportunity to surface these dependencies explicitly in a number of extension Maven configurations.

The introduction of `nifi-bom` also provided the opportunity to remove explicit version numbers and scopes from most referencing modules. This reduces the number of changes required when releasing new project versions.

The initial scope of `nifi-bom` is limited to public and framework API modules, as well as a subset of modules under the `nifi-commons` directory. For `nifi-commons` in particular, there is a smaller subset of modules with wider usage and greater stability, providing sufficient reason to include them in `nifi-bom`.

The nature of this new `nifi-bom` module limits the scope to the main branch, targeted for NiFi 2.0.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/minifi/minifi-nar-bundles/minifi-provenance-repository-bundle/minifi-provenance-repositories/pom.xml
+++ b/minifi/minifi-nar-bundles/minifi-provenance-repository-bundle/minifi-provenance-repositories/pom.xml
@@ -31,5 +31,9 @@ limitations under the License.
             <artifactId>nifi-data-provenance-utils</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-api</artifactId>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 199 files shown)


markap14:
Thanks @exceptionfactory I think this is a good idea, using a separate bom module instead of putting huge numbers of dependency versions in dependency management. This way, others can depend on that instead of having to add nifi as a parent. The cleanup that you did here, eliminating unnecessary versions in pom's and explicitly declaring the otherwise-implicit dependencies is very helpful! +1 merged to main. Thanks!

PR #7858, opened 2023-10-08
NIFI-12178 add integration-tests and docker-tests github actions

ChrisSamo632:
# Summary

[NIFI-12178](https://issues.apache.org/jira/browse/NIFI-12178) add integration-tests and docker-tests github actions

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

- ~[ ] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)`~

### Documentation


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,248 @@
...
+    -pl -:nifi-runtime-manifest-test
+    -pl -:nifi-stateless-assembly
+    -pl -:nifi-stateless-processor-tests
+    -pl -:nifi-stateless-system-test-suite
+    -pl -:nifi-system-test-suite
+    -pl -:nifi-nar-provider-assembly
+    -pl -:nifi-py4j-integration-tests
+# skip individual failing ITs (some won't work in a CI environment)
+# always run the (failsafe default) IT*, *IT, *ITCase integration-tests
+  MAVEN_FAILSAFE_SKIP_ITS: >-
```
exceptionfactory, line 78: Did you consider enumerating these failing tests in a profile, or is that not possible at a root level? It would be helpful to run a similar profile in a local build environment.

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,164 @@
...
+      - name: Clean Modules
+        env:
+          MAVEN_OPTS: >-
+            ${{ env.DEFAULT_MAVEN_OPTS }}
+        # clean non-assembly modules before running Docker builds and delete unpacked binaries, otherwise we'll run out of disk space
+        run: >
+          ${{ env.MAVEN_COMMAND }}
+          ${{ env.MAVEN_CLEAN_ARGUMENTS }}
+          ${{ env.MAVEN_CLEAN_PROJECTS }}
+          && rm -rf nifi-assembly/target/nifi-*-bin/ nifi-registry/nifi-registry-assembly/target/nifi-registry-*-bin/
```
exceptionfactory, line 156: Instead of manually removing the binaries, building the registry-assembly binary should be disabled for this particular workflow.

exceptionfactory (changes requested):
Thanks for working on these additional workflows @ChrisSamo632. Having automated execution of working integration tests and Docker builds will be very useful.

On initial review, I raised several implementation questions. At a high level, it would be very useful to run integration tests locally, so it seems like a more Maven-centric approach to integration test selection would be better. That may present some challenges, but I was wondering whether you considered those options. I also have a few questions on some of the particular build steps and scripting commands.

.github/workflows/system-tests.yml
```diff
@@ -64,71 +64,45 @@ env:
...
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.os }} Java ${{ matrix.version }}
+    env:
+      # default to Amazon Corretto JDK
+      JAVA_DISTRIBUTION: corretto
     steps:
+      # Zulu JDK appears to be more reliable for MacOS
+      - name: Use Java distribution Zulu
+        if: ${{ runner.os == 'macos' }}
+        run: echo "JAVA_DISTRIBUTION=zulu" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
ChrisSamo632: You'd have hoped so. Apparently there was a `set-env` action a while back, but that was deemed unsafe to use, so this appears to be the GitHub suggested way to achieve this between steps - see https://docs.github.com/en/actions/learn-github-actions/variables#passing-values-between-steps-and-jobs-in-a-workflow

Open to other/better approaches if you find one

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,248 @@
...
+
+      # many integration-tests use Docker to run Testcontainers; MacOS doesn't come with Docker, so we need to install it
+      - name: Set up Docker
+        if: ${{ runner.os == 'macos' }}
+        run: |-
+          /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+          brew install docker
+          colima start
+
+          if [ ! -e /var/run/docker.sock ]; then sudo ln -s ~/.colima/docker.sock /var/run/docker.sock; fi
```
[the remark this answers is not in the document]
ChrisSamo632: I tried installing Docker a few different ways and they had different setups, but I think this step is always necessary for Colima, although it then doesn't behave quite the same way as `docker` in Linux, so this step is effectively skipped for the minute as the `macos-latest` runner won't be used until NIFI-12191 is implemented

I'll remove the conditional, but we could consider removing the step completely, and the other macos conditionals, although I thought leaving them in might help with the future implemtnation of follow-on tickets *if* we see sufficient benefit (vs. cost) in running across multiple platforms as we do for other worfklows?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,248 @@
...
+      - name: Confirm Docker
+        run: id && docker info
```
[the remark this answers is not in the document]
ChrisSamo632: `id` was definitely a debugging thing, `docker info` was to help diagnose potiential issues encountered during builds, a bit like having the `-V` flag on `mvn` commands

I'll remove the `id` part for now, but can remove the entire step if you would prefer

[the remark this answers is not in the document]
exceptionfactory: Thanks, perhaps as long as we have the manual Docker installation, the `docker info` command could be part of that step?

.github/workflows/docker-tests.yml
ChrisSamo632: We need the binaries, but only the Zip file versions, these `rm` commands remove the unzipped directories from `target` (which exist prior to the `.zip` being created by the assembly plugin)

This *may not* be required for the space saving any longer, but I had a lot of issues with disk space during development, so deleting unnecessary files between steps seemed most sensible

.github/workflows/integration-tests.yml
ChrisSamo632: I agree that it would be better to make it easier to run the same set of tests locally, so I'll try adding a new profile, e.g. `integration-tests-ci`

exceptionfactory: Thanks @ChrisSamo632!

.github/workflows/integration-tests.yml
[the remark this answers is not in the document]
ChrisSamo632: `docker info` could still be useful for the `ubuntu-latest` runner where we're *not* explicitly installing it - if the tests start failing at some point without source code changes, e.g. Ubuntu release, Docker packages used by Ubunutu runner updated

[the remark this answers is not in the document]
exceptionfactory: That's a good point, in that way, it is along the lines of the `System Information` step in the `ci-workflow`, sounds good.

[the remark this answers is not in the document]
ChrisSamo632: Yes, exactly that - in fact, I'll rename this step and copy over the same `System Info` commands from the `ci` workflow, as they'll likely be worth having in these workflows too

.github/workflows/integration-tests.yml
ChrisSamo632: Actually, the [test](https://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html#test) configuration allows us to do this - I'd have preferred the use of `includes` and `excludes` I think, but with `test` we at least retain the method-level filtering

ChrisSamo632:
@exceptionfactory I've rebased from latest `main` as there have been quite a few changes over the past few days.

I also noticed I'd missed the `nifi-toolkit` tests from the new `docker-tests` workflow, which are failing on `main` due to the `file-manager` command no longer being present in the CLI - something I spotted and corrected also in [NIFI-12175](https://github.com/apache/nifi/pull/7864)

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,189 @@
...
+          ${{ env.MAVEN_BUILD_PROJECTS }}
+      # run nifi-toolkit docker build and tests before cleaning modules
+      - name: Run NiFi Toolkit Docker Tests
+        env:
+          MAVEN_OPTS: >-
+            ${{ env.DEFAULT_MAVEN_OPTS }}
+        run: >
+          ${{ env.MAVEN_COMMAND }}
+          ${{ env.MAVEN_DOCKER_ARGUMENTS }}
+          -pl nifi-toolkit/nifi-toolkit-assembly
```
ChrisSamo632, line 170: Blocked by NIFI-12175 (see #7864) - tests will fail due to Java version mis-match between assembly compiler and the Docker Image used to create `apache/nifi-toolkit` Image

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,189 @@
...
+    paths:
+      - '.github/workflows/docker-tests.yml'
+      - 'nifi-assembly/**'
+      - 'nifi-docker/**'
+      - 'nifi-registry/nifi-registry-assembly/**'
+      - 'nifi-registry/nifi-registry-docker-maven/**'
+      - 'nifi-toolkit/nifi-toolkit-assembly/**'
+      - 'nifi-toolkit/nifi-toolkit-cli/**'
+      - 'nifi-toolkit/nifi-toolkit-encrypt-config/**'
+      - 'nifi-toolkit/nifi-toolkit-tls/**'
```
ChrisSamo632, line 47: NIFI-12200 (see #7879) will lead to this needing to be changed

.github/workflows/ci-workflow.yml
```diff
@@ -75,7 +76,7 @@ env:
     -pl -nifi-toolkit/nifi-toolkit-assembly
 
 concurrency:
-  group: ${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
```
exceptionfactory, line 79: Can you provide some background for this particular change?

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,189 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  build_and_test:
+    strategy:
+      fail-fast: false
+      matrix:
+        # TODO: macos-latest doesn't come with Docker and Colima seems to have problems with the nifi-docker integration_test scripts - see NIFI-12177/NIFI-12191
```
exceptionfactory, line 116: Recommend avoiding the inclusion of the `TODO` comment as they quickly become ignored. The existence of the Jira issues seems sufficient.

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,189 @@
...
+        os: [ ubuntu-latest ]
+        version: [ 21 ]
+    timeout-minutes: 120
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.os }} Java ${{ matrix.version }}
+    env:
+      # default to Amazon Corretto JDK
+      JAVA_DISTRIBUTION: corretto
```
exceptionfactory, line 124: In this case, it seems much simpler to use `Zulu` for all builds. The JDK version is less important for this workflow because the focus is running Docker containers.

.github/workflows/docker-tests.yml
exceptionfactory: Can you revisit this based on the latest version and see if it is still necessary? It would be best to avoid including the removal if it is not needed, or call the appropriate build settings to skip Zip binary generation.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,143 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  build_and_test:
+    strategy:
+      fail-fast: false
+      matrix:
+        # TODO: macos-latest doesn't come with Docker and Colima seems to have problems with Test Containers - see NIFI-12191
```
exceptionfactory, line 89: See note above on removing the `TODO` given the existence of the Jira issue.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,143 @@
...
+        os: [ ubuntu-latest ]
+        version: [ 21 ]
+    timeout-minutes: 120
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.os }} Java ${{ matrix.version }}
+    env:
+      # default to Amazon Corretto JDK
+      JAVA_DISTRIBUTION: corretto
```
exceptionfactory, line 97: Although JDK differences are a bit more important for integration tests, it is still less important than system tests or workflow builds, so I also recommend using one JDK version and simplifying this workflow.

exceptionfactory (changes requested):
Thanks for making the updates @ChrisSamo632. I noted a few more things on this review, in particular, selecting just Azul Zulu as the JDK should simplify the Docker and Integration Test workflows.

.github/workflows/ci-workflow.yml
ChrisSamo632: Adding a `concurrency` entry for the new Workflows (and existing `system-tests`) seemed sensible to avoid running duplicate Workflows for PRs that trigger them (the `cron` schedule, of course, should not result in duplicate runs)

Having a `group` of just `${{ github.ref }}` means that the Workflows cancel each other because they're all running in the same `group`. Adding the `${{ github.workflow }}` name prefix means that a Workflow can only cancel an existing run of itself. There's not really a *need* to add the prefix to the `ci-workflow` Workflow definition, as it could be left as the only Workflow without the prefix, meaning it would still be in a `group` of its own. But as I was upting this Workflow definition for other reasons, then making this consistent with all other NiFi Workflows seemed sensible for the sake of future maintenance.

.github/workflows/docker-tests.yml
ChrisSamo632: Yeah that's fine now that I've raised the Jira tickets (I've also, I think, got a fix for these TODOs, but don't want to extend this existing PR further), will remove the TODOs

ChrisSamo632:
@exceptionfactory RE https://github.com/apache/nifi/pull/7858#discussion_r1362094730 (skipping the `rm -rf` Clean step for the `docker-tests`)

The last run of the Workflow in this PR that didn't include the `rm -rf` ran out of disk space, it would appear these removals are still required.

Building without the ZIP wouldn't work - the `dockermaven` builds use the ZIP files (I don't want to change that either as that's what the `dockerhub` builds do when downloading the build convenience artefacts from the Apache Archive servers), so the clean up step:
- cleans the Maven module `target` folders (except for the `assembly` modules as that's what we need for the Docker Image build step)
- deletes the un-zipped files from the assembly module `target` folders

The Docker Image builds each create a new Image (TAR file) containing the assembled ZIP, then unzips them before deleting the ZIP file - this uses a lot of disk space

exceptionfactory:
> @exceptionfactory RE [#7858 (comment)](https://github.com/apache/nifi/pull/7858#discussion_r1362094730) (skipping the `rm -rf` Clean step for the `docker-tests`)
> 
> The last run of the Workflow in this PR that didn't include the `rm -rf` ran out of disk space, it would appear these removals are still required.
> 
> Building without the ZIP wouldn't work - the `dockermaven` builds use the ZIP files (I don't want to change that either as that's what the `dockerhub` builds do when downloading the build convenience artefacts from the Apache Archive servers), so the clean up step:
> 
>     * cleans the Maven module `target` folders (except for the `assembly` modules as that's what we need for the Docker Image build step)
> 
>     * deletes the un-zipped files from the assembly module `target` folders
> 
> 
> The Docker Image builds each create a new Image (TAR file) containing the assembled ZIP, then unzips them before deleting the ZIP file - this uses a lot of disk space

Thanks for the additional background @ChrisSamo632, that makes sense now, it sounds like the current approach is the best option.

.github/workflows/docker-tests.yml
ChrisSamo632: Rebased from `main` after NIFI-12175 was merged - all tests passing 👍

.github/workflows/docker-tests.yml
ChrisSamo632: References to `toolkit-tls` removed after NIFI-12200 merged to `main`

ChrisSamo632:
@exceptionfactory have rebased from `main` latest after you merged NIFI-12175 - all new & existing test Workflows now passing without errors or warnings (from the Github Runners) 👍

exceptionfactory (approved):
Thanks for addressing the feedback and making adjustments @ChrisSamo632, this will be a helpful way to improve maintenance visibility for Docker builds and integration tests. +1 merging

PR #7859, opened 2023-10-09
Nifi-12193 Implement boolean switch for python processor support

taz1988:
# Summary

[NIFI-12193](https://issues.apache.org/jira/browse/NIFI-12193)

# Tracking

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-12193) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-properties/src/main/java/org/apache/nifi/util/NiFiProperties.java
+++ b/nifi-commons/nifi-properties/src/main/java/org/apache/nifi/util/NiFiProperties.java
@@ -314,6 +314,7 @@ public class NiFiProperties extends ApplicationProperties {
     public static final String DIAGNOSTICS_ON_SHUTDOWN_MAX_DIRECTORY_SIZE = "nifi.diagnostics.on.shutdown.max.directory.size";
 
     // python properties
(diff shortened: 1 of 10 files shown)


exceptionfactory (changes requested):
Thanks for proposing this change @taz1988, however, it seems unnecessary since the presence or absence of the `nifi.python.command` property value already controls whether Python support is enabled or disabled.

I also have some concerns related to the method naming of the unit test, but I will skip over those for now, pending discussion of the property itself.

Is there a particular problem with simply not setting the `nifi.python.command` property as the way to disable Python support?

markap14:
I would tend to agree, @exceptionfactory. Already, nifi.properties is extremely verbose with a huge number of configuration options. We should aim to make this as simple as we can.

exceptionfactory:
> I would tend to agree, @exceptionfactory. Already, nifi.properties is extremely verbose with a huge number of configuration options. We should aim to make this as simple as we can.

Thanks for the reply @markap14, I agree that keeping `nifi.properties` as focused as possible is an important part of avoiding greater complexity for administrators.

PR #7874, opened 2023-10-12
NIFI-12220: Added ability to create Controller Services from migrateProperties.

markap14:
Added ability to get raw property values from PropertyConfiguration instead of just effective values. Updated TestRunner to allow for testing these migration methods. Eliminated authentication properties from AWS processors and migrated all processors to using Controller Service or authentication. Eliminated Proxy properties in all AWS processors and instead just make use of the Proxy Configuration controller service

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


nifi-api/src/main/java/org/apache/nifi/migration/PropertyConfiguration.java
```diff
@@ -128,10 +128,83 @@ default Optional<String> getPropertyValue(PropertyDescriptor descriptor) {
...
+     * <p>
+     * This allows for properties that were previously defined in the extension to be moved to a Controller Service. For example,
+     * consider a Processor that has "Username" and "Password" properties. In the next version of the Processor, we want to support
+     * multiple types of authentication, and we delegate the authentication to a Controller Service. Consider that the Controller Service
+     * implementation we wish to use has a classname of {@code org.apache.nifi.services.authentication.UsernamePassword}. We might then
+     * use this method as such:
+     * </p>
+     *
+     * <pre><code>
+     *     // Create a new Controller Service ofo type org.apache.nifi.services.authentication.UsernamePassword whose Username and Password
```
exceptionfactory, line 180:
```suggestion
     *     // Create a new Controller Service of type org.apache.nifi.services.authentication.UsernamePassword whose Username and Password
```

nifi-api/src/main/java/org/apache/nifi/migration/PropertyConfiguration.java
```diff
@@ -128,10 +128,83 @@ default Optional<String> getPropertyValue(PropertyDescriptor descriptor) {
...
+     *
+     * <p>
+     * Also note that this method expects the classname of the implementation, not the classname of the interface.
+     * </p>
+     *
+     * @param implementationClassName the fully qualified classname of the Controller Service implementation
+     * @param serviceProperties       the property values to configure the newly created Controller Service with
+     * @return an identifier for the Controller Service
+     */
+    String createControllerService(String implementationClassName, Map<String, String> serviceProperties);
```
exceptionfactory, line 209: The implementation can throw an `IllegalArgumentException`, but that makes sense in the end as this is an operation that must succeed based on requested implementation.

This ends up resulting a stronger requirement between a particular component and a Controller Service implementation, but as shown in the AWS use cases, that seems reasonable.

nifi-mock/src/main/java/org/apache/nifi/util/MockPropertyConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+public class MockPropertyConfiguration implements PropertyConfiguration {
+    private final Map<String, String> propertyRenames = new HashMap<>();
+    private final Set<String> propertiesRemoved = new HashSet<>();
+    private final Set<String> propertiesUpdated = new HashSet<>();
+    private final Map<String, String> rawProperties;
+    private final Set<CreatedControllerService> createdControllerServices = new HashSet<>();
```
exceptionfactory, line 35: Recommend using `Linked` implementations to avoid unexpected ordering differences in some potential comparisons.
```suggestion
    private final Map<String, String> propertyRenames = new LinkedHashMap<>();
    private final Set<String> propertiesRemoved = new LinkedHashSet<>();
    private final Set<String> propertiesUpdated = new LinkedHashSet<>();
    private final Map<String, String> rawProperties;
    private final Set<CreatedControllerService> createdControllerServices = new LinkedHashSet<>();
```

nifi-mock/src/main/java/org/apache/nifi/util/MockPropertyConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+
+
+    public MockPropertyConfiguration(final Map<String, String> propertyValues) {
+        this.rawProperties = new HashMap<>(propertyValues);
```
exceptionfactory, line 39:
```suggestion
        this.rawProperties = new LinkedHashMap<>(propertyValues);
```

nifi-mock/src/main/java/org/apache/nifi/util/TestRunner.java
```diff
@@ -1064,4 +1064,6 @@ void assertAttributes(
      * @param eventType Provenance event type
      */
      void assertProvenanceEvent(ProvenanceEventType eventType);
+
```
exceptionfactory, line 1067: It would be helpful to add a basic Java Doc comment on this method to assist developers in using this method.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/ITPutKinesisStreamWithEndpointOverride.java
```diff
@@ -29,24 +31,24 @@
 
 // This integration test can be run against a mock Kenesis such as
 // https://github.com/mhart/kinesalite or https://github.com/localstack/localstack
+@Disabled("Required external service be running. Needs to be updated to make use of Localstack TestContainer")
```
exceptionfactory, line 34: I recommend deleting this test class instead of marking it disabled. There is not a lot included, and it would be better to start with a new integration test class.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/migration/ControllerServiceFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.apache.nifi.migration;
+
+import org.apache.nifi.controller.service.ControllerServiceNode;
+
+import java.util.Map;
+
+public interface ControllerServiceFactory {
+    ControllerServiceCreationDetails creationDetails(String implementationClassName, Map<String, String> propertyValues);
```
exceptionfactory, line 25: Minor naming question, is there a reason for naming this `creationDetails` as opposed to `getCreationDetails()`?

nifi-system-tests/nifi-system-test-suite/src/test/resources/conf/default/bootstrap.conf
```diff
@@ -27,7 +27,7 @@ java.arg.3=-Xmx512m
 
 java.arg.14=-Djava.awt.headless=true
 
-#java.arg.debug=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=8002
+java.arg.debug=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=8002
```
exceptionfactory, line 30: Reset to commented?

exceptionfactory (commented):
Thanks for introducing the new framework feature @markap14. As evidenced by the number of AWS components updated, the ability to create a Controller Service definition as part of migration will make it much easier to move component implementation in an optimal direction.

I had a few minor questions and comments, and I plan to do some runtime testing, but this looks solid in general.

nifi-api/src/main/java/org/apache/nifi/migration/PropertyConfiguration.java
markap14: Yeah, agreed. I don't love that it is needing to provide an explicit implementation of the service, but at the same time - the whole point is to allow backward compatibility and so it makes sense to declare that a specific implementation is what is necessary. Since it does not affect what is allowed to be configured, I think it's fair game.

nifi-mock/src/main/java/org/apache/nifi/util/MockPropertyConfiguration.java
markap14: I think I would disagree on this one. I believe that the order in which properties are removed, renamed, updated, etc. is not something that should be relied upon in a test. Tests should verify that "given this input, the expected output is produced." They should not attempt to assert the order in which the operations were performed in order to obtain that output - doing so would make the test very brittle.

nifi-mock/src/main/java/org/apache/nifi/util/TestRunner.java
markap14: Good catch. Will do.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/ITPutKinesisStreamWithEndpointOverride.java
markap14: Yeah, fair enough.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/migration/ControllerServiceFactory.java
markap14: Nope - `getCreationDetails` is probably a better name. Will change.

nifi-mock/src/main/java/org/apache/nifi/util/MockPropertyConfiguration.java
exceptionfactory: Thanks for the reply. For additional background, in the past we received some pull requests flagging potential issues surfaced through the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, which introduces randomness for certain collections to highlight the impact of element ordering. This exercise is admittedly somewhat contrived. I agree that tests and runtime code should not depend on the ordering of properties changed, which could surface through calls to method that return a Map of keys and values. With that background, I am fine with leaving the implementation as it is for now, and we could revisit later if necessary.

nifi-mock/src/main/java/org/apache/nifi/util/MockPropertyConfiguration.java
exceptionfactory: Thanks for the reply, I agree that it is fine for now as noted in the reply to the other similar comment.

exceptionfactory (approved):
Thanks for the work on this new migration feature @markap14! Support for migrating from individual properties to standard Controller Services will be very helpful in streamlining a number of components. Thanks for responding to the feedback, the code looks good, including the cleanup of the AWS components. Runtime testing with S3 components worked as expected.

I rebased to address a merge conflict with VersionedFlowSynchronizer, which was just a one-line change for an import, so planning to merge soon.

PR #7880, opened 2023-10-13
NIFI-12201 - Deprecation marking for nifi-toolkit-tls in NiFi 1.x

greyp9:
# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-toolkit/nifi-toolkit-tls/src/main/java/org/apache/nifi/toolkit/tls/TlsToolkitMain.java
+++ b/nifi-toolkit/nifi-toolkit-tls/src/main/java/org/apache/nifi/toolkit/tls/TlsToolkitMain.java
@@ -86,7 +86,7 @@ public class TlsToolkitMain {
             printUsageAndExit("Unknown service: " + service, ExitCode.INVALID_ARGS);
         }
 
-        System.out.println("The TLS Toolkit is deprecated and targeted for removal in Apache NiFi 2.0.");
+        System.err.println("The TLS Toolkit is deprecated and targeted for removal in Apache NiFi 2.0.");
 
         try {
             return mainClass.getDeclaredMethod("main", String[].class);
--- a/nifi-toolkit/nifi-toolkit-tls/src/main/java/org/apache/nifi/toolkit/tls/TlsToolkitMain.java
+++ b/nifi-toolkit/nifi-toolkit-tls/src/main/java/org/apache/nifi/toolkit/tls/TlsToolkitMain.java
@@ -86,6 +86,8 @@ public class TlsToolkitMain {
             printUsageAndExit("Unknown service: " + service, ExitCode.INVALID_ARGS);
         }
 
+        System.out.println("The TLS Toolkit is deprecated and targeted for removal in Apache NiFi 2.0.");
+
         try {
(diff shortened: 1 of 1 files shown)


nifi-toolkit/nifi-toolkit-tls/src/main/java/org/apache/nifi/toolkit/tls/TlsToolkitMain.java
```diff
@@ -86,6 +86,8 @@ protected Method getMain(String service) {
             printUsageAndExit("Unknown service: " + service, ExitCode.INVALID_ARGS);
         }
 
+        System.out.println("The TLS Toolkit is deprecated and targeted for removal in Apache NiFi 2.0.");
```
ChrisSamo632, line 89: Can/should we use the Deprecation Logger here?

And consider adding the `@Deprecated` annotation for `public` classes/methods referenced elsewhere so people get more warnings about this utility being removed in the future. The simple `System.out` could be easily missed (but I've not checked whether the standard NiFi Deprecation Logger would be available for us easily here)

ChrisSamo632 (changes requested):
Thanks for the update @greyp9, can we use the Deprecation Logger instead of `System.out` for noting the pending removal of this utility?

exceptionfactory (commented):
Thanks for introducing this message @greyp9, this approach looks good under the circumstances.

@ChrisSamo632 For other components, the DeprecationLogger and the Deprecated annotation make sense, but it does seem necessary in this particular scenario. The `TlsToolkitMain` is a command class that gets called through wrapping scripts, as opposed to a class that other components use. From that perspective, using DeprecationLogger does not add much, and it requires a configured logger implementation. For the same reasons, the Deprecated annotation does not provide a warning for end users of the TLS Toolkit, as it is developer-facing.

With that background, this approach seems sufficient to me. What do you think?

ChrisSamo632:
That's probably fair enough, thanks for the explanation @exceptionfactory 

I wonder whether this should be a `System.err` call instead though, to prevent problems for anyone capturing and parsing STDOUT currently for this utility. But I'm not going to push hard for such a change

greyp9:
> That's probably fair enough, thanks for the explanation @exceptionfactory
> 
> I wonder whether this should be a `System.err` call instead though, to prevent problems for anyone capturing and parsing STDOUT currently for this utility. But I'm not going to push hard for such a change

Thanks much for the input.  I'm switching it over to `System.err`.

ChrisSamo632:
Thanks for the update, LGTM 👍 

Will merge tomorrow (if I'm not beaten to it), when hopefully the build failures will have been fixed on the support branch (unrelated to this change)

ChrisSamo632 (approved):

exceptionfactory (approved):
Thanks for making the adjustments @greyp9, and thanks for the review @ChrisSamo632! The build failures are unrelated, and now resolved in a recent commit. +1 merging

ChrisSamo632:
This has been merged to `support/nifi-1.x`

PR #7883, opened 2023-10-13
NIFI-12030 Property pop-up window is horizontally resizable

annanys23:
# Summary

[NIFI-12030](https://issues.apache.org/jira/browse/NIFI-12030)

When a property of a controller service or processor is a list of values, the combo box is now horizontally resizable when open.

# Tracking

### Issue Tracking

- [x ] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/combo/jquery.combo.css
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/jquery/combo/jquery.combo.css
@@ -26,6 +26,7 @@
     height: 30px;
     line-height: 30px;
     border: 1px solid #eaeef0;
+    resize: horizontal;
 }
 
 .combo-options {
@@ -37,6 +38,7 @@
     z-index: 2000;
     border-top: none;
     font-size: 13px;
+    resize: horizontal;
 }
 
 .combo-options .fa {


Lehel44 (approved):
Thanks @annanys23 for the contribution. I tested the feature in Chrome and Safari and it works fine.
Screenshot 2023-10-26 at 17 33 12

exceptionfactory (changes requested):
Thanks for the contribution @annanys23, and thanks for testing this @Lehel44.

Although resizing the combo option box may have some usefulness, the user experience around resizing the combo option does not seem like the best approach. Others with additional user interface experience may have thoughts.

koccs:
Probably it is less comfortable because it requires hover over the options, but if any text is truncated with an ellipsis then a tooltip (with the full text) is the most commonly used solution in the industry.
Although it is not the best in the perspective of accessibility, but neither the resize handle.

scottyaslan:
This case is already addressed in the UI. For long values, the text is truncated with the ellipsis:

Screenshot 2023-10-31 at 1 53 40 PM

The user can hover over the value to see the title tooltip:

Screenshot 2023-10-31 at 1 53 11 PM

exceptionfactory (changes requested):
I agree with you @scottyaslan, this is already handled, so from that perspective, I don't think there are any better approaches.

Do you have any additional thoughts @annanys23? Otherwise it would be best to close the pull request, and perhaps continue the discussion on the associated Jira issue.

annanys23:
I don't have any other thoughts on this.  Thank you all for your input.

PR #7891, opened 2023-10-17
NIFI-12239 - Add OS/Java details at startup in bootstrap log

pvillard31:
# Summary

[NIFI-12239](https://issues.apache.org/jira/browse/NIFI-12239) - Add OS/Java details at startup in bootstrap log

Example of the added log:

````
2023-10-17 17:26:33,501 INFO [main] org.apache.nifi.bootstrap.RunNiFi {"javaVersion":"21","cores":16,"totalPhysicalMemoryMB":16384,"maxOpenFileDescriptors":10240,"xms":"1g","xmx":"1g"}
````

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
+++ b/nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
@@ -25,6 +25,10 @@ import org.apache.nifi.util.file.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
@@ -37,6 +41,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.Reader;
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
 import java.lang.reflect.Method;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
@@ -1165,6 +1171,9 @@ public class RunNiFi {
 
         nifiPropsFilename = nifiPropsFilename.trim();
 
+        String xmx = null;
+        String xms = null;
+
         final List<String> javaAdditionalArgs = new ArrayList<>();
         for (final Map.Entry<String, String> entry : props.entrySet()) {
             final String key = entry.getKey();
@@ -1172,6 +1181,12 @@ public class RunNiFi {
 
             if (key.startsWith("java.arg")) {
                 javaAdditionalArgs.add(value);
+                if(value.toLowerCase().startsWith("-xms")) {
+                    xms = StringUtils.substringAfter(value.toLowerCase(), "-xms");
+                }
+                if(value.toLowerCase().startsWith("-xmx")) {
+                    xmx = StringUtils.substringAfter(value.toLowerCase(), "-xmx");
+                }
             }
         }
 
@@ -1197,8 +1212,7 @@ public class RunNiFi {
             cpFiles.add(file.getAbsolutePath());
         }
 
-        String runtimeJavaVersion = System.getProperty("java.version");
-        defaultLogger.info("Runtime Java version: {}", runtimeJavaVersion);
+        defaultLogger.info(printBasicOSJavaDetails(xms, xmx));
 
         final StringBuilder classPathBuilder = new StringBuilder();
         for (int i = 0; i < cpFiles.size(); i++) {
@@ -1251,6 +1265,7 @@ public class RunNiFi {
         cmd.add("-classpath");
         cmd.add(classPath);
         cmd.addAll(javaAdditionalArgs);
+
         cmd.add("-Dnifi.properties.file.path=" + nifiPropsFilename);
         cmd.add("-Dnifi.bootstrap.listen.port=" + listenPort);
         cmd.add("-Dapp=NiFi");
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by exceptionfactory — COMMENTED** [review]
Thanks for working on this improvement @pvillard31.

Is there a particular reason for using JSON for serialization of this information? One important thing to note is that `jackson-databind` is not a direct dependency in `nifi-bootstrap`, so that should be declared if it is referenced in the class. Alternatively, what do you think about something else, such as a basic key = value approach?

The available processors information is also available from [Runtime.availableProcessors()](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#availableProcessors--). Although the MXBean approach is used for system diagnostics, it would be better to evaluate alternative options for the bootstrap command.

File: nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java [hunk 1]
```diff
@@ -1392,6 +1407,45 @@ public boolean accept(final File dir, final String filename) {
...
+        // num cores
+        final OperatingSystemMXBean os = ManagementFactory.getOperatingSystemMXBean();
+        details.put("cores", os.getAvailableProcessors());
+
+        // max file descriptor count + total physical memory
+        try {
+            final OperatingSystemMXBean osStats = ManagementFactory.getOperatingSystemMXBean();
+            final Class<?> unixOsMxBeanClass = Class.forName("com.sun.management.UnixOperatingSystemMXBean");
+            if (unixOsMxBeanClass.isAssignableFrom(osStats.getClass())) {
+                final Method totalPhysicalMemory = unixOsMxBeanClass.getMethod("getTotalPhysicalMemorySize");
```
exceptionfactory, line 1426: In this case, is the total physical memory useful? The `Runtime` memory methods provide heap information, which is actually addressable for the process, versus the physical memory, which may or may not be available to the JVM.

nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
```diff
@@ -1165,13 +1171,22 @@ public void start(final boolean monitor) throws IOException {
...
             final String value = entry.getValue();
 
             if (key.startsWith("java.arg")) {
                 javaAdditionalArgs.add(value);
+                if(value.toLowerCase().startsWith("-xms")) {
+                    xms = StringUtils.substringAfter(value.toLowerCase(), "-xms");
+                }
+                if(value.toLowerCase().startsWith("-xmx")) {
+                    xmx = StringUtils.substringAfter(value.toLowerCase(), "-xmx");
+                }
```
exceptionfactory, line 1189: Instead of parsing `Xmx`, what about using [Runtime.maxMemory()](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#maxMemory--)?

pvillard31: That would be the value for the bootstrap process (48MB), not for the NiFi JVM

nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
pvillard31: I believe that some users tend to set a very high heap, not leaving enough room for off-heap. Having the total memory available on the host may be useful. If the information is not available, we would see a null value.

nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
exceptionfactory: That's a good point, thanks!

pvillard31:
Yeah JSON was kind of a default choice to ease automatic parsing of this data but a KV approach is fine with me too, will update

pvillard31:
New format:

````
2023-10-17 18:24:24,923 INFO [main] org.apache.nifi.bootstrap.RunNiFi {totalPhysicalMemoryMB=16384, xmx=1g, cores=16, javaVersion=21, maxOpenFileDescriptors=10240, xms=1g}
````

Switched to ``Runtime`` for available processors. Open to suggestions for the two other values unless we believe we should just drop those values because not important enough to justify the use of ``MXBean``

**Review by exceptionfactory — COMMENTED** [review]
Thanks for the updates @pvillard31.

The [DiagnosticAnalysisTask](https://github.com/apache/nifi/blame/574c2b21682c78eb0b38bdeec6752b1688ce95a5/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/diagnostics/bootstrap/tasks/DiagnosticAnalysisTask.java#L146) has a method for retrieving the file descriptors using the object name as opposed to the private class. This approach should be more broadly compatible, and should also support retrieving the value of `TotalPhysicalMemorySize`.

**Comment by pvillard31** [comment]
Updated with your suggestion

File: nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java [hunk 3]
```diff
@@ -1392,6 +1404,33 @@ public boolean accept(final File dir, final String filename) {
         }
     }
 
+    private String printBasicOSJavaDetails(final String xms, final String xmx) {
+        Map<String, String> details = new HashMap<String, String>(6);
```
exceptionfactory, line 1408: Recommend `LinkedHashMap` for guaranteed ordering.
```suggestion
        Map<String, String> details = new LinkedHashMap<String, String>(6);
```

nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
```diff
@@ -1165,13 +1168,22 @@ public void start(final boolean monitor) throws IOException {
...
+        String xms = null;
+
         final List<String> javaAdditionalArgs = new ArrayList<>();
         for (final Map.Entry<String, String> entry : props.entrySet()) {
             final String key = entry.getKey();
             final String value = entry.getValue();
 
             if (key.startsWith("java.arg")) {
                 javaAdditionalArgs.add(value);
+                if(value.toLowerCase().startsWith("-xms")) {
```
exceptionfactory, line 1181:
```suggestion
                if (value.toLowerCase().startsWith("-xms")) {
```

nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
```diff
@@ -1165,13 +1168,22 @@ public void start(final boolean monitor) throws IOException {
...
+                    xms = StringUtils.substringAfter(value.toLowerCase(), "-xms");
+                }
+                if(value.toLowerCase().startsWith("-xmx")) {
```
exceptionfactory, line 1184:
```suggestion
                if (value.toLowerCase().startsWith("-xmx")) {
```

nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
exceptionfactory: Is there a reason for the `toLowerCase()` comparison? The case should be consistent.

nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
```diff
@@ -1392,6 +1404,33 @@ public boolean accept(final File dir, final String filename) {
...
+            details.put("maxOpenFileDescriptors", String.valueOf(maxOpenFileCount));
+            final Object totalPhysicalMemory = ManagementFactory.getPlatformMBeanServer().getAttribute(osObjectName, "TotalPhysicalMemorySize");
+            details.put("totalPhysicalMemoryMB", String.valueOf(((Long) totalPhysicalMemory) / (1024*1024)));
+        } catch (final Throwable t) {
+            // Ignore. This will throw either ClassNotFound or NoClassDefFoundError if unavailable in this JVM.
+        }
+
+        // min/max heap
+        details.put("xms", xms);
+        details.put("xmx", xmx);
```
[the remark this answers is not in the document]
exceptionfactory: It looks like these values could be `null`, correct? They should be checked before inserting into the Map.

exceptionfactory (changes requested):
Thanks for the updates @pvillard31. The simplified approach using `Map.toString()` looks good. Just noted a couple more recommendations and questions on naming, and then this should be ready to go.

nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/RunNiFi.java
pvillard31: One could be configuring NiFi using -Xms and -Xmx, no? The default being:
````
java.arg.2=-Xms1g
java.arg.3=-Xmx1g
````

pvillard31: OK I understand now... this is case sensitive so I can directly check against Xmx and Xms. Updating.

exceptionfactory: Thanks!

exceptionfactory (approved):

PR #7893, opened 2023-10-18
NIFI-12241 Efficient Parquet Splitting

takraj:
# Summary

[NIFI-12241](https://issues.apache.org/jira/browse/NIFI-12241)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


pvillard31:
I'll giving this a try later today but yeah, the naming can certainly be better as this new processor should always be followed by a record processor using the ParquetWriter so that the actual "splitting" is happening. Could be good to leverage the new @UseCase annotation to describe this.

takraj:
@markap14 , @pvillard31 Thank you for your comments. I have added use cases, and renamed the processor in 9d7fd48. Please check again.

pvillard31:
I did the below test:

- Large Parquet file with about 50M records -> ConvertRecord (Parquet Reader / Parquet Writer)
File size: 944MB
Duration: 7'01

- Same file -> CalculateParquetOffsets (1M records split) -> same ConvertRecord with 8 concurrent tasks
The CalculateParquetOffsets generates immediately 50 flow files (CLONE)
Duration: 4'40

I was expecting the second case to be much faster given that we use 8 concurrent threads. Any idea? Thoughts?

takraj:
@pvillard31

* I guess creating the 50 clones is actually copying the input 50 times. You could also give it a try with 'Zero Content Output" setup.
* To determine how many records are in the Parquet file,  CalculateParquetOffsets needs to read some parts of the file, for which it uses the same library as ParquetReader does. I'm not sure how this library determines the record count, but maybe it scans the whole file and counts them. But, if your input FlowFile has a 'record.count' attribute, then this step is skipped. Give it a try, I'd expect the whole process faster.
* ParquetReader uses RecordFilter to 'jump' to the wanted record. But as this is a general concept in the library we use, this means that the whole file is scanned through, and RecordFilter is evaluated to each of the records. Unfortunately, there is no concept like 'Jump to a record index'. Although I have an optimization idea of stop reading the records after we have read the desired number of records. RecordFilter then would be used only to find the first record to be read. I'll add a new commit soon with this improvement.

pvillard31:
The CalculateParquetOffsets step is taking less than 1 millisecond to complete: the 50 clones are not copying the data, they get the same reference to actual claim (where the data is stored in the content repo).

So I think it's really around the ParquetReader and how we can "jump" to the right location as you said. Will keep an eye out for your improvement for additional testing.

pvillard31:
New tests with 10M records (190MB):
- normal ConvertRecord (Parquet/Parquet) - 1 min 34 sec
- Calculate Parquet Offsets (1M records) - ConvertRecord 8 concurrent tasks - 10FFs - 38 seconds

So definitely better. I'm a +1 from a testing pov but would be nice for someone to have a look at the code changes. @bbende maybe?

takraj:
@pvillard31 I've been working on further performance improvements in the last couple of days, and created a new variant of this processor, which calculates the offsets of the row group boundaries, and updated the reader to take these offsets and seek onto these positions in the input file. This is achieved with a special configuration option of the Parquet reader, called 'File Range', that selects only row groups that overlap with the specified start and end offset range. Hopefully this brings major improvement.

Could you share how did you extract the processing times from NiFi in your benchmarks? Did you simply take the 'Tasks/Time' indicator, or is there something else that I can monitor?

takraj:
I have added a commit, which includes the alternative approach, that I have already mentioned in my previous comment. Please check.

pvillard31:
Thanks @takraj - giving another try to it.

pvillard31:
Test with 15,598,440 records (295MB):
- Normal ConvertRecord (Parquet/Parquet) - 2 min 13 sec
- Calculate Parquet Offsets (1M records) -> ConvertRecord 8 concurrent tasks - 16FFs - 38 seconds

It seems to be very similar to the previous test since the time is really related to how many records should be in every split (1M in this case). It's a 4x improvement when optimizing the number of splits versus the number of concurrent tasks so it's definitely a nice improvement.

takraj:
@pvillard31 Did you try the new `CalculateParquetRowGroupOffsets` processor too? In this one, you cannot configre the number of records in splits, because it splits by row group boundaries. But since it calculates where the `ParquetReader` should seek in the file, the data is distributed and processed much more efficiently. My measurements showed major improvement in performance.

You can also combine the two as `CalculateParquetRowGroupOffsets -> CalculateParquetOffsets`, so you would still have the control of how many records fall into a single FlowFile too.

I can't really do any further improvement to this, due to API limitations of the underlying library.

pvillard31:
Oh, I'm sorry, I didn't realize it was another processor. Will try again.

pvillard31:
FF with 24M records (452.85MB)
- ConvertRecord (Parquet/Parquet) - 3'31
- CalculateParquetOffsets (24 FFs) + ConvertRecord (8 CT) - 1'02
- CalculateParquetRowGroupOffsets (4 FFs) + ConvertRecord (8 CT) - 1'09
- CalculateParquetRowGroupOffsets (4 FFs) + CalculateParquetOffsets (25 FFs) + ConvertRecord (8 CT) - 50'

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/parquet/record/ParquetRecordReader.java
```diff
@@ -16,45 +16,79 @@
...
     private GenericRecord lastParquetRecord;
-    private RecordSchema recordSchema;
+    private final RecordSchema recordSchema;
 
     private final InputStream inputStream;
-    private final InputFile inputFile;
     private final ParquetReader<GenericRecord> parquetReader;
-
-    public ParquetRecordReader(final InputStream inputStream, final long inputLength, final Configuration configuration) throws IOException {
+    private final Long count;
```
tpalfy, line 47:
```suggestion
    private final Long recordsToRead;
```

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/processors/parquet/CalculateParquetOffsets.java
```diff
@@ -0,0 +1,249 @@
...
+
+    @Override
+    public Set<Relationship> getRelationships() {
+        return RELATIONSHIPS;
+    }
+
+    @Override
+    public void onTrigger(ProcessContext context, ProcessSession session) throws ProcessException {
+        final FlowFile original = session.get();
+        if (original == null) {
```
tpalfy, line 165:
```suggestion
        if (inputFlowFile == null) {
```
"original" makes sense when used as a relationship because that is a very specific context. In the code "original" can mean too many things. Also "input..." describes more accurately what it is. (In contrast, in case of a relationship calling it "input" would be very confusing.)

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/processors/parquet/CalculateParquetOffsets.java
```diff
@@ -0,0 +1,249 @@
...
+        ) {
+            return reader.getRecordCount();
+        } catch (IOException e) {
+            throw new ProcessException(e);
+        }
+    }
+
+    private List<FlowFile> getPartitions(
+            ProcessSession session,
+            FlowFile flowFile,
```
tpalfy, line 217:
```suggestion
            FlowFile inputFlowFile,
```

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/processors/parquet/CalculateParquetOffsets.java
```diff
@@ -0,0 +1,249 @@
...
+            long partitionSize,
+            long recordCount,
+            long recordOffset,
+            boolean zeroContentOutput
+    ) {
+        final long numberOfPartitions = (recordCount / partitionSize) + ((recordCount % partitionSize) > 0 ? 1 : 0);
```
tpalfy, line 223:
```suggestion
        final long numberOfPartitions = Math.ceilDiv(recordCount, partitionSize);
```

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/test/java/org/apache/nifi/parquet/ParquetTestUtils.java
```diff
@@ -0,0 +1,95 @@
...
+                }
+            });
+        }
+        return parquetFile;
+    }
+
+    private static Schema getSchema() throws IOException {
+        final File schemaFile = new File(SCHEMA_PATH);
+        final String schemaString = IOUtils.toString(new FileInputStream(schemaFile), StandardCharsets.UTF_8);
+        return new Schema.Parser().parse(schemaString);
```
tpalfy, line 80:
```suggestion
        try (InputStream schemaInputStream = ParquetTestUtils.class.getClassLoader().getResourceAsStream("avro/user.avsc")) {
            final String schemaString = IOUtils.toString(schemaInputStream, StandardCharsets.UTF_8);
            return new Schema.Parser().parse(schemaString);
        }
```

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/processors/parquet/CalculateParquetOffsets.java
takraj: Done in 9f340c9

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/processors/parquet/CalculateParquetOffsets.java
takraj: Done in 9f340c9

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/processors/parquet/CalculateParquetOffsets.java
takraj: Done in 9f340c9

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/parquet/hadoop/AvroParquetHDFSRecordReader.java
```diff
@@ -35,11 +35,18 @@ public class AvroParquetHDFSRecordReader implements HDFSRecordReader {
...
+    private final Long count;
+    private long recordsRead = 0;
 
     private final ParquetReader<GenericRecord> parquetReader;
 
     public AvroParquetHDFSRecordReader(final ParquetReader<GenericRecord> parquetReader) {
+        this(parquetReader, null);
+    }
+
+    public AvroParquetHDFSRecordReader(final ParquetReader<GenericRecord> parquetReader, Long count) {
```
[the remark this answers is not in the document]
takraj: Done in 9f340c9

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/processors/parquet/FetchParquet.java
```diff
@@ -63,9 +79,35 @@ public class FetchParquet extends AbstractFetchHDFSRecord {
...
+
+        if (offset != null) {
+            readerBuilder.withFilter(FilterCompat.get(OffsetRecordFilter.offset(offset)));
+        }
+
+        if (count == null) {
+            return new AvroParquetHDFSRecordReader(readerBuilder.build());
+        } else {
+            return new AvroParquetHDFSRecordReader(readerBuilder.build(), count);
+        }
```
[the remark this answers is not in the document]
takraj: Done in 9f340c9

nifi-nar-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/parquet/hadoop/AvroParquetHDFSRecordReader.java
```diff
@@ -35,11 +35,18 @@ public class AvroParquetHDFSRecordReader implements HDFSRecordReader {
     private GenericRecord lastRecord;
     private RecordSchema recordSchema;
     private boolean initialized = false;
+    private final Long count;
```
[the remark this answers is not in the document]
takraj: Done in 9f340c9

tpalfy:
LGTM
Thank you @takraj for your work.
Merged to main and merging to nifi-1.x

tpalfy:
Pushed to nifi-1.x

PR #7894, opened 2023-10-18
NIFI-12240, NIFI-12195: Created new python processors for text embedd…

markap14:
…ings, inserting into Chroma, querying Chroma, querying ChatGPT, inserting into and querying Pinecone. Fixed some bugs in the Python framework. Added Python extensions to assembly. Also added ability to load dependencies from a requirements.txt as that was important for making the different vectorstore implementations play more nicely together.

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-python-extensions/nifi-text-embeddings-module/src/main/python/ChunkDocuments.py
+++ b/nifi-python-extensions/nifi-text-embeddings-module/src/main/python/ChunkDocuments.py
@@ -1,0 +1,211 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+
+from langchain.text_splitter import Language
+from nifiapi.flowfiletransform import FlowFileTransform, FlowFileTransformResult
+from nifiapi.properties import PropertyDescriptor, StandardValidators, PropertyDependency, ExpressionLanguageScope
+
+SPLIT_BY_CHARACTER = 'Split by Character'
+SPLIT_CODE = 'Split Code'
+RECURSIVELY_SPLIT_BY_CHARACTER = 'Recursively Split by Character'
+
+TEXT_KEY = "text"
+METADATA_KEY = "metadata"
+
+
+class ChunkDocuments(FlowFileTransform):
+    class Java:
+        implements = ['org.apache.nifi.python.processor.FlowFileTransform']
+    class ProcessorDetails:
+        version = '2.0.0-SNAPSHOT'
+        description = """Splits incoming documents into chunks that are appropriately sized for creating Text Embeddings. The input is expected to be in "json-lines" format, with
+            each line having a 'text' and a 'metadata' element. Each line will then be split into one or more lines in the output."""
+        tags = ["text", "split", "chunk", "langchain", "embeddings", "vector", "machine learning", "ML", "artificial intelligence", "ai", "document"]
+        dependencies = ['langchain']
+
+
+    CHUNK_STRATEGY = PropertyDescriptor(
+        name="Chunking Strategy",
+        description="Specifies which splitter should be used to split the text",
+        allowable_values=[RECURSIVELY_SPLIT_BY_CHARACTER, SPLIT_BY_CHARACTER, SPLIT_CODE],
+        required=True,
+        default_value=RECURSIVELY_SPLIT_BY_CHARACTER
+    )
+    SEPARATOR = PropertyDescriptor(
+        name="Separator",
+        description="Specifies the character sequence to use for splitting apart the text. If using a Chunking Strategy of Recursively Split by Character, " +
+                    "it is a comma-separated list of character sequences. Meta-characters \\n, \\r and \\t are automatically un-escaped.",
+        required=True,
+        default_value="\\n\\n,\\n, ,",
+        validators=[StandardValidators.NON_EMPTY_VALIDATOR],
+        dependencies=[PropertyDependency(CHUNK_STRATEGY, SPLIT_BY_CHARACTER, RECURSIVELY_SPLIT_BY_CHARACTER)],
+        expression_language_scope=ExpressionLanguageScope.FLOWFILE_ATTRIBUTES
+    )
+    SEPARATOR_FORMAT = PropertyDescriptor(
+        name="Separator Format",
+        description="Specifies how to interpret the value of the <Separator> property",
+        required=True,
+        default_value="Plain Text",
+        allowable_values=["Plain Text", "Regular Expression"],
+        dependencies=[PropertyDependency(CHUNK_STRATEGY, SPLIT_BY_CHARACTER, RECURSIVELY_SPLIT_BY_CHARACTER)]
+    )
+    CHUNK_SIZE = PropertyDescriptor(
+        name="Chunk Size",
+        description="The maximum size of a chunk that should be returned",
+        required=True,
+        default_value="4000",
+        validators=[StandardValidators.POSITIVE_INTEGER_VALIDATOR]
+    )
+    CHUNK_OVERLAP = PropertyDescriptor(
+        name="Chunk Overlap",
+        description="The number of characters that should be overlapped between each chunk of text",
+        required=True,
+        default_value="200",
+        validators=[StandardValidators.NON_NEGATIVE_INTEGER_VALIDATOR]
+    )
+    KEEP_SEPARATOR = PropertyDescriptor(
+        name="Keep Separator",
+        description="Whether or not to keep the text separator in each chunk of data",
+        required=True,
+        default_value="false",
+        allowable_values=["true", "false"],
+        dependencies=[PropertyDependency(CHUNK_STRATEGY, SPLIT_BY_CHARACTER, RECURSIVELY_SPLIT_BY_CHARACTER)]
+    )
+    STRIP_WHITESPACE = PropertyDescriptor(
+        name="Strip Whitespace",
+        description="Whether or not to strip the whitespace at the beginning and end of each chunk",
+        required=True,
+        default_value="true",
+        allowable_values=["true", "false"],
+        dependencies=[PropertyDependency(CHUNK_STRATEGY, SPLIT_BY_CHARACTER, RECURSIVELY_SPLIT_BY_CHARACTER)]
+    )
+    LANGUAGE = PropertyDescriptor(
+        name="Language",
+        description="The language to use for the Code's syntax",
+        required=True,
+        default_value="python",
+        allowable_values=[e.value for e in Language],
+        dependencies=[PropertyDependency(CHUNK_STRATEGY, SPLIT_CODE)]
+    )
+
+    property_descriptors = [CHUNK_STRATEGY,
+                            SEPARATOR,
+                            SEPARATOR_FORMAT,
+                            CHUNK_SIZE,
+                            CHUNK_OVERLAP,
+                            KEEP_SEPARATOR,
+                            STRIP_WHITESPACE]
+
+
+    def __init__(self, **kwargs):
+        pass
+
+    def getPropertyDescriptors(self):
+        return self.property_descriptors
+
+
+    def split_docs(self, context, flowfile, documents):
+        from langchain.text_splitter import CharacterTextSplitter
+        from langchain.text_splitter import RecursiveCharacterTextSplitter
+
+        strategy = context.getProperty(self.CHUNK_STRATEGY).getValue()
+        if strategy == SPLIT_BY_CHARACTER:
+            text_splitter = CharacterTextSplitter(
+                separator = context.getProperty(self.SEPARATOR).evaluateAttributeExpressions(flowfile).getValue(),
+                keep_separator = context.getProperty(self.KEEP_SEPARATOR).asBoolean(),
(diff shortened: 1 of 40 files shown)


nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryPinecone.py
```diff
@@ -0,0 +1,165 @@
...
+import pinecone
+
+
+class QueryPinecone(FlowFileTransform):
+    class Java:
+        implements = ['org.apache.nifi.python.processor.FlowFileTransform']
+
+    class ProcessorDetails:
+        version = '2.0.0-SNAPSHOT'
+        description = "Queries Pinecone in order to gather a specified number of documents that are most closely related to the given query."
```
krisztina-zsihovszki, line 30: In my view it'd be useful to mention that QueryPinecone, QueryChroma and PromptChatGPT require an incoming flowfile, otherwise the queries are not triggered (no output flow file is created, even if there was a match based on the processor properties).

krisztina-zsihovszki (commented):
Not being a Python expert, added a few comments regarding the processors usage only.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryPinecone.py
markap14: This should actually have been done at the higher level, the Python Processor Proxy level - it should require input. Good catch. Will address that.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/ParseUnstructuredDocument.py
```diff
@@ -0,0 +1,259 @@
...
+PARSING_STRATEGY_FAST = "Fast"
+
+SINGLE_DOCUMENT = "Single Document"
+DOCUMENT_PER_ELEMENT = "Document Per Element"
+
+TEXT_KEY = "text"
+METADATA_KEY = "metadata"
+
+
+class ParseUnstructuredDocument(FlowFileTransform):
```
joewitt, line 43: would ParseDocument be a better name?  It doesn't seem critical to think of it as structured or unstructured and makes the name quite long.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/ChromaUtils.py
```diff
@@ -0,0 +1,155 @@
...
+    name="Password",
+    description="The password to use for authenticating to Chroma server",
+    validators=[StandardValidators.NON_EMPTY_VALIDATOR],
+    required=True,
+    sensitive=True,
+    dependencies=[PropertyDependency(AUTH_STRATEGY, BASIC_AUTH)]
+)
+COLLECTION_NAME = PropertyDescriptor(
+    name="Collection Name",
+    description="The name of the Chroma Collection to update",
```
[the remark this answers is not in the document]
markap14: Will mention that in the description of the PutChroma Processor. It does not belong here, as this Property Descriptor is used by both PutChroma and QueryChroma. Which means it's a typo to say "the Collection to update" so will address that as well.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/ParseUnstructuredDocument.py
markap14: Thanks @joewitt. I went back and forth on whether to have the word 'unstructured' there or not. The only concern that I had with 'ParseDocument' is that it might be vague. But if you think ParseDocument fits better, will change to that.

joewitt: yeah with a user hat on - i'm parsing a doc.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/ChunkDocuments.py
```diff
@@ -0,0 +1,211 @@
...
+
+SPLIT_BY_CHARACTER = 'Split by Character'
+SPLIT_CODE = 'Split Code'
+RECURSIVELY_SPLIT_BY_CHARACTER = 'Recursively Split by Character'
+
+TEXT_KEY = "text"
+METADATA_KEY = "metadata"
+
+
+class ChunkDocuments(FlowFileTransform):
```
exceptionfactory, line 30: Following the convention of other components, it seems like this should be named `ChunkDocument` as opposed to `ChunkDocuments`.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/PutChroma.py
```diff
@@ -0,0 +1,125 @@
...
+import EmbeddingUtils
+
+
+class PutChroma(FlowFileTransform):
+    class Java:
+        implements = ['org.apache.nifi.python.processor.FlowFileTransform']
+
+    class ProcessorDetails:
+        version = '2.0.0-SNAPSHOT'
+        description = """Publishes JSON data to a Chroma VectorDB. The Incoming data must be in single JSON per Line format, each with two keys: 'text' and 'metadata'.
```
exceptionfactory, line 30: The requirement for this specific structure with ` text` and `metadata`, with a single JSON object per line seems limiting. What do you think about introducing a Record Reader? That would provide at least one layer of flexibility. If those are the only object keys required, perhaps they don't need to be configurable, but that could be another option to consider.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/PutChroma.py
```diff
@@ -0,0 +1,125 @@
...
+        collection_name = context.getProperty(ChromaUtils.COLLECTION_NAME).evaluateAttributeExpressions(flowfile).getValue()
+        distance_method = context.getProperty(self.DISTANCE_METHOD).getValue()
+        id_field_name = context.getProperty(self.DOC_ID_FIELD_NAME).evaluateAttributeExpressions(flowfile).getValue()
+
+        collection = client.get_or_create_collection(
+            name=collection_name,
+            embedding_function=embedding_function,
+            metadata={"hnsw:space": distance_method})
+
+        json_lines = flowfile.getContentsAsBytes().decode()
```
exceptionfactory, line 91: Are there any concerns here related to memory consumption with large files? It seems like this should be streamed.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryChroma.py
```diff
@@ -0,0 +1,159 @@
...
+        name="Query",
+        description="The query to issue to the Chroma VectorDB. The query is always converted into embeddings using the configured embedding function, and the embedding is " +
+                    "then sent to Chroma. The text itself is not sent to Chroma.",
+        required=True,
+        validators=[StandardValidators.NON_EMPTY_VALIDATOR],
+        expression_language_scope=ExpressionLanguageScope.FLOWFILE_ATTRIBUTES
+    )
+    NUMBER_OF_RESULTS = PropertyDescriptor(
+        name="Number of Results",
+        description="The number of results to return from Chroma",
```
exceptionfactory, line 46: Is this a maximum or limit? That would be worth clarifying in the description.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryChroma.py
```diff
@@ -0,0 +1,159 @@
...
+        required=True,
+        validators=[StandardValidators.POSITIVE_INTEGER_VALIDATOR],
+        default_value="10",
+        expression_language_scope=ExpressionLanguageScope.FLOWFILE_ATTRIBUTES
+    )
+    WHERE_CLAUSE = PropertyDescriptor(
+        name="Where Clause (Metadata Filter)",
```
exceptionfactory, line 53: Recommend avoiding the use of parentheses in the property name. What do you think about naming this simply `Metadata Filter Clause` or `Metadata Filter`? With the value needing to be JSON, `Metadata Filter` seems straightforward.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryChroma.py
```diff
@@ -0,0 +1,159 @@
...
+        description="A JSON representation of a Metadata Filter that can be applied against the Chroma documents in order to narrow down the documents that can be returned. " +
+                    "For example: { \"metadata_field\": \"some_value\" }",
+        validators=[StandardValidators.NON_EMPTY_VALIDATOR],
+        expression_language_scope=ExpressionLanguageScope.FLOWFILE_ATTRIBUTES,
+        required=False
+    )
+    WHERE_DOCUMENT_CLAUSE = PropertyDescriptor(
+        name="Where Document (Document Filter)",
```
exceptionfactory, line 61: As above, what about `Document Filter` for this property name?

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryChroma.py
```diff
@@ -0,0 +1,159 @@
...
+        ids = query_results['ids'][0]
+        distances = None if (not self.include_distances or query_results['distances'] is None) else query_results['distances'][0]
+        metadatas = None if (not self.include_metadatas or query_results['metadatas'] is None) else query_results['metadatas'][0]
+        documents = None if (not self.include_documents or query_results['documents'] is None) else query_results['documents'][0]
+        embeddings = None if (not self.include_embeddings or query_results['embeddings'] is None) else query_results['embeddings'][0]
+
+        (output_contents, mime_type) = self.query_utils.create_json(flowfile, documents, metadatas, embeddings, distances, ids)
+
+        # Return the results
+        attributes = {"mime.type": mime_type}
```
exceptionfactory, line 158: Will the response `mime.type` always be `application/json`? It would be helpful to indicate the WritesAttribute status for documentation purposes.

nifi-system-tests/nifi-system-test-suite/src/test/resources/conf/default/bootstrap.conf
```diff
@@ -27,7 +27,7 @@ java.arg.3=-Xmx512m
 
 java.arg.14=-Djava.awt.headless=true
 
-#java.arg.debug=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=8002
+java.arg.debug=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=8002
```
exceptionfactory, line 30: It looks like the comment needs to be restored.
```suggestion
#java.arg.debug=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=8002
```

exceptionfactory (commented):
Thanks for introducing these new Processors @markap14! This covers a lot of ground in terms of generative AI capabilities, and also providing examples of Python-based components.

I highlighted a few minor questions and recommendations related to naming.

At a general level, two areas of concern are reading input JSON content into memory for the `PutChroma` and `PutPinecone` Processors, and the input format requirements for JSON fields and lines in those components. Introducing more configuration options also brings more complexity, but particularly in relation to the JSON object-per-line requirement, it seems more limiting.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryChroma.py
markap14: It will always return this number of results. (Unless there are fewer than this number in the db).

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryChroma.py
markap14: Fair.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/PutChroma.py
markap14: In all Gen AI examples that I've come across, the structure is JSON and has a 'text' or similar property along with a 'metadata' property. So I tried to make this as intuitive as possible. This is also the format that is produced by the ParseDocument processor. Generally, I expect the flow to look like:
`(some source) -> ParseDocument -> ChunkDocument -> PutChroma`
And so this format keeps things simple. Unfortunately, we do not yet have the @UseCase and @MultiProcessorUseCase capabilities built out yet for Python based processors, so this hasn't been that clearly documented yet. But this will definitely be something we'll want to highlight.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/PutChroma.py
markap14: I'd say yes and no. In this space, we tend to see code that assumes that one document can fit well within the memory limits of the application. While in Java based processors, this would never fly, it feels a bit safer here, because of the fact that there is no shared heap. Unfortunately, if we were to use streaming, with the current architecture, it would require an RPC call per line of text, which would result in unacceptable performance. I think this is something that we can likely improve over time. For example, we might actually end up passing a reference to the content repository/offset/length and then make a `BytesIO` available to the Processor. But I didn't see a cleaner way to handle this in the meantime.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/vectorstores/QueryChroma.py
markap14: It will not always be `application/json`. The mime type will depend on the chosen Output Strategy.

nifi-python-extensions/nifi-text-embeddings-module/src/main/python/ChunkDocuments.py
markap14: Agreed, good catch!

exceptionfactory (approved):
Thanks the responses @markap14. These new components are certainly resource intensive based on testing, as expected for unstructured document processing and vector manipulation. I am sure we will continue to evolve these capabilities, but these components together definitely highlight the capabilities of Python integration, and provide the building blocks for GenAI flows! +1 merging

PR #7903, opened 2023-10-19
NIFI-11909: Clearing password when login fails

mcgilman:
NIFI-11909:
- Clearing the password in the login form when login fails or unable to verify access once authentication succeeds.


exceptionfactory (approved):
Thanks for the improvement @mcgilman, verified that clearing the password works as expected. +1 merging

exceptionfactory:
Merged in https://github.com/apache/nifi/commit/97dfe2d812a9fa8b0fcc2b1d0b7f352f656b530e

PR #7905, opened 2023-10-19
NIFI-12247: Reverting to the initial value if there is no selected parameter in the combo editor

mcgilman:
NIFI-12247:
- Reverting to the initial value if there is no selected parameter in the combo editor.


markap14:
Thanks @mcgilman was able to replicate issue and verify fix. +1 merged to main

PR #7920, opened 2023-10-23
NIFI-12191 simplify integration-tests and docker-tests GitHub Action workflows

ChrisSamo632:
# Summary

[NIFI-12191](https://issues.apache.org/jira/browse/NIFI-12191) simplify integration-tests and docker-tests GitHub Action workflows

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/.github/workflows/docker-tests.yml
+++ b/.github/workflows/docker-tests.yml
@@ -107,36 +107,23 @@ permissions:
 
 jobs:
   build_and_test:
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ ubuntu-latest ]
-        version: [ 21 ]
     timeout-minutes: 120
-    runs-on: ${{ matrix.os }}
-    name: ${{ matrix.os }} Java ${{ matrix.version }}
+    runs-on: ubuntu-latest
+    name: Ubuntu Java 21
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
-      - name: Set up Java Zulu ${{ matrix.version }}
+      - name: Set up Java Zulu 21
         uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: ${{ matrix.version }}
+          java-version: 21
           cache: 'maven'
-
-      - name: Set up Docker
-        if: ${{ runner.os == 'macOS' }}
-        run: |-
-          /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-          brew install docker
-          colima start
-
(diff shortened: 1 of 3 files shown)


exceptionfactory (commented):
Thanks for working on this @ChrisSamo632.

I am wondering about the value of testing multiple platforms. For the Docker tests, there seems to be little value on running the builds on multiple platforms, since the tests involve the container, not necessarily the host OS.

For the integration tests, there may be some additional value, but there is also a cost in terms of time to execute the tests.

What do you think of limiting the Docker tests to just Linux, while keeping the integration tests for both Linux and macOS?

ChrisSamo632:
@exceptionfactory I agree that there's concern over the increased cost vs. possible utility - to a degree, I was more wondering whether things would work at all with `colima` instead of `docker`

I suspect (the Checks will tell in time) that using MacOS for the ITs will prove more tempremental that worthwhile, so for that reason alone I'm leaning on the side of not including those unless there are obvious benefits - the only real one I can come up with is that MacOS as a development environment is probably more common that Linux (but that's a assumption on my part). How useful regular IT runs are in that environment though, I'm as yet unconvinced

For similar reasons, I was thinking the `docker-tests` might be worthwhile, and they're also a lot faster to run so would be cheaper in terms of build resources. But that said, I suspect that building and running the Docker Images with `colima` (or any tool other than `docker`) is probably relatively rare / not something we're looking to directly support here

So I think it probably comes down to whether we think there is sufficient benefit in (more regularly) running the extended tests in the (assumed) more common development environment

exceptionfactory:
Thanks for the thoughtful reply @ChrisSamo632. Those are definitely points worth considering as far as Colima versus standard Docker.

I have been in touch with Apache Infrastructure support about testing their new hosted GitHub Runners, which will hopefully provide some performance improvements. However, I'm not sure about the operating systems supported, so that may be limiting.

With that background, and considering the resources used, I think it would be better to keep the Docker and Integration Test workflows limited to one OS (Linux). Although it might not cover every possible situation, it is much better than no coverage.

What do you think about re-scoping this pull request to removing the matrix configuration and just targeting the one OS?

ChrisSamo632:
@exceptionfactory with the discussion and the continued failure of the Docker-based tests on MacOS runners, I'm certainly happy to re-scope this to be a simplification of the existing Workflows to focus on Ubuntu-based testing only

exceptionfactory (approved):
Thanks for work on this and re-scoping the changes @ChrisSamo632, the latest version looks good! +1 merging

PR #7922, opened 2023-10-23
NIFI-12121 added the ability to modify the output filter in …

mosermw:
… JSLTTransformJSON, which can be used to keep null values or empty arrays

# Summary

The JSLT java library supports a filter on the results of a JSLT transform.  By default, that filter removes
all objects with a null value, all empty objects, and all empty arrays.  This PR supports a new Property
on the JSLTTransformJSON processor that lets users configure that filter, in case the default behavior
is not desired.

Also, this PR fixes a potential NPE in OnShutdown if the cache has never been created.

[NIFI-12121](https://issues.apache.org/jira/browse/NIFI-12121)

This PR should cherry-pick cleanly to the nifi-1.x support branch, though I don't know if it's wanted there.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
+++ b/nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
@@ -115,7 +115,7 @@ public class JSLTTransformJSON extends AbstractProcessor {
             .displayName("Transform Result Filter")
             .description("A filter of output results using another JSLT. This property allows you to change the built-in filter,"
                     + " which removes objects with null values, empty objects and empty arrays from the output."
-                    + " Use a filter such as \"1 == 1\" to disable all filtering.")
+                    + " Use a filter such as \"true\" to disable all filtering.")
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .identifiesExternalResource(ResourceCardinality.SINGLE, ResourceType.TEXT, ResourceType.FILE)
             .required(true)
--- a/nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
+++ b/nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
@@ -114,8 +114,8 @@ public class JSLTTransformJSON extends AbstractProcessor {
             .name("jslt-transform-result-filter")
             .displayName("Transform Result Filter")
             .description("A filter of output results using another JSLT. This property allows you to change the built-in filter,"
-                    + " which removes objects with null values, empty objects and empty arrays from the output."
-                    + " This JSLT must return true/false"
+                    + " which removes JSON objects with null values, empty objects and empty arrays from the output."
+                    + " This JSLT must return true for each JSON object to be included and false for each object to be removed."
                     + " Use a filter such as \"true\" to disable all filtering.")
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .required(true)
--- a/nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
+++ b/nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
@@ -88,6 +88,8 @@ import static org.apache.nifi.processors.jslt.JSLTTransformJSON.TransformationSt
         + "fails, the original FlowFile is routed to the 'failure' relationship.")
 public class JSLTTransformJSON extends AbstractProcessor {
 
+    public static String JSLT_FILTER_DEFAULT = ". != null and . != {} and . != []";
+
     public static final PropertyDescriptor JSLT_TRANSFORM = new PropertyDescriptor.Builder()
             .name("jslt-transform-transformation")
             .displayName("JSLT Transformation")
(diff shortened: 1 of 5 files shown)


nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
```diff
@@ -107,6 +108,16 @@ public class JSLTTransformJSON extends AbstractProcessor {
             .defaultValue(EACH_OBJECT.getValue())
             .build();
 
+    public static final PropertyDescriptor RESULT_FILTER = new PropertyDescriptor.Builder()
+            .name("jslt-transform-result-filter")
+            .displayName("Transform Result Filter")
+            .description("An optional filter of output results using another JSLT, allowing you to change the default filter."
+                    + " The default filter is \". != null and . != {} and . != []\" which excludes objects with null values, empty objects and empty arrays")
```
exceptionfactory, line 115: Instead of making this an optional property, what do you think about setting it as required and providing the default value? This would make the default behavior more obvious, and provide a quick example of what it can do.

exceptionfactory (commented):
Thanks for the improvement @mosermw. As noted on a detail comment, what do you think about changing the approach to make the new property required with the default value provided?

nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
mosermw: I like your suggested approach, and I will make the change.  I certainly didn't know that the JSLT library had a default output filter, and it would be nice to make this obvious.  I'm a bit concerned about performance of using a JSLT filter instead of the default filter, so I'll investigate whether that makes a significant difference.

mosermw: @exceptionfactory I made the new property required, gave it a default value, and updated the description.  My performance tests showed a small penalty (roughly 5%) using a simple JSLT filter over the built-in default.  To mitigate this, if the property is set to the default filter value, then I use the library's built-in default filter.  This should preserve current performance.

exceptionfactory: Thanks for the updates @mosermw. Can you clarify the difference between the "default filter" and the default configured value of the new property? I may have misunderstood the initial property description. Is the difference based on the way the internal "default filter" is implemented, versus the same functional result using the property default JSLT expression?

mosermw: No problem @exceptionfactory it took me a bit of reading and testing to learn this.  The default filter in the JSLT library is  [DefaultJsonFilter.java](https://github.com/schibsted/jslt/blob/master/core/src/main/java/com/schibsted/spt/data/jslt/filters/DefaultJsonFilter.java).  It uses the [NodeUtils::isValue](https://github.com/schibsted/jslt/blob/afb0cc037daac1ac69d9348d42a62d09a12cc4f5/core/src/main/java/com/schibsted/spt/data/jslt/impl/NodeUtils.java#L59) method to determine whether to output a JsonNode.  This method is a bit faster than processing a JSLT that does the same thing.

The [documentation about this filter](https://github.com/schibsted/jslt/blob/master/docs/api.md#object-key-filter) shows an equivalent JSLT which is what you see in the JSLT_FILTER_DEFAULT variable.  Even though implementing a Java based filter would be faster, I think using JSLT for the filter gives users maximum flexibility.

exceptionfactory: Thanks for the clarification and references @mosermw, that makes sense now. I agree that using JSLT for the filter definition itself makes the most sense from a configuration point of view.

nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
```diff
@@ -107,6 +110,18 @@ public class JSLTTransformJSON extends AbstractProcessor {
...
             .build();
 
+    public static final PropertyDescriptor RESULT_FILTER = new PropertyDescriptor.Builder()
+            .name("jslt-transform-result-filter")
+            .displayName("Transform Result Filter")
+            .description("A filter of output results using another JSLT. This property allows you to change the built-in filter,"
+                    + " which removes objects with null values, empty objects and empty arrays from the output."
+                    + " Use a filter such as \"true\" to disable all filtering.")
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .identifiesExternalResource(ResourceCardinality.SINGLE, ResourceType.TEXT, ResourceType.FILE)
```
exceptionfactory, line 120: Is the intent to support an external reference to a filter? That would require additional changes to read an external resource, and it seems better to start off with this being a standard property value.

exceptionfactory (commented):
Thanks for making the adjustments @mosermw. The latest version looks good except for one question I noted on the `identifiesExternalResource()` setting on the new Property Descriptor.

nifi-nar-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
mosermw: I admit I made the new property support external resources to make the customValidate logic simpler, and didn't think about it much after that.  Since the JSLT result filter must return true/false for each JSON object, I can't imagine it can get complicated enough to require needing external resources.  I modified the property descriptor to make it a standard property.

exceptionfactory (approved):
Thanks for working through the feedback @mosermw, the latest version looks good! +1 merging

PR #7929, opened 2023-10-25
NIFI-12249 FetchFTP and FetchSFTP processors now write failure reason to 'failure' attribute

annanys23:
# Summary

[NIFI-12249](https://issues.apache.org/jira/browse/NIFI-12249)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

Unit tests updated to ensure failure attribute is set on the flow file.  Some tests were updated to check for specific error conditions cited in the failure reasons written to the attribute.

Also, verified by creating the following flow:
GenerateFlowFile -> success -> FetchFTP
FetchFTP -> comms.failure, not.found, permission.denied -> LogAttribute
FetchFTP -> success -> funnel
LogAttribute -> success -> funnel

Configure FetchFTP with a bogus Hostname, Username, and Remote File
Ran the flow the saw the attribute 'failure' get logged:

Key: 'failure'
        Value: 'Failed to fetch content for StandardFlowFileRecord[uuid=a363e000-3c24-429a-8c0b-4e2d12a37345,claim=,offset=0,name=a363e000-3c24-429a-8c0b-4e2d12a37345,size=0] from filename file1 on remote host somehost:21 due to org.apache.nifi.processors.standard.socket.ClientConnectException: FTP Client connection failed [somehost:21]; routing to comms.failure'

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFTP.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFTP.java
@@ -48,7 +48,7 @@ import org.apache.nifi.processors.standard.util.FileTransfer;
     @WritesAttribute(attribute = "ftp.remote.filename", description = "The name of the remote file that was pulled"),
     @WritesAttribute(attribute = "filename", description = "The filename is updated to point to the filename fo the remote file"),
     @WritesAttribute(attribute = "path", description = "If the Remote File contains a directory name, that directory name will be added to the FlowFile using the 'path' attribute"),
-    @WritesAttribute(attribute = "failure", description = "If the transfer failed, the failure reason is written to this attribute")
+    @WritesAttribute(attribute = "fetch.failure.reason", description = "The name of the failure relationship applied when routing to any failure relationship")
 })
 @MultiProcessorUseCase(
     description = "Retrieve all files in a directory of an FTP Server",
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFTP.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFTP.java
@@ -47,7 +47,8 @@ import org.apache.nifi.processors.standard.util.FileTransfer;
     @WritesAttribute(attribute = "ftp.remote.port", description = "The port that was used to communicate with the remote FTP server"),
     @WritesAttribute(attribute = "ftp.remote.filename", description = "The name of the remote file that was pulled"),
     @WritesAttribute(attribute = "filename", description = "The filename is updated to point to the filename fo the remote file"),
-    @WritesAttribute(attribute = "path", description = "If the Remote File contains a directory name, that directory name will be added to the FlowFile using the 'path' attribute")
+    @WritesAttribute(attribute = "path", description = "If the Remote File contains a directory name, that directory name will be added to the FlowFile using the 'path' attribute"),
+    @WritesAttribute(attribute = "failure", description = "If the transfer failed, the failure reason is written to this attribute")
 })
 @MultiProcessorUseCase(
     description = "Retrieve all files in a directory of an FTP Server",
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFileTransfer.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFileTransfer.java
@@ -61,7 +61,7 @@ public abstract class FetchFileTransfer extends AbstractProcessor {
     static final AllowableValue COMPLETION_NONE = new AllowableValue("None", "None", "Leave the file as-is");
     static final AllowableValue COMPLETION_MOVE = new AllowableValue("Move File", "Move File", "Move the file to the directory specified by the <Move Destination Directory> property");
     static final AllowableValue COMPLETION_DELETE = new AllowableValue("Delete File", "Delete File", "Deletes the original file from the remote system");
-
+    static final String FAILURE_REASON_ATTRIBUTE = "fetch.failure.reason";
 
     static final PropertyDescriptor HOSTNAME = new PropertyDescriptor.Builder()
         .name("Hostname")
@@ -254,7 +254,6 @@ public abstract class FetchFileTransfer extends AbstractProcessor {
             transfer = transferWrapper.getFileTransfer();
         }
 
-        String failureReason = null;
         Relationship failureRelationship = null;
         boolean closeConnOnFailure = false;
         boolean provenanceEventOnFailure = false;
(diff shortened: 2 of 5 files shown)


mosermw:
I think the main motivation for requests like this are dataflow managers who don't want a separate UpdateAttribute processor creating a detailed notification message for each of the failure relationships.  "If I have 10 FetchSFTP processors then I need 30 UpdateAttribute processors, when really I only want 1".  At a large scale, little design decisions like this make a difference.

I wonder what the community would think of consolidating the FetchSFTP "comms.failure", "not.found" and "permission.denied" relationships into 1 "failure" relationship and moving forward with an attribute that describes the failure reason?  It would definitely be for NiFi 2.0 since it's a big breaking change. I suppose we would want @markap14 to weigh in on this as the original author.

exceptionfactory:
I can appreciate the desire to have fewer UpdateAttribute Processors in the case of handling different types of failures. Different Processors take different approaches in terms of failure handling, thinking of InvokeHTTP as one example of multiple categories of relationships.

The value of specific relationships is a clear differentiation of error types. Using FlowFile attributes can be helpful where there are well known status properties, such as HTTP response codes.

SFTP has some error codes, but even communication failures are different than other issues, like file not found or permission denied. One of the main reasons for having different relationships is that some errors could be retried.

From that angle, keeping the separate relationships has value for a number of use cases, even though it might complicate notification uses cases.

mosermw:
You make some very thoughtful points @exceptionfactory and it makes sense to provide fine grained error handling if the underlying library provides different error codes.  I spot checked some other Fetch processors which only have a single failure relationship, and I imagine their libraries don't enable finer grained handling as easily.

How about one last idea before I join you with opposing this?  What if we do not include the log message in the attribute but only include the relationship name in the attribute?  And change the attribute name to 'fetch.failure.reason' which aligns with the FetchParquet processor.  This should enable this PR to be much simpler and easier to maintain.

fetch.failure.reason=comms.failure or fetch.failure.reason=not.found or fetch.failure.reason=permission.denied

exceptionfactory:
Thanks for the reply @mosermw.

Introducing a FlowFile attribute that uses the relationship name seems like a good middle way in this scenario. It would support the use case of a single output Processor for failures if desired, providing a clear programmatic way to handle different types of failures.

What do you think @annanys23?

annanys23:
That works for me.  Thanks for the discussion.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFTP.java
```diff
@@ -47,7 +47,8 @@
     @WritesAttribute(attribute = "ftp.remote.port", description = "The port that was used to communicate with the remote FTP server"),
     @WritesAttribute(attribute = "ftp.remote.filename", description = "The name of the remote file that was pulled"),
     @WritesAttribute(attribute = "filename", description = "The filename is updated to point to the filename fo the remote file"),
-    @WritesAttribute(attribute = "path", description = "If the Remote File contains a directory name, that directory name will be added to the FlowFile using the 'path' attribute")
+    @WritesAttribute(attribute = "path", description = "If the Remote File contains a directory name, that directory name will be added to the FlowFile using the 'path' attribute"),
+    @WritesAttribute(attribute = "failure", description = "If the transfer failed, the failure reason is written to this attribute")
```
exceptionfactory, line 51: This should be changed to reflect the updated attribute name:
```suggestion
    @WritesAttribute(attribute = "fetch.failure.reason", description = "The name of the failure relationship applied when routing to any failure relationship")
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFileTransfer.java
```diff
@@ -296,6 +298,18 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
+            if (failureRelationship != null) {
+                attributes.put(FAILURE_REASON_ATTRIBUTE, failureRelationship.getName());
+                flowFile = session.putAllAttributes(flowFile, attributes);
+                session.transfer(session.penalize(flowFile), failureRelationship);
+                if (provenanceEventOnFailure) {
+                    session.getProvenanceReporter().route(flowFile, failureRelationship);
+                }
+                cleanupTransfer(transfer, closeConnOnFailure, transferQueue, host, port);
+                return;
+            }
```
exceptionfactory, line 311: Moving this logic to a separate block outside of the catch blocks introduces some potential logic changes, such as the addition of other attributes to the FlowFile, is that intentional? Otherwise it would be less of an impact to apply the attribute value in the appropriate catch blocks.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchSFTP.java
```diff
@@ -47,7 +47,8 @@
     @WritesAttribute(attribute = "sftp.remote.port", description = "The port that was used to communicate with the remote SFTP server"),
     @WritesAttribute(attribute = "sftp.remote.filename", description = "The name of the remote file that was pulled"),
     @WritesAttribute(attribute = "filename", description = "The filename is updated to point to the filename fo the remote file"),
-    @WritesAttribute(attribute = "path", description = "If the Remote File contains a directory name, that directory name will be added to the FlowFile using the 'path' attribute")
+    @WritesAttribute(attribute = "path", description = "If the Remote File contains a directory name, that directory name will be added to the FlowFile using the 'path' attribute"),
+    @WritesAttribute(attribute = "failure", description = "If the transfer failed, the failure reason is written to this attribute")
```
exceptionfactory, line 51: This should be changed to reflect the updated attribute name:
```suggestion
    @WritesAttribute(attribute = "fetch.failure.reason", description = "The name of the failure relationship applied when routing to any failure relationship")
```

exceptionfactory (changes requested):
Thanks for making the adjustments @annanys23. This looks closer to completion, but I noted updates needed for the `WritesAttribute` annotations, and also a concern about the logic change for handling attributes.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFileTransfer.java
annanys23: Yes, this was intentional.  I would think the other attributes are needed for a follow on processor to have context (ex: generate email notifications with similar content to what is being logged in the catch blocks).  Also, this was meant to remove some redundant code.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFileTransfer.java
```diff
@@ -254,36 +254,38 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
             transfer = transferWrapper.getFileTransfer();
         }
 
+        Relationship failureRelationship = null;
+        boolean closeConnOnFailure = false;
+        boolean provenanceEventOnFailure = false;
```
exceptionfactory, line 259: Reviewing the changes, it looks like this conditional check can be removed. Although the previous failure handling block did not call the Provenance Reporter on communications failure, this should be changed to always call `getProvenanceReporter().route()` for failures.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FetchFileTransfer.java
exceptionfactory: Thanks, that makes sense, and sounds good.

exceptionfactory (approved):
Thanks for working through the feedback and bringing this to completion @annanys23, and thanks for the additional perspective @mosermw! The latest version looks good! +1 merging

PR #7931, opened 2023-10-25
NIFI-12272 Add Formatter for Certificate Distinguished Names

exceptionfactory:
# Summary

[NIFI-12272](https://issues.apache.org/jira/browse/NIFI-12272) Adds a `PrincipalFormatter` interface and `StandardPrincipalFormatter` implementation that retrieves and formats Certificate Subject and Issuer Distinguished Names according to [RFC 1779](https://www.rfc-editor.org/rfc/rfc1779).

This approach preserves compatibility with the historical `getSubjectDN()` and `getIssuerDN()` methods from `X509Certificate` while using the recommended `getSubjectX500Principal()` and `getIssuerX500Principal()` methods.

These changes restore compatibility with existing NiFi versions, recently changed for [NIFI-12142](https://issues.apache.org/jira/browse/NIFI-12142).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-security-cert/src/main/java/org/apache/nifi/security/cert/StandardPrincipalFormatter.java
+++ b/nifi-commons/nifi-security-cert/src/main/java/org/apache/nifi/security/cert/StandardPrincipalFormatter.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.security.cert;
+
+import javax.security.auth.x500.X500Principal;
+import java.security.cert.X509Certificate;
+import java.util.Objects;
+
+/**
+ * Standard Principal Formatter implementation returns Subject and Issuer formatted according to RFC 1779 following the convention of getSubjectDN and getIssuerDN methods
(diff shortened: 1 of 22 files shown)


nifi-commons/nifi-security-cert/src/main/java/org/apache/nifi/security/cert/StandardPrincipalFormatter.java
```diff
@@ -0,0 +1,69 @@
...
+     * @return Issuer Distinguished Name formatted according to RFC 1779
+     */
+    @Override
+    public String getIssuer(final X509Certificate certificate) {
+        Objects.requireNonNull(certificate, "Certificate required");
+        return getFormatted(certificate.getIssuerX500Principal());
+    }
+
+    private String getFormatted(final X500Principal principal) {
+        return principal.getName(X500Principal.RFC1779);
```
ChrisSamo632, line 67: Do you foresee a need for other formats in the future, e.g. `RFC2253`? If so, should we make the format configurable for the client?

If there's no apparent need right now (I can't see an immediate one if the LDAP connectivity wouldn't benefit from the new formatter/approach), then happy for this to be something that can be added later if/when the need arises

ChrisSamo632 (commented):
Looks OK to me @exceptionfactory, although one question about possible formatting configurability to consider

nifi-commons/nifi-security-cert/src/main/java/org/apache/nifi/security/cert/StandardPrincipalFormatter.java
exceptionfactory: If there is a need for `RFC2253` in other contexts, that could be a separate implementation of the `PrincipalFormatter` interface. The purpose of the `StandardPrincipalFormatter` was to maintain historical consistency when it comes to string representations, without ruling out the possibility of other options in different contexts.

exceptionfactory: Based on current references, making this configurable would probably create more confusion, given the nature of identity mapping and FlowFile attribute evaluation. However, this certainly doesn't rule out other approaches if needed.

ChrisSamo632: That makes sense - a separate implementation of the interface for a different formatting if/when needed in future is fine

PR #7940, opened 2023-10-26
NIFI-12255 refactor PutElasticsearchRecord and PutElasticsearchJson relationships to be more consistent with other processors

ChrisSamo632:
# Summary

[NIFI-12255](https://issues.apache.org/jira/browse/NIFI-12255) refactor PutElasticsearchRecord and PutElasticsearchJson relationships to be more consistent with other processors

Additionally:
- added ability to unit test `migrateRelationships` for processor refactoring
- addressed issue with `StandardTestRunner#migrateProperties` where some properties may be `null`
- addressed some warnings in `MockProcessContext` and `StandardTestRunner` implementations, improving some logging for easier test failure diagnostics

**Breaking Change**: addition of the `successfull` relationship to `PutElasticsearchJson` will require the connection to be added to a downstream processor (or terminated within the processor config). Relationship has been added for consistency with `PutElasticsearchRecord`, along with `original` (automatically renamed from `success`) and `errors`.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-mock/src/main/java/org/apache/nifi/util/RelationshipMigrationResult.java
+++ b/nifi-mock/src/main/java/org/apache/nifi/util/RelationshipMigrationResult.java
@@ -25,10 +25,10 @@ public interface RelationshipMigrationResult {
     /**
      * @return a mapping of previous relationship names to the new names of those relationships
      */
-    Map<String, Set<String>> getRelationshipsSplit();
+    Map<String, Set<String>> getPreviousRelationships();
 
     /**
      * @return a mapping of previous relationship names to the new names of those relationships
      */
-    Map<String, String> getRelationshipsRenamed();
+    Map<String, String> getRenamedRelationships();
 }
--- a/nifi-mock/src/main/java/org/apache/nifi/util/RelationshipMigrationResult.java
+++ b/nifi-mock/src/main/java/org/apache/nifi/util/RelationshipMigrationResult.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nifi.util;
+
+import java.util.Map;
+import java.util.Set;
+
+public interface RelationshipMigrationResult {
+
+    /**
+     * @return a mapping of previous relationship names to the new names of those relationships
+     */
+    Map<String, Set<String>> getRelationshipsSplit();
+
+    /**
+     * @return a mapping of previous relationship names to the new names of those relationships
+     */
+    Map<String, String> getRelationshipsRenamed();
+}
--- a/nifi-mock/src/main/java/org/apache/nifi/util/StandardProcessorTestRunner.java
+++ b/nifi-mock/src/main/java/org/apache/nifi/util/StandardProcessorTestRunner.java
@@ -63,6 +63,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
@@ -241,8 +242,7 @@ public class StandardProcessorTestRunner implements TestRunner {
                     unscheduledRun = true;
                     unSchedule();
                 }
-            } catch (final Exception e) {
-                e.printStackTrace();
+            } catch (final InterruptedException | ExecutionException e) {
             }
         }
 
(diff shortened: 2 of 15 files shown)


nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/AbstractPutElasticsearch.java
```diff
@@ -55,10 +55,41 @@
 import java.util.stream.Collectors;
 
 public abstract class AbstractPutElasticsearch extends AbstractProcessor implements ElasticsearchRestProcessor {
+    static final Relationship REL_ORIGINAL = new Relationship.Builder()
+            .name("original")
+            .description("All flowfiles that are sent to Elasticsearch without request failures go to this relationship.")
+            .build();
+
+    static final Relationship REL_SUCCESSFUL = new Relationship.Builder()
+            .name("successful")
```
exceptionfactory, line 64: Is there a particular reason for calling this `successful` as opposed to the much more common `success`?

exceptionfactory (commented):
Thanks for working on these changes @ChrisSamo632.

The general renaming makes sense, but I raised a particular question about `successful` as opposed to the more conventional `success`. I see the the current `success` is renamed to `original`, so is that part of the consideration? Even so, it seems like this is an opportunity to align with standard conventions.

Regarding the unit testing of migration, I'm not sure about the relative value of unit testing those changes. The unit tests as they stand appear to be fairly involved, so if this is going to be a first-class `nifi-mock` capability, I think it would be better to evaluate that on its own, as opposed to including in this pull request.

nifi-nar-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/AbstractPutElasticsearch.java
ChrisSamo632: That, and I think renaming the relationship might impact other ES processors (I'll have to double check)

ChrisSamo632:
RE the migration testing, I discussed this with @markap14 at the time. I find some issues with the property mock processing, so fixed that (I'll have to reverse engineer what the problem was now, if more details are needed).

Adding the relationship mocking is something that could be separated out, although at the time it helped me confirm the updates I wanted to make and I thought it might be helpful for others in the future. It does, of course, increase this PR a bit though. Happy to create a separate jira ticket and either link it here (to reflect the additional change), or try to separate out the changes if people think it's necessary

ChrisSamo632:
I've rebased from latest `main` to address merge conflicts.

Let me know your thoughts on previous comments @exceptionfactory.

nifi-mock/src/main/java/org/apache/nifi/util/StandardProcessorTestRunner.java
```diff
@@ -229,33 +226,32 @@ public void run(final int iterations, final boolean stopOnFinish, final boolean
...
-                        unscheduledRun = true;
-                        unSchedule();
-                    }
-                } catch (final Exception e) {
+                if (++finishedCount == 1 && stopOnFinish) {
+                    unscheduledRun = true;
+                    unSchedule();
                 }
+            } catch (final Exception e) {
+                e.printStackTrace();
```
exceptionfactory, line 245: This use of `printStackTrace()` seems problematic, better throw a more particular type of exception?

nifi-mock/src/main/java/org/apache/nifi/util/RelationshipMigrationResult.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Map;
+import java.util.Set;
+
+public interface RelationshipMigrationResult {
+
+    /**
+     * @return a mapping of previous relationship names to the new names of those relationships
+     */
+    Map<String, Set<String>> getRelationshipsSplit();
```
exceptionfactory, line 28: The `Split` naming is not quite clear. Perhaps `getPreviousRelationships()`?

nifi-mock/src/main/java/org/apache/nifi/util/RelationshipMigrationResult.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * @return a mapping of previous relationship names to the new names of those relationships
+     */
+    Map<String, String> getRelationshipsRenamed();
```
exceptionfactory, line 33: Aligning with the other method, perhaps `getRenamedRelationships()`?

exceptionfactory (commented):
Thanks for your patience on the pull request @ChrisSamo632. I noted a couple more questions, but this seems close to completion.

nifi-mock/src/main/java/org/apache/nifi/util/StandardProcessorTestRunner.java
ChrisSamo632: The `Exception` capture is nothing new, just previously it was being silently ignored, so I thought I'd at least indicate that the `catch` had been triggered, without potentially causing problems for lots of tests across the existing estate

Happy to revert this and silently swallow the `InterruptedException` & `ExecutionException`s, although I'll change it from catching all `Exception`s at the least

This approach matches other similar handling elsewhere within the `TestRunner`

exceptionfactory: Thanks, I think it is best to avoid the `printStackTrace()` for now.

ChrisSamo632:
Rebased against latest `main` to refactor changed into new `nifi-extension-bundles` module

joewitt:
@ChrisSamo632 Thanks for rebasing and also leveraging migrate properties!

exceptionfactory (approved):
Thanks for working through the feedback and keeping this PR updated with the latest changes @ChrisSamo632, the latest version looks good. +1 merging

PR #7952, opened 2023-10-29
NIFI-8932: Add capability to skip first N rows in CSVReader

mattyb149:
# Summary

[NIFI-8932](https://issues.apache.org/jira/browse/NIFI-8932) This PR adds the capability to skip the first N rows of an incoming file to CSVReader, in the case of headers or other invalid records at the top of the FlowFile.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/AbstractCSVRecordReader.java
+++ b/nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/AbstractCSVRecordReader.java
@@ -33,6 +33,8 @@ abstract public class AbstractCSVRecordReader implements RecordReader {
     protected final boolean ignoreHeader;
     private final boolean trimDoubleQuote;
 
+    protected final int skipTopRows;
+
     protected final Supplier<DateFormat> LAZY_DATE_FORMAT;
     protected final Supplier<DateFormat> LAZY_TIME_FORMAT;
     protected final Supplier<DateFormat> LAZY_TIMESTAMP_FORMAT;
@@ -44,12 +46,13 @@ abstract public class AbstractCSVRecordReader implements RecordReader {
     protected final RecordSchema schema;
 
     AbstractCSVRecordReader(final ComponentLog logger, final RecordSchema schema, final boolean hasHeader, final boolean ignoreHeader,
-                            final String dateFormat, final String timeFormat, final String timestampFormat, final boolean trimDoubleQuote) {
+                            final String dateFormat, final String timeFormat, final String timestampFormat, final boolean trimDoubleQuote, final int skipTopRows) {
         this.logger = logger;
         this.schema = schema;
         this.hasHeader = hasHeader;
         this.ignoreHeader = ignoreHeader;
         this.trimDoubleQuote = trimDoubleQuote;
+        this.skipTopRows = skipTopRows;
 
         if (dateFormat == null || dateFormat.isEmpty()) {
             this.dateFormat = null;
--- a/nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/AbstractCSVRecordReader.java
+++ b/nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/AbstractCSVRecordReader.java
@@ -17,12 +17,19 @@
 
 package org.apache.nifi.csv;
 
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.io.input.BOMInputStream;
 import org.apache.nifi.logging.ComponentLog;
 import org.apache.nifi.serialization.RecordReader;
 import org.apache.nifi.serialization.record.DataType;
 import org.apache.nifi.serialization.record.RecordFieldType;
 import org.apache.nifi.serialization.record.RecordSchema;
 import org.apache.nifi.serialization.record.util.DataTypeUtils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
 import java.text.DateFormat;
 import java.util.function.Supplier;
 
@@ -45,8 +52,12 @@ abstract public class AbstractCSVRecordReader implements RecordReader {
 
     protected final RecordSchema schema;
 
-    AbstractCSVRecordReader(final ComponentLog logger, final RecordSchema schema, final boolean hasHeader, final boolean ignoreHeader,
-                            final String dateFormat, final String timeFormat, final String timestampFormat, final boolean trimDoubleQuote, final int skipTopRows) {
+    protected final Reader inputStreamReader;
+
+    AbstractCSVRecordReader(final InputStream in, final ComponentLog logger, final RecordSchema schema, final CSVFormat csvFormat, final boolean hasHeader, final boolean ignoreHeader,
+                            final String dateFormat, final String timeFormat, final String timestampFormat, final String encoding, final boolean trimDoubleQuote, final int skipTopRows)
+            throws IOException {
+
         this.logger = logger;
         this.schema = schema;
         this.hasHeader = hasHeader;
@@ -76,6 +87,14 @@ abstract public class AbstractCSVRecordReader implements RecordReader {
         } else {
             this.timestampFormat = timestampFormat;
             LAZY_TIMESTAMP_FORMAT = () -> DataTypeUtils.getDateFormat(timestampFormat);
+        }
+
+        final InputStream bomInputStream = BOMInputStream.builder().setInputStream(in).get();
+        inputStreamReader = new InputStreamReader(bomInputStream, encoding);
+
+        // Skip the number of rows at the "top" as specified
+        for (int i = 0; i < skipTopRows; i++) {
+            readNextRecord(inputStreamReader, csvFormat.getRecordSeparator());
         }
     }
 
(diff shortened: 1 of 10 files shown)


exceptionfactory (changes requested):
Thanks for working on this new feature @mattyb149. The concept makes sense, but the tests are failing and the approach implemented in `restNextRecord` seems problematic.

Instead of creating an `InputStreamReader` with the platform-default encoding, it seems like the raw `InputStream` could be used, and wrapped in a [PushbackInputStream](https://docs.oracle.com/javase/8/docs/api/java/io/PushbackInputStream.html). The Pushback approach would support returning extra bytes to the stream, which actual Reader would then handle.

exceptionfactory:
@mattyb149 Just to streamline the reviews, I am closing this pull request for now given the test failures across the board, and the implementation concerns. Feel free to reopen whenever you are ready. Thanks!

mattyb149:
Reopening this as I'm actively working it. I realized I hadn't passed in the `Character Set` property value into the InputStreamReader so I'm trying to fix the code/tests using that first. If not (or if you still object to the Reader at all) I can try the PushbackInputStream. The only caveat there is that the record separator can be an arbitrary string so I need to create the pushback buffer of that size ("n") and push back only "n-1" bytes. I figured the reader would do something similar and it makes the code easier to read, so if it works with the Reader and the correct encoding I'd like to go with that.

exceptionfactory:
> Reopening this as I'm actively working it. I realized I hadn't passed in the `Character Set` property value into the InputStreamReader so I'm trying to fix the code/tests using that first. If not (or if you still object to the Reader at all) I can try the PushbackInputStream. The only caveat there is that the record separator can be an arbitrary string so I need to create the pushback buffer of that size ("n") and push back only "n-1" bytes. I figured the reader would do something similar and it makes the code easier to read, so if it works with the Reader and the correct encoding I'd like to go with that.

That's a good point about the arbitrary string separator. The Reader will do character decoding based on the configured Character Set, so the best approach probably depends on how to evaluate the string separator.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/CSVRecordReader.java
```diff
@@ -47,11 +47,17 @@ public class CSVRecordReader extends AbstractCSVRecordReader {
...
+            throws IOException {
+        super(logger, schema, hasHeader, ignoreHeader, dateFormat, timeFormat, timestampFormat, trimDoubleQuote, skipTopRows);
 
         final InputStream bomInputStream = BOMInputStream.builder().setInputStream(in).get();
-        final Reader reader = new InputStreamReader(bomInputStream, encoding);
+        final Reader inputStreamReader = new InputStreamReader(bomInputStream, encoding);
+
+        // Skip the number of rows at the "top" as specified
+        for (int i = 0; i < skipTopRows; i++) {
+            readNextRecord(inputStreamReader, csvFormat.getRecordSeparator());
```
exceptionfactory, line 59: This use of `readNextRecord` ignores the return, which makes sense in this scenario. Are there other places where the return value is used, or would it be worth removing that from `readNextRecord` to avoid unnecessary string building?

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/CSVRecordReader.java
```diff
@@ -155,4 +159,59 @@ private List<RecordField> getRecordFields() {
...
+     * @param reader the Reader providing the input
+     * @param recordSeparator the String specifying the end of a record in the input
+     * @return a String created from the input until the record separator is reached.
+     * @throws IOException if an error occurs during reading
+     */
+    protected String readNextRecord(Reader reader, String recordSeparator) throws IOException {
+        int indexIntoSeparator = 0;
+        int recordSeparatorLength = recordSeparator.length();
+        StringBuilder lineBuilder = new StringBuilder();
+        StringBuilder separatorBuilder = new StringBuilder();
```
exceptionfactory, line 176: Is it necessary to build the separator as opposed to appending the `recordSeparator` to the `lineBuilder`?

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/CSVRecordReader.java
```diff
@@ -155,4 +159,59 @@ private List<RecordField> getRecordFields() {
...
+            // This defensive check limits a record size to 2GB, this prevents out-of-memory errors if the record separator
+            // is not present in the input (or at least in the first 2GB)
+            if (indexIntoSeparator == Integer.MAX_VALUE) {
+                throw new IOException("2GB input threshold reached, the record is either larger than 2GB or the separator "
+                        + "is not found in the first 2GB of input. Ensure the Record Separator is correct for this FlowFile.");
+            }
+            code = reader.read();
+        }
+
+        // The end of input has been reached without the record separator being found, throw an exception with the string so far
```
exceptionfactory, line 213: This comment does not appear to match the implementation.

exceptionfactory (commented):
Thanks for making the adjustments @mattyb149. I noted a couple questions related to the `readNextRecord()` implementation, particularly whether it is necessary to maintain and return the built string. I may be missing other uses, but otherwise it seems like it could be simplified if the return string is not used.

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/AbstractCSVRecordReader.java
```diff
@@ -74,6 +88,14 @@ abstract public class AbstractCSVRecordReader implements RecordReader {
             this.timestampFormat = timestampFormat;
             LAZY_TIMESTAMP_FORMAT = () -> DataTypeUtils.getDateFormat(timestampFormat);
         }
+
+        final InputStream bomInputStream = BOMInputStream.builder().setInputStream(in).get();
+        inputStreamReader = new InputStreamReader(bomInputStream, encoding);
```
dan-s1, line 93:
```suggestion
        this.inputStreamReader = new InputStreamReader(bomInputStream, encoding);
```

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/AbstractCSVRecordReader.java
```diff
@@ -158,4 +180,46 @@ protected String trim(String value) {
...
+     */
+    protected void readNextRecord(Reader reader, String recordSeparator) throws IOException {
+        int indexIntoSeparator = 0;
+        int recordSeparatorLength = recordSeparator.length();
+        int code = reader.read();
+        while (code != -1) {
+            char nextChar = (char)code;
+            if (recordSeparator.charAt(indexIntoSeparator) == nextChar) {
+                if (++indexIntoSeparator == recordSeparatorLength) {
+                    // We have matched the separator, return the string built so far
```
dan-s1, line 199:
```suggestion
                    // Short circuit as the separator indicates a record has been read
```

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/AbstractCSVRecordReader.java
```diff
@@ -158,4 +180,46 @@ protected String trim(String value) {
...
+                // and check if it is the first character of the separator.
+                indexIntoSeparator = 0;
+                if (recordSeparator.charAt(indexIntoSeparator) == nextChar) {
+                    // This character is the beginning of the record separator, keep it
+                    if (++indexIntoSeparator == recordSeparatorLength) {
+                        // We have matched the separator, return the string built so far
+                        return;
+                    }
+                }
+            }
```
dan-s1, line 213: Instead of all the comments here, make a method which conveys the intention of what you are trying to accomplish.


```
char nextChar = (char)code;
if (recordSeparator.charAt(indexIntoSeparator) == nextChar) {
    ++indexIntoSeparator;
    if (hasReachedEndOfRecord(indexIntoSeparator, recordSeparatorLength) ) {
        return;
    }
} else {
    // The character didn't match the expected one in the record separator, reset the separator matcher
    // and check if it is the first character of the separator.
    indexIntoSeparator = 0;
    if (recordSeparator.charAt(indexIntoSeparator) == nextChar) {
        ++indexIntoSeparator;
        if (hasReachedEndOfRecord(indexIntoSeparator, recordSeparatorLength) ) {
         return;
        }
    }
}

private boolean hasReachedEndOfRecord(int indexIntoSeparator, int recordSeparatorLength) {
        return indexIntoSeparator == recordSeparatorLength;
}
```

nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/AbstractCSVRecordReader.java
```diff
@@ -158,4 +180,46 @@ protected String trim(String value) {
...
+        int indexIntoSeparator = 0;
+        int recordSeparatorLength = recordSeparator.length();
+        int code = reader.read();
+        while (code != -1) {
+            char nextChar = (char)code;
+            if (recordSeparator.charAt(indexIntoSeparator) == nextChar) {
+                if (++indexIntoSeparator == recordSeparatorLength) {
+                    // We have matched the separator, return the string built so far
+                    return;
+                }
```
dan-s1, line 201: @mattyb149 @exceptionfactory A concern I have with this logic is what happens when the record separator is escaped in the data? How will you tell whether you have an end of record or an escaped record separator.  I did a quick look on line and found that a new line character which usually is the end of a CSV record can be embedded in the data.

mattyb149: I'll look at capturing escape characters, maybe some sort of flag indicating that we're in an escape sequence. Good catch!

mattyb149: Actually come to think of it, this capability is to skip lines aren't AREN'T valid CSV, otherwise you can use SampleRecord. Because we aren't assuming the skipped rows to be valid CSV, we should be able to just match on the newline character. I'll put in some code anyway but what are your thoughts?

markap14:
@mattyb149 I think I'm a -1 on the idea of skipping lines in the CSV Reader. Any sort of workarounds like this for invalid data significantly complicate the codebase. Rather, a simple RouteText should be used to strip out the unwanted line. I commented to this effect on the Jira.

mattyb149:
That's a good point, if the CSV file has gone through a record-based processor you can also use RouteText to skip the last N lines too using the `${record.count}` attribute. Closing this and the Jira.

PR #7965, opened 2023-10-31
NIFI-12229 Refine NiFiRegistry client lifecycle: replacing actual client when SSL service changes

simonbence:
# Summary

[NIFI-12229](https://issues.apache.org/jira/browse/NIFI-12229)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-flow-registry-client-bundle/nifi-flow-registry-client-services/src/main/java/org/apache/nifi/registry/flow/NifiRegistryFlowRegistryClient.java
+++ b/nifi-nar-bundles/nifi-flow-registry-client-bundle/nifi-flow-registry-client-services/src/main/java/org/apache/nifi/registry/flow/NifiRegistryFlowRegistryClient.java
@@ -64,14 +64,34 @@ public class NifiRegistryFlowRegistryClient extends AbstractFlowRegistryClient {
             .identifiesControllerService(SSLContextService.class)
             .build();
 
+    private volatile Integer sslContextServiceHash;
     private volatile String registryUrl;
     private volatile NiFiRegistryClient registryClient;
+
+    private boolean shouldGetNewRegistryClient(final FlowRegistryClientConfigurationContext context) {
+        final String proposedUrl = getProposedUri(context);
+
+        if (!proposedUrl.equals(registryUrl)) {
+            return true;
+        }
+
+        final SSLContextService sslContextService = context.getProperty(SSL_CONTEXT_SERVICE).asControllerService(SSLContextService.class);
+        final Integer newSslContextServiceHash = getSslContextServiceHash(sslContextService);
+
+        if ((sslContextServiceHash == null && newSslContextServiceHash != null)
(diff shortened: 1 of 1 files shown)


pvillard31:
I feel like the reported problem could happen in any implementation of the Registry Client, not just this one. Instead of taking this approach that is quite specific to the SSL controller service, should we consider adding something at the framework level to trigger a "refresh" method anytime the configuration of the component is changed? It would then be up to the component to implement whatever makes sense in this refresh method. Thoughts?

Note: I'm fine with this approach to solve the immediate problem but I feel like we could have a more generic approach in case there are some other implementations of the Registry Client interface.

exceptionfactory (changes requested):
Thanks for working on this issue @simonbence, and thanks for the perspective @pvillard31.

I am concerned about this approach because it seems to be attempting to solve a framework-level concern at the component level.

Some of this appears to be the result of introducing the Flow Registry Client as a component without having a lifecycle for the component, such as enabling and disabling of Controller Services. This is apparent from the fallback to the System SSL Context, which would have similar concerns if selected as the configuration approach.

Instead of this approach, it seems like the Flow Registry Client should be reinitialized when a referencing Controller Service is enabled. Although this approach would be more involved, it seems like something along this line is necessary.

simonbence:
@pvillard31 @exceptionfactory  Thanks for the feedback!

As of this given case, the behaviour is specific to the way this particular implementation works. Other implementations, if they are caching some internal state depending on "controller service" property might have a similar challenge.

@exceptionfactory You are perfectly right, the Flow Registry Client currently has no such lifecycle states like active-deactivated, that is correct. At the time of changing the Client to a component, this was an intentional choice due to the number of questions would come with this behavioural change (and due to avoiding any break in the previous behaviour). The questions include: what should happen with the existing versioned flows when the registry is deactivated? It looks like a different case compared to the situation when the registry is unintenionally unreachable and might result some confusion. An other possibility you mention is to "hook on" some existing event like validation or as you mention, the (de)activation of the dependent Controller Services but I am not aware of any support in the framework for this.

What I propose to do is to move on with this fix (which in some cases can cause misbehaviour as it is now) and for a wider scope solution it would be beneficial to start a discussion about the level of support we aim for: I can easily imagine that an additional hookpoint might be useful for other components as well.

markap14:
Yeah, I think it's better to solve this at the framework level as well. The approach I would recommend would probably be to update the Registry Client Node object so that when validation is performed, if it goes from INVALID to VALID, we call any method annotated with `@OnEnabled`. That way, if a Controller Service that it uses were to be disabled and re-enabled, it would call the `@OnEnabled` method and that, in turn, would be responsible for recreating the underlying client based on the new SSL Context configuration.

simonbence:
As suggested I will raise a new PR with a framework level solution. I think, @markap14 's suggestion might be the best balance between feature richness and unnecessary complexity I will aim for that.

PR #7980, opened 2023-11-04
NIFI-12261 Remove Hortonworks Schema References

exceptionfactory:
# Summary

[NIFI-12261](https://issues.apache.org/jira/browse/NIFI-12261) Removes Hortonworks Schema references from Schema Registry Services, including the removal of the `Schema Protocol Version` property, which was specific to Hortonworks Schema Registry formats. The removal includes use of `migrateProperties()` to ensure that the Schema Protocol Version property is removed from existing services. This removal follows along with the removal of the Hortonworks Schema Registry components completed in [NIFI-11208](https://issues.apache.org/jira/browse/NIFI-11208).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


pvillard31:
I'm not sure this is the approach we want to take here. We have users with a mix of hortonworks / confluent setup (schema encoding following the hortonworks approach but sending the data into confluent, and vice versa). We removed the HWX Schema Registry but I think removing the schema encoding strategies does not have the same level of impact and one could wonder why we keep Confluent but not HWX/Cloudera. A cleaner approach would probably to have the schema encoding as an extension point so that it can be part of the NARs but it sounds like a significant effort that is not really worth it. Wouldn't be easier here to just rename the classes using the Cloudera name and keep both Cloudera and Confluent? I could even see value in reverting the removal of the HWX Schema Registry and just rename it given that we keep the equivalent for Confluent and we're adding support for additional Schema Registries (Glue, Apicurio, etc). Thoughts?

exceptionfactory:
Thanks for the feedback @pvillard31.

I think the best solution would be to create a new Controller Service interface for writing schemas.

I agree that we don't want to artificially limit support for any given type of Schema Registry. From that perspective, having formatting capabilities specific to any particular vendor at this level makes it challenging without turning this into an extension point.

It is worth noting that the current capabilities include several older versions, and I'm not sure if there are any other more recent changes. It is also worth noting that the [Registry](https://github.com/hortonworks/registry) code still uses the `hortonworks` namespace, which is understandable for historical reasons.

With that in mind, perhaps we should avoid any renaming, and instead target the introduction of an extension point before the 2.0.0 GA release?

pvillard31:
Yeah I don't feel like there is any perfect solution. Some time ago, I discussed with @bbende the option of having an extension point but I think I remember that it would also raise some questions, can't exactly remember right now though.

exceptionfactory:
Thanks again for the feedback @pvillard31, I will evaluate the changes necessary to create a new extensible Controller Service interface for reading Schema Registry Access information.

PR #7995, opened 2023-11-07
NIFI-12192 Add Codecov Reporting with JaCoCo

exceptionfactory:
# Summary

[NIFI-12192](https://issues.apache.org/jira/browse/NIFI-12192) Adds code coverage for unit tests using the [JaCoCo Maven Plugin](https://www.jacoco.org/jacoco/trunk/doc/maven.html) reporting to [Codecov](https://about.codecov.io/).

Codecov provides free reporting capabilities for open source projects and enables integration with GitHub workflows using the [Codecov Action](https://github.com/codecov/codecov-action).

The implementation of code coverage reporting updates the GitHub `ci-workflow` to run the Ubuntu Linux Java 21 job with JaCoCo coverage enabled. The JaCoCo plugin writes an aggregated report in XML, using the `nifi-code-coverage` module to aggregate coverage information based on an enumeration of dependencies. The Maven module is necessary for selecting and aggregating coverage reporting information. The Codecov Action reads the JaCoCo XML and sends it to Codecov for processing.

The JaCoCo prepare-agent goal updates the `argLine` property, which required removing custom locale information from the Ubuntu Linux job, instead relying on the default settings of the GitHub runner.

Updates include adding a Codecov badge to the README for displaying coverage on the main branch.

Current code coverage information is available in [Codecov for this feature branch](https://app.codecov.io/gh/exceptionfactory/nifi/tree/NIFI-12192).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


.github/workflows/ci-workflow.yml
```diff
@@ -160,22 +159,19 @@ jobs:
           ${{ env.MAVEN_COMPILE_COMMAND }}
       - name: Maven Verify
         env:
-          NIFI_CI_LOCALE: >-
-            -Duser.language=en
-            -Duser.country=AU
-          SUREFIRE_OPTS: >-
```
ChrisSamo632, line 166: It might be possible to retain these args (thus allowing use of JaCoCo in other Actions in future if desired, without needing to change the `mvn` args, by using something like:

```xml
                    <plugin>
                        <groupId>org.apache.maven.plugins</groupId>
                        <artifactId>maven-surefire-plugin</artifactId>
                        <configuration>
                            <argLine>@{jaCoCoArgLine} -Xmx1024m</argLine>
                        </configuration>
                    </plugin>
```

within `<pluginManagement/>`, and (within `<build/>`):

```xml
                    <plugin>
                        <groupId>org.jacoco</groupId>
                        <artifactId>jacoco-maven-plugin</artifactId>
                        <executions>
                            <execution>
                                <id>default-prepare-agent</id>
                                <goals>
                                    <goal>prepare-agent</goal>
                                </goals>
                                <configuration>
                                    <propertyName>jaCoCoArgLine</propertyName>
                                </configuration>
                            </execution>
                        </executions>
                    </plugin>
```

pom.xml
```diff
@@ -29,6 +29,7 @@
         <module>nifi-framework-api</module>
         <module>nifi-server-api</module>
         <module>nifi-bootstrap</module>
+        <module>nifi-code-coverage</module>
```
ChrisSamo632, line 32: I think the other possibie option, to avoid the new `nifi-code-coverage` sub-module, which may become out of step when new things are added to the repo, would be to add:

```xml
<properties>
  <jacoco.skip>true</jacoco.skip>
</properties>
```

to all sub-modules where we don't want to collect coverage, e.g. NAR and Bundle sub-modules. So this would remove the need for a new sub-module to maintain, but would mean updating lots of other existing sub-modules instead, then people understanding when to include the `skip` property for new sub-modules in future.

I don't think either option is ideal, but thought I'd raise the alternative for consideration - I'd be happy with either considered option

ChrisSamo632:
@exceptionfactory thanks for the update, the code coverage will help improve confidence in the code base going forward.

I've suggeted a couple of options to think about for improving the use of JaCoCo - more for consideration than blockers to merging this PR.

As discussed elsewhere, would it be sensible to update the `integration-tests-ci` profile to skip all `ITRedisDistributedMapCacheClientService` ITs as part of this PR - more of them failed the first time in the PR checks, but then succeeded on the retry?

pom.xml
exceptionfactory: Thanks for the feedback. I considered the option of individually running the JaCoCo [report](https://www.jacoco.org/jacoco/trunk/doc/report-mojo.html) goal, but I went with the [report-aggregate](https://www.jacoco.org/jacoco/trunk/doc/report-aggregate-mojo.html) goal and aggregating module to avoid changing individual Maven configurations.

I agree that neither solution is perfect, as each approach requires an extra element to consider when adding new modules. However, after making recent changes to remove the need for JavaDoc and source plugin skipping in individual modules, I went with the centralized approach. Another positive of this option is that if we don't see coverage information for a particular module, there is only one place to check, versus going through multiple modules.

.github/workflows/ci-workflow.yml
exceptionfactory: I considered similar options as described in the [prepare-agent](https://www.jacoco.org/jacoco/trunk/doc/prepare-agent-mojo.html) goal documentation, but part of the difficulty is that the build workflow was setting a System property, which JaCoCo also needs to set. We could change the System property in the build workflow to a different name, and then use that name for late property evaluation in the maven-surefire-plugin configuration, but that would have involved changes all of the build profiles. I kept the scope of changes narrow in this case, but we could revisit in the future if we need to introduce other arguments.

exceptionfactory:
> As discussed elsewhere, would it be sensible to update the `integration-tests-ci` profile to skip all `ITRedisDistributedMapCacheClientService` ITs as part of this PR - more of them failed the first time in the PR checks, but then succeeded on the retry?

Thanks for the review and feedback @ChrisSamo632! I have also noticed that methods in that class have been unreliable in automated tests, so I changed the exclusion to apply to the entire `ITRedisDistributedMapCacheClientService` class.

ChrisSamo632 (approved):
@exceptionfactory thanks for the updates and useful discussion RE options

LGTM 👍 will merge

PR #7998, opened 2023-11-08
NIFI-12335 Fix MiNiFi startup issue

ferencerdei:
# Summary

[NIFI-12335](https://issues.apache.org/jira/browse/NIFI-12335)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
+++ b/minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
@@ -26,6 +26,7 @@ import static java.util.Optional.ofNullable;
 import static java.util.UUID.randomUUID;
 import static java.util.stream.Collectors.toMap;
 import static org.apache.commons.lang3.StringUtils.EMPTY;
+import static org.apache.nifi.flow.ScheduledState.ENABLED;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonFactory;
@@ -90,6 +91,9 @@ public class FlowEnrichService {
         }
 
         VersionedDataflow versionedDataflow = parseVersionedDataflow(flowCandidate);
+        versionedDataflow.setReportingTasks(ofNullable(versionedDataflow.getReportingTasks()).orElseGet(ArrayList::new));
+        versionedDataflow.setRegistries(ofNullable(versionedDataflow.getRegistries()).orElseGet(ArrayList::new));
+        versionedDataflow.setControllerServices(ofNullable(versionedDataflow.getControllerServices()).orElseGet(ArrayList::new));
 
         Optional<Integer> maxConcurrentThreads = ofNullable(minifiProperties.getProperty(MiNiFiProperties.NIFI_MINIFI_FLOW_MAX_CONCURRENT_THREADS.getKey()))
             .map(Integer::parseInt);
@@ -103,10 +107,12 @@ public class FlowEnrichService {
             rootGroup.setInstanceIdentifier(randomUUID().toString());
         }
 
+        rootGroup.getControllerServices().forEach(cs -> cs.setScheduledState(ENABLED));
+
         Optional<VersionedControllerService> commonSslControllerService = createCommonSslControllerService();
         commonSslControllerService
             .ifPresent(sslControllerService -> {
-                List<VersionedControllerService> currentControllerServices = ofNullable(versionedDataflow.getControllerServices()).orElseGet(ArrayList::new);
+                List<VersionedControllerService> currentControllerServices = new ArrayList<>(versionedDataflow.getControllerServices());
                 currentControllerServices.add(sslControllerService);
                 versionedDataflow.setControllerServices(currentControllerServices);
             });
@@ -118,7 +124,7 @@ public class FlowEnrichService {
 
         createProvenanceReportingTask(commonSslControllerService.map(VersionedComponent::getInstanceIdentifier).orElse(EMPTY))
             .ifPresent(provenanceReportingTask -> {
-                List<VersionedReportingTask> currentReportingTasks = ofNullable(versionedDataflow.getReportingTasks()).orElseGet(ArrayList::new);
+                List<VersionedReportingTask> currentReportingTasks = new ArrayList<>(versionedDataflow.getReportingTasks());
                 currentReportingTasks.add(provenanceReportingTask);
                 versionedDataflow.setReportingTasks(currentReportingTasks);
             });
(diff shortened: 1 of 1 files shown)


minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
```diff
@@ -90,6 +91,9 @@ public byte[] enrichFlow(byte[] flowCandidate) {
         }
 
         VersionedDataflow versionedDataflow = parseVersionedDataflow(flowCandidate);
+        versionedDataflow.setReportingTasks(ofNullable(versionedDataflow.getReportingTasks()).orElseGet(ArrayList::new));
+        versionedDataflow.setRegistries(ofNullable(versionedDataflow.getRegistries()).orElseGet(ArrayList::new));
+        versionedDataflow.setControllerServices(ofNullable(versionedDataflow.getControllerServices()).orElseGet(ArrayList::new));
```
bejancsaba, line 96: Do we need to do this for the others (e.g.: parameters) as well or only these are needed to be initialised on case of an empty flow

minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
```diff
@@ -103,10 +107,12 @@ public byte[] enrichFlow(byte[] flowCandidate) {
             rootGroup.setInstanceIdentifier(randomUUID().toString());
         }
 
+        rootGroup.getControllerServices().forEach(cs -> cs.setScheduledState(ENABLED));
```
briansolo1985, line 110: Very minor, but in my opinion lamdbas are more expresseive if meaningiful parameter names are used, in this case I would use controllerService, but leave it you.

bejancsaba, line 110: This seems unrelated to the startup issue, is this specific to NiFi 2.0?

bejancsaba (commented):
Looks good thanks for the change I had 2 quick questions.

minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
```diff
@@ -103,10 +107,12 @@ public byte[] enrichFlow(byte[] flowCandidate) {
...
+
         Optional<VersionedControllerService> commonSslControllerService = createCommonSslControllerService();
         commonSslControllerService
             .ifPresent(sslControllerService -> {
-                List<VersionedControllerService> currentControllerServices = ofNullable(versionedDataflow.getControllerServices()).orElseGet(ArrayList::new);
+                List<VersionedControllerService> currentControllerServices = new ArrayList<>(versionedDataflow.getControllerServices());
```
briansolo1985, line 115: Since the initialization was moved outside now it is guaranteed that we will get a non-null list object here.
Is it necessary to recreate the list object for some reasons?
If not we could simplify the ifPresent block by reducing it to 
```
commonSslControllerService.ifPresent(versionedDataflow.getControllerServices()::add);
```

minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
```diff
@@ -118,7 +124,7 @@ public byte[] enrichFlow(byte[] flowCandidate) {
 
         createProvenanceReportingTask(commonSslControllerService.map(VersionedComponent::getInstanceIdentifier).orElse(EMPTY))
             .ifPresent(provenanceReportingTask -> {
-                List<VersionedReportingTask> currentReportingTasks = ofNullable(versionedDataflow.getReportingTasks()).orElseGet(ArrayList::new);
+                List<VersionedReportingTask> currentReportingTasks = new ArrayList<>(versionedDataflow.getReportingTasks());
```
briansolo1985, line 127: And this one also could be simplified
```
createProvenanceReportingTask(commonSslControllerService.map(VersionedComponent::getInstanceIdentifier).orElse(EMPTY))
            .ifPresent(versionedDataflow.getReportingTasks()::add);
```

minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
```diff
@@ -26,6 +26,7 @@
 import static java.util.UUID.randomUUID;
 import static java.util.stream.Collectors.toMap;
 import static org.apache.commons.lang3.StringUtils.EMPTY;
+import static org.apache.nifi.flow.ScheduledState.ENABLED;
```
briansolo1985, line 29: Could we use the static import in line 159 as well?

briansolo1985 (commented):
Hi Ferenc, thank you for this change. The approach looks good, and will make this part of the code more fault tolerant.
Wrote some  comments please take a look when you'll have the time.

minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
ferencerdei: For ParameterProviders and FlowAnalysisRules it is handled in the VersionedFlowSynchronizer

minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
ferencerdei: This also caused issues because the controller services were not enabled automatically so I got validation errors. I forgot to mention this in the ticket. Because it makes no sense to disable separate components in the context of MiNiFi, I think it's safe to always enable everything.

minifi/minifi-commons/minifi-commons-framework/src/main/java/org/apache/nifi/minifi/commons/service/FlowEnrichService.java
ferencerdei: I wanted to avoid passing the reference for the original list, but maybe you are right. As this object doesn't exist outside of this class we can safely use the value directly. I'm going to update this and the other as well.

bejancsaba:
Thanks for the details and the changes will merge as soon as the workflows conclude. +1 from my side

PR #8003, opened 2023-11-10
NIFI-12296 Write the data to Apache Doris in a StreamLoad fashion

wanghangyu817:
# Summary

[NIFI-12296](https://issues.apache.org/jira/browse/NIFI-12296)

# Tracking

### Issue Tracking

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-12296`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-12296`

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for resubmitting the pull request and cleaning up the initial set of commits @wanghangyu817.

There are a number of design and convention issue to address in order for this to go forward. At a high level, use of the Apache HttpClient library should be removed and replaced with use of the `nifi-web-client-service-api` and `nifi-web-client-api` interfaces. The `WebClientService` interface provides an abstraction for generalized HTTP operations, and it also allows standard HTTP settings like TLS and timeouts to be configured through the `StandardWebClientServiceProvider` implementation.

With moving to that interface, it seems like the `DorisClientService` may not be necessary. However, it may still be useful to abstract Doris connection settings like username and password, so it could be retained. However, the interface definition needs to be revisited. In particular, the `testConnectivity` method does not seem necessary, as that should be the responsibility of the service implementation.

At another level, making this Processor specific to JSON might be reasonable, but it would be much better to support the `RecordReader` Controller Service interface, which allows reading in more generic data in a variety of formats. As Doris appears to need JSON for the actual HTTP communication, the Processor would translate the generic NiFi Record into JSON.

There are some other naming and usage conventions, but those can be flagged and addressed after the high-level issues.

I realize the WebClientService and Record abstractions are specific to NiFi, so it may take some work to refactor, but the end result will be much more usable in the context of NiFi flows. There are a number of other components that make use of these features, so you should be able to find some examples.

If you are willing to work through a few rounds of feedback and evaluate some design questions along the way, it will take some time, but this could be integrated. On the other hand, in some cases other projects have maintained their own NiFi extensions in their own repositories, so that might be another approach to consider.

wanghangyu817:
@exceptionfactory Ok, let me sort out the rest of the work:

1. Replace Apache HttpClient with nifi-web-client-service-api and nifi-web-client-api

2.DorisClientService refactoring

3.testConnectivity removed

4. Use RecordReader support or change to PutDorisJSON?

5. Naming conventions overhaul

If there are other additions, please provide them

exceptionfactory:
You're welcome @wanghangyu817. Yes, that is a good summary for now. That should cover the high-level recommendations, and then we could evaluate more of the implementation details.

wanghangyu817:
This will require some refactoring, but I'll keep the details up to date

exceptionfactory:
@wanghangyu817 Are you still planning on revisiting this pull request? If you do not expect to get to it for some time, it would be best to close it for now and open it again when you are ready.

sdhzwc:
> This will require some refactoring, but I'll keep the details up to date

大佬 还会继续更新吗

PR #8011, opened 2023-11-12
NIFI-4550 Detect charset for FlowFile with MIME type text/*

EndzeitBegins:
# Summary

Identifies `charset`as part of `IdentifyMimeType` processor, when the detected MIME type is `text/*`.

[NIFI-4550](https://issues.apache.org/jira/browse/NIFI-4550)
[NIFI-1874](https://issues.apache.org/jira/browse/NIFI-1874)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for working on this improvement @EndzeitBegins.

It looks like reading the character set information requires reading the input stream again, which presents some efficiency concerns. It should be possible to reset the same stream used for reading the MIME Type, and then pass it to the Encoding Detector. Have you evaluated that approach?

EndzeitBegins:
Thank you for taking a look at this @exceptionfactory, I highly appreciate it.

I've made some changes regarding the stream re-use as you've pointed out.
I replaced the use of `ProcessSession.read` that receives a `InputStreamCallback` with the version that returns the `InputStream` instead, to be able to use final fields instead of AtomicReferences.     
While doing so, I've noticed that some code paths were impossible to reach (especially the mimeType being null) and removed them. This further streamlines the processors implementation.

Please let me know what you think about these changes.

EndzeitBegins:
@exceptionfactory Is there any chance to include this into 1.24.0, now that the RC1 is called off? This way, we wouldn't need to wait for the following release which will be some weeks in the future, I assume.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
```diff
@@ -233,53 +239,62 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
+            mediaTypeString = mediaType.getBaseType().toString();
+            extension = lookupExtension(mediaTypeString, logger);
+            charset = identifyCharset(tikaStream, metadata, mediaType);
+        } catch (IOException e) {
+            throw new ProcessException("IOException thrown identifying mime-type of FlowFile content", e);
+        }
+
+        flowFile = session.putAttribute(flowFile, CoreAttributes.MIME_TYPE.key(), mediaTypeString);
+        flowFile = session.putAttribute(flowFile, "mime.extension", extension);
+        flowFile = session.putAttribute(flowFile, "mime.charset", charset == null ? "" : charset.name());
```
exceptionfactory, line 266: It seems better to avoid setting this attribute at all, as opposed to setting an empty value when the detected character set is null.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
```diff
@@ -233,53 +239,62 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
             }
-        });
 
-        String mimeType = mimeTypeRef.get();
+            final MediaType mediaType = detector.detect(tikaStream, metadata);
+            mediaTypeString = mediaType.getBaseType().toString();
+            extension = lookupExtension(mediaTypeString, logger);
+            charset = identifyCharset(tikaStream, metadata, mediaType);
+        } catch (IOException e) {
+            throw new ProcessException("IOException thrown identifying mime-type of FlowFile content", e);
```
exceptionfactory, line 261: It is not necessary to repeat `IOException` in the message since that will be part of the stack trace. Recommend the following:
```suggestion
            throw new ProcessException("Failed to identify MIME type from content stream", e);
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
```diff
@@ -89,14 +89,17 @@
...
-@WritesAttribute(attribute = "mime.extension", description = "This Processor sets the FlowFile's mime.extension attribute to the file "
-        + "extension associated with the detected MIME Type. "
-        + "If there is no correlated extension, the attribute's value will be empty")
+        @WritesAttribute(attribute = "mime.type", description = "This Processor sets the FlowFile's mime.type attribute to the detected MIME Type. "
+                + "If unable to detect the MIME Type, the attribute's value will be set to application/octet-stream"),
+        @WritesAttribute(attribute = "mime.extension", description = "This Processor sets the FlowFile's mime.extension attribute to the file "
+                + "extension associated with the detected MIME Type. "
+                + "If there is no correlated extension, the attribute's value will be empty"),
+        @WritesAttribute(attribute = "mime.charset", description = "This Processor sets the FlowFile's mime.charset attribute to the detected charset. "
+                + "If unable to detect the charset or the detected MIME type is not of type text/*, the attribute's value will be empty")
```
exceptionfactory, line 102: As noted below, it seems better to avoid setting the attribute instead of setting an empty value for null.

EndzeitBegins: Thank you for the feedback. 
I adjusted the implementation and documentation. In the original commit, I followed the implementation of `mime.extension`, which is set to an empty string, when it cannot be determined. Should its implementation as is or don't be set like `mime.charset` as well?

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
EndzeitBegins: I adjusted the message as suggested.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
EndzeitBegins: I added a commit, no longer setting the attribute at all, when the charset cannot be detected.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
exceptionfactory: Thanks for making the change to the new `mime.charset` approach. I think the existing approach for `mime.extension` should be left unchanged to avoid unexpected changes for existing flows. Since this is a new attribute that only applies for text types, it seems better to set it only when it is not null.

EndzeitBegins: Perfect. Yes, that makes sense. Thank you for your quick responses.

exceptionfactory (commented):
Thanks for the updates @EndzeitBegins. I noticed that the new Tika module includes `juniversalcharset` as a transitive dependency, which is licensed under the Mozilla Public License 1.1. This library is also used in the `nifi-media-nar`, so it is already covered at the `nifi-assembly` level, but it needed to be noted in the `nifi-standard-nar`, so I pushed that change. I should be able to finish the review soon.

It is important to note, however, that this will apply to the main branch only, because of other recent changes to IdentifyMimeType. After this is merged, you could consider opening a new issue to backport the changes to the `support/nifi-1.x`, but it will require some adjustments.

exceptionfactory (approved):
Thanks again for your work on this improvement @EndzeitBegins, the latest version looks good! +1 merging to the main branch as mentioned previously.

EndzeitBegins:
Thank you for your reviews and the fast merge @exceptionfactory.

I'll take a look at what's required for the backport in the next days presumably and open a new pull-request if the work required is manageable.

EndzeitBegins:
The backport was more straightforward than I've expected. I've opened #8031.
I'd highly appreciate if you'd take a look at it @exceptionfactory, as you're familiar with this original pull-request already.

PR #8023, opened 2023-11-14
NIFI-12350 Implement Azure Active Directory authentication option whe…

vuong2407:
…n using Apache Kafka in Azure Event Hubs

# Summary

[NIFI-12350](https://issues.apache.org/jira/browse/NIFI-12350)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-nar-bundles/nifi-kafka-bundle/nifi-kafka-shared/src/main/java/org/apache/nifi/kafka/shared/aad/AADAuthenticateCallbackHandler.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+public class AADAuthenticateCallbackHandler implements AuthenticateCallbackHandler {
+
+    public static String authority;
+    public static String appId;
+    public static String appSecret;
+    public static String bootstrapServer;
```
exceptionfactory, line 30: Setting static variables is not suitable for multiple instances of this Processor, although the way the Kafka library instantiates the class makes it difficult to configure. Some further design consideration will be necessary.

exceptionfactory (changes requested):
Thanks for the contribution @vuong2407. This feature for Azure AD authentication looks useful in general, and aligns with optional supports for AWS MSK. However, the initial implementation raises several concerns about shared configuration properties. Work is in progress to redesign the Kafka integration strategy using an extensible Controller Service to handle authentication, and that may provide a better foundation for this kind of feature. For this pull request, some further consideration is necessary to determine whether it can fit in the existing structure given the dependencies required.

exceptionfactory (changes requested):
Thanks for your work on this improvement @vuong2407.

On further review of these changes, it would be best to evaluate an alternative that is not specific to the Azure EventHubs Kafka environment. The `OAUTHBEARER` authentication strategy is generic, as described in [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876), although some implementations may have certain particular approaches. The proposed changes would add a direct dependency on `msal4j`, which would not apply to many Kafka client deployments.

One option is to revisit this capability after completion of new Kafka components, described in [NIFI-11259](https://issues.apache.org/jira/browse/NIFI-11259). There is no precise timeline for that implementation, so the other option is to evaluate an alternative approach that avoids vendor-specific dependencies. That may not be achievable without introducing some new Controller Service extension interface, which again points back to the redesign. As it stands, however, introducing these substantive changes to the current components does not appear to be the best way forward in terms of future maintainability.

exceptionfactory:
Thanks again for the work on this change @vuong2407.

In light of the current conflicts, and the work in progress on refactored Kafka Processors for [NIFI-11259](https://issues.apache.org/jira/browse/NIFI-11259) now in review, it would be better to revisit this implementation after the refactored Processors have been reviewed and merged. That should provide a much better baseline for integrating alternative authentication strategies.

PR #8028, opened 2023-11-15
NIFI-12372 MiNiFi C2 Encrypt Flow Configuration Properties when Transferring

briansolo1985:
# Summary

[NIFI-12372](https://issues.apache.org/jira/browse/NIFI-12372)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/main/java/org/apache/nifi/minifi/c2/command/DefaultUpdateConfigurationStrategy.java
+++ b/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/main/java/org/apache/nifi/minifi/c2/command/DefaultUpdateConfigurationStrategy.java
@@ -47,9 +47,12 @@ import org.apache.nifi.controller.ComponentNode;
 import org.apache.nifi.controller.FlowController;
 import org.apache.nifi.controller.ProcessorNode;
 import org.apache.nifi.controller.flow.FlowManager;
+import org.apache.nifi.controller.flow.VersionedDataflow;
 import org.apache.nifi.groups.ProcessGroup;
 import org.apache.nifi.groups.RemoteProcessGroup;
 import org.apache.nifi.minifi.commons.service.FlowEnrichService;
+import org.apache.nifi.minifi.commons.service.FlowPropertyEncryptor;
+import org.apache.nifi.minifi.commons.service.FlowSerDeService;
 import org.apache.nifi.services.FlowService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -66,15 +69,20 @@ public class DefaultUpdateConfigurationStrategy implements UpdateConfigurationSt
     private final FlowController flowController;
     private final FlowService flowService;
     private final FlowEnrichService flowEnrichService;
+    private final FlowPropertyEncryptor flowPropertyEncryptor;
+    private final FlowSerDeService flowSerDeService;
     private final Path flowConfigurationFile;
     private final Path backupFlowConfigurationFile;
     private final Path rawFlowConfigurationFile;
     private final Path backupRawFlowConfigurationFile;
 
-    public DefaultUpdateConfigurationStrategy(FlowController flowController, FlowService flowService, FlowEnrichService flowEnrichService, String flowConfigurationFile) {
+    public DefaultUpdateConfigurationStrategy(FlowController flowController, FlowService flowService, FlowEnrichService flowEnrichService,
+                                              FlowPropertyEncryptor flowPropertyEncryptor, FlowSerDeService flowSerDeService, String flowConfigurationFile) {
         this.flowController = flowController;
         this.flowService = flowService;
         this.flowEnrichService = flowEnrichService;
+        this.flowPropertyEncryptor = flowPropertyEncryptor;
+        this.flowSerDeService = flowSerDeService;
         Path flowConfigurationFilePath = Path.of(flowConfigurationFile).toAbsolutePath();
         this.flowConfigurationFile = flowConfigurationFilePath;
         this.backupFlowConfigurationFile = Path.of(flowConfigurationFilePath + BACKUP_EXTENSION);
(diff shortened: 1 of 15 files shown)


briansolo1985:
A flow incoming via C2 may contain unencrypted sensitive properties as the agents can have there our encryption key.
The goal is to check the properties in the incoming flow, and if there are unencrypted sensitive properties there, use the configured encryption key to encrypt them before persisting the flow.json.raw to disk.
The logic first tries to utilise the property descriptors sent with the flow to detect the sensitive property names. However the property descriptors can omitted. In this case the logic uses the runtime manifest to get the list of sensitive properties.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/main/java/org/apache/nifi/minifi/c2/command/FlowPropertyEncryptor.java
```diff
@@ -0,0 +1,167 @@
...
+import org.apache.nifi.c2.protocol.component.api.RuntimeManifest;
+import org.apache.nifi.controller.flow.VersionedDataflow;
+import org.apache.nifi.encrypt.PropertyEncryptor;
+import org.apache.nifi.flow.VersionedConfigurableExtension;
+import org.apache.nifi.flow.VersionedParameter;
+import org.apache.nifi.flow.VersionedProcessGroup;
+import org.apache.nifi.flow.VersionedPropertyDescriptor;
+import org.apache.nifi.minifi.commons.service.FlowSerDeService;
+
+public class FlowPropertyEncryptor {
```
exceptionfactory, line 47: I recommend defining an interface with a Standard implementation.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/test/java/org/apache/nifi/minifi/c2/command/FlowPropertyEncryptorTest.java
```diff
@@ -0,0 +1,338 @@
...
+    @BeforeEach
+    public void setup() {
+        when(mockPropertyEncryptor.encrypt(anyString())).thenReturn(randomAlphabetic(5));
+        flowSerDeService = FlowSerDeService.defaultInstance();
+        testEncryptor = new FlowPropertyEncryptor(mockPropertyEncryptor, flowSerDeService, mockRunTimeManifest);
+    }
+
+    @Test
+    public void shouldEncryptParameterContextsSensitiveVariables() {
+        // given
```
exceptionfactory, line 124: The `given/when/then` comments should be removed as they are not standard across the repository and do not necessarily add much to the test definition.

exceptionfactory (commented):
Thanks for working on this improvement @briansolo1985.

The general approach makes sense, and I will defer to others with more familiarity for closer review. The use of the standard PropertyEncryptor looks good.

I recommend defining an interface named `FlowPropertyEncryptor` with a `StandardFlowPropertyEncryptor` implementation, particularly since the test class mocks the class right now.

bejancsaba (approved):
@briansolo1985 thank you for this enhancement the code and tests are clean and straightforward. I tried it via C2 and the downloaded raw flow that is persisted contained the encrypted property on the file system while the decrypted value was present runtime. So works as expected. +1 from my side.

ferencerdei:
Thanks for the feature, this really useful. I have only one question.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/main/java/org/apache/nifi/minifi/c2/command/DefaultUpdateConfigurationStrategy.java
```diff
@@ -90,11 +93,12 @@ public boolean update(byte[] rawFlow) {
         }
 
         try {
-            byte[] enrichedFlowCandidate = flowEnrichService.enrichFlow(rawFlow);
+            byte[] propertyEncryptedRawDataFlow = flowPropertyEncryptor.encryptSensitiveProperties(rawFlow);
```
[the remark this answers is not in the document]
briansolo1985: We only need to encrypt the raw flow.json. Sensitive properties in the enriched flow are automatically encrypted when loaded by flowService.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/test/java/org/apache/nifi/minifi/c2/command/FlowPropertyEncryptorTest.java
briansolo1985: Removed the comments

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/main/java/org/apache/nifi/minifi/c2/command/FlowPropertyEncryptor.java
briansolo1985: Introduced the interface

briansolo1985:
Thanks your all the reviews. Addressed the comments, and pushed the most recent changes. Please re-review.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/test/java/org/apache/nifi/minifi/c2/command/FlowPropertyEncryptorTest.java
```diff
@@ -0,0 +1,329 @@
...
+import org.apache.nifi.flow.VersionedPropertyDescriptor;
+import org.apache.nifi.minifi.commons.service.FlowSerDeService;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class FlowPropertyEncryptorTest {
```
exceptionfactory, line 61: This class should be renamed to `StandardFlowPropertyEncryptorTest` to match the new implementation class name.

exceptionfactory (commented):
Thanks for defining the interface @briansolo1985, that is helpful.

The interface definition itself is very generic, and presumes that the implementation handles parsing and serializing the JSON. Are there any performance concerns related to flow size? It looks like the enrichment service expects the flow as a byte array, but should this change also prompt further adjustments to avoid multiple rounds for conversion from byte array to JSON objects and back again?

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/test/java/org/apache/nifi/minifi/c2/command/FlowPropertyEncryptorTest.java
briansolo1985: Thanks for the catch, renamed.

briansolo1985:
I addressed the second round of comments. Decided to make a bigger refactor: moved services to minifi-commons, created interfaces for those services to become consistent. Also removed the flow serialization and deserialization from those particular services. Performance was not an issue there as that code was not on a critical path, eg flow updates are expected maximum a few times a day, but most frequently we talk about weeks or months here. However having the serde in the services hurt SRP at some level, so moved the logic to an upper layer. Also now the flow serde logic is encapsulated in one class.
Rebased onto the latest main. Tested a flow with C2, unecrypted properties were encrypted correctly.
Please review the latest changes.

ferencerdei (approved):
Thanks for the updates. +1 from my side.
I'm happy to merge if there is no other objection.

PR #8042, opened 2023-11-16
NIFI-12382: Add DatabaseSchemaRegistry service

mattyb149:
# Summary

[NIFI-12382](https://issues.apache.org/jira/browse/NIFI-12382) This PR adds a DatabaseSchemaRegistry controller service which gets the metadata for a specified database table and creates a RecordSchema from it.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for proposing this new feature @mattyb149.

The concept of a database-backed Schema Registry makes sense, but this initial implementation appears to be very specific in terms of table definition, and does not include any documentation on those expectations. With that background, it seems like this could make supporting such a component potentially difficult with concerns around the potential need for schema migration.

For these reasons, it seems like a generalized implementation needs to more configuration options and documentation. Even then, some table definition decisions raise other questions, like `IS_NULLABLE` as a string versus a tiny integer or something similar. Perhaps it would be helpful to address some of these design considerations before going further.

mattyb149:
I'll add more documentation around the use case and what this service does, but I didn't see any use for a DB-backed registry that just stores schema text (such as a String column containing an Avro schema) since we have other implementations that basically do that. I can rename the service DatabaseTableSchemaRegistry if that helps make it clearer what this service is for, but the idea is to generate the schema from metadata rather than store and retrieve schemas. This can be used by things like ValidateRecord in order to ensure that records will successfully go into a database with a downstream PutDatabaseRecord for example.

exceptionfactory:
Thanks for the reply @mattyb149. Renaming the class could be helpful.

Another potential concern is using the schema name as the table name. Aside from schema naming conventions potentially conflicting with database table naming requirements, this would also require a table for every schema, which seems like it could be difficult to maintain. Although this could make sense in some specific environments, it seems like this might not be a good fit for inclusion as a component for general usage.

pvillard31:
I'm jumping in the conversation here but I don't think this is a very specific use case. The idea is that, in many cases, data would be sent into a database by NiFi and some users may not want to deal with schemas in NiFi or introduce an external Schema Registry. The idea here is to be able to retrieve the schema of the destination table in the NiFi flow so that it can be leveraged instead of the Infer Schema option which comes with the limitations and risks that we know.

exceptionfactory:
Thanks for the additional input @pvillard31, that is helpful, and makes more sense now.

I realize that I was not evaluating all the implementation details. Now I see that the implementation uses the Database Metadata results, and that is what defines the information. Sorry for missing that detail earlier @mattyb149.

With that in mind, I can now see how this makes more sense as a way to use database table metadata for the schema definition. Perhaps including `Metadata` in Service class name would also help describe how this works.

With that background, are there any concerns about JDBC driver support across vendors? I would expect that more popular database vendors should work with this approach, but some may not, so that may be worth highlighting under additional details.

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-dbcp-service-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-dbcp-base</artifactId>
+            <version>2.0.0-SNAPSHOT</version>
+        </dependency>
```
exceptionfactory, line 36: Is this dependency necessary? It should only be required for DB Controller Service implementation, but I may be missing a reference.

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-kerberos-credentials-service-api</artifactId>
+            <version>2.0.0-SNAPSHOT</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-kerberos-user-service-api</artifactId>
+        </dependency>
```
exceptionfactory, line 46: Are these dependencies used?

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/src/main/java/org/apache/nifi/db/schemaregistry/DatabaseSchemaRegistry.java
```diff
@@ -0,0 +1,179 @@
...
+@Tags({"schema", "registry", "database", "table"})
+@CapabilityDescription("Provides a service for generating a record schema from a database table definition. The service is configured "
+        + "to use a table name and a database connection fetches the table metadata (i.e. table definition) such as column names, data types, "
+        + "nullability, etc.")
+public class DatabaseSchemaRegistry extends AbstractControllerService implements SchemaRegistry {
+
+    private static final Set<SchemaField> schemaFields = EnumSet.of(SchemaField.SCHEMA_NAME);
+
+    static final PropertyDescriptor DBCP_SERVICE = new PropertyDescriptor.Builder()
+            .name("dbcp-service")
```
exceptionfactory, line 59: As a new Service, this is a good opportunity to use the same value for the name and display name.
```suggestion
            .name("Database Connection Pooling Service")
```

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/src/main/java/org/apache/nifi/db/schemaregistry/DatabaseSchemaRegistry.java
```diff
@@ -0,0 +1,179 @@
...
+                        recordFields.add(new RecordField(
+                                columnName,
+                                DataTypeUtils.getDataTypeFromSQLTypeValue(dataType),
+                                defaultValue,
+                                isNullable));
+                    }
+
+                    // If no columns are found, check that the table exists
+                    if (recordFields.isEmpty()) {
+                        try (final ResultSet tblrs = dmd.getTables(dbCatalogName, dbSchemaName, tableName, null)) {
```
exceptionfactory, line 151: Recommend adjusting the variable name for readability:
```suggestion
                        try (final ResultSet tablesResultSet = dmd.getTables(dbCatalogName, dbSchemaName, tableName, null)) {
```

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/src/main/java/org/apache/nifi/db/schemaregistry/DatabaseSchemaRegistry.java
```diff
@@ -0,0 +1,179 @@
...
+
+                            if (!tblrs.next()) {
+                                throw new SchemaNotFoundException("Table "
+                                        + String.join(".", qualifiedNameSegments)
+                                        + " not found");
+                            } else {
+                                getLogger().warn("Table "
+                                        + String.join(".", qualifiedNameSegments)
+                                        + " found but no columns were found, if this is not expected then check the user permissions for getting table metadata from the database");
+                            }
```
exceptionfactory, line 169: To avoid deep levels of code nesting, what do you think about breaking this out to a separate method?

exceptionfactory (commented):
Thanks for adding the capability description @mattyb149, that definitely helps clarify how things work now. I noted a couple more detail recommendations.

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/pom.xml
mattyb149: It's only for the common properties. If there's a better spot to move those to I can refactor it as part of this.

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/pom.xml
mattyb149: I got an error when running tests, but it might be because of including nifi-dbcp-base. If I can refactor out the common properties I probably don't need these either.

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/src/main/java/org/apache/nifi/db/schemaregistry/DatabaseSchemaRegistry.java
mattyb149: Is that the new convention? We used to do the opposite when requiring both `name()` and `displayName()` to be set, `name` was a more machine-friendly name for things like internationalization and `displayName` was the display name in English.

exceptionfactory: Yes, it is newer with part of the changes for migrating properties in NiFi 2.0. I know there have been past discussions about internationalization, but unfortunately it appears there was never any significant progress in that direction. For now, the best approach seems to be using the same name in both places, and if there is ever momentum around internationalization, then we could revisit a migration strategy. Unfortunately the disconnect between property and display names right now makes it more difficult to translate from a UI to programmatic approach, so keeping the same names will help under the current circumstances.

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/pom.xml
exceptionfactory: Thanks for clarifying, defining the properties for the test class looks good.

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/src/test/java/org/apache/nifi/db/schemaregistry/DatabaseTableSchemaRegistryTest.java
```diff
@@ -0,0 +1,224 @@
...
+        for (String createTableStatement : CREATE_TABLE_STATEMENTS) {
+            stmt.execute(createTableStatement);
+        }
+
+        stmt.close();
+        conn.close();
+    }
+
+    @AfterAll
+    public static void shutdownDatabase() throws Exception {
```
exceptionfactory, line 128: The `Exception` is not thrown and can be removed:
```suggestion
    public static void shutdownDatabase() {
```

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/src/test/java/org/apache/nifi/db/schemaregistry/DatabaseTableSchemaRegistryTest.java
```diff
@@ -0,0 +1,224 @@
...
+        runner.setProperty(dbSchemaRegistry, DatabaseTableSchemaRegistry.DBCP_SERVICE, SERVICE_ID);
+        runner.setProperty(dbSchemaRegistry, DatabaseTableSchemaRegistry.SCHEMA_NAME, "SCHEMA1");
+        runner.enableControllerService(dbSchemaRegistry);
+        SchemaIdentifier schemaIdentifier = new StandardSchemaIdentifier.Builder()
+                .name("NOT_A_TABLE")
+                .build();
+        assertThrows(SchemaNotFoundException.class, () -> dbSchemaRegistry.retrieveSchema(schemaIdentifier));
+    }
+
+    private static class DBCPServiceSimpleImpl extends AbstractControllerService implements DBCPService {
```
exceptionfactory, line 200: Recommend avoiding the use of `Impl` on class names, even for testing.
```suggestion
    private static class SimpleDBCPService extends AbstractControllerService implements DBCPService {
```

nifi-nar-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-db-schema-registry-service/src/test/java/org/apache/nifi/db/schemaregistry/DatabaseTableSchemaRegistryTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+        @Override
+        public String getIdentifier() {
+            return this.getClass().getName();
+        }
+
+        @Override
+        public Connection getConnection() throws ProcessException {
+            try {
+                Class.forName("org.apache.derby.jdbc.EmbeddedDriver");
```
exceptionfactory, line 216: Recommend declaring this class name once and reusing throughout the test class.

exceptionfactory (changes requested):
Thanks for making the code adjustments @mattyb149. The latest version works well in runtime testing with MySQL.

I noted a couple minor convention recommendations. The only concern remaining is the location of the class itself. Although it is convenient to have it with the Connection Pools, it seems out of place. As mentioned in the detail comments, moving it to a separate NAR seems like a better long-term solution, with minimal impact on the size. Thoughts?

exceptionfactory (approved):
Thanks for restructuring the bundle @mattyb149, the latest version looks good! +1 merging

PR #8071, opened 2023-11-28
NIFI-12420: Add Sawmill transformation processors

mattyb149:
# Summary

[NIFI-12420](https://issues.apache.org/jira/browse/NIFI-12420) This PR offers the SawmillTransformJSON and SawmillTransformRecord processors, using the Sawmill library to transform input.

Couple of questions for the reviewer(s):
- The NAR is 9.3 MB, should this be a profile disabled by default?
- The SawmillTransformJSON processor is faster for JSON input/output, but is it useful enough to include?
- There is a Guava override to eliminate vulnerabilities, seems to work fine but would like to confirm?

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
+++ b/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-resources/src/main/resources/conf/logback.xml
@@ -90,6 +90,9 @@
     <!-- Suppress non-error messages due to known warning about redundant path annotation (NIFI-574) -->
     <logger name="com.sun.jersey.spi.inject.Errors" level="ERROR"/>
 
+    <!-- Suppress non-error messages due to Sawmill ProcessorFactoriesLoader emitting a large amount of WARNS. -->
+    <logger name="io.logz.sawmill.ProcessorFactoriesLoader" level="ERROR"/>
+
     <!--
         Logger for capturing Bootstrap logs and MiNiFi's standard error and standard out.
     -->
--- a/nifi-assembly/pom.xml
+++ b/nifi-assembly/pom.xml
@@ -906,6 +906,12 @@ language governing permissions and limitations under the License. -->
             <version>2.0.0-SNAPSHOT</version>
             <type>nar</type>
         </dependency>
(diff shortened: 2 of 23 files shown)


exceptionfactory (changes requested):
Thanks for proposing this new Processor @mattyb149.

The Sawmill library includes a number of additional dependencies, and it seems like many of them have not been updated recently. This raises some general concerns from a maintainability point of view, as it impacts both the size of the NAR as well as the security posture of the distribution.

The Guava library is a good example, although there should not be any compatibility concerns between 27 at 32. There are a number of other transitive dependencies that do not appear to be updated in Sawmill directly, so every dependency should be evaluated for a newer version and overridden if possible. Some dependencies appear to be specific to certain pipeline operations, so one potential strategy could be excluding specific dependencies with a note that those particular pipeline operations are not supported.

The composable nature of Sawmill pipelines looks like it has some advantages, but it also raises questions about specific features, such as the Geo IP database.

With that background, I recommend a more detailed review of all transitive dependencies before moving forward. I need to go back through several other pull requests, but I could also take a closer look soon.

EndzeitBegins:
If I remember correctly, there have been issues with the size of the default distribution of NiFi in the past. 

As the Sawmill nar seems to include a fair amount of dependencies, I advocate for a separate nar, not included in the default distribution, in case the processors should be merged into the codebase.

joewitt:
I'd avoid adding to default build for now.

I also think the dependencies need further review from a L&N point of view.  The L&N is updated but seems pretty light compared to the deps involved here.  Are we sure we checked?

exceptionfactory:
As mentioned earlier, I'm hesitant on bringing in this component and bundle, even optionally, given the number of outdated dependencies. These findings still show up in reviews, even if not included in standard convenience binaries. It may be possible to address the old dependencies through overrides, but that is something that should be evaluated first, before going forward. If we can override old dependencies, then it seems like we would be in a better place to consider bringing this into the repository as an optional module.

mattyb149:
Closing this as it isn't a good fit as a NiFi processor.

PR #8094, opened 2023-11-30
NIFI-12270: Add new provenance event: UPLOAD

Lehel44:
# Summary

[NIFI-12270](https://issues.apache.org/jira/browse/NIFI-12270)
[See discussion](https://lists.apache.org/thread/hwt87hz4679vbmtv9yg78dw38jrwc861)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


joewitt:
Similar to Matt's comment why does a SEND event not cover this case sufficiently?

In this case it is that NiFi is effectively being used to take a given external reference and send that data to another system all without having stored locally (within nifi) the bytes.  Never the less NiFi is the system that sends the data it seems and that is the purpose of the SEND event.  A send without having had a CREATE/RECEIVE is fair game now given this paradigm...  that is ok.

I think we should avoid adding an event type unless the need/case is really strong.

pvillard31:
This PR is following @markap14 's recommendation in this email thread:
https://lists.apache.org/thread/hwt87hz4679vbmtv9yg78dw38jrwc861

joewitt:
Ahh that discussion context is helpful - would make sense to have that already linked in either the JIRA itself or this PR.  If it was there and I missed it sorry about that.  Otherwise, we need to put links to such good/useful discussions on these to help others who might review later.

To the issue at hand, I do not share the view we should create a different event type here.  The meaning of 'SEND' is that NiFi is responsible for having taken the bits of a thing and sent it to some external system.  That is precisely what is happening here.  Now there is value in knowing as well then that NiFi itself never actually stored the bits.  This is known though already from the provenance trail of this object.  Presumably it never had a CREATE/RECEIVE and the bytes will of course have never been populated other than perhaps a link/reference.

Given there was already a decent discussion and some apparent consensus and clearly this was a lot of work I am not saying I am a -1 on this but it does add more code/more maintenance/more semantics and I'm not clear what the value add will be in that sense.  If the issue of size is being worked around here then ideally we'd have a different/simpler answer.  

If we continue with this as-is the description of a SEND should ensure devs use that when sending content from within NiFi to an external system and an upload only when sending external content to an external system.  Because as written in teh docs/guide right now UPLOAD looks like a subcase of SEND which it is not.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
```diff
@@ -468,6 +468,11 @@
             <groupId>org.glassfish.jaxb</groupId>
             <artifactId>jaxb-runtime</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-data-provenance-utils</artifactId>
+            <version>2.0.0-SNAPSHOT</version>
+        </dependency>
```
markap14, line 475: Processors should not need to have a dependency on nifi-data-provenance-utils. The API should include anything necessary in order to emit the provenance event.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/repository/StandardProvenanceReporter.java
```diff
@@ -119,7 +122,7 @@ public void receiveMigration(final Set<ProvenanceEventRecord> events) {
      * be created by the session itself, as well as by the Processor
      *
      * @param parents parents
-     * @param child child
+     * @param child   child
```
markap14, line 125: This whitespace shouldn't be added

nifi-commons/nifi-data-provenance-utils/src/main/java/org/apache/nifi/provenance/upload/StandardUploadContext.java
```diff
@@ -0,0 +1,83 @@
...
+            return this;
+        }
+
+        public StandardUploadContext build() {
+            StandardUploadContext event = new StandardUploadContext();
+            event.transitUri = transitUri;
+            event.details = details;
+            event.transmissionMillis = transmissionMillis;
+            event.force = force;
+            return event;
```
markap14, line 79: Preferred method here would be to pass the Builder to the constructor of StandardUploadContext and mark all member variables in StandardUploadContext as final

nifi-commons/nifi-data-provenance-utils/src/main/java/org/apache/nifi/provenance/upload/StandardFileResource.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public Long getSize() {
+        return size;
+    }
+
+    @Override
+    public String toString() {
+        return "location=" + path + ", " +
+                "size=" + size;
```
markap14, line 40: The toString() should include the fact that it's a StandardFileResource. E.g., `return "StandardFileResource[location=" + location + ", size=" + size + "]";`

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
markap14: We should probably eliminate this newly introduced `UploadContext`. While it does keep the `ProvenanceReporter` cleaner by avoiding introducing several potential overrides, it introduces more complexity into the Processors that might want to call the method. I.e., the overrides make the ProvenanceReporter more complex but make the Processors simpler. This is one of those situations where I could go either way. But the fact that interface already is laid out to use several overrides means it's cleaner to keep consistent. Additionally, it eliminates the need for something like this in data-provenance-utils and also avoids introducing something like this into the nifi-api

markap14 (changes requested):
Thanks for the contribution, @Lehel44! Overall, I think it's reasonable to add the new UPLOAD event but see inline comments about implementation details.

Lehel44:
@joewitt I added the discussion link to the Jira, thanks for pointing out. The main issue here was that the SEND event used the size of the incoming FlowFile but instead of a copy of this FlowFile an external file was uploaded and sent to the external system, therefore its size should have been represented instead of the FlowFile's size. From the documentation: "SEND: Emits a Provenance Event of type that indicates that a copy of the given FlowFile was sent to an external destination.". The UPLOAD event would describe an external resource being uploaded to an external destination. While the SEND event automatically uses the size of the FlowFile argument when creating the event, the UPLOAD takes a FlowFile for necessary event attributes (flowfile UUID etc.), but also takes a FileResource object which represents the external resource and uses its attributes when constructing the event: for instance the size and the path are will be displayed in the provenance event.

nifi-api/src/main/java/org/apache/nifi/provenance/upload/FileResource.java
```diff
@@ -20,15 +20,26 @@
  * Holds information of a file resource for UPLOAD
  * provenance events.
  */
-public interface FileResource {
+public class FileResource {
```
exceptionfactory, line 23: This is a breaking change, so it should be reverted.

exceptionfactory (changes requested):
Thanks for the update @Lehel44. I noticed that `FileResource` was changed from an interface to a class, but this is a breaking change at the `nifi-api` level, so it should be reverted. I also recommend only adding a minimal number of new methods to ProvenanceReporter, instead of all possible versions.

nifi-api/src/main/java/org/apache/nifi/provenance/upload/FileResource.java
Lehel44: @exceptionfactory I included the FileResource interface in this pull request as an interface, but later realized it's unnecessary as it functions as a straightforward DTO. Moreover, if I introduce the StandardFileResource implementation to nifi-data-provenance-utils (the place I thought to add it), it will create a dependency on nifi processors.

Lehel44: I can reduce the number of methods, I followed the signatures of the send method. I can check which ones are more frequently used and eliminate the others.

exceptionfactory: Thanks for the clarification, I should have reviewed the full history, that makes sense.

Yes, I recommend defining the methods required for the common use cases.

Lehel44: I excluded methods based on if they were either unused or less commonly utilized in the SEND events.

nifi-mock/src/main/java/org/apache/nifi/util/MockProvenanceReporter.java
```diff
@@ -225,6 +228,49 @@ public void send(final FlowFile flowFile, final String transitUri, final String
...
+            if (force) {
+                sharedSessionState.addProvenanceEvents(Collections.singleton(record));
+            } else {
+                events.add(record);
+            }
+        } catch (final Exception e) {
+            logger.error("Failed to generate Provenance Event due to " + e);
+            if (logger.isDebugEnabled()) {
+                logger.error("", e);
+            }
```
exceptionfactory, line 270: The `due to` convention should be avoided, I also recommend avoid the conditional debug for the stack trace. Although it is present in some places, we should move away from it in general.

Lehel44:
Thanks for the review @markap14 @exceptionfactory. Is this one good to go in?

exceptionfactory (approved):
Thanks for working through the feedback and scoping down the number of new methods @Lehel44, the latest version looks good. +1 merging

PR #8097, opened 2023-12-01
NIFI-12419: Applying a copy of the nifi-web-frontend rat config in the parent module when the new ui profile is not activated

mcgilman:
NIFI-12419:
- Applying a copy of the nifi-web-frontend rat config in the parent module when the new ui profile is not activated. Previously, if the new ui profile was not activated, the rat check would fail.


rfellows:
will review.

rfellows (approved):
LGTM 👍 Thanks, @mcgilman. Will merge after CI completes.

PR #8064, opened 2023-11-26
NIFI-12386 Adds processor FilterAttribute (Java 1.8 / NiFi 1.x)

EndzeitBegins:
# Summary

Adds a new processor `FilterAttribute` to the `nifi-standard-processors` bundle.
See [NIFI-12386](https://issues.apache.org/jira/browse/NIFI-12386).

This is backport of #8049 compatible with Java 1.8 / NiFi 1.x.

# Tracking

Please complete the following tracking steps prior to pull request creation.

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI) issue created

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-00000`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-00000`

### Pull Request Formatting

- [x] Pull Request based on current revision of the `main` branch
- [x] Pull Request refers to a feature branch with one commit containing changes

# Verification

Please indicate the verification steps performed prior to pull request creation.

### Build

- [x] Build completed using `mvn clean install -P contrib-check`
  - [x] JDK ~~21~~ 8

### Licensing

- [x] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [x] New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [x] Documentation formatting appears as expected in rendered files

PR #8098, opened 2023-12-01
NIFI-12446 Refactor FilterAttribute to align with code conventions

EndzeitBegins:
# Summary

Refactors the recently added `FilterAttribute` processor to align with common code practices followed by other processors, before it's included in a first release version.

[NIFI-12446](https://issues.apache.org/jira/browse/NIFI-12446)

Note: This targets the `support/nifi-1.x` branch. The `main` will be receive a PR once the changes have finalized and this PR is merged.  

# Tracking

### Issue Tracking

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-12446`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-12446`

### Pull Request Formatting

- [x] Pull Request based on current revision of the ~~`main`~~`support/nifi-1.x` branch

# Verification

### Build

  - [x] ~~JDK 21~~JDK 1.8

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FilterAttribute.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FilterAttribute.java
@@ -23,7 +23,7 @@ import org.apache.nifi.annotation.behavior.SupportsBatching;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.Tags;
 import org.apache.nifi.annotation.lifecycle.OnScheduled;
-import org.apache.nifi.components.AllowableValue;
+import org.apache.nifi.components.DescribedValue;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
@@ -34,7 +34,6 @@ import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


EndzeitBegins:
I've taken a first shot at refactoring the `FilterAttribute` processor.

I'd appreciate if you'd take a look at the changes @mosermw / @exceptionfactory as you both are familiar with the topic.

I've encountered that `allowableValues` actually support passing in an enum class (that implements `DescribedValue`) and that's used by a multitude of existing processors. I thought that is a good match to align the declaration and usage, instead of passing `String` values around. Let me know what you think about that.

Also I noticed that we did a trim where that was not needed, as were trimming all values beforehand:
```Java
                .map(String::trim)
                .filter(attributeName -> !attributeName.trim().isEmpty())`
```
It seems there is no `not()` function available (at least for Java 1.8). I think the expression without the trim is now readable enough to: `!attributeName.isEmpty()`. Do you agree @exceptionfactory?

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FilterAttribute.java
```diff
@@ -257,17 +215,93 @@ private static String getDelimiter() {
...
+        private final String description;
+
+        FilterMode(final String value, final String description) {
+            this.value = value;
+            this.description = description;
+        }
+
+        @Override
+        public String getValue() {
+            return this.value;
```
exceptionfactory, line 242: Instead of returning `value`, this can return `name()` so that the enum value matches the configured property value. That allows FilterMode.valueOf() to work directly with the configured property value.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FilterAttribute.java
```diff
@@ -257,17 +215,93 @@ private static String getDelimiter() {
...
-        REGEX,
+    private static <E extends Enum<E> & DescribedValue> E enumConstantFromValue(final Class<E> enumClass, final String value) {
+        for (E enumConstant : enumClass.getEnumConstants()) {
+            if (enumConstant.getValue().equals(value)) {
+                return enumConstant;
+            }
+        }
+
+        throw new IllegalArgumentException(String.format("Unknown %s value [%s]", enumClass.getSimpleName(), value));
     }
```
exceptionfactory, line 306: As mentioned in relation to the property value, switching `getValue()` to return the enum `name()` should obviated the need for this method.

exceptionfactory (changes requested):
Thanks for making the adjustments @EndzeitBegins, this looks good overall. The switch to use enum with `DescribedValue` is helpful. I noted one more recommendation about the enum definition usage, otherwise this looks good.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FilterAttribute.java
EndzeitBegins: I thought about this as well. However this prevents providing more readable names to the users of the processor. 
As far as I know best practice of naming enum entries is using all UPPERCASE. 
Thus, all properties using allowableValues based on an enum would be restricted to be in uppercase. I'm not sure if that's a favourable user experience, additionally it somewhat reveals implementation details. 

In case NIFI-12452 / #8102 gets accepted, this method would be provided "globally", reducing the complexity to implement something like this time and time again. 
Scanning through the code I found multiple places where a method similar to this is used. I was hoping NIFI-12452 could be the basis to simplify and align the implementation at these places.

What are you thoughts on this? Or do you think it's not worth "only" to support more readable names?

In case NIFI-12452 is agreed upon, I would like to backport it into the 1.x line as well and refactor this PR to base on those changes.

exceptionfactory: The changes proposed for NIFI-12452 look useful, so thanks for working on those improvements.

In either case, the question comes down to differentiating between the value itself and the display name. In this scenario, the value could match the enum, such as `RETAIN`, but the display name could be `Retain` for presentation purposes. When you refer to the user experience, are you referencing the UI, or programmatic interaction with the REST API?

EndzeitBegins: I was talking about the UI users first and foremost.

I'm somewhat confused though. Due to [this comment](https://github.com/apache/nifi/pull/8064#discussion_r1411195114) I was under the impression one should aim to use the same value for both `value` and `display name` for new implementations, where possible. 
I assumed this would be applicable to `AllowableValue`s as well. 
Or was that comment directed at `PropertyDescriptor`s exclusively?

exceptionfactory: Thanks for the reply @EndzeitBegins, and for referencing the comments in relation to the Property Names.

Although this is focusing on the property value, you are right to reference the same question as it relates to user experience in both cases. That's also a good point about the default value.

On further consideration, maintaining parity between the value and display name of the property value seems the most consistent, so your current approach looks good.

Thanks again for the thoughtful interaction on this question. With this covered, I think the changes should be ready to go, so I will follow up soon.

exceptionfactory (approved):
Thanks again for making these changes @EndzeitBegins! +1 merging

exceptionfactory:
Merged in https://github.com/apache/nifi/commit/c565e08ccca6c1fce17b1d3de49d7a160cfbee38

PR #8101, opened 2023-12-01
NIFI-11308 Adding expression language function to return nifi version

annanys23:
# Summary

[NIFI-11308](https://issues.apache.org/jira/browse/NIFI-11308)

New ExpL function added: ${nifiVersion()}

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-commons/nifi-expression-language/pom.xml
```diff
@@ -132,5 +137,10 @@
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-api</artifactId>
+            <scope>compile</scope>
+        </dependency>
```
exceptionfactory, line 144: The `nifi-framework-api` should not be a dependency of this module.

nifi-commons/nifi-expression-language/pom.xml
```diff
@@ -101,6 +101,11 @@
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-utils</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-nar-utils</artifactId>
+            <version>2.0.0-SNAPSHOT</version>
+        </dependency>
```
exceptionfactory, line 108: The `nifi-nar-utils` should not be a dependency of this module.

exceptionfactory (changes requested):
Thanks for proposing this new feature @annanys23.

The concept seems useful, but the implementation is problematic because it introduces framework-level dependencies to the expression language library, which is contrary to the general design of the system.

A simpler approach would be to configure the Maven JAR Plugin to write the standard manifest entries, and then use standard Java Package implementation version methods to retrieve the version from that expression language library itself.

Although `nifiVersion` is fairly clear, it might be better to name this `frameworkVersion` instead.

markap14:
I agree, we definitely want to avoid adding any sort of framework-specific dependencies to the nifi-expression-language module. Ideally, though, I think we should actually avoid updating the Expression Language at all. There's no need to expose a new function for this, but instead just update `NarClassLoaders.InitContext` so that when it sets the framework bundle it also sets a system property: `System.setProperty("nifi.version", frameworkBundle.getBundleDetails().getCoordinate().getVersion());`
This should make it available to the Expression Language as `${'nifi.version'}`

annanys23:
Thanks for the review and suggestions!

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-nar-utils/src/main/java/org/apache/nifi/nar/NarClassLoaders.java
```diff
@@ -74,6 +74,7 @@ private InitContext(
             this.jettyBundle = jettyBundle;
             this.serverInstance = serverInstance;
             this.bundles = bundles;
+            System.setProperty("nifi.version", frameworkBundle.getBundleDetails().getCoordinate().getVersion());
```
exceptionfactory, line 77: Recommend adding a comment indicating the reason for this addition. I would recommend naming this `framework.version` to avoid potential confusion that could result from different NAR extension versions, which are also part of NiFi, but could be different in particular installations.

annanys23: agreed. where are the properties documented?

mosermw: Could we name this system property `nifi.framework.version`?  This is more specific and may avoid confusion with generic system properties that are not related to NiFi.

annanys23: changed prop name to: nifi.framework.version and added a section to the Expression Language Guide.  Are there any other system properties that need to be documented?  I did not find any, outside of test classes.

exceptionfactory (approved):
Thanks for working through the feedback and updating the documentation @annanys23, the latest version looks good. Thanks for the recommendation on the implementation @markap14. +1 merging

PR #8105, opened 2023-12-02
NIFI-12411: Update PublishAMQP to read AMQP headers value from FlowFile attributes and `amq$headers` string

umarhussain15:
# Summary

[NIFI-12411](https://issues.apache.org/jira/browse/NIFI-12411)

Processor can now also create message headers using attributes matching the given Regex property. The name of the attribute is used as header key. In addition to that it can read values from two sources i.e. attributes matching Regex and the headers present in `amq$headers` string. In case of key duplication in both sources, the precedence property decide which value will be used.

AMQP attributes keys are now set as constants strings and reused where needed.

Added `onScheduled` to PublishAMQP and ConsumeAMQP to read all properties which don't change during running state.

Test cases added for `PublishAMQP` headers source changes.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
+++ b/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
@@ -19,7 +19,6 @@ package org.apache.nifi.amqp.processors;
 import com.rabbitmq.client.AMQP;
 import com.rabbitmq.client.AMQP.BasicProperties;
 import com.rabbitmq.client.Connection;
-import java.util.EnumSet;
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
 import org.apache.nifi.annotation.behavior.ReadsAttribute;
@@ -62,7 +61,8 @@ import java.util.regex.Pattern;
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_APPID_ATTRIBUTE, description = "The App ID field to set on the AMQP Message"),
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_CONTENT_ENCODING_ATTRIBUTE, description = "The Content Encoding to set on the AMQP Message"),
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_CONTENT_TYPE_ATTRIBUTE, description = "The Content Type to set on the AMQP Message"),
-    @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_HEADERS_ATTRIBUTE, description = "The headers to set on the AMQP Message"),
+    @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_HEADERS_ATTRIBUTE, description = "The headers to set on the AMQP Message, if 'Header Source' is set to use it. "
+        + "See additional details of the processor."),
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_DELIVERY_MODE_ATTRIBUTE, description = "The numeric indicator for the Message's Delivery Mode"),
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_PRIORITY_ATTRIBUTE, description = "The Message priority"),
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_CORRELATION_ID_ATTRIBUTE, description = "The Message's Correlation ID"),
@@ -87,7 +87,6 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             .build();
     public static final PropertyDescriptor ROUTING_KEY = new PropertyDescriptor.Builder()
             .name("Routing Key")
-            .displayName("Routing Key")
             .description("The name of the Routing Key that will be used by AMQP to route messages from the exchange to a destination queue(s). "
                     + "Usually provided by the administrator (e.g., 'myKey')In the event when messages are sent to a default exchange this property "
                     + "corresponds to a destination queue name, otherwise a binding from the Exchange to a Queue via Routing Key must be set "
@@ -98,31 +97,17 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             .build();
     public static final PropertyDescriptor HEADERS_SOURCE = new PropertyDescriptor.Builder()
             .name("Headers Source")
-            .displayName("Headers Source")
-            .description(
-                    "The source of the headers which will be put in the published message. They can come either from the processor property as a string or they can be " +
-                            "picked from flow file attributes based on Regex expression.")
+            .description("The source of the headers which will be applied to the published message.")
             .required(true)
-            .allowableValues(InputHeaderSource.getAllowedValues())
-            .defaultValue(InputHeaderSource.STRING)
-            .build();
-    public static final PropertyDescriptor HEADERS_SOURCE_PRECEDENCE = new PropertyDescriptor.Builder()
-            .name("Headers Source Precedence")
-            .displayName("Headers Source Precedence")
-            .description(
-                    "In case of key duplication in header sources, this defines which value to take.")
-            .required(true)
-            .dependsOn(HEADERS_SOURCE, InputHeaderSource.BOTH)
-            .allowableValues(InputHeaderSource.STRING, InputHeaderSource.ATTRIBUTES)
-            .defaultValue(InputHeaderSource.STRING)
+            .allowableValues(InputHeaderSource.class)
+            .defaultValue(InputHeaderSource.AMQP_HEADERS_ATTRIBUTE)
             .build();
     public static final PropertyDescriptor HEADERS_ATTRIBUTES_REGEX = new PropertyDescriptor.Builder()
             .name("Attributes To Headers Regular Expression")
-            .displayName("Attributes To Headers Regular Expression")
-            .description("Regular expression that will be evaluated against the flow file attributes to select "
-                    + "the matching attributes and put as AMQP headers.")
+            .description("Regular expression that will be evaluated against the flow file attributes to select the matching attributes and put as AMQP headers. "
+                + "Attribute name will be used as header key.")
             .required(true)
-            .dependsOn(HEADERS_SOURCE, InputHeaderSource.ATTRIBUTES, InputHeaderSource.BOTH)
+            .dependsOn(HEADERS_SOURCE, InputHeaderSource.FLOWFILE_ATTRIBUTES)
             .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
             .addValidator(StandardValidators.createRegexValidator(0, Integer.MAX_VALUE, true))
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
@@ -133,7 +118,7 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             .description("The character that is used to split key-value for headers. The value must only one character. "
                     + "Otherwise you will get an error message")
             .defaultValue(",")
-            .dependsOn(HEADERS_SOURCE, InputHeaderSource.STRING, InputHeaderSource.BOTH)
+            .dependsOn(HEADERS_SOURCE, InputHeaderSource.AMQP_HEADERS_ATTRIBUTE)
             .addValidator(StandardValidators.SINGLE_CHAR_VALIDATOR)
             .required(false)
             .build();
@@ -157,7 +142,6 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
         properties.add(EXCHANGE);
         properties.add(ROUTING_KEY);
         properties.add(HEADERS_SOURCE);
-        properties.add(HEADERS_SOURCE_PRECEDENCE);
         properties.add(HEADERS_ATTRIBUTES_REGEX);
         properties.add(HEADER_SEPARATOR);
         properties.addAll(getCommonPropertyDescriptors());
@@ -172,8 +156,8 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
      * <br>
      * <p>
      * NOTE: Attributes extracted from {@link FlowFile} are considered candidates for AMQP properties if their names are prefixed with
-     * "amq$" (e.g., amqp$contentType=text/xml). For "amqp$headers" it depends on the value of
-     * {@link PublishAMQP#HEADERS_SOURCE}, if the value is {@link InputHeaderSource#ATTRIBUTES} then message headers are created from this attribute value,
+     * "amqp$" (e.g., amqp$contentType=text/xml). For "amqp$headers" it depends on the value of
+     * {@link PublishAMQP#HEADERS_SOURCE}, if the value is {@link InputHeaderSource#FLOWFILE_ATTRIBUTES} then message headers are created from this attribute value,
      * otherwise this attribute will be ignored.
      */
     @Override
@@ -188,21 +172,17 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             throw new IllegalArgumentException("Failed to determine 'routing key' with provided value '"
                 + context.getProperty(ROUTING_KEY) + "' after evaluating it as expression against incoming FlowFile.");
         }
-        String selectedHeaderSource = context.getProperty(HEADERS_SOURCE).getValue();
-        String headerSourcePrecedence = null;
+        InputHeaderSource selectedHeaderSource = context.getProperty(HEADERS_SOURCE).asAllowableValue(InputHeaderSource.class);
         Character headerSeparator = null;
         Pattern pattern = null;
         if (context.getProperty(HEADERS_ATTRIBUTES_REGEX).isSet()) {
             pattern = Pattern.compile(context.getProperty(HEADERS_ATTRIBUTES_REGEX).evaluateAttributeExpressions().getValue());
         }
-        if (context.getProperty(HEADERS_SOURCE_PRECEDENCE).isSet()) {
-            headerSourcePrecedence = context.getProperty(HEADERS_SOURCE_PRECEDENCE).getValue();
-        }
         if (context.getProperty(HEADER_SEPARATOR).isSet()) {
             headerSeparator = context.getProperty(HEADER_SEPARATOR).getValue().charAt(0);
         }
 
-        final BasicProperties amqpProperties = extractAmqpPropertiesFromFlowFile(flowFile, selectedHeaderSource, headerSourcePrecedence, headerSeparator, pattern);
+        final BasicProperties amqpProperties = extractAmqpPropertiesFromFlowFile(flowFile, selectedHeaderSource, headerSeparator, pattern);
 
         final String exchange = context.getProperty(EXCHANGE).evaluateAttributeExpressions(flowFile).getValue();
         final byte[] messageContent = extractMessage(flowFile, session);
@@ -240,7 +220,7 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
     /**
      * Extracts contents of the {@link FlowFile} as byte array.
      */
-    private byte[] extractMessage(FlowFile flowFile, ProcessSession session) {
+    private byte[] extractMessage(final FlowFile flowFile, ProcessSession session) {
         final byte[] messageContent = new byte[(int) flowFile.getSize()];
         session.read(flowFile, in -> StreamUtils.fillBuffer(in, messageContent, true));
         return messageContent;
@@ -270,9 +250,9 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
     /**
      * Extracts AMQP properties from the {@link FlowFile} attributes. Attributes
      * extracted from {@link FlowFile} are considered candidates for AMQP
-     * properties if their names are prefixed with "amq$" (e.g., amqp$contentType=text/xml).
+     * properties if their names are prefixed with "amqp$" (e.g., amqp$contentType=text/xml).
      */
-    private BasicProperties extractAmqpPropertiesFromFlowFile(FlowFile flowFile, String selectedHeaderSource, String headerSourcePrecedence, Character separator, Pattern pattern) {
+    private BasicProperties extractAmqpPropertiesFromFlowFile(final FlowFile flowFile, final InputHeaderSource selectedHeaderSource, final Character separator, final Pattern pattern) {
         final AMQP.BasicProperties.Builder builder = new AMQP.BasicProperties.Builder();
 
         readAmqpAttribute(flowFile, AbstractAMQPProcessor.AMQP_CONTENT_TYPE_ATTRIBUTE, builder::contentType);
@@ -289,7 +269,7 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
         readAmqpAttribute(flowFile, AbstractAMQPProcessor.AMQP_APPID_ATTRIBUTE, builder::appId);
         readAmqpAttribute(flowFile, AbstractAMQPProcessor.AMQP_CLUSTER_ID_ATTRIBUTE, builder::clusterId);
 
-        Map<String, Object> headers = prepareAMQPHeaders(flowFile, selectedHeaderSource, headerSourcePrecedence, separator, pattern);
+        Map<String, Object> headers = prepareAMQPHeaders(flowFile, selectedHeaderSource, separator, pattern);
         builder.headers(headers);
 
         return builder.build();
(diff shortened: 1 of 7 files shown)


nifi-nar-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/ConsumeAMQP.java
```diff
@@ -191,6 +204,19 @@ public class ConsumeAMQP extends AbstractAMQPProcessor<AMQPConsumer> {
...
+    public void onScheduled(ProcessContext context) {
+        super.onScheduled(context);
+        batchSize = context.getProperty(BATCH_SIZE).asInteger();
+        queueName = context.getProperty(QUEUE).getValue();
+        headerFormat = context.getProperty(HEADER_FORMAT).getValue();
+        headerAttributePrefix = context.getProperty(HEADER_KEY_PREFIX).getValue();
+        removeCurlyBraces=context.getProperty(REMOVE_CURLY_BRACES).asBoolean();
+        valueSeparatorForHeaders = context.getProperty(HEADER_SEPARATOR).getValue();
+        autoAcknowledge = context.getProperty(AUTO_ACKNOWLEDGE).asBoolean();
+        prefetchCount =  context.getProperty(PREFETCH_COUNT).asInteger();
```
exceptionfactory, line 217: Is there a particular reason for moving these properties to class member variables? In general, it is best to keep that as method-local variables unless there is some other reason to move them around.

nifi-nar-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -58,36 +59,42 @@
...
                     + "It is an optional property. If kept empty the messages will be sent to a default AMQP exchange.")
             .required(true)
             .defaultValue("")
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .addValidator(Validator.VALID)
             .build();
-
     public static final PropertyDescriptor ROUTING_KEY = new PropertyDescriptor.Builder()
-            .name("Routing Key")
+            .name("routing.key")
```
exceptionfactory, line 96: Property Names cannot be changed with introducing an automated migration method, or following a deprecation process. In general, the display name and property name should match for new properties, so these property name changes should be reverted.

nifi-nar-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -121,30 +159,48 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
 
     private final static Set<Relationship> relationships;
 
+    private String selectedHeaderSource;
+    private String headerSourcePrecedence;
+    private Character headerSeparator;
```
exceptionfactory, line 164: See note on relocating property values.

nifi-nar-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -58,36 +59,42 @@
...
 })
 public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
-    private static final String ATTRIBUTES_PREFIX = "amqp$";
 
+    public static final AllowableValue HEADERS_FROM_ATTRIBUTES = new AllowableValue("headersFromAttributes", "Attributes Matching Regex",
+            "Select attributes based on regex pattern to put in rabbitmq headers. Key of the attribute will be used as header key");
+    public static final AllowableValue HEADERS_FROM_STRING = new AllowableValue("headersFromString", "Attribute 'amp$headers' Value",
+            "Prepare headers from 'amp$headers' attribute string");
+    public static final AllowableValue HEADERS_FROM_BOTH = new AllowableValue("headersFromBoth", "Regex Match And 'amp$headers' Value",
+            "Take headers from both sources: 'amp$headers' attribute and attributes matching Regex. In case of key duplication precedence property will define which value to take.");
```
exceptionfactory, line 84: Instead of declaring instances of `AllowableValue`, it would be helpful to define an enum of options that implemented `DescribedValue`.

exceptionfactory (changes requested):
Thanks for the proposed improvement @umarhussain15.

The general focus on the change makes sense, but there are some other stylistic recommendations I noted. It is helpful to improve the overall Processor implementation through the definition and reuse of static values for things like FlowFile attributes, so thanks for including those changes.

nifi-nar-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
umarhussain15: Good point this will lead to breaking change. I will revert this change.

nifi-nar-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/ConsumeAMQP.java
umarhussain15: I moved these to class level variables since they are only evaluated when processor is scheduled and will not change during `onTrigger`. Most of them don't have `ExpressionLanguageScope.FLOWFILE_ATTRIBUTES` or only have `ExpressionLanguageScope.ENVIRONMENT`.
But I can move them back to be method local, if they don't provide any benefit performance wise.

joewitt:
@umarhussain15 Can you please rebase and consider the feedback shared from @exceptionfactory ?

Thanks

umarhussain15:
Hi @joewitt @exceptionfactory,
I have reverted the changes of making class level variables based on processor properties. 
Also, the branch is rebased on latest changes from main

umarhussain15 (commented):
Changes applied in new commits

umarhussain15:
@exceptionfactory @joewitt 

I have added one more change, which will allow users to set amqp processor's password via env variables.
I think this will be helpful in cases where we use AMQP processors in multiple NiFI instances. Instead of copying and pasting different passwords in processor properties, we can reference the env variable via its name.
Let me know your thoughts on it.
Thanks.

exceptionfactory:
> @exceptionfactory @joewitt
> 
> I have added one more change, which will allow users to set amqp processor's password via env variables. I think this will be helpful in cases where we use AMQP processors in multiple NiFI instances. Instead of copying and pasting different passwords in processor properties, we can reference the env variable via its name. Let me know your thoughts on it. Thanks.

Thanks for calling out this particular change @umarhussain15. Given the scope of the current pull request, I recommend considering that change separately.

In general, sensitive properties such as passwords do not support reading from environment variables. The reason is that environment variables are also accessible through Expression Language, so they are not necessarily secure. The use case is still supported through Parameter Contexts and Parameter Providers, where the Environment Variable Parameter Provider can be used. This is an additional layer of indirection, but it follows the general convention for other sensitive properties.

umarhussain15:
Thanks for the explanation, @exceptionfactory. I will revert the change here then.

exceptionfactory (commented):
Thanks for following up @umarhussain15. Please note, the Checkstyle rules were recently updated to include consistency checking for whitespace, so the latest static analysis build noted these issues:

```
Warning:  src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java:[115,38] (whitespace) WhitespaceAfter: ',' is not followed by whitespace.
Warning:  src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java:[307,76] (whitespace) WhitespaceAfter: ',' is not followed by whitespace.
Warning:  src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java:[314,76] (whitespace) WhitespaceAfter: ',' is not followed by whitespace.
Warning:  src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java:[316,76] (whitespace) WhitespaceAfter: ',' is not followed by whitespace.
Warning:  src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java:[380,68] (whitespace) WhitespaceAround: '{' is not preceded with whitespace.
Warning:  src/main/java/org/apache/nifi/amqp/processors/ConsumeAMQP.java:[349,69] (whitespace) WhitespaceAround: '{' is not preceded with whitespace.
```

umarhussain15:
Checkstyle reported issues fixed

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -206,46 +247,94 @@ private byte[] extractMessage(FlowFile flowFile, ProcessSession session) {
...
+        final String attributeValue = flowFile.getAttribute(attributeKey);
         if (attributeValue == null) {
             return;
         }
 
         try {
             updater.accept(attributeValue);
         } catch (final Exception e) {
-            getLogger().warn("Failed to update AMQP Message Property " + attribute, e);
+            getLogger().warn("Failed to update AMQP Message Property " + attributeKey, e);
```
exceptionfactory, line 266:
```suggestion
            getLogger().warn("Failed to update AMQP Message Property [{}]", attributeKey, e);
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/PublishAMQPTest.java
```diff
@@ -191,6 +192,159 @@ public void validateFailedPublishAndTransferToFailure() {
...
+        runner.setProperty(PublishAMQP.HEADERS_ATTRIBUTES_REGEX, "test.*|tmp\\..*|foo2|foo3");
+
+        final Map<String, String> expectedHeaders = new HashMap<>() {{
+            put("test1", "value1");
+            put("test2", "value2");
+            put("foo2", "");
+            put("foo3", null);
+            put("tmp.test1", "tmp1");
+            put("tmp.test2.key", "tmp2");
+        }};
```
exceptionfactory, line 210: This should be changed to use `Map.of()`

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/PublishAMQPTest.java
```diff
@@ -191,6 +192,159 @@ public void validateFailedPublishAndTransferToFailure() {
...
+    public void validateSuccessWithHeaderFromBothSourcesToSuccess() throws Exception {
+        final PublishAMQP pubProc = new LocalPublishAMQP();
+        final TestRunner runner = TestRunners.newTestRunner(pubProc);
+        setConnectionProperties(runner);
+        runner.setProperty(PublishAMQP.HEADERS_SOURCE, InputHeaderSource.BOTH);
+        // header keys from `amq$headers will be put in output if present in both sources
+        runner.setProperty(PublishAMQP.HEADERS_SOURCE_PRECEDENCE, InputHeaderSource.STRING);
+        runner.setProperty(PublishAMQP.HEADERS_ATTRIBUTES_REGEX, "test.*|tmp\\..*|foo2|foo3");
+        runner.setProperty(PublishAMQP.HEADER_SEPARATOR, "|");
+        final Map<String, String> expectedHeaders = new HashMap<>() {{
```
exceptionfactory, line 260:
```suggestion
        final Map<String, String> expectedHeaders = Map.of(
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/PublishAMQPTest.java
```diff
@@ -191,6 +192,159 @@ public void validateFailedPublishAndTransferToFailure() {
...
+    public void validateSuccessAttributesPrecedenceWithHeaderFromBothSourcesToSuccess() throws Exception {
+        final PublishAMQP pubProc = new LocalPublishAMQP();
+        final TestRunner runner = TestRunners.newTestRunner(pubProc);
+        setConnectionProperties(runner);
+        runner.setProperty(PublishAMQP.HEADERS_SOURCE, InputHeaderSource.BOTH);
+        // header keys from `amq$headers will be put in output if present in both sources
+        runner.setProperty(PublishAMQP.HEADERS_SOURCE_PRECEDENCE, InputHeaderSource.ATTRIBUTES);
+        runner.setProperty(PublishAMQP.HEADERS_ATTRIBUTES_REGEX, "test.*|tmp\\..*|foo2|foo3");
+        runner.setProperty(PublishAMQP.HEADER_SEPARATOR, "|");
+        final Map<String, String> expectedHeaders = new HashMap<>() {{
```
exceptionfactory, line 309:
```suggestion
        final Map<String, String> expectedHeaders = Map.of(
```

exceptionfactory (changes requested):
@umarhussain15 This looks close to completion. There is one minor merge conflict right now related to logging that should be easy to address in a rebase and force push. The only other suggested change is to use `Map.of()` instead of anonymous HashMaps in the test methods.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -85,9 +85,9 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .addValidator(Validator.VALID)
             .build();
-
     public static final PropertyDescriptor ROUTING_KEY = new PropertyDescriptor.Builder()
             .name("Routing Key")
+            .displayName("Routing Key")
```
exceptionfactory, line 90:
```suggestion
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -96,13 +96,44 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
-
+    public static final PropertyDescriptor HEADERS_SOURCE = new PropertyDescriptor.Builder()
+            .name("Headers Source")
+            .displayName("Headers Source")
```
exceptionfactory, line 101:
```suggestion
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -96,13 +96,44 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
...
+            .description(
+                    "The source of the headers which will be put in the published message. They can come either from the processor property as a string or they can be " +
+                            "picked from flow file attributes based on Regex expression.")
```
exceptionfactory, line 104:
```suggestion
            .description("The source of the headers which will be applied to the published message.")
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -96,13 +96,44 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
...
+            .required(true)
+            .allowableValues(InputHeaderSource.getAllowedValues())
```
exceptionfactory, line 106:
```suggestion
            .allowableValues(InputHeaderSource.class)
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -96,13 +96,44 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
...
+            .defaultValue(InputHeaderSource.STRING)
+            .build();
+    public static final PropertyDescriptor HEADERS_SOURCE_PRECEDENCE = new PropertyDescriptor.Builder()
```
exceptionfactory, line 109:
```suggestion

    public static final PropertyDescriptor HEADERS_SOURCE_PRECEDENCE = new PropertyDescriptor.Builder()
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -125,26 +156,25 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
...
      * {@link FlowFile} and converted to {@link BasicProperties} to be sent along with the message. Upon success the incoming {@link FlowFile} is
      * transferred to 'success' {@link Relationship} and upon failure FlowFile is penalized and transferred to the 'failure' {@link Relationship}
      * <br>
-     *
-     * NOTE: Attributes extracted from {@link FlowFile} are considered
-     * candidates for AMQP properties if their names are prefixed with
-     * {@link PublishAMQP#ATTRIBUTES_PREFIX} (e.g., amqp$contentType=text/xml)
+     * <p>
+     * NOTE: Attributes extracted from {@link FlowFile} are considered candidates for AMQP properties if their names are prefixed with
+     * "amq$" (e.g., amqp$contentType=text/xml). For "amqp$headers" it depends on the value of
```
exceptionfactory, line 175:
```suggestion
     * "amqp$" (e.g., amqp$contentType=text/xml). For "amqp$headers" it depends on the value of
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -158,10 +188,21 @@ protected void processResource(final Connection connection, final AMQPPublisher
             throw new IllegalArgumentException("Failed to determine 'routing key' with provided value '"
                 + context.getProperty(ROUTING_KEY) + "' after evaluating it as expression against incoming FlowFile.");
         }
+        String selectedHeaderSource = context.getProperty(HEADERS_SOURCE).getValue();
```
exceptionfactory, line 191: This should use `asDescribedValue()` and return `HeaderSource` instead of the string value.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -206,46 +247,94 @@ private byte[] extractMessage(FlowFile flowFile, ProcessSession session) {
...
         return builder.build();
     }
 
+    /**
+     * Extract AMQP headers from incoming {@link FlowFile} based on selected headers source value.
+     *
+     * @param flowFile used to extract headers
+     * @return {@link Map}
+     */
+    private Map<String, Object> prepareAMQPHeaders(FlowFile flowFile, String selectedHeaderSource, String headerSourcePrecedence, Character headerSeparator, Pattern pattern) {
```
exceptionfactory, line 304: This should be changed to take the `HeaderSource` enum instead of the string value and all arguments should be declared as `final`.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -206,46 +247,94 @@ private byte[] extractMessage(FlowFile flowFile, ProcessSession session) {
...
+        }
+        return headers;
+    }
+
+    /**
+     * Matches the pattern to keys of input attributes and output the amqp headers map
+     * @param attributes flowFile attributes to scan for match
+     * @return Map with entries matching the pattern
+     */
+    private Map<String, String> attributesToHeaders(Map<String, String> attributes, Pattern pattern) {
```
exceptionfactory, line 328:
```suggestion
    private Map<String, String> getMatchedAttributes(final Map<String, String> attributes, final Pattern pattern) {
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -268,4 +357,43 @@ private Map<String, Object> validateAMQPHeaderProperty(String amqpPropValue, Cha
...
+        STRING("headersFromString", "Attribute 'amp$headers' Value",
+            "Prepare headers from 'amp$headers' attribute string"),
+        BOTH("headersFromBoth", "Regex Match And 'amp$headers' Value",
+            "Take headers from both sources: 'amp$headers' attribute and attributes matching Regex. In case of key duplication precedence property will define which value to take.");
+
+        private final String value;
+        private final String displayName;
+        private final String description;
+
+        InputHeaderSource(String value, String displayName, String description) {
```
exceptionfactory, line 373: The `value` is unnecessary and the `name()` from the enum can be used instead.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -268,4 +357,43 @@ private Map<String, Object> validateAMQPHeaderProperty(String amqpPropValue, Cha
...
+            this.description = description;
+        }
+
+        public static EnumSet<InputHeaderSource> getAllowedValues() {
+            return EnumSet.of(STRING, ATTRIBUTES, BOTH);
+        }
+
+        @Override
+        public String getValue() {
+            return value;
```
exceptionfactory, line 386:
```suggestion
            return name();
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -268,4 +357,43 @@ private Map<String, Object> validateAMQPHeaderProperty(String amqpPropValue, Cha
         }
         return headers;
     }
+    public enum InputHeaderSource implements DescribedValue {
+
+        ATTRIBUTES("headersFromAttributes", "Attributes Matching Regex",
+            "Select attributes based on regex pattern to put in rabbitmq headers. Key of the attribute will be used as header key"),
+        STRING("headersFromString", "Attribute 'amp$headers' Value",
+            "Prepare headers from 'amp$headers' attribute string"),
+        BOTH("headersFromBoth", "Regex Match And 'amp$headers' Value",
```
exceptionfactory, line 366: The value of `BOTH` is confusing and would not translate well if some new value were introduced. It seems like it should be removed, along with the Precedence property.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -96,13 +96,44 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
...
+            .displayName("Headers Source")
+            .description(
+                    "The source of the headers which will be put in the published message. They can come either from the processor property as a string or they can be " +
+                            "picked from flow file attributes based on Regex expression.")
+            .required(true)
+            .allowableValues(InputHeaderSource.getAllowedValues())
+            .defaultValue(InputHeaderSource.STRING)
+            .build();
+    public static final PropertyDescriptor HEADERS_SOURCE_PRECEDENCE = new PropertyDescriptor.Builder()
+            .name("Headers Source Precedence")
```
exceptionfactory, line 110: It seems like it would be better to remove this property and just prefer one option or the other.

exceptionfactory (changes requested):
Thanks for the updates @umarhussain15. On further review, the implementation and property names raise a few more questions, particular in terms of the `Precedence` property. If you can address the syntax issues, that would be helpful, and consider whether the `Precedence` property is necessary.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
umarhussain15: I have removed this option now

umarhussain15:
Hi @exceptionfactory,
I have applied your suggestions and also removed the precedence option from the processor. 
I have also updated the additional documentation to describe the headers source setting of the processor.

exceptionfactory (approved):
Thanks for working through the feedback and making adjustments @umarhussain15, the latest version looks ready to go. +1 merging

PR #8123, opened 2023-12-05
NIFI-12236 Improving fault tolerancy of the QuestDB backed metrics repository

simonbence:
# Summary

This is a reworking of the original `QuestDB` based status history based on community feedback. The main intention is to improve the maturity of the feature in order to be a possible default setting for 2.0. This improvement mainly focuses on but not limited to:
- Centralisation and separation of the `QuestDB` related code in order to easier maintenance
- Add capabilities to the client and database management which helps us dealing with issues (prioritising the continuity of the `NiFi` core functionality over the availability of the database)
- Clarifications on the related repository code including exposing more customisation options in the properties among other things.

The review is fairly big and complicated. My suggested order for easier understanding is the following:
- The retry capability is something might be tackled separately, it provides a robust solution to retry smaller operations and used by the `quest-db` module. The `SynchronousRetryTemplateTest` is suggested for quicker grasping.
- The newly added `quest-db` module is a good starting point for the actual `QuestDB` functionality. The included package description provides some overview of the main areas. The unit tests most helpful for understanding are the `CompositeClientTest` and the `EmbeddedDatabaseManagerTest`. Up to these point the code are considered standalone and however it follows `NiFi` conventions in the most regards, not depending on `NiFi` features.
- Last part of the PR is the changes around the application of the code above. The tests were covering the previous implementation (`EmbeddedQuestDbStatusHistoryRepositoryForComponentsTest` and `EmbeddedQuestDbStatusHistoryRepositoryForNodeTest`) are still in place and other than some smaller technical changes they still guard the function the same way. Other then these, the code organisation changed heavily. Most of the business logic around handling the snapshots are the same but as the `QuestDB` specific code are moved out from the core, the structure evolved. I suggest to start with the `EmbeddedQuestDbStatusHistoryRepository` itself and go top-down. Majority of the related code is within the same package, other changes around are to integrate it properly.

Please if any question is raising about either the approach or the implementation, do not hesitate to leave a comment about that, some additional context might be helpful to communicate my intent with the details of the change. Thanks!

[NIFI-12236](https://issues.apache.org/jira/browse/NIFI-12236)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/pom.xml
```diff
@@ -108,7 +108,7 @@
         <nifi.provenance.repository.buffer.size>100000</nifi.provenance.repository.buffer.size>
 
         <!-- Component status repository properties -->
-        <nifi.components.status.repository.implementation>org.apache.nifi.controller.status.history.VolatileComponentStatusRepository</nifi.components.status.repository.implementation>
+        <nifi.components.status.repository.implementation>org.apache.nifi.controller.status.history.questdb.EmbeddedQuestDbStatusHistoryRepository</nifi.components.status.repository.implementation>
```
exceptionfactory, line 111: I recommend reverting this particular change for this pull request. There is a good deal to review, so if we get to the place where this is ready to be the default setting, we should consider that after reviewing the substantive changes.

nifi-api/src/main/java/org/apache/nifi/controller/status/ConnectionStatus.java
```diff
@@ -22,6 +22,7 @@
 /**
  */
 public class ConnectionStatus implements Cloneable {
+    private long createdAtInMs;
```
exceptionfactory, line 25: This addition raises several questions and concerns. It does not seem like the NiFi API definition should include the creation timestamp of components in all circumstances. It would make much more sense to track this as a feature internal to the framework.

As far as naming the `InMs` suffix is awkward, and it would be much better to use `java.time.Instant` for model classes.

exceptionfactory (changes requested):
Thanks for your work on this improvement @simonbence! This is a lot to consider here, so thank you for the background description in the pull request introduction.

Before getting into the implementation details, I recommend two changes. First, it would be better to consider changing the default implementation in a separate pull request. That raises a number of questions, it would be helpful to consider those apart from the implementation itself.

Second, on initial evaluation, adding a creation timestamp to NiFi API components seems problematic, and would be better served at a framework implementation level.

Thanks again for the work here, and hopefully more reviews can evaluate the details.

joewitt:
I have questions about the direction in general so I will take those over to the JIRA.  This PR obviously was a ton of work and my questions/thoughts aren't about that right now so I think the JIRA is a better place to discuss/share those.

simonbence:
Hi @exceptionfactory !

Thanks for taking care! The separation of the default setting sounds a great idea, I will apply it with other requested changes.

As for the API changes, I looked around and the usage of the *Status classes looks to be limited to some dedicated things. It did not seem harmful, however in the other hand in many times it travels together with the capture time of the status making it somewhat coincidental. An other idea I had (to keep the API intact) is to provide a wrapper class extending the original ones with this information but as I see it would come with a the creation of a huge amount of unnecessary object creation. I am open for ideas but this looked the most fitting solution so far

nifi-api/src/main/java/org/apache/nifi/controller/status/ConnectionStatus.java
simonbence: Based on most usage I would think a time related information would make sense within status (at what time the status snapshot is relevant). I am not sure this is true for all usages but the current setup on main (these two coupled information travels together) does not seem straightforward to me. I cannot rule out however a possible third solution which complies with both things.

Instant is a good idea. I think I went down this road because some already contained creation time in this flavor but the `Instant` really makes more sense (especially that we convert it to instant many places already)

exceptionfactory: Although it might require some additional replication in the QuestDB implementation itself, it is important to avoid introducing specific concepts to the NiFi API that do not have general applicability. I would have to take a closer look at how this is used, but in general, I believe any changes to the nifi-api should be avoided for the purpose of this implementation.

nifi-api/src/main/java/org/apache/nifi/controller/status/NodeStatus.java
```diff
@@ -35,7 +35,6 @@ public class NodeStatus implements Cloneable {
     private double processorLoadAverage;
 
     private long totalThreads;
-    private long eventDrivenThreads;
```
exceptionfactory, line 38: This change is a helpful cleanup, which would be useful to handle separately. As this particular pull request may take some time to review, would you consider separating out this removal to its own pull request?

nifi-commons/nifi-questdb/src/main/java/org/apache/nifi/questdb/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,294 @@
...
+
+    }
+
+    private void ensurePersistLocationIsAccessible() throws CorruptedDatabaseException {
+        try {
+            FileUtils.ensureDirectoryExistAndCanAccess(context.getPersistLocationAsPath().toFile());
+        } catch (final Exception e) {
+            final String errorMessage = String.format("Database directory creation failed [%s]", context.getPersistLocationAsPath());
+            LOGGER.error(errorMessage, e.toString());
+            throw new CorruptedDatabaseException(errorMessage);
```
exceptionfactory, line 134: As a general implementation note, the exception cause should be passed to the new exception for tracking. As it stands, the source of the stack is neither logged nor passed, which limits potential debugging.

nifi-commons/nifi-questdb/src/test/java/org/apache/nifi/questdb/CompositeClientTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+    private Client testSubject;
+
+    @BeforeEach
+    public void setUp() {
+        databaseStructureLock = new ReentrantReadWriteLock();
+
+        testSubject = getTestSubject();
+
+        Mockito.when(condition.check()).thenReturn(true);
```
exceptionfactory, line 72: I recommend using static imports for Mockito and Assertion methods for more concise test code.

nifi-commons/nifi-questdb/src/test/java/org/apache/nifi/questdb/QuestDbTestUtil.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.io.File;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+
+public class QuestDbTestUtil {
+    static final File TEST_DB_PATH = new File("src/test/resources/testdb");
```
exceptionfactory, line 27: The JUnit 5 `@TempDir` annotation is preferable to hard-coded paths that reference source directories, since that annotation provides lifecycle management for temporary directories.

exceptionfactory (changes requested):
Thanks for the reply @simonbence.

I noted a few more general recommendations at several points. This is also a good opportunity to move the QuestDB implementation to a separate NAR. Although this relates to the largest discussion about the default setting, moving the implementation to a separate NAR would make it easier to differentiate and maintain, as opposed to continuing to have the implementation and dependency in the standard framework NAR bundle.

nifi-api/src/main/java/org/apache/nifi/controller/status/ConnectionStatus.java
simonbence: The implementation was the trigger for the change but not the sole reason. The idea was -based on how we use these- that the state snapshot and the time of relevance travels together in many times which makes it suspicious that they should be boundled together. I will not touch until you make your rounds, let's discuss after that how you feel about the approach.

simonbence:
As the changes discussed on the ticket and were resulted by NIFI-12492 would make this PR hard to follow, I close it and will open a new one

PR #8125, opened 2023-12-05
NIFI-12474: remove TriggerWhenEmpty for DeleteHDFS

emiliosetiadarma:
# Summary

[NIFI-12474](https://issues.apache.org/jira/browse/NIFI-12474)
- Added back `context.yield()` to DeleteHDFS in order to make processor yield when there is no incoming flowfiles in an incoming connection. To alleviate the high task count, I removed `@TriggerWhenEmpty` since the Kerberos login/ticket renewal has been handled properly in the following `getUserGroupInformation()` call in `onTrigger`

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/DeleteHDFS.java
+++ b/nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/DeleteHDFS.java
@@ -25,7 +25,6 @@ import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.Restricted;
 import org.apache.nifi.annotation.behavior.Restriction;
-import org.apache.nifi.annotation.behavior.TriggerWhenEmpty;
 import org.apache.nifi.annotation.behavior.WritesAttribute;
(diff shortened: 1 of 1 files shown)


nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/DeleteHDFS.java
```diff
@@ -139,6 +137,7 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
         // If this processor has an incoming connection, then do not run unless a
         // FlowFile is actually sent through
         if (originalFlowFile == null && context.hasIncomingConnection()) {
+            context.yield();
```
bbende, line 140: I don't think we need to add the `yield()` back right?

Since we are removing `@TriggerWhenEmpty` the only way we can now get a null flow file would be a concurrent task execution where one task executes the processor at a split second where another task executed and grabbed the flow file. In this case we don't need/want to yield.

emiliosetiadarma: That makes sense.. going to remove it

bbende (approved):

PR #8127, opened 2023-12-06
NIFI-7505 Add InvokeHTTPRecord processor

Aerilym:
# Summary

[NIFI-7505](https://issues.apache.org/jira/browse/NIFI-7505)
Continuation of PR #7532 

This PR isn't complete yet, but I've created it in case anybody has any feedback on the approach I'm taking. I've kept the commit from the previous PR for tracking purposes, but they will be merged before the PR merge to conform to the PR rules.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/InvokeHTTPRecord.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/InvokeHTTPRecord.java
@@ -14,21 +14,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.nifi.processors.standard;
 
-import okhttp3.*;
 import org.apache.nifi.annotation.behavior.*;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.Tags;
-import org.apache.nifi.annotation.lifecycle.OnScheduled;
 import org.apache.nifi.components.AllowableValue;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.PropertyValue;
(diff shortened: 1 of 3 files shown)


exceptionfactory (changes requested):
Thanks for providing a work-in-progress proposal for consideration @Aerilym. It is helpful to consider the implementation at this stage before considering additional work.

As I noted in several comments on the previous pull request #7532, this presents a number of implementation concerns and raises questions about whether such a processor provides a reasonable level of abstraction for general use.

InvokeHTTP has many options, which is somewhat understandable given the nature of the HTTP protocol. Bringing a similar level of flexibility to a record-oriented Processor seems less useful due to the nature of record-oriented handling. Most HTTP services have very specific contracts for communication, which may make it difficult to adapt when considering both record-oriented and HTTP protocol abstractions together. The Jira issue itself highlight different types of general use cases, whether bulk requests or requests per record.

Using the WebClientServiceProvider is helpful, but unfortunately InvokeHTTP does not provide a good general starting point. From that perspective, there are a large number of implementation concerns at the code level that would need to be addressed, beyond the concept itself.

For these reasons, I am not in favor of moving forward with an InvokeHTTPRecord Processor as currently implemented. In order to make the best of your effort, I recommend continuing the discussion on the associated Jira issue, describing some concrete use cases, which would provide opportunity for discussion. After working through some of those details, then it would be better to come back to consider implementation.

Aerilym:
Thank you for your thorough review and the time you've taken to give feedback and comment again, your tact is appreciated :)  I understand the work in its current form being rejected and the desire to return the requirements and design. Though I would have ~~like~~ liked to be able to try and initiate and participate in that process, I've been in a position where I just needed to "scratch my own itch" as quickly as possible and won't be able to devote more time to it.

Therefore, I wanted to share what I have implemented in case it is of any benefit or in case anyone else can use *it* in its current form with 1.x. Would still love to see an InvokeHTTPRecord processor at some point in future because I believe there is a very clear need, but understand it evidently needs some more experienced sponsors than myself to make that case and design it so that it has wider acceptance.

PR #8131, opened 2023-12-06
NIFI-12478 Return Message Type as body for JMS Object Messages

exceptionfactory:
# Summary

[NIFI-12478](https://issues.apache.org/jira/browse/NIFI-12478) Updates supporting utilities for the `ConsumeJMS` Processor to return the Message Type as the body when receiving JMS [ObjectMessage](https://docs.oracle.com/javaee/7/api/javax/jms/ObjectMessage.html) types. This approach avoids various problems with attempting to serialize a wrapped opaque `Object` from a JMS `ObjectMessage`. Java Object serialization is not suitable for inter-operation with subsequent Processors, and the wrapped Object may reference classes that are not available on a given Java Virtual Machine. Rather than writing the message as an error, this approach allows NiFi flows to continue receiving `ObjectMessage` types and handling the associated metadata, while using a standard value of `ObjectMessage` as the bytes for the message body.

Additional changes include correcting expected values in a JMS integration test that previously resulted in failures.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/JMSConsumer.java
+++ b/nifi-nar-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/JMSConsumer.java
@@ -38,6 +38,7 @@ import javax.jms.StreamMessage;
 import javax.jms.TextMessage;
 import javax.jms.Topic;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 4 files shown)


markap14:
Thanks @exceptionfactory I agree it's dangerous to enable using an ObjectMessage and serializing the bytes like that. I'm not really sure what exactly else to do with such a message. This approach at least ensures that the message doesn't disappear, as it passes *something* on with metadata, etc. Will merge to main.

PR #8135, opened 2023-12-06
NIFI-12484: Bumping dependency to latest minor/incremental release

mcgilman:
NIFI-12484:
- Bumping dependency to latest minor/incremental release.
- Adding an explicit override for vite which is needed until Angular can be upgraded to version 17.


rfellows:
reviewing...

rfellows (approved):
Version bumps look good and there are no longer any vulnerabilities reported. Thanks, @mcgilman .

PR #8136, opened 2023-12-07
NIFI-12487 Add CSRF Protection to Registry

exceptionfactory:
# Summary

[NIFI-12487](https://issues.apache.org/jira/browse/NIFI-12487) Adds Cross-Site Request Forgery protection to NiFi Registry using the Spring Security CSRF Filter and several components that follow the same approach currently implemented for NiFi CSRF protection.

NiFi Registry does not use cookies for passing Application Bearer Tokens, and instead relies on the HTTP Authorization header to be populated using a custom JavaScript request interceptor. This approach mitigates a number of potential threats. Introducing the CSRF Filter provides an additional layer of protection. Following a strategy similar to NiFi, the CSRF Request Matcher is based on the default Spring Security HTTP Method matching plus the presence of the Request Token cookie. This enables programmatic clients to continue working without reconfiguration, while requiring browser-based clients to pass the `Request-Token` header, as implemented through updates to the JavaScript request interceptor.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/csrf/CsrfRequestMatcher.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/csrf/CsrfRequestMatcher.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.registry.web.security.authentication.csrf;
+
+import jakarta.servlet.http.Cookie;
+import jakarta.servlet.http.HttpServletRequest;
+import org.springframework.security.web.csrf.CsrfFilter;
+import org.springframework.security.web.util.matcher.RequestMatcher;
+
+import static org.springframework.web.util.WebUtils.getCookie;
+
+/**
+ * Cross-Site Request Forgery Mitigation Request Matcher
+ */
+public class CsrfRequestMatcher implements RequestMatcher {
+
+    @Override
+    public boolean matches(final HttpServletRequest request) {
+        final boolean matches;
+
+        if (CsrfFilter.DEFAULT_CSRF_MATCHER.matches(request)) {
+            // Presence of Request Token Cookie requires invoking the CsrfFilter
+            final Cookie requestTokenCookie = getCookie(request, CsrfCookieName.REQUEST_TOKEN.getCookieName());
+            matches = requestTokenCookie != null;
+        } else {
+            matches = false;
+        }
+
+        return matches;
+    }
+}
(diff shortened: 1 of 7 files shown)


nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/csrf/CsrfRequestMatcher.java
```diff
@@ -0,0 +1,45 @@
...
+        if (CsrfFilter.DEFAULT_CSRF_MATCHER.matches(request)) {
+            // Presence of Request Token Cookie requires invoking the CsrfFilter
+            final Cookie requestTokenCookie = getCookie(request, CsrfCookieName.REQUEST_TOKEN.getCookieName());
+            matches = requestTokenCookie != null;
+        } else {
+            matches = false;
+        }
+
+        return matches;
+    }
```
dan-s1, line 44: By default when declaring `boolean matches` it is already false. Setting to false I would think only clarifies its actual value. Hence the code could be simplified with

```suggestion
        final boolean matches = false;

        if (CsrfFilter.DEFAULT_CSRF_MATCHER.matches(request)) {
            // Presence of Request Token Cookie requires invoking the CsrfFilter
            final Cookie requestTokenCookie = getCookie(request, CsrfCookieName.REQUEST_TOKEN.getCookieName());
            matches = requestTokenCookie != null;
        } 

        return matches;
    }
```

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/csrf/StandardCsrfTokenRequestAttributeHandler.java
```diff
@@ -0,0 +1,66 @@
...
+
+        final String resolvedToken;
+        if (StringUtils.hasText(headerTokenValue)) {
+            resolvedToken = super.resolveCsrfTokenValue(request, csrfToken);
+        } else {
+            resolvedToken = null;
+        }
+
+        return resolvedToken;
+    }
```
dan-s1, line 65: The default value of `resolvedToken `is already null without even setting it to null when the variable is declared. Setting to null I would think only clarifies its actual value. Hence the code could be simplified with

```suggestion
        final String resolvedToken = null;
        if (StringUtils.hasText(headerTokenValue)) {
            resolvedToken = super.resolveCsrfTokenValue(request, csrfToken);
        } 

        return resolvedToken;
    }
```

dan-s1 (commented):
I just noted some style issues.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/csrf/CsrfRequestMatcher.java
exceptionfactory: > By default when declaring `boolean matches` it is already false. Setting to false I would think only clarifies its actual value. Hence the code could be simplified with

Thanks for the feedback. The `matches` variable is declared `final`, so that is the reason for the explicit `false` assignment, which would otherwise be disallowed unless the `final` keyword were removed.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/csrf/StandardCsrfTokenRequestAttributeHandler.java
exceptionfactory: Similar to the `matches` situation, the `resolvedToken` variable is declared `final`, so it needs to be assigned explicitly in the conditionals.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/csrf/CsrfRequestMatcher.java
dan-s1: Thanks for clarifying that. Just curios, what is the advantage of using `final` here?

exceptionfactory: There is not a great benefit in this particular case with a small method and minimal conditionals, but it is more of a general approach to use `final` as the standard, and only define re-assignable variables when absolutely necessary.

bbende (approved):
Looks good to me, verified the request token is being set in the Cookie and sent back in the header

PR #8138, opened 2023-12-07
NIFI-12488: Add support for dynamic sensitive properties for the ExecuteProcess processor

lfrancke:
# Summary

[NIFI-12488](https://issues.apache.org/jira/browse/NIFI-12488)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ExecuteProcess.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ExecuteProcess.java
@@ -22,6 +22,7 @@ import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
 import org.apache.nifi.annotation.behavior.Restricted;
 import org.apache.nifi.annotation.behavior.Restriction;
+import org.apache.nifi.annotation.behavior.SupportsSensitiveDynamicProperties;
 import org.apache.nifi.annotation.behavior.WritesAttribute;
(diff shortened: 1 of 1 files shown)


lfrancke:
The build failed for me locally but on totally unrelated tests (Registry) so I assume it's okay.

exceptionfactory (changes requested):
Thanks for proposing this improvement @lfrancke.

The change itself looks useful, but there are some important additional changes required before this can go forward. Reviewing the Processor implementation, it currently logs commands at the info level and also reports commands to Provenance. These are at least two items that need to be changed.

Part of the general contract for sensitive property handling is that components will not leak sensitive information to the best of their ability. In a Processor like ExecuteProcess, called commands are still responsible for proper handling, but at least in terms of the Processor itself, it should not log sensitive information, which would be visible to NiFi operators.

For other Processors, the most straightforward approach was to remove the logging. That may be acceptable in this case, or ideally, only the base command would be logged, without the arguments. That may require more changes. If you are willing to work on making those additional changes and review the code or any other potential references writing configured arguments, then this can go forward.

exceptionfactory:
Thanks for the quick reply @lfrancke. If you get to the point where you don't think you will be able to complete the changes, I recommend noting these details on the Jira issue for future implementation.

exceptionfactory:
Thanks again for suggesting this improvement @lfrancke. Based on the current state of the PR and previous comments, I am closing it for now, but feel free to open it again if you have the opportunity to work through the rest of the Processor and implement the additional changes necessary.

lfrancke:
I'm sorry for the silence. Thank you!
I'm back with the customer that requested this only in February and might pick it up then, we'll see if they are still interested.

PR #8141, opened 2023-12-07
NIFI-12492 Move QuestDB Status Repository to Separate NAR

exceptionfactory:
# Summary

[NIFI-12492](https://issues.apache.org/jira/browse/NIFI-12492) Moves the QuestDB implementation of the Status History Repository from `nifi-framework-core` to a new `nifi-framework-questdb-nar` bundle for improved maintainability. This also provides a clear separation for the 8 MB QuestDB library from the standard framework NAR.

The changes introduce a new `nifi-framework-status-history-shared` module, which is limited to API dependencies that can be shared between the standard volatile implementation and the QuestDB implementation.

The change introduces a new `include-questdb` build profile for the `nifi-framework-questdb-nar` and also upgrades QuestDB from 7.2 to 7.3.7, addressing [NIFI-12435](https://issues.apache.org/jira/browse/NIFI-12435).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


simonbence (commented):
I reviewed the code and tested the changes. It looks good to me and I am willing to merge it. The only question I have is in this case what is the benefit of using a "nar" instead of a "jar" for `nifi-framework-questdb-status-history`? It contains no pluggable services at this point and based on my tests it should work as a simple jar. I see that the other profiles are refering to nars as well but I assume they contain some processors and controller services. What is the resason for that in this case?

exceptionfactory:
> I reviewed the code and tested the changes. It looks good to me and I am willing to merge it. The only question I have is in this case what is the benefit of using a "nar" instead of a "jar" for `nifi-framework-questdb-status-history`? It contains no pluggable services at this point and based on my tests it should work as a simple jar. I see that the other profiles are refering to nars as well but I assume they contain some processors and controller services. What is the resason for that in this case?

Thanks for the feedback @simonbence.

The extension point class is `org.apache.nifi.controller.status.history.EmbeddedQuestDbStatusHistoryRepository`, which implements the `StatusHistoryRepository` interfaces, as enumerated in the file under `META-INF/services`. This follows the same pattern as Processors and Controller Services, and is also similar to other framework extension points like the Kubernetes State Manager and Leader Election implementations. With the NAR packaging, this has the standard benefits of ClassLoader isolation and discrete packaging.

PR #8142, opened 2023-12-07
NIFI-12486: Registry Clients

mcgilman:
NIFI-12486:
- Registry Clients.
- General authorization guard.
- Additional authorization checks in the existing Settings tabs.


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/settings/ui/registry-clients/registry-client-table/registry-client-table.component.html
```diff
@@ -0,0 +1,119 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<div class="relative h-full border">
+    <div class="registry-client-table listing-table absolute inset-0 overflow-y-auto">
+        <table mat-table [dataSource]="dataSource" matSort matSortDisableClear>
```
rfellows, line 20: probably should add an initial sort:
```suggestion
        <table mat-table [dataSource]="dataSource" matSort matSortDisableClear matSortActive="name" matSortDirection="asc">
```

but, of course that would introduce ExpressionChangedAfterItHasBeenCheckedError if we continue to use the `even` helper in the row def on line 116 below.

So we should probably implement the sort function here.

rfellows (changes requested):
A few changes/suggestions...

rfellows (approved):
Looks good. Thanks, @mcgilman.

PR #8152, opened 2023-12-12
NIFI-12236 Improving fault tolerancy of the QuestDB backed metrics repository

simonbence:
This is a refined PR for NIFI-12236 after feedback and discussion on [PR 8123](https://github.com/apache/nifi/pull/8123). The general approach and purpose remained the same: the PR aims to refine error handling of the original impelmentation and in order to reach this, the QuestDb handling was refactored:
- All QuestDB related code has been moved to a separate NAR
- QuestDB usage is moved beding a Client interface, with some changes of the original usage
Most of the guidelines from the original PR are still relevant, but here are some key things have changed:
- NiFi API changes have been reverted
- Now the Volatile  implementation is the default again
- Following the changes from NIFI-12492 all QuestDB related code is not within the QuestDB bundle. Note: it has been moved out from under the framework and is part of the nifi-nar-bundles module, as other "includable" NARs.

Thank you for taking your time with the PR!

# Summary

[NIFI-12236](https://issues.apache.org/jira/browse/NIFI-12236)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/EmbeddedQuestDbStatusHistoryRepositoryDefinitions.java
```diff
@@ -0,0 +1,343 @@
...
+        put(2, ConnectionStatusDescriptor.INPUT_BYTES.getDescriptor());
+        put(3, ConnectionStatusDescriptor.INPUT_COUNT.getDescriptor());
+        put(4, ConnectionStatusDescriptor.OUTPUT_BYTES.getDescriptor());
+        put(5, ConnectionStatusDescriptor.OUTPUT_COUNT.getDescriptor());
+        put(6, ConnectionStatusDescriptor.QUEUED_BYTES.getDescriptor());
+        put(7, ConnectionStatusDescriptor.QUEUED_COUNT.getDescriptor());
+        put(8, ConnectionStatusDescriptor.TOTAL_QUEUED_DURATION.getDescriptor());
+        put(9, ConnectionStatusDescriptor.MAX_QUEUED_DURATION.getDescriptor());
+        put(10, ConnectionStatusDescriptor.AVERAGE_QUEUED_DURATION.getDescriptor());
+    }};
```
exceptionfactory, line 97: The anonymous HashMap creation with nested `put` should not be used. Recommend either `Map.of()` or using a static initializer. If the order is important, this should also be a `LinkedHashMap`.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/QuestDbStatusHistoryStorage.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private List<StatusSnapshot> getSnapshot(final String query, final QueryResultProcessor<List<StandardStatusSnapshot>> rowProcessor) {
+        return new ArrayList<>(getResult(query, rowProcessor,  Collections.emptyList()));
+    }
+
+    private <T> T getResult(final String query, final QueryResultProcessor<T> rowProcessor, final T errorResult) {
+        try {
+            return client.query(query, rowProcessor);
+        } catch (final DatabaseException e) {
+            LOGGER.error("Error during returning results for query " + query, e);
```
exceptionfactory, line 167: Log statements should always use placeholder values instead of string concatenation.
```suggestion
            LOGGER.error("Error during returning results for query {}", query, e);
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/StatusHistoryStorage.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Collection;
+import java.util.Date;
+import java.util.List;
+
+interface StatusHistoryStorage {
+
+    default void init() {};
+    default void close() {};
+
+    List<StatusSnapshot> getConnectionSnapshots(final String componentId, final Date start, final Date end);
```
exceptionfactory, line 36: Is there a reason for using `java.util.Date` as opposed to `java.time.Instant` in this new interface definition?

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+            <artifactId>commons-io</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.questdb</groupId>
+            <artifactId>questdb</artifactId>
+            <version>7.3.5</version>
```
exceptionfactory, line 47: The latest version in the main branch is 7.3.7, so this should match that version.
```suggestion
            <version>7.3.7</version>
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/ClientIsDisconnectedException.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.questdb.embedded;
+
+import org.apache.nifi.questdb.DatabaseException;
+
+final class ClientIsDisconnectedException extends DatabaseException {
```
exceptionfactory, line 21: Recommend adjusting the name to `ClientDisconnectedException`

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedClient.java
```diff
@@ -0,0 +1,152 @@
...
+            while (rowDataSource.hasNextToInsert()) {
+                context.addRow(rowDataSource);
+            }
+
+            LOGGER.debug("Committing {} rows", tableWriter.getRowCount());
+            tableWriter.commit();
+        } catch (final Exception | CairoError e) {
+            // CairoError might be thrown in extreme cases, for example when no space left on the disk
+            LOGGER.error("Add rows to table [{}] failed", tableName, e);
+            throw new DatabaseException(e);
```
exceptionfactory, line 97: Is there a reason for logging the error and throwing the exception? The caller should also catch and log the exception.

exceptionfactory (commented):
Thanks for creating the new pull request and making the adjustments @simonbence. I have yet not review the entire implementation, but I noted a few recommendations.

Having this in a separate module and avoiding changes to the nifi-api is helpful. I am not sure about the general reuse of the retry interfaces, so I'm not sure whether it warrants adding them to `nifi-utils`. Perhaps they would be better in a separate `nifi-retry-utils` under `nifi-commons`, or just kept as part of the QuestDB module in a separate package until it gets to the point where they could be reused.

simonbence:
> Thanks for creating the new pull request and making the adjustments @simonbence. I have yet not review the entire implementation, but I noted a few recommendations.
> 
> Having this in a separate module and avoiding changes to the nifi-api is helpful. I am not sure about the general reuse of the retry interfaces, so I'm not sure whether it warrants adding them to `nifi-utils`. Perhaps they would be better in a separate `nifi-retry-utils` under `nifi-commons`, or just kept as part of the QuestDB module in a separate package until it gets to the point where they could be reused.

I hope this PR will be more fit to go on with :)

As of retry: even if it is currently used only by the QuestDB module, I would consider it as a general util or tool, just like other parts of the given module, like for example the `InputStream` implementations. The concept is simple and might be applied in many place and I think we would benefit from having one extracted and tested code piece do this work if necessary. So I would not think moving it to the QuestDB module would be the right way but if you insist, I can separate it to a distinct module under `nifi-commons` (however I think that would be a bit of an exaggeration). Please share your thoughts, thanks!

nifi-commons/nifi-utils/src/main/java/org/apache/nifi/retry/NoReturnCallable.java
```diff
@@ -14,25 +14,16 @@
...
-import java.time.Instant;
+package org.apache.nifi.retry;
 
 /**
- * Readable status storage for garbage collection status entries.
+ * Represents a portion of callable code without expected return value.
  */
-public interface GarbageCollectionStatusStorage extends StatusStorage<GarbageCollectionStatus> {
+@FunctionalInterface
+public interface NoReturnCallable {
```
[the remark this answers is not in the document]
simonbence: I think, from Java Gerenics standpoint, `Void` could work out. Where it fails however is the way of intended usage. The `RetryTemplate` is primarily designed to be called not with explicit implementations of `Callable` or `NoReturnCallable` for that sake, but to give lambda expressions as arguments, like the following (from `SynchronousRetryTemplateTest`):

`testSubject.execute(() -> action.getValue("okay"), FALLBACK_ACTION);`

Creating an expression with `Void` return value to fit this intended use look to be cumbersome. So by adding `NoReturnCallable` as a functional interface, which internally wraps these "void return" or "no return" lambdas  into a 'Callable' will solve the question of easy use. With this, a no return expression can be as simple as:

`testSubject.executeWithoutValue(() -> action.consumeValue("okay"));`

nifi-docs/src/main/asciidoc/administration-guide.adoc
```diff
@@ -3538,15 +3538,19 @@ of 576.
...
 
 |====
 |*Property*|*Description*
 |`nifi.status.repository.questdb.persist.node.days`|The number of days the node status data (such as Repository disk space free, garbage collection information, etc.) will be kept. The default values
 is `14`.
 |`nifi.status.repository.questdb.persist.component.days`|The number of days the component status data (i.e., stats for each Processor, Connection, etc.) will be kept. The default value is `3`.
 |`nifi.status.repository.questdb.persist.location`|The location of the persistent Status History Repository. The default value is `./status_repository`.
+|`nifi.status.repository.questdb.persist.location.backup`|The location of the database backup in case the database is being corrupted and recreated. The default value is `./status_repository_backup`.
+|`nifi.status.repository.questdb.persist.batchsize`|The QuestDb based status history repository persists the collected status information in batches. The batch size determines the maximum number of persisted status records at a given time. The default value is `1000`.
```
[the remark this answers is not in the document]
simonbence: In most situations I think the default settings should work well and this is the reason they were not exposed in the original implementation. In cases where there are high number of processor, or the node is under high load I find it useful to give the opportunity to fine tune. What I am thinking about for example is in case of 10-30k of processors might use bigger batch size for persisting. These can be more extreme cases but I find it useful to have the opportunity for affecting on the behaviour.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-status-history-shared/pom.xml
```diff
@@ -30,5 +30,9 @@
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-framework-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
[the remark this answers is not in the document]
simonbence: You are all right, it is not necessary any more.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/CapturedStatus.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+package org.apache.nifi.controller.status.history.questdb;
+
+import java.time.Instant;
+import java.util.Objects;
+
+final class CapturedStatus<T> {
+    private final T status;
+    private final Instant capturedAt;
```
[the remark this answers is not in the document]
simonbence: The implementation uses the term "captured at" from the database tables over all. In some places I think the "at" suffix fastens understanding that it is a time based attribute. So I can change it but in that case I think we should change it everywhere. What are your thoughts?

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/EmbeddedQuestDbStatusHistoryRepositoryDefinitions.java
```diff
@@ -0,0 +1,343 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+final class EmbeddedQuestDbStatusHistoryRepositoryDefinitions {
+    /**
+     * Date format expected by the storage.
+     */
+    static final String CAPTURE_DATE_FORMAT = "yyyy-MM-dd:HH:mm:ss Z";
```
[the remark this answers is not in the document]
simonbence: I do not have the details in hand, as this is from the original implementation, but I think at the time of adding QuestDB this was either the supported or properly working format.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/EmbeddedQuestDbStatusHistoryRepositoryDefinitions.java
simonbence: The order is important but not without gaps, so the `Map.of()` will not be sufficient here. I do change it to static initializers but I would be curious what you have against the nested put here?

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/StatusHistoryStorage.java
simonbence: It is called from `public StatusHistory getConnectionStatusHistory(final String connectionId, final Date start, final Date end, final int preferredDataPoints)` which is defined by the `StatusHistoryRepository` interface. I made it to the responsiblity of the underlaying storage to convert the dates as it needs.

nifi-commons/nifi-utils/src/main/java/org/apache/nifi/retry/NoReturnCallable.java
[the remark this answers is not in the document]
exceptionfactory: Thanks for the additional background. With that in mind, it seems like having the custom `Callable` as a nested interface on the RetryTemplate would be a better approach.

Did you consider using [spring-retry](https://www.baeldung.com/spring-retry#retrytemplate) instead? Since the abstraction applies to QuestDB, it seems worth considering as it has a minimal number of dependencies.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/CapturedStatus.java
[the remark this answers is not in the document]
exceptionfactory: Taking the `nifi-client-dto` as an example, timestamp properties in those classes due not have the `At` suffix, and include properties such as `created` and `lastUpdated`. Following that convention, I recommend removing all occurrences of the `At` suffix from property names.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/EmbeddedQuestDbStatusHistoryRepositoryDefinitions.java
exceptionfactory: Although not all of the bullet points may be applicable, this is a helpful list of the disadvantages of this approach, particularly the creation new new class instances. Less of an issue with this static list, but more along the lines of an approach to be avoided.

https://www.baeldung.com/java-double-brace-initialization#disadvantages-of-using-double-braces

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb-status-history/src/main/java/org/apache/nifi/controller/status/history/questdb/QuestDbStatusHistoryStorage.java
```diff
@@ -0,0 +1,181 @@
...
+    public void storeProcessorStatuses(final Collection<CapturedStatus<ProcessorStatus>> statuses) {
+        store(TABLE_NAME_PROCESSOR_STATUS, EmbeddedQuestDbStatusHistoryRepositoryDefinitions.getProcessorStatusDataSource(statuses));
+        store(TABLE_NAME_COMPONENT_COUNTER, EmbeddedQuestDbStatusHistoryRepositoryDefinitions.getCounterStatisticsDataSource(statuses));
+    }
+
+    private <T> void store(final String tableName, final InsertRowDataSource source) {
+        try {
+            client.insert(tableName, source);
+        } catch (final DatabaseException e) {
+            LOGGER.error("Error during storing snapshots to " + tableName, e);
```
exceptionfactory, line 150: Placeholders should be used instead of string concatenation for logs.
```suggestion
            LOGGER.error("Error during storing snapshots to table [{}]", tableName, e);
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/DatabaseManager.java
```diff
@@ -14,25 +14,25 @@
...
-     * Returns with the status history of the node for the specified time range.
-     *
-     * @param start Start date of the history, inclusive.
-     * @param end End date of the history, inclusive.
-     *
-     * @return Status history.
+     * Finishes maintenance of the database. After calling, manager does not guarantee any connection with the database.
      */
-    StatusHistory read(Instant start, Instant end);
+    void close();
```
exceptionfactory, line 37: It looks this interface could extend `Closeable` given the void `close()` method.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedClient.java
```diff
@@ -0,0 +1,151 @@
...
+        } catch (final Exception e) {
+            throw new DatabaseException(e);
+        }
+    }
+
+    @Override
+    public void disconnect() throws DatabaseException {
+        checkConnectionState();
+        disconnected.set(true);
+        LOGGER.info("Client disconnects");
```
exceptionfactory, line 135: This looks like it should be at the debug level.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,297 @@
...
+            .newScheduledThreadPool(2, new BasicThreadFactory.Builder().namingPattern("EmbeddedQuestDbManagerWorker-" + id + "-%d").build());
+
+    EmbeddedDatabaseManager(final EmbeddedDatabaseManagerContext context) {
+        this.context = context;
+    }
+
+    @Override
+    public void init() {
+        if (state.get() != EmbeddedDatabaseManagerStatus.UNINITIALIZED) {
+            throw new IllegalStateException("Manager is already initialized!");
```
exceptionfactory, line 67: It is best to avoid including exclamations in messages.
```suggestion
            throw new IllegalStateException("Manager is already initialized");
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,297 @@
...
+            try {
+                ensurePersistLocationIsAccessible();
+                ensureConnectionEstablished();
+                ensureTablesAreInPlaceAndHealthy();
+                successful = true;
+            } catch (final CorruptedDatabaseException e) {
+                boolean couldMoveOldToBackup = false;
+
+                try {
+                    LOGGER.error("Database is corrupted. Recreation is triggered.", e);
```
exceptionfactory, line 90: It would be helpful to include the `backupFolder` location in this message.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+    }
+
+    private void ensurePersistLocationIsAccessible() throws CorruptedDatabaseException {
+        try {
+            FileUtils.ensureDirectoryExistAndCanAccess(context.getPersistLocationAsPath().toFile());
+        } catch (final Exception e) {
+            final String errorMessage = String.format("Database directory creation failed [%s]", context.getPersistLocationAsPath());
+            LOGGER.error(errorMessage, e.toString());
```
exceptionfactory, line 136: Is this log necessary, throwing the exception result in a log from calling code.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,297 @@
...
+                .collect(Collectors.toMap(f -> f.getAbsolutePath().substring(context.getPersistLocationAsFile().getAbsolutePath().length() + 1), f -> f));
+        final Client client = getUnmanagedClient();
+
+        try {
+            for (final ManagedTableDefinition tableDefinition : context.getTableDefinitions()) {
+                if (!databaseFiles.containsKey(tableDefinition.getName())) {
+                    try {
+                        LOGGER.info("Creating table {}", tableDefinition.getName());
+                        client.execute(tableDefinition.getDefinition());
+                        LOGGER.info("Table {} is created", tableDefinition.getName());
```
exceptionfactory, line 170: This look like they should be debug messages.
```suggestion
                        LOGGER.debug("Creating table {}", tableDefinition.getName());
                        client.execute(tableDefinition.getDefinition());
                        LOGGER.debug("Table {} created", tableDefinition.getName());
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,297 @@
...
+                    } catch (DatabaseException e) {
+                        throw new CorruptedDatabaseException(e);
```
exceptionfactory, line 172: It would be helpful to add a message indicating the table creation failed with the table name.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,297 @@
...
+            }
+        }
+    }
+
+    private void startRollover() {
+        final RolloverWorker rolloverWorker = new RolloverWorker(acquireClient(), context.getTableDefinitions());
+        final ScheduledFuture<?> rolloverFuture = scheduledExecutorService.scheduleWithFixedDelay(
+                rolloverWorker, context.getRolloverFrequency(), context.getRolloverFrequency(), context.getRolloverFrequencyTimeUnit());
+        scheduledFutures.add(rolloverFuture);
+        LOGGER.debug("Rollover is started");
```
exceptionfactory, line 205:
```suggestion
        LOGGER.debug("Rollover started");
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,297 @@
...
+    private Client getUnmanagedClient() {
+        return new EmbeddedClient(() -> engine.get());
+    }
+
+    public Client acquireClient() {
+        checkIfManagerIsInitialised();
+        final Client fallback = new DummyClient();
+
+        if (state.get() == EmbeddedDatabaseManagerStatus.CORRUPTED) {
+            LOGGER.warn("The database is corrupted. Dummy client is returned");
```
exceptionfactory, line 236: What is the behavior of the DummyClient? Perhaps NoOpClient would be clearer?

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerContext.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.concurrent.TimeUnit;
+
+interface EmbeddedDatabaseManagerContext {
+    String getPersistLocation();
+    Path getPersistLocationAsPath();
+    File getPersistLocationAsFile();
+
+    String getBackupLocation();
+    Path getBackupLocationAsPath();
+    File getBackupLocationAsFile();
```
exceptionfactory, line 31: These `asFile()` methods seem unnecessary since Path has a `toFile()` method.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerContext.java
```diff
@@ -0,0 +1,39 @@
...
+
+    int getNumberOfAttemptedRetries();
+    int getLockAttemptTime();
+    TimeUnit getLockAttemptTimeUnit();
```
exceptionfactory, line 35: This could be replaced with a single method that returns a `java.time.Duration`.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerContext.java
```diff
@@ -0,0 +1,39 @@
...
+    int getRolloverFrequency();
+    TimeUnit getRolloverFrequencyTimeUnit();
```
exceptionfactory, line 37: As above, this could be replaced with Duration.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/LockedClient.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+
+final class LockedClient implements Client {
+    private static final Logger LOGGER = LoggerFactory.getLogger(LockedClient.class);
+    private final Lock lock;
+    private final long lockAttemptTime;
+    private final TimeUnit lockAttemptTimeUnit;
```
exceptionfactory, line 34: This could be replaced with a single Duration.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/LockedClient.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.nifi.questdb.InsertRowDataSource;
+import org.apache.nifi.questdb.QueryResultProcessor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+
+final class LockedClient implements Client {
```
exceptionfactory, line 30: Is this class necessary because QuestDB does not have locking support itself?

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/SimpleEmbeddedDatabaseManagerContext.java
```diff
@@ -0,0 +1,142 @@
...
+final class SimpleEmbeddedDatabaseManagerContext implements EmbeddedDatabaseManagerContext {
+    final private Set<ManagedTableDefinition> tableDefinitions = new HashSet<>();
+
+    private String persistLocation;
+    private String backupLocation;
+    private int numberOfAttemptedRetries;
+    private int lockAttemptTime;
+    private TimeUnit lockAttemptTimeUnit;
+    private int rolloverFrequency;
+    private TimeUnit rolloverFrequencyTimeUnit;
```
exceptionfactory, line 35: See notes on using Duration.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/mapping/SimpleRequestMapping.java
```diff
@@ -0,0 +1,57 @@
...
+        return factory.get();
+    }
+
+    @Override
+    public int getNumberOfFields() {
+        return fieldMappings.size();
+    }
+
+    @Override
+    public Class<?> getFieldTypeAt(final int position) {
```
exceptionfactory, line 49: Recommend removing the `At` suffix from these methods.

exceptionfactory (commented):
Thanks for the initial adjustments @simonbence, I went through the code again and noted several recommendations and questions.

nifi-commons/nifi-utils/src/main/java/org/apache/nifi/retry/NoReturnCallable.java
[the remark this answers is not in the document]
simonbence: Good idea, I tried and it looks to me that the sprint-retry can be configured to provide the same behaviour.

nifi-commons/nifi-utils/src/test/java/org/apache/nifi/retry/ExceptionExcludingRetryConditionTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import javax.naming.OperationNotSupportedException;
+import java.io.IOException;
+import java.util.Arrays;
+
+class ExceptionExcludingRetryConditionTest {
+    final ExceptionExcludingRetryCondition testSubject = getTestSubject();
```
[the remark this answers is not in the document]
simonbence: I think in most cases (especially when the test scenario is more complicated) having `testSubject` as name leads the eye making it easier and faster to understand the test. If you do not insist, I would prefer to keep it as it is.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/DatabaseManager.java
simonbence: Yeah, that is true. I added it but kept the original method signiture without the exception as well. Currently we do not build on the "closability" and where it is considered as `DatabaseManager` we do not expect exception when closing.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedClient.java
simonbence: Closing a client is a relatively rare event so it will not pollute the log. In the other hand it can be a really valuable information when the behaviour needs to be examined.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
simonbence: Yeah, `NoOp` is a better name indeed, thanks for the idea! The original `Dummy` prefix came from the "dummy components" family, like 'DummyProcessor' but describes the intention way better.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerContext.java
simonbence: It was for convinience but I removed that

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/LockedClient.java
simonbence: This lock server different purposes than the database's internal lock. Due to the embedded nature of the database, the maintenance of it is a responsibility taken care by the `EmbeddedDatabaseManager`. This comes with time periods where the database is tampered with (kind of under maintenance) when we do not want to interact with the database. This is prevented and encapsulated by the `LockedClient`. Together with the retry, this should make shorter outages (recreating database) unnoticable for clients.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+        </dependency>
+
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-core</artifactId>
+            <version>6.1.2</version>
```
exceptionfactory, line 58: This version can be removed because it is managed at the root pom.xml. Is this dependency required?

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedClient.java
exceptionfactory, line 135: In that case, recommend the following wording adjustment:
```suggestion
        LOGGER.info("Client disconnected");
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,294 @@
...
+    private Client getUnmanagedClient() {
+        return new EmbeddedClient(() -> engine.get());
+    }
+
+    public Client acquireClient() {
+        checkIfManagerIsInitialised();
+        final Client fallback = new NoOpClient();
+
+        if (state.get() == EmbeddedDatabaseManagerStatus.CORRUPTED) {
+            LOGGER.warn("The database is corrupted. Dummy client is returned");
```
exceptionfactory, line 234: I recommend changing this to and error and adjusting the wording to indicate the behavior.
```suggestion
            LOGGER.error("The database is corrupted: Status History will not be stored");
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,294 @@
...
+            throw new IllegalStateException("The state of the database manager is not initialized");
+        }
+    }
+
+    private void errorAction(final int attemptNumber, final Throwable throwable) {
+        if (shouldRestoreDatabase(attemptNumber, throwable)) {
+            LOGGER.error("Database manager tries to restore database after the first failed attempt if necessary");
+            ensureDatabaseIsReady();
+        } else {
+            LOGGER.warn("Error happened at attempt {}: {}", attemptNumber, throwable);
```
exceptionfactory, line 258:
```suggestion
            LOGGER.warn("Error happened at attempt {}", attemptNumber, throwable);
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/rollover/DeleteOldRolloverStrategy.java
```diff
@@ -0,0 +1,95 @@
...
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.function.Supplier;
+
+final class DeleteOldRolloverStrategy implements RolloverStrategy {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DeleteOldRolloverStrategy.class);
+    private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd").withZone(ZoneOffset.UTC);
```
exceptionfactory, line 35: Is there a reason for using `UTC` in this case? It seems like it would be more intuitive to use `LocalDateTime`, matching the system time.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/util/Event.java
```diff
@@ -0,0 +1,83 @@
...
+package org.apache.nifi.questdb.util;
+
+import java.time.Instant;
+import java.util.Objects;
+
+/**
+ * This class exists for test purposes
+ */
+public class Event {
+    private Instant capturedAt;
```
exceptionfactory, line 26:
```suggestion
    private Instant captured;
```

exceptionfactory (changes requested):
Thanks for making the adjustments, this looks cleaner with the move to `spring-retry`.

I noted a few minor things, but also there appears to be a file-locking problem resulting in test failures on Windows. Linux and macOS are generally more tolerant in this area, but it highlights either a test problem, or a problem with open files in the runtime class. See the Windows build for more details.

```
Error:  org.apache.nifi.questdb.embedded.EmbeddedClientTest.testInsertAndQuery -- Time elapsed: 0.121 s <<< ERROR!
java.io.IOException: Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit16271263950658316199. The following paths could not be deleted (see suppressed exceptions for details): <root>, event, event\2024-01-14, event\2024-01-14\capturedAt.d, event\2024-01-14\subject.d, event\2024-01-14\value.d, event\subject.c, event\subject.k, event\subject.o, event\subject.v, event\_cv, event\_meta, event\_txn, event\_txn_scoreboard, tables.d.0, _tab_index.d
```

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/rollover/DeleteOldRolloverStrategy.java
simonbence: If I remember correctly (but this decision was made during the original implementation) there was a concern for nodes in different time zones.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/pom.xml
simonbence: `spring-retry` needs it and it is not a transitive for that. I removed the version that is a good point

simonbence:
> Thanks for making the adjustments, this looks cleaner with the move to `spring-retry`.
> 
> I noted a few minor things, but also there appears to be a file-locking problem resulting in test failures on Windows. Linux and macOS are generally more tolerant in this area, but it highlights either a test problem, or a problem with open files in the runtime class. See the Windows build for more details.
> 
> ```
> Error:  org.apache.nifi.questdb.embedded.EmbeddedClientTest.testInsertAndQuery -- Time elapsed: 0.121 s <<< ERROR!
> java.io.IOException: Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit16271263950658316199. The following paths could not be deleted (see suppressed exceptions for details): <root>, event, event\2024-01-14, event\2024-01-14\capturedAt.d, event\2024-01-14\subject.d, event\2024-01-14\value.d, event\subject.c, event\subject.k, event\subject.o, event\subject.v, event\_cv, event\_meta, event\_txn, event\_txn_scoreboard, tables.d.0, _tab_index.d
> ```

It looks like it has something to do with the `@TempDir` annotation. I am looking after it.

exceptionfactory:
> > Thanks for making the adjustments, this looks cleaner with the move to `spring-retry`.
> > I noted a few minor things, but also there appears to be a file-locking problem resulting in test failures on Windows. Linux and macOS are generally more tolerant in this area, but it highlights either a test problem, or a problem with open files in the runtime class. See the Windows build for more details.
> > ```
> > Error:  org.apache.nifi.questdb.embedded.EmbeddedClientTest.testInsertAndQuery -- Time elapsed: 0.121 s <<< ERROR!
> > java.io.IOException: Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit16271263950658316199. The following paths could not be deleted (see suppressed exceptions for details): <root>, event, event\2024-01-14, event\2024-01-14\capturedAt.d, event\2024-01-14\subject.d, event\2024-01-14\value.d, event\subject.c, event\subject.k, event\subject.o, event\subject.v, event\_cv, event\_meta, event\_txn, event\_txn_scoreboard, tables.d.0, _tab_index.d
> > ```
> 
> It looks like it has something to do with the `@TempDir` annotation. I am looking after it.

Thanks for looking into it. The `@TempDir` annotation is useful because JUnit will automatically clear the directory after running a test method, so those types of locking errors usually indicate that there is some other thread which has not closed a file stream.

exceptionfactory (changes requested):
Thanks for the updates @simonbence, unfortunately it looks like the cleanup approach is still failing on Windows builds.

simonbence:
The failed checks look to be in connection with Python processors and minify

exceptionfactory (changes requested):
Thanks for making the updates @simonbence.

On closer review of the test changes, there is still a problem with manager tests not cleaning up temporary files on completion. Disabling the methods on Windows masks this problem, but running the tests on macOS shows `junit` directories left behind after completion. I recommend removing the `DisabledOnOS` annotations and tracking down the source of the problem. Whether in the test class, or the implementation itself, temporary files should not remain after successful test completion.

I also noticed several QuestDB table definitions that still use `capturedAt` as the column name, so this should be checked across the set of changes and updated.

simonbence:
> Thanks for making the updates @simonbence.
> 
> On closer review of the test changes, there is still a problem with manager tests not cleaning up temporary files on completion. Disabling the methods on Windows masks this problem, but running the tests on macOS shows `junit` directories left behind after completion. I recommend removing the `DisabledOnOS` annotations and tracking down the source of the problem. Whether in the test class, or the implementation itself, temporary files should not remain after successful test completion.
> 
> I also noticed several QuestDB table definitions that still use `capturedAt` as the column name, so this should be checked across the set of changes and updated.

I take a look on the leftover temporary files.

As for disable: It was a deliberate choice after looking into the reasons. I picked adding them due to the test setup, not because of production reasons. In order to emulate corrupted database files I directly touch the files and write some gibberish into that. This is something normally would happened outside of our code, most reasonably would happen within the database. I did not found other way to trigger this however Windows does not let one to touch a file is under lock by an other process. If there is an other way to deliberatly corrupt the database (not directly working with the underlying files) than it could work out, but I am not aware of such a method.

exceptionfactory:
> > Thanks for making the updates @simonbence.
> > On closer review of the test changes, there is still a problem with manager tests not cleaning up temporary files on completion. Disabling the methods on Windows masks this problem, but running the tests on macOS shows `junit` directories left behind after completion. I recommend removing the `DisabledOnOS` annotations and tracking down the source of the problem. Whether in the test class, or the implementation itself, temporary files should not remain after successful test completion.
> > I also noticed several QuestDB table definitions that still use `capturedAt` as the column name, so this should be checked across the set of changes and updated.
> 
> I take a look on the leftover temporary files.
> 
> As for disable: It was a deliberate choice after looking into the reasons. I picked adding them due to the test setup, not because of production reasons. In order to emulate corrupted database files I directly touch the files and write some gibberish into that. This is something normally would happened outside of our code, most reasonably would happen within the database. I did not found other way to trigger this however Windows does not let one to touch a file is under lock by an other process. If there is an other way to deliberatly corrupt the database (not directly working with the underlying files) than it could work out, but I am not aware of such a method.

Thanks for looking into the details. Manual file changes for the test methods seems like a reasonable strategy, so if you can track down and resolve the leftover files issue, I would expect that should also resolve the test failures on Windows.

simonbence:
Notable changes:
- CapturedAt database fields were renamed to captured
- I found the reason of the remaining test files and fixed it
- I adjusted the NAR structure, now it is inherited from framework again, fixing an instance of issue
- With this I renamed the NAR. I would prefer the more generic `nifi-questdb-nar` as name but with the framework as parent we would have challenges with processors in the NAR. So probably if we want to add them, it will be a separate NAR thus the more specific name is beneficial.

exceptionfactory (changes requested):
Thanks for continued work on these changes @simonbence.

On further review, I returned to a situation that came up in an earlier comment regarding the `NoOpClient`.

In the current revision, the implementation logs an error and return the `NoOpClient`, which effectively disables all Status History tracking. Although this avoids causing NiFi to crash when the QuestDB database is corrupted, is also disables an important function. This might be acceptable as a non-default configuration option, but it would be a blocker for considering this as a default setting. It also seems problematic in general, even as an opt-in solution.

One option is to throw an exception that causes NiFi to fail on startup, similar to current behavior with the QuestDB repository. Another option would be to fall back to in-memory tracking, which seems better. The latter option could be more involved, but I mention it for consideration.

exceptionfactory (changes requested):
One additional note on test behavior, running tests on Ubuntu Linux 22.04 still leaves behind backup directories in a directory named `questDbBackup`. There is a new backup directory for each subsequent build of the `nifi-questdb` module.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerTest.java
```diff
@@ -0,0 +1,397 @@
...
+        testSubject.close();
+
+        // Ensuring that not the fallback client answers
+        assertEquals(3, result.size());
+    }
+
+    @Test
+    @DisabledOnOs(OS.WINDOWS)
+    // This testcase cannot be reproduced under Windows using Junit
+    public void testDatabaseRestorationAfterCorruptedFiles() throws DatabaseException, IOException {
```
exceptionfactory, line 232: For additional reference, this is the test method that leaves the `questDbBackup` directory with `backup` sub-directories after test completion. This is probably the reason for the failure on Windows, but on both macOS and Linux, the directory remains after all tests have completed. It may be necessary to implement some manual deletion logic to ensure the directory is cleared. If the directory is cleared on macOS and Linux, it might be worth enabling this test on Windows again, as it more clearly surfaces the problem.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerTest.java
```diff
@@ -0,0 +1,397 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertIterableEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class EmbeddedDatabaseManagerTest extends EmbeddedQuestDbTest {
+    private static final Logger LOGGER = LoggerFactory.getLogger(EmbeddedDatabaseManagerTest.class);
+    private static final int DAYS_TO_KEEP_EVENT = 1;
+    private static final String NON_EXISTING_PLACE = SystemUtils.IS_OS_WINDOWS ? "T:/nonExistingPlace" : "/nonExistingPlace";
```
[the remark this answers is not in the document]
simonbence: The case `testFallsBackToDummyWhenCannotEnsureDatabaseHealth` still uses this. Others usages are guarded by `@DisabledOnOs`

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/main/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManager.java
```diff
@@ -0,0 +1,294 @@
...
+    private Client getUnmanagedClient() {
+        return new EmbeddedClient(() -> engine.get());
+    }
+
+    public Client acquireClient() {
+        checkIfManagerIsInitialised();
+        final Client fallback = new NoOpClient();
+
+        if (state.get() == EmbeddedDatabaseManagerStatus.CORRUPTED) {
+            LOGGER.error("The database is corrupted: Status History will not be stored");
```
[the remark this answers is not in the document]
simonbence: The NoOp behaviour was picked after careful consideration: the original discussion on the dev mail list emphasised the need of keeping NiFi core functionality in case of the repository misbehaviours. So I needed to find a fallback strategy, which ended up the NoOp.

I considered having a temporary inmemory storage or having the Volatile implementation as backup but I ended up not using those. When NiFi is configured, including heap size, if users does not count with the possible memory consumption of the status history repository (which can be significant with high processor count) and the fallback can become the very reason for a heap overflow.

To be absolute clear on the fallback: this is considered a last resort solution. Before that, NiFi attempts to reach QuestDB multiple times. As the database instance is embedded, most possible issues are related to the environment like inaccessible disk or no free space. These are situations which should be rare and my understanding is that we do not want to deal with theses within NiFi. So all in all, NoOp seems to be the best bargain from the options I am aware of.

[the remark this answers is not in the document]
exceptionfactory: Thanks for the helpful reply. In light of the QuestDB implementation being an optional configuration, this approach looks reasonable for now.

I agree there are heap memory constraints to consider for a fallback solution using memory, and in the end, that would be analogous to the default volatile status repository.

Based on the discussion, this approach seems sufficient for this pull request, and other improvements could be evaluated later.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerTest.java
exceptionfactory: The latest updates in https://github.com/apache/nifi/pull/8152/commits/e33dbe91571ef3373f356f028641f366cb4bb6c7 now clear out the `questDbBackup` directory, but the directory itself still remains. It looks like adding one more step to delete the directory itself should resolve this particular issue.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerTest.java
```diff
@@ -0,0 +1,397 @@
...
+        final List<Event> testData = new ArrayList<>();
+        testData.add(new Event(Instant.now().minus((DAYS_TO_KEEP_EVENT + 1), ChronoUnit.DAYS), "A", 1));
+        testData.add(new Event(Instant.now(), "B", 2));
+        final DatabaseManager testSubject = getTestSubject();
+
+        final Client client = testSubject.acquireClient();
+        client.insert(EVENT_TABLE_NAME, QuestDbTestUtil.getEventTableDataSource(testData));
+
+        // The rollover runs in every 5 seconds
+        Thread.sleep(TimeUnit.SECONDS.toMillis(6));
```
exceptionfactory, line 104: This is a long time to sleep for a unit test, by adjusting the rollover time, can this value be reduced?

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerTest.java
```diff
@@ -0,0 +1,397 @@
...
+
+        final List<Event> result = client.query(SELECT_QUERY, RequestMapping.getResultProcessor(QuestDbTestUtil.EVENT_TABLE_REQUEST_MAPPING));
+        testSubject.close();
+
+        assertEquals(1, result.size());
```
exceptionfactory, line 109: When building, the module, this test method fails intermittently. It seems like the `Thread.sleep()` approach is not a strong enough guarantee of rollover completion. It may be necessary to do some additional checking on the directory content itself to determine if a rollover has occurred. If all else fails, it would be better to remove the test method, as opposed to having a test that fails intermittently.

```
[ERROR] Tests run: 13, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 18.24 s <<< FAILURE! -- in org.apache.nifi.questdb.embedded.EmbeddedDatabaseManagerTest
[ERROR] org.apache.nifi.questdb.embedded.EmbeddedDatabaseManagerTest.testRollover -- Time elapsed: 7.888 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <1> but was: <2>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:527)
	at org.apache.nifi.questdb.embedded.EmbeddedDatabaseManagerTest.testRollover(EmbeddedDatabaseManagerTest.java:109)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

exceptionfactory (changes requested):
Thanks for the latest updates and responses @simonbence.

This looks close to completion, the last remaining items are related to test behavior and stability. The directory cleanup is looking better, but still leaves an empty directory.

The other concern is the timing and intermittent failures for `testRollover()`. As mentioned in the detailed comment, the `Thread.sleep()` approach takes a lot of time for a unit test, and appears to be unreliable, so another more robust strategy for determining rollover completion seems necessary.

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerTest.java
simonbence: This is an answer for both this and the previous one. 

In general I would lean toward to use something like a latch, which guarantees the end of the background task. Due to the nature of the test case I could not use a spy for this, and I rely on waiting. As I left some gap between the roundtrip and wait time, I did not have issue with it so far, but based on your concerns I proposed a more refined way to do this. The fundamentals are the same: the test thread waits, but it polls multiple times to shorten the usual test duration and now it has 2 sec of gaps between the scheduled round trips and the waiting times. In case 3 seconds is not enough for the positive result I assume we have an issue indeed. Either with the test or the environment

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerTest.java
simonbence: Good catch, however the folder was left behind a different test case. I double checked if any leaves anything behind in multiple platform and I found that working well. Thanks for the finding!

nifi-nar-bundles/nifi-questdb-bundle/nifi-questdb/src/test/java/org/apache/nifi/questdb/embedded/EmbeddedDatabaseManagerTest.java
exceptionfactory: Thanks for adjusting the approach, the polling strategy looks good, and avoids the timing issues.

exceptionfactory (commented):
Thanks for making the adjustments to the test method @simonbence.

In the course of local build execution, I still noticed intermittent remnants of a JUnit temporary directory containing QuestDB files. I could reproduce it fairly often by just building the `nifi-questdb` module with associated tests. It is not clear which test method is leaving files behind, but the directory name indicates it is a random temporary directory from the `@TempDir` annotation. The intermittent nature of the behavior seems to indicate one remaining timing issue.

I will take a closer look at narrowing down the source of the behavior, but if we can get that isolated and resolved, then this should be ready to go.

exceptionfactory (commented):
After additional local build evaluation, the remnant temporary directory only seems to occur on Linux, not on macOS. Given that appears test-related, and does not cause failures, I don't think that it needs to be a blocker.

I pushed one minor adjustment to the test logger configuration, setting the level to `CRITICAL`, which avoids writing unnecessary QuestDB output that can result in corrupting the forked Surefire runner stream. Planning to merge pending one more successful build cycle.

exceptionfactory (commented):
@simonbence Following one more review, I tracked down the source of the orphaned files to a particular test method that was missing a close call on the `DatabaseManager`. After adding the `close()`, I could run the tests multiple times without seeing orphaned temporary directories.

I pushed one more commit to address that behavior in the test class.

exceptionfactory (approved):
Thanks for working through the feedback and addressing the various issues @simonbence. The latest version looks good. +1 merging

PR #8154, opened 2023-12-13
NIFI-12509 Changing default TTL of HazelcastMapCacheClient

simonbence:
Current default TTL is 0, which in terms of Hazelcast means that the cache will be ever-growing. Having Max Size Policy on the imap (backs the cache) can be reached via instance (server side) so it looks to be a valid setup to user 0 TTL. This makes it unreasonable to not allow the value at this point. Note: `EmbeddedHazelcastCacheManager` has currently no opportunity to set such a Max Size Policy.

# Summary

[NIFI-12509](https://issues.apache.org/jira/browse/NIFI-12509)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-hazelcast-bundle/nifi-hazelcast-services/src/main/java/org/apache/nifi/hazelcast/services/cacheclient/HazelcastMapCacheClient.java
+++ b/nifi-nar-bundles/nifi-hazelcast-bundle/nifi-hazelcast-services/src/main/java/org/apache/nifi/hazelcast/services/cacheclient/HazelcastMapCacheClient.java
@@ -78,10 +78,11 @@ public class HazelcastMapCacheClient extends AbstractControllerService implement
             .name("hazelcast-entry-ttl")
             .displayName("Hazelcast Entry Lifetime")
             .description("Indicates how long the written entries should exist in Hazelcast. Setting it to '0 secs' means that the data" +
-                    "will exists until its deletion or until the Hazelcast server is shut down.")
+                    "will exists until its deletion or until the Hazelcast server is shut down. Using `EmbeddedHazelcastCacheManager` as" +
+                    "cache manager will not provide policies to limit the size of the cache.")
             .required(true)
             .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
-            .defaultValue("0 secs") // Note: in case of Hazelcast IMap, negative value would mean "map default" which might be overridden by a different client.
+            .defaultValue("5 min") // Note: in case of Hazelcast IMap, negative value would mean "map default" which might be overridden by a different client.
             .build();
 
     private static final long STARTING_REVISION = 1;


nifi-nar-bundles/nifi-hazelcast-bundle/nifi-hazelcast-services/src/main/java/org/apache/nifi/hazelcast/services/cacheclient/HazelcastMapCacheClient.java
```diff
@@ -78,10 +78,11 @@ public class HazelcastMapCacheClient extends AbstractControllerService implement
             .name("hazelcast-entry-ttl")
             .displayName("Hazelcast Entry Lifetime")
             .description("Indicates how long the written entries should exist in Hazelcast. Setting it to '0 secs' means that the data" +
-                    "will exists until its deletion or until the Hazelcast server is shut down.")
+                    "will exists until its deletion or until the Hazelcast server is shut down. Using `EmbeddedHazelcastCacheManager` as" +
+                    "cache manager will not provide policies to limit the size of the cache.")
             .required(true)
             .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
-            .defaultValue("0 secs") // Note: in case of Hazelcast IMap, negative value would mean "map default" which might be overridden by a different client.
+            .defaultValue("5 min") // Note: in case of Hazelcast IMap, negative value would mean "map default" which might be overridden by a different client.
```
turcsanyip, line 85: @simonbence What is the exact reason for changing the default value?
Some processors store state-like data using `DistributedMapCacheClient` and they do not want to loose the cached items after some expiration time. So the default 0 (no TTL) looks reasonable for me. The user can change it if they have a different use case.
It may make sense to allow "no value" for the property and use the map default in that case (negative value in the Hazelcast API). As far as I understand, currently it is not possible to fall back to the default server side setting of the map.

simonbence: The original default value (0) is handled as "no expiration" by Hazelcast. This is by itself not something we should worry about so I intend to keep it an allowed value. Together with the `EmbeddedHazelcastCacheManager` which is one of the simplest way to back the cache this results an ever growing memory consumption. The case is the same with external Hazelcast instances where no Max Size Policy is set for the iMap behind. So having a reasonable staring value looks to be more safe as users needs to think about the expected TTL instead of just leaving it on a seemingly innocent value.

turcsanyip: @simonbence Thanks for the clarification!

+1 LGTM

turcsanyip (approved):

PR #8158, opened 2023-12-14
NIFI-12506 /nifi-api/flow/metrics endpoint times out if flow is big

timeabarna:
# Summary
If a flow is relatively big, having more than 10K processors, /nifi-api/flow/metrics endpoint times out if nifi.analytics.predict.enabled set to true as gathering predictions takes long. 
This PR modify the prediction gathering to run in parallel to speed up the processing.

[NIFI-12506](https://issues.apache.org/jira/browse/NIFI-12506)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/pom.xml
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/pom.xml
@@ -252,6 +252,8 @@
         <nifi.analytics.connection.model.implementation>org.apache.nifi.controller.status.analytics.models.OrdinaryLeastSquares</nifi.analytics.connection.model.implementation>
         <nifi.analytics.connection.model.score.name>rSquared</nifi.analytics.connection.model.score.name>
         <nifi.analytics.connection.model.score.threshold>.90</nifi.analytics.connection.model.score.threshold>
+        <nifi.analytics.prediction.parallel.processing.timeout>60 secs</nifi.analytics.prediction.parallel.processing.timeout>
+        <nifi.analytics.prediction.parallel.processing.threads>4</nifi.analytics.prediction.parallel.processing.threads>
 
         <!-- nifi.properties: flow analysis properties -->
         <nifi.flow.analysis.background.task.schedule>5 mins</nifi.flow.analysis.background.task.schedule>
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/pom.xml
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/pom.xml
@@ -252,8 +252,6 @@
         <nifi.analytics.connection.model.implementation>org.apache.nifi.controller.status.analytics.models.OrdinaryLeastSquares</nifi.analytics.connection.model.implementation>
         <nifi.analytics.connection.model.score.name>rSquared</nifi.analytics.connection.model.score.name>
         <nifi.analytics.connection.model.score.threshold>.90</nifi.analytics.connection.model.score.threshold>
-        <nifi.analytics.prediction.parallel.processing.timeout>60 secs</nifi.analytics.prediction.parallel.processing.timeout>
-        <nifi.analytics.prediction.parallel.processing.threads>4</nifi.analytics.prediction.parallel.processing.threads>
 
         <!-- nifi.properties: flow analysis properties -->
         <nifi.flow.analysis.background.task.schedule>5 mins</nifi.flow.analysis.background.task.schedule>
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
@@ -17,6 +17,7 @@
 package org.apache.nifi.web;
 
 import io.prometheus.client.CollectorRegistry;
+import jakarta.annotation.PostConstruct;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.nifi.action.Action;
 import org.apache.nifi.action.Component;
@@ -461,7 +462,18 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
 
     private RuleViolationsManager ruleViolationsManager;
 
+    private boolean analyticsEnabled;
     private Executor parallelProcessingExecutor;
+
+    @PostConstruct
+    public void postConstruct() {
+        analyticsEnabled = Boolean.parseBoolean(
+                properties.getProperty(NiFiProperties.ANALYTICS_PREDICTION_ENABLED, Boolean.FALSE.toString()));
+
+        if(analyticsEnabled) {
+            parallelProcessingExecutor = Executors.newVirtualThreadPerTaskExecutor();
+        }
+    }
 
     // -----------------------------------------
     // Synchronization methods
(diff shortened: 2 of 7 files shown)


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/pom.xml
```diff
@@ -252,6 +252,8 @@
         <nifi.analytics.connection.model.implementation>org.apache.nifi.controller.status.analytics.models.OrdinaryLeastSquares</nifi.analytics.connection.model.implementation>
         <nifi.analytics.connection.model.score.name>rSquared</nifi.analytics.connection.model.score.name>
         <nifi.analytics.connection.model.score.threshold>.90</nifi.analytics.connection.model.score.threshold>
+        <nifi.analytics.prediction.parallel.processing.timeout>60 secs</nifi.analytics.prediction.parallel.processing.timeout>
```
markap14, line 255: I would avoid adding new properties for this. Instead, recommend just hard-coding the number of threads to some value like 4. And wait up to `nifi.web.request.timeout` amount of time.
I can't see anyone really having a need to configure these properties, and the more that we add, the more overwhelming it is to configure.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -6233,27 +6237,54 @@ protected Collection<AbstractMetricsRegistry> populateFlowMetrics() {
...
+                                    c.getDestination().getName(),
+                                    c.getDestination().getIdentifier()
+                            );
+                            predictions.add(statusAnalytics.getPredictions());
+                        }));
+            } finally {
+                parallelProcessingThreadPool.shutdown();
+                try {
+                    boolean finished = parallelProcessingThreadPool.awaitTermination(parallelProcessingTimeout, TimeUnit.MILLISECONDS);
+                    if (!finished) {
```
markap14, line 6278: If it doesn't finish in time, the result is irrelevant so we should use `parallelProcessingThreadPool.shutdownNow()` in this instance also.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -6233,27 +6237,54 @@ protected Collection<AbstractMetricsRegistry> populateFlowMetrics() {
...
+        Collection<Map<String, Long>> predictions = new ConcurrentLinkedDeque<>();
+
+        final boolean analyticsEnabled = Boolean.parseBoolean(properties.getProperty(NiFiProperties.ANALYTICS_PREDICTION_ENABLED, Boolean.FALSE.toString()));
+
+        if (analyticsEnabled) {
+            final int parallelProcessingThreads = properties.getIntegerProperty(NiFiProperties.ANALYTICS_PREDICTION_PARALLEL_PROCESSING_THREADS, 4);
+            final long parallelProcessingTimeout = Math.round(FormatUtils.getPreciseTimeDuration(
+                    properties.getProperty(NiFiProperties.ANALYTICS_PREDICTION_PARALLEL_PROCESSING_TIMEOUT, "1 min"), TimeUnit.MILLISECONDS));
+
+            final ForkJoinPool.ForkJoinWorkerThreadFactory factory = pool -> {
```
exceptionfactory, line 6249: Creating Thread Pools is generally something that should happen at outside a particular method invocation so that threads can be reused. This may also be an opportunity to use virtual threads.

exceptionfactory (commented):
Thanks for working on this issue @timeabarna.

One general note, creating a thread pool for each method invocation can be expensive, and is not ideal from a resource reuse perspective. This may be a good opportunity to use virtual threads, as opposed to creating another long-lived thread pool. One of those options should be considered.

timeabarna:
Thanks @markap14 and @exceptionfactory for your review. I've removed the new properties and modified the code to reuse the thread pool as this functionality is heavily used with the 1.x version where Virtual Threads not available.

exceptionfactory (commented):
Thanks for making the adjustments @timeabarna. Creating the thread pool once is helpful, but it also highlights a concern about whether this is necessary when analytic prediction is disabled. That's the value of the virtual-thread approach, but I understand the concern for a solution that also works on NiFi 1.

This seems like a case where it would be best to implement this pull request for NiFi 2 with virtual threads, and then submit a second pull request for the support branch. The support branch could optionally create the thread pool. Theoretically the same approach could be taken in this pull request, but even making thread pool creation conditional on the analytics prediction is not as clean as virtual threads.

timeabarna:
Thanks @exceptionfactory for your review. I've updated this PR with Virtual Threads and opened [this PR ](https://github.com/apache/nifi/pull/8251) for the support branch.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -6233,27 +6239,54 @@ protected Collection<AbstractMetricsRegistry> populateFlowMetrics() {
...
         Set<Connection> connections = controllerFacade.getFlowManager().findAllConnections();
-        for (Connection c : connections) {
-            // If a ResourceNotFoundException is thrown, analytics hasn't been enabled
+        Collection<Map<String, Long>> predictions = Collections.synchronizedList(new ArrayList<>());
+        final boolean analyticsEnabled = Boolean.parseBoolean(properties.getProperty(NiFiProperties.ANALYTICS_PREDICTION_ENABLED, Boolean.FALSE.toString()));
+
+        if (analyticsEnabled) {
+            if (parallelProcessingExecutor == null) {
+                parallelProcessingExecutor = Executors.newVirtualThreadPerTaskExecutor();
+            }
```
exceptionfactory, line 6248: This conditional instantiation is not thread safe for multiple invocations of the same method. The property needs to be checked earlier in the process.

exceptionfactory (changes requested):
Thanks for making the adjustments @timeabarna, unfortunately the approach of initializing the executor in the method is not thread-safe, so it should be initialized when the NiFiProperties are first provided to the class.

timeabarna:
Thanks @exceptionfactory I've updated both pull requests based on your recommendation.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -457,6 +462,19 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
 
     private RuleViolationsManager ruleViolationsManager;
 
+    private boolean analyticsEnabled;
+    private Executor parallelProcessingExecutor;
```
exceptionfactory, line 466: This should be `ExecutorService` instead of `Executor`.

exceptionfactory (changes requested):
Thanks for the adjustment @timeabarna. I should have mentioned this previously, but creating an ExecutorService requires another lifecycle operation to close the service on shutdown.

Taking step back from this, it looks like a cleaner solution, in light of the lifecycle concern, would be to define the ExecutorService outside of the StandardNiFiServiceFacade itself, and instead inject in the Spring context configuration that provides other services. That avoids the new PostConstruct method, and allows Spring to use standard lifecycle handling to manage the ExecutorService. Unfortunately this means another change is necessary, but this will make the approach more robust when starting and stopping NiFi.

timeabarna:
Thanks @exceptionfactory , I've created a processing service and updated both PRs.

exceptionfactory (changes requested):
Thanks for making the adjustments @timeabarna, moving the processing to a separate service looks like a reasonable approach.

The `VirtualThreadParallelProcessingService` still needs to close the `ExecutorService` that gets created in the constructor. If Processing Service implements the `Closeable` interface and calls `ExecutorService.close()` in that method, then Spring will shutdown the service cleanly. That should be the last change required.

timeabarna:
Thanks @exceptionfactory, I've updated both PRs with shutdown.

exceptionfactory (approved):
Thanks for working through the feedback @timeabarna, the latest version looks good and manages the executor lifecycle as expected. +1 merging

PR #8166, opened 2023-12-15
[NIFI-12519] - deselect parameter when it is deleted.

rfellows:
# Summary

[NIFI-12519](https://issues.apache.org/jira/browse/NIFI-12519)

* deselect parameter when it is deleted so it no longer shows its referencing components.
* also, fixed a minor bug preventing goto controller service linking from parameter reference.


mcgilman:
Will review...

mcgilman (approved):
Looks great. Thanks @rfellows! +1

PR #8185, opened 2023-12-22
NIFI-12529: Enable/Disable Controller Service

mcgilman:
NIFI-12529:
- Enable/Disable Controller Services.
- Resetting state when destroying Controller Service listings, General, Registry Clients, and Reporting Tasks.


rfellows:
will review

rfellows:
reviewing

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/app.module.ts
```diff
@@ -56,7 +57,13 @@ import { MatDialogModule } from '@angular/material/dialog';
             routerState: RouterState.Minimal,
             navigationActionTiming: NavigationActionTiming.PostActivation
         }),
-        EffectsModule.forRoot(UserEffects, ExtensionTypesEffects, AboutEffects, StatusHistoryEffects),
+        EffectsModule.forRoot(
+            UserEffects,
+            ExtensionTypesEffects,
+            AboutEffects,
+            StatusHistoryEffects,
+            EnableControllerServiceEffects
```
rfellows, line 65: This name, `EnableControllerServiceEffects`, gives the impression the effects are only useful for enabling controller services. However, all of the disable-related effects are defined in there as well.
Can we find a more appropriate name? I suppose this would include the actions, reducers, and selectors as well.

rfellows (changes requested):
A couple of things flagged inline. The main concern I have is regarding a user-experience difference between the new UI and the original. When enabling with referencing components 2 different CS's that have the same referencing component... the 2nd one reports a failure in a way that is more jarring than NiFi did.

There technically is an error when attempting to start the referencing processors they fail as were already set to a scheduled state of RUNNING. However, since not all of its controller service references where enabled, it couldn't fully start until the second CS was enabled. The backend returns a 409 CONFLICT in this case.

Original NiFi opens a new dialog with the error message whereas the new UI displays the error inline in the initial dialog. This is nice except that the red text and red `X` make it seem like there is something wrong. When in reality, everything worked as they intended.

This is more of a conversation starter for some ideas on an alternative approach.
* Can we handle the 409 specifically and ignore it?
* Should we display this message in a more subtle way?
* Am I making more out of this than I should?


Here is the new use case I tried to describe:
Kapture 2023-12-27 at 14 42 14


Here is the original:
Kapture 2023-12-27 at 14 55 31

mcgilman:
Thanks for the review @rfellows! I've pushed a new commit addressing the items you outlined.

rfellows (approved):
Excellent changes, @mcgilman. I think it looks really good now with the smaller, more subtle inline error messages. Thanks for the updates!

PR #8188, opened 2023-12-24
NIFI-12547: Set the width of the Property editor.

mcgilman:
- Set the width of the Property editor. This ensures the width looks correct and the overflow scroll policy works as expected.
- Fixed table layout issues where width: 100% from listing table wasn't correct.

property-editor-width


rfellows:
reviewing...

rfellows (approved):
Looks goo to me. thanks for the fix, @mcgilman

PR #8192, opened 2023-12-29
[NIFI-12552] - Support client-side pagination on Summary tables

rfellows:
# Summary

[NIFI-12552](https://issues.apache.org/jira/browse/NIFI-12552)

Also:
* updated the filter control to only show the label indicating matches when a filter is active. similar to how it is done for the events table
* aligned cells like Run Status so the icon and text are vertically centered in the table cell
* made minor changes suggested by eslint in files touched.


mcgilman (approved):
Thanks for the updates @rfellows! +1

PR #8211, opened 2024-01-07
NIFI-12573 Improve support for Enum values in PropertyDescriptor.Builder

EndzeitBegins:
# Summary

[NIFI-12573](https://issues.apache.org/jira/browse/NIFI-12573)
Lowers type restrictions on three methods in `PropertyDescriptor.Builder` to allow to work with `DescribedValue` instead of child class `AllowableValue` and with both `Enum` classes that do and do not implement `DescribedValue`.

Also addresses [NIFI-12574](https://issues.apache.org/jira/browse/NIFI-12574) by adding a new method `clearDefaultValue` to `PropertyDescriptor.Builder`. 

Adds several test cases for the affected code.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/components/AllowableValue.java
+++ b/nifi-api/src/main/java/org/apache/nifi/components/AllowableValue.java
@@ -43,9 +43,8 @@ public class AllowableValue implements DescribedValue {
      * Constructs a new AllowableValue with the given value and display name and
      * no description
      *
-     * @param value that is allowed
+     * @param value       that is allowed
      * @param displayName to display for the value
-     *
      * @throws NullPointerException if either argument is null
      */
     public AllowableValue(final String value, final String displayName) {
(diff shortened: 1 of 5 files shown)


exceptionfactory (commented):
Thanks for the contribution @EndzeitBegins. Supporting any `enum` would follow a similar pattern to the existing methods that take a set of strings. Is there any other particular reason prompting these changes? In the course of new implementations, providing an enum that implements `DescribedValue` is always preferred because it requires methods for display name and description.

EndzeitBegins:
Thanks for the questions @exceptionfactory.
First of all I agree that it should be preferred to let the enums implement `DescribedValue` whenever feasible.
Maybe that's something worth to better highlight in the code documentation. 

In particular I thought about two situations, where the more adaptive approach introduced in this PR would be beneficial.
1. There might be cases we're an existing enum (e.g. from a library) is used that cannot be altered. Even though I'd personally prefer creating one's own enum (implementing DescribedValue) that maps to the library one (similar to the azure bundle) I can see how someone might not want to go "the extra mile".
2. And by far more important to me, I noticed that when using the existing `public <E extends Enum<E>> Builder allowableValues(final E[] values)` with an enum that actually implements DescribedValue, those values will be ignored, which might come as a surprise.

exceptionfactory:
Thanks for the quick reply @EndzeitBegins. That's a very good point about the second scenario, and the first one is also reasonable. With that background, I agree with the changes and will proceed with merging. +1

exceptionfactory (approved):

EndzeitBegins:
As always, thank you for input and taking on the merge process @exceptionfactory. 🙂👍🏻

PR #8237, opened 2024-01-11
NIFI-12597: Introducing a common navigation bar across all pages

mcgilman:
NIFI-12597:
- Introducing a common navigation bar across all pages.


scottyaslan (approved):
LGTM! Thanks for these changes @mcgilman! I really like the direction!

One thought for the future: I think it may be useful to also see the flow status bar in each page. It may also be beneficial to have the search available as well.

PR #8249, opened 2024-01-15
NIFI-12554 Placed all Jolt related code under a new nifi-jolt-bundle and refactored to reduce duplicate code.

dan-s1:
- Created a dedicated bundle of all Jolt related software
   - The new structure is the top level directory named nifi-jolt-bundle with sub  directories
		- nifi-jolt-processors
                - nifi-jolt-transform-json-ui
                - nifi-jolt-utils (instead of nifi-standard-utils)
                - nifi-jolt-nar
   - The JoltTransformJSON and JoltTransformRecord record processors were moved under one module nifi-jolt-processors.
   - The TransformFactory which was duplicated originally in the original modules for JoltTransformJSON and JoltTransformRecord is no longer duplicated and is found in nifi-jolt-utils.
   - Common code between JoltTransformJSON and JoltTransformRecord was pulled up to an abstract class AbstractJoltTransform and contains:
     - Definitions of common properties
     - Instantiation of a specs cache.
     - Retrieval from cache
     - custom validate
   - Defined strings representing the various types of Jolt transforms in one place in the JoltTransformStrategy enum. This allows for their use with a common enum across the nifi-jolt-utils (in TransformFactory) module and the nifi-jolt-processors module (AbstractJoltTransform).
   
 - Upgraded the JUnit tests with JUnit 5 features e.g.
   public class, non public test methods and the use of assertThrows instead
   of try catch statements. Also upgraded to Java 21 features e.g. Files.readString
 - In TestJoltTransformJSON removed duplicate instation of processor and testrunner in each test method and placed in setUp method annotated with Jupitor Junit BeforeEach.
 - Removed the duplicate Jolt specs and placed the common ones in a directory so both tests can share them.
 - In TestJolTransformRecord replaced the hard coded string for pretty print with the JsonRecordSetWriter.PRETTY_PRINT_JSON property.

# Summary

[NIFI-12554](https://issues.apache.org/jira/browse/NIFI-12554)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-nar-bundles/nifi-jolt-bundle/pom.xml
```diff
@@ -57,6 +45,22 @@
...
+            <dependency>
+                <groupId>javax.servlet.jsp</groupId>
+                <artifactId>javax.servlet.jsp-api</artifactId>
+                <version>2.3.3</version>
+            </dependency>
+            <dependency>
+                <groupId>javax.ws.rs</groupId>
+                <artifactId>javax.ws.rs-api</artifactId>
+                <version>2.1.1</version>
+            </dependency>
```
exceptionfactory, line 57: These dependencies are the old versions that are no longer current following the upgrade to Spring 6 and Jetty 12. These should be replaced with the new Jakarta equivalent versions.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-nar/src/main/resources/META-INF/NOTICE
```diff
@@ -10,7 +10,7 @@ This product includes the following work from the Apache Hadoop project under Ap
 This includes derived works from the Apache Software License V2 library Jolt (https://github.com/bazaarvoice/jolt)
   Copyright 2013-2014 Bazaarvoice, Inc
   The derived work is adapted from com.bazaarvoice.jolt.chainr.ChainrBuilder.java, com.bazaarvoice.jolt.chainr.spec.ChainrSpec.java,
-  com.bazaarvoice.jolt.chainr.spec.ChainrEntry.java and can be found in the org.apache.nifi.processors.standard.util.jolt.TransformFactory.java class.
+  com.bazaarvoice.jolt.chainr.spec.ChainrEntry.java and can be found in the util.org.apache.nifi.processors.jolt.TransformFactory.java class.
```
exceptionfactory, line 13: This appears to be incorrect and should be reverted.

nifi-nar-bundles/nifi-standard-bundle/pom.xml
```diff
@@ -80,18 +77,19 @@
...
+                <groupId>javax.servlet</groupId>
+                <artifactId>javax.servlet-api</artifactId>
+                <version>3.1.0</version>
             </dependency>
             <dependency>
-                <groupId>jakarta.ws.rs</groupId>
-                <artifactId>jakarta.ws.rs-api</artifactId>
+                <groupId>javax.ws.rs</groupId>
+                <artifactId>javax.ws.rs-api</artifactId>
+                <version>2.1.1</version>
```
exceptionfactory, line 92: These javax dependencies are no longer current and should be removed.

nifi-nar-bundles/nifi-standard-bundle/pom.xml
```diff
@@ -101,7 +99,7 @@
             <dependency>
                 <groupId>com.hierynomus</groupId>
                 <artifactId>sshj</artifactId>
-                <version>0.38.0</version>
+                <version>0.37.0</version>
```
exceptionfactory, line 102: This change should be reverted.

nifi-nar-bundles/nifi-standard-bundle/pom.xml
```diff
@@ -210,6 +208,33 @@
...
+                    <exclusion>
+                        <groupId>org.apache.geronimo.specs</groupId>
+                        <artifactId>geronimo-jpa_2.0.spec</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.apache.bval</groupId>
+                        <artifactId>bval-xstream</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
```
exceptionfactory, line 237: This dependency should be removed as it is not used.

nifi-nar-bundles/nifi-standard-bundle/pom.xml
```diff
@@ -265,7 +290,7 @@
             <dependency>
                 <groupId>com.networknt</groupId>
                 <artifactId>json-schema-validator</artifactId>
-                <version>1.1.0</version>
+                <version>1.0.87</version>
```
exceptionfactory, line 293: This change should be reverted.

exceptionfactory (changes requested):
Thanks for working on this migration of the Jolt components @dan-s1.

Before taking a closer look at the code changes, it appears that there are several Maven configuration changes that do not align with the current main branch. It appears to be the result of not taking the latest changes from the current main branch, so there are a handful of dependency versions and files that need to be corrected before going forward.

nifi-nar-bundles/nifi-standard-bundle/pom.xml
dan-s1: Kind of weird as I updated main and then on my branch I rebased with main, I am not sure why all these changes were not picked up.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-nar/src/main/resources/META-INF/NOTICE
dan-s1: Yeah I noticed that also after I pushed.  You beat me to it :)

EndzeitBegins:
Thank you for working on this @dan-s1. 👍🏻 

I haven't done a full code review so I might've just overlooked something.
But I noticed that the package names of the processors change due to the move from the standard-nar to the new jolt-nar. 
From my understanding this will result in the processors needing to be replaced manually on the canvas between version updates, or is there a migration in place for this sort of change?
If not, and the user needs to replace those processors themself, I don't think we need to migrate properties as there won't be any instance of the processor existing anyway.

exceptionfactory (changes requested):
It looks like the `nifi-code-coverage` module still references `nifi-jolt-record-processors`, resulting in build failures, so that also needs to be updated to the new module coordinates.

dan-s1:
@exceptionfactory Thanks for clarifying what the builds were failing on as I could not see that. I only saw Error:

dan-s1:
@exceptionfactory Do I need to add the new modules I created to `nifi-code-coverage` `pom.xml`? Which ones do I add `nifi-jolt-processors` and `nifi-jolt-utils` ?

exceptionfactory:
> @exceptionfactory Do I need to add the new modules I created to `nifi-code-coverage` `pom.xml`? Which ones do I add `nifi-jolt-processors` and `nifi-jolt-utils` ?

Yes, those two should be added and the old one should be removed.

dan-s1:
@exceptionfactory I am now able to see the stacktraces from the builds. Apparently there was also an issue with the `nifi-standard-utils` which no longer exists. So I removed both `nifi-standard-utils` and  `nifi-jolt-record-processors` and replaced them with `nifi-jolt-processors` and `nifi-jolt-utils`.

dan-s1:
> Thank you for working on this @dan-s1. 👍🏻
> 
> I haven't done a full code review so I might've just overlooked something. But I noticed that the package names of the processors change due to the move from the standard-nar to the new jolt-nar. From my understanding this will result in the processors needing to be replaced manually on the canvas between version updates, or is there a migration in place for this sort of change? If not, and the user needs to replace those processors themself, I don't think we need to migrate properties as there won't be any instance of the processor existing anyway.

@EndzeitBegins I am not sure the answer to your question. I was following what @exceptionfactory had told me as seen in [NIFI-12554](https://issues.apache.org/jira/browse/NIFI-12554) in the comment on 1/9/2024.

exceptionfactory (changes requested):
@dan-s1 in the process of testing, it is clear that the migrateProperties method is not useful in this scenario. I wasn't quite clear on the initial comment, but that method can be removed. With moving these components to a new NAR, they are effectively treated as new components, so automated migration is not possible.

The only other high-level change is to have the `nifi-jolt-bundle` use `nifi-standard-shared-bom` as the parent, and set `nifi-standard-shared-nar` as the NAR dependency in `nifi-jolt-nar` so that the Jolt NAR does not include additional copies of the Jackson JAR libraries.

dan-s1:
> nifi-standard-shared-nar



> @dan-s1 in the process of testing, it is clear that the migrateProperties method is not useful in this scenario. I wasn't quite clear on the initial comment, but that method can be removed. With moving these components to a new NAR, they are effectively treated as new components, so automated migration is not possible.
> 
> The only other high-level change is to have the `nifi-jolt-bundle` use `nifi-standard-shared-bom` as the parent, and set `nifi-standard-shared-nar` as the NAR dependency in `nifi-jolt-nar` so that the Jolt NAR does not include additional copies of the Jackson JAR libraries.

@exceptionfactory  I thought the `nifi-standard-shared-nar` is already a dependency in the `nifi-jolt-nar`. Below is a cut and paste from the `nifi-jolt-nar` `pom.xml`
```
 <dependency>
            <artifactId>nifi-standard-shared-nar</artifactId>
            <type>nar</type>
        </dependency>
```

exceptionfactory:
> > nifi-standard-shared-nar
> 
> > @dan-s1 in the process of testing, it is clear that the migrateProperties method is not useful in this scenario. I wasn't quite clear on the initial comment, but that method can be removed. With moving these components to a new NAR, they are effectively treated as new components, so automated migration is not possible.
> > The only other high-level change is to have the `nifi-jolt-bundle` use `nifi-standard-shared-bom` as the parent, and set `nifi-standard-shared-nar` as the NAR dependency in `nifi-jolt-nar` so that the Jolt NAR does not include additional copies of the Jackson JAR libraries.
> 
> @exceptionfactory I thought the `nifi-standard-shared-nar` is already a dependency in the `nifi-jolt-nar`. Below is a cut and paste from the `nifi-jolt-nar` `pom.xml`
> 
> ```
>  <dependency>
>             <groupId>org.apache.nifi</groupId>
>             <artifactId>nifi-standard-shared-nar</artifactId>
>             <type>nar</type>
>         </dependency>
> ```

Thanks @dan-s1, you are correct, I just noted that for completeness. The other element still needs to be changed so that the bundle uses `nifi-standard-shared-bom` as the parent, which will mark the appropriate dependencies as provided, thus reducing the NAR size.

If you could address that, and rebase on the current main branch, that would be helpful.

exceptionfactory (approved):
Thanks for making the changes @dan-s1. In the course of testing, I had to change the property name references in the custom UI, but that was a simple adjustment. Otherwise, the configuration of both components works as expected.

This will be a breaking change from previous NiFi versions as highlighted in previous discussion, but given the custom UI and Jolt itself, this seems warranted to decouple from the `nifi-standard-nar`. This is something to be highlighted in the migration guide, which I will note as part of the release process.

+1 merging

PR #8266, opened 2024-01-18
NIFI-9677 Fixed LookupRecord issue that an empty JSON array caused mismatch

jrsteinebrey:
### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-9677) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

I added two new unit tests to verify my fix.
I also manually tested the new code works as intended in a local build of NiFi editor.

### Build

  - [X] JDK 21

### Licensing

N/A

### Documentation

N/A

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/LookupRecord.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/LookupRecord.java
@@ -487,7 +487,13 @@ public class LookupRecord extends AbstractProcessor {
                 final RecordPath recordPath = entry.getValue();
 
                 final RecordPathResult pathResult = recordPath.evaluate(record);
-                final List<FieldValue> lookupFieldValues = pathResult.getSelectedFields()
+                final List<FieldValue> allFieldValues = pathResult.getSelectedFields().toList();
+                if (allFieldValues.isEmpty()) {
+                    // An empty JSON array was found which counts as a match.
+                    // Since array is empty, no further processing is needed, so continue.
+                    continue;
+                }
+                final List<FieldValue> lookupFieldValues = allFieldValues.stream()
                     .filter(fieldVal -> fieldVal.getValue() != null)
                     .collect(Collectors.toList());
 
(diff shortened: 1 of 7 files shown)


mattyb149:
Reviewing...

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/LookupRecord.java
```diff
@@ -487,7 +487,13 @@ public Set<Relationship> lookup(final Record record, final ProcessContext contex
                 final RecordPath recordPath = entry.getValue();
 
                 final RecordPathResult pathResult = recordPath.evaluate(record);
-                final List<FieldValue> lookupFieldValues = pathResult.getSelectedFields()
+                final List<FieldValue> allFieldValues = pathResult.getSelectedFields().toList();
+                if (allFieldValues.isEmpty()) {
```
mattyb149, line 491: I was able to get this "working" by just changing line 503 to `continue;` after setting `hasUnmatchedValue = true`, removing the temp `rels` variable and updating the debug line. That avoids having to call `toList()` and then `stream()` later. But since it should not be an unmatched value, that's less than ideal too.

I think the crux of the issue is that `getSelectedFields()` returns an empty list, when it should return a field with an `Object[0]` in it. This is a result of `WildcardIndexPath:67` which generates the empty list instead of a singleton list with a empty array in it. We could add a check for `array.length == 0` there and return the one-element empty array instead.

jrsteinebrey: @mattyb149 I refactored the code to avoid streaming twice. Please review again.

mattyb149:
When running the unit tests with Java 8 (for a `support/nifi-1.x` backport), I get an error because the `foo` entry shows up as `null` in the output of `testLookupMissingJsonField()`:

`[{"foo":null,"unmentioned":{"foo":"original"}}]`

I checked some commits and everything looks the same, but I'm guessing there was some change made to the record-based stuff in `main` that didn't make its way back to the 1.x branch. Either way JSON is not guaranteed to have "ordered" fields, so we should figure out why the `"foo": null` is in there (if it shouldn't be) or check each field for its contents rather than the whole output FlowFile.

mattyb149:
+1 LGTM, thanks for the fix! Merging to support/nifi-1.x and main

PR #8273, opened 2024-01-19
NIFI-11520: Add a menu to display Flow Analysis report results

sardell:
# Summary
[NIFI-11520](https://issues.apache.org/jira/browse/NIFI-11520)

This PR introduces a menu to the canvas that displays the results of a Flow Analysis report. This will enable UI users to see a list of their established rules and violations (scoped to the currently selected Process Group in the canvas). I'm not sure how this will translate over to the new UI work that's currently in-progress, but I'll gladly take responsibility for adding it there as well.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,554 @@
...
+                /**
+                 * Initialize the flow analysis controller.
+                 */
+                init: function () {
+                    var flowAnalysisCtrl = this;
+                    var drawer = $('#flow-analysis-drawer');
+                    var requiredRulesEl = $('#required-rules');
+                    var recommendedRulesEl = $('#recommended-rules');
+                    var newFlowAnalsysisBtnEl = $('#flow-analysis-check-now-btn');
+                    var flowAnalysisRefreshIntervalSeconds = 5;
```
sardell, line 809: Is there a constant elsewhere used for polling that I should be using here?  `5` is an arbitrary value used for development. I'm guessing we don't want polling that often.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,554 @@
...
+                        }
+
+                        function openRuleMoreInfoDialog() {
+                            $('#rule-menu').hide();
+                            $('#rule-type-pill').empty()
+                                                .removeClass()
+                                                .addClass(ruleInfo.enforcementPolicy.toLowerCase() + ' rule-type-pill')
+                                                .append(ruleInfo.enforcementPolicy);
+                            $('#rule-display-name').empty().append(ruleInfo.name);
+                            $('#rule-description').empty().append(ruleInfo.descriptors['component-type'].description);
```
tpalfy, line 711: This refers to the description of the PropertyDescriptor named 'component-type'.
That is very specific to the DisallowComponentType rule.
If a rule doesn't have that descriptor this  fails with an error.

Important to not that some Rules don't have any PropertyDescriptors at all.

In general adding the PropertyDescriptors to this dialog page is not desirable in my opinion.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,554 @@
...
+                        $(document).on('click', closeViolationWindow);
+
+                        function closeViolationWindow(e) {
+                            if ($(e.target).parents("#violation-menu").length === 0) {
+                                $("#violation-menu").hide();
+                                unbindViolationMenuHandling();
+                            }
+                        }
+
+                        function openRuleMoreInfoDialog() {
```
tpalfy, line 762:
```suggestion
                        function openRuleViolationMoreInfoDialog() {
```

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,554 @@
...
+                            var rule = response.rules.find(function(rule){ 
+                                return rule.id === violationInfo.ruleId;
+                            })
+                            $('#violation-menu').hide();
+                            $('#violation-type-pill').empty()
+                                                    .removeClass()
+                                                    .addClass(violationInfo.enforcementPolicy.toLowerCase() + ' violation-type-pill')
+                                                    .append(violationInfo.enforcementPolicy);
+                            $('#violation-display-name').empty().append(violationInfo.violationMessage);
+                            $('#violation-description').empty().append(rule.descriptors['component-type'].description);
```
tpalfy, line 772: The same issue here as above:
This refers to the description of the PropertyDescriptor named 'component-type'.
That is very specific to the DisallowComponentType rule.
If a rule doesn't have that descriptor this fails with an error.

But in general I don't think adding the Rule's PropertyDescriptors'  info to the rule violation info dialog page is desirable.

tpalfy (changes requested):

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
sardell: Just to be clear, I think the dialog would be really valuable for a user if we did have descriptions for rules so they don't have to constantly look up documentation to find a short description for what a rule does, but if the property doesn't exist or isn't always present it would probably be better just to direct the user to the rule documentation.

sardell:
> Hey @sardell 👋 Thanks for the PR! Would love to have you help contribute this into the new UI work. That work is still in progress and is being tracked here [1]. Please add a line item there and file a sub-task. Please see [2] for a high-level overview of layout, structure, and dataflow throughout the application. These details will be formalized in a README soon, but wanted to pass along a pointer in the meantime.
> 
> The UX in this new effort is largely unchanged from the previous UI. So any designs for this PR should be directly applicable to the new UI. The only real difference is moving away from the large full-screen modal (though that may still be used for component documentation or custom UIs yet to be developed) to separate pages.
> 
> [1] https://issues.apache.org/jira/browse/NIFI-12400 [2] #8053

@mcgilman Thanks for that quick rundown. I'm very excited to see this UI framework update coming together so fast! I've added a line item and sub-task to NIFI-12400 and will begin work shortly.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
tpalfy: For now I'd go with directing the user to the rule documentation.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,554 @@
...
+                            $('#rule-menu').hide();
+                            $('#rule-type-pill').empty()
+                                                .removeClass()
+                                                .addClass(ruleInfo.enforcementPolicy.toLowerCase() + ' rule-type-pill')
+                                                .append(ruleInfo.enforcementPolicy);
+                            $('#rule-display-name').empty().append(ruleInfo.name);
+                            $('#rule-description').empty().append(ruleInfo.descriptors['component-type'].description);
+                            $( "#rule-menu-more-info-dialog" ).modal( "show" );
+                            $('.rule-docs-link').click(function () {
+                                // open the documentation for this flow analysis rule
```
tpalfy, line 714: This "View documentation" is has some error in it.
2 symptoms (I assume both require multiple enable rules of different types):
1. Many time the documentation of a wrong rule is shown.
2. When debugging this function runs multiple times.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,554 @@
...
+                            $('#violation-menu').hide();
+                            $('#violation-type-pill').empty()
+                                                    .removeClass()
+                                                    .addClass(violationInfo.enforcementPolicy.toLowerCase() + ' violation-type-pill')
+                                                    .append(violationInfo.enforcementPolicy);
+                            $('#violation-display-name').empty().append(violationInfo.violationMessage);
+                            $('#violation-description').empty().append(rule.descriptors['component-type'].description);
+                            $('#violation-menu-more-info-dialog').modal( "show" );
+                            $('.violation-docs-link').click(function () {
+                                // open the documentation for this flow analysis rule
```
tpalfy, line 775: Similarly to Rule details, this "View documentation" is has some error in it.
2 symptoms (I assume both require multiple enable rules of different types):
1. Many time the documentation of a wrong rule is shown.
2. When debugging this function runs multiple times.

tpalfy (changes requested):

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
mcgilman: In `nf-canvas-bootstrap.js` there is a variable `autoRefreshIntervalSeconds` that you'll have to plumb down a few layers starting from [1]. Try setting a breakpoint in the `init` function in `nf-ng-canvas-flow-status-controller.js` and you see how it needs to be plumbed down.

[1] https://github.com/apache/nifi/blob/main/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/nf-canvas-bootstrap.js#L353

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
sardell: I'm still not sure what caused this issue, but I can't reproduce it at all. Perhaps the changes from my last commit, which moves the documentation link to the rule options menu, resolved this issue. Could you test again with the latest when you have a chance, @tpalfy?

sardell: Please disregard my last comment. I was finally able to reproduce the issue (happens when clicking from one rule options menu trigger then immediately clicking on another with no breakpoint set).

sardell: Thanks for pointing this out, @tpalfy. Seems like the issue was an event handler that wasn't removed properly. I've since pushed a fix for the issue.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
sardell: Seems like the issue was an event handler that wasn't removed properly. I've since pushed a fix for the issue.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,554 @@
...
+                        $('#violation-menu').show();
+                        $('#violation-menu').position({
+                            my: "left top",
+                            at: "left top",
+                            of: event
+                        });
+
+                        // violation menu bindings
+                        $('#violation-menu-more-info').on( "click", openRuleMoreInfoDialog);
+                        $('#violation-menu-go-to').on('click', goToComponent);
```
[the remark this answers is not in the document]
sardell: This was caused by an incorrect value being passed for the groupId. Should be fixed now.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
sardell: @mcgilman Based on your comment above, I ended up creating a getter/setter in nf-common and calling the set function in nf-canvas-bootstrap. If the approach you had in mind is much different, let me know and I can update accordingly.

sardell:
@tpalfy I believe I addressed all of the issues you pointed out. When you have time, could you please take another look?

tpalfy:
LGTM
Tested with built-in and some custom rules. All issues raised by me seem to be addressed.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
mcgilman: The change is fine. But I believe it could have been passed down through the `init` functions.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,556 @@
...
+                            }
+                        }
+
+                        function openRuleDetailsDialog() {
+                            $('#rule-menu').hide();
+                            nfSettings.showSettings().done(function() {
+                                nfSettings.selectFlowAnalysisRule(ruleInfo.id);
+                            });
+                            unbindRuleMenuHandling();
+                        }
```
mcgilman, line 716: Users that lack permission to the controller won't be able to view the rules. Currently, the application allows all users to attempt this. If the user lacks permission they will be shown a dialog with a corresponding error message. But I think could we should be able to check this ahead of time and hide this menu item accordingly.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/flow-analysis-drawer.jsp
```diff
@@ -0,0 +1,85 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+--%>
+<%@ page contentType="text/html" pageEncoding="UTF-8" session="false" %>
+<section id="flow-analysis-drawer">
+    <div class="flow-analysis-header">
+        <div class="flow-analysis-refresh-container">
+            <button id="flow-analysis-check-now-btn" class="flow-analysis-check-now-btn">Start a new analysis</button>
```
mcgilman, line 21: When the user clicks this button there is no feedback. With no feedback, I was questioning whether it worked or not. As I repeatedly click on the link, new flow analysis reports seem to be generated. At a minimum, we should probably limit a client from having more than one outstanding request.

Alternatively, should be feature be simplified and the UI only show whatever the current results are? If so, we could just remove this link.

mcgilman (commented):
Thanks for the PR @sardell! I've noted a few items below. For some of them, we should probably get input from others that have worked on this feature as well.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/flow-status.jsp
```diff
@@ -71,6 +71,7 @@
             <button id="search-button" ng-click="appCtrl.serviceProvider.headerCtrl.flowStatusCtrl.search.toggleSearchField();"><i class="fa fa-search"></i></button>
             <input id="search-field" type="text" placeholder="Search"/>
         </div>
+        <button id="flow-analysis" class="flow-analysis"><i class="fa fa-lightbulb-o flow-analysis-notification-icon"></i></button>
```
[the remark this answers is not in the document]
sardell: Good catch. I'll update the background color styling to match the hover/focus states of the search icon.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
sardell: Yeah, great point. I'll update the client side to show/hide that button based on a user's permission to the controller.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,556 @@
...
+                        var violationEl = $('<div class="violation-list-item"></div>');
+                        var violationListItemWrapperEl = $('<div class="violation-list-item-wrapper"></div>');
+                        var violationRuleEl = $('<div class="rule-violations-list-item-name"></div>');
+                        var violationListItemNameEl = $('<div class="violation-list-item-name"></div>');
+                        var violationListItemIdEl = $('<span class="violation-list-item-id"></span>');
+                        var violationInfoButtonEl = $('<button class="violation-menu-btn"><i class="fa fa-ellipsis-v rules-list-item-menu-target" aria-hidden="true"></i></button>');
+
+                        // add text content and button data
+                        $(violationRuleEl).text(rule.name);
+                        $(violationListItemNameEl).text(violation.subjectDisplayName);
```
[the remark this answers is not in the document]
sardell: > In past features, referencing components where the user lacked permissions were rendered as unauthorized

I'm assuming that those past features would handle the permission logic on the backend so details like the name aren't returned in the request response. 

cc @tpalfy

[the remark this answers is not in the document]
tpalfy: @mcgilman this is a good observation. I'm trying to figure out how to solve this. In cluster mode usually the Merger (and in this case the FlowAnalysisResultEntityMerger) actually filters the result based on authorization. But in non-cluster mode I don't see what the usual practice is.
Do you have a suggestion?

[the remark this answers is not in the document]
tpalfy: @mcgilman I have opened https://issues.apache.org/jira/browse/NIFI-12862 and going to open a PR for the backend to fix this issue.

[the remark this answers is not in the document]
tpalfy: @mcgilman I opened a PR https://issues.apache.org/jira/browse/NIFI-12862 that should take care of this issue.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/flow-analysis-drawer.jsp
sardell: I updated my work to disable this button and show a spinner while a report is being run.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,556 @@
...
+                                })).done(function () {});
+                            });
+                            unbindViolationMenuHandling();
+                        }
+
+                        function goToComponent() {
+                            $('#violation-menu').hide();
+                            nfCanvasUtils.showComponent(violationInfo.groupId, violationInfo.subjectId);
+                            unbindViolationMenuHandling();
+                        }
```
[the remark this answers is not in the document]
sardell: I've tested canvas components and those seemed to work fine for me. Controller Services definitely did not work. @tpalfy added this commit[1] in another PR as a way of identifying the component type. I added a client-side check so the Go To Component button is disabled if the violating component is a CS. As a result, [his PR](https://github.com/apache/nifi/pull/8475) will need to be merged first before this one. 

[1] https://github.com/apache/nifi/pull/8475/commits/054c39cf4a1155c5932d84b830deee8ed95a40e1

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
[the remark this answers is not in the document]
sardell: I've updated this PR to show 'Unauthorized' for the fields that will return `null` if a user doesn't have read permission. @tpalfy's PR will need to be merged first as a result.

sardell:
@mcgilman I believe I addressed all your comments. Thanks for the quick assist, @tpalfy!

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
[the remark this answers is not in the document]
sardell: @mcgilman Quick update: @tpalfy's work has been reviewed and merged into `main`. I've rebased this PR to include those changes.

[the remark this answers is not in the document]
mcgilman: @tpalfy Typically we would just report it as unauthorized and clear out component details. Please see the Controller Service references, Parameter references, Parameter Provider references, etc for examples.

mcgilman (commented):
Thanks for the updates @sardell! I've left a few notes below. I'm also still wondering if we need the ability for the user to initiate a new analysis. The front end will poll for the latest every 30 secs. Does the analysis run on the back end less frequently than that? 

I also have some concerns over users with limited permissions being able to submit analysis request. Given the size of the flow or the number rules, they may not be aware of the implication of their request. In my opinion they shouldn't be allow to initiate analysis using rules they aren't allowed to see against components they aren't allowed to use. 

It feels like this should be more aligned with how validation works. The back end is responsible for performing the analysis when appropriate (on an interval or possibly when certain actions occur) and the front end simply fetches the most recent results.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +407,602 @@
...
+                            $('#violation-menu-more-info').on( "click", openRuleViolationMoreInfoDialog);
+                        } else {
+                            $('#violation-menu-more-info').addClass('disabled');
+                            $('#violation-menu-more-info .violation-menu-option-icon').addClass('disabled');
+                        }
+                        
+                        // If the groupId and subjectId are not the same, we can select the component
+                        var isRootGroup = violationInfo.groupId === violationInfo.subjectId;
+                        var isProcessor = violationInfo.subjectComponentType === 'PROCESSOR';
+                        if (!isRootGroup && isProcessor) {
```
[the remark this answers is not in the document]
sardell: I'm also used to null being returned for the root group in other areas of the application. This logic is guidance I received from @tpalfy based on offline discussions we had, but I can't remember the exact reasoning why he suggested this approach. I'll gladly change it to evaluating if null/unset like we're used to, but I'm wondering if @tpalfy remembers why we went this direction.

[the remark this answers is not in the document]
mcgilman: What type of components could be the subject? Since the `subjectId` is being compared to the `groupId` it suggested that a Process Group could be the subject. If a child Process Group were the subject wouldn't the same condition be `true` even though it isn't the root group?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/canvas/flow-analysis-drawer.jsp
```diff
@@ -0,0 +1,84 @@
...
+            <ul>
+                <li class="rule-menu-option" id="rule-menu-view-documentation"><i class="fa fa-info-circle rule-menu-option-icon" aria-hidden="true"></i>View Documentation</li>
+                <li class="rule-menu-option" id="rule-menu-edit-rule"><i class="fa fa-pencil rule-menu-option-icon" aria-hidden="true"></i>Edit Rule</li>
+            </ul>
+        </div>
+
+        <div class="violation-menu" id="violation-menu">
+            <ul>
+                <li class="violation-menu-option" id="violation-menu-more-info"><i class="fa fa-info-circle violation-menu-option-icon" aria-hidden="true"></i>Violation details</li>
+                <li class="violation-menu-option" id="violation-menu-go-to"><i class="fa fa-pencil violation-menu-option-icon" aria-hidden="true"></i>Go to component</li>
```
mcgilman, line 80: Minor but I noticed inconsistent enabled states on the icons in the context menu.

Screenshot 2024-04-12 at 6 57 04 PM

Screenshot 2024-04-12 at 6 57 15 PM

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,562 @@
...
+                            $('#violation-menu-more-info').removeClass('disabled');
+                            $('#violation-menu-more-info .violation-menu-option-icon').removeClass('disabled');
+                            $('#violation-menu-more-info').on( "click", openRuleViolationMoreInfoDialog);
+                        } else {
+                            $('#violation-menu-more-info').addClass('disabled');
+                            $('#violation-menu-more-info .violation-menu-option-icon').addClass('disabled');
+                        }
+                        
+                        var isProcessor = violationInfo.subjectComponentType === 'PROCESSOR';
+                        if (violationInfo.groupId && isProcessor) {
```
mcgilman, line 770: Thanks for updating the `isRootGroup` logic here but I'm still unsure of this condition. A Processor will always have a parent group. Other extension types will not have a parent group or will conditionally have a parent group. If the intent is to only support Go To for Processors then I think we only need to check if it's a Processor.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,562 @@
...
+                            $('#violation-menu-go-to').removeClass('disabled');
+                            $('#violation-menu-go-to .violation-menu-option-icon').removeClass('disabled');
+                            $('#violation-menu-go-to').on('click', goToComponent);
+                        } else {
+                            $('#violation-menu-go-to').addClass('disabled');
+                            $('#violation-menu-go-to .violation-menu-option-icon').addClass('disabled');
```
mcgilman, line 776: I know we often go back and forth with hidden vs disabled but in this case but in this case, we're dealing with something that isn't a Processor and will not support Go To. Should we consider hiding this menu item? There is a class `hidden` that is `display: none;` that should work.

I think disabling makes sense for scenarios when the user lacks permissions (like Edit Rule) but am interested in your thoughts on this one.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
```diff
@@ -400,6 +406,562 @@
...
+
+                        // violation menu bindings
+                        if (violationInfo.subjectPermissionDto.canRead) {
+                            $('#violation-menu-more-info').removeClass('disabled');
+                            $('#violation-menu-more-info .violation-menu-option-icon').removeClass('disabled');
+                            $('#violation-menu-more-info').on( "click", openRuleViolationMoreInfoDialog);
+                        } else {
+                            $('#violation-menu-more-info').addClass('disabled');
+                            $('#violation-menu-more-info .violation-menu-option-icon').addClass('disabled');
+                        }
```
mcgilman, line 767: I'm not sure we need to prevent navigation to a component when the user lacks permissions. If you compare this to other Referencing Component scenarios we allow the user to navigate to them. We just hide the component configuration and details.

Also, in scenario like this we've disabled Violation Details in one menu but not the other even though they are from the same Rule.

Screenshot 2024-04-12 at 7 25 08 PM

mcgilman (commented):
Thanks for the updates @sardell! Just a couple minor things I noted while reviewing.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/canvas/controllers/nf-ng-canvas-flow-status-controller.js
mcgilman: I see. The `violationMessage` is subject dependent. Thanks for clarifying.

mcgilman (approved):
Thanks for the updates @sardell! +1 will merge once CI completes.

PR #8280, opened 2024-01-22
NIFI-12651: Add case-insensitive field retrieval option for Record and Schema

mark-bathori:
# Summary

The idea is to create a separate map in the Schema with lowercase names and a separate method where the ‘fieldName’ parameter is also lowercased to make the retrieval case-insensitive.
The new value map is created with **lazy evaluation** strategy so it’s only generated when there is a method call for it. With this approach we are not storing the same mapping twice.

[NIFI-12651](https://issues.apache.org/jira/browse/NIFI-12651)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/SimpleRecordSchema.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/SimpleRecordSchema.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
@@ -34,6 +35,7 @@ import org.apache.nifi.serialization.record.SchemaIdentifier;
 public class SimpleRecordSchema implements RecordSchema {
     private List<RecordField> fields = null;
     private Map<String, RecordField> fieldMap = null;
+    private Map<String, RecordField> lowerCaseFieldMap = null;
     private boolean textAvailable;
     private final AtomicReference<String> text = new AtomicReference<>();
     private String schemaFormat;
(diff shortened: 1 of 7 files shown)


exceptionfactory (changes requested):
Thanks for proposing this new feature @mark-bathori.

The concept seems useful, but the initial implementation raises some questions. Although the lowercase field map is created lazily, it effectively duplicates the existing field map, which could have notable impacts on memory usage when processing files with hundreds of thousands of records.

Introducing a new `caseInsensitiveGetField` method makes this more accessible, but it does raise a question about whether this should be a first-class API method.

As an alternative approach, one option could be supporting an additional constructor argument to handle fields while ignoring the case sensitivity.

Either way, this seems worth further consideration before going forward.

mark-bathori:
Thanks a lot @exceptionfactory for commenting on this. I'm happy to discuss possible solutions for this task since I was thinking a lot about how should this be implemented properly.

I was also considering the constructor parameter solution, my main concern was that usually when we are creating a `Record` ( e.g. through a `RecordReader`) we don't know how other components would like to use it so we can't determine if it should be created in case-sensitive or insensitive way. Possible solution could be if it is modifiable on the `Record` and maybe track the state with a boolean. 

Possible solution with the current approach could be if we set the original case-sensitive `Map` to be also **lazy-initialized**. This would solve the duplicated storing issue. There could be further restriction added so only one type of map can be initialized at once but both could be supported if needed.

I can also see further improvements in the current `SimpleRecordSchema` implementation in terms of memory usage. Currently it stores the schema fields in a `List` (**fields**) and also in a `Map` (**fieldMap**) where the key is the name of the record field and the value is the `RecordField` object. Since the `RecordField` object is stored twice it could be changed to store the index of the `RecordField` in the field list as the map's value instead of the RecordField itself.

I agree on moving this only to the `SimplRecordeSchema` implementation if we don't want to add this on interface level.

What is your opinion about these?

exceptionfactory (commented):
Thanks for the thoughtful reply @mark-bathori.

Revisiting this discussion, making case-insensitive retrieval a first-class concept raises a number of questions.

In the current form, Records provide a format-agnostic model that maintains a level of fidelity to the original structure. There are some other Map implementation approaches that support case-insensitive lookup, which could be part of a solution.

Perhaps it would be helpful to describe some use cases. I know the Jira references some particular workarounds, but describing those might be a good way to highlight whether or not this is a good candidate for a generalized solution.

mark-bathori:
Thanks for the reply @exceptionfactory. 

For example Kudu([NIFI-6680](https://issues.apache.org/jira/browse/NIFI-6680)) and Iceberg([NIFI-11263](https://issues.apache.org/jira/browse/NIFI-11263)) processor contains some kind of workaround for this. Both of them requires this when the NiFi side record schema and the database schema needs to be matched.

I've checked other Map implementations like Apache’s CaseInsensitiveMap but the problem I found with them is their performance is usually worse than HashMap's and it would lead to unneccesary performance degredation.

I think we can close this PR since we have a working solution currently for every module which needs a case-insensitive field retrieval. This can be reopened if we will have a new feature requiring this kind of functionality in the future.

PR #8294, opened 2024-01-24
[NIFI-12655] theme NiFi following Material Design spec and leveraging Angular Material themes

scottyaslan:
NIFI-12655 initial theming of nifi leveraging Angular Material themes and following the Material Design spec. The Canvas, Codemirror, Angular Material components, as well as our custom components have **ALL** been themed.

To change the theme open the `nifi-web-fronted/pom.xml` and update the theme to 'purple' by setting the `frontend.theme` maven variable to:

`<frontend.theme>purple</frontend.theme>`

Then run a full build with:

`mvn -T 2 install -DskipTests -P\!minify-and-compress,avoid-archive-formats,include-new-ui`

and visit https://localhost:8443/nf/# to see the new purple theme!

To develop a new theme create your theme file in `nifi-web-frontend/src/main/nifi/src/assets/themes/`. Each theme file should define each of the variables for each of the primary, accent, and warn color palettes. Once this new theme file has been created it can be referenced as the value for `frontend.theme`.

The theme may also be changed outside of a full build in order to support fast developer cycles when designing a new theme. To do so `npm run start` and visit the new UI. Notice that the original NiFi theme is still applied even though we built with the 'purple' theme above. This is because the angular dev server is running in the src/ not in the target/ so in this location the 'purple' theme file has not overwritten the 'nifi' theme file. Now, to change to the 'purple' theme in `nifi-web-frontend/src/main/nifi/src/styles.scss` edit line #75 from:

`@import 'assets/themes/nifi';`

to

`@import 'assets/themes/purple';`

or if you have created your own theme file you could include that and see your custom theme applied.

Please note that for this initial effort the birdseye, default label color, and default processor fill color have **not** been included in the theming.


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/access-policies/ui/common/add-tenant-to-policy-dialog/add-tenant-to-policy-dialog.component.html
```diff
@@ -42,7 +42,7 @@ <h2 mat-dialog-title>Add Users/Groups To Policy</h2>
         </div>
     </mat-dialog-content>
     <mat-dialog-actions align="end" *ngIf="{ value: (saving$ | async)! } as saving">
-        <button mat-raised-button mat-dialog-close color="accent">Cancel</button>
+        <button mat-raised-button mat-dialog-close color="primary">Cancel</button>
```
rfellows, line 45: we shouldn't have 2 primary buttons on the same dialog. this one should probably be:
```suggestion
        <button mat-stroked-button mat-dialog-close color="primary">Cancel</button>
```

rfellows (commented):
Looks pretty good to me. Left a couple of comments inline.
The only thing that really stands out to me but I didn't understand how it all works to really comment inline... the text on primary buttons is too dark IMO. Makes the text  hard to read. Can we customize the text color in those scenarios?

scottyaslan:
> Looks pretty good to me. Left a couple of comments inline. The only thing that really stands out to me but I didn't understand how it all works to really comment inline... the text on primary buttons is too dark IMO. Makes the text hard to read. Can we customize the text color in those scenarios?

Let me take a look and see what else we can do here...

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/service/manager/processor-manager.service.ts
```diff
@@ -669,20 +669,20 @@ export class ProcessorManager {
...
                 } else if (d.status.aggregateSnapshot.runStatus === 'Running') {
-                    fill = '#7dc7a0';
+                    clazz = 'accent-200';
                 } else if (d.status.aggregateSnapshot.runStatus === 'Stopped') {
-                    fill = '#d18686';
+                    clazz = 'warn-200';
                 }
 
-                return fill;
+                return clazz;
```
rfellows, line 685: I think this is overwriting the `run-status-icon` class already on this element.
It seems to be preventing the icon from updating when the processor is started/stopped.

Ports have the same issue.

probably want something like:
```
            .attr('class', function (d: any) {
                let clazz: string = 'primary-500';

                if (d.status.aggregateSnapshot.runStatus === 'Validating') {
                    clazz = 'warn-contrast-300';
                } else if (d.status.aggregateSnapshot.runStatus === 'Invalid') {
                    clazz = 'accent-A200';
                } else if (d.status.aggregateSnapshot.runStatus === 'Running') {
                    clazz = 'accent-200';
                } else if (d.status.aggregateSnapshot.runStatus === 'Stopped') {
                    clazz = 'warn-200';
                }

                return `run-status-icon ${clazz}`;
            })
```

scottyaslan: Great catch! I have updated these in my latest commit. Thanks!

rfellows (changes requested):
Found a couple of things:


Create Process Group dialog isn't right. Its inlined, the form elements should be on separate lines:
Screenshot 2024-01-29 at 5 33 56 PM

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/styles.scss
```diff
@@ -122,341 +169,332 @@ $appFontPath: '~roboto-fontface/fonts';
...
-.enabled {
-    color: #44a3cf !important;
-    fill: #44a3cf !important;
-    text-shadow: 0 0 4px rgba(255, 255, 255, 1);
-}
+    .running {
+        color: $accent-palette-200 !important;
+        fill: $accent-palette-200 !important;
+        text-shadow: 0 0 4px $canvas-primary-palette-900;
+    }
```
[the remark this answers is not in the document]
scottyaslan: Ok great catch! I updated the accent palette to be $nifi-canvas-accent-light-palette but didn't update the places we look those up.

scottyaslan:
> Create Process Group dialog isn't right. Its inlined, the form elements should be on separate lines: Screenshot 2024-01-29 at 5 33 56 PM

This is fixed in my latest commit.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/status-history/_status-history.component-theme.scss
```diff
@@ -0,0 +1,63 @@
...
+    $canvas-primary-palette-A700: mat.get-color-from-palette($canvas-primary-palette, 'A700');
+
+    #status-history-chart-container,
+    #status-history-chart-control-container {
+        background-color: $canvas-primary-palette-900;
+
+        .axis path,
+        .axis line {
+            fill: none;
+            stroke: $canvas-primary-palette-A100;
```
rfellows, line 43: This translates into a transparent stroke for the chart axis lines (invisible).

New UI, no x or y axis:
Screenshot 2024-01-30 at 8 34 06 AM


Old UI, both x and y axis
Screenshot 2024-01-30 at 8 34 33 AM

rfellows (commented):
@scottyaslan, I only found one thing. I'm happy if we want to include a fix for it in a future PR. Just don't want to let it get lost.

Let me know how you want to proceed.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/status-history/_status-history.component-theme.scss
rfellows: Let's just come back to this in the future. This PR is too big to keep rebasing over and over.

rfellows (approved):
Excellent work, @scottyaslan !

PR #8305, opened 2024-01-26
NIFI-12663: Error Handling in CS Listing

mcgilman:
NIFI-12663:
- Handling API error responses in the Management Controller Services page.

The following items are covered in this PR
1) Snackbar errors for management service creation or deletion failure.
2) Banner errors in edit management service dialog. (enable/disable not covered in this PR)
3) Full screen error when initially loading management service listing.
4) Full screen error with missing or expired token.


rfellows:
will review...

rfellows (changes requested):
There isn't a great place for this comment, so i'll add it generally.
If the call to get the property descriptor fails while creating a new property, nothing happens and the new property dialog stays visible.

It seems the source of this is in the property-table-helper where we aren't handling any errors from the service call. We might want to consider adding some. While a bit crude, this seems to at least open the snackbar with the error and closes the dialog:

```
                            catchError((e) => {
                                this.store.dispatch(
                                    snackBarError({
                                        error: e.error
                                    })
                                );
                                newPropertyDialogReference.close();
                                return NEVER;
                            })
```

this should go in the pipe after making the call to `propertyDescriptorService.getPropertyDescriptor`.

The same issue exists when creating a new inline service if something fails. That one has 2 service calls, but adding the catch on the outer one handles a failure in both:

```
                                    catchError((e) => {
                                        this.store.dispatch(
                                            snackBarError({
                                                error: e.error
                                            })
                                        );
                                        createServiceDialogReference.close();
                                        return NEVER;
                                    })
```

mcgilman:
Thanks for the review @rfellows! I hadn't targeted the property descriptor and new inline service error handling but it makes sense to address it now since it's exposed through the edit service dialog. I've already addressed a few of the items above but still need to consider the descriptor and inline service concerns. Will comment again once it's all ready for re-review.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/page-content/page-content.component.scss
```diff
@@ -15,7 +15,9 @@
  * limitations under the License.
  */
 
-.message {
+.page-content {
+    //width: 500px;
```
rfellows, line 19: can this be removed?

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/service/property-table-helper.service.ts
```diff
@@ -153,6 +180,20 @@ export class PropertyTableHelperService {
...
+                                                tap({
+                                                    error: (errorResponse: HttpErrorResponse) => {
+                                                        // we've errored getting the descriptor but since the service
+                                                        // was already created, we should close the create service dialog
+                                                        // so multiple service instances are not inadvertently created
+                                                        createServiceDialogReference.close();
+
+                                                        this.store.dispatch(
+                                                            snackBarError({
+                                                                error: `Service created but until to reload Property Descriptor: ${errorResponse.error}`
```
rfellows, line 192: I'm a bit confused by this wording. Can we clarify it a bit more?

rfellows (approved):
Changes look good. Thanks @mcgilman. I'll let CI do its thing and then will merge.

PR #8306, opened 2024-01-26
NIFI-12679: Prompt to copy policy during override

mcgilman:
NIFI-12679:
- When overriding a policy prompting whether the new policy should copy the users and groups or if it should be empty.


rfellows (approved):
Thanks, @mcgilman.

PR #8313, opened 2024-01-30
[NIFI-12686] eslint configuration

rfellows:
also added linting to the default maven build. errors will make the build fail, warning will not.

[NIFI-12686](https://issues.apache.org/jira/browse/NIFI-12686)


mcgilman:
Will review...

PR #8319, opened 2024-01-30
NIFI-12697: Allow disabling scientific notation when writing JSON; us…

markap14:
…e Jackson's internal implementations for parsing big integers and floating-point numbers more efficiently

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/WriteJsonResult.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/WriteJsonResult.java
@@ -19,6 +19,7 @@ package org.apache.nifi.json;
 
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonGenerator.Feature;
 import com.fasterxml.jackson.core.util.MinimalPrettyPrinter;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.nifi.NullSuppression;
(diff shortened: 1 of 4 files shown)


Lehel44:
Thanks @markap14 for making this. I went through the code and unit tests, as well did some manual testing and verifying property migrations. The changes made to the JsonWriter appear to function as intended.

One concern is that while the changes to the JsonWriter fixed the scientific notation problem, the RecordReader still uses scientific notation by default. This mismatch could cause unexpected issues when using both components together. In my tests, I noticed that the `WriteJsonResult::writeRecord` received records in scientific notation, which might not be what users expect. To avoid confusion and ensure consistency, it would be wise to update the RecordReader to match the changes made to the JsonWriter. What do you think?

markap14:
@Lehel44 I'm not sure that I understand the concern. Can you provide a concrete use case? The JSON Reader should be able to read the JSON regardless of whether the incoming data is in scientific notation or not.

nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/WriteJsonResult.java
```diff
@@ -19,6 +19,7 @@
```
Lehel44, line 19: I think in the section of WriteJsonResult from line 176 to 189, the scientific notation logic needs to be added for cases where the Record is already serialized.

Lehel44 (changes requested):

nifi-nar-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/WriteJsonResult.java
markap14: Got it. Thanks @Lehel44 that makes sense. Great catch! Will update shortly, with new test cases.

PR #8321, opened 2024-01-30
NIFI-12684: Handling error responses in the service listing on the canvas

mcgilman:
NIFI-12684:
- Handling error responses in the service listing on the canvas.
- Handling error responses when converting a Property value to a Parameter.


rfellows (approved):
Looks good. Thanks @mcgilman

PR #8332, opened 2024-02-01
NIFI-12710: Allow for microsecond precision timestamps when pushing t…

markap14:
…o database

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
@@ -21,7 +21,6 @@ import org.apache.nifi.serialization.record.util.IllegalTypeConversionException;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
-import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.Date;
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
@@ -21,6 +21,7 @@ import org.apache.nifi.serialization.record.util.IllegalTypeConversionException;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.Date;
@@ -30,6 +31,8 @@ import java.util.Optional;
  * Convert Object to java.time.LocalDateTime using instanceof evaluation and optional format pattern for DateTimeFormatter
  */
 class ObjectLocalDateTimeFieldConverter implements FieldConverter<Object, LocalDateTime> {
+    private static final long YEAR_TEN_THOUSAND = 253_402_300_800_000L;
+
     /**
      * Convert Object field to java.sql.Timestamp using optional format supported in DateTimeFormatter
      *
@@ -81,14 +84,14 @@ class ObjectLocalDateTimeFieldConverter implements FieldConverter<Object, LocalD
     }
 
     private LocalDateTime tryParseAsNumber(final String value, final String fieldName) {
-        // If decimal, treat as a double and convert to seconds and nanoseconds.
-        if (value.contains(".")) {
-            final double number = Double.parseDouble(value);
-            return toLocalDateTime(number);
-        }
-
-        // attempt to parse as a long value
         try {
+            // If decimal, treat as a double and convert to seconds and nanoseconds.
+            if (value.contains(".")) {
+                final double number = Double.parseDouble(value);
+                return toLocalDateTime(number);
+            }
+
+            // attempt to parse as a long value
             final long number = Long.parseLong(value);
             return toLocalDateTime(number);
         } catch (final NumberFormatException e) {
@@ -98,10 +101,10 @@ class ObjectLocalDateTimeFieldConverter implements FieldConverter<Object, LocalD
 
     private LocalDateTime toLocalDateTime(final double secondsSinceEpoch) {
         // Determine the number of micros past the second by subtracting the number of seconds from the decimal value and multiplying by 1 million.
-        final long micros = (long) (1_000_000 * (secondsSinceEpoch - (long) secondsSinceEpoch));
+        final double micros = 1_000_000 * (secondsSinceEpoch - (long) secondsSinceEpoch);
         // Convert micros to nanos. Note that we perform this as a separate operation, rather than multiplying by 1_000,000,000 in order to avoid
         // issues that occur with rounding at high precision.
-        final long nanos = micros * 1000L;
+        final long nanos = (long) micros * 1000L;
 
         return toLocalDateTime((long) secondsSinceEpoch, nanos);
     }
@@ -112,13 +115,14 @@ class ObjectLocalDateTimeFieldConverter implements FieldConverter<Object, LocalD
     }
 
     private LocalDateTime toLocalDateTime(final long value) {
-        final Instant instant = Instant.ofEpochMilli(value);
-        final LocalDateTime localDateTime = ofInstant(instant);
-        if (localDateTime.getYear() > 10_000) {
+        if (value > YEAR_TEN_THOUSAND) {
             // Value is too large. Assume microseconds instead of milliseconds.
             final Instant microsInstant = Instant.ofEpochSecond(value / 1_000_000, (value % 1_000_000) * 1_000);
             return ofInstant(microsInstant);
         }
+
+        final Instant instant = Instant.ofEpochMilli(value);
+        final LocalDateTime localDateTime = ofInstant(instant);
 
         return localDateTime;
     }
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
@@ -54,9 +54,9 @@ class ObjectLocalDateTimeFieldConverter implements FieldConverter<Object, LocalD
         if (field instanceof final Number number) {
             // If value is a floating point number, we consider it as seconds since epoch plus a decimal part for fractions of a second.
             if (field instanceof Double || field instanceof Float) {
-                // Use long value as seconds; use everything after it as fractions of a second and multiply by 1_000_000_000 to get nanoseconds.
-                return toLocalDateTime(number.longValue(), (long) ((number.doubleValue() - number.longValue()) * 1_000_000_000));
+                return toLocalDateTime(number.doubleValue());
             }
+
             return toLocalDateTime(number.longValue());
         }
         if (field instanceof String) {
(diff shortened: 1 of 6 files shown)


exceptionfactory (changes requested):
Thanks for working on this improvement to timestamp handling @markap14.

The new integration test has several failures related to the precision of fractional seconds in compared values.

```
Error:  org.apache.nifi.processors.standard.PutDatabaseRecordIT.testWithNumericTimestampUsingMicros -- Time elapsed: 0.044 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <2024-02-01T03:06:17.555052498Z> but was: <2024-02-01T03:06:17.555052Z>
```

It looks like the Timestamp returned from PostgreSQL has microsecond precision, but the expected Instant has nanosecond precision.

Related to the above, it would be helpful to add some unit tests for the Field Converter itself to ensure that the new number parsing strategies work as expected at that level.

markap14:
Thanks @exceptionfactory have addressed.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/PutDatabaseRecordIT.java
```diff
@@ -0,0 +1,312 @@
...
+              "dob": "1975-01-01"
+            }
+            """);
+        runner.run();
+        runner.assertAllFlowFilesTransferred(PutDatabaseRecord.REL_SUCCESS, 1);
+
+
+        final Map<String, Object> results = getResults();
+        assertEquals("John Doe", results.get("name"));
+        assertEquals(50, results.get("age"));
```
exceptionfactory, line 133: These comparisons are repeated in most of the test methods, along with the same data. It looks like the test methods could be refactored so that the shared behavior is not repeated. That would make this test more maintainable, avoid some of the duplication, and highlight what is unique about the particular test method (Date, Timestamp, Milliseconds, Microseconds, etc).

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/TestObjectLocalDateTimeFieldConverter.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    public void testConvertTimestampMillis() {
+        final ObjectLocalDateTimeFieldConverter converter = new ObjectLocalDateTimeFieldConverter();
+        final long now = System.currentTimeMillis();
+        final Instant instant = Instant.ofEpochMilli(now);
+        final LocalDateTime expected = LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
+
+        final LocalDateTime result = converter.convertField(now, Optional.empty(), "test");
+        assertEquals(expected, result);
+        assertEquals(now, result.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli());
```
exceptionfactory, line 40: Is there a particular reason for this comparison after comparing the expected value above? Converting to a different representation is effectively mutating the value, versus testing the `LocalDateTime` conversion on its own.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/TestObjectLocalDateTimeFieldConverter.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Test
+    public void testDoubleAsEpochSecondsAsString() {
+        final ObjectLocalDateTimeFieldConverter converter = new ObjectLocalDateTimeFieldConverter();
+        final long millis = System.currentTimeMillis();
+        final double seconds = millis / 1000.0;
+        final double withMicros =  seconds + 0.000567;
+
+        final LocalDateTime result = converter.convertField(Double.toString(withMicros), Optional.empty(), "test");
+        assertEquals(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneId.systemDefault()).toLocalDate(), result.toLocalDate());
```
exceptionfactory, line 78: This comparison is a bit difficult to follow, but it appears to be comparing only the `LocalDate` portion of `LocalDateTime`. Is there a reason for this comparison as opposed to the nanosecond comparison below?

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
```diff
@@ -51,10 +51,13 @@ public LocalDateTime convertField(final Object field, final Optional<String> pat
...
         }
-        if (field instanceof Number) {
-            final Number number = (Number) field;
-            final Instant instant = Instant.ofEpochMilli(number.longValue());
-            return ofInstant(instant);
+        if (field instanceof final Number number) {
+            // If value is a floating point number, we consider it as seconds since epoch plus a decimal part for fractions of a second.
+            if (field instanceof Double || field instanceof Float) {
+                // Use long value as seconds; use everything after it as fractions of a second and multiply by 1_000_000_000 to get nanoseconds.
+                return toLocalDateTime(number.longValue(), (long) ((number.doubleValue() - number.longValue()) * 1_000_000_000));
```
exceptionfactory, line 58: Recommend breaking out nanosecond assignment to a separate line for readability, as it is a bit difficult to follow the levels of nested calculation.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
```diff
@@ -67,22 +70,49 @@ public LocalDateTime convertField(final Object field, final Optional<String> pat
...
         }
 
         throw new FieldConversionException(LocalDateTime.class, field, name);
     }
 
+    private LocalDateTime tryParseAsNumber(final String value, final String fieldName) {
+        // If decimal, treat as a double and convert to seconds and nanoseconds.
+        if (value.contains(".")) {
+            final double number = Double.parseDouble(value);
+            return toLocalDateTime((long) number, (long) ((number - (long) number) * 1_000_000_000));
```
exceptionfactory, line 87: Similar to above, recommend assigning nanosPastSecond separately for readability.

exceptionfactory (changes requested):
Thanks for the updates @markap14. The basic implementation looks good.

Dealing with date and time is challenging, even with the more recent Java APIs. For this reason, I made some more detailed comments on the test classes, as keeping them straightforward will be helpful for future maintenance.

The variety of test methods for the FieldConverter is great, but I recommend minimizing the amount of conversion necessary to perform the comparison, as it can lead to challenges trying to untangle uncertain behavior. For example, avoiding and minimizing conversion to other types helps avoid failures related to incorrect test code, versus incorrect conversion code. Glad to take a closer look and contribute some changes if that would be helpful.

markap14:
Thanks for the review @exceptionfactory @dan-s1 . Agreed, the tests needed some cleanup. I actually simplified them significantly by just avoiding the use of the current time and trying to manipulate it into the format desired and instead just create predefined timestamps. This is simpler, cleaner, and far less error prone.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
```diff
@@ -67,22 +70,59 @@ public LocalDateTime convertField(final Object field, final Optional<String> pat
...
             }
         }
 
         throw new FieldConversionException(LocalDateTime.class, field, name);
     }
 
+    private LocalDateTime tryParseAsNumber(final String value, final String fieldName) {
+        // If decimal, treat as a double and convert to seconds and nanoseconds.
+        if (value.contains(".")) {
+            final double number = Double.parseDouble(value);
```
exceptionfactory, line 86: The `Double.parseDouble()` method could throw a `NumberFormatException`, similar to below with `Long.parseLong()`, so it would be helpful to have the same try-catch that throws a `FieldConversionException` for troubleshooting. For example, if the string contains two period characters, instead of one, which would be a good test case.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
```diff
@@ -67,22 +70,59 @@ public LocalDateTime convertField(final Object field, final Optional<String> pat
...
+    }
+
+    private LocalDateTime toLocalDateTime(final double secondsSinceEpoch) {
+        // Determine the number of micros past the second by subtracting the number of seconds from the decimal value and multiplying by 1 million.
+        final long micros = (long) (1_000_000 * (secondsSinceEpoch - (long) secondsSinceEpoch));
+        // Convert micros to nanos. Note that we perform this as a separate operation, rather than multiplying by 1_000,000,000 in order to avoid
+        // issues that occur with rounding at high precision.
+        final long nanos = micros * 1000L;
+
+        return toLocalDateTime((long) secondsSinceEpoch, nanos);
```
exceptionfactory, line 106: Given the multiple casts to `long` for `secondsSinceEpoch`, it seems like it would help readability to declare that separately and reuse the value.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
```diff
@@ -67,22 +70,59 @@ public LocalDateTime convertField(final Object field, final Optional<String> pat
...
+
+    private LocalDateTime toLocalDateTime(final long epochSeconds, final long nanosPastSecond) {
+        final Instant instant = Instant.ofEpochSecond(epochSeconds).plusNanos(nanosPastSecond);
+        return ofInstant(instant);
+    }
+
+    private LocalDateTime toLocalDateTime(final long value) {
+        final Instant instant = Instant.ofEpochMilli(value);
+        final LocalDateTime localDateTime = ofInstant(instant);
+        if (localDateTime.getYear() > 10_000) {
```
exceptionfactory, line 117: Instead of comparing the year value after the conversion, what about checking that the initial `long value` is greater than a certain number equivalent to 10K years in the future? That would avoid multiple calls to `ofInstant()` in the case of handling microseconds.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/TestObjectLocalDateTimeFieldConverter.java
```diff
@@ -0,0 +1,87 @@
...
+    private final Instant INSTANT_MICROS_PRECISION = Instant.ofEpochMilli(MILLIS_TIMESTAMP_LONG).plusNanos(NANOS_AFTER_SECOND).minusMillis(MILLIS_TIMESTAMP_LONG % 1000);
+    private final LocalDateTime LOCAL_DATE_TIME_MILLIS_PRECISION = LocalDateTime.ofInstant(INSTANT_MILLIS_PRECISION, ZoneId.systemDefault());
+    private final LocalDateTime LOCAL_DATE_TIME_MICROS_PRECISION = LocalDateTime.ofInstant(INSTANT_MICROS_PRECISION, ZoneId.systemDefault());
+
+    private final ObjectLocalDateTimeFieldConverter converter = new ObjectLocalDateTimeFieldConverter();
+
+
+    @Test
+    public void testConvertTimestampMillis() {
+        final LocalDateTime result = converter.convertField(MILLIS_TIMESTAMP_LONG, Optional.empty(), "test");
```
exceptionfactory, line 47: Just a minor bit, but recommend declaring a static `FIELD_NAME` value and reusing across all test methods.

exceptionfactory (changes requested):
Thanks for updating the tests @markap14, they are now much easier to follow with the expected values.

I noted one more edge case handling question for string parsing, and a couple minor recommendations for clarity, but otherwise this looks close to completion.

exceptionfactory (approved):
Thanks for working through the feedback and making the adjustments @markap14, and thanks for the additional feedback @dan-s1. The latest version looks good! +1 merging

PR #8350, opened 2024-02-02
NIFI-12708 UnpackContent should allow the user to specify a character set to apply in reading paths and filenames

umarhussain15:
# Summary

[NIFI-12708](https://issues.apache.org/jira/browse/NIFI-12708): add option in UnpackContent to specify encoding charset for filenames in zip unpacking

The processor can now take a filename encoding parameter and pass it to zip unpacking. This will allow
user to unzip files with specific encoding to get correct filenames in output.

This for example help with zip files created on Windows which by default uses Cp437 for filename encoding.
If the filename contains special character like German alphabet ä, ü etc., decoding this with Linux's
default encoding (usually UTF8) will contain `?` in output. When the same file is processed with property
set with `Cp437`, the processor outputs correct filenames with special characters preserved.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
```diff
@@ -139,6 +145,20 @@ public class UnpackContent extends AbstractProcessor {
                     PackageFormat.FLOWFILE_STREAM_FORMAT_V2.toString(), PackageFormat.FLOWFILE_TAR_FORMAT.toString())
             .defaultValue(PackageFormat.AUTO_DETECT_FORMAT.toString())
             .build();
+    public static final PropertyDescriptor ZIP_FILENAMES_ENCODING = new PropertyDescriptor.Builder()
+            .name("zip-filenames-encoding")
```
joewitt, line 149: Consider just 'filename-charset' for (name) and 'Filename Character Set' for Display name.

This only applies when it is zip format selection so no need to add 'zip' to the name as the context for use is already established.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
```diff
@@ -139,6 +145,20 @@ public class UnpackContent extends AbstractProcessor {
...
+            .description(
+                "The encoding used by zip creating utility, for the file names inside the zip. Processor will pass this encoding to Zip unpacker. For example 'Cp437', 'UTF8' etc. Default is to "
+                    + "use platform's encoding. This can be useful for example incase a zip was created on Windows with Cp437 and unpacked on Linux machine. Without correct encoding value special "
+                    + "characters in filenames will be outputted as `?`")
+            .required(false)
+            .dependsOn(
+                PACKAGING_FORMAT,
+                PackageFormat.ZIP_FORMAT.toString(),
+                PackageFormat.AUTO_DETECT_FORMAT.toString())
+            .addValidator(new CharsetStringValidator())
```
joewitt, line 160: Consider changing this to StandardValidators.CHARACTER_SET_VALIDATOR and removing the custom validator created below.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
```diff
@@ -139,6 +145,20 @@ public class UnpackContent extends AbstractProcessor {
                     PackageFormat.FLOWFILE_STREAM_FORMAT_V2.toString(), PackageFormat.FLOWFILE_TAR_FORMAT.toString())
             .defaultValue(PackageFormat.AUTO_DETECT_FORMAT.toString())
             .build();
+    public static final PropertyDescriptor ZIP_FILENAMES_ENCODING = new PropertyDescriptor.Builder()
+            .name("zip-filenames-encoding")
+            .displayName("Zip Filenames Encoding")
+            .description(
+                "The encoding used by zip creating utility, for the file names inside the zip. Processor will pass this encoding to Zip unpacker. For example 'Cp437', 'UTF8' etc. Default is to "
```
joewitt, line 152: Consider alternative text

'If supplied this character set will be supplied to the Zip utility to attempt to decode filenames using the specific character set.  If not specified the default platform character set will be used.  This is useful if a Zip was created with a different character set than the platform default and the zip uses non standard values to specify.

joewitt: Instead of describing the Cp437 scenario in the property description consider creating an '@UseCase' annotation to describe the specific case and how the whole processor would be configured.  See 'PublishSlack' for an example of this done well.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
```diff
@@ -353,7 +374,7 @@ public void unpack(final ProcessSession session, final FlowFile source, final Li
                 int fragmentCount = 0;
                 try (final TarArchiveInputStream tarIn = new TarArchiveInputStream(new BufferedInputStream(inputStream))) {
                     TarArchiveEntry tarEntry;
-                    while ((tarEntry = tarIn.getNextTarEntry()) != null) {
+                    while ((tarEntry = tarIn.getNextEntry()) != null) {
```
joewitt, line 377: Is there some documentation you can point to suggesting changing this method call is preferable?  also changing this tar one here given this is about zip changes is not ideal.  Generally fine but if there is a clear value to doing so

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
```diff
@@ -473,24 +495,28 @@ protected void processEntry(final InputStream zipInputStream, final boolean dire
...
             }
 
             @Override
             public void process(final InputStream inputStream) throws IOException {
-                try (final ZipArchiveInputStream zipInputStream = new ZipArchiveInputStream(new BufferedInputStream(inputStream), null, true, allowStoredEntriesWithDataDescriptor)) {
+                try (final ZipArchiveInputStream zipInputStream = new ZipArchiveInputStream(new BufferedInputStream(inputStream),
+                    filenameEncoding.toString(), true, allowStoredEntriesWithDataDescriptor)) {
                     ZipArchiveEntry zipEntry;
-                    while ((zipEntry = zipInputStream.getNextZipEntry()) != null) {
+                    while ((zipEntry = zipInputStream.getNextEntry()) != null) {
```
joewitt, line 519: Same ask as in the tar case.  Why change the method call here?

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
```diff
@@ -643,4 +672,20 @@ public static PackageFormat getFormat(String textValue) {
             };
         }
     }
+
```
joewitt, line 675: Can remove this block as per above comment on standard validator available.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
```diff
@@ -139,6 +145,20 @@ public class UnpackContent extends AbstractProcessor {
...
+                "The encoding used by zip creating utility, for the file names inside the zip. Processor will pass this encoding to Zip unpacker. For example 'Cp437', 'UTF8' etc. Default is to "
+                    + "use platform's encoding. This can be useful for example incase a zip was created on Windows with Cp437 and unpacked on Linux machine. Without correct encoding value special "
+                    + "characters in filenames will be outputted as `?`")
+            .required(false)
+            .dependsOn(
+                PACKAGING_FORMAT,
+                PackageFormat.ZIP_FORMAT.toString(),
+                PackageFormat.AUTO_DETECT_FORMAT.toString())
+            .addValidator(new CharsetStringValidator())
+            .build();
```
dan-s1, line 161: Add a default value so can simplify code on lines 255-257

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestUnpackContent.java
```diff
@@ -222,6 +225,88 @@ public void testInvalidZip() throws IOException {
             flowFile.assertContentEquals(path.toFile());
         }
     }
+    @Test
+    public void testZipEncodingField() {
```
joewitt, line 229: Super wonderful that you created a unit test to cover this and provided a simple example zip to do so!  But the checks will likely fail because the zip file doesn't have a license header so I believe you'll need to add it to an excludes list for the RAT process to not try to extract a header.  You'll see other examples in some pom files for such exclusions.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
```diff
@@ -231,7 +252,10 @@ public void onScheduled(ProcessContext context) throws ProcessException {
             }
             final PropertyValue allowStoredEntriesWithDataDescriptorVal = context.getProperty(ALLOW_STORED_ENTRIES_WITH_DATA_DESCRIPTOR);
             final boolean allowStoredEntriesWithDataDescriptor = allowStoredEntriesWithDataDescriptorVal.isSet() ? allowStoredEntriesWithDataDescriptorVal.asBoolean() : false;
-            zipUnpacker = new ZipUnpacker(fileFilter, password, allowStoredEntriesWithDataDescriptor);
+            final PropertyValue filenamesEncodingVal = context.getProperty(ZIP_FILENAMES_ENCODING);
+
+            Charset filenamesEncoding =filenamesEncodingVal.isSet() ? Charsets.toCharset(filenamesEncodingVal.getValue()) : Charset.defaultCharset();
```
dan-s1, line 257: See prior comment
```suggestion
            final String filenamesEncodingVal = context.getProperty(ZIP_FILENAMES_ENCODING).getValue();
            final Charset filenamesEncoding = Charsets.toCharset(filenamesEncodingVal);
```

joewitt:
@umarhussain15 Great start here.  Left you a series of comments plus Dan just left one to consider as well.  Thanks!

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
exceptionfactory: Actually, using `Filename Character Set` for both the name and display name would be optimal.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UnpackContent.java
umarhussain15: I will revert these method name changes

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestUnpackContent.java
```diff
@@ -222,6 +225,88 @@ public void testInvalidZip() throws IOException {
...
+
+        final List<MockFlowFile> unpacked =
+            unpackRunner.getFlowFilesForRelationship(UnpackContent.REL_SUCCESS);
+        for (final MockFlowFile flowFile : unpacked) {
+            final String filename = flowFile.getAttribute(CoreAttributes.FILENAME.key());
+            assertTrue(StringUtils.containsNone(filename, "?"), "filename contains '?': " + filename);
+            assertTrue(StringUtils.containsNone(filename, "�"), "filename contains '�': " + filename);
+            final String path = flowFile.getAttribute(CoreAttributes.PATH.key());
+            assertTrue(StringUtils.containsNone(path, "?"), "path contains '?': " + path);
+            assertTrue(StringUtils.containsNone(path, "�"), "path contains '�': " + path);
```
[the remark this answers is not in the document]
umarhussain15: Thanks. I have updated the test cases to generate zip with known Unicode character and to look for it in the output filename.
The test case with Windows zip file now only look for `?` in test case as the input contains both files with special character in names and without any special character in name. Also, UnpackContent is always outputting `?` for this zip file (the original issue reported in the ticket).

exceptionfactory (approved):

joewitt:
Thanks @umarhussain15 - good work and good discussion

PR #8358, opened 2024-02-06
NIFI-12743: Property Table Fixes

mcgilman:
NIFI-12743: 
- Updating property editing so that Shift-Enter inserts a new line and Enter commits the edit.
- Fixing bug that prevented deleting a Property and then re-adding a new Property with the same name.


rfellows (approved):
Looks good to me. Thanks @mcgilman

PR #8362, opened 2024-02-06
NIFI-12674 Modified ValidateCSV to make the schema optional if a head…

Freedom9339:
…er is provided. Added validate on attribute option.

# Summary

[NIFI-12674](https://issues.apache.org/jira/browse/NIFI-12674)

Made the schema optional for the ValidateCSV processor if a header is provided. In this case, only the structure of the CSV will validated, using the header to determine how many fields each line should have. 

Additionally, a new validation strategy was implemented, Validate on Attribute. This works similar to the way ValidateXML works in which the value of a given attribute of a FlowFile will be treated as the contents of a CSV file. Validation will be done on that attribute and not on the content of the FlowFile.

I would also like to note that I made the stream a variable that can be assigned to either the FlowFile content or the value of the attribute. In doing this, I removed the need for an inner method, and thus all of the variables that previously needed to be Atomic References could now be regular variables. 

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
@@ -100,7 +100,6 @@ public class ValidateCsv extends AbstractProcessor {
 
     private static final String ROUTE_WHOLE_FLOW_FILE = "FlowFile validation";
     private static final String ROUTE_LINES_INDIVIDUALLY = "Line by line validation";
-    private static final String ROUTE_BY_ATTRIBUTE = "Attribute validation";
 
     public static final AllowableValue VALIDATE_WHOLE_FLOWFILE = new AllowableValue(ROUTE_WHOLE_FLOW_FILE, ROUTE_WHOLE_FLOW_FILE,
             "As soon as an error is found in the CSV file, the validation will stop and the whole flow file will be routed to the 'invalid'"
@@ -112,15 +111,12 @@ public class ValidateCsv extends AbstractProcessor {
                     + "the incorrect lines. Take care if choosing this option while using Unique cell processors in schema definition:"
                     + "the first occurrence will be considered valid and the next ones as invalid.");
 
-    public static final AllowableValue VALIDATE_ATTRIBUTE_AS_CSV = new AllowableValue(ROUTE_BY_ATTRIBUTE, ROUTE_BY_ATTRIBUTE,
-            "Validation will be done on the specified attribute of the FlowFile. The attribute will be treated as CSV text. ");
-
     public static final PropertyDescriptor SCHEMA = new PropertyDescriptor.Builder()
             .name("validate-csv-schema")
             .displayName("Schema")
             .description("The schema to be used for validation. Is expected a comma-delimited string representing the cell "
                     + "processors to apply. The following cell processors are allowed in the schema definition: "
-                    + ALLOWED_OPERATORS + ". Note: cell processors cannot be nested except with Optional.")
+                    + ALLOWED_OPERATORS + ". Note: cell processors cannot be nested except with Optional. Schema is required if Header is false.")
             .required(false)
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .addValidator(StandardValidators.NON_EMPTY_EL_VALIDATOR)
@@ -200,6 +196,7 @@ public class ValidateCsv extends AbstractProcessor {
 
     private static final List<PropertyDescriptor> PROPERTIES = List.of(
             SCHEMA,
+            CSV_SOURCE_ATTRIBUTE,
             HEADER,
             DELIMITER_CHARACTER,
             QUOTE_CHARACTER,
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
@@ -126,6 +126,15 @@ public class ValidateCsv extends AbstractProcessor {
             .addValidator(StandardValidators.NON_EMPTY_EL_VALIDATOR)
             .build();
 
+    public static final PropertyDescriptor CSV_SOURCE_ATTRIBUTE = new PropertyDescriptor.Builder()
+            .name("CSV Source Attribute")
+            .displayName("CSV Source Attribute")
+            .description("The name of the attribute containing CSV data to be validated. If this property is blank, the FlowFile content will be validated.")
+            .required(false)
+            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
+            .addValidator(StandardValidators.ATTRIBUTE_KEY_VALIDATOR)
+            .build();
+
     public static final PropertyDescriptor HEADER = new PropertyDescriptor.Builder()
             .name("validate-csv-header")
             .displayName("Header")
@@ -172,16 +181,7 @@ public class ValidateCsv extends AbstractProcessor {
             .description("Strategy to apply when routing input files to output relationships.")
             .required(true)
             .defaultValue(VALIDATE_WHOLE_FLOWFILE.getValue())
-            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE, VALIDATE_ATTRIBUTE_AS_CSV)
-            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
-            .build();
-
-    public static final PropertyDescriptor VALIDATION_ATTRIBUTE = new PropertyDescriptor.Builder()
-            .name("validate-csv-attribute")
-            .displayName("Validation attribute")
-            .description("FlowFile attribute to validate. The value of this attribute will be treated as CSV text.")
-            .required(true)
-            .dependsOn(VALIDATION_STRATEGY, VALIDATE_ATTRIBUTE_AS_CSV)
+            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
 
@@ -216,6 +216,7 @@ public class ValidateCsv extends AbstractProcessor {
             .name("invalid")
             .description("FlowFiles that are not valid according to the specified schema are routed to this relationship")
             .build();
+
 
     private static final Set<Relationship> RELATIONSHIPS = Set.of(
             REL_VALID,
@@ -491,8 +492,8 @@ public class ValidateCsv extends AbstractProcessor {
         }
 
         InputStream stream;
-        if (validationStrategy.equals(VALIDATE_ATTRIBUTE_AS_CSV.getValue())) {
-            stream = new ByteArrayInputStream(flowFile.getAttribute(context.getProperty(VALIDATION_ATTRIBUTE).getValue()).getBytes(StandardCharsets.UTF_8));
+        if (context.getProperty(CSV_SOURCE_ATTRIBUTE).isSet()) {
+            stream = new ByteArrayInputStream(flowFile.getAttribute(context.getProperty(CSV_SOURCE_ATTRIBUTE).getValue()).getBytes(StandardCharsets.UTF_8));
         } else {
             stream = session.read(flowFile);
         }
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
@@ -36,8 +36,6 @@ import org.apache.nifi.processor.AbstractProcessor;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
-import org.apache.nifi.processor.io.InputStreamCallback;
-import org.apache.nifi.processor.io.OutputStreamCallback;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.supercsv.cellprocessor.Optional;
 import org.supercsv.cellprocessor.ParseBigDecimal;
@@ -67,11 +65,12 @@ import org.supercsv.io.CsvListReader;
 import org.supercsv.prefs.CsvPreference;
 import org.supercsv.util.CsvContext;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.io.OutputStream;
 import java.io.Reader;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -98,8 +97,10 @@ public class ValidateCsv extends AbstractProcessor {
             "Strlen", "StrMinMax", "StrNotNullOrEmpty", "StrRegEx", "Unique", "UniqueHashCode", "IsIncludedIn"
     );
 
+
     private static final String ROUTE_WHOLE_FLOW_FILE = "FlowFile validation";
     private static final String ROUTE_LINES_INDIVIDUALLY = "Line by line validation";
+    private static final String ROUTE_BY_ATTRIBUTE = "Attribute validation";
 
     public static final AllowableValue VALIDATE_WHOLE_FLOWFILE = new AllowableValue(ROUTE_WHOLE_FLOW_FILE, ROUTE_WHOLE_FLOW_FILE,
             "As soon as an error is found in the CSV file, the validation will stop and the whole flow file will be routed to the 'invalid'"
@@ -111,13 +112,16 @@ public class ValidateCsv extends AbstractProcessor {
                     + "the incorrect lines. Take care if choosing this option while using Unique cell processors in schema definition:"
                     + "the first occurrence will be considered valid and the next ones as invalid.");
 
+    public static final AllowableValue VALIDATE_ATTRIBUTE_AS_CSV = new AllowableValue(ROUTE_BY_ATTRIBUTE, ROUTE_BY_ATTRIBUTE,
+            "Validation will be done on the specified attribute of the FlowFile. The attribute will be treated as CSV text. ");
+
     public static final PropertyDescriptor SCHEMA = new PropertyDescriptor.Builder()
             .name("validate-csv-schema")
             .displayName("Schema")
             .description("The schema to be used for validation. Is expected a comma-delimited string representing the cell "
                     + "processors to apply. The following cell processors are allowed in the schema definition: "
                     + ALLOWED_OPERATORS + ". Note: cell processors cannot be nested except with Optional.")
-            .required(true)
+            .required(false)
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .addValidator(StandardValidators.NON_EMPTY_EL_VALIDATOR)
             .build();
@@ -168,7 +172,16 @@ public class ValidateCsv extends AbstractProcessor {
             .description("Strategy to apply when routing input files to output relationships.")
             .required(true)
             .defaultValue(VALIDATE_WHOLE_FLOWFILE.getValue())
-            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE)
+            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE, VALIDATE_ATTRIBUTE_AS_CSV)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor VALIDATION_ATTRIBUTE = new PropertyDescriptor.Builder()
+            .name("validate-csv-attribute")
+            .displayName("Validation attribute")
+            .description("FlowFile attribute to validate. The value of this attribute will be treated as CSV text.")
+            .required(true)
+            .dependsOn(VALIDATION_STRATEGY, VALIDATE_ATTRIBUTE_AS_CSV)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
 
@@ -223,6 +236,7 @@ public class ValidateCsv extends AbstractProcessor {
     protected Collection<ValidationResult> customValidate(ValidationContext context) {
 
         PropertyValue schemaProp = context.getProperty(SCHEMA);
+        PropertyValue headerProp = context.getProperty(HEADER);
         String schema = schemaProp.getValue();
         String subject = SCHEMA.getName();
 
@@ -231,7 +245,11 @@ public class ValidateCsv extends AbstractProcessor {
         }
         // If no Expression Language is present, try parsing the schema
         try {
-            this.parseSchema(schema);
+            if (schema != null) {
+                this.parseSchema(schema);
+            } else if (!headerProp.asBoolean()) {
+                throw(new Exception("Schema cannot be empty if header is false."));
+            }
         } catch (Exception e) {
             final List<ValidationResult> problems = new ArrayList<>(1);
             problems.add(new ValidationResult.Builder().subject(subject)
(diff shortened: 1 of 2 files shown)


pvillard31:
Any reason for not using ValidateRecord processor if the only requirement is to confirm that the data is valid CSV without any specific constraint?

Freedom9339:
@pvillard31 The ValidateRecord processor splits the input FlowFile into 2 Flowfiles, one for valid and another for invalid records. With the change to ValidateCSV, the whole file will be routed to either valid or invalid.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -171,7 +174,16 @@ public class ValidateCsv extends AbstractProcessor {
...
             .required(true)
             .defaultValue(VALIDATE_WHOLE_FLOWFILE.getValue())
-            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE)
+            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE, VALIDATE_ATTRIBUTE_AS_CSV)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor VALIDATION_ATTRIBUTE = new PropertyDescriptor.Builder()
+            .name("validate-csv-attribute")
+            .displayName("Validation attribute")
```
dan-s1, line 183:
```suggestion
            .name("Validation Attribute")
            .displayName("Validation Attribute")
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -458,156 +476,148 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
         final CsvPreference csvPref = getPreference(context, flowFile);
         final boolean header = context.getProperty(HEADER).asBoolean();
         final ComponentLog logger = getLogger();
-        final String schema = context.getProperty(SCHEMA).evaluateAttributeExpressions(flowFile).getValue();
-        final CellProcessor[] cellProcs = this.parseSchema(schema);
-        final boolean isWholeFFValidation = context.getProperty(VALIDATION_STRATEGY).getValue().equals(VALIDATE_WHOLE_FLOWFILE.getValue());
+        String schema = context.getProperty(SCHEMA).evaluateAttributeExpressions(flowFile).getValue();
```
dan-s1, line 479:
```suggestion
        final String schema = context.getProperty(SCHEMA).evaluateAttributeExpressions(flowFile).getValue();
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -458,156 +476,148 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
+        CellProcessor[] cellProcs = null;
+        if (schema != null) {
+            cellProcs = this.parseSchema(schema);
+        }
```
dan-s1, line 483:
```suggestion
        final CellProcessor[] cellProcs = schema != null ? this.parseSchema(schema) : null;
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: The schema cannot be final because if it's null, I modify it in line 519.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: cellProcs cannot be final because I modify it in line 520 if the schema was null.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
dan-s1: Good point though instead of modifying it couldn't you just have
`cellProcs = this.parseSchema(newSchema.substring(0, newSchema.length() - 1));`

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
dan-s1: Why isn't that code pulled up here?

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: I could, I just didn't want to make it confusing if down the line the schema variable needs to be used.

dan-s1:
@exceptionfactory Can you please restart the failed job? It does not seem related to the changes. Thanks!

Freedom9339:
Any updates on reviewing this change?

mattyb149:
There are merge conflicts that need to be resolved

Freedom9339:
@mattyb149 I've rebased against main. Thank You

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -240,7 +254,11 @@ protected Collection<ValidationResult> customValidate(ValidationContext context)
         }
         // If no Expression Language is present, try parsing the schema
         try {
-            this.parseSchema(schema);
+            if (schema != null) {
+                this.parseSchema(schema);
+            } else if (!headerProp.asBoolean()) {
+                throw(new Exception("Schema cannot be empty if header is false."));
```
mattyb149, line 260: Rather than throwing an exception, a ValidationResult should be added to a list to be returned at the end of the method.

Freedom9339: @mattyb149 Thank you for the review. A ValidationResult is created and returned once the thrown exception is catched. I could replace the throw exception with a ValidationResult, but it would be a duplicate of what the catch block already does.

https://github.com/apache/nifi/blob/9edb476aea715666a1e63d47996c911596e8eadb/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java#L262-L270

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -171,7 +174,16 @@ public class ValidateCsv extends AbstractProcessor {
             .description("Strategy to apply when routing input files to output relationships.")
             .required(true)
             .defaultValue(VALIDATE_WHOLE_FLOWFILE.getValue())
-            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE)
+            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE, VALIDATE_ATTRIBUTE_AS_CSV)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor VALIDATION_ATTRIBUTE = new PropertyDescriptor.Builder()
```
jrsteinebrey, line 181: I would change this property name and description to be consistent with ValidateXML because I feel it will make it easier to understand how to use it. Plus it will make it easier for users who use both ValidateXML and ValidateCSV to learn one way of doing it instead of two.
This is the property in ValidateXML that to use. Just change XML to CSV.
    public static final PropertyDescriptor XML_SOURCE_ATTRIBUTE = new PropertyDescriptor.Builder()
            .name("XML Source Attribute")
            .displayName("XML Source Attribute")
            .description("The name of the attribute containing XML to be validated. If this property is blank, the FlowFile content will be validated.")
            .required(false)
            .addValidator(StandardValidators.ATTRIBUTE_KEY_VALIDATOR)
            .build();

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -171,7 +174,16 @@ public class ValidateCsv extends AbstractProcessor {
             .description("Strategy to apply when routing input files to output relationships.")
             .required(true)
             .defaultValue(VALIDATE_WHOLE_FLOWFILE.getValue())
-            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE)
+            .allowableValues(VALIDATE_LINES_INDIVIDUALLY, VALIDATE_WHOLE_FLOWFILE, VALIDATE_ATTRIBUTE_AS_CSV)
```
jrsteinebrey, line 177: Remove VALIDATE_ATTRIBUTE_AS_CSV from VALIDATION_STRATEGY and use the CSV_SOURCE_ATTRIBUTE property described below to trigger validating an attribute. Using CSV_SOURCE_ATTRIBUTE removes the need for VALIDATE_ATTRIBUTE_AS_CSV.  VALIDATE_ATTRIBUTE_AS_CSV does not seem to logically belong with the other two existing options in VALIDATION_STRATEGY.
Sidenote: Can the Attribute to Validate as CSV contain more than 1 line of csv data?

jrsteinebrey (changes requested):

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: I've made this change.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: I've made this change. Also, yes, the attribute can contain as many lines as the attribute value can hold.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -458,155 +473,147 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
         if (!isWholeFFValidation) {
-            invalidFF.set(session.create(flowFile));
-            validFF.set(session.create(flowFile));
+            invalidFF = session.create(flowFile);
+            validFF = session.create(flowFile);
+        }
+
+        InputStream stream;
+        if (context.getProperty(CSV_SOURCE_ATTRIBUTE).isSet()) {
+            stream = new ByteArrayInputStream(flowFile.getAttribute(context.getProperty(CSV_SOURCE_ATTRIBUTE).getValue()).getBytes(StandardCharsets.UTF_8));
```
jrsteinebrey, line 501: @Freedom9339  Thanks for making that change. It turned out well except you need to call .evaluateAttributeExpressions() after the .getProperty(CSV_SOURCE_ATTRIBUTE) call. After that, the change looks complete.

Freedom9339: Done.

jrsteinebrey: @Freedom9339 Thanks for your contribution.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -120,11 +120,20 @@ public class ValidateCsv extends AbstractProcessor {
             .description("The schema to be used for validation. Is expected a comma-delimited string representing the cell "
                     + "processors to apply. The following cell processors are allowed in the schema definition: "
                     + allowedOperators.toString() + ". Note: cell processors cannot be nested except with Optional.")
-            .required(true)
+            .required(false)
```
mattyb149, line 123: This should have a `dependsOn(HEADER, "false")`, unless you can provide the explicit schema yet there be a header line that should be ignored.

mattyb149: I think the "unless" part is true here. But now there are merge conflicts, sorry I lost track of this

exceptionfactory:
@mosermw noting that you removed the `stale` label, are you planning to review this pull request? Although this has had some activity, the lack involvement over four months seems to indicate that it should be closed, with the option to revisit if there is a positive comment on the Jira issue that a committer is planning to review and guide this to completion.

mosermw:
> @mosermw noting that you removed the `stale` label, are you planning to review this pull request?

It looks like Matt B. made some comments that should be addressed, along with the merge conflicts.  I contacted the PR author to ask if they still want this change, and will be happy to work with them to finish this.

exceptionfactory:
Thanks @mosermw!

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: I've updated the documentation.

Freedom9339:
@mosermw @exceptionfactory @mattyb149 Sorry about loosing track of this. I've rebased and addressed the comments provided.

mosermw:
Reviewing ...

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -449,155 +465,147 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-        final AtomicReference<FlowFile> validFF = new AtomicReference<>(null);
-        final AtomicReference<String> validationError = new AtomicReference<>(null);
+        boolean valid = true;
+        int okCount = 0;
+        int totalCount = 0;
+        FlowFile invalidFF = null;
+        FlowFile validFF = null;
+        String validationError = null;
+        final AtomicReference<Boolean> isFirstLineValid = new AtomicReference<Boolean>(true);
+        final AtomicReference<Boolean> isFirstLineInvalid = new AtomicReference<Boolean>(true);
```
mosermw, line 484: A recent commit used diamond notation for the constructor and I suspect your rebase didn't pick this up.  This is especially important since new PMD rules were recently committed that checks for this.
```suggestion
        final AtomicReference<Boolean> isFirstLineValid = new AtomicReference<>(true);
        final AtomicReference<Boolean> isFirstLineInvalid = new AtomicReference<>(true);
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestValidateCsv.java
mosermw, line 1: Please add a unit test where the CSV Source Attribute is defined but is missing from the flowfile.  Also verify that flowfile content is not modified when validating an attribute.

mosermw, line 1: Please add a unit test to verify flowfile goes to "valid" when validating an attribute and the attribute value is missing or empty and a schema is defined.  This explicitly defines expected behavior when there is no CSV data to validate.  Also verify that flowfile content is not modified when validating an attribute.

mosermw, line 1: Please add a unit test to verify a 0 byte flowfile goes to "valid" when validating flowfile content and a schema is defined.  This explicitly defines expected behavior when there is no CSV data to validate.  This doesn't test any changes you made in this PR but would really help fill out this unit test.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -449,155 +465,147 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
         if (!isWholeFFValidation) {
-            invalidFF.set(session.create(flowFile));
-            validFF.set(session.create(flowFile));
+            invalidFF = session.create(flowFile);
+            validFF = session.create(flowFile);
+        }
+
+        InputStream stream;
+        if (context.getProperty(CSV_SOURCE_ATTRIBUTE).isSet()) {
+            stream = new ByteArrayInputStream(flowFile.getAttribute(context.getProperty(CSV_SOURCE_ATTRIBUTE).evaluateAttributeExpressions().getValue()).getBytes(StandardCharsets.UTF_8));
```
mosermw, line 493: The call to flowFile.getAttribute() can return null if the attribute does not exist on the flowfile.  If that happens, this line will cause a NullPointerException.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -204,6 +214,7 @@ public class ValidateCsv extends AbstractProcessor {
             .description("FlowFiles that are not valid according to the specified schema are routed to this relationship")
             .build();
 
+
```
mosermw, line 217: This extra blank line doesn't appear to be necessary.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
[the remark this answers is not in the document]
Freedom9339: Done.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -98,6 +97,7 @@ public class ValidateCsv extends AbstractProcessor {
             "Strlen", "StrMinMax", "StrNotNullOrEmpty", "StrRegEx", "Unique", "UniqueHashCode", "IsIncludedIn"
     );
 
+
```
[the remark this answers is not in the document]
Freedom9339: Removed

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
```diff
@@ -231,7 +243,11 @@ protected Collection<ValidationResult> customValidate(ValidationContext context)
         }
         // If no Expression Language is present, try parsing the schema
         try {
-            this.parseSchema(schema);
+            if (schema != null) {
+                this.parseSchema(schema);
+            } else if (!headerProp.asBoolean()) {
+                throw(new Exception("Schema cannot be empty if header is false."));
```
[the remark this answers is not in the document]
Freedom9339: Done.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: Done.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestValidateCsv.java
Freedom9339: Done.

Freedom9339: Done.

Freedom9339: Done.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: Fixed.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ValidateCsv.java
Freedom9339: Removed.

mosermw:
Thanks for making these changes @Freedom9339.  One or two last things that I noticed.

When validating a CSV Source Attribute, we can set the "Validation Strategy = Line by line".  When we do this, the flowfile content is replaced with the value of the CSV attribute.  This is clearly not good!  I cannot think of a use case where we want to allow "Validation Strategy = Line by line" on an attribute ... it should always validate the whole attribute and not change anything.

We can make the CSV Source Attribute depend on "Validation Strategy = FlowFile" but then we would have to defend ourselves from the case when someone enters a CSV Source Attribute and then changes the Validation Strategy.  The CSV Source Attribute will disappear and I'm worried that it will still be set and we will still try to use it.

So, if using dependsOn() gets too complicated, just add something to customValidate to disallow CSV Source Attribute with "Validation Strategy = Line by line"

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestValidateCsv.java
```diff
@@ -164,6 +168,107 @@ public void testBigDecimalBoolCharIntLong() {
...
+    @Test
+    public void testNoSchema() {
+        final TestRunner runner = TestRunners.newTestRunner(new ValidateCsv());
+        runner.setProperty(ValidateCsv.DELIMITER_CHARACTER, ",");
+        runner.setProperty(ValidateCsv.END_OF_LINE_CHARACTER, "\r\n");
+        runner.setProperty(ValidateCsv.QUOTE_CHARACTER, "\"");
+        runner.setProperty(ValidateCsv.HEADER, "true");
+
+        runner.enqueue("bigdecimal,bool,char,integer,long\r\n10.0001,true,c,1,92147483647");
+        runner.enqueue(new byte[0]);
```
mosermw, line 180: This second runner.enqueue() isn't necessary.

Freedom9339: Removed.

Freedom9339:
@mosermw I've added the depends on and made the validate attribute only show when using whole ff validation. I've added a check to make sure whole ff validation is being used before using the attribute as a source.

mosermw:
Thanks for sticking with this @Freedom9339.  Everything looks goo now.

+1 merging.

PR #8372, opened 2024-02-07
[NIFI-12746] fix search UX bug

scottyaslan:
This PR fixes a bug that causes the search input to close when the user clicks on the field.


mcgilman:
Reviewing...

mcgilman (approved):
Thanks @scottyaslan!

PR #8390, opened 2024-02-09
NIFI-12772 Expose REMOTE_POLL_BATCH_SIZE property for ListSFTP

tombrisland:
# Summary

[NIFI-12772](https://issues.apache.org/jira/browse/NIFI-12772)

Allows use of the REMOTE_POLL_BATCH_SIZE property in the ListSFTP processor. Previously this was supported but not exposed for configuration.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FileTransfer.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FileTransfer.java
@@ -29,6 +29,7 @@ import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.processor.util.list.AbstractListProcessor;
 
 public interface FileTransfer extends Closeable {
 
--- a/nifi-nar-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FileTransfer.java
+++ b/nifi-nar-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FileTransfer.java
@@ -29,7 +29,6 @@ import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
-import org.apache.nifi.processor.util.list.AbstractListProcessor;
 
 public interface FileTransfer extends Closeable {
 
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
@@ -317,12 +317,6 @@ public class SFTPTransfer implements FileTransfer {
             final RemoteResourceFilter filter = (entry) -> {
                 final String entryFilename = entry.getName();
 
-                // Since SSHJ does not have the concept of BREAK that JSCH had, we need to move this before the call to listing.add
-                // below, otherwise we would keep adding to the listings since returning false here doesn't break
-                if (listing.size() >= maxResults) {
-                    return false;
-                }
-
                 // skip over 'this directory' and 'parent directory' special files regardless of ignoring dot files
                 if (RELATIVE_CURRENT_DIRECTORY.equals(entryFilename) || RELATIVE_PARENT_DIRECTORY.equals(entryFilename)) {
                     return false;
@@ -335,6 +329,12 @@ public class SFTPTransfer implements FileTransfer {
 
                 if (isIncludedDirectory(entry, recurse, symlink)) {
                     subDirs.add(entry);
+                    return false;
+                }
+
+                // Since SSHJ does not have the concept of BREAK that JSCH had, we need to move this before the call to listing.add
+                // below, otherwise we would keep adding to the listings since returning false here doesn't break
+                if (listing.size() >= maxResults) {
                     return false;
                 }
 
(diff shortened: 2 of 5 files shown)


tombrisland:
Need to do some manual testing

EndzeitBegins:
Thank you for working on this @tombrisland.
I'm really looking forward to this improvement and hope we can backport it to 1.x as well. 

I haven't had time to do any testing, but just glanced over the code changes and they look straightforward.

It seems that even though the `SFTPTransfer` supports the property already, the underlying `sshj` does not support a limit to `ls` directly but instead we rather stop adding results to our list. The listing is continued nonetheless. 
I think supporting the property is still an improvement, as we at least can limit the amount of FlowFiles / records created this way. 

However, I wonder if it makes sense to move the existing max result check
```Java
if (listing.size() >= maxResults) {
    return false;
}
```
to the top of the `RemoteResourceFilter`? The limit only ever gets relevant, when there is a large amount of files at the remote source. Even though we cannot abort the listing prematurely, at least not without modifications to sshj, this way the checks needed for all files after the limit is reached is kept to a minimum.

What do you think about that?

tombrisland:
Build above failed with a heap error.

Makes sense to me @EndzeitBegins, would also be great if we can get your early termination support in!

We may also need to have a think about which tracking strategies are supported when you select batch size. E.g. I don't think that time tracking will work with batch size since it relies on keeping track of the last timestamp processed?

EndzeitBegins:
Thank you for the adjustments. 
While I hope we can get the changes to land in `sshj` I wouldn't wait for them. After introducing the general support for `REMOTE_POLL_BATCH_SIZE` in `ListSFTP`, we still can integrate a new version that includes the improvements in a following PR.

That's a very sensible remark. I think it makes sense to limit the support for `REMOTE_POLL_BATCH_SIZE` to the strategies `BY_ENTITIES` and `NO_TRACKING` only.  I'm not familiar with `BY_TIME_WINDOW` but I think combining the batch size with one of the other strategies might result in some files being skipped / never be picked up.

nifi-nar-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FileTransfer.java
```diff
@@ -172,6 +173,8 @@ default String getAbsolutePath(FlowFile flowFile, String remotePath) throws IOEx
             + "than normal.")
         .defaultValue("5000")
         .addValidator(StandardValidators.NON_NEGATIVE_INTEGER_VALIDATOR)
+        // Limiting to batches means that TIME_TRACKING and TIME_WINDOW strategies will leave files behind
+        .dependsOn(AbstractListProcessor.LISTING_STRATEGY, AbstractListProcessor.BY_ENTITIES, AbstractListProcessor.NO_TRACKING)
```
exceptionfactory, line 177: This should not be changed because it also impacts the `ListFTP` Processor.

exceptionfactory (changes requested):
Thanks for working on this change @tombrisland, and thanks for the feedback thus far @EndzeitBegins.

Regarding making the property dependent, this should not be changed in `FileTransfer` because it also impacts `ListFTP` and `GetFTP`.

It does not seem necessary to make it dependent, because those other strategies could still work under some circumstances. The fact that it is not currently a dependent property for the FTP Processors is also worth noting, so at least for the initial change, it seems better to avoid making it a dependent property.

nifi-nar-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FileTransfer.java
tombrisland: I did intend for the changes to cover both, I can't think how `TIME_TRACKING` can work with batches for either `ListSFTP` or `ListFTP`. Have you got an example where it does?

I checked out `GetFTP` and `GetSFTP` and they retain the property even with the dependency, it seems that if the 'depended on' property doesn't exist then it defaults to visible.

Happy to make the change of course, can you think of a different way to communicate the limitation to the user?

joewitt:
I don't have a lot to add to the comments I see already.  But the main gist is if the intent of the JIRA is now broader than the original title please change the JIRA.  If the intent has not changed then please narrow the PR.  So either this is about adding this option to ListSFTP or it is about others.  Would be important to validate each case which expands the scope of the review.

Given we don't actually get to, it appears, change the behavior of the SFTP commands executed on the remote end (ie: once you start a listing in a large directory...much of the damage is already done in terms of time/waiting) then the benefit of this property being available is mostly to protect the NiFi side.  Still useful just probably not the truly desired/most beneficial impact.

I'd avoid *any* changes for now that aren't central to the specific desired improvement.

nifi-nar-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FileTransfer.java
```diff
@@ -29,6 +29,7 @@
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.processor.util.list.AbstractListProcessor;
```
exceptionfactory, line 32: This change can be reverted

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
```diff
@@ -317,6 +317,12 @@ protected void getListing(final String path, final int depth, final int maxResul
             final RemoteResourceFilter filter = (entry) -> {
                 final String entryFilename = entry.getName();
 
+                // Since SSHJ does not have the concept of BREAK that JSCH had, we need to move this before the call to listing.add
+                // below, otherwise we would keep adding to the listings since returning false here doesn't break
+                if (listing.size() >= maxResults) {
+                    return false;
+                }
+
```
exceptionfactory, line 325: Is this change necessary? It seems useful in some circumstances, but not needed for this particular property change.

exceptionfactory (commented):
Thanks for the update @tombrisland. I noticed one import that no longer seems necessary, and a had a question on one other relocation.

Also, if you could do this change as a regular push, instead of a force-push, that would be helpful for tracking. The initial PR should have a single commit, but after the review process is complete, then all the changes will be squashed into a single commit for merging into the main branch.

EndzeitBegins:
Thanks for taking on the merge process @joewitt. 
Could we backport these changes to the `support/nifi-1.x` branch as well?

joewitt:
@EndzeitBegins It would need a separate JIRA/PR would be needed and I dont think the 'removeIf' language mechanism would work as NiFi 1.x is Java 8 or higher.

EndzeitBegins:
Thank you for the fast response @joewitt and letting me know. 

Hey @tombrisland, are you interested to work on a backport for 1.x or should I have a go at it?

PR #8405, opened 2024-02-14
NIFI-12725 Updated json-schema-validator to 1.3.0. 

dan-s1:
With the upgrade in version the change in json-schema-validator of URIs for schemas Draft 4, Draft 6 and Draft 7 had to be taken into account.

# Summary

[NIFI-12725](https://issues.apache.org/jira/browse/NIFI-12725)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-json-schema-api/src/main/java/org/apache/nifi/json/schema/SchemaVersion.java
+++ b/nifi-commons/nifi-json-schema-api/src/main/java/org/apache/nifi/json/schema/SchemaVersion.java
@@ -19,9 +19,9 @@ package org.apache.nifi.json.schema;
 import org.apache.nifi.components.DescribedValue;
 
 public enum SchemaVersion implements DescribedValue {
-    DRAFT_4("Draft Version 4", "Draft 4", "https://json-schema.org/draft-04/schema"),
-    DRAFT_6("Draft Version 6", "Draft 6", "https://json-schema.org/draft-06/schema"),
-    DRAFT_7("Draft Version 7", "Draft 7", "https://json-schema.org/draft-07/schema"),
+    DRAFT_4("Draft Version 4", "Draft 4", "http://json-schema.org/draft-04/schema#"),
+    DRAFT_6("Draft Version 6", "Draft 6", "http://json-schema.org/draft-06/schema#"),
+    DRAFT_7("Draft Version 7", "Draft 7", "http://json-schema.org/draft-07/schema#"),
     DRAFT_2019_09("Draft Version 2019-09", "Draft 2019-09", "https://json-schema.org/draft/2019-09/schema"),
     DRAFT_2020_12("Draft Version 2020-12", "Draft 2020-12", "https://json-schema.org/draft/2020-12/schema");
 
--- a/nifi-nar-bundles/nifi-registry-bundle/nifi-registry-service/pom.xml
+++ b/nifi-nar-bundles/nifi-registry-bundle/nifi-registry-service/pom.xml
@@ -60,7 +60,7 @@ language governing permissions and limitations under the License. -->
         <dependency>
             <groupId>com.networknt</groupId>
             <artifactId>json-schema-validator</artifactId>
-            <version>1.1.0</version>
+            <version>1.3.0</version>
         </dependency>
         <dependency>
             <groupId>org.apache.nifi</groupId>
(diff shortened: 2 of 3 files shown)


exceptionfactory (commented):
@dan-s1 Can you provide some additional background on the reason for this change based on the changes in the validator library? The adjusted URI values seem odd in light of the unchanged values for 2019-09 and 2020-12, which makes me think there was some unexpected change in the validator library that should be considered.

exceptionfactory:
The latest version of json-schema-validator is now 1.3.2, so recommend changing to that version as part of this upgrade.

The following issue indicates some changes in the library for schema fragment handling, perhaps related, but it calls into question whether the reference URI values should be changed, or whether there is a better approach.

https://github.com/networknt/json-schema-validator/pull/930

dan-s1:
> @dan-s1 Can you provide some additional background on the reason for this change based on the changes in the validator library? The adjusted URI values seem odd in light of the unchanged values for 2019-09 and 2020-12, which makes me think there was some unexpected change in the validator library that should be considered.

I had to make these changes based on your suggestions in https://github.com/apache/nifi/pull/8005#discussion_r1389887565
Now that json-schema-validator changed some of the underlying URI strings that is why when Mike tried to upgrade there were issues as there were no matches for schemas 4, 6 and 7.

exceptionfactory:
> > @dan-s1 Can you provide some additional background on the reason for this change based on the changes in the validator library? The adjusted URI values seem odd in light of the unchanged values for 2019-09 and 2020-12, which makes me think there was some unexpected change in the validator library that should be considered.
> 
> I had to make these changes based on your suggestions in [#8005 (comment)](https://github.com/apache/nifi/pull/8005#discussion_r1389887565) Now that json-schema-validator changed some of the underlying URI strings that is why when Mike tried to upgrade there were issues as there were no matches for schemas 4, 6 and 7.

For clarification, did you find the related changes in the json-schema-validator library that required the changes in NiFi? Recent versions included several other breaking changes, so this is something where it would be helpful to validate the changes in the library itself, as opposed to just making changes in NiFi immediately.

dan-s1:
Yes I found the changes to the URI strings [here](https://github.com/networknt/json-schema-validator/blob/1.3.0/src/main/java/com/networknt/schema/SchemaId.java)

dan-s1:
@exceptionfactory Can you clarify what you mean by "breaking changes" in regards to the NIFI code? As far as I can see the NIFI code is not adding functionality it, merely a pass through so the json-schema-validator API could be used.

exceptionfactory:
> @exceptionfactory Can you please clarify what you mean by "breaking changes" in regards to the NIFI code? As far as I can see the NIFI code is not adding functionality, it merely is a pass through so the json-schema-validator API could be used.

Sure, the breaking changes in `json-schema-validator` apply to other aspects of the library, not used in NiFi. Thankfully, it looks like the only issue is the Schema URI references, which now appear to be aligned with the URI values from [JSON Schema Specifications](https://json-schema.org/specification-links), which makes sense from an official alignment point of view.

dan-s1:
@exceptionfactory I wasn't sure though would this affect anyone who already is using the [StandardJsonSchemaRegistry](https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-registry-nar/1.25.0/org.apache.nifi.schemaregistry.services.StandardJsonSchemaRegistry/index.html) as the registry does store `org.apache.nifi.json.schema.SchemaVersion` objects.

exceptionfactory:
> @exceptionfactory I wasn't sure though would this affect anyone who already is using the [StandardJsonSchemaRegistry](https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-registry-nar/1.25.0/org.apache.nifi.schemaregistry.services.StandardJsonSchemaRegistry/index.html) as the registry does store `org.apache.nifi.json.schema.SchemaVersion` objects.

Thanks for asking @dan-s1. Thankfully the `StandardJsonSchemaRegistry` does not use the URI for referencing the version, so just changing the URI value does not impact existing behavior.

exceptionfactory (approved):
Thanks again for tracking down the required changes @dan-s1, looks good. +1 merging

PR #8407, opened 2024-02-14
NIFI-11443 Route Python Framework Logging to SLF4J

exceptionfactory:
# Summary

[NIFI-11443](https://issues.apache.org/jira/browse/NIFI-11443) Updates Python Process log handling to route messages to SLF4J so that logging levels and appenders can be managed using Logback.

These changes remove the need for separate Python log settings in `nifi.properties` and support translating Python log levels and logger names to Logback.

The implementation changes the Python logging format to use the Python [level numbers](https://docs.python.org/3/library/logging.html#levels) for optimized parsing. The implementation changes Python logging to write messages to the standard output stream, and the `PythonProcess` builder creates a Java Virtual Thread to read lines from the `BufferedReader` containing process output stream messages.

These changes eliminate the dedicated `nifi-python.log` and enable Python logging to be handled along with other framework logging that defaults to the `nifi-app.log` file appender.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
+++ b/nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
@@ -22,8 +22,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.BufferedReader;
 import java.io.IOException;
+import java.util.ArrayDeque;
 import java.util.Arrays;
-import java.util.LinkedList;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Queue;
@@ -74,16 +74,20 @@ class PythonProcessLogReader implements Runnable {
      */
     @Override
     public void run() {
-        final Queue<ParsedRecord> parsedRecords = new LinkedList<>();
+        final Queue<ParsedRecord> parsedRecords = new ArrayDeque<>();
 
         try {
             String line = processReader.readLine();
             while (line != null) {
-                processLine(line, parsedRecords);
+                try {
+                    processLine(line, parsedRecords);
 
-                if (parsedRecords.size() == 2 || !processReader.ready()) {
-                    final ParsedRecord parsedRecord = parsedRecords.remove();
-                    log(parsedRecord);
+                    if (parsedRecords.size() == 2 || !processReader.ready()) {
+                        final ParsedRecord parsedRecord = parsedRecords.remove();
+                        log(parsedRecord);
+                    }
+                } catch (final Exception e) {
+                    processLogger.error("Failed to handle log from Python Process", e);
                 }
 
                 // Read and block for subsequent lines
--- a/nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
+++ b/nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
@@ -58,7 +58,7 @@ class PythonProcessLogReader implements Runnable {
                 line = processReader.readLine();
             }
         } catch (final IOException e) {
-            processLogger.error("Read Process output failed", e);
+            processLogger.error("Failed to read output of Python Process", e);
         }
     }
 
--- a/nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
+++ b/nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
@@ -82,9 +82,18 @@ class PythonProcessLogReader implements Runnable {
                 try {
                     processLine(line, parsedRecords);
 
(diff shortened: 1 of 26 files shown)


nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
```diff
@@ -0,0 +1,140 @@
...
+        try {
+            String line = processReader.readLine();
+            while (line != null) {
+                if (line.length() > MINIMUM_LINE_LENGTH) {
+                    processLine(line);
+                }
+                line = processReader.readLine();
+            }
+        } catch (final IOException e) {
+            processLogger.error("Read Process output failed", e);
```
markap14, line 61: The message here is quite vague; perhaps "Failed to read output of Python Process"?

nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/test/java/org/apache/nifi/py4j/PythonProcessLogReaderTest.java
```diff
@@ -0,0 +1,127 @@
...
+            final BufferedReader reader = new BufferedReader(new StringReader(LINE_UNFORMATTED));
+            final Runnable command = new PythonProcessLogReader(reader);
+            command.run();
+        }
+
+        verify(logger).info(eq(LINE_UNFORMATTED));
+    }
+
+    @Test
+    void testMissingSeparator() {
```
markap14, line 101: You mind adding a test for logging an empty string? I was wondering when reading the code above, if the `:` separator is the last character in the line, is that going to throw some sort of ArrayIndexOutOfBoundsException or something?

nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
exceptionfactory: The logger name may help, but your suggestion is definitely clearer.

exceptionfactory:
Thanks for the initial feedback @markap14! Based on further discussion, I will be following up with additional changes to incorporate log level updates from Logback.

nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
```diff
@@ -0,0 +1,163 @@
...
+     * Read lines from Process Reader and write log messages based on parsed level and named logger
+     */
+    @Override
+    public void run() {
+        final Queue<ParsedRecord> parsedRecords = new LinkedList<>();
+
+        try {
+            String line = processReader.readLine();
+            while (line != null) {
+                processLine(line, parsedRecords);
```
markap14, line 82: I think within this `while` loop we should have a try/catch (Exception e)
Without it, if there's any Exception thrown from parsing, etc. the execution will escape the run() method without any sort of warning, and the logging will simply stop. Which will eventually cause the Python process' STDOUT to fill up and then the Python process will block.
Need to ensure that this thread never dies unexpectedly.

nifi-nar-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcessLogReader.java
```diff
@@ -0,0 +1,163 @@
...
+
+                if (parsedRecords.size() == 2 || !processReader.ready()) {
+                    final ParsedRecord parsedRecord = parsedRecords.remove();
+                    log(parsedRecord);
+                }
```
markap14, line 87: I think the logic here is a bit off.
I created this Processor for testing:
```
from nifiapi.flowfiletransform import FlowFileTransform, FlowFileTransformResult
import logging

root_logger = logging.getLogger('')

class LogContents(FlowFileTransform):
    class Java:
        implements = ['org.apache.nifi.python.processor.FlowFileTransform']
    class ProcessorDetails:
        version = '0.0.1-SNAPSHOT'

    def __init__(self, **kwargs):
        pass

    def transform(self, context, flowFile):
        contents = flowFile.getContentsAsBytes().decode("utf-8")
        root_logger.info("Contents:\n1\n2\n3")
        root_logger.info(contents)
        return FlowFileTransformResult(relationship = "success")
```

I then sent in some FlowFiles. Each time, I get the log output of:
```
Contents
1
2
3
```

but not the output of the actual logger.
If I change the first log message instead to:
```
        root_logger.info("Contents:")
```

I get just the line "Contents:" on the first iteration. The second time the processor is run, it outputs the contents of the first FlowFile, followed by "Contents" again.

exceptionfactory: Thanks for the thorough feedback and testing. I made a change to check the readiness of the Process Reader separately, and then the example `LogContents` Processor writes the messages as expected.

markap14:
Thanks @exceptionfactory all seems to be working well now. This is a huge improvement! +1 will merge to main

PR #8417, opened 2024-02-15
NIFI-12017 add ability to choose to output to single line for base32 base64 contents

knguyen1:
# Summary

[NIFI-12017](https://issues.apache.org/jira/browse/NIFI-12017)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -16,6 +16,14 @@
  */
 package org.apache.nifi.processors.standard;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
 import org.apache.commons.codec.DecoderException;
 import org.apache.commons.codec.binary.Base32InputStream;
 import org.apache.commons.codec.binary.Base32OutputStream;
@@ -29,24 +37,22 @@ import org.apache.nifi.annotation.behavior.SupportsBatching;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.Tags;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.Validator;
+import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.processor.AbstractProcessor;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.io.StreamCallback;
+import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.processors.standard.encoding.EncodingMode;
+import org.apache.nifi.processors.standard.encoding.EncodingType;
+import org.apache.nifi.processors.standard.encoding.LineOutputMode;
 import org.apache.nifi.processors.standard.util.ValidatingBase32InputStream;
 import org.apache.nifi.processors.standard.util.ValidatingBase64InputStream;
 import org.apache.nifi.stream.io.StreamUtils;
 import org.apache.nifi.util.StopWatch;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Set;
-import java.util.concurrent.TimeUnit;
 
 @SideEffectFree
 @SupportsBatching
@@ -55,28 +61,63 @@ import java.util.concurrent.TimeUnit;
 @CapabilityDescription("Encode or decode the contents of a FlowFile using Base64, Base32, or hex encoding schemes")
 public class EncodeContent extends AbstractProcessor {
 
-    public static final String ENCODE_MODE = "Encode";
-    public static final String DECODE_MODE = "Decode";
-
-    public static final String BASE64_ENCODING = "base64";
-    public static final String BASE32_ENCODING = "base32";
-    public static final String HEX_ENCODING = "hex";
-
     public static final PropertyDescriptor MODE = new PropertyDescriptor.Builder()
             .name("Mode")
             .description("Specifies whether the content should be encoded or decoded")
             .required(true)
-            .allowableValues(ENCODE_MODE, DECODE_MODE)
-            .defaultValue(ENCODE_MODE)
+            .allowableValues(EncodingMode.class)
+            .defaultValue(EncodingMode.ENCODE.getValue())
             .build();
 
     public static final PropertyDescriptor ENCODING = new PropertyDescriptor.Builder()
             .name("Encoding")
             .description("Specifies the type of encoding used")
             .required(true)
-            .allowableValues(BASE64_ENCODING, BASE32_ENCODING, HEX_ENCODING)
-            .defaultValue(BASE64_ENCODING)
+            .allowableValues(EncodingType.class)
+            .defaultValue(EncodingType.BASE64_ENCODING.getValue())
             .build();
+
+    static final PropertyDescriptor LINE_OUTPUT_MODE = new PropertyDescriptor.Builder()
+            .name("line-output-mode")
+            .displayName("Line Output Mode")
+            .description("If set to 'single-line', the encoded FlowFile content will output as a single line. If set to 'multiple-lines', "
+                + "it will output as multiple lines. This property is only applicable when Base64 or Base32 encoding is selected.")
+            .required(false)
+            .defaultValue(LineOutputMode.SINGLE_LINE.getValue())
+            .allowableValues(LineOutputMode.class)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .dependsOn(MODE, EncodingMode.ENCODE)
+            .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
+            .build();
+
+    static final PropertyDescriptor ENCODED_LINE_SEPARATOR = new PropertyDescriptor.Builder()
+        .name("line-separator")
+        .displayName("Encoded Content Line Separator")
+        .description("Each line of encoded data will be terminated with this byte sequence (e.g. \\r\\n"
+                + "). This property defaults to the system-dependent line separator string.  If `line-length` <= 0, "
+                + "the `line-separator` property is not used. This property is not used for `hex` encoding.")
+        .required(false)
+        .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+        .defaultValue(System.lineSeparator())
+        .addValidator(Validator.VALID)
+        .dependsOn(MODE, EncodingMode.ENCODE)
+        .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
+        .build();
+
+    static final PropertyDescriptor ENCODED_LINE_LENGTH = new PropertyDescriptor.Builder()
+        .name("encoded-line-length")
+        .displayName("Encoded Content Line Length")
+        .description("Each line of encoded data will contain `encoded-line-length` characters (rounded down to the nearest multiple of 4). "
+            + "If `encoded-line-length` <= 0, the encoded data is not divided into lines. This property is "
+            + "ignored if `line-output-mode` is set to `single-line`.")
+        .required(false)
+        .defaultValue("76")
+        .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+        .addValidator(StandardValidators.INTEGER_VALIDATOR)
+        .dependsOn(MODE, EncodingMode.ENCODE)
+        .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
+        .dependsOn(LINE_OUTPUT_MODE, LineOutputMode.MULTIPLE_LINES)
+        .build();
 
     public static final Relationship REL_SUCCESS = new Relationship.Builder()
             .name("success")
@@ -90,7 +131,7 @@ public class EncodeContent extends AbstractProcessor {
     private static final int BUFFER_SIZE = 8192;
 
     private static final List<PropertyDescriptor> properties = List.of(MODE,
-        ENCODING);
+        ENCODING, LINE_OUTPUT_MODE, ENCODED_LINE_SEPARATOR, ENCODED_LINE_LENGTH);
 
     private static final Set<Relationship> relationships = Set.of(REL_SUCCESS,
         REL_FAILURE);
@@ -112,9 +153,14 @@ public class EncodeContent extends AbstractProcessor {
             return;
         }
 
-        final boolean encode = context.getProperty(MODE).getValue().equalsIgnoreCase(ENCODE_MODE);
+        final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
         final String encoding = context.getProperty(ENCODING).getValue();
-        final StreamCallback callback = getStreamCallback(encode, encoding);
+
+        final Boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
+        final int lineLength = context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
+        final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
+
+        final StreamCallback callback = getStreamCallback(encode, encoding, Boolean.TRUE.equals(singleLineOutput) ? -1 : lineLength, lineSeparator);
 
         try {
             final StopWatch stopWatch = new StopWatch(true);
@@ -129,31 +175,35 @@ public class EncodeContent extends AbstractProcessor {
         }
     }
 
-    private static StreamCallback getStreamCallback(final boolean encode, final String encoding) {
-        if (encode) {
-            if (encoding.equalsIgnoreCase(BASE64_ENCODING)) {
-                return new EncodeBase64();
-            } else if (encoding.equalsIgnoreCase(BASE32_ENCODING)) {
-                return new EncodeBase32();
-            } else {
-                return new EncodeHex();
-            }
-        } else {
-            if (encoding.equalsIgnoreCase(BASE64_ENCODING)) {
-                return new DecodeBase64();
-            } else if (encoding.equalsIgnoreCase(BASE32_ENCODING)) {
-                return new DecodeBase32();
-            } else {
-                return new DecodeHex();
-            }
+    private static StreamCallback getStreamCallback(final Boolean encode, final String encoding,
+        final int lineLength, final String lineSeparator) {
+        switch(encoding) {
+            case "BASE64_ENCODING":
+                return encode ? new EncodeBase64(lineLength, lineSeparator) : new DecodeBase64();
+            case "BASE32_ENCODING":
+                return encode ? new EncodeBase32(lineLength, lineSeparator) : new DecodeBase32();
+            default:
+                return encode ? new EncodeHex() : new DecodeHex();
         }
     }
 
     private static class EncodeBase64 implements StreamCallback {
 
+        private int lineLength;
+        private String lineSeparator;
+
+        public EncodeBase64(final int lineLength,
+            final String lineSeparator) {
+            this.lineLength = lineLength;
+            this.lineSeparator = lineSeparator;
+        }
+
         @Override
         public void process(final InputStream in, final OutputStream out) throws IOException {
-            try (Base64OutputStream bos = new Base64OutputStream(out)) {
+            try (Base64OutputStream bos = new Base64OutputStream(out,
+                true,
+                this.lineLength,
+                this.lineSeparator.getBytes())) {
                 StreamUtils.copy(in, bos);
             }
         }
@@ -171,9 +221,22 @@ public class EncodeContent extends AbstractProcessor {
 
     private static class EncodeBase32 implements StreamCallback {
 
+        private int lineLength;
+        private String lineSeparator;
+
+        public EncodeBase32(final int lineLength,
+            final String lineSeparator) {
+
+            this.lineLength = lineLength;
+            this.lineSeparator = lineSeparator;
+        }
+
         @Override
         public void process(final InputStream in, final OutputStream out) throws IOException {
-            try (Base32OutputStream bos = new Base32OutputStream(out)) {
+            try (Base32OutputStream bos = new Base32OutputStream(out,
+                true,
+                this.lineLength,
+                this.lineSeparator.getBytes())) {
                 StreamUtils.copy(in, bos);
             }
         }
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -186,8 +186,8 @@ public class EncodeContent extends AbstractProcessor {
 
     private static class EncodeBase64 implements StreamCallback {
 
-        private int lineLength;
-        private String lineSeparator;
+       private final int lineLength;
+       private final String lineSeparator;
 
         private EncodeBase64(final int lineLength,
             final String lineSeparator) {
@@ -218,8 +218,8 @@ public class EncodeContent extends AbstractProcessor {
 
     private static class EncodeBase32 implements StreamCallback {
 
-        private int lineLength;
-        private String lineSeparator;
+        private final int lineLength;
+        private final String lineSeparator;
 
         public EncodeBase32(final int lineLength,
             final String lineSeparator) {
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -153,7 +153,7 @@ public class EncodeContent extends AbstractProcessor {
         }
 
         final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
-        final EncodingType encoding = getEncodingTypeFromString(context.getProperty(ENCODING).getValue());
+        final EncodingType encoding = getEncodingType(context.getProperty(ENCODING).getValue());
         final boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
         final int lineLength = singleLineOutput ? -1 : context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
         final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -298,10 +298,10 @@ public class EncodeContent extends AbstractProcessor {
         }
     }
 
-    private static EncodingType getEncodingTypeFromString(final String encodingTypeValue) {
-        if (encodingTypeValue.compareToIgnoreCase("base64") == 0) {
+    private static EncodingType getEncodingType(final String encodingTypeValue) {
+        if (EncodingType.BASE64.getValue().equals(encodingTypeValue)) {
             return EncodingType.BASE64;
-        } else if (encodingTypeValue.compareToIgnoreCase("base32") == 0) {
+        } else if (EncodingType.BASE32.getValue().equals(encodingTypeValue)) {
             return EncodingType.BASE32;
         } else {
             return EncodingType.HEXADECIMAL;
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -63,7 +63,7 @@ public class EncodeContent extends AbstractProcessor {
 
     public static final PropertyDescriptor MODE = new PropertyDescriptor.Builder()
             .name("Mode")
-            .description("Specifies whether the content should be encoded or decoded")
+            .description("Specifies whether the content should be encoded or decoded.")
             .required(true)
             .allowableValues(EncodingMode.class)
             .defaultValue(EncodingMode.ENCODE)
@@ -71,10 +71,10 @@ public class EncodeContent extends AbstractProcessor {
 
     public static final PropertyDescriptor ENCODING = new PropertyDescriptor.Builder()
             .name("Encoding")
-            .description("Specifies the type of encoding used")
+            .description("Specifies the type of encoding used.")
             .required(true)
             .allowableValues(EncodingType.class)
-            .defaultValue(EncodingType.BASE64_ENCODING)
+            .defaultValue(EncodingType.BASE64)
             .build();
 
     static final PropertyDescriptor LINE_OUTPUT_MODE = new PropertyDescriptor.Builder()
@@ -86,7 +86,7 @@ public class EncodeContent extends AbstractProcessor {
             .allowableValues(LineOutputMode.class)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .dependsOn(MODE, EncodingMode.ENCODE)
-            .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
+            .dependsOn(ENCODING, EncodingType.BASE64, EncodingType.BASE32)
             .build();
 
     static final PropertyDescriptor ENCODED_LINE_SEPARATOR = new PropertyDescriptor.Builder()
@@ -100,7 +100,7 @@ public class EncodeContent extends AbstractProcessor {
         .defaultValue(System.lineSeparator())
         .addValidator(Validator.VALID)
         .dependsOn(MODE, EncodingMode.ENCODE)
-        .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
+        .dependsOn(ENCODING, EncodingType.BASE64, EncodingType.BASE32)
         .build();
 
     static final PropertyDescriptor ENCODED_LINE_LENGTH = new PropertyDescriptor.Builder()
@@ -114,7 +114,7 @@ public class EncodeContent extends AbstractProcessor {
         .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
         .addValidator(StandardValidators.INTEGER_VALIDATOR)
         .dependsOn(MODE, EncodingMode.ENCODE)
-        .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
+        .dependsOn(ENCODING, EncodingType.BASE64, EncodingType.BASE32)
         .dependsOn(LINE_OUTPUT_MODE, LineOutputMode.MULTIPLE_LINES)
         .build();
 
@@ -153,7 +153,7 @@ public class EncodeContent extends AbstractProcessor {
         }
 
         final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
-        final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
+        final EncodingType encoding = getEncodingTypeFromString(context.getProperty(ENCODING).getValue());
         final boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
         final int lineLength = singleLineOutput ? -1 : context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
         final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
@@ -175,9 +175,9 @@ public class EncodeContent extends AbstractProcessor {
     private static StreamCallback getStreamCallback(final boolean encode, final EncodingType encoding,
         final int lineLength, final String lineSeparator) {
         switch(encoding) {
-            case BASE64_ENCODING:
+            case BASE64:
                 return encode ? new EncodeBase64(lineLength, lineSeparator) : new DecodeBase64();
-            case BASE32_ENCODING:
+            case BASE32:
                 return encode ? new EncodeBase32(lineLength, lineSeparator) : new DecodeBase32();
             default:
                 return encode ? new EncodeHex() : new DecodeHex();
@@ -295,6 +295,16 @@ public class EncodeContent extends AbstractProcessor {
                 }
             }
             out.flush();
+        }
+    }
+
+    private static EncodingType getEncodingTypeFromString(final String encodingTypeValue) {
+        if (encodingTypeValue.compareToIgnoreCase("base64") == 0) {
+            return EncodingType.BASE64;
+        } else if (encodingTypeValue.compareToIgnoreCase("base32") == 0) {
+            return EncodingType.BASE32;
+        } else {
+            return EncodingType.HEXADECIMAL;
         }
     }
 }
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -37,7 +37,6 @@ import org.apache.nifi.annotation.behavior.SupportsBatching;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.Tags;
 import org.apache.nifi.components.PropertyDescriptor;
-import org.apache.nifi.components.Validator;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.processor.AbstractProcessor;
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -78,7 +78,7 @@ public class EncodeContent extends AbstractProcessor {
             .build();
 
     static final PropertyDescriptor LINE_OUTPUT_MODE = new PropertyDescriptor.Builder()
-            .name("line-output-mode")
+            .name("Line Output Mode")
             .displayName("Line Output Mode")
             .description("If set to 'single-line', the encoded FlowFile content will output as a single line. If set to 'multiple-lines', "
                 + "it will output as multiple lines. This property is only applicable when Base64 or Base32 encoding is selected.")
@@ -91,7 +91,7 @@ public class EncodeContent extends AbstractProcessor {
             .build();
 
     static final PropertyDescriptor ENCODED_LINE_SEPARATOR = new PropertyDescriptor.Builder()
-        .name("line-separator")
+        .name("Encoded Content Line Separator")
         .displayName("Encoded Content Line Separator")
         .description("Each line of encoded data will be terminated with this byte sequence (e.g. \\r\\n"
                 + "). This property defaults to the system-dependent line separator string.  If `line-length` <= 0, "
@@ -105,7 +105,7 @@ public class EncodeContent extends AbstractProcessor {
         .build();
 
     static final PropertyDescriptor ENCODED_LINE_LENGTH = new PropertyDescriptor.Builder()
-        .name("encoded-line-length")
+        .name("Encoded Content Line Length")
         .displayName("Encoded Content Line Length")
         .description("Each line of encoded data will contain `encoded-line-length` characters (rounded down to the nearest multiple of 4). "
             + "If `encoded-line-length` <= 0, the encoded data is not divided into lines. This property is "
@@ -155,8 +155,7 @@ public class EncodeContent extends AbstractProcessor {
 
         final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
         final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
-
-        final Boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
+        final boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
         final int lineLength = context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
         final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
 
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -66,7 +66,7 @@ public class EncodeContent extends AbstractProcessor {
             .description("Specifies whether the content should be encoded or decoded")
             .required(true)
             .allowableValues(EncodingMode.class)
-            .defaultValue(EncodingMode.ENCODE.getValue())
+            .defaultValue(EncodingMode.ENCODE)
             .build();
 
     public static final PropertyDescriptor ENCODING = new PropertyDescriptor.Builder()
@@ -74,7 +74,7 @@ public class EncodeContent extends AbstractProcessor {
             .description("Specifies the type of encoding used")
             .required(true)
             .allowableValues(EncodingType.class)
-            .defaultValue(EncodingType.BASE64_ENCODING.getValue())
+            .defaultValue(EncodingType.BASE64_ENCODING)
             .build();
 
     static final PropertyDescriptor LINE_OUTPUT_MODE = new PropertyDescriptor.Builder()
@@ -83,7 +83,7 @@ public class EncodeContent extends AbstractProcessor {
             .description("If set to 'single-line', the encoded FlowFile content will output as a single line. If set to 'multiple-lines', "
                 + "it will output as multiple lines. This property is only applicable when Base64 or Base32 encoding is selected.")
             .required(false)
-            .defaultValue(LineOutputMode.SINGLE_LINE.getValue())
+            .defaultValue(LineOutputMode.SINGLE_LINE)
             .allowableValues(LineOutputMode.class)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .dependsOn(MODE, EncodingMode.ENCODE)
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -80,8 +80,7 @@ public class EncodeContent extends AbstractProcessor {
     static final PropertyDescriptor LINE_OUTPUT_MODE = new PropertyDescriptor.Builder()
             .name("Line Output Mode")
             .displayName("Line Output Mode")
-            .description("If set to 'single-line', the encoded FlowFile content will output as a single line. If set to 'multiple-lines', "
-                + "it will output as multiple lines. This property is only applicable when Base64 or Base32 encoding is selected.")
+            .description("Controls the line formatting for encoded content based on selected property values.")
             .required(false)
             .defaultValue(LineOutputMode.SINGLE_LINE)
             .allowableValues(LineOutputMode.class)
@@ -190,7 +189,7 @@ public class EncodeContent extends AbstractProcessor {
         private int lineLength;
         private String lineSeparator;
 
-        public EncodeBase64(final int lineLength,
+        private EncodeBase64(final int lineLength,
             final String lineSeparator) {
             this.lineLength = lineLength;
             this.lineSeparator = lineSeparator;
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -156,10 +156,9 @@ public class EncodeContent extends AbstractProcessor {
         final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
         final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
         final boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
-        final int lineLength = context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
+        final int lineLength = singleLineOutput ? -1 : context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
         final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
-
-        final StreamCallback callback = getStreamCallback(encode, encoding, (singleLineOutput ? -1 : lineLength), lineSeparator);
+        final StreamCallback callback = getStreamCallback(encode, encoding, lineLength, lineSeparator);
 
         try {
             final StopWatch stopWatch = new StopWatch(true);
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
@@ -159,7 +159,7 @@ public class EncodeContent extends AbstractProcessor {
         final int lineLength = context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
         final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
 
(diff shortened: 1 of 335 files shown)


knguyen1:
FYI: @exceptionfactory  @dan-s1

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/encoding/EncodingMode.java
```diff
@@ -0,0 +1,56 @@
...
+        public String getDisplayName() {
+            return "Decode";
+        }
+
+        @Override
+        public String getDescription() {
+            return "Sets the operation mode to 'decode'.";
+        }
+    }
+}
```
dan-s1, line 56: While how you defined this enum is not incorrect it is not the standard way enums which extend `DescribedValue` are implemented in the code base. I will demonstrate for this class and similar changes should be made for the `EncodingType` and `LineOutputMode` classes. For the `value` you can use `name()` method which should simplify your if statements and obviate the need to use `equalIgnoreCase`.

```suggestion
public enum EncodingMode implements DescribedValue {
     ENCODE("Encode", "Sets the operation mode to 'encode'."),
     DECODE("Decode", "Sets the operation mode to 'decode'.");
     
     EncodingMode(String displayName, String description) {
            this.displayName = displayName;
            this.description = description;
      }
     
      private final String displayName;
      private final String description;

      @Override
       public String getValue() {
            return name();
       }

       @Override
       public String getDisplayName() {
            return displayName;
       }

       @Override
       public String getDescription() {
            return description;
       }
}
```

dan-s1: Please do not forget to make similar changes for `EncodingType` and `LineOutputMode` and rework the logic in EncodeContent.

knguyen1: Done.  Please check now.

dan-s1: @knguyen1 Please only squash on the initial commit. After the review starts, only additional commits should be added without squashing to allow for tracking. Thanks!

knguyen1: Whooops, sorry. Was following the "only 1 commit rule"--didn't know about wanting to squash.  Next time...

dan-s1:
@knguyen1 It looks like there are Checkstyle violations that must be corrected.

knguyen1:
> @knguyen1 It looks like there are Checkstyle violations that must be corrected.

Fixed. [95f6c0a](https://github.com/apache/nifi/pull/8417/commits/95f6c0a9e38a49bea4fc7194c329ec715d79577d)

knguyen1:
@dan-s1 I just realised our change to `enum` introduces a breaking change.

```
case "BASE64_ENCODING":
case "BASE32_ENCODING":
```

It's because we're using `name()` from the `enum`.  Thoughts on overriding `getValue()` and return the old `string` values instead?

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -112,9 +153,14 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
             return;
         }
 
-        final boolean encode = context.getProperty(MODE).getValue().equalsIgnoreCase(ENCODE_MODE);
+        final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
         final String encoding = context.getProperty(ENCODING).getValue();
```
dan-s1, line 157:
```suggestion
               final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -129,31 +175,35 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-            } else {
-                return new DecodeHex();
-            }
+    private static StreamCallback getStreamCallback(final Boolean encode, final String encoding,
+        final int lineLength, final String lineSeparator) {
+        switch(encoding) {
+            case "BASE64_ENCODING":
+                return encode ? new EncodeBase64(lineLength, lineSeparator) : new DecodeBase64();
+            case "BASE32_ENCODING":
+                return encode ? new EncodeBase32(lineLength, lineSeparator) : new DecodeBase32();
```
dan-s1, line 184:
```suggestion
    private static StreamCallback getStreamCallback(final Boolean encode, final EncodingType encoding,
        final int lineLength, final String lineSeparator) {
        switch(encoding) {
            case BASE64_ENCODING:
                return encode ? new EncodeBase64(lineLength, lineSeparator) : new DecodeBase64();
            case BASE32_ENCODING:
                return encode ? new EncodeBase32(lineLength, lineSeparator) : new DecodeBase32();
```

dan-s1 (commented):
The suggested changes addresses your latest question

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
knguyen1: Interesting.  Okay, didn't know enums behaved that way.  Committing.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -112,9 +153,14 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
             return;
         }
 
-        final boolean encode = context.getProperty(MODE).getValue().equalsIgnoreCase(ENCODE_MODE);
-        final String encoding = context.getProperty(ENCODING).getValue();
-        final StreamCallback callback = getStreamCallback(encode, encoding);
+        final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
+               final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
```
dan-s1, line 157: Please align the corrected variable with the others
```suggestion
        final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
        final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -16,65 +16,50 @@
...
     private static final Path FILE_PATH = Paths.get("src/test/resources/hello.txt");
+    private TestRunner testRunner;
+
+    @BeforeEach
+    public void setUp() {
+        testRunner = TestRunners.newTestRunner(EncodeContent.class);
+    }
 
     @Test
     public void testBase64RoundTrip() throws IOException {
```
dan-s1, line 50: Since you are adding other tests which do not have `public`, please remove `public` from this test and any other test which has `public` since for JUnit 5 `public` is not needed.
```suggestion
    void testBase64RoundTrip() throws IOException {
```

dan-s1 (commented):
Hopefully we are getting closer. I still need to look over the unit tests but here are some formatting suggestions. Now with Junit5 none of the test methods need to be public so the `public` modifier should be removed from the tests similar to the ones you added. In addition the standard in the code base is to have the `@Test` annotation above the test method signature so most of the current suggestions are to align with that . Thanks!

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+        testRunner.setProperty(EncodeContent.LINE_OUTPUT_MODE, outputToSingleLine.getValue());
+        testRunner.setProperty(EncodeContent.ENCODED_LINE_LENGTH, Integer.toString(lineLength));
+        testRunner.setProperty(EncodeContent.ENCODED_LINE_SEPARATOR, lineSeparator);
+
+        testRunner.enqueue(input);
         testRunner.run();
 
-        testRunner.assertAllFlowFilesTransferred(EncodeContent.REL_FAILURE, 1);
+        final MockFlowFile result = testRunner.getFlowFilesForRelationship(routedTo).get(0);
+        assertEquals(expectedOutput, result.getContent());
```
dan-s1, line 393: Intellij has the following issue 

> Call to 'assertEquals()' from 'Assert' should be replaced with a call to a method from 'org.junit.jupiter.api.Assertions'

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+    @Test void testBasicEncodeHex1() {
+        executeTestSuccessHelper(EncodingMode.ENCODE, EncodingType.HEX_ENCODING, "foo", "666F6F");
     }
 
-    @Test
-    public void testFailDecodeNotHex() throws IOException {
-        final TestRunner testRunner = TestRunners.newTestRunner(new EncodeContent());
+    @Test void testBasicEncodeBase320() {
+        executeTestSuccessHelper(EncodingMode.ENCODE, EncodingType.BASE32_ENCODING, "hello", "NBSWY3DP" + System.lineSeparator());
+    }
```
dan-s1, line 156: Instead of this test and the test on line 158-160 use JUnit 5 parameterize test
 ```suggestion
    @ParameterizedTest
    @MethodSource("encodeBase32Args")
    void testBasicEncodeBase32(String input, String expectedOutput) {
        executeTestSuccessHelper(EncodingMode.ENCODE, EncodingType.BASE32_ENCODING, input, expectedOutput);
    }
    
    private static Stream<Arguments> encodeBase32Args() {
        return Stream.of(
                Arguments.of("hello", "NBSWY3DP" + System.lineSeparator()),
                Arguments.of("foo", "MZXW6===" + System.lineSeparator())
        );
    }
```

Please remember to remove the test on lines 158-160 and include imports
```
import java.util.stream.Stream;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.Arguments;
import org.junit.jupiter.params.provider.MethodSource;
```

File: nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java [hunk 8]
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
 
-        testRunner.setProperty(EncodeContent.MODE, EncodeContent.DECODE_MODE);
-        testRunner.enqueue(flowFile);
-        testRunner.clearTransferState();
-        testRunner.run();
-        testRunner.assertAllFlowFilesTransferred(EncodeContent.REL_SUCCESS, 1);
+    @Test void testBasicEncodeHex0() {
+        executeTestSuccessHelper(EncodingMode.ENCODE, EncodingType.HEX_ENCODING, "hello", "68656C6C6F");
+    }
```
dan-s1, line 148: Replace this test and the test on lines 150-151 with a single parameterized test as was demonstrated.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+        testRunner.clearTransferState(); // clear the state for the next test
+        executeTestSuccessHelper(EncodingMode.DECODE, EncodingType.BASE64_ENCODING, expectedOutput, specialChars);
+    }
 
-        testRunner.enqueue(FILE_PATH);
-        testRunner.clearTransferState();
-        testRunner.run();
+    @Test void testBasicDecodeBase32() {
+        executeTestSuccessHelper(EncodingMode.DECODE, EncodingType.BASE32_ENCODING, "NBSWY3DP", "hello");
+    }
```
dan-s1, line 136: This test as well as tests on lines 138-140 and 142-144 should be combined into a single parameterized test.

dan-s1 (commented):
Further suggestions to refactor the multiple unit test into single parameterized tests.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+            "hello",
+            "aGVsbG8=" + System.lineSeparator());
+    }
+
+    @Test void testBasicEncodeBase641() {
+        executeTestSuccessHelper(EncodingMode.ENCODE,
+            EncodingType.BASE64_ENCODING,
+            "foo",
+            "Zm9v" + System.lineSeparator());
+    }
```
dan-s1, line 174: Refactor both of these tests to a single parameterized test.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -29,55 +37,88 @@
...
             .build();
 
     public static final PropertyDescriptor ENCODING = new PropertyDescriptor.Builder()
             .name("Encoding")
             .description("Specifies the type of encoding used")
             .required(true)
-            .allowableValues(BASE64_ENCODING, BASE32_ENCODING, HEX_ENCODING)
-            .defaultValue(BASE64_ENCODING)
+            .allowableValues(EncodingType.class)
+            .defaultValue(EncodingType.BASE64_ENCODING.getValue())
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/c5e0e4d97cddd2604e264f93ed30a8024ba4be67

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -29,55 +37,88 @@
...
+            .build();
+
+    static final PropertyDescriptor LINE_OUTPUT_MODE = new PropertyDescriptor.Builder()
+            .name("line-output-mode")
+            .displayName("Line Output Mode")
+            .description("If set to 'single-line', the encoded FlowFile content will output as a single line. If set to 'multiple-lines', "
+                + "it will output as multiple lines. This property is only applicable when Base64 or Base32 encoding is selected.")
+            .required(false)
+            .defaultValue(LineOutputMode.SINGLE_LINE.getValue())
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/c5e0e4d97cddd2604e264f93ed30a8024ba4be67

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -29,55 +37,88 @@
...
-    public static final String HEX_ENCODING = "hex";
-
     public static final PropertyDescriptor MODE = new PropertyDescriptor.Builder()
             .name("Mode")
             .description("Specifies whether the content should be encoded or decoded")
             .required(true)
-            .allowableValues(ENCODE_MODE, DECODE_MODE)
-            .defaultValue(ENCODE_MODE)
+            .allowableValues(EncodingMode.class)
+            .defaultValue(EncodingMode.ENCODE.getValue())
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/c5e0e4d97cddd2604e264f93ed30a8024ba4be67

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
         testRunner.assertAllFlowFilesTransferred(EncodeContent.REL_FAILURE, 1);
     }
 
-    @Test
-    public void testHexRoundTrip() throws IOException {
-        final TestRunner testRunner = TestRunners.newTestRunner(new EncodeContent());
+    @Test void testEncodeDecodeSpecialCharsBase64() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+    }
+
+    @Test void testBlankValueShouldNotFail() {
+        executeTestSuccessHelper(EncodingMode.ENCODE,
+            EncodingType.BASE64_ENCODING,
+            StringUtils.EMPTY,
+            StringUtils.EMPTY);
+    }
+
+    @Test void testEncodeContentMultipleLinesBase64() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+        // Execute the test using the helper method
+        executeTestHelper(EncodingMode.ENCODE,
+            EncodingType.BASE64_ENCODING,
+            LOREM_IPSUM,
+            LineOutputMode.MULTIPLE_LINES,
+            expectedOutput,
+            EncodeContent.REL_SUCCESS);
+    }
+
+    @Test void testEncodeContentSingleLineBase64() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+        // Execute the test using the helper method
+        executeTestHelper(EncodingMode.ENCODE,
+            EncodingType.BASE64_ENCODING,
+            LOREM_IPSUM,
+            LineOutputMode.SINGLE_LINE,
+            expectedOutput,
+            EncodeContent.REL_SUCCESS);
+    }
+
+    @Test void testEncodeContentSingleLineBase32() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+        // Execute the test using the helper method
+        executeTestHelper(EncodingMode.ENCODE,
+            EncodingType.BASE32_ENCODING,
+            LOREM_IPSUM,
+            LineOutputMode.SINGLE_LINE,
+            expectedOutput,
+            EncodeContent.REL_SUCCESS);
+    }
+
+    @Test void testEncodeContentMultipleLinesBase32() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+        // Execute the test using the helper method
+        executeTestHelper(EncodingMode.ENCODE,
+            EncodingType.BASE32_ENCODING,
+            LOREM_IPSUM,
+            LineOutputMode.MULTIPLE_LINES, // set false to output multiple lines
+            expectedOutput,
+            EncodeContent.REL_SUCCESS);
+    }
+
+    @Test void testEncodeContentMultipleLinesNonStandardLengthBase32() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+            EncodingType.BASE32_ENCODING,
+            LOREM_IPSUM,
+            LineOutputMode.MULTIPLE_LINES,
+            80,
+            System.lineSeparator(),
+            expectedOutput,
+            EncodeContent.REL_SUCCESS);
+    }
+
+    @Test void testThatLineLengthIsIgnoredIfSingleLineOutputTrueBase32() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+            EncodingType.BASE64_ENCODING,
+            LOREM_IPSUM,
+            LineOutputMode.MULTIPLE_LINES, // set false to output multiple lines
+            80,
+            System.lineSeparator(),
+            expectedOutput,
+            EncodeContent.REL_SUCCESS);
+    }
+
+    @Test void testOverrideLineSeparatorBase64() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+            EncodingType.BASE64_ENCODING,
+            LOREM_IPSUM,
+            LineOutputMode.MULTIPLE_LINES, // set false to output multiple lines
+            80,
+            "|",
+            expectedOutput,
+            EncodeContent.REL_SUCCESS);
+    }
+
+    @Test void testOverrideLineSeparatorBase32() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -123,43 +122,275 @@ public void testFailDecodeNotBase32() throws IOException {
...
+            EncodingType.BASE32_ENCODING,
+            LOREM_IPSUM,
+            LineOutputMode.MULTIPLE_LINES, // set false to output multiple lines
+            80,
+            "|",
+            expectedOutput,
+            EncodeContent.REL_SUCCESS);
+    }
+
+    @Test void testThatLineLengthIsIgnoredIfSingleLineOutputTrueBase64() {
```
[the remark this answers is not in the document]
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
knguyen1: Fixed https://github.com/apache/nifi/pull/8417/commits/8c16ea83e7ab83c19f4e4f685b0b30dd0d5edc50

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
knguyen1: https://github.com/apache/nifi/pull/8417/commits/4e7732fccb2b70b8f44fb571c0d3d86e8320fb40

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
knguyen1: Refactored https://github.com/apache/nifi/pull/8417/commits/9fd11b7ae031fe416783f0135b7a94752b534ade

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
knguyen1: Refactored https://github.com/apache/nifi/pull/8417/commits/ed7bd5ad7d3c0072d0eb839f1897c993d6730666

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
knguyen1: Refactored https://github.com/apache/nifi/pull/8417/commits/ed7bd5ad7d3c0072d0eb839f1897c993d6730666

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
knguyen1: Refactored https://github.com/apache/nifi/pull/8417/commits/a4ee842160677e6cefc544bdab2995700ca26bd2

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -16,25 +16,85 @@
...
+
+    @Test
+    void testBase64RoundTrip() throws IOException {
+        runTestRoundTrip(EncodingType.BASE64_ENCODING.getValue());
+    }
+
+    @Test
+    void testFailDecodeNotBase64() throws IOException {
+        runTestDecodeFailure(EncodingType.BASE64_ENCODING.getValue());
+    }
```
dan-s1, line 61: We are getting closer. We can collapse six of the unit tests into two parameterized tests and remove two methods.
Tests`testBase64RoundTrip(), testBase32RoundTrip() and testHexRoundTrip()` can be collapsed into a paramaterized test with the body from `runTestRoundTrip`. In this case you would use the `@EnumSource` annotation for the argument.

```
    @ParameterizedTest
    @EnumSource(value = EncodingType.class)
    void testRoundTrip(EncodingType encoding) throws IOException {
        testRunner.setProperty(EncodeContent.MODE, EncodingMode.ENCODE);
        testRunner.setProperty(EncodeContent.ENCODING, encoding);

        testRunner.enqueue(FILE_PATH);
        testRunner.clearTransferState();
        testRunner.run();

        testRunner.assertAllFlowFilesTransferred(EncodeContent.REL_SUCCESS, 1);

        MockFlowFile flowFile = 
 testRunner.getFlowFilesForRelationship(EncodeContent.REL_SUCCESS).get(0);
        testRunner.assertQueueEmpty();

        testRunner.setProperty(EncodeContent.MODE, EncodingMode.DECODE.getValue());
        testRunner.enqueue(flowFile);
        testRunner.clearTransferState();
        testRunner.run();
        testRunner.assertAllFlowFilesTransferred(EncodeContent.REL_SUCCESS, 1);

        flowFile = testRunner.getFlowFilesForRelationship(EncodeContent.REL_SUCCESS).get(0);
        flowFile.assertContentEquals(FILE_PATH);
    }
```

Please add this and implement similar code for `testFailDecodeNotHex(), testFailDecodeNotBase32(), testFailDecodeNotHex() and runTestDecodeFailure`.
Also do not forget the import
`import org.junit.jupiter.params.provider.EnumSource;`

knguyen1: Refactored and collapsed: https://github.com/apache/nifi/pull/8417/commits/e994e3c46c3dae2fb4bed944826a22c2bb5fc0e3 and https://github.com/apache/nifi/pull/8417/commits/1a15112f86842f64ecf72e56371bef62c685614b

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -29,55 +37,88 @@
...
             .description("Specifies the type of encoding used")
             .required(true)
-            .allowableValues(BASE64_ENCODING, BASE32_ENCODING, HEX_ENCODING)
-            .defaultValue(BASE64_ENCODING)
+            .allowableValues(EncodingType.class)
+            .defaultValue(EncodingType.BASE64_ENCODING)
+            .build();
+
+    static final PropertyDescriptor LINE_OUTPUT_MODE = new PropertyDescriptor.Builder()
+            .name("line-output-mode")
```
dan-s1, line 81:
```suggestion
            .name("Line Output Mode")
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -29,55 +37,88 @@
...
+            .required(false)
+            .defaultValue(LineOutputMode.SINGLE_LINE)
+            .allowableValues(LineOutputMode.class)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .dependsOn(MODE, EncodingMode.ENCODE)
+            .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
             .build();
 
+    static final PropertyDescriptor ENCODED_LINE_SEPARATOR = new PropertyDescriptor.Builder()
+        .name("line-separator")
```
dan-s1, line 94:
```suggestion
        .name("Encoded Content Line Separator")
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -29,55 +37,88 @@
...
+        .required(false)
+        .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+        .defaultValue(System.lineSeparator())
+        .addValidator(Validator.VALID)
+        .dependsOn(MODE, EncodingMode.ENCODE)
+        .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
+        .build();
+
+    static final PropertyDescriptor ENCODED_LINE_LENGTH = new PropertyDescriptor.Builder()
+        .name("encoded-line-length")
```
dan-s1, line 108:
```suggestion
        .name("Encoded Content Line Length")
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -112,9 +153,14 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
             return;
         }
 
-        final boolean encode = context.getProperty(MODE).getValue().equalsIgnoreCase(ENCODE_MODE);
-        final String encoding = context.getProperty(ENCODING).getValue();
-        final StreamCallback callback = getStreamCallback(encode, encoding);
+        final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
+        final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
+
+        final Boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
```
dan-s1, line 159:
```suggestion
        final boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -45,7 +78,7 @@ public void testBase64RoundTrip() throws IOException {
         MockFlowFile flowFile = testRunner.getFlowFilesForRelationship(EncodeContent.REL_SUCCESS).get(0);
         testRunner.assertQueueEmpty();
 
-        testRunner.setProperty(EncodeContent.MODE, EncodeContent.DECODE_MODE);
+        testRunner.setProperty(EncodeContent.MODE, EncodingMode.DECODE.getValue());
```
dan-s1, line 81:
```suggestion
        testRunner.setProperty(EncodeContent.MODE, EncodingMode.DECODE);
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -55,12 +88,11 @@ public void testBase64RoundTrip() throws IOException {
...
-    @Test
-    public void testFailDecodeNotBase64() throws IOException {
-        final TestRunner testRunner = TestRunners.newTestRunner(new EncodeContent());
-
-        testRunner.setProperty(EncodeContent.MODE, EncodeContent.DECODE_MODE);
-        testRunner.setProperty(EncodeContent.ENCODING, EncodeContent.BASE64_ENCODING);
+    @ParameterizedTest
+    @EnumSource(value = EncodingType.class)
+    void testDecodeFailure(final EncodingType encoding) throws IOException {
+        testRunner.setProperty(EncodeContent.MODE, EncodingMode.DECODE.getValue());
```
dan-s1, line 94:
```suggestion
        testRunner.setProperty(EncodeContent.MODE, EncodingMode.DECODE);
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -55,12 +88,11 @@ public void testBase64RoundTrip() throws IOException {
...
+        testRunner.setProperty(EncodeContent.ENCODING, encoding.getValue());
```
dan-s1, line 95:
```suggestion
        testRunner.setProperty(EncodeContent.ENCODING, encoding);
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -70,96 +102,310 @@ public void testFailDecodeNotBase64() throws IOException {
...
+        final String input,
+        final DescribedValue outputToSingleLine,
+        final Integer lineLength,
+        final String lineSeparator,
+        final String expectedOutput,
+        final Relationship routedTo) {
+
+        testRunner.setProperty(EncodeContent.MODE, mode.getValue());
+        testRunner.setProperty(EncodeContent.ENCODING, encodingType.getValue());
+        testRunner.setProperty(EncodeContent.LINE_OUTPUT_MODE, outputToSingleLine.getValue());
```
dan-s1, line 400:
```suggestion
        testRunner.setProperty(EncodeContent.MODE, mode);
        testRunner.setProperty(EncodeContent.ENCODING, encodingType);
        testRunner.setProperty(EncodeContent.LINE_OUTPUT_MODE, outputToSingleLine);
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -16,25 +16,58 @@
...
+    void setUp() {
+        testRunner = TestRunners.newTestRunner(EncodeContent.class);
+    }
 
     @Test
-    public void testBase64RoundTrip() throws IOException {
-        final TestRunner testRunner = TestRunners.newTestRunner(new EncodeContent());
+    void testFailDecodeNotBase64ButIsAMultipleOfFourBytes() {
+        testRunner.setProperty(EncodeContent.MODE, EncodingMode.DECODE.getValue());
+        testRunner.setProperty(EncodeContent.ENCODING, EncodingType.BASE64_ENCODING.getValue());
```
dan-s1, line 57:
```suggestion
        testRunner.setProperty(EncodeContent.MODE, EncodingMode.DECODE);
        testRunner.setProperty(EncodeContent.ENCODING, EncodingType.BASE64_ENCODING);
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestEncodeContent.java
```diff
@@ -16,25 +16,58 @@
...
+        testRunner.assertAllFlowFilesTransferred(EncodeContent.REL_FAILURE, 1);
+    }
 
-        testRunner.setProperty(EncodeContent.MODE, EncodeContent.ENCODE_MODE);
-        testRunner.setProperty(EncodeContent.ENCODING, EncodeContent.BASE64_ENCODING);
+    @ParameterizedTest
+    @EnumSource(value = EncodingType.class)
+    void testRoundTrip(final EncodingType encoding) throws IOException {
+        testRunner.setProperty(EncodeContent.MODE, EncodingMode.ENCODE.getValue());
+        testRunner.setProperty(EncodeContent.ENCODING, encoding.getValue());
```
dan-s1, line 70:
```suggestion
        testRunner.setProperty(EncodeContent.MODE, EncodingMode.ENCODE);
        testRunner.setProperty(EncodeContent.ENCODING, encoding);
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -112,9 +153,14 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
             return;
         }
 
-        final boolean encode = context.getProperty(MODE).getValue().equalsIgnoreCase(ENCODE_MODE);
-        final String encoding = context.getProperty(ENCODING).getValue();
-        final StreamCallback callback = getStreamCallback(encode, encoding);
+        final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
+        final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
+
```
dan-s1, line 158:
```suggestion
```

dan-s1 (commented):
A few more things which should be changed.

dan-s1:
@exceptionfactory When you get a chance can you please check things over to see if all necessary changes were made? Thanks!

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -112,9 +153,13 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-        final boolean encode = context.getProperty(MODE).getValue().equalsIgnoreCase(ENCODE_MODE);
-        final String encoding = context.getProperty(ENCODING).getValue();
-        final StreamCallback callback = getStreamCallback(encode, encoding);
+        final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
+        final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
+        final boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
+        final int lineLength = context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
+        final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
+
+        final StreamCallback callback = getStreamCallback(encode, encoding, Boolean.TRUE.equals(singleLineOutput) ? -1 : lineLength, lineSeparator);
```
dan-s1, line 162:
```suggestion
        final StreamCallback callback = getStreamCallback(encode, encoding, (singleLineOutput ? -1 : lineLength), lineSeparator);
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -129,31 +175,35 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-            }
-        } else {
-            if (encoding.equalsIgnoreCase(BASE64_ENCODING)) {
-                return new DecodeBase64();
-            } else if (encoding.equalsIgnoreCase(BASE32_ENCODING)) {
-                return new DecodeBase32();
-            } else {
-                return new DecodeHex();
-            }
+    private static StreamCallback getStreamCallback(final Boolean encode, final EncodingType encoding,
```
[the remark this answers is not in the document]
dan-s1: I do not see the suggested change in the commit listed above.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -112,9 +153,13 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-        final boolean encode = context.getProperty(MODE).getValue().equalsIgnoreCase(ENCODE_MODE);
-        final String encoding = context.getProperty(ENCODING).getValue();
-        final StreamCallback callback = getStreamCallback(encode, encoding);
+        final boolean encode = context.getProperty(MODE).getValue().equals(EncodingMode.ENCODE.getValue());
+        final EncodingType encoding = EncodingType.valueOf(context.getProperty(ENCODING).getValue());
+        final boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
+        final int lineLength = context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
+        final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
+
+        final StreamCallback callback = getStreamCallback(encode, encoding, (singleLineOutput ? -1 : lineLength), lineSeparator);
```
dan-s1, line 162: Sorry to drive you nuts on this but I thought the code reads better with the following changes (I did not realize this before).
```suggestion
        final boolean singleLineOutput = context.getProperty(LINE_OUTPUT_MODE).getValue().equals(LineOutputMode.SINGLE_LINE.getValue());
        final int lineLength = singleLineOutput ? -1 : context.getProperty(ENCODED_LINE_LENGTH).evaluateAttributeExpressions(flowFile).asInteger();
        final String lineSeparator = context.getProperty(ENCODED_LINE_SEPARATOR).evaluateAttributeExpressions(flowFile).getValue();
        final StreamCallback callback = getStreamCallback(encode, encoding, lineLength, lineSeparator);
```

exceptionfactory (changes requested):
Thanks for the work on these new properties @knguyen1, and thanks for the feedback thus far @dan-s1.

The general capability is useful, but there are still several refinements needed to maintain backward compatibility. I also recommended a few documentation changes. One particular note is the Line Separator property, which seems unnecessary in general. Unless there is a strong reason to support either LF or CRLF, just using LF across the board seems simpler. Sometimes less is more when it comes to configuration, but feel free to provide more details.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -29,55 +37,88 @@
...
+        .description("Each line of encoded data will be terminated with this byte sequence (e.g. \\r\\n"
+                + "). This property defaults to the system-dependent line separator string.  If `line-length` <= 0, "
+                + "the `line-separator` property is not used. This property is not used for `hex` encoding.")
+        .required(false)
+        .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+        .defaultValue(System.lineSeparator())
+        .addValidator(Validator.VALID)
+        .dependsOn(MODE, EncodingMode.ENCODE)
+        .dependsOn(ENCODING, EncodingType.BASE64_ENCODING, EncodingType.BASE32_ENCODING)
+        .build();
```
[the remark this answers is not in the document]
knguyen1: I exposed it because it's available in the class.  Need to solicit additional feedback from the community on its usefulness; otherwise can delete.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -29,55 +37,88 @@
...
+
+    static final PropertyDescriptor ENCODED_LINE_LENGTH = new PropertyDescriptor.Builder()
+        .name("Encoded Content Line Length")
+        .displayName("Encoded Content Line Length")
+        .description("Each line of encoded data will contain `encoded-line-length` characters (rounded down to the nearest multiple of 4). "
+            + "If `encoded-line-length` <= 0, the encoded data is not divided into lines. This property is "
+            + "ignored if `line-output-mode` is set to `single-line`.")
```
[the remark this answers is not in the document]
knguyen1: The "nearest multiple of 4" is straight from documentation.  Quote:

> If doEncode is true, each line of encoded data will contain lineLength characters (rounded down to the nearest multiple of 4). If lineLength <= 0, the encoded data is not divided into lines. If doEncode is false, lineLength is ignored.

Source: https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base32OutputStream.html#%3Cinit%3E(java.io.OutputStream,boolean,int,byte%5B%5D)
Source: https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64OutputStream.html#%3Cinit%3E(java.io.OutputStream,boolean,int,byte%5B%5D,org.apache.commons.codec.CodecPolicy)

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/encoding/EncodingType.java
```diff
@@ -0,0 +1,48 @@
...
+    BASE32_ENCODING("Base32", "Sets the encoding type to 'Base32'."),
+    HEX_ENCODING("Hexadecimal", "Sets the encoding type to 'Hexadecimal'.");
+
+    private final String displayName;
+    private final String description;
+
+    EncodingType(String displayName, String description) {
+        this.displayName = displayName;
+        this.description = description;
+    }
```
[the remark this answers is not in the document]
knguyen1: Added.  See: https://github.com/apache/nifi/pull/8417/commits/ecc32631d35eb10cef24c8d2a270051fdcabbfd8

knguyen1:
I fixed failing tests and build failures from the latest code review.  Everything's passing now. @exceptionfactory 

I had to add this method: `EncodingType getEncodingTypeFromString(final String encodingTypeValue)` because previous implementation used `hex` instead of `hexadecimal` as in @exceptionfactory's suggestion.  So we can't easily use `EncodingType.valueOf` anymore.

Latest is pending the resolution of comments regarding.

> Is this property necessary? It seems best to hard-code the line-feed character `\n`, rather than making this platform-dependent or configurable.

> This description needs some rewording and clarification. It is not clear why it is rounded to a multiple of 4, and again, the conditional nature of the property does not need to be stated.

I have left those threads unresolved.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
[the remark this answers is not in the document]
exceptionfactory: Thanks for the clarification, I didn't make that connection earlier when reviewing the implementation, but it makes sense now. Thanks for the background!

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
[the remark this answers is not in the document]
exceptionfactory: Thanks, I think in this case, the first version of this implementation could hard-code `\n` instead of making it a property. If later on, there is a clear need for this property, it would be added.

Tagging @dan-s1 for additional comment since he has provided feedback thus far.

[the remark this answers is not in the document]
knguyen1: @dan-s1 `Base32OutputStream` and `Base64OutputStream`.  Look for the argument `lineSeparator`.  The thinking at the time was: if it's available in the API, someone might find it useful.

Source: https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base32OutputStream.html#%3Cinit%3E(java.io.OutputStream,boolean,int,byte%5B%5D)
Source: https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64OutputStream.html#%3Cinit%3E(java.io.OutputStream,boolean,int,byte%5B%5D,org.apache.commons.codec.CodecPolicy)

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -237,4 +297,14 @@ public void process(final InputStream in, final OutputStream out) throws IOExcep
             out.flush();
         }
     }
+
+    private static EncodingType getEncodingTypeFromString(final String encodingTypeValue) {
```
dan-s1, line 301: No need to add suffix `FromString` as it is obvious from the argument.
```suggestion
    private static EncodingType getEncodingType(final String encodingTypeValue) {
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -237,4 +297,14 @@ public void process(final InputStream in, final OutputStream out) throws IOExcep
...
+        if (encodingTypeValue.compareToIgnoreCase("base64") == 0) {
+            return EncodingType.BASE64;
+        } else if (encodingTypeValue.compareToIgnoreCase("base32") == 0) {
+            return EncodingType.BASE32;
```
dan-s1, line 305:
```suggestion
        if (EncodingType.BASE64.getValue().equals(encodingTypeValue)) {
            return EncodingType.BASE64;
        } else if (EncodingType.BASE32.getValue().equals(encodingTypeValue)) {
            return EncodingType.BASE32;
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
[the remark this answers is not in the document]
dan-s1: @knguyen1 Thanks for pointing out those classes.
@exceptionfactory  Based on the above classes there is another property which I am not sure should be exposed or not and that is the decoding policy. There are two options lenient and strict. The default is lenient. The wording from the `Base32OutputStream `Javadoc is

> You can set the decoding behavior when the input bytes contain leftover trailing bits that cannot be created by a valid encoding. These can be bits that are unused from the final character or entire characters. The default mode is lenient decoding.
> 
> Lenient: Any trailing bits are composed into 8-bit bytes where possible. The remainder are discarded.
> Strict: The decoding will raise an [IllegalArgumentException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html) if trailing bits are not part of a valid encoding. Any unused bits from the final character must be zero. Impossible counts of entire final characters are not allowed.
> When strict decoding is enabled it is expected that the decoded bytes will be re-encoded to a byte array that matches the original, i.e. no changes occur on the final character. This requires that the input bytes use the same padding and alphabet as the encoder.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -129,31 +172,35 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
+                return encode ? new EncodeBase32(lineLength, lineSeparator) : new DecodeBase32();
+            default:
+                return encode ? new EncodeHex() : new DecodeHex();
         }
     }
 
     private static class EncodeBase64 implements StreamCallback {
 
+        private int lineLength;
+        private String lineSeparator;
```
dan-s1, line 190: Intellij points out these should be final
```suggestion
       private final int lineLength;
       private final String lineSeparator;
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
```diff
@@ -171,9 +218,22 @@ public void process(final InputStream in, final OutputStream out) throws IOExcep
 
     private static class EncodeBase32 implements StreamCallback {
 
+        private int lineLength;
+        private String lineSeparator;
```
dan-s1, line 222:
```suggestion
        private final int lineLength;
        private final String lineSeparator;
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/encoding/EncodingMode.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package org.apache.nifi.processors.standard.encoding;
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum EncodingMode implements DescribedValue {
+     ENCODE("Encode", "Transform original input to encoded representation"),
+     DECODE("Decode", "Transform encoded input to original representation");
+
+     EncodingMode(String value, String description) {
```
exceptionfactory, line 25: The spacing in this file appears to be off in several places. Each block should be indented with a multiple of four spaces.

exceptionfactory (commented):
Thanks for updating the property values @knguyen1.

The last remaining item to change is the `Encoded Content Line Separator` property. I recommend removing the property and using a hard-coded `\n` value so that the behavior is the same regardless of the operating system on which NiFi is running.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/encoding/EncodingMode.java
knguyen1: Fixed indentations: https://github.com/apache/nifi/pull/8417/commits/b8df280eb8bc79c78fa0eb91fbba94cd7ceedd95

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/EncodeContent.java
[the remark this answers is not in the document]
knguyen1: @exceptionfactory [Remove ENCODED_LINE_SEPARATOR property and tests](https://github.com/apache/nifi/pull/8417/commits/ce32323242615a93e70c22a51ba00ce4b7d84a08)

exceptionfactory (commented):
Thanks for working through the feedback @knguyen1. One more final review, I adjusted the new properties to be required, since the default values align with current behavior. I also set the `Encoded Line Length` property to use the Positive Integer Validator, ensuring that the value will never be zero or lower.

With these changes, I plan to merge soon, pending one more successful build.

exceptionfactory (approved):
Thanks again for taking the time to work through all the feedback @knguyen1! +1 merging

exceptionfactory:
Thanks for the reviews as well @dan-s1!

PR #8442, opened 2024-02-21
NIFI-12832: Eliminated unnecessary dependencies from nifi-mock; moved…

markap14:
… StandardValidators to nifi-api; broke apart FormatUtils into FormatUtils and TimeFormat classes, with TimeFormat existing in nifi-api

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


nifi-api/src/main/java/org/apache/nifi/time/TimeFormat.java
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        final StringBuilder sb = new StringBuilder();
+        sb.append(values[0]);
+        for (int i = 1; i < values.length; i++) {
+            sb.append(delimiter).append(values[i]);
+        }
+
+        return sb.toString();
+    }
```
dan-s1, line 249: Does this method do something different than `String.join` ? If not I would suggest replacing the use of this with `String.join`.

nifi-api/src/main/java/org/apache/nifi/processor/util/StandardValidators.java
```diff
@@ -843,7 +843,7 @@ public ValidationResult validate(final String subject, final String input, final
             final boolean validSyntax = pattern.matcher(lowerCase).matches();
             final ValidationResult.Builder builder = new ValidationResult.Builder();
             if (validSyntax) {
-                final long nanos = FormatUtils.getTimeDuration(lowerCase, TimeUnit.NANOSECONDS);
+                final long nanos = new TimeFormat().getTimeDuration(lowerCase, TimeUnit.NANOSECONDS);
```
dan-s1, line 846: `TimeFormat`, does not have any instance variables, Couldn't all the methods be static hence the constructor for it should be private and no instantiation should be allowed.

nifi-api/src/main/java/org/apache/nifi/processor/util/StandardValidators.java
```diff
@@ -540,9 +540,9 @@ public ValidationResult validate(final String subject, final String input, final
                 }
 
                 try {
+                    // Check that we can parse the value as a URL
                     final String evaluatedInput = context.newPropertyValue(input).evaluateAttributeExpressions().getValue();
-                    final URI uri = UriUtils.create(evaluatedInput);
-                    uri.toURL();
+                    URI.create(evaluatedInput).toURL();
```
dan-s1, line 545: Lines 544-545 which were removed were meant to fix the bug discovered in NIFI-12513

nifi-api/src/main/java/org/apache/nifi/time/TimeFormat.java
markap14: Yeah, good point. I pulled it over directly from FormatUtils, but I think now that we depend on Java 21 we can just use String.join

nifi-api/src/main/java/org/apache/nifi/processor/util/StandardValidators.java
markap14: Yeah, I intentionally didn't make the method static, because I was thinking there were more methods, some of which might benefit from providing a timezone, etc. But now that the refactoring is complete, the methods are really only about durations. Probably makes sense to use static methods and name it `DurationFormat` rather than `TimeFormat`.

nifi-api/src/main/java/org/apache/nifi/processor/util/StandardValidators.java
markap14: Got it. I saw it was used in InvokeHTTP but didn't think it made sense for validation. But I guess it does.

markap14:
Thanks @dan-s1 for catching my oversight on the URL Validator issue. I restored the logic from UriUtils, but we don't want to have Utils types of classes in the API module. Because of that, I just moved the logic into the URLValidator class, so that it becomes part of the Validator itself.

nifi-api/src/main/java/org/apache/nifi/components/ValidationContext.java
```diff
@@ -21,9 +21,13 @@
 import org.apache.nifi.controller.ControllerService;
 import org.apache.nifi.controller.ControllerServiceLookup;
 import org.apache.nifi.expression.ExpressionLanguageCompiler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
exceptionfactory, line 25: This introduces a dependency on `slf4j-api` in `nifi-api`, which is coming through right now from `slf4j-simple`. As all of the logging statements are for debugging, it seems like the most straightforward option is to remove the debug logging.

nifi-api/src/test/java/org/apache/nifi/time/TestTimeFormat.java
```diff
@@ -0,0 +1,278 @@
...
+
+import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class TestTimeFormat {
```
exceptionfactory, line 36: This should be renamed to `TestDurationFormat`.

PR #8445, opened 2024-02-22
NIFI-12828 Add mapping  for BIT SQL Type in DataTypeUtils

ravinarayansingh:
# Summary

[NIFI-12828](https://issues.apache.org/jira/browse/NIFI-12828)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
@@ -1917,8 +1917,9 @@ public class DataTypeUtils {
         switch (sqlType) {
             case Types.BIGINT:
                 return RecordFieldType.BIGINT.getDataType();
-            case Types.BOOLEAN:
             case Types.BIT:
+                return RecordFieldType.INT.getDataType();
+            case Types.BOOLEAN:
                 return RecordFieldType.BOOLEAN.getDataType();
             case Types.TINYINT:
                 return RecordFieldType.BYTE.getDataType();
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
@@ -1918,6 +1918,7 @@ public class DataTypeUtils {
             case Types.BIGINT:
                 return RecordFieldType.BIGINT.getDataType();
             case Types.BOOLEAN:
+            case Types.BIT:
                 return RecordFieldType.BOOLEAN.getDataType();
             case Types.TINYINT:
                 return RecordFieldType.BYTE.getDataType();
--- a/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
@@ -558,6 +558,7 @@ public class TestDataTypeUtils {
     @Test
     public void testGetDataTypeFromSQLTypeValue() {
         assertEquals(RecordFieldType.STRING.getDataType(), DataTypeUtils.getDataTypeFromSQLTypeValue(Types.CLOB));
+        assertEquals(RecordFieldType.BOOLEAN.getDataType(), DataTypeUtils.getDataTypeFromSQLTypeValue(Types.BIT));
         assertEquals(RecordFieldType.ARRAY.getArrayDataType(RecordFieldType.BYTE.getDataType()), DataTypeUtils.getDataTypeFromSQLTypeValue(Types.BLOB));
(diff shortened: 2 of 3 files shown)


Lehel44:
@ravinarayansingh Thanks for making this change. However, I'd like to address a concern regarding its scope as this change only adresses a subset of the issue. Postgres maps `BOOLEAN` types to `BIT` due to historical reasons. By mapping `java.sql.Types.BIT` to `RecordFieldType.Boolean` in `DataTypeUtils::getDataTypeFromSQLTypeValue`, we may encounter a scenario where a field defined as `BIT` in the PostgreSQL table with a size specification such as `BIT(n)` where `n>2` cannot be represented as a `BOOLEAN` type, and is not appropriately converted by the JDBC driver. Consequently, the schema for such a field will also be defined as Boolean. @mattyb149, do you have any thoughts on this?

Lehel44:
Postgres does not recommend storing boolean values in BIT(1) columns because it introduces overhead, so we neither should support a conversion from BIT(1) to Boolean and everything else to Integer, I also think that all the BIT(n) and BIT VARYING(n) can be just converted to Integer. This means that when a BOOLEAN data type is converted to BIT by the JDBC driver and it arrives as a BIT datatype, we cannot distinguish if it was originally a BIT datatype or it was converted by the JDBC driver, we have to map both of them to Intege.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
```diff
@@ -1918,6 +1918,7 @@ public static DataType getDataTypeFromSQLTypeValue(final int sqlType) {
             case Types.BIGINT:
                 return RecordFieldType.BIGINT.getDataType();
             case Types.BOOLEAN:
+            case Types.BIT:
```
Lehel44, line 1921: The mapping for the `BIT` datatype should be changed to `Integer`. This is because the `BIT` datatype can represent not only BIT(1) but also BIT(n>1), making it impossible to distinguish between them. Therefore, using `Integer` provides a more suitable and consistent mapping.

Lehel44 (changes requested):

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
mattyb149: Agreed

ravinarayansingh: hi @Lehel44 and @mattyb149

Have a look on below  sample table definition (Postgres14)

    CREATE TABLE test_table (
        id int4 NOT NULL,
        active bool NULL,
        col_varbit varbit(2) NULL,
        col_bit bit(1) NULL,
        CONSTRAINT temp_market_cap_pkey PRIMARY KEY (id)
    );



| colName     | postgresType | javaSqlType | finalType |
|-------------|--------------|-------------|-----------|
| id          | int          | 4           | INT       |
| active      | bool         | -7          | BOOLEAN   |
| col_bit     | bit(1)       | -7          | BOOLEAN   |
| col_varbit  | varbit(2)    | 1111        | STRING    |

So I believe that we are doing right ,let me know yours thought

References: https://www.postgresql.org/message-id/Pine.BSO.4.61.0511231340310.14387%40leary.csoft.net

ravinarayansingh: hi @Lehel44 and @mattyb149

Have a look on below  sample table definition (Postgres14)

    CREATE TABLE test_table (
        id int4 NOT NULL,
        active bool NULL,
        col_bit bit(1) NULL,
        col_bit1 bit(1) NULL,
        col_varbit varbit(2) NULL,
        CONSTRAINT temp_market_cap_pkey PRIMARY KEY (id)
    );

We can use something like below:

    String typeName = columnResultSet.getString("TYPE_NAME");
    final int dataType;
    if (typeName.equalsIgnoreCase("bool")) {
        dataType = 16;
    } else {
        dataType = columnResultSet.getInt("DATA_TYPE");
    }
            
then below will be result 

| colName    | postgresType | javaSqlType | javaSqlTypeName | finalType |
|------------|--------------|-------------|-----------------|-----------|
| id         | int          | 4           | int4            | INT       |
| active     | bool         | -7          | bool            | BOOLEAN   |
| col_bit    | bit(1)       | -7          | bit             | INT       |
| col_bit1   | bit(2)       | -7          | bit             | INT       |
| col_varbit | varbit(2)    | 1111        | varbit          | STRING    |

let me know your thoughts

Lehel44: @ravinarayansingh I tried it and it looks good to me!

mattyb149:
+1 LGTM, merging to support/nifi-1.x and main. In the future please include the Jira case in the actual commit message(s), I will add them on merge. Thanks for the fix!

PR #8454, opened 2024-02-27
[NIFI-12537] Open cluster/node dialog from Summary screen.

rfellows:
[NIFI-12537](https://issues.apache.org/jira/browse/NIFI-12537)

* Refactored the Summary Listing tables to all extend an abstract base class to reduce code duplication
* Created an abstract base class for the cluster dialog tables
* Added cluster node filtering dropdown when in clustered mode on the summary screen

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/summary/feature/summary.component.html
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/summary/feature/summary.component.html
@@ -19,8 +19,18 @@
     <header class="nifi-header">
         <navigation></navigation>
     </header>
-    <div class="px-5 flex-1 flex flex-col">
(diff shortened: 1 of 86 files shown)


mcgilman:
Will review...

mcgilman (commented):
Nice work @rfellows! The shared filter component and base class for the cluster status table nicely avoid repeated boilerplate as much as possible. I did notice a couple of things detailed below and a separate observation that affects all tables. As I was verifying the sort (and multi-sort) capabilities and clicking on the headers the text kept selecting. I'd be interested to get your thoughts on possibly disabling that by setting `user-select: none` for all `mat-sort-header-content` in the base styles.scss.

rfellows:
> As I was verifying the sort (and multi-sort) capabilities and clicking on the headers the text kept selecting. I'd be interested to get your thoughts on possibly disabling that by setting `user-select: none` for all `mat-sort-header-content` in the base styles.scss.

I agree. We have many issues in the app allowing too much text selection. I can add this to the base listing table style since it happens so regularly when just clicking on the headers to sort them.

mcgilman (approved):
Thanks for the updates @rfellows! Will merge once CI completes.

PR #8455, opened 2024-02-27
[NIFI-12848] - Status History - clustered node support

rfellows:
* fixed ExpressionChanged error in Status History dialog
* color the legend text to match the color of the corresponding line in the chart for each node

[NIFI-12848](https://issues.apache.org/jira/browse/NIFI-12848)


mcgilman:
Reviewing...

PR #8458, opened 2024-02-28
NIFI-12785 Refactored the code to avoid double encoding.

dan-s1:
# Summary

[NIFI-12785](https://issues.apache.org/jira/browse/NIFI-12785)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/InvokeHTTP.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/InvokeHTTP.java
@@ -203,7 +203,8 @@ public class InvokeHTTP extends AbstractProcessor {
 
     public static final PropertyDescriptor HTTP_URL = new PropertyDescriptor.Builder()
             .name("HTTP URL")
-            .description("HTTP remote URL including a scheme of http or https, as well as a hostname or IP address with optional port and path elements.")
+            .description("HTTP remote URL including a scheme of http or https, as well as a hostname or IP address with optional port and path elements." +
+                    " Any encoding of the URL must be done by the user.")
             .required(true)
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .addValidator(StandardValidators.URL_VALIDATOR)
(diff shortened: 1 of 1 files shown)


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/InvokeHTTP.java
```diff
@@ -1012,7 +1013,7 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
         }
     }
 
-    private Request configureRequest(final ProcessContext context, final ProcessSession session, final FlowFile requestFlowFile, URL url) {
+    private Request configureRequest(final ProcessContext context, final ProcessSession session, final FlowFile requestFlowFile, String url) {
```
exceptionfactory, line 1016: For clarification, passing a `URL` results in double-encoding, whereas passing the raw `String` property does not? How was this working before the implementation of the custom UriUtils?

exceptionfactory (commented):
Thanks for working on this @dan-s1. I have one question on implementation to make sure I am following the changes.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/InvokeHTTP.java
dan-s1: @exceptionfactory Not exactly. Only when creating a URL with the UriUtils (v1) / UrlValidator (v2) is there the double encoding. Before my changes to how the URL was created there was never a double encoding but if the URL is created with the URI seven argument constructor it will always encode "illegal" characters.

exceptionfactory: Thanks for the additional explanation. So this should restore the previous behavior prior to the multiple iterations that resulted from the need to move away from `java.net.URL` construction.

With this background, it would be helpful to add or update a unit test that evaluates the Transit URI on a Provenance Event, as one way to confirm the expected behavior.

dan-s1: Would it be okay to check the value of the `invokehttp.request.url ` attribute which the author of the ticket focused on?

exceptionfactory: Yes, that sounds like a good approach.

dan-s1: I have actually incorporated both

dan-s1:
@exceptionfactory Can you please look at this PR when you get a chance? Our team is also having the issue detailed in this ticket and we would like to see this fix  for this. Thank you!

exceptionfactory (approved):
Thanks for adding the test @dan-s1, that definitely clarifies and confirms the expected behavior! +1 merging

PR #8462, opened 2024-02-29
NIFI-12300: Add oauth2 support for RestLookupService

gforeman02:
# Summary

[NIFI-12300](https://issues.apache.org/jira/browse/NIFI-12300): Enable the RestLookupService to use StandardOauth2AccessTokenProvider to support REST services that require oauth2 tokens to access.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

- [x] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for the work on this improvement @gforeman02. Can you rebase this pull request to resolve the current conflicts?

The inclusion of OAuth 2 support is a useful improvement. Rather than include the `customValidate()` method, I recommend adding one more property named `Authentication Strategy`. The default value could be `Username and Password`, and the other value would be `OAuth 2`. The current Basic Authentication properties, plus the `Use Digest Authentication` property would depend on the default value, and then the new OAuth 2 Token Provider could depend on the `OAuth 2` option. That should maintain existing compatibility, and allow for a clearer distinction between which properties apply to which authentication approach.

gforeman02:
@exceptionfactory I appreciate the review.  The modifications are ready for review.  I believe migrateProperties() method is required to account for controller properties that currently have Basic Authentication selected, but let me know if you see differently.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -125,11 +127,30 @@ public class RestLookupService extends AbstractControllerService implements Reco
...
+            .allowableValues(AuthenticationStrategy.class)
+            .defaultValue(AuthenticationStrategy.NONE)
+            .build();
+
+    public static final PropertyDescriptor OAUTH2_ACCESS_TOKEN_PROVIDER = new PropertyDescriptor.Builder()
+        .name("rest-lookup-oauth2-access-token-provider")
+        .displayName("OAuth2 Access Token Provider")
+        .description("Enables managed retrieval of OAuth2 Bearer Token applied to HTTP requests using the Authorization Header.")
+        .identifiesControllerService(OAuth2AccessTokenProvider.class)
+        .required(false)
```
exceptionfactory, line 144: This can be changed to `true` since the `dependOn()` check makes it conditional, and it is required when the associated strategy is selected.
```suggestion
        .required(true)
```

exceptionfactory (changes requested):
Thanks for the updates @gforeman02. The latest version looks close to completion, I noted one more minor adjustment and then this should be ready to go.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
gforeman02: @exceptionfactory I assume this will also apply to the basic auth username and password fields as well now that they also depend on the selection.

exceptionfactory: That could apply to those properties, but since they are existing, it seems reasonable to leave them unchanged for the scope of this particular pull request so as not to introduce any unexpected changes.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/pom.xml
```diff
@@ -148,6 +148,11 @@
             <artifactId>nifi-schema-registry-service-api</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-oauth2-provider-api</artifactId>
+            <scope>compile</scope>
```
exceptionfactory, line 154: The scope should be removed so that it receives the default `provided` from the parent configuration.
```suggestion
```

exceptionfactory (changes requested):
Thanks for the update @gforeman02. On final review, I noticed one more issue with the scope of the `nifi-oauth2-provider-api`, which should be removed so that it will inherit the `provided` scope from the parent Maven configuration.

Along with that update, it would also be helpful to rebase and squash existing commits to remove any merge commits. That's not absolutely necessary for a final merge, but in general, it is better to rebase and force-push as opposed to including merge commits.

exceptionfactory (approved):
Thanks again for the work on this improvement @gforeman02! +1 merging

gforeman02:
@exceptionfactory you are very welcome!  I was working on trying to squash the commits but time slipped away from me.  I will work on that in the future.

PR #8468, opened 2024-03-04
NIFI-12860 Null exception from ExtensionMetadata caused by NIFI-12113

pradden:
# Summary

[NIFI-12860](https://issues.apache.org/jira/browse/NIFI-12860)


exceptionfactory (approved):
Thanks for the correction @pradden! +1 merging

PR #8475, opened 2024-03-05
NIFI-12862 FlowAnalysisResults should not leak unauthorized component details

tpalfy:
When building FlowAnalysisRuleViolationDTO objects (in StandardNiFiServiceFacade), violating component details will be left blank when user has no read permission for that component.

# Summary

[NIFI-12862](https://issues.apache.org/jira/browse/NIFI-12862)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-client-dto/src/main/java/org/apache/nifi/web/api/dto/FlowAnalysisRuleViolationDTO.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-client-dto/src/main/java/org/apache/nifi/web/api/dto/FlowAnalysisRuleViolationDTO.java
@@ -35,6 +35,8 @@ public class FlowAnalysisRuleViolationDTO {
     private String issueId;
     private String violationMessage;
 
+    private String subjectComponentType;
+
     private PermissionsDTO subjectPermissionDto;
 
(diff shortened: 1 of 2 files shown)


sardell:
@tpalfy Thanks for the quick turnaround. How would the client know if a user has read access permission to a component? By checking all those field returns on each violation and implicitly determining based on all those fields values being `null`? I think a better approach would be to include `permissions.canRead` or some other explicit property in the response data model like we do elsewhere.

cc @mcgilman

tpalfy:
> @tpalfy Thanks for the quick turnaround. How would the client know if a user has read access permission to a component? By checking all those field returns on each violation and implicitly determining based on all those fields' values being `null`? I think a better approach would be to include `permissions.canRead` or some other explicit property in the response data model like we do elsewhere.
> 
> cc @mcgilman

Check `result.subjectPermission.canRead` on client side and hide/discard the result when false is a good approach.
As far as UX goes, that would work without this PR. This PR is needed to make sure that unauthorized data isn't even received by the client at all.

sardell:
@tpalfy My mistake. I missed the subjectPermission property in the already-implemented response. Disregard.

pvillard31 (approved):

PR #8477, opened 2024-03-06
NIFI-12854: Add S3 user metadata via flowfile attributes

gforeman02:
# Summary

[NIFI-12854](https://issues.apache.org/jira/browse/NIFI-12854) The PutS3Object processor currently supports adding user metadata to objects via dynamic properties.  This PR adds support for using FlowFile attributes to generate S3 user metadata.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

- [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/PutS3Object.java
+++ b/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/PutS3Object.java
@@ -252,6 +252,27 @@ public class PutS3Object extends AbstractS3Processor {
             .defaultValue("false")
             .build();
 
+    public static final PropertyDescriptor OBJECT_METADATA_PREFIX = new PropertyDescriptor.Builder()
+            .name("s3-object-metadata-prefix")
+            .displayName("Object Metadata Prefix")
+            .description("Specifies the prefix which would be scanned against the incoming FlowFile's attributes and the matching attribute's " +
+                    "name and value would be considered as the outgoing S3 object's Metadata name and Metadata value respectively. For Ex: If the " +
+                    "incoming FlowFile carries the attributes metaS3country, metaS3PII, the metadata prefix to be specified would be 'metaS3'")
+            .required(false)
+            .addValidator(StandardValidators.NON_EMPTY_EL_VALIDATOR)
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .build();
+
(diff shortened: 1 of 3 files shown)


markap14:
Hey @gforeman02 thanks for putting this forward. I'm not sure that I agree that it's a good idea to add this, though. In general, we should not support multiple different ways to do the same thing. Doing so just leads to confusion. Is there a reason that the existing method does not work for you?

gforeman02:
Hello @markap14 , thanks for posting.  Yes, the existing method (dynamic properties) requires knowing what the user metadata keys will be in advance.  My client has a scenario where one flow has metadata key A and another that has key B and this is only known by reading the flowfile attributes.  It would be great for the processor to work in the same way it does for tags.

gforeman02:
@markap14 one other approach could be to have a special dynamic property that the processor uses to read key/value pairs from the dynamic property value.  Maybe something like a key "metadata.json" with a value of `{ "a": "a_val", "b": "b_val" .... }`.

gforeman02:
@markap14 i created a custom s3 nar with the following mod:
```
final Map<String, String> userMetadata = new HashMap<>();
for (final Entry<PropertyDescriptor, String> entry : context.getProperties().entrySet()) {
    if (entry.getKey().isDynamic()) {
        final String value = context.getProperty(
                entry.getKey()).evaluateAttributeExpressions(ff).getValue();
        if( "metadata.json".equals(entry.getKey().getName())){
            ObjectReader reader = new ObjectMapper().readerFor(Map.class);
            Map<String, String> map = reader.readValue(value);
            userMetadata.putAll(map);
        } else {
            userMetadata.put(entry.getKey().getName(), value);
        }
    }
}
```
I am happy to modify this PR and resubmit it for consideration.

exceptionfactory (commented):
@gforeman02 and @markap14, one additional way forward could be the support pattern-based dynamic properties. In other words, if the dynamic property name ends with an asterisk character `*`, it could be handled as a wildcard match against FlowFile attributes. That would retain existing behavior and avoid introducing new properties.

I defer to @markap14 for additional comments on whether that approach could be worth considering.

gforeman02:
@exceptionfactory thanks for the idea on an approach.  I guess a decision needs to be made:

Should the PutS3Object processor support dynamic metadata keys?

+1 here for supporting dynamic metadata keys since I have a scenario where this is used.  @markap14 if the concern is unique to me then I have a workaround in place.

exceptionfactory:
Thanks for the reply @gforeman02, if you willing to rework the PR along the lines of using dynamic properties with wildcards as a way to support dynamic metadata keys, that sounds like a good way forward, without introducing new properties.

exceptionfactory:
@gforeman02 Given the current conflicts and the different direction, I'm closing this pull request for now. Feel free to reopen when you are ready with a different approach, or open a new PR if you prefer. Thanks!

PR #8484, opened 2024-03-07
NIFI-12875: Add unsuccessful HTTP status code handling logic

gforeman02:
# Summary

[NIFI-12875](https://issues.apache.org/jira/browse/NIFI-12875)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

- [x] JDK 21

### Licensing

### Documentation


Lehel44:
Reviewing...

Lehel44 (approved):
Thanks @gforeman02 for making this improvement. I performed some manual testing and it looks good to me. (The error is logged in the RestLookUpService and is also passed up to the processor using it, which results in the error to be logged twice. Since this is legacy code I would not change it but probably the extra log could be removed.)

exceptionfactory (changes requested):
Thanks for the proposed changes @gforeman02.

Reviewing the implementation, however, throwing an `IOException` for a non-successful response could introduce problems from a compatibility point of view. For example, it is legitimate for a REST service to return an HTTP 404 not found if a requested resource does not exist, but that may be acceptable for this Lookup Service use case where information may or may be available.

For that reason, the proposed approach does not look like it is ready to go forward.

One option is to consider the introduction of a Response Code Handling Strategy property, which could enable deciding whether a non-200-series response should be treated as an error. Some additional consideration should be given the the wider variety of use cases before selecting a particular behavior.

gforeman02:
@exceptionfactory thanks for responding.

I see your point about the 404 scenario, good catch.  I will add the property and submit the changes.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -172,6 +173,20 @@ public class RestLookupService extends AbstractControllerService implements Reco
         .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
         .build();
 
+    static final AllowableValue PASS_THROUGH = new AllowableValue("Pass-through",
+            "Pass-through","sends successful and unsuccessful HTTP responses to the caller");
+    static final AllowableValue HANDLE = new AllowableValue("Handle Service Errors",
+            "Handle Service Errors", "generates an exception when an unsuccessful HTTP response code is received");
```
exceptionfactory, line 179: Declaring these options as an `enum` that implements `DescribedValue` makes it easier to refer to the parsing value of the property. Recommend naming the enum `ResponseHandlingStrategy`.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -172,6 +173,20 @@ public class RestLookupService extends AbstractControllerService implements Reco
...
+    public static final PropertyDescriptor PROP_RESPONSE_CODE_HANDLING = new PropertyDescriptor.Builder()
+        .name("rest-lookup-response-code-handling")
+        .displayName("Response Code Handling Strategy")
```
exceptionfactory, line 182: On further consideration, what do you think of naming this `Response Handling Strategy` with values of `Returned` and `Evaluated`? Open to additional suggestions, but having a single word could be useful, with the description indicating all responses and response codes are returned, or the response code is evaluated for success or failure based on standard HTTP conventions?

exceptionfactory (changes requested):
Thanks for the updates @gforeman02. The new property approach looks good, but it would be worth giving a bit more consideration to the naming. I suggested some possible one-word values, but open to additional ideas to communicate the behavior as clearly and succinctly as possible.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
gforeman02: Sounds good.  This change has been incorporated in the latest commit.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
gforeman02: The values are fine with me.  I updated the text descriptions...let me know what you think in the latest commit.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -214,6 +226,7 @@ protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
     private volatile String basicUser;
     private volatile String basicPass;
     private volatile boolean isDigest;
+    private volatile PropertyValue responseHandler;
```
exceptionfactory, line 229: This can be changed to use the `ResponseHandlingStrategy` enum:
```suggestion
    private volatile ResponseHandlingStrategy responseHandlingStrategy;
```

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -172,6 +172,17 @@ public class RestLookupService extends AbstractControllerService implements Reco
         .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
         .build();
 
+
+    public static final PropertyDescriptor PROP_RESPONSE_CODE_HANDLING = new PropertyDescriptor.Builder()
```
exceptionfactory, line 176: Recommend removing the `PROP_` prefix:
```suggestion
    public static final PropertyDescriptor RESPONSE_HANDLING_STRATEGY = new PropertyDescriptor.Builder()
```

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -251,6 +264,8 @@ public void onEnabled(final ConfigurationContext context) {
         buildHeaders(context);
 
         urlTemplate = context.getProperty(URL);
+
+        responseHandler = context.getProperty(PROP_RESPONSE_CODE_HANDLING);
```
exceptionfactory, line 268: After changing the member variable type, the `asAllowableValue()` method can be used to resolve the `enum` value from the property value.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -320,13 +335,20 @@ public Optional<Record> lookup(Map<String, Object> coordinates, Map<String, Stri
...
+            final ResponseBody responseBody = response.body();
 
             if (getLogger().isDebugEnabled()) {
                 getLogger().debug("Response code {} was returned for coordinate {}",
                         new Object[]{response.code(), coordinates});
             }
 
-            final ResponseBody responseBody = response.body();
+            if (!response.isSuccessful()
+                    && ResponseHandlingStrategy.EVALUATED.equals(responseHandler.asAllowableValue(ResponseHandlingStrategy.class))) {
```
exceptionfactory, line 346: After changing the member variable, this comparison can be simplified.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -320,13 +335,20 @@ public Optional<Record> lookup(Map<String, Object> coordinates, Map<String, Stri
...
+                final String responseText = responseBody == null ? "<No Message Received from Server>" : responseBody.string();
```
exceptionfactory, line 347: Recommend replacing  angle brackets with square brackets in the placeholder message:
```suggestion
                final String responseText = responseBody == null ? "[No Message Received]" : responseBody.string();
```

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -320,13 +335,20 @@ public Optional<Record> lookup(Map<String, Object> coordinates, Map<String, Stri
...
+                throw new IOException("Failed to download content from URL " + request.url() +
+                        ": Response code was " + response.code() + ": " + responseText);
```
exceptionfactory, line 349: Recommend using String formatting instead of concatenation for readability:
```suggestion
                throw new IOException("Request failed with HTTP %d for [%s]: %s".formatted(response.code(), request.url(), responseText));
```

exceptionfactory (changes requested):
Thanks for the update @gforeman02. The changes look good, I noted a few more minor stylistic recommendations, and then this should be ready to go.

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/test/java/org/apache/nifi/lookup/TestRestLookupService.java
```diff
@@ -170,6 +171,30 @@ void testLookupRecordFoundPostMethod() throws Exception {
...
+        when(recordReaderFactory.createRecordReader(any(), any(), anyLong(), any())).thenReturn(recordReader);
+        when(recordReader.nextRecord()).thenReturn(record);
+
+        mockWebServer.enqueue(new MockResponse().setResponseCode(HTTP_NOT_FOUND)
+                .setHeader("Content-type", "application/json")
+                .setBody("{\"error\": { \"code\": 404 } }"));
+
+        final Optional<Record> recordFound = restLookupService.lookup(Collections.emptyMap());
+        assertTrue(recordFound.isPresent());
+    }
```
dan-s1, line 188:
```suggestion
    }
    
```

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/test/java/org/apache/nifi/lookup/TestRestLookupService.java
```diff
@@ -170,6 +171,30 @@ void testLookupRecordFoundPostMethod() throws Exception {
...
+    @Test
+    void testLookupErrorHandle() {
+        runner.setProperty(restLookupService, RestLookupService.PROP_RESPONSE_CODE_HANDLING, ResponseHandlingStrategy.EVALUATED);
+        runner.enableControllerService(restLookupService);
+        mockWebServer.enqueue(new MockResponse().setResponseCode(HTTP_NOT_FOUND));
+
+        final LookupFailureException exception = assertThrows(LookupFailureException.class, () -> restLookupService.lookup(Collections.emptyMap()));
+        assertInstanceOf(IOException.class, exception.getCause());
+    }
```
dan-s1, line 197:
```suggestion
    }
    
```

nifi-nar-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/RestLookupService.java
```diff
@@ -172,6 +172,17 @@ public class RestLookupService extends AbstractControllerService implements Reco
...
 
+
+    public static final PropertyDescriptor RESPONSE_HANDLING_STRATEGY = new PropertyDescriptor.Builder()
+        .name("rest-lookup-response-code-handling")
+        .displayName("Response Code Handling Strategy")
+        .description("How to handle response codes from remote service.  'Pass-through' sends successful and unsuccessful HTTP responses to the caller.  " +
+                "'Handle Service Errors' generates an exception when an unsuccessful HTTP response code is received.")
+        .required(true)
+        .defaultValue(ResponseHandlingStrategy.RETURNED)
+        .allowableValues(ResponseHandlingStrategy.RETURNED, ResponseHandlingStrategy.EVALUATED)
```
dan-s1, line 183:
```suggestion
        .allowableValues(ResponseHandlingStrategy.class)
```

dan-s1 (commented):
Some additional minor changes requested.

gforeman02:
@dan-s1 and @exceptionfactory thanks for the reviews/suggestions.  Updates committed.

exceptionfactory (approved):
Thanks for working through the feedback @gforeman02! I pushed one more commit adjusting the property description to remove the wording related to the original values, and I will proceed to merge this version.

PR #8491, opened 2024-03-11
NIFI-10707 Added proxy support in PutBigQuery

turcsanyip:
Bumped GCP client library version
Added grpc-* jars in service api nar in order to avoid CNFE warning in io.grpc.LoadBalancerRegistry

# Summary

[NIFI-10707](https://issues.apache.org/jira/browse/NIFI-10707)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


exceptionfactory (commented):
Thanks for working on this improvement @turcsanyip. The implementation looks straightforward.

Regarding the gRPC libraries, do you know why they are required in the API module? Are there other gRPC libraries loaded through the API module, thus requiring these to be available in the Class Loader? If these are required, it should be possible to exclude them in the service implementation NAR to avoid packaging duplicate copies.

turcsanyip:
> Regarding the gRPC libraries, do you know why they are required in the API module? Are there other gRPC libraries loaded through the API module, thus requiring these to be available in the Class Loader? If these are required, it should be possible to exclude them in the service implementation NAR to avoid packaging duplicate copies.

`com.google.auth:google-auth-library-oauth2-http` brings in io.grpc:grpc-context and `io.grpc:grpc-api`. The latter contains `io.grpc.LoadBalancerRegistry` which tries to load classes from `io.grpc:grpc-core` and `io.grpc:grpc-util` via reflection:
https://github.com/grpc/grpc-java/blob/b3475a0e4625d518434f7d89eae7965a44a427cf/api/src/main/java/io/grpc/LoadBalancerRegistry.java#L140-L152

It seems to work without these classes loaded but the warning log can be confusing so I would fix it.

I agree that jar duplicates should be avoided. Actually, almoast all jars from the `nifi-gcp-services-api-nar` were duplicated in the `nifi-gcp-nar`. I excluded all of them.

pvillard31 (approved):

PR #8493, opened 2024-03-12
NIFI-12887 Additional ways for handling Strings as binary in PutDatabaseRecord

tpalfy:
Add option to treat Strings as hexadecimal character sequences or base64-encoded binary data when inserting into a binary type column.

# Summary

[NIFI-12887](https://issues.apache.org/jira/browse/NIFI-12887)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
@@ -18,6 +18,7 @@ package org.apache.nifi.processors.standard;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.io.BaseEncoding;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
@@ -238,6 +239,34 @@ public class PutDatabaseRecord extends AbstractProcessor {
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
 
+    static final AllowableValue BINARY_STRING_FORMAT_UTF8 = new AllowableValue(
+            "UTF8",
+            "UTF8",
+            "String values for binary columns contain the original value as text via UTF-8 character encoding"
+    );
+
+    static final AllowableValue BINARY_STRING_FORMAT_HEX_STRING = new AllowableValue(
+            "Hex String",
+            "Hex String",
+            "String values for binary columns contain the original value in hexadecimal format"
+    );
+
+    static final AllowableValue BINARY_STRING_FORMAT_BASE64 = new AllowableValue(
+            "Base64-encoded String",
+            "Base64-encoded String",
+            "String values for binary columns contain the original value in Base64 encoded format"
+    );
+
+    static final PropertyDescriptor BINARY_STRING_FORMAT = new Builder()
+            .name("put-db-record-binary-format")
+            .displayName("Binary String Format")
+            .description("How to interpret string data for binary columns.")
+            .required(false)
+            .expressionLanguageSupported(FLOWFILE_ATTRIBUTES)
+            .allowableValues(BINARY_STRING_FORMAT_UTF8, BINARY_STRING_FORMAT_HEX_STRING, BINARY_STRING_FORMAT_BASE64)
+            .defaultValue(BINARY_STRING_FORMAT_UTF8.getValue())
+            .build();
+
     static final PropertyDescriptor TRANSLATE_FIELD_NAMES = new Builder()
             .name("put-db-record-translate-field-names")
             .displayName("Translate Field Names")
@@ -388,6 +417,7 @@ public class PutDatabaseRecord extends AbstractProcessor {
         pds.add(CATALOG_NAME);
         pds.add(SCHEMA_NAME);
         pds.add(TABLE_NAME);
+        pds.add(BINARY_STRING_FORMAT);
         pds.add(TRANSLATE_FIELD_NAMES);
         pds.add(UNMATCHED_FIELD_BEHAVIOR);
         pds.add(UNMATCHED_COLUMN_BEHAVIOR);
@@ -606,6 +636,8 @@ public class PutDatabaseRecord extends AbstractProcessor {
         final int maxBatchSize = context.getProperty(MAX_BATCH_SIZE).evaluateAttributeExpressions(flowFile).asInteger();
         final int timeoutMillis = context.getProperty(QUERY_TIMEOUT).evaluateAttributeExpressions().asTimePeriod(TimeUnit.MILLISECONDS).intValue();
 
+        final String binaryStringFormat = context.getProperty(BINARY_STRING_FORMAT).evaluateAttributeExpressions(flowFile).getValue();
+
         // Ensure the table name has been set, the generated SQL statements (and TableSchema cache) will need it
         if (StringUtils.isEmpty(tableName)) {
             throw new IllegalArgumentException(format("Cannot process %s because Table Name is null or empty", flowFile));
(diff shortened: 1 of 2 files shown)


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
```diff
@@ -764,7 +796,15 @@ private void executeDML(final ProcessContext context, final ProcessSession sessi
                                             }
                                             currentValue = dest;
                                         } else if (currentValue instanceof String) {
-                                            currentValue = ((String) currentValue).getBytes(StandardCharsets.UTF_8);
+                                            final String stringValue = (String) currentValue;
+
+                                            if (BINARY_STRING_FORMAT_BASE64.getValue().equals(binaryStringFormat)) {
+                                                currentValue = BaseEncoding.base64().decode(stringValue);
```
exceptionfactory, line 802: The standard Java [Base64](https://docs.oracle.com/en/java/javase/21/docs//api/java.base/java/util/Base64.html) class should be used instead of the the Google class.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
```diff
@@ -764,7 +796,15 @@ private void executeDML(final ProcessContext context, final ProcessSession sessi
...
+                                            } else if (BINARY_STRING_FORMAT_HEX_STRING.getValue().equals(binaryStringFormat)) {
+                                                currentValue = BaseEncoding.base16().decode(stringValue.toUpperCase());
```
exceptionfactory, line 804: The standard Java [HexFormat](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/HexFormat.html) class should be used instead of the Google class.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
```diff
@@ -238,6 +239,34 @@ public class PutDatabaseRecord extends AbstractProcessor {
...
+    static final AllowableValue BINARY_STRING_FORMAT_BASE64 = new AllowableValue(
+            "Base64-encoded String",
+            "Base64-encoded String",
+            "String values for binary columns contain the original value in Base64 encoded format"
+    );
+
+    static final PropertyDescriptor BINARY_STRING_FORMAT = new Builder()
+            .name("put-db-record-binary-format")
+            .displayName("Binary String Format")
+            .description("How to interpret string data for binary columns.")
```
exceptionfactory, line 263: Recommend adjusting the description for clarity to indicate decoding.
```suggestion
            .description("The format to be applied when decoding string values to binary.")
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
```diff
@@ -238,6 +239,34 @@ public class PutDatabaseRecord extends AbstractProcessor {
...
 
+    static final AllowableValue BINARY_STRING_FORMAT_UTF8 = new AllowableValue(
+            "UTF8",
+            "UTF8",
+            "String values for binary columns contain the original value as text via UTF-8 character encoding"
+    );
+
+    static final AllowableValue BINARY_STRING_FORMAT_HEX_STRING = new AllowableValue(
+            "Hex String",
+            "Hex String",
```
exceptionfactory, line 250: Recommend using `Hexadecimal` to align with other options.
```suggestion
            "Hexadecimal",
            "Hexadecimal",
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
```diff
@@ -238,6 +239,34 @@ public class PutDatabaseRecord extends AbstractProcessor {
...
+            "String values for binary columns contain the original value in hexadecimal format"
+    );
+
+    static final AllowableValue BINARY_STRING_FORMAT_BASE64 = new AllowableValue(
+            "Base64-encoded String",
+            "Base64-encoded String",
```
exceptionfactory, line 256: Recommend using `Base64` to align with other options.
```suggestion
            "Base64",
            "Base64",
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutDatabaseRecord.java
exceptionfactory: Java 8 support is not a requirement for the main branch, and HexFormat was introduced in Java 17, so this should be the direction for future integration. If the goal is to support this capability for the 1.x branch, then the feature could be backported and the Google library could be used. Going forward, however, we should make use of available Java capabilities as opposed to placing more dependencies on Google Guava or other libraries.

tpalfy: Yes, I'm thinking about having this change on the 1.x branch.

So not sure what is your suggestion. If we want it on 1.x branch is it okay to use the Google library?

Or you suggest it's worth the additional work to use vanilla Java on main and instead of simply doing a cherry-pick, the committer should do the code rewrite during backport to 1.x?

pvillard31:
@tpalfy - I think the easiest is to keep this PR for support/1.x, and open another PR targeted at main that does not use Google library. I think it's safer compared to letting the merger do the changes. Happy to merge once we're on the same page.

tpalfy:
> @tpalfy - I think the easiest is to keep this PR for support/1.x, and open another PR targeted at main that does not use Google library. I think it's safer compared to letting the merger do the changes. Happy to merge once we're on the same page.

This PR is based on and is open against `main` already. So I'll need to change this one and open a new one against 1.x.

exceptionfactory (commented):
Thanks for the update @tpalfy, the change to the standard Java classes looks good, the only thing remaining appears to be removing the unused import for `com.google.common.io.BaseEncoding`.

pvillard31:
Yeah the option of opening another PR for 1.x is also OK :) As a FYI, from the Github UI, on the PR, you can change the branch against which the PR is submitted, that is why I suggested to use this one against 1.x and open a new one with the changes. But all good, the end result is the same ;)

exceptionfactory (changes requested):
Thanks for making the adjustments @tpalfy.

The last item to address it the `required()` statue for the new property. As mentioned in the comment, it can be set to `true` since the default value provided maintains compatibility with existing configuration.

Other than that, the changes look good.

tpalfy:
> The last item to address it the `required()` status for the new property. As mentioned in the comment, it can be set to `true` since the default value provided maintains compatibility with existing configuration.

Technically we could make it required, but we don't automatically do that in such cases. When a property is essential on a conceptual level, we make it required but leave it optional otherwise, even if it has a default value.

This is more about UX but it makes sense. This can't be considered a conceptually essential parameter, like 'Table Name' for example. Let's leave it optional.

exceptionfactory:
> > The last item to address it the `required()` status for the new property. As mentioned in the comment, it can be set to `true` since the default value provided maintains compatibility with existing configuration.
> 
> Technically we could make it required, but we don't automatically do that in such cases. When a property is essential on a conceptual level, we make it required but leave it optional otherwise, even if it has a default value.
> 
> This is more about UX but it makes sense. This can't be considered a conceptually essential parameter, like 'Table Name' for example. Let's leave it optional.

Thanks for the reply.

Although use of the property is optional in one respect, in that it does not apply if the record does not relate to a binary column, it is required in terms of understanding the default behavior.

Right now, the default behavior is implicit, always using UTF-8 decoding. From another angle, making the property optional leaves the behavior ambiguous, because the property value could be an empty string. In that case, it could fall back to UTF-8, but that is less clear than marking the property as required.

exceptionfactory (approved):
Thanks for making the adjustments @tpalfy, the latest version looks good. +1 merging

PR #8498, opened 2024-03-13
NIFI-12855: Add more information to provenance events to facilitate full graph traversal

mattyb149:
# Summary

[NIFI-12855](https://issues.apache.org/jira/browse/NIFI-12855) This PR augments the provenance capabilities to include the following features:

- A reference in a provenance event to any parent events ("previousEventIds")
- Add methods to GraphClientService to generate queries/statements in popular graph languages such as Tinkerpop/Gremlin, Cypher, and SQL
- Add ArcadeDB service as reference implementation for SQL generation

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/minifi/minifi-nar-bundles/minifi-provenance-repository-bundle/minifi-provenance-repositories/src/main/java/org/apache/nifi/provenance/NoOpProvenanceRepository.java
+++ b/minifi/minifi-nar-bundles/minifi-provenance-repository-bundle/minifi-provenance-repositories/src/main/java/org/apache/nifi/provenance/NoOpProvenanceRepository.java
@@ -29,7 +29,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 
-import static java.util.Collections.EMPTY_SET;
 import static java.util.Collections.emptyList;
 
 /**
@@ -37,12 +36,12 @@ import static java.util.Collections.emptyList;
  * store events.
  *
  */
-public class NoOpProvenanceRepository implements ProvenanceRepository {
+public class NoOpProvenanceRepository extends AbstractProvenanceRepository {
 
   @Override
   public void initialize(EventReporter eventReporter, Authorizer authorizer,
-      ProvenanceAuthorizableFactory factory, IdentifierLookup identifierLookup)
-      throws IOException {
+                         ProvenanceAuthorizableFactory factory, IdentifierLookup identifierLookup)
+          throws IOException {
 
   }
 
@@ -68,13 +67,13 @@ public class NoOpProvenanceRepository implements ProvenanceRepository {
 
   @Override
   public List<ProvenanceEventRecord> getEvents(long firstRecordId, int maxRecords)
-      throws IOException {
+          throws IOException {
     return emptyList();
   }
 
   @Override
   public List<ProvenanceEventRecord> getEvents(long firstRecordId,
-      int maxRecords, NiFiUser niFiUser) throws IOException {
+                                               int maxRecords, NiFiUser niFiUser) throws IOException {
     return emptyList();
   }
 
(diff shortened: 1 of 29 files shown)


exceptionfactory (changes requested):
Thanks for adjusting the approach in this pull request @mattyb149.

Two quick notes that will make this more straightforward to review:

1. There are large number of white-space formatting changes that make this more difficult to evaluate. It would be very helpful to revert formatting changes so the scope of substantive changes is clearer
2. The Graph Client component changes should be extracted to a separate pull request. Although it is helpful to see the relationship, it would be much better to isolate framework changes from component changes that introduce features not directly related.

mattyb149:
The graph client stuff has been removed, I'll do a separate PR but not yet in case the reviews here affect the clients. Once this is in it will facilitate more clients such as perhaps an RDF/SPARQL client. Thanks for the inputs!

exceptionfactory (changes requested):
Thanks for scoping down the changes in this pull request @mattyb149, this makes the core changes much easier to follow.

Reviewing the addition of the `previousEventIds` raises several questions. The `ProvenanceEventRecord` already contains Parent and Child FlowFile identifiers, used for linking, and it seems like this should be sufficient to support graph traversal use cases. The Provenancen Event ID is more of an internal database identifier, so it seems less than optimal to promote additional usage of this field.

From a performance perspective, querying the Provenance Repository for previous identifiers also seems like a potential performance problem.

If you can provide more background on why this is necessary, that would be helpful, but otherwise, the proposed approach does not look like a good candidate to go forward.

exceptionfactory (changes requested):
@mattyb149 Given the current merge conflicts and the unresolved discussion on the need for these changes, I recommend closing this pull request and revisiting it when ready. We could continue the discussion on the Jira issue if that would be helpful to scope out the types of changes needed.

minifi/minifi-nar-bundles/minifi-provenance-repository-bundle/minifi-provenance-repositories/src/main/java/org/apache/nifi/provenance/NoOpProvenanceRepository.java
```diff
@@ -29,20 +29,19 @@
...
  *
  */
-public class NoOpProvenanceRepository implements ProvenanceRepository {
+public class NoOpProvenanceRepository extends AbstractProvenanceRepository {
 
   @Override
   public void initialize(EventReporter eventReporter, Authorizer authorizer,
-      ProvenanceAuthorizableFactory factory, IdentifierLookup identifierLookup)
-      throws IOException {
+                         ProvenanceAuthorizableFactory factory, IdentifierLookup identifierLookup)
```
joewitt, line 43: there are a number of entries here that look like whitespace additions.  Please double check and if so remove

nifi-api/src/main/java/org/apache/nifi/provenance/ProvenanceEventBuilder.java
```diff
@@ -283,6 +283,13 @@ public interface ProvenanceEventBuilder {
      */
     ProvenanceEventBuilder setRelationship(Relationship relationship);
 
+    /**
+     * Sets the IDs for the event that happened previously to this event for the given FlowFile
+     * @param previousEventIds The previous event IDs (usually one except for JOIN events and such)
+     * @return the builder
+     */
+    ProvenanceEventBuilder setPreviousEventIds(List<Long> previousEventIds);
```
joewitt, line 291: Does this imply we're storing event ids for the before and after event on both the before and after event?

joewitt: Ah maybe it is that we had parent FlowFile identifiers and now we'll have previous Event Identfiers.

joewitt:
@mattyb149 It is important to undo all the whitespace changes to improve the reviewer efficiency and avoid the whitespace in general.

PR #8500, opened 2024-03-13
NIFI-12514: Add config python virtual env binary directory for windows.

bobpaulin:
# Summary

[NIFI-12514](https://issues.apache.org/jira/browse/NIFI-12514)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

Started nifi on windows with:

nifi.python.command=python
nifi.python.virtual.env.binary.directory=Scripts

Started fine.  Ran a simple Python extension.

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-properties/src/main/java/org/apache/nifi/util/NiFiProperties.java
+++ b/nifi-commons/nifi-properties/src/main/java/org/apache/nifi/util/NiFiProperties.java
@@ -322,6 +322,7 @@ public class NiFiProperties extends ApplicationProperties {
     public static final String PYTHON_MAX_PROCESSES = "nifi.python.max.processes";
     public static final String PYTHON_MAX_PROCESSES_PER_TYPE = "nifi.python.max.processes.per.extension.type";
     public static final String PYTHON_COMMS_TIMEOUT = "nifi.python.comms.timeout";
+    public static final String PYTHON_VIRTUAL_ENV_BINARY_DIRECTORY = "nifi.python.virtual.env.binary.directory";
 
     public static final String PYTHON_CONTROLLER_DEBUGPY_ENABLED = "nifi.python.controller.debugpy.enabled";
     public static final String PYTHON_CONTROLLER_DEBUGPY_PORT = "nifi.python.controller.debugpy.port";
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
@@ -878,7 +878,7 @@ public class FlowController implements ReportingTaskProvider, FlowAnalysisRulePr
         final boolean enableControllerDebug = Boolean.parseBoolean(nifiProperties.getProperty(NiFiProperties.PYTHON_CONTROLLER_DEBUGPY_ENABLED, "false"));
         final int debugPort = nifiProperties.getIntegerProperty(NiFiProperties.PYTHON_CONTROLLER_DEBUGPY_PORT, 5678);
         final String debugHost = nifiProperties.getProperty(NiFiProperties.PYTHON_CONTROLLER_DEBUGPY_HOST, "localhost");
-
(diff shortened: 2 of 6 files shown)


exceptionfactory (commented):
Thanks for the contribution @bobpaulin, enabling Python extensions on Windows is a helpful step forward.

Although the new property provides a level of flexibility, it also presents some complexity point of view. I do not run on Windows as frequently, but if `bin` is the common directory on macOS and Linux, and `Scripts` is common on Windows, it seems possible to check for the existence of each sub-directory and select one accordingly. That would avoid the need for a new property and the potential complexity. What do you think about that approach?

bobpaulin:
> Thanks for the contribution @bobpaulin, enabling Python extensions on Windows is a helpful step forward.
> 
> Although the new property provides a level of flexibility, it also presents some complexity. I do not run on Windows as frequently, but if `bin` is the common directory on macOS and Linux, and `Scripts` is common on Windows, it seems possible to check for the existence of each sub-directory and select one accordingly. That would avoid the need for a new property and the potential complexity. What do you think about that approach?

Hi @exceptionfactory  thanks for the quick reply.  Yes I am open to a more automated approach see [1] .  I think checking the parent directory for the sub-directory rather than specifically checking for OS might be much simpler.  However not being familiar with how stable venv is do you think it makes sense to implement this as a filter config for which directories to check?  EX

```
nifi.python.virtual.env.binary.directory.filter=bin,Scripts
```

That way if venv introduces a change users can simply add the directory to the list?

[1] https://lists.apache.org/thread/gfdm5w3dw4zm6rhsswk71gj8pg3d21m3

exceptionfactory:
Thanks for the reply and reference to the mailing list thread @bobpaulin.

Reviewing the [Python Virtual Environments documentation](https://docs.python.org/3/tutorial/venv.html) and a [Python venv tutorial](https://python.land/virtual-environments/virtualenv), `Scripts` seems to be the standard directory for Windows and `bin` seems to be standard for macOS and Linux.

Given that background, I think directory-existence checking would be the ideal way forward. It avoids having to evaluate the OS System properties, and also avoids the need for a new NiFi property.

If we run into issues down the road, then something along the lines of the filter property you described seems worth considering. For now, however, checking first for `bin` and then for `Scripts` seems like it should cover current behavior.

bobpaulin:
> Thanks for the reply and reference to the mailing list thread @bobpaulin.
> 
> Reviewing the [Python Virtual Environments documentation](https://docs.python.org/3/tutorial/venv.html) and a [Python venv tutorial](https://python.land/virtual-environments/virtualenv), `Scripts` seems to be the standard directory for Windows and `bin` seems to be standard for macOS and Linux.
> 
> Given that background, I think directory-existence checking would be the ideal way forward. It avoids having to evaluate the OS System properties, and also avoids the need for a new NiFi property.
> 
> If we run into issues down the road, then something along the lines of the filter property you described seems worth considering. For now, however, checking first for `bin` and then for `Scripts` seems like it should cover current behavior.

Hi @exceptionfactory .  Yes I think we're aligned on using directory over OS detection.  And I agree there is less changes required if we simply hard code the check for first bin then Scripts into the code.  Naturally the tradeoff is less complexity with requiring a Nifi release to adapt if venv changes vs a more complex change upfront .  It sounds like you're saying at this point your read is the property is more trouble than it's worth.  I'll bring this conversation back on list and adjust my PR as advised.  Thanks again for the quick reply.

exceptionfactory:
> > Thanks for the reply and reference to the mailing list thread @bobpaulin.
> > Reviewing the [Python Virtual Environments documentation](https://docs.python.org/3/tutorial/venv.html) and a [Python venv tutorial](https://python.land/virtual-environments/virtualenv), `Scripts` seems to be the standard directory for Windows and `bin` seems to be standard for macOS and Linux.
> > Given that background, I think directory-existence checking would be the ideal way forward. It avoids having to evaluate the OS System properties, and also avoids the need for a new NiFi property.
> > If we run into issues down the road, then something along the lines of the filter property you described seems worth considering. For now, however, checking first for `bin` and then for `Scripts` seems like it should cover current behavior.
> 
> Hi @exceptionfactory . Yes I think we're aligned on using directory over OS detection. And I agree there is less changes required if we simply hard code the check for first bin then Scripts into the code. Naturally the tradeoff is less complexity with requiring a Nifi release to adapt if venv changes vs a more complex change upfront . It sounds like you're saying at this point your read is the property is more trouble than it's worth. I'll bring this conversation back on list and adjust my PR as advised. Thanks again for the quick reply.

Yes, that's a good summary. Thanks for moving this forward!

PR #8503, opened 2024-03-14
NIFI-12900 Avoid unnecessary directory listing in PutSFTP

EndzeitBegins:
# Summary

Adjusts `SFTPTransfer` to avoid listing whole target directory before upload in `PutSFTP`. 

[NIFI-12900](https://issues.apache.org/jira/browse/NIFI-12900)

I'd appreciate if we could backport these changes to 1.x as well. Please let me know if I should open a separate pull-request for that.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
@@ -708,8 +708,6 @@ public class SFTPTransfer implements FileTransfer {
             }
         }
 
-        // Previously JSCH would perform a listing on the full path (path + filename) and would get an exception when it wasn't
-        // a file and then return null, so to preserve that behavior we return null if the matchingEntry is a directory
         if (fileAttributes == null || isDirectory(fileAttributes)) {
             return null;
         } else {
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
@@ -111,10 +111,12 @@ public class SFTPTransfer implements FileTransfer {
     }
 
     private static String buildFullPath(String path, String filename) {
-        return (path == null) ? filename : (path.endsWith("/")) ? path + filename : path + "/" + filename;
+        if (path == null) return filename;
+        if (path.endsWith("/")) return path + filename;
+        return path + "/" + filename;
     }
 
-    private static boolean identifiesDirectory(FileAttributes attributes) {
+    private static boolean isDirectory(FileAttributes attributes) {
         return attributes.getType() == FileMode.Type.DIRECTORY;
     }
 
@@ -348,7 +350,7 @@ public class SFTPTransfer implements FileTransfer {
 
                 if (isIncludedFile(entry, symlink) && (filteringDisabled || pathMatched)) {
                     if (filteringDisabled || fileFilterPattern == null || fileFilterPattern.matcher(entryFilename).matches()) {
-                        listing.add(newFileInfo(entry, path));
+                        listing.add(newFileInfo(path, entry.getName(), entry.getAttributes()));
                     }
                 }
 
@@ -428,14 +430,7 @@ public class SFTPTransfer implements FileTransfer {
         return FileMode.Type.SYMLINK == remoteResourceInfo.getAttributes().getType();
     }
 
-    private FileInfo newFileInfo(final RemoteResourceInfo entry, String path) {
-        if (entry == null) {
-            return null;
-        }
-        return newFileInfo(entry.getAttributes(), path, entry.getName());
-    }
-
-    private FileInfo newFileInfo(final FileAttributes attributes, String path, String filename) {
+    private FileInfo newFileInfo(String path, String filename, final FileAttributes attributes) {
         final File newFullPath = new File(path, filename);
         final String newFullForwardPath = newFullPath.getPath().replace("\\", "/");
 
@@ -457,7 +452,7 @@ public class SFTPTransfer implements FileTransfer {
         final FileInfo.Builder builder = new FileInfo.Builder()
             .filename(filename)
             .fullPathFileName(newFullForwardPath)
-            .directory(identifiesDirectory(attributes))
+            .directory(isDirectory(attributes))
             .size(attributes.getSize())
             .lastModifiedTime(attributes.getMtime() * 1000L)
             .permissions(permsBuilder.toString())
@@ -465,7 +460,6 @@ public class SFTPTransfer implements FileTransfer {
             .group(Integer.toString(attributes.getGID()));
         return builder.build();
     }
-
 
     private void appendPermission(final StringBuilder builder, final Set<FilePermission> permissions, final FilePermission filePermission, final String permString) {
         if (permissions.contains(filePermission)) {
@@ -710,10 +704,10 @@ public class SFTPTransfer implements FileTransfer {
 
         // Previously JSCH would perform a listing on the full path (path + filename) and would get an exception when it wasn't
         // a file and then return null, so to preserve that behavior we return null if the matchingEntry is a directory
-        if (fileAttributes == null || identifiesDirectory(fileAttributes)) {
+        if (fileAttributes == null || isDirectory(fileAttributes)) {
             return null;
         } else {
-            return newFileInfo(fileAttributes, path, filename);
+            return newFileInfo(path, filename, fileAttributes);
         }
     }
 
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
@@ -111,8 +111,14 @@ public class SFTPTransfer implements FileTransfer {
     }
 
     private static String buildFullPath(String path, String filename) {
-        if (path == null) return filename;
-        if (path.endsWith("/")) return path + filename;
+        if (path == null) {
+             return filename;
+        }
+        
+        if (path.endsWith("/")) { 
+            return path + filename;
+        }
+         
         return path + "/" + filename;
     }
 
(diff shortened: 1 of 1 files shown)


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
```diff
@@ -110,6 +110,14 @@ private static String convertFactorySetToString(Set<String> factorySetNames) {
                 .collect(Collectors.joining(", "));
     }
 
+    private static String buildFullPath(String path, String filename) {
+        return (path == null) ? filename : (path.endsWith("/")) ? path + filename : path + "/" + filename;
+    }
+
+    private static boolean identifiesDirectory(FileAttributes attributes) {
```
dan-s1, line 117: The usual convention to name methods that return boolean is to ask a question in the name usually by using is/has as the prefix.
```suggestion
    private static boolean isDirectory(FileAttributes attributes) {
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
```diff
@@ -443,17 +455,18 @@ private FileInfo newFileInfo(final RemoteResourceInfo entry, String path) {
         appendPermission(permsBuilder, permissions, FilePermission.OTH_X, "x");
 
         final FileInfo.Builder builder = new FileInfo.Builder()
-            .filename(entry.getName())
+            .filename(filename)
             .fullPathFileName(newFullForwardPath)
-            .directory(entry.isDirectory())
-            .size(entry.getAttributes().getSize())
-            .lastModifiedTime(entry.getAttributes().getMtime() * 1000L)
+            .directory(identifiesDirectory(attributes))
```
dan-s1, line 460: Per earlier comment
```suggestion
            .directory(isDirectory(attributes))
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
```diff
@@ -694,20 +708,12 @@ public FileInfo getRemoteFileInfo(final FlowFile flowFile, final String path, St
...
-            if (entry.getName().equalsIgnoreCase(filename)) {
-                matchingEntry = entry;
-                break;
-            }
-        }
-
         // Previously JSCH would perform a listing on the full path (path + filename) and would get an exception when it wasn't
         // a file and then return null, so to preserve that behavior we return null if the matchingEntry is a directory
-        if (matchingEntry != null && matchingEntry.isDirectory()) {
+        if (fileAttributes == null || identifiesDirectory(fileAttributes)) {
```
dan-s1, line 713: Per earlier comment
```suggestion
        if (fileAttributes == null || isDirectory(fileAttributes)) {
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
```diff
@@ -110,6 +110,14 @@ private static String convertFactorySetToString(Set<String> factorySetNames) {
                 .collect(Collectors.joining(", "));
     }
 
+    private static String buildFullPath(String path, String filename) {
+        return (path == null) ? filename : (path.endsWith("/")) ? path + filename : path + "/" + filename;
+    }
```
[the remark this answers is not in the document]
EndzeitBegins: Thank you for the review.

Yes, I thought about that myself, but figured I keep the changes to a minimum. 
However, I think you're right that this should be refactored. I replaced it with an if statement.

EndzeitBegins:
Thank you for your review @dan-s1.

I've pushed another commit based on your suggestions. I'd appreciate if you'd take another look.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
```diff
@@ -110,6 +110,16 @@ private static String convertFactorySetToString(Set<String> factorySetNames) {
                 .collect(Collectors.joining(", "));
     }
 
+    private static String buildFullPath(String path, String filename) {
+        if (path == null) return filename;
+        if (path.endsWith("/")) return path + filename;
+        return path + "/" + filename;
```
dan-s1, line 116: While this code is valid I have noticed in the NIFI code the importance for readability, I would suggest expanding to the code below or what I originally suggested using `else if` and `else`
```suggestion
        if (path == null) {
             return filename;
        }
        
        if (path.endsWith("/")) { 
            return path + filename;
        }
         
        return path + "/" + filename;
```

dan-s1 (commented):
@EndzeitBegins Changes look good thank you for making them. I had one more suggestion.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
EndzeitBegins: Yeah, I guess that's a matter of preference.
Personally I feel eaving the brackets out is less verbose and more readable.
But I'm working with `Kotlin` mostly, so I'm used to a more consice syntax I guess. 

I'm fine with both either way, so I applied your suggestion.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
```diff
@@ -340,7 +356,7 @@ protected void getListing(final String path, final int depth, final int maxResul
 
                 if (isIncludedFile(entry, symlink) && (filteringDisabled || pathMatched)) {
                     if (filteringDisabled || fileFilterPattern == null || fileFilterPattern.matcher(entryFilename).matches()) {
-                        listing.add(newFileInfo(entry, path));
+                        listing.add(newFileInfo(path, entry.getName(), entry.getAttributes()));
```
exceptionfactory, line 359: This change does not account for the possibility of `entry` being `null` as implied by the previous behavior.

exceptionfactory: In this particular context, the `RemoteResourceInfo` may not return directory status if the path is a symbolic link.

exceptionfactory (changes requested):
Thanks for your work on this improvement @EndzeitBegins. This check is important to get correct in all cases, and there seems to be a potential issue with the changes in the `getListing()` method. That method returns a list of `RemoteResourceInfo` objects, but some of those objects may be symbolic links, which do not indicate directory status. Calling `stat()` on every entry would be very expensive, so that is not an option. It seems like scoping down the changes to the `getRemoteFileInfo()` method could be better, even if that means backing out some of the refactoring to share the `newFileInfo()` method. Have you evaluated the behavior with symbolic links in various scenarios?

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
EndzeitBegins: Thank you for your feedback @exceptionfactory.

The call to `newFileInfo` is preceded by a call to `isIncludedFile` inside the if expression, which in turn calls `.isRegularFile()` on the `RemoteResourceInfo entry` passed in. 
In case the `entry` was `null`, a `NullPointerException` would be thrown before reaching `newFileInfo`. 

Thus I figured the null check is not required in `newFileInfo`, but I can add an explicit `null` check into the if clause, if that's desired.

exceptionfactory: Very good point @EndzeitBegins, on closer inspection, I agree with you that the null check there should be unnecessary given that it would have failed in `isIncludedFile()` based on current behavior.

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
```diff
@@ -694,20 +708,12 @@ public FileInfo getRemoteFileInfo(final FlowFile flowFile, final String path, St
...
-        RemoteResourceInfo matchingEntry = null;
-        for (final RemoteResourceInfo entry : remoteResources) {
-            if (entry.getName().equalsIgnoreCase(filename)) {
-                matchingEntry = entry;
-                break;
-            }
-        }
-
         // Previously JSCH would perform a listing on the full path (path + filename) and would get an exception when it wasn't
         // a file and then return null, so to preserve that behavior we return null if the matchingEntry is a directory
```
exceptionfactory, line 712: Given the change in behavior, these comments should be removed.
```suggestion
```

nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java
exceptionfactory: Regarding the second question, it appears that the behavior remains the same. Thanks for the helpful reply, this looks good.

exceptionfactory (commented):
Thanks for the replies @EndzeitBegins.

On closer evaluation of the existing `getListing()` behavior, it does appear to be unchanged.

With that, I think the only adjustment is removing the legacy comment regarding historical JSch behavior, which is no longer relevant.

EndzeitBegins:
Perfect. 🙂 I applied your changes regarding the comment removal as is.
As always, thank you for your review @exceptionfactory. 👍🏻

exceptionfactory (approved):
Thanks again for the improvement @EndzeitBegins! +1 merging

PR #8509, opened 2024-03-14
NIFI-12908 provide powershell versions of the nifi-env.cmd and nifi.cmd

phreed:
# Summary

[NIFI-12908](https://issues.apache.org/jira/browse/NIFI-12908)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/bin/nifi-env.ps1
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/bin/nifi-env.ps1
@@ -1,0 +1,29 @@
+# 
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+# 
+#    http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# 
+
+#  Set Java version
+#  $env:JAVA_HOME = "C:\Program Files\Java\jdk"
+
+#  Set application home directory based on parent directory of script location
+$myHome = $PSScriptRoot
+$env:NIFI_HOME = (Split-Path -Path $myHome -Parent)
+
+#  Set run directory for process identifier tracking
+$env:NIFI_PID_DIR = "${env:NIFI_HOME}\run"
+
+#  Set application log directory
+$env:NIFI_LOG_DIR = "${env:NIFI_HOME}\logs"
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/bin/nifi.ps1
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/bin/nifi.ps1
@@ -1,0 +1,100 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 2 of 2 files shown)


exceptionfactory (commented):
Thanks for the contribution @phreed.

As you noted in the Jira issue, the `nifi.cmd` script is not clearly highlighted in documentation, so that could also be a useful improvement.

Can you provide some additional background on the reason for introducing a PowerShell alternative to the `nifi.cmd` script for Windows? It would be ideal to maintain a unified approach for Windows, as opposed to maintaining two separate scripts for the same OS.

phreed:
> Can you provide some additional background on the reason for introducing a PowerShell alternative to the `nifi.cmd` script for Windows? It would be ideal to maintain a unified approach for Windows, as opposed to maintaining two separate scripts for the same OS.

I am just starting out with NiFi, so I will have to speak in generalities.

Making changes to *.cmd/*.bat files is fraught.
Every time I think I am going to make a small change to one it ends up being a time sink.

I am planning on extensively using NiFi for a project I am working on.
A lot of the work will involve dynamic deployment of NiFi components on a dynamic cluster.
I anticipate building tooling to support this work.
I would much prefer to use powershell over CMD/BAT.
I plan on contributing this tooling back to Apache.

It may be that the tooling would be better written in one of the JVM scripting languages (i.e. Groovy, Kotlin, Clojure).
That said I expect that some of the tooling would still best be written as powershell modules on Windows.

Obviously, the `nifi.ps1` is not (yet) a powershell module, I am getting a feel for how the NiFi project operates.
I would like to see the `nifi.cmd` deprecated in favor of  powershell modules.

exceptionfactory:
> > Can you provide some additional background on the reason for introducing a PowerShell alternative to the `nifi.cmd` script for Windows? It would be ideal to maintain a unified approach for Windows, as opposed to maintaining two separate scripts for the same OS.
> 
> I am just starting out with NiFi, so I will have to speak in generalities.
> 
> Making changes to _.cmd/_.bat files is fraught (for me). Every time I think I am going to make a small change to one it ends up being a time sink.
> 
> I am planning on extensively using NiFi for a project I am working on. A lot of the work will involve dynamic deployment of NiFi components on a dynamic cluster. I anticipate building tooling to support this work. I would much prefer to use powershell over CMD/BAT. I plan on contributing this tooling back to Apache.
> 
> It may be that the tooling would be better written in one of the JVM scripting languages (i.e. Groovy, Kotlin, Clojure). That said I expect that some of the tooling would still best be written as powershell modules on Windows.
> 
> Obviously, the `nifi.ps1` is not (yet) a powershell module, I am getting a feel for how the NiFi project operates. I would like to see the `nifi.cmd` deprecated in favor of powershell modules.

Thanks for additional background @phreed.

The shell scripts are intended to provide a minimum wrapper around the environment settings necessary to start NiFi Java processes. For that reason, these scripts should be as minimal as possible, and provide the broadest base of compatibility. That is why the current scripts for Windows do not use PowerShell, and should avoid additional complexity. That is also the reason why they do not use any type of JVM scripting.

Running NiFi on Windows is supported in general, but running scalable production deployments most often depends on Linux as the operating system.

So with that background, any shell script changes should continue to follow the minimalist approach to maintain compatibility across modern distributions of Windows. With that in mind, it may be best to avoid introducing PowerShell scripts in favor of retaining the limited scope of command scripting. This does highlight the need for reviewing the general approach of Windows startup scripts, including the removal of the `BAT` scripts.

If the PowerShell approach works without installing any additional Windows packages on current versions of both Home and Pro, then this seems like a potential option. Are you aware of any compatibility issues with PowerShell on recent Windows versions, particularly on non-server editions?

phreed:
Not compatibility issues per se but powershell does have a stronger security model requiring changing the default security settings. So, maybe dropping this PR would be the best move. I originally make this port because there was a problem with the nifi.cmd and rather than debug nifi.cmd I did this port. 
I mostly use linux, so I am far from a powershell expert (but I really do not like bat/cmd).

exceptionfactory:
> Not compatibility issues per se but powershell does have a stronger security model requiring changing the default security settings. So, maybe dropping this PR would be the best move. I originally make this port because there was a problem with the nifi.cmd and rather than debug nifi.cmd I did this port. I mostly use linux, so I am far from a powershell expert (but I really do not like bat/cmd).

That's good to know. NiFi usage is much more common on Linux and macOS, but development on Windows is common enough. Based on the security requirements you mentioned, I agree that closing the current pull request is the best option for now.

If you are able to track down issues with the `nifi.cmd`, that would be helpful to maintain.

exceptionfactory:
@phreed I am closing this pull request for now based on the discussion. As mentioned, if you find particular problems with the current `nifi.cmd`, it would be helpful to highlight those in the Jira issue so we can track down the problem.

PR #8532, opened 2024-03-19
NIFI-5302 Add Support for Client Credentials Flow with OIDC

exceptionfactory:
# Summary

[NIFI-5302](https://issues.apache.org/jira/browse/NIFI-5302) Adds supports for the [OAuth 2.0 Client Credentials Flow](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/) when NiFi is configured for [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html) authentication.

The implementation changes maintain existing support for the [Authorization Code Grant Flow](https://www.oauth.com/oauth2-servers/server-side-apps/authorization-code/) as part of standard browser-based access, but enable machine-to-machine authentication using Client Credentials. Most of the code changes involve reorganizing the OIDC and JWT configuration objects for reuse between configuration classes.

The configuration changes support verifying Access Tokens based on the provided Issuer claim. When the Token Issuer matches the OIDC Client Registration Issuer, NiFi uses the Identity Provider signing key for verification. In all other cases, NiFi uses the existing application signing key for verification, maintaining current behavior.

Enabling machine-to-machine access with the Client Credentials Flow may also require authorization configuration changes depending on the deployed environment. The OpenID Connect fallback claims identifying user property may also be necessary to use the `sub` claim as the identity instead of the default `email` claim. These configuration elements can be enabled using existing configuration properties and do not require additional code changes.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-docs/src/main/asciidoc/administration-guide.adoc
+++ b/nifi-docs/src/main/asciidoc/administration-guide.adoc
@@ -549,6 +549,19 @@ Access Token has expired, indicating that the Resource Owner has terminated the 
 termination occurs when the user closes the browser without initiating the logout process. The scheduled process avoids
 extended storage of Refresh Tokens for users who are no longer interacting with the application.
 
+The OpenID Connect implementation also supports the OAuth 2 Client Credentials Grant Type as described in
+link:https://www.rfc-editor.org/rfc/rfc6749#section-4.4[RFC 6749 Section 4.4^]. With OpenID Connect integration enabled,
+NiFi evaluates the JSON Web Token Issuer Claim named `iss` and delegates to either the configured Authorization Server
+or internal processing for signature verification. When the `iss` claim value matches the `issuer` from the OpenID
+Connect Discovery Configuration, NiFi uses the JSON Web Keys from the Authorization Server for signature verification.
+In all other cases, NiFi verifies JSON Web Token signatures using an internal public key.
+
+The Client Credentials Grant Type enables machine-to-machine authentication and requires token request processing outside
+of NiFi itself to obtain an Access Token. NiFi must also be configured to authorize requests based on the identity
+defined in a signed Access Token. Access Tokens obtained using the Client Credentials Grant Type do not include the
+standard `email`, which requires configuring a fallback claim to identify the machine user. The most common claim for
+identification is the Subject Claim named `sub`, which contains the Client ID.
+
 OpenID Connect integration supports the following settings in _nifi.properties_.
 
 [options="header"]
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-security/src/main/java/org/apache/nifi/web/security/configuration/AuthenticationSecurityConfiguration.java
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-security/src/main/java/org/apache/nifi/web/security/configuration/AuthenticationSecurityConfiguration.java
@@ -34,7 +34,9 @@ import org.springframework.security.authentication.AuthenticationManager;
  */
 @Configuration
 @Import({
+        ClientRegistrationConfiguration.class,
         JwtAuthenticationSecurityConfiguration.class,
+        JwtDecoderConfiguration.class,
         KerberosAuthenticationSecurityConfiguration.class,
         KnoxAuthenticationSecurityConfiguration.class,
         OidcSecurityConfiguration.class,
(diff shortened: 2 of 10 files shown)


pvillard31:
Hey @exceptionfactory - does that mean that it provides a way to use the NiFi CLI without mTLS when NiFi is configured with OIDC for Authentication? If yes, is it possible to provide examples? It also could be worth updating the documentation to reflect that option. I know it's always painful for users to leverage the CLI when both keystore and truststore are required (even if you can use the proxied entity feature).

exceptionfactory:
> Hey @exceptionfactory - does that mean that it provides a way to use the NiFi CLI without mTLS when NiFi is configured with OIDC for Authentication? If yes, is it possible to provide examples? It also could be worth updating the documentation to reflect that option. I know it's always painful for users to leverage the CLI when both keystore and truststore are required (even if you can use the proxied entity feature).

Thanks for asking @pvillard31.

This improvement provides the foundation for alternative authentication using the NiFi CLI, but additional work is required in the CLI itself to enable an alternative to mTLS.

I agree that mTLS can be challenging to configure for automated integration, so this pull request provides a foundation for future improvements. With this change, it will be possible to enhance the CLI to support OAuth 2 Client Credentials or Device Flow authentication. That will require additional consideration, as obtaining an Access Token often involves multiple steps. However, with these changes in place, there will be more options for integration.

I pushed a commit updating the OpenID Connect section of the Admin Guide highlighting support for the Client Credentials Grant Type and outlining the basics of the implementation.

pvillard31:
Thanks @exceptionfactory that makes sense! Can you share some details on how this can be leveraged as-is? Trying to understand how this can be tested and how this improvement would be used with NiFi deployments.

exceptionfactory:
> Thanks @exceptionfactory that makes sense! Can you share some details on how this can be leveraged as-is? Trying to understand how this can be tested and how this improvement would be used with NiFi deployments.

Sure, good question.

The exact details vary between Identity Providers, but the basic process for exercising this new capability is as follows:

1. Configure an Application Client on the Identity Provider, enabled for the Client Credentials Grant Type
2. Get the Client ID and Client Secret values associated with the Application Client from the Identity Provider
3. Make an HTTP request to the Identity Provider requesting an Access Token using the Client ID and Client Secret values
4. Make an HTTP request to the NiFi REST API using the `access_token` value returned and passing it to NiFi using the `Authorization` header with the `Bearer` scheme

The first two steps are specific to the Identity Provider configured.

Step 3 is from the OAuth 2 specification for Client Credentials as described in [RFC 6749 Section 4.4.2](https://www.rfc-editor.org/rfc/rfc6749#section-4.4.2).

Step 4 uses existing NiFi REST API handling with the HTTP `Authorization` header.

Here are some concrete examples for several Identity Providers on how to configure a Client Credentials Grant Flow:

- [Okta Client Credentials Flow](https://developer.okta.com/docs/guides/implement-grant-type/clientcreds/main/#client-credentials-flow)
- [AWS Cognito Token Endpoint for Client Credentials](https://docs.aws.amazon.com/cognito/latest/developerguide/token-endpoint.html#post-token-positive-exchanging-client-credentials-for-an-access-token)

Step 3 could be part of future enhancements to the NiFi CLI. For example, with an OpenID Connect Discovery URL, a Client ID, and a Client Secret, the NiFi CLI could make a request for an Access Token and then make a request to the NiFi REST API with the customized Authorization header.

As a way to test this using NiFi itself, the [StandardOauth2AccessTokenProvider](https://nifi.apache.org/documentation/nifi-2.0.0-M2/components/org.apache.nifi/nifi-oauth2-provider-nar/2.0.0-M2/org.apache.nifi.oauth2.StandardOauth2AccessTokenProvider/index.html) could be configured with the `Client Credentials` Grant Type, and InvokeHTTP could be used to invoke NiFi REST API methods.

This would be worth a blog post at some point, but hopefully this provides sufficient background.

exceptionfactory:
As mentioned in the updated documentation, testing this capability also requires the following changes:

1. Set the property `nifi.security.user.oidc.fallback.claims.identifying.user` to `sub` so that NiFi will use the Access Token subject claim to identify users in addition to the standard `email` claim
2. Add the Client ID of the Application Client to the list of users authorized to invoke NiFi REST API methods. This can be done using the standard Managed Authorizer, or other Authorize that supports multiple users.

pvillard31:
Understood, thanks, will give it a try asap if no one else can do it before.

exceptionfactory:
Rebased to incorporate latest updates to Spring Framework 6.1. Replaced deprecated `OkHttp3ClientHttpRequestFactory` with `JdkClientHttpRequestFactory`.

raditpan:
Hi @exceptionfactory , I have tested your branch with my test Keycloak server and my group authorization configs. And it works as expected. I can make calls to Nifi REST API with token from client credentials flow.

> Add the Client ID of the Application Client to the list of users authorized to invoke NiFi REST API methods.

I'd like to added that, to make the Nifi REST API access works properly, the username added to NiFi's list of users authorized need to have the same name as the `sub` value of Token subject claim. This part is not clear to me when I first tried it.

exceptionfactory:
Thanks for the feedback @raditpan! Glad to hear it works as expected.

You are correct that the Client ID from the `sub` claim needs to be authorized, which follows the pattern of other authentication strategies such as client certificates, where the certificate subject needs to be authorized for invoking REST API resources.

bbende (approved):
Also verified running this with Keycloak and was able to get the access token and use it to make API calls to NiFi, and also could do normal web login to NiFi through Keycloak as a regular user. Looks good!

PR #8544, opened 2024-03-22
NIFI-12923 Added append avro mode to PutHDFS

balazsgerner:
# Summary

[NIFI-12923](https://issues.apache.org/jira/browse/NIFI-12923)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/PutHDFS.java
```diff
@@ -397,10 +432,28 @@ public void process(InputStream in) throws IOException {
...
-                                StreamUtils.copy(bis, fos);
-                                bis = null;
-                                fos.flush();
+
+                                final String appendMode = context.getProperty(APPEND_MODE).getValue();
+                                if (APPEND_RESOLUTION.equals(conflictResponse)
+                                        && AVRO_APPEND_MODE.equals(appendMode)
+                                        && destinationExists) {
+                                    getLogger().info("Appending avro record to existing avro file");
+                                    try (final var reader = new DataFileStream<>(in, new GenericDatumReader<>());
```
mattyb149, line 441: This won't backport to 1.x because `var` wasn't introduced until Java 10 and we still have to support Java 8 on 1.x for now.

exceptionfactory: One additional note, use of `var` should be avoided in general, even when supported in the main branch.

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/PutHDFS.java
```diff
@@ -160,6 +170,14 @@ public class PutHDFS extends AbstractHadoopProcessor {
             .allowableValues(WRITE_AND_RENAME_AV, SIMPLE_WRITE_AV)
             .build();
 
+    public static final PropertyDescriptor APPEND_MODE = new PropertyDescriptor.Builder()
+            .name("Append mode")
```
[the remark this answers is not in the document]
balazsgerner: Fixed

nifi-nar-bundles/nifi-hadoop-bundle/nifi-hdfs-processors/src/main/java/org/apache/nifi/processors/hadoop/PutHDFS.java
```diff
@@ -160,6 +170,14 @@ public class PutHDFS extends AbstractHadoopProcessor {
             .allowableValues(WRITE_AND_RENAME_AV, SIMPLE_WRITE_AV)
             .build();
 
+    public static final PropertyDescriptor APPEND_MODE = new PropertyDescriptor.Builder()
+            .name("Append Mode")
+            .description("Defines the append strategy to use when the Conflict Resolution Strategy is set to 'append'.")
+            .allowableValues(DEFAULT_APPEND_MODE, AVRO_APPEND_MODE)
+            .defaultValue(DEFAULT_APPEND_MODE)
+            .required(true)
+            .build();
```
turcsanyip, line 179: `Append Mode` property should depend on `Conflict Resolution Strategy` and should only be shown if the strategy is `append`.
I also suggest placing the new property just below `Conflict Resolution Strategy`.

balazsgerner: Fixed

mattyb149:
In order for the append to work I had to set `Writing Strategy` to `Simple write`, if I leave the default `Write and rename`, it actually deletes the file. Is this intended? If not we should add another dependent property so if AVRO is chosen, the Writing Strategy must be Simple Write.

balazsgerner:
> In order for the append to work I had to set `Writing Strategy` to `Simple write`, if I leave the default `Write and rename`, it actually deletes the file. Is this intended? If not we should add another dependent property so if AVRO is chosen, the Writing Strategy must be Simple Write.

To be honest, during the whole dev testing I left the 'Writing Strategy' property on default value which is 'Write and rename', and it worked completely fine for me. I double checked it now and it still does. As I see it, my ticket does not affect the connection between writing strategy and conflict resolution strategy so it should work just as before.

I also checked 'Writing strategy'='Simple write', and that also works for me without any problems. 

If you experience file deletion I assume your flow might need a tweak. I do not know the processors you use exactly, but I had file deletion problems when I left the 'GetHDFS' or the 'GetFile' processors 'Keep source file' on default which is 'false'. In order to make my flow reexecutable I changed that value to 'true'.

For testing I used the following flow:
1. GetFile('Keep source file'='true') -> PutHDFS(append,AVRO,'Write and rename'), but it also works with 'Simple write'
2. GetHDFS('Keep source file'='true') -> ConvertAvroToJSON (I just used this step to validate the content of the avro file, not sure if the convert processor exists upstream, I used a custom nifi instance for this)

balazsgerner:
I had to rebase to `main` in order resolve merge conflicts. Sorry for the force push.

mattyb149:
Now I'm running into [this](https://github.com/apache/hadoop/blob/rel/release-3.4.0/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/ChecksumFileSystem.java#L590) in my test environment, trying to change the default filesystem now.

mattyb149:
+1 LGTM, verified the Avro append capability works as expected. Thanks for the improvement! Merging to main and support/nifi-1.x

PR #8564, opened 2024-03-25
NIFI-12948: Various bugs discovered when migrating to Angular 17

mcgilman:
NIFI-12948:
- Action is optional.
- Removing duplicate check in counter table.
- Ensure flow configuration is loaded in route guard.
- Only show user/group list when there are users or groups.
- Default override policy to copy.


scottyaslan (approved):
LGTM! Thanks @mcgilman!

PR #8565, opened 2024-03-25
NIFI-12950: Updating front end to use the uiOnly flag where possible

mcgilman:
NIFI-12950:
- Updating front end to use the uiOnly flag where possible.

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
@@ -20,6 +20,7 @@ import { FlowService } from '../../service/flow.service';
 import { Actions, concatLatestFrom, createEffect, ofType } from '@ngrx/effects';
 import * as FlowActions from './flow.actions';
 import * as StatusHistoryActions from '../../../../state/status-history/status-history.actions';
+import * as ErrorActions from '../../../../state/error/error.actions';
 import {
     asyncScheduler,
     catchError,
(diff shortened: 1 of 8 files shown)


rfellows:
reviewing...

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
```diff
@@ -1069,6 +1069,16 @@ export class FlowEffects {
...
                 map((action) => action.request),
+                switchMap((request) =>
+                    from(this.flowService.getProcessor(request.entity.id)).pipe(
+                        map((entity) => {
+                            return {
+                                ...request,
+                                entity
+                            };
+                        })
+                    )
```
rfellows, line 1080: Do we need to consider the possibility of this endpoint failing?
When I force an error in the service and try to edit a processor config, nothing happens. Should we pop up a snackbar in this rare case? Or is ignoring it sufficient?

rfellows (commented):
It looks like we didn't need to change any consumer of these services that now accept the `uiOnly` argument... So, maybe we don't need to make it an argument at all if nothing overrides the default value. However, I'm not opposed to keeping it if we think there might be a scenario in the future where we think this might be needed.

Also found one place we might need some error handling.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
rfellows: I'm fine with not addressing it here as long it is accounted for.

rfellows (approved):
Looks good. I'll merge once CI finishes 👍

PR #8580, opened 2024-03-28
[NIFI-12941] - adding in jest-preset-angular to enable unit testing in Idea based IDEs

rfellows:
[NIFI-12941](https://issues.apache.org/jira/browse/NIFI-12941)


mcgilman (approved):
Thanks @rfellows!

PR #8582, opened 2024-03-29
NIFI-12972 - Only show selected relationships in read-only connection details

pvillard31:
# Summary

[NIFI-12972](https://issues.apache.org/jira/browse/NIFI-12972) - Only show selected relationships in read-only connection details

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/connection-details.jsp
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/connection-details.jsp
@@ -39,7 +39,6 @@
                     <div id="read-only-relationship-names-container" class="setting">
                         <div class="setting-name">
                             Relationships
-                            <div class="fa fa-question-circle" alt="Info" title="Selected relationships are in bold."></div>
                         </div>
                         <div class="setting-field">
                             <div id="read-only-relationship-names"></div>
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/connection-details.jsp
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/connection-details.jsp
@@ -38,7 +38,7 @@
                     </div>
                     <div id="read-only-relationship-names-container" class="setting">
                         <div class="setting-name">
-                            Relationships
+                            Selected relationships
                         </div>
                         <div class="setting-field">
                             <div id="read-only-relationship-names"></div>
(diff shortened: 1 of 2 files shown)


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/nf-connection-details.js
```diff
@@ -496,10 +496,10 @@
...
 
-                        // show the available relationship if applicable
+                        // show the selected relationships if applicable
                         if (nfCommon.isDefinedAndNotNull(availableRelationships) || nfCommon.isDefinedAndNotNull(selectedRelationships)) {
-                            // populate the available connections
-                            $.each(availableRelationships, function (i, name) {
+                            // populate the selected connections
+                            $.each(selectedRelationships, function (i, name) {
                                 createRelationshipOption(name);
                             });
```
mcgilman, line 504: We need to account for dynamic Relationships that are selected within this Connection but have since been deleted from the source Processor. Previously we would render all Relationships here and in the block below we would identify any selected relationships that are no longer available and assign them the `undefined` class. Since now we're only rendering the selected relationships we'll probably want to handle that here and remove the block below.

Note: There is admittedly a bug in the existing implementation because `createRelationshipOption` is not actually returning the new option. It will need to be updated like:

```
    var createRelationshipOption = function (name) {
        return $('<div class="available-relationship-container"></div>').append(
            $('<div class="relationship-name"></div>').text(name)).appendTo('#read-only-relationship-names');
    };
```

You should see the Relationship that was removed is rendered in a lighter color and in italic.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/WEB-INF/partials/connection-details.jsp
```diff
@@ -39,7 +39,6 @@
                     <div id="read-only-relationship-names-container" class="setting">
                         <div class="setting-name">
                             Relationships
-                            <div class="fa fa-question-circle" alt="Info" title="Selected relationships are in bold."></div>
```
mcgilman, line 42: Should this label also be updated to `Selected Relationships` so there's no confusion over what we're showing?

mcgilman (commented):
Thanks for the PR @pvillard31! I've left a few notes below. There is a corner case that we should account for.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui/src/main/webapp/js/nf/nf-connection-details.js
pvillard31: Thanks Matt for catching this. I just pushed a new commit with the suggested changes.

Screenshot 2024-04-01 at 09 50 11
Screenshot 2024-04-01 at 09 50 20

mcgilman (approved):
Thanks for the updates @pvillard31! +1

PR #8584, opened 2024-03-29
NIFI-7085 add flowFile batching to ConsumeJMS and PublishJMS

mosermw:
# Summary

[NIFI-7085](https://issues.apache.org/jira/browse/NIFI-7085)

This PR adds batching to PublishJMS and ConsumeJMS in the form of a new property. ConsumeJMS can read more than one message from the JMS broker per triggered task. PublishJMS can publish more than one flowFile to the JMS broker per triggered task. This helps get higher message throughput to JMS in busy NiFi environments.

The SupportsBatching annotation was considered, but it is not appropriate for processors that send/receive data from an external source.

The new Maximum Batch Size property has a default value of 1, to match old behavior. ConsumeJMS in record processing mode used to have a 10,000 max batch size which is replaced by this property value. ConsumeJMS will migrate the new Maximum Batch Size property to 10000 if it was not set and it was in record processing mode.

Fixed an bug in JMSConsumer:consumeMessageSet where it could read, but not acknowledge or commit, an extra message when batchCounter = MAX_MESSAGES_PER_FLOW_FILE

Fixed JMSPublisherConsumerIT integration test which didn't account for when consumeMessage() did not read a message.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/AbstractJMSProcessor.java
+++ b/nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/AbstractJMSProcessor.java
@@ -130,7 +130,7 @@ public abstract class AbstractJMSProcessor<T extends JMSWorker> extends Abstract
 
     static final PropertyDescriptor MAX_BATCH_SIZE = new PropertyDescriptor.Builder()
             .name("Maximum Batch Size")
-            .description("The number of messages to publish or consume in a single iteration of the processor.")
+            .description("The maximum number of messages to publish or consume in each invocation of the processor.")
             .required(true)
             .defaultValue("1")
             .addValidator(StandardValidators.createLongValidator(1, 10_000, true))
--- a/nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/AbstractJMSProcessor.java
+++ b/nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/AbstractJMSProcessor.java
@@ -128,6 +128,14 @@ public abstract class AbstractJMSProcessor<T extends JMSWorker> extends Abstract
             .identifiesControllerService(JMSConnectionFactoryProviderDefinition.class)
             .build();
 
+    static final PropertyDescriptor MAX_BATCH_SIZE = new PropertyDescriptor.Builder()
+            .name("Maximum Batch Size")
+            .description("The number of messages to publish or consume in a single iteration of the processor.")
+            .required(true)
+            .defaultValue("1")
+            .addValidator(StandardValidators.createLongValidator(1, 10_000, true))
+            .build();
+
     static final List<PropertyDescriptor> JNDI_JMS_CF_PROPERTIES = Collections.unmodifiableList(
             JndiJmsConnectionFactoryProperties.getPropertyDescriptors().stream()
                     .map(pd -> new PropertyDescriptor.Builder()
--- a/nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/PublishJMS.java
+++ b/nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/PublishJMS.java
@@ -29,6 +29,7 @@ import org.apache.nifi.annotation.behavior.SystemResourceConsideration;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.SeeAlso;
 import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.annotation.lifecycle.OnScheduled;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.RequiredPermission;
 import org.apache.nifi.expression.ExpressionLanguageScope;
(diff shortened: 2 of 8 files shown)


mosermw:
Rebased to account for nifi-nar-bundles moving to nifi-extension-bundles.  Removed the JMSPublisherConsumerIT testMultipleThreads integration test fixes since that test itself was removed.

This MR may not seem like an appealing change, but ConsumeJMS already supported batching messages when using a Record Reader/Writer.  This MR promotes batching to be available in all configurations.  And as a reminder this MR fixes an bug in JMSConsumer consumeMessageSet() where it could read then drop an extra message when batchCounter reaches MAX_MESSAGES_PER_FLOW_FILE.

mosermw:
Rebased onto main to resolve latest conflicts.

nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/PublishJMS.java
```diff
@@ -211,8 +212,12 @@ public class PublishJMS extends AbstractJMSProcessor<JMSPublisher> {
...
     @Override
     protected void rendezvousWithJms(ProcessContext context, ProcessSession processSession, JMSPublisher publisher) throws ProcessException {
-        FlowFile flowFile = processSession.get();
-        if (flowFile != null) {
+        final List<FlowFile> flowFiles = processSession.get(context.getProperty(MAX_BATCH_SIZE).asInteger());
+        if (flowFiles.isEmpty()) {
+            return;
+        }
+
+        flowFiles.forEach(flowFile -> {
```
jrsteinebrey, line 220: @mosermw Thanks for these enhancements to JMS consume and publish. 
Please review the existing code inside the new added loop and look for ways to optimize it. 
For example, move these variables to be created once before the loop if possible:
  allowIllegalChars
  attributeHeaderRegex 
  pattern
  readerFactory
  writerFactory
  flowFileReader

nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/AbstractJMSProcessor.java
```diff
@@ -128,6 +128,14 @@ public abstract class AbstractJMSProcessor<T extends JMSWorker> extends Abstract
             .identifiesControllerService(JMSConnectionFactoryProviderDefinition.class)
             .build();
 
+    static final PropertyDescriptor MAX_BATCH_SIZE = new PropertyDescriptor.Builder()
+            .name("Maximum Batch Size")
+            .description("The number of messages to publish or consume in a single iteration of the processor.")
```
jrsteinebrey, line 133: Thanks for making batch size consistent in these processors.
NiFi does not have for loops in the ui so I suggest using "invocation" instead of "iteration" if you don't mind.

jrsteinebrey (changes requested):

nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/PublishJMS.java
mosermw: Thanks for looking at this @jrsteinebrey.  Great suggestion.  I did this for the vars you mentioned except for flowFileReader.  Creating flowFileReader outside the loop would require checking the RECORD_READER property both outside and inside the loop.  I can do it if you feel strongly about it, but the code felt ugly when I had it in there like that.

nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/AbstractJMSProcessor.java
mosermw: I fixed the wording, and also stealthily made the same change to the GetFile Batch Size description.

mosermw:
I reinstated the JMSPublisherConsumerIT testMultipleThreads integration test, with fixes to test consuming a batch of messages and making it reliable.

nifi-extension-bundles/nifi-jms-bundle/nifi-jms-processors/src/main/java/org/apache/nifi/jms/processors/PublishJMS.java
jrsteinebrey: @mosermw Thanks for the changes. Your points on flowFileReader make perfect sense so I agree to leave that one as is.

joewitt:
@mosermw Sorry not sure why this has sat so long.  From a quick visual scan it seems ok.  Am building locally and will run all the tests/integration and otherwise.

Can you share more on whether you tested this fix before/after to verify that, in fact, performance is demonstrably better?

mosermw:
Thanks for looking at this @joewitt.  I was going to ask for eyes after the 2.0.0 push was over, so I appreciate you finding this now.

I probably spent >60 hours testing this, and my notes are on my work computer so I'll try to remember.  My setup was activemq and nifi on separate EC2 instances, with PublishJMS pushing messages as fast as it can and ConsumeJMS reading messages in various scenarios.  Under ideal conditions (100B message size, nothing else running in NiFi), 1 batch size versus 25 did **160k versus 350k** (messages per 5 mins), so roughly double.  Under more real-world conditions (5KB message size, NiFi busy doing lots of things) 1 versus 25 batch size did **30k versus 300k**, which was more noticeable.  Super big batch size like 10,000 didn't perform much different than 25 in my environment.

I even put a speedbump latency generator between activemq and nifi, but results were predictable.  Though it did seem like PublishJMS was more affected by latency than ConsumeJMS, I didn't dive deeper there.

PR #8589, opened 2024-03-31
NIFI-12982 Extend test suite of MockProcessSession

EndzeitBegins:
# Summary

[NIFI-12982](https://issues.apache.org/jira/browse/NIFI-12982)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-mock/src/test/java/org/apache/nifi/util/TestMockProcessSession.java
+++ b/nifi-mock/src/test/java/org/apache/nifi/util/TestMockProcessSession.java
@@ -16,6 +16,9 @@
  */
 package org.apache.nifi.util;
 
+import org.apache.nifi.annotation.behavior.Stateful;
+import org.apache.nifi.components.state.Scope;
+import org.apache.nifi.components.state.StateMap;
 import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.flowfile.attributes.CoreAttributes;
 import org.apache.nifi.processor.AbstractProcessor;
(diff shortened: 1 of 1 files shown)


nifi-mock/src/test/java/org/apache/nifi/util/TestMockProcessSession.java
```diff
@@ -45,189 +52,588 @@
...
-        fillBuffer(in, buffer);
+    private final Processor processor = new TestProcessor();
+    private final SharedSessionState sharedState = new SharedSessionState(processor, new AtomicLong(0L));
+    private final MockStateManager stateManager = new MockStateManager(processor);
+    private final MockProcessSession session = new MockProcessSession(sharedState, processor, stateManager);
+
+    private final Processor statefulProcessor = new StatefulTestProcessor();
+    private final SharedSessionState sharedStateOfStatefulProcessor = new SharedSessionState(statefulProcessor, new AtomicLong(0L));
+    private final MockStateManager stateManagerOfStatefulProcessor = new MockStateManager(statefulProcessor);
+    private final MockProcessSession sessionOfStatefulProcessor = new MockProcessSession(sharedStateOfStatefulProcessor, statefulProcessor, stateManagerOfStatefulProcessor);
```
exceptionfactory, line 63: As a general rule, reusing an instance of an object across test method invocations can lead to unexpected results.

exceptionfactory (commented):
Thanks for working on additional tests for `MockProcessSession` @EndzeitBegins. Instead of creating `final` instances of various objects, did you consider creating them in a more localized scope? Building on the `Nested` approach, it is possible to have before and after methods that could apply to particular nested groups of tests if needed.

nifi-mock/src/test/java/org/apache/nifi/util/TestMockProcessSession.java
EndzeitBegins: Thank you for your feedback @exceptionfactory. 
I totally agree!

However, as far as I'm aware the default `@TestInstance` lifecycle of JUnit 5 is `PER_METHOD`. 
That is, JUnit 5 is creating a new instance of the `TestMockProcessSession` class for every test method invoked. 
As a result, those objects aren't actually reused between test methods. 

I started out having the "stateful" variants in a single nested scope, but ended up needing them in another nested scope, as far as I remember, which is why I moved them up. 

I'm fine fine with making these field non final and moving those instantiations in a `@BeforeEach` if that's preferred. From my perspective (using mainly Kotlin) it's just more verbose.

exceptionfactory: Thanks for the helpful reply @EndzeitBegins, very good point regarding the standard per-method behavior. In light of this, and the fact that there are no Mockito-based mock objects, the current approach makes sense.

exceptionfactory (approved):
Thanks again for the work on this test class, and for the helpful reply. Based on a further review, this looks good. +1 merging

EndzeitBegins:
You're welcome. Thank you for the review and merge, @exceptionfactory.

PR #8598, opened 2024-04-03
WIP NIFI-12996 removing commons-compress from shared nar/bom and ensuring…

joewitt:
… optional deps depended on are synced on version

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
@@ -150,6 +150,10 @@
             <artifactId>commons-compress</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.github.luben</groupId>
+            <artifactId>zstd-jni</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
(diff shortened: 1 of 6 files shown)


exceptionfactory (commented):
Thanks for evaluating the details of these optional dependencies @joewitt.

Based on the number of references to Commons Compress listed, it seems like it would be better to move the optional compression libraries into the `nifi-standard-shared-nar`. That would make them available in the same ClassLoader, as required for standard operation.

I agree that the Netty dependencies, in particular `netty-codec`, need to be evaluated along these lines.

joewitt:
paraphrasing you're basically saying "Hey this is great.  How about doing it exactly the opposite way" :).  Haha all good.  You're probably right.  Will give that a go.  only means we need zstd as the other deps are direct deps and bypass commons compress anyway.

joewitt:
@exceptionfactory Ok I believe I've updated to what you've described.  Tested on a live install and it works as well.  Checked where commons-compress and zstd libs are now as well and checks out

exceptionfactory (approved):
Thanks for making the adjustments @joewitt, the latest version looks good and resolves the issue. We can address any other issues with other libraries separately if needed. +1 merging

PR #8606, opened 2024-04-05
NIFI-13001: Allowing the user to acknowledge changes to the cluster connection state

mcgilman:
NIFI-13001:
- Allowing the user to acknowledge changes to the cluster connection state.


rfellows (commented):
Overall, really good. I found one or two minor things that really aren't an issue but just simplify the ClusterConnectionService a bit.

rfellows (changes requested):
This needs rebased.

PR #8609, opened 2024-04-05
NIFI-13005: Cluster State in Component State Dialog

mcgilman:
NIFI-13005:
- Adding support for showing cluster state in the component state dialog.

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/state/component-state/index.ts
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/state/component-state/index.ts
@@ -42,6 +42,12 @@ export interface StateEntry {
     clusterNodeAddress?: string;
 }
 
+export interface StateItem {
+    key: string;
+    value: string;
+    scope?: string;
+}
+
 export interface StateMap {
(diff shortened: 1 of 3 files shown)


rfellows (changes requested):
This is good, however I found one unrelated thing that we should clean up while updating this dialog. We should set the Close button to be `color="primary"` as it is the only button on the dialog.

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/settings/ui/registry-clients/registry-client-table/registry-client-table.component.html
```diff
@@ -101,10 +101,7 @@
                 <td mat-cell *matCellDef="let item">
                     <div class="flex items-center gap-x-3">
                         @if (canConfigure(item)) {
-                            <div
-                                class="pointer fa fa-cog"
-                                (click)="configureClicked(item, $event)"
-                                title="Edit"></div>
+                            <div class="pointer fa fa-cog" (click)="configureClicked(item, $event)" title="Edit"></div>
```
rfellows, line 104: when rebased, this will need to add `primary-color` class

rfellows (changes requested):
This needs rebased.

PR #8620, opened 2024-04-09
NIFI-12986 Tidy up JavaDoc of ProcessSession

EndzeitBegins:
# Summary

Adjusts JavaDoc of `ProcessSession` to fix some minor typos and and align documentation drifts between method overloads, in order to make the JavaDoc more consistent.

[NIFI-12986](https://issues.apache.org/jira/browse/NIFI-12986)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/processor/ProcessSession.java
+++ b/nifi-api/src/main/java/org/apache/nifi/processor/ProcessSession.java
@@ -38,123 +38,70 @@ import org.apache.nifi.processor.io.InputStreamCallback;
 import org.apache.nifi.processor.io.OutputStreamCallback;
 import org.apache.nifi.processor.io.StreamCallback;
 import org.apache.nifi.provenance.ProvenanceReporter;
+import org.apache.nifi.provenance.ProvenanceEventType;
 
 /**
- * <p>
  * A process session encompasses all the behaviors a processor can perform to
- * obtain, clone, read, modify remove FlowFiles in an atomic unit. A process
- * session is always tied to a single processor at any one time and ensures no
- * FlowFile can ever be accessed by any more than one processor at a given time.
- * The session also ensures that all FlowFiles are always accounted for. The
- * creator of a ProcessSession is always required to manage the session.
- * </p>
- *
+ * obtain, clone, read, modify remove {@link FlowFile}s in an atomic unit.
+ * A process session is always tied to a single {@link Processor} at any one time
+ * and ensures no FlowFile can ever be accessed by any more than one processor at a given time.
+ * The session also ensures that all FlowFiles are always accounted for.
+ * The creator of a process session is always required to manage the session.
  * <p>
- * A session is not considered thread safe. The session supports a unit of work
- * that is either committed or rolled back
- * </p>
- *
+ * A session is not considered thread safe. The session supports a unit of work that is either committed or rolled back.
  * <p>
- * As noted on specific methods and for specific exceptions automated rollback
- * will occur to ensure consistency of the repository. However, several
- * situations can result in exceptions yet not cause automated rollback. In
- * these cases the consistency of the repository will be retained but callers
- * will be able to indicate whether it should result in rollback or continue on
- * toward a commit.
- * </p>
- *
+ * As noted on specific methods and for specific exceptions
+ * automated rollback will occur to ensure consistency of the repository.
+ * However, several situations can result in exceptions yet not cause automated rollback.
+ * In these cases the consistency of the repository will be retained
+ * but callers will be able to indicate whether it should result in rollback or continue on toward a commit.
  * <p>
- * A process session has two 'terminal' methods that will result in the process session
- * being in a 'fresh', containing no knowledge or any FlowFile, as if the session were newly
- * created. After one of these methods is called, the instance may be used again. The terminal
- * methods for a Process Session are the {@link #commit()} and {@link #rollback()}. Additionally,
- * the {@link #migrate(ProcessSession, Collection)} method results in {@code this} containing
- * no knowledge of any of the FlowFiles that are provided, as if the FlowFiles never existed in
- * this ProcessSession. After each commit or rollback, the session can be used again. Note, however,
- * that even if all FlowFiles are migrated via the {@link #migrate(ProcessSession, Collection)} method,
- * this Process Session is not entirely cleared, as it still has knowledge of Counters that were adjusted
- * via the {@link #adjustCounter(String, long, boolean)} method. A commit or rollback will clear these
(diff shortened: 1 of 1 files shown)


nifi-api/src/main/java/org/apache/nifi/processor/ProcessSession.java
```diff
@@ -173,846 +120,783 @@ public interface ProcessSession {
...
+     *              {@link #write(FlowFile, OutputStreamCallback)}) or while a read or write stream is open
+     *              (see {@link #read(FlowFile)}, {@link #write(FlowFile)}) for any of the given {@code sources} and {@code destination} {@link FlowFile}s
+     * @throws FlowFileHandlingException if any of the given {@link FlowFile}s is already transferred or removed or doesn't belong to this session.
+     *              Automatic rollback will occur.
+     * @throws MissingFlowFileException if any of the given {@link FlowFile}'s content cannot be found.
+     *              The FlowFile should no longer be referenced, will be internally destroyed. The session is automatically rolled back.
+     * @throws FlowFileAccessException if some IO problem occurs accessing {@link FlowFile} content;
+     *              the state of the {@code destination} {@link FlowFile} will be as it was prior to this call.
+     */
+    default FlowFile merge(Collection<FlowFile> sources, FlowFile destination) {
```
exceptionfactory, line 654: This appears to be a new method.

nifi-api/src/main/java/org/apache/nifi/processor/ProcessSession.java
```diff
@@ -173,846 +120,783 @@ public interface ProcessSession {
...
-     * @throws IllegalStateException if called from within a callback (See {@link #write(FlowFile, StreamCallback)}, {@link #write(FlowFile, OutputStreamCallback)},
-     * {@link #read(FlowFile, InputStreamCallback)}).
+     * @throws IllegalStateException if detected that this method is being called from within a read or write callback
+     *              (see {@link #read(FlowFile, InputStreamCallback)}, {@link #write(FlowFile, StreamCallback)},
+     *              {@link #write(FlowFile, OutputStreamCallback)}) or while a read or write stream is open
+     *              (see {@link #read(FlowFile)}, {@link #write(FlowFile)}).
+     * @throws FlowFileHandlingException if not all {@link FlowFile}s acted upon within this session are accounted for
+     *              such that they have a transfer identified or where marked for removal. Automated rollback occurs.
+     */
+    default void commitAsync() {
```
exceptionfactory, line 133: This appears to be a new method.

exceptionfactory (changes requested):
Thanks for the documentation updates @EndzeitBegins. 

The documentation updates look good in general. However, there appear to be some new methods introduced, which goes beyond the scope of the issue description. Were those changes intentional?

EndzeitBegins:
Thank you for taking a look @exceptionfactory. I know this is a large change in terms of lines changes, so I really appreciate that.

While going through the interface declaration, I noticed that those functions are just a shorthand overload of their respective functions with more parameters.

The behaviour of calling `commitAsync()` is defined equal to calling `commitAsync(Runnable onSuccess, Consumer<Throwable> onFailure)` with both parameters set to `null`, as both parameters are optional. In fact, the same applies to the "sibling" method `commitAsync(Runnable onSuccess)`  which has an existing default implementation of `commitAsync(onSuccess, null);`. I figured adding a default for `commitAsync()` would further align both function overloads.

A similar reasoning can be applied to `merge(Collection<FlowFile> sources, FlowFile destination)` who's behavior is defined akin to that of `merge(Collection<FlowFile> sources, FlowFile destination, byte[] header, byte[] footer, byte[] demarcator)` when passing `null` for all of `header`, `footer`, and `demarcator`. However, this time there is no other sibling overload with a default.

As adding a default implementation is not a breaking change nor alters the behavior for any existing implementation I figured that's a sensible addition. However, you're right that it's out of scope of the original issue description.
With this background, what are your thoughts on this? Should I revert the addition of those default implementations or do you follow the reasoning above?

exceptionfactory:
Thanks for the helpful reply @EndzeitBegins.

I evaluated the current method implementations that align with the proposed default implementations, and it seems like introducing the new default implementation methods could be helpful. However, it probably means that the existing implementations should be removed to avoid confusion. From that perspective, it would be better to consider those changes as a separate PR, keeping this one focused solely on documentation updates.

EndzeitBegins:
Makes perfect sense @exceptionfactory.

I removed the default implementations for now and might open a separate PR once the documentation adjustments are merged.

exceptionfactory (changes requested):
Thanks for removing the default methods @EndzeitBegins. This looks close to completion, but on further review, I noted the addition of a number of opening `<p>` tags without a corresponding closing `</p>` tag. Although this can work, best practice HTML should include the closing tag. The other option is using the `<br>` tag. If you could go through and make adjustments, that would be helpful.

EndzeitBegins:
To be honest I was (and still am really) somewhat confused regarding the paragraphs. 😅

I've seen both styles applied throughout the NiFi codebase, e.g. [here](https://github.com/apache/nifi/blob/cc7af91f97dd1844cc6c8d30bd47d6865cb20aaa/nifi-api/src/main/java/org/apache/nifi/time/DurationFormat.java#L62-L85) vs. [here](https://github.com/apache/nifi/blob/62a4ece55d08d5488e42819f27bbecd864e9e7dc/nifi-api/src/main/java/org/apache/nifi/reporting/ReportingTask.java#L24-L58).
A web search didn't help that much either as it yielded ambiguous results. This [article from Oracle](https://www.oracle.com/de/technical-resources/articles/java/javadoc-tool.html#format) uses single `<p>` notation. Others suggested that that wouldn't be valid due to the closing tag missing. 

I don't have hard feelings about that. 🤷🏻‍♂️ I just took the less verbose option. 

Regardless of what we choose to use, it may be a good idea to unify on one declaration style inside NiFi, not in this PR of course, to avoid that confusion in the future. 

Do you have a verdict on what style should be preferred for NiFi @exceptionfactory?

exceptionfactory:
Thanks for taking the time to research and evaluate current approaches @EndzeitBegins. I believe this ultimately goes back to the roots of HTML where the `<p>` is not required in most cases, since HTML is not a subset of XML, requiring properly closed elements.

I lean in the direction of clarity, including the closing tag, but given that fact that it is valid without, and the example from the Oracle documentation, I am fine with leaving it as it stands, without the closing tag so we can proceed.

Thanks again for the thoughtful interaction.

PR #8628, opened 2024-04-10
[NIFI-13022] rename semantic color classes to avoid conflict with angular material class names

scottyaslan:


rfellows (commented):
@scottyaslan this looks good. I only have one question/suggestion... is there a specific reason you chose to suffix the default colors with `-default`? it just seems a bit extraneous to me.

PR #8630, opened 2024-04-11
NIFI-13026 - Add a read only mode to non secured NiFi Registry

pvillard31:
# Summary

[NIFI-13026](https://issues.apache.org/jira/browse/NIFI-13026) - Add a read only mode to non secured NiFi Registry

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/AuthorizerFactory.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/AuthorizerFactory.java
@@ -523,6 +523,9 @@ public class AuthorizerFactory implements UserGroupProviderLookup, AccessPolicyP
         return new Authorizer() {
             @Override
             public AuthorizationResult authorize(final AuthorizationRequest request) throws AuthorizationAccessException {
+                if(properties.isReadOnly() && !request.getAction().equals(RequestAction.READ)) {
+                    return AuthorizationResult.denied("NiFi Registry is configured with read only.");
+                }
                 return AuthorizationResult.approved();
             }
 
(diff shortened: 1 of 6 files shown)


nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/AuthorizerFactory.java
```diff
@@ -523,6 +523,9 @@ private Authorizer createDefaultAuthorizer() {
         return new Authorizer() {
             @Override
             public AuthorizationResult authorize(final AuthorizationRequest request) throws AuthorizationAccessException {
+                if(properties.isReadOnly() && !request.getAction().equals(RequestAction.READ)) {
```
exceptionfactory, line 526: Injecting this property check into the default is not an optimal approach because it requires evaluating the property on every request, as opposed to configuring a specific authorizer.

exceptionfactory (changes requested):
Thanks for proposing this option @pvillard31.

Having a read-only configuration for Registry seems like it could be useful in some limited scenarios, so the main question here is how common this would be, and how involved it is to maintain.

It does not seem to be common enough to warrant a first-class application property, since the majority of users would need to read and write flows. For that reason, injecting the property and checking it in the default unsecured authorizer does not seem like the right solution.

It seems like this capability could be implemented using a custom Authorizer, so then the question is whether that should be part of the standard Registry application, or whether it should be maintained outside of this repository.

pvillard31:
Thanks @exceptionfactory! What about the following approach where I have the DefaultAuthorizer moved to its own class and in authorizers.xml, have something like:

````
    <authorizer>
        <identifier>default-authorizer</identifier>
        <class>org.apache.nifi.registry.security.authorization.DefaultAuthorizer</class>
        <property name="Read Only">false</property>
    </authorizer>
````
And then instead of
````
    public Authorizer getAuthorizer() throws AuthorizerFactoryException {
        if (authorizer == null) {
            if (properties.getSslPort() == null) {
                // use a default authorizer... only allowable when running not securely
                authorizer = createDefaultAuthorizer();
            } else {
````
I would instantiate the DefaultAuthorizer the same way as the other ones are instantiated.
Thoughts?

If we think this is not worth it because it is for a very specific use case, I'll maintain my own version of the NiFi Registry.

pvillard31:
Closing as this is a very particular edge case.

PR #8631, opened 2024-04-11
NIFI-13027 - Warn users for small files processing in PutIceberg

pvillard31:
# Summary

[NIFI-13027](https://issues.apache.org/jira/browse/NIFI-13027) - Warn users for small files processing in PutIceberg

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/PutIceberg.java
+++ b/nifi-nar-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/PutIceberg.java
@@ -36,11 +36,13 @@ import org.apache.nifi.annotation.documentation.Tags;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.ValidationContext;
 import org.apache.nifi.components.ValidationResult;
+import org.apache.nifi.components.PropertyDescriptor.Builder;
 import org.apache.nifi.context.PropertyContext;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.flowfile.attributes.CoreAttributes;
 import org.apache.nifi.hadoop.SecurityUtil;
+import org.apache.nifi.processor.DataUnit;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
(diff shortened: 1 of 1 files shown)


joewitt:
@exceptionfactory The problem with doing that in the flow with logmessage is that you're putting the onus on the user to do that.  And the point is they might not know.  

@pvillard31 I think this makes sense but to my response to exceptionfactory the whole idea here is to convey information that might not be obvious to the end user.  This seems like it could be more thoughtfully done by skipping new properties altogether.  Is there a somewhat established "datasets on average going to Iceberg should be X bytes" big?  If there is something like that then just have the processor keep a bit of state about this and after a bit of time and set of files go by periodically kick out such guidance.

pvillard31:
From discussions with Iceberg experts, a good size is 10MB. Problem is that we've seen users using PutIceberg without any kind of Merge operation before (imagine ConsumeKafka (not record based) to PutIceberg) creating thousands of small (hundreds of bytes) files and associated snapshots. This has basically killed the performances and made the query engine on top of the Iceberg tables pretty much unusable. While I do agree that the approach with a specific property sets a precedent that is not great, I don't really see another good approach to make sure the flow designer is aware of this as soon as the designer tries to run the processor to make some tests. I definitely don't think we should go in the direction of adding the Merge capabilities into the processor itself (that is going to be a mess and extremely complicated). The rules engine would be great but you can't be sure that the rule would be set and enforced by users before someone starts designing a flow where Iceberg is a destination. The reason for this change in this case is really because creating so many snapshot files has basically killed the downstream system.

Another approach that I can think of: if the processor is processing a flow file that is below the agreed size, then we would yield the processor, and the default yield duration would be set to something like 1 minute. But we would still want to emit a bulletin to explain the behavior or something...

joewitt:
@pvillard31 I was meaning to say it is good to warn the user about usage not making sense with that processor.  But it does not need to be configurable what the size is - we already have too many dials/knobs.  Instead the processor should keep some simple state about the rate/avg size of files seen and if over say a 1-5 minute interval the data is on average 'too small' then log a warn/bulletin.  The bulletin could certainly advise the user that data to this processor should be batched together and that a component such as MergeContent or MergeRecord can be helpful.

pvillard31:
My concern with changing the routing approach is that it'd be a significant breaking change for existing flows. 

Having a logic that is not managed through properties also means that if a user does want small files (no idea why...) that would not be an option. But yeah I guess having a built-in rate limiter is probably the best option if we don't want to add properties.

joewitt:
@pvillard31 Can you please rebase and advise whether you are good removing the configurability of the size?

I think the warning is fair game but the configurable size just seems unnecessary.  I'm fine leaving it so the only thing really needing done is a rebase - but I'm asking if you'd consider dropping that.

Thanks

exceptionfactory:
Just following up on this, glad to take another look after the rebase, but I still disagree with adding a property to enable warning, since the warning is not really actionable.

pvillard31:
Hey @joewitt & @exceptionfactory - will close for now. I won't have the opportunity to get back to this before quite some time and I believe @mark-bathori may be able to come up with a PR that follows your recommendations.

PR #8639, opened 2024-04-12
NIFI-5519 Allow ListDatabaseTables to accept incoming connections and tweak GenerateTableFetch to work the same.

jrsteinebrey:
…nc up code in GenerateTableFetch to have its incoming connection work better like ListDatabaseTables.

# Summary

[NIFI-5519](https://issues.apache.org/jira/browse/NIFI-5519)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-5519) issue created

### Pull Request Tracking

- [X] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-5519`
- [X] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-5519`

### Pull Request Formatting

# Verification

I added new unit tests for incoming flow files for ListDatabaseTables.
I also tested ListDatabaseTables and GenerateTableFetch in the NiFi UI.

### Build

  - [X] JDK 21

### Licensing

No new dependencies

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ListDatabaseTables.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ListDatabaseTables.java
@@ -17,7 +17,6 @@
 package org.apache.nifi.processors.standard;
 
 import org.apache.nifi.annotation.behavior.InputRequirement;
-import org.apache.nifi.annotation.behavior.PrimaryNodeOnly;
 import org.apache.nifi.annotation.behavior.Stateful;
 import org.apache.nifi.annotation.behavior.TriggerSerially;
 import org.apache.nifi.annotation.behavior.WritesAttribute;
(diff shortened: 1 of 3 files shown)


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ListDatabaseTables.java
```diff
@@ -270,24 +297,40 @@ public Set<Relationship> getRelationships() {
...
+        final boolean isRefreshTracked = fileToProcess == null;
+        final long refreshInterval = !isRefreshTracked ? 0 : context.getProperty(REFRESH_INTERVAL).asTimePeriod(TimeUnit.MILLISECONDS);
 
         final StateMap stateMap;
         final Map<String, String> stateMapProperties;
         try {
-            stateMap = session.getState(Scope.CLUSTER);
-            stateMapProperties = new HashMap<>(stateMap.toMap());
+            stateMap = !isRefreshTracked ? null : session.getState(Scope.CLUSTER);
+            stateMapProperties = !isRefreshTracked ? null : new HashMap<>(stateMap.toMap());
```
dan-s1, line 333:
```suggestion
            stateMap = isRefreshTracked ? session.getState(Scope.CLUSTER) : null;
            stateMapProperties = isRefreshTracked ? new HashMap<>(stateMap.toMap()) : null;
```

jrsteinebrey: I make the change

joewitt:
These components are quite widely used so in review it is important to carefully consider whether this degree of change is desirable vs new components more specific to the use case(s) intended.

mattyb149:
I think that's why this Jira has sat around for so long, most ListXYZ processors are true source processors because they mostly deal with filesystems which are hierarchical and can be recursed for a single overall resource (top-level folder, etc.) Although databases tend to be less hierarchical, I think between the Category, Schema, and Table Name properties being able to match on wildcards, we can cover all tables for a single overall resource (the top-level database). The usual convention is to have multiple instances of the processor for each top-level resource (i.e. database connection).

Having said that, if we want to leverage LookupServices to allow a single instance to perform more diverse "work", we may want to look at another set of processors that better support swapping out controller services based on attribute values.

PR #8652, opened 2024-04-15
NIFI-13047: Adding property history to the property tooltip

mcgilman:
NIFI-13047:
- Adding property history to the property tooltip in the Edit dialogs for Processors, Controller Services, Reporting Tasks, Parameter Providers, and Flow Analysis Rules.

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/property-table/property-table.component.ts
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/property-table/property-table.component.ts
@@ -350,14 +350,6 @@ export class PropertyTable implements AfterViewInit, ControlValueAccessor {
         return 'property-' + item.id;
     }
 
-    hasInfo(descriptor: PropertyDescriptor): boolean {
-        return (
-            !this.nifiCommon.isBlank(descriptor.description) ||
-            !this.nifiCommon.isBlank(descriptor.defaultValue) ||
-            descriptor.supportsEl
-        );
-    }
-
     isSensitiveProperty(descriptor: PropertyDescriptor): boolean {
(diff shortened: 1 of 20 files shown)


nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/property-table/property-table.component.ts
```diff
@@ -396,7 +398,8 @@ export class PropertyTable implements AfterViewInit, ControlValueAccessor {
 
     getPropertyTipData(item: PropertyItem): PropertyTipInput {
         return {
-            descriptor: item.descriptor
+            descriptor: item.descriptor,
+            propertyHistory: this.propertyHistory?.propertyHistory[item.property]
```
rfellows, line 402: Not really an issue with this code exactly, but the closest place to make the comment contextual...

We only display the tooltip if the property `hasInfo`. In my case, I have a CS with a dynamic property and it fails that test so the history is inaccessible via tooltip. I think we might want to add a new boolean helper like:
```
    hasHistory(item: PropertyItem): boolean {
        return !!this.propertyHistory?.propertyHistory[item.property];
    }
```

then update the logic in the template to be something like:
```
                            <div>
                                @if (hasInfo(item.descriptor) || hasHistory(item)) {
                                    <div
                                        class="fa fa-question-circle primary-color"
                                        nifiTooltip
                                        [tooltipComponentType]="PropertyTip"
                                        [tooltipInputData]="getPropertyTipData(item)"
                                        [delayClose]="false"></div>
                                }
                            </div>

```

It looks like this will solve the same issue for all of the various Controller Settings configuration dialogs

rfellows (changes requested):
Original NiFi does not display property history for Registry Clients, however if the endpoint is called with a registry client component id, it does return results. We should add support for it in the new UI.

A few other things inline...

rfellows (approved):
Looks good, @mcgilman!

PR #8669, opened 2024-04-19
NIFI-13072: Fix MonitorActivity problems with cluster scope flow monitoring

takraj:
# Summary

[NIFI-13072](https://issues.apache.org/jira/browse/NIFI-13072)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/MonitorActivity.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/MonitorActivity.java
@@ -16,6 +16,20 @@
  */
 package org.apache.nifi.processors.standard;
 
+import static java.util.Collections.singletonMap;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
 import org.apache.nifi.annotation.behavior.SideEffectFree;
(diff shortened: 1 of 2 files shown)


turcsanyip:
Reviewing...

turcsanyip:
@takraj Could you please rebase you PR onto the latest main? The module structure has been refactored since you opened the PR.

I also noticed that MonitorActivity's logic has been changed a bit:
- original version: processor restart resets the state
- new version: inactive state preserved upon restart

I suggest introducing a new property: `Reset State on Restart` (or similar) with default value `true` in order to keep the original behaviour.

takraj:
@turcsanyip

I've made it optional, with the original behavior as default. Thanks for the suggestion. I also had to fix a bug, that prevented ths feature to work in case NiFi is not clustererd.

Could you please revisit?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/MonitorActivity.java
```diff
@@ -217,224 +230,221 @@ protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
...
             }
+        }
+    }
 
-            if (sendInactiveMarker && shouldThisNodeReport(isClusterScope, shouldReportOnlyOnPrimary, context)) {
-                lastInactiveMessage.set(System.currentTimeMillis());
+    private void onTriggerInactiveFlow(ProcessContext context, ProcessSession session, boolean isClusterScope, long lastActivity) {
+        final ComponentLog logger = getLogger();
+        final boolean shouldReportOnlyOnPrimary = shouldReportOnlyOnPrimary(isClusterScope, context);
+        final boolean shouldThisNodeReport = shouldThisNodeReport(isClusterScope, shouldReportOnlyOnPrimary, context);
```
turcsanyip, line 373: `shouldThisNodeReport()` could call `shouldReportOnlyOnPrimary()` internally because all the parameters are passed to that method too and `shouldReportOnlyOnPrimary` local variable is not used anywhere else.

Also in `onTriggerActiveFlow()`.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/MonitorActivity.java
```diff
@@ -217,224 +230,221 @@ protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
...
 
-                session.getProvenanceReporter().create(inactiveFlowFile);
-                session.transfer(inactiveFlowFile, REL_INACTIVE);
-                logger.info("Transferred {} to 'inactive'", new Object[]{inactiveFlowFile});
-            } else {
-                context.yield();    // no need to dominate CPU checking times; let other processors run for a bit.
+        if (!wasActive) {
+            if (shouldThisNodeReport) {
```
turcsanyip, line 389: `shouldThisNodeReport` local variable could be moved into `if (!wasActive) { ... }`.

turcsanyip (changes requested):
@takraj I tested the processor and found the following issues and improvement points:

1. If `Reporting Node = primary`, the flow is in inactive state and then it is activated on a non-primary node, then it takes for a while for the primary node to detect the change and to send the activity marker. The marker should be sent on the next execution.
2. If `Reporting Node = all`, the flow is in inactive state, a node is disconnected, the flow is activated on that node and then the node is reconnected, then all cluster members send the activity marker except the reconnected node. The reconnected node should send the marker too. Similar wrong scenario: both the connected cluster and the disconnected node is activated independently. Similar working scenario: the connected cluster is activated and then the inactive node reconnects. In this case there is an activity marker on the reconnected node.
3. `Copy Attributes` property cloud be required (similar to the other true/false properties of the processor).
4. `Reporting Node` property should depend on `Monitoring Scope = cluster`.

Please also find my comments on the code below.

turcsanyip (approved):
Thanks @takraj!

+1 LGTM

PR #8670, opened 2024-04-19
NIFI-13030 Adding endpoint for comparing versions of registered flows

simonbence:
# Summary

[NIFI-13030](https://issues.apache.org/jira/browse/NIFI-13030)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


pgyori (changes requested):
Thank you for this feature. I added some review comments, please consider them. Thanks!

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/NiFiServiceFacade.java
```diff
@@ -1498,6 +1498,18 @@ Set<DocumentedTypeDTO> getControllerServiceTypes(final String serviceType, final
...
+     * Returns the differences of version B from version A.
+     *
+     * @param registryId the ID of the registry
+     * @param bucketId the ID of the bucket
+     * @param flowId the ID of the flow
+     * @param versionA version of the baseline snapshot of the comparison
+     * @param versionB version of the compared snapshot
+     * @return the differences between the snapshots
+     */
+    FlowComparisonEntity getVersionDifference(String registryId, String bucketId, String flowId, String versionA, String versionB);
```
simonbence, line 1511: I might need to check if the new branching causes changes here

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/util/ClosedOpenInterval.java
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.web.util;
+
+/**
+ * This implementation includes the lower boundary bud does not include the higher boundary.
```
pgyori, line 20: Typo in the comment: "bud" = "but".

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/util/ClosedOpenInterval.java
```diff
@@ -0,0 +1,82 @@
...
+        if (otherIntervalLowerBoundary < 0) {
+            throw new IllegalArgumentException("Lower boundary cannot be negative");
+        }
+
+        if (otherIntervalHigherBoundary <= 0) {
+            // Note: as a design decision the implementation currently does not support comparison with unspecified higher boundary
+            throw new IllegalArgumentException("Higher boundary must be positive");
+        }
+
+        if (otherIntervalLowerBoundary > otherIntervalHigherBoundary) {
```
pgyori, line 60: `otherIntervalLowerBoundary >= otherIntervalHigherBoundary` needs to be used in order to check for 0-length intervals where lower==higher (currently this case is not covered).

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/util/Interval.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Relative position of the "other" interval compared to this.
+     *
+     * @param otherIntervalLowerBoundary Lower boundary of the compared interval.
+     * @param otherIntervalHigherBoundary Higher boundary of the compared interval.
+     *
+     * @return Returns the relative position of the "other" interval compared to this interval. For example: if the result
+     *         is BEFORE, read it as: the other interval ends BEFORE the actual (and there is no intersection between them).
+     */
+    RelativePosition getRelativePositionOf(final int otherIntervalLowerBoundary, final int otherIntervalHigherBoundary);
```
pgyori, line 62: I'll leave this thought here for consideration:
I believe it would make sense for this method to accept another Interval object as parameter instead of boundary integers. That way in the subclass we do not have to implement again the checks for the reations between the lower and higher boundary. The only extra check in the sublcass' method would be to not allow the other interval to have undefined higher boundary.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/util/ClosedOpenIntervalTest.java
```diff
@@ -0,0 +1,93 @@
...
+        Assertions.assertThrows(IllegalArgumentException.class, () -> new ClosedOpenInterval(-1, 3));
+    }
+
+    @Test
+    public void testNegativeHigherBoundary() {
+        Assertions.assertThrows(IllegalArgumentException.class, () -> new ClosedOpenInterval(0, -1));
+    }
+
+    @Test
+    public void testReorderedBoundaries() {
```
pgyori, line 40: I believe a more appropriate name would be `testSwitchedBoundaries`, just like the word "switched" is used in `testCompareWhenOtherBoundariesAreSwitched`.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/util/Interval.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @return Returns an interval instance with closed low and open high boundary.
+     */
+    static Interval getClosedOpenInterval(final int lowerBoundary, final int higherBoundary) {
```
pgyori, line 67: I do not agree with having a factory method in the generic interface: if further classes were to be added that implement the interface, then the interface would need to be modified each time to contain factory methods for each implementing type if we wanted to maintain consistency. I think this method could be removed and instead of calling `Interval.getClosedOpenInterval(offset, higherBoundary)` in PaginationHelper, `new ClosedOpenInterval(offset, higherBoundary)` constructor could be used.

pgyori (changes requested):
Thank you @simonbence for considering all my comments and making changes. Much appreciated!
I like the new enum items. Also, throughout the code it is now very clear what is the basis of interval comparison and what is being compared to it.
I went through the new modifications and added comments, please have a look at them.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/util/Interval.java
pgyori: I understand your thought that since there is a finite number of combinations of boundaries, the interface could have methods for instantiating each combination and would not need to expose which implementing class is used. However, this current implementation violates 3 of the SOLID principles, namely:
- Single Responsibility: the interface now not only defines the capabilities of an Interval, but is also responsible for creating instances.
- Open-Closed Principle: Adding new implementations other than ClosedOpenInterval requires modification of Interval (to add new factory methods or modify exiting one(s)).
- Dependency Inversion: currently the interface depends on the implementing class.

simonbence: I slightly disagree with this primarily because of the static method usage but I have no strong feelings about it. Please expect a commit soonish

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/FlowResource.java
```diff
@@ -2057,6 +2062,100 @@ public Response getDetails(
         return generateOkResponse(flowDetails).build();
     }
 
+    @GET
+    @Consumes(MediaType.WILDCARD)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("registries/{registry-id}/branches/{branch-id-a}/buckets/{bucket-id-a}/flows/{flow-id-a}/{version-a}/diff/branches/{branch-id-b}/buckets/{bucket-id-b}/flows/{flow-id-b}/{version-b}")
```
bbende, line 2068: Is the intended use case for this to right-click on a process group under version control and be able to choose something like `Compare With` and then choose another version to compare with? is there any other scenario where we want to compare two flow versions?

I'm asking because if it is limited to the first scenario I described, we may want to consider a more specific API to the PG, similar to `process-groups/{id}local-changes`, maybe `/process-groups/{id}/diff`. Then it would only need the parameters for the other flow version to compare to, which could be added to the path, or as query params.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/FlowResource.java
```diff
@@ -2104,6 +2203,24 @@ public Response getVersions(
...
+
+    private static ComponentDifferenceDTO limitDifferences(final ComponentDifferenceDTO original, final List<DifferenceDTO> partial) {
+        final ComponentDifferenceDTO result = new ComponentDifferenceDTO();
+        result.setComponentType(original.getComponentType());
+        result.setComponentId(original.getComponentId());
+        result.setComponentName(original.getComponentName());
+        result.setProcessGroupId(original.getProcessGroupId());
+        result.setDifferences(partial);
+        return result;
+    }
```
bbende, line 2222: Just want to ask whether we think we truly need paging here. Generally most of NiFi's REST APIs don't page anything, and all of the paging is done on the client side. The reason is because the slow part is not sending a large response from server to client, it is the rendering of the whole response on the client. So as long as the client is rendering pages, it usually works totally fine. 

If we do want paging, then shouldn't the entity being returned contain some paging info? At a minimum I would think we need to return the total rows in the result set, otherwise how can a client know when to page to?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/FlowResource.java
simonbence: The main use case is to provide a possibility to make comparisons even _before_ picking a version. As for the user's perspective this will result a similar listing as we have with `Local changes`. The intended place for this is the version list (for example when exeucting `Change Version` action on a versioned group as an addition.

The URL deliberately consists this many parameteres: the simplest situation is where we want to compers consecurtive versions of the same flow (in the same branch, etc.). A more complex, but still useful scenario is to compare flow snapshots between branches, like between a production and a development branch. As the branching strategy does not provide rescritions to follow the same structure for every branch, stricting this might make it impossible to make useful comparions with certain setups.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/FlowResource.java
simonbence: A significant difference from the `Local Changes` is that, we have a bigger chance to have more changed components, so I would expect more situation where we reach the limit of the first page (which is 1000, similar to the limitations of the `Local Changes`). 

As of this I do not find it unlikely that we want to support pagination in the long run. As for now, I aim to serve the first page only, but I wanted to provide an implementation which will be a good basis for furher changes if we want to allow pagination in the future.

Upon the start of this improvement, I reached out to @markap14 in order to ask his opinion about the possible support of pagination within the API (which would either result an extension of the current entity or adding a new one) and he voted against to touch the API for now.

The solution I provided looked to satisfy this expectation as well (next to the expectations from the original purpose) and provides enough option for later adjustments. With this, if based on the community's experience the pagination will be a more significant need, we can add it easily, even without NiFi REST API changes.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/FlowResource.java
bbende: So just to clarify, you are saying that there is some PG under version control, lets say it is at v1. The user can select Change Version and see a versions list which has (1, 2, 3), and they can somehow select 2 and 3 and compare those even though the current flow is at 1? or can they only compare 1 against 2 or 3?

simonbence: As for URL perspective, other then the registry id, it is expected to provide branch-bucket-flow-version information for both sides of the comparson.

As for the user interface, the idea is to have a button with every version, like `Compare To...` which leads to the comparison view, which should be similart to the `Local Changes`, expect additional controls to pick the information above. The left side of the comparison should be already filled based on the picked version.

PR #8671, opened 2024-04-19
NIFI-13065: Adding initial bend points for self looping connections and connections that collide with existing connections

mcgilman:
NIFI-13065:
- Adding initial bend points for self looping connections and connections that collide with existing connections.
- Merging two actions into one for opening the new connection dialog.


rfellows:
reviewing

rfellows (approved):
Looks great. Thanks @mcgilman

PR #8676, opened 2024-04-19
NIFI-13054: Restoring Property Editor popups that provide EL Function documentation and Parameter descriptions

mcgilman:
NIFI-13054:
- Restoring Property Editor popups that provide EL Function documentation and Parameter descriptions.


scottyaslan (approved):
LGTM

PR #8677, opened 2024-04-19
NIFI-12998 Restructured project/build to drive better enforcement/maintainability of framework vs extensions.  Makes it easier for people to build extensions outside our codebase as well with boms they can leverage just as we ourselves do.  Build for NiFi ends up a bit smaller building on great work that created nifi-shared-bundle layer for common dependencies

joewitt:
# Summary

[NIFI-12998](https://issues.apache.org/jira/browse/NIFI-12998)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


joewitt:
1. need to fix dependency regressions that now are vuln versions.  
2. need to fix errors in https://github.com/apache/nifi/actions/runs/8758685134/job/24040051585?pr=8677.  This stems from us using pre-built nars in our test images.  We should likely stop this.

Note to reviewer:
The PR is large but in major part this is true to directory creation/renames.

What used to be in nifi-nar-bundles is now in two directories.

All contents were moved from nifi-nar-bundles to nifi-extension-bundles

All contents in nifi-nar-bundles/nifi-framework-bundle were moved to the top level.  This was a key part of separating the build of the core framework from the build of all extensions.

There are other minor directory creations for new bom(s) but otherwise the structure is in tact.

joewitt:
'nifi-jetty-bundle' was also renamed to 'nifi-jetty-nar' to be consistent with all other nar naming

joewitt:
So far unable to understand why github ci failed the nifi-jetty-nar test.  Locally it works fine and the source that creates those nar's does not reference nifi-jetty-bundle or nifi-jetty-nar.  What that is old and should be eliminated as a binary in our source tree our current full built works, all tests pass including the ci integration tests, system tests, and nifi runs.  So not sure what the deal is there.  Will focus on the security related version bumps

joewitt:
Reduced build size another few MB.  Tightened up and cleaned boilerplate in poms.  Fixed a few lingering test issues related to nifi-jetty-bundle vs nifi-jetty-nar.  Some strange issue with nifi-mock in TestTailFile on Windows where processor var could be null.  

The system test suite has some issue i still need to run down after the latest round of pom cleanup.  Getting a 503 during LoadBalanceIT

joewitt:
Will rebase for latest commit that landed on main.

Will remove dependency:analyze-only from pom.  It can be activated on the command line during builds whenever people need.

Will restore usage of slf4j-simple where it appears safe to do so.  

Thanks!

joewitt:
Squashed all the commits so any further improvements can be seen more easily.

Fixed the runtime-manifest-generator output which was getting conflicting log provider implementations.

exceptionfactory (commented):
Thanks for keeping this pull request updated with recent changes @joewitt.

I have made some progress reviewing the substance of the changes. The root Maven configuration looks good, along with the general approach of introducing new Bill of Materials definitions, using them as project parents, and making dependencies explicit in a number of places.

Moving extension modules under the new `nifi-extension-bundles` parent at the root level is very helpful for differentiating between them and the `nifi-framework-bundle` at the same level.

In the course of spot checking, I found a couple dependency versions that should use properties instead of explicit versions:

1. `nifi-framework-kubernetes-nar` should use the `kotlin.version` property instead of declaring `1.9.23`
2. `nifi-web-api` should use the `prometheus.version` property instead of declaring `0.16.0`

That's a small list for so many changes, and the majority of the updates look good on the review thus far.

joewitt:
Sounds good.  I tried to honor those throughout so not sure why these weren't.   But updated.

joewitt:
the check failures seem spurious/unrelated.   meanwhile i am rebasing to get these latest commits from main and will push up in about 11 minutes after all checks out locally

exceptionfactory (approved):
Thanks for the extensive work on restructuring module hierarchy and clarifying the dependency relationships @joewitt. Moving the framework bundle to the root level and separating the extension bundles was a much-needed change. The latest builds and tests look good. +1 merging

PR #8678, opened 2024-04-20
NIFI-13074: Resetting the transform when navigating from the lineage to the event listing

mcgilman:
NIFI-13074:
- Resetting the transform when navigating from the lineage to the event listing.


scottyaslan:
Reviewing...

scottyaslan (approved):
Works great! Thanks @mcgilman !

PR #8680, opened 2024-04-22
NIFI-13078: Adding support to Enable and Disable through the context menu and operation control

mcgilman:
NIFI-13078:
- Adding support to Enable and Disable through the context menu and operation control.

### DIFF
--- a/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/flow/flow.reducer.ts
+++ b/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/flow/flow.reducer.ts
@@ -30,6 +30,10 @@ import {
     createProcessGroup,
     createProcessor,
     deleteComponentsSuccess,
+    disableComponent,
+    disableComponentSuccess,
+    enableComponent,
+    enableComponentSuccess,
     flowApiError,
     flowVersionBannerError,
     groupComponents,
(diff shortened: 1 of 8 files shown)


rfellows:
reviewing

nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/flow/flow.reducer.ts
```diff
@@ -400,20 +415,26 @@ export const flowReducer = createReducer(
...
         operationCollapsed
     })),
-    on(startComponentSuccess, stopComponentSuccess, (state, { response }) => {
-        return produce(state, (draftState) => {
-            const collection: any[] | null = getComponentCollection(draftState, response.type);
+    on(
+        startComponentSuccess,
+        stopComponentSuccess,
+        enableComponentSuccess,
+        disableComponentSuccess,
```
rfellows, line 422: We are missing a reducer for `enableProcessGroupSuccess` and `disableProcessGroupSuccess` that sets the flowState saving back to false.

If you enable/disable the current process group (right click on the canvas) then try to edit one of the components in that PG, the apply button is a spinner as it thinks it is still "saving"
Screenshot 2024-04-23 at 10 17 08 AM

However, if you were to enable/disable a child pg this isn't an issue as the child pg is reloaded and causes `loadChildProcessGroupSuccess` to fire which ultimately sets the saving state back to false.

rfellows (changes requested):

rfellows (approved):
Changes look good, thanks @mcgilman 👍

PR #8683, opened 2024-04-22
NIFI-12986 Remove depreciation warning from ProcessSession.commit

EndzeitBegins:
# Summary

Removes deprecation notice as discussed in issue 
[NIFI-12986](https://issues.apache.org/jira/browse/NIFI-12986).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/processor/ProcessSession.java
+++ b/nifi-api/src/main/java/org/apache/nifi/processor/ProcessSession.java
@@ -76,10 +76,10 @@ public interface ProcessSession {
      * <p>
      * As soon as the commit completes the session is again ready to be used.
      * <p>
-     * As of NiFi 1.14.0, an asynchronous API was introduced, see {@link #commitAsync()}, {@link #commitAsync(Runnable)}, and {@link #commitAsync(Runnable, Consumer)}.
-     * In most cases, users should prefer to use the asynchronous API. The synchronous API is less suited for execution in different runtimes, e.g. MiNiFi or Stateless NiFi,
-     * and may cause the dataflow in such runtimes to get stuck.
-     *
+     * See {@link #commitAsync()}, {@link #commitAsync(Runnable)}, and {@link #commitAsync(Runnable, Consumer)}. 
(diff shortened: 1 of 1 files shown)


nifi-api/src/main/java/org/apache/nifi/processor/ProcessSession.java
```diff
@@ -75,10 +75,11 @@ public interface ProcessSession {
      * All FlowFiles operated on within this session must be accounted for by transfer or removal or the commit will fail.
      * <p>
      * As soon as the commit completes the session is again ready to be used.
+     * <p>
+     * As of NiFi 1.14.0, an asynchronous API was introduced, see {@link #commitAsync()}, {@link #commitAsync(Runnable)}, and {@link #commitAsync(Runnable, Consumer)}.
+     * In most cases, users should prefer to use the asynchronous API. The synchronous API is less suited for execution in different runtimes, e.g. MiNiFi or Stateless NiFi,
+     * and may cause the dataflow in such runtimes to get stuck.
```
joewitt, line 81: @EndzeitBegins The 'nifi-api' is the most definitive/canonical API we have.  We do not want to be talking about how a given method would cause things to get stuck and if that were the case we should fix that.  This is 'the api' so it should just work.  The downside of commit is that it is synchronous and immediate.  The upside of asyncCommit is that it is asynchronous meaning the framework gets to choose the optimal time to make the commit happen.

I recommend the following phrasing be considered instead of this whole block.

See {@link #commitAsync()}, {@link #commitAsync(Runnable)}, and {@link #commitAsync(Runnable, Consumer)}. The asynchronous session commit methods are preferable to this commit call for most cases as it defers when the actual commit happens to the framework so that it can optimize when the commit call takes place.

EndzeitBegins: Thanks for the recommendation. I applied the change you suggested.

PR #8684, opened 2024-04-22
NIFI-13083: Parameter sensitivity issue when deleting and re-adding

mcgilman:
NIFI-13083:
- Fixing import preventing the dialog from warning the user that they must apply changes before deleting/adding the same parameter with different sensitivities.
- Allowing the sensitivity to change when deleting/adding the same parameter as long as it hasn't been saved yet.
- Adding the Parameters context menu item for Process Groups.


scottyaslan:
Reviewing...

scottyaslan (approved):
LGTM +1!

PR #8687, opened 2024-04-22
NIFI-13085: When viewing content, defaulting the Mode to 'Formatted'

mcgilman:
NIFI-13085:
- When viewing content, defaulting the Mode to 'Formatted'.


scottyaslan (approved):
Verified that the content viewer is opened with the `View as:` in
 'Formatted' mode. Thanks!

PR #8708, opened 2024-04-27
NIFI-13108 Update a variety of dependencies

joewitt:
# Summary

[NIFI-13108](https://issues.apache.org/jira/browse/NIFI-13108)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### UI Contributions

- [ ] NiFi is modernizing its UI. Any contributions that update the [current UI](https://github.com/apache/nifi/tree/main/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui) also need to be implemented in the [new UI](https://github.com/apache/nifi/tree/main/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi).  

### Licensing

### Documentation

### DIFF
--- a/nifi-assembly/pom.xml
+++ b/nifi-assembly/pom.xml
@@ -101,12 +101,12 @@ language governing permissions and limitations under the License. -->
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <version>${logback.version}</version>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-core</artifactId>
-            <version>${logback.version}</version>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
--- a/nifi-bom/pom.xml
+++ b/nifi-bom/pom.xml
@@ -149,6 +149,12 @@
         </dependencies>
     </dependencyManagement>
     <dependencies>
+        <!--Include by default as dependencies anything that will be in the root lib and
+        which we are ok with nifi components actually having a compile time dependency on.
+        For instance we will not list logback-classic or logback-core as we only want these
+        relied on within an execution environment - code should use the slf4j or related
+        bridging APIs and not a concrete logging implementation for example.  Similarly,
+        we do not include nifi-runtime and such here even though they're in the root lib -->
         <dependency>
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-api</artifactId>
@@ -160,14 +166,6 @@
         <dependency>
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-property-utils</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-core</artifactId>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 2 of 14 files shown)


exceptionfactory (commented):
Thanks for the incremental upgrades @joewitt.

None of the updates appear introduce notable changes, however, two attempts to run the integration tests resulted in failures. Unfortunately, there appears to be a logging configuration problem resulting in an extremely large number of debug log messages for the `PeerChannel` class. This may need to be addressed separately, but for now, it appears to be obscuring the failure. The integration tests pass on the current main branch, so unless we can confirm that the failures are intermittent on GitHub, it appears that one of these updates could have introduced an unexpected change.

joewitt:
@exceptionfactory Yeah that is an ultra bummer because i run the same integration tests locally and they work fine. I'll run them again though and see if the same log info comes out.  I noticed what you're talking about as well.  Will try to run it down

joewitt:
have noticed a good bit of failures on the windows build as well with something related to the otel stuff.

joewitt:
the wild log output is found in at least nifi-framework-core integration tests.  And it is due to

SLF4J(W): Class path contains multiple SLF4J providers.
SLF4J(W): Found provider [ch.qos.logback.classic.spi.LogbackServiceProvider@7c24b813]
SLF4J(W): Found provider [org.slf4j.simple.SimpleServiceProvider@2c35e847]
SLF4J(W): See https://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J(I): Actual provider is of type [ch.qos.logback.classic.spi.LogbackServiceProvider@7c24b813]

This stems from the fact that we have in our lib (and now in the nifi-bom) logback-classic/logback-core AND there is a test dependency in the root pom on slf4j-simple.   I dont know why we need/want the slf4j-simple so am looking into that.

joewitt:
Adjusted nifi-bom to remove logback-classic and logback-core from the dependency list even as provided as it means test runs have both slf4j-simple and logback-classic/core.  The nifi-bom now is listing things which we are ok with components having a dependency on wheras things like logback-classic/core are really meant as runtime dependencies anyway.

Might add these things back into the nifi-bom but mark them as 'runtime' instead of provided to better articulate this intent.

joewitt:
@exceptionfactory This changeset should do it.  Hilariously reducing the number of log impls changes the behavior of aws-processors test.  I had to change that in NIFI-12998 but now that it is back to one the assertions are different.  I didn't look into why but that is...funny.

The root of the adjustment here is we took logback-classic and logback-core out out of the dependency set (still in dep mgmt) of nifi-bom.  This makes sense given we dont want things actually having compile time deps on them anyway.  I did mark them as 'runtime' scope instead of provided.  Meanwhile the nifi-py4j-bridge DOES have a direct code based requirement for logback-classic (and thus core).  This seems undesirable but I didn't look into the full story there and instead just explained this and marked it as provided scope for that one module.

Log output is dramatically reduced now thanks to the default behavior of test scoped slf4j-simple.

Anyway this should be in much better shape.  The integration tests seem fairly flakey in multiple local runs but they are able to pass most of the time.  And now we should see why it fails instead of being buried in noisy logs.

joewitt:
grrr.  Need to clean these up...

❯ find . -type f | grep \\.jar | grep logback
./nar/framework/nifi-framework-nar-2.0.0-SNAPSHOT.nar-unpacked/NAR-INF/bundled-dependencies/logback-core-1.5.6.jar
./nar/extensions/nifi-server-nar-2.0.0-SNAPSHOT.nar-unpacked/NAR-INF/bundled-dependencies/logback-core-1.5.6.jar
./nar/extensions/nifi-framework-kubernetes-nar-2.0.0-SNAPSHOT.nar-unpacked/NAR-INF/bundled-dependencies/logback-core-1.5.6.jar

pvillard31:
@joewitt - it feels like some of the dep upgrades included in that PR could also apply to NiFi 1.26 - are you planning for a PR that would apply to 1.x and should I wait for this before kicking off the NiFi 1.26 RC?

joewitt:
The change-set for the main branch is ready to go and looks good.

In general I do not plan to make hygiene updates on the 1.x line at this point.  The dependency management there is substantially less certain than it is now on the 2/x main line post NIFI-12998 and various related JIRAs which reduced build size/etc...  But some of these should be safe-ish and I'll flag those and put up a different JIRA and PR for it.

joewitt:
@pvillard31 For the 1.x line i put together https://issues.apache.org/jira/browse/NIFI-13131.  This *should be* safe.  Full clean build and a basic runs suggests so.

exceptionfactory (approved):
Thanks for resolving the logging issues @joewitt, the latest set of updates looks good. +1 merging

PR #8710, opened 2024-04-29
NIFI-13111: Removing/Hiding unimplemented features in preparation for next milestone release

mcgilman:
NIFI-13111:
- Removing/Hiding unimplemented features in preparation for next milestone release.


scottyaslan (commented):
Works as advertised. However, I think you need to run prettier and I see that several lint warnings have been added with this PR.

Do you think we should add some TODO above the commented out code that points to a jira or something?

mcgilman:
Thanks for the review @scottyaslan! Let me fix the formatting issues. I can add some TODOs so they can be easily found. We don't yet have JIRAs for each individual item but they are already called out in the parent JIRA [1].

[1] - https://issues.apache.org/jira/browse/NIFI-12400

PR #8711, opened 2024-04-29
NIFI-13112: Fixing linking to non canvas items

mcgilman:
NIFI-13112:
- Fixing linking to non canvas items.
- Adding a tooltip to render search match details.


scottyaslan (approved):
LGTM! Tested go to processor, funnel, label, connection, PG, ports, and RPG. Works well with the browser back button. I also tested go to parameter and CS. Works well.

PR #8713, opened 2024-04-29
NIFI-13114: Always show primary node indicator when configured

mcgilman:
NIFI-13114:
- Always show primary node indicator when configured


rfellows:
Will review

mcgilman:
Just discussed this more with @rfellows offline. Considering we allow for configuring primary node only execution and support filtering for this in the Summary page, it makes more sense to always show the primary node icon if configured. Updating this PR to reflect this and do some additional clean up.

rfellows (approved):
LGTM 👍

PR #8719, opened 2024-04-30
NIFI-13110: Auto-selecting the relationship when the source processor only offers a single relationship

mcgilman:
NIFI-13110:
- Auto-selecting the relationship when the source processor only offers a single relationship.


scottyaslan (approved):
LGTM! Thanks @mcgilman!

PR #8722, opened 2024-05-01
[NIFI-12763] - Theme mat paginator - unset the background color so it blends in with the theme

rfellows:


scottyaslan:
Reviewing...

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/assets/styles/_app.scss
```diff
@@ -136,6 +136,10 @@
         flex-basis: 100%;
         flex: 1;
     }
+
+    .mat-mdc-paginator {
+        background-color: unset !important;
```
scottyaslan, line 141: I think we likely want to just set it to the mat-app-background-color:

```suggestion
        background-color: var(--mat-app-background-color) !important;
```

scottyaslan (changes requested):
Thanks @rfellows this is a good improvement! I left a minor comment below.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/assets/styles/_app.scss
rfellows: That assumes we always want to put the paginator on our app background. setting is as `unset` allows it to be used on any background.

scottyaslan: Fair enough. Let's roll.

PR #8724, opened 2024-05-01
NIFI-13119: When evaluating dependent Properties, the UI should identify when the Property value is a parameter reference and resolve the value accordingly

mcgilman:
NIFI-13119:
- When evaluating dependent Properties, the UI should identify when the Property value is a parameter reference and resolve the value accordingly.


rfellows:
reviewing

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/property-table/property-table.component.ts
```diff
@@ -215,18 +220,50 @@ export class PropertyTable implements AfterViewInit, ControlValueAccessor {
...
 
+        // if the dependent item is sensitive, in this case we are lenient and
+        // consider the dependency met
+        if (dependentItem.descriptor.sensitive) {
+            return true;
+        }
+
         // if the dependent item is visible, but does not require a specific
         // dependent value consider the dependency met
         if (this.nifiCommon.isEmpty(dependency.dependentValues)) {
```
rfellows, line 231: We need to handle the scenario where the dependentItem.value is null so the dependency is not shown. An example of this is the `AWSCredentialsProviderControllerService` and the `Assume Role ARN` property. If it is unset, the other properties below it should not be shown (like `Assume Role Session Name` among others).

rfellows (approved):
Nice work, @mcgilman. I'll merge after CI completes.

PR #8733, opened 2024-05-03
[NIFI-13134] - Support No Value Set in Edit Parameter Context. 

rfellows:
[NIFI-13134](https://issues.apache.org/jira/browse/NIFI-13134)

Also included in this PR:
  [NIFI-13135](https://issues.apache.org/jira/browse/NIFI-13135) - Allow parameter table to be sortable by parameter name

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/edit-parameter-dialog/edit-parameter-dialog.component.ts
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/edit-parameter-dialog/edit-parameter-dialog.component.ts
@@ -110,11 +110,14 @@ export class EditParameterDialog {
 
         this.editParameterForm = this.formBuilder.group({
             name: this.name,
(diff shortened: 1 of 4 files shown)


mcgilman:
Will review...

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/edit-parameter-dialog/edit-parameter-dialog.component.ts
```diff
@@ -163,8 +166,8 @@ export class EditParameterDialog {
         this.editParameter.next({
             parameter: {
                 name: this.editParameterForm.get('name')?.value,
-                value,
-                valueRemoved: value == '' && !empty,
+                value: value === '' && !empty ? null : value,
+                valueRemoved: value === '' && !empty,
```
mcgilman, line 170: When applying the Parameter Context update, this `valueRemoved` field is getting lost and doesn't not get included in the update request. As a result, it's not possible when the user attempts to clear a field it is not removed.

PR #8734, opened 2024-05-03
NIFI-13136: Allowing users to unset optional property values

mcgilman:
NIFI-13136:
- Allowing users to unset optional property values.
- Only selecting value and applying focus if it is not read only.


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/property-table/editors/nf-editor/nf-editor.component.ts
```diff
@@ -207,8 +218,18 @@ export class NfEditor implements OnDestroy {
...
+            const value = valueControl.value;
+            if (value === '') {
+                if (emptyStringChecked.value) {
+                    this.ok.next('');
+                } else {
+                    this.ok.next(null);
+                }
+            } else {
+                this.ok.next(value);
+            }
```
rfellows, line 232: Not sure where else to leave this comment, but we need to prevent `Enter` from submitting the okClicked action if the form is invalid (required value not entered)

rfellows (approved):
Looks great! Nice work, @mcgilman 👍

PR #8744, opened 2024-05-03
[NIFI-12802] - Update to use node 22

rfellows:
[NIFI-12802](https://issues.apache.org/jira/browse/NIFI-12802)


scottyaslan (approved):
LGTM

PR #8749, opened 2024-05-06
NIFI-13147: Remove include-new-ui profile 

mcgilman:
NIFI-13147:
- Remove include-new-ui profile and include it by default always.


joewitt:
reviewing

PR #8752, opened 2024-05-06
NIFI-13141: Adding a dragging flag to not reset positioning for labels and connection bends if the user is currently performing an action

mcgilman:
NIFI-13141:
- Adding a dragging flag to not reset positioning for labels and connection bends if the user is currently performing an action.


rfellows:
reviewing...

rfellows (approved):
Looks good. Nice fixes, @mcgilman 👍

PR #8760, opened 2024-05-07
NIFI-13164: Moving from jsdom to happy dom for test environment execution

mcgilman:
NIFI-13164:
- Moving from jsdom to happy dom for test environment execution.
- Removing test for the Documentation component that was failing because it was attempting to load an iframe with the nifi documentation that wasn't available.


scottyaslan (approved):
LGTM. I confirmed that the unit test output no longer displays any errors.

PR #8762, opened 2024-05-07
[NIFI-13162] horizontal and vertical canvas component alignment

scottyaslan:
Kapture 2024-05-07 at 10 28 35

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/service/canvas-context-menu.service.ts
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/service/canvas-context-menu.service.ts
@@ -47,7 +47,7 @@ import {
     requestRefreshRemoteProcessGroup,
     runOnce,
     stopVersionControlRequest,
-    terminateThreads
+    terminateThreads, updatePositions
 } from '../state/flow/flow.actions';
 import { ComponentType } from '../../../state/shared';
 import {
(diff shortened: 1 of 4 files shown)


joewitt:
Hello there.  It says review feedback but I dont see any.  It is essential to capture fredback openly.  Thanks

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/service/canvas-context-menu.service.ts
```diff
@@ -321,24 +326,193 @@ export class CanvasContextMenu implements ContextMenuDefinitionProvider {
...
+                                        !updates.has(connection.id) &&
+                                        connectionSelection.classed('selected') &&
+                                        this.canvasUtils.canModify(connectionSelection)
+                                    ) {
+                                        // this is a selected connection that hasn't been updated by the delta yet
+                                        if (
+                                            this.canvasUtils.getConnectionSourceComponentId(connection) === d.id ||
+                                            !this.canvasUtils.isSourceSelected(connection, selection)
+                                        ) {
+                                            // the connection is either outgoing or incoming when the source of the connection is not part of the selection
```
mcgilman, line 392: I'm struggling with this logic and the comment. What are we attempting to check here?

mcgilman (commented):
Thanks for the PR @scottyaslan! Some minor comments below and looking for a clarification about one part.

Also, what are your thoughts on updating this to also include connections (even if they aren't selected) if both the source and destination are selected? It seems like this would be helpful, but I might be overlooking a scenario.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/service/canvas-context-menu.service.ts
scottyaslan: Happy to revisit this UX but from legacy nifi if the user has a connection selected it is aligned. If not, it is not:

Kapture 2024-05-08 at 11 51 32

Except for self looping connections. Those are treated differently since they move with a component when it is moved so they are also moved in the align use case as well:

Kapture 2024-05-08 at 11 54 35

scottyaslan:
> Also, what are your thoughts on updating this to also include connections (even if they aren't selected) if both the source and destination are selected? It seems like this would be helpful, but I might be overlooking a scenario.

Happy to discuss this further. I simply copied most of this code over from legacy nifi and maintained the UX from there. I think it makes sense as is. If the user wants to align the connection they should included it in the selection.

mcgilman (approved):
Thanks for the updates @scottyaslan! Will merge once CI completes.

PR #8768, opened 2024-05-07
[NIFI-13165] hide resize handle is status history when insufficient h…

scottyaslan:
…istory
Screenshot 2024-05-07 at 2 35 03 PM
Screenshot 2024-05-07 at 2 35 18 PM

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/status-history/status-history.component.html
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/status-history/status-history.component.html
@@ -170,7 +170,7 @@
             <mat-dialog-actions align="start">
                 <div class="flex w-full">
                     <div class="flex-1">
-                        @if (instances.length > 0) {
+                        @if (instances.length === 0 || fieldDescriptors.length === 0) {
                             <div class="refresh-container flex items-center gap-x-2">
                                 <button mat-icon-button color="primary" (click)="refresh()">
                                     <i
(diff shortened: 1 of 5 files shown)


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/status-history/status-history.component.html
```diff
@@ -14,7 +14,13 @@
...
 ~  limitations under the License.
 -->
-<div resizable (resized)="resized()" [minHeight]="720" [minWidth]="887" class="flex flex-col status-history-container">
+<div
+    resizable
+    [resizeDisabled]="instances.length === 0 || fieldDescriptors.length === 0"
+    (resized)="resized()"
+    [minHeight]="720"
+    [minWidth]="887"
+    class="flex flex-col status-history-container">
```
rfellows, line 23: Not specific to your changes, but while you are in here... does it make sense to update the logic when deciding to show/hide the refresh/last loaded timestamp in the footer?
Currently it only bases its decision on the number of instances and ignores the number of fieldDescriptors so we end up with an odd dialog like so:
Screenshot 2024-05-08 at 10 35 06 AM

technically they can keep refreshing so it isn't terrible as it is. just not ideal. i suppose if they hit refresh and there happens to be data at that point, the dialog would just grown and look like the other.

what are your thoughts here, @scottyaslan?

scottyaslan: Sure, I can hide it in this case.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/status-history/status-history.component.html
```diff
@@ -170,7 +170,7 @@ <h2 mat-dialog-title>Status History</h2>
             <mat-dialog-actions align="start">
                 <div class="flex w-full">
                     <div class="flex-1">
-                        @if (instances.length > 0) {
+                        @if (instances.length === 0 || fieldDescriptors.length === 0) {
```
rfellows, line 173: this is inverted. should be 
```suggestion
                        @if (instances.length > 0 && fieldDescriptors.length > 0) {
```

rfellows (changes requested):

rfellows (approved):
LGTM 👍 
Will merge once CI completes

PR #8771, opened 2024-05-07
NIFI-13155: Adding revision checking in flow reducer

mcgilman:
NIFI-13155: 
- Handling newer revisions in flow reducer to ensure that the appropriate version of the component is saved in case responses are received out of order.

This addresses an issue that occurs for slow connections and there are very large flows. Under these circumstances, a user would issue a request to add or remove a component between when the canvas polling starts and completes. This can be easily simulated by adding a `delay` to the `pipe` following the API call and before `loadProcessGroupSuccess` is dispatched.

This PR additionally adds an override to skip the revision check when a nodes cluster connection state changes. This is important because the basis for component revisions changes when the node is connected to the cluster from when it's disconnected.


rfellows:
will review

PR #8774, opened 2024-05-07
[NIFI-13117] update label resize handle to match other resize handles…

scottyaslan:
… and replace comments indicator with icon

Kapture 2024-05-07 at 18 16 50


rfellows (approved):
Looks good. I like the updates. Thanks, @scottyaslan

PR #8778, opened 2024-05-08
[NIFI-13183] add drop shadow to connections to increase visibility on…

scottyaslan:
… top of colored labels in both light and dark mode


rfellows (changes requested):
I can tell what is causing this issue, but if the connection line is vertical (or nearly vertical) the arrows disappear and the line get narrower and the drop shadow goes away.

Screenshot 2024-05-08 at 3 21 30 PM

Screenshot 2024-05-08 at 3 21 39 PM

scottyaslan:
@rfellows thanks for the thorough review! Great catch. I found this STO that I think hints at the issue you are seeing. Applying the drop shadow via css seems to have fixed it.

rfellows (approved):
These changes are a good improvement over the old UI and look great in both light and dark modes 👍

PR #8780, opened 2024-05-08
NIFI-13176: Relaxing the scale that triggers when details for components on the canvas are not rendered

mcgilman:
NIFI-13176: 
- Relaxing the scale that triggers when details for components on the canvas are not rendered.


scottyaslan (approved):
LGTM! Thanks for this contribution @mcgilman I can confirm that the user can zoom out farther before the canvas details are hidden.

PR #8781, opened 2024-05-08
[NIFI-13158] - Fix: need to prevent/disable keyboard hotkeys when a dialog is open or when typing in the search bar.

rfellows:
[NIFI-13158](https://issues.apache.org/jira/browse/NIFI-13158)


mcgilman (approved):
Thanks @rfellows!

PR #8784, opened 2024-05-09
NIFI-13188 updated to latest dependency check plugin 9.1.0

joewitt:
# Summary

[NIFI-13188](https://issues.apache.org/jira/browse/NIFI-13188)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### UI Contributions

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-provenance-repository-bundle/nifi-persistent-provenance-repository/pom.xml
+++ b/nifi-extension-bundles/nifi-provenance-repository-bundle/nifi-persistent-provenance-repository/pom.xml
@@ -52,13 +52,16 @@
             <artifactId>nifi-repository-encryption</artifactId>
             <version>2.0.0-SNAPSHOT</version>
         </dependency>
-
         <dependency>
             <groupId>org.apache.lucene</groupId>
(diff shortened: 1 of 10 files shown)


joewitt:
I again created this PR on the back of another PR.  So this trivial change will need to be cherry-picked.

Instructions on how to use this new version are on the JIRA.

It will execute very slowly without an API key as the download is substantially throttled but it does appear to work.  With an API key it is much faster.

We should consider getting a key for the NiFi CI builds and will want to cache the results for several hours/full day.

exceptionfactory (commented):
Thanks for highlighting the slowness introduced in version 9 of the dependency check plugin @joewitt, that was part of the reason this has not yet been updated.

I noted the GitHub Action now takes 25 minutes, versus 5 minutes previously. Although that is notably slower, it is faster than the other workflows. I agree that we should pursue a caching strategy. It should be possible to coordinate storing an API Key with Apache Infra so that it can be accessed by GitHub workflows.

exceptionfactory (approved):

PR #8790, opened 2024-05-09
NIFI-13196: Added new isAutoTerminated(Relationship) method to Proces…

markap14:
…sContext, which simply delegates to Connectable.isAutoTerminated(Relationship)

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### UI Contributions

### Licensing

### Documentation

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/processor/ProcessContext.java
+++ b/nifi-api/src/main/java/org/apache/nifi/processor/ProcessContext.java
@@ -109,6 +109,13 @@ public interface ProcessContext extends PropertyContext, ClusterContext {
     Set<Relationship> getAvailableRelationships();
 
     /**
+     * Indicates whether or not the given relationship is configured to be auto-terminated
+     * @param relationship the relationship
+     * @return <code>true</code> if the given relationship is auto-terminated, <code>false</code> otherwise
+     */
+    boolean isAutoTerminated(Relationship relationship);
+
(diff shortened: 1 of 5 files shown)


joewitt:
@markap14 one has to wonder how on earth we did not have this mechanism long ago.  So if I'm writing a processor, such as a custom processor or otherwise, I can now ask "Hey if I make this potentially expensive output is it going to get thrown away anyway so I can just bypass doing so?"     Very nice!  

Once this lands makes sense to have follow-on JIRA(s) to take advantage of it.

markap14:
@joewitt yes correct, I was sure that we had this method already, then I found that it exists on Connectable but was not surfaced in the ProcessContext. I am almost sure that we have processors that could take advantage of it already, and will do a scan through, creating any Jiras as necessary.

markap14:
@joewitt just created a Jira for RouteText, for example. https://issues.apache.org/jira/browse/NIFI-13198
Will take some thinking through which processors can make use of this, but for several Route types processors, especially those with an 'unmatched' type of relationship it will make sense.

joewitt:
Thanks.  This makes sense and I'm a +1.  Will merge tomorrow pending any further inputs.  The change is specific, well documented, and the two JIRAs filed are good examples of how it can be applied well.

PR #8792, opened 2024-05-09
NIFI-13060: Using ellipsis and normal line height in the property table

mcgilman:
NIFI-13060:
- Using ellipsis and normal line height in the property table.


rfellows (approved):
LGTM 👍

PR #8809, opened 2024-05-10
NIFI-13184 - on ProcessGroup.start(), mark invalid processors to start once they become valid

greyp9:
# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### UI Contributions

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/groups/ProcessGroup.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/groups/ProcessGroup.java
@@ -18,7 +18,6 @@ package org.apache.nifi.groups;
 
 import org.apache.nifi.authorization.resource.ComponentAuthorizable;
 import org.apache.nifi.components.VersionedComponent;
-import org.apache.nifi.components.validation.ValidationStatus;
 import org.apache.nifi.connectable.Connectable;
 import org.apache.nifi.connectable.Connection;
 import org.apache.nifi.connectable.Funnel;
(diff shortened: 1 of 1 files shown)


joewitt:
@greyp9 Can you clarify the behavior you're seeing.  I have a system i can try to reproduce this on but I have some python components that take some time to load and they show up invalid.  But they did start on their own I thought. 

Are you saying right now such components even if set to auto start would on restart not launch?

greyp9:
[app1.log](https://github.com/apache/nifi/files/15277123/app1.log)

greyp9:
I've not tested the case where the processors are set to auto-start.  

In the case where a flow is imported, during the period of time where the dependencies are being resolved, the action "ProcessGroup / Start" excludes the processors with dependencies.  So they end up Stopped, even after the dependencies become available.

greyp9:
I suspect that, if set to auto-start in the flow definition, things would work as expected.

exceptionfactory:
@greyp9 Is this still something to be resolved, or does the change no longer appear to be required. If it isn't too involved, having a unit test associated with the change would be helpful to illustrate the expected behavior.

greyp9:
> @greyp9 Is this still something to be resolved, or does the change no longer appear to be required. If it isn't too involved, having a unit test associated with the change would be helpful to illustrate the expected behavior.

Yes, my read is that this is still an issue.  My understanding is that the current behavior differs from the behavior when a node rejoins a cluster.

When I did this research, I didn't immediately see a natural place to add a unit test.  Maybe you could point me at an existing test that I could use to model a new one?

markap14:
I'm not sure that there's anywhere that a unit test, per se, would be fruitful, but a system test might make a lot of sense. Add some components that are invalid but that will become valid, start the PG, and verify that they start running shortly after. But given the limited scope of changes here, and the fact that all existing tests still pass (and the fact that things appear to work when manually testing) i'm good with it as-is. Could always do a follow-on for system tests. You good with that @exceptionfactory ?

exceptionfactory (approved):
Thanks for the update @greyp9, and thanks for the additional perspective @markap14.

On further consideration, it sounds like an automated test would be a bit more challenging to implement, given that something would be required to make a Processor become valid.

With that background, I'm supportive of the change as it stands.

markap14:
Thanks @greyp9 and @exceptionfactory all good from my side. +1 merged to main!

PR #8811, opened 2024-05-10
[NIFI-13197] - Clicking outside of a dialog no longer closes the dialog

rfellows:
Also, escape will only close the dialog if the underlying form is not dirty.

[NIFI-13197](https://issues.apache.org/jira/browse/NIFI-13197)

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/close-on-escape-dialog/close-on-escape-dialog.component.ts
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/close-on-escape-dialog/close-on-escape-dialog.component.ts
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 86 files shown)


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/close-on-escape-dialog/close-on-escape-dialog.component.ts
```diff
@@ -0,0 +1,52 @@
...
+                .pipe(
+                    filter((event: KeyboardEvent) => event.key === 'Escape'),
+                    takeUntilDestroyed()
+                )
+                .subscribe(() => {
+                    if (!this.isDirty()) {
+                        this.dialogRef.close();
+                    }
+                });
+        }
```
scottyaslan, line 46: If the form is dirty and the user types 'Esc' the user has no feedback as to why the dialog will not close. I think we could use the snackbar to inform the user to either apply the change or 'Cancel' to discard the changes.

I also think the UX is a little inconsistent. The backdrop click should close the dialog when the form is not dirty same as the 'Esc' button. The following change will implement this:


```suggestion
        const store: Store<CanvasState> = inject(Store<any>);
        if (this.dialogRef) {
            this.dialogRef
                .keydownEvents()
                .pipe(
                    filter((event: KeyboardEvent) => event.key === 'Escape'),
                    takeUntilDestroyed()
                )
                .subscribe(() => {
                    if (!this.isDirty()) {
                        this.dialogRef.close();
                    } else {
                        store.dispatch(
                            ErrorActions.snackBarError({
                                error: 'You have unsaved changes. Please apply the change or click close to discard.'
                            })
                        );
                    }
                });

            this.dialogRef
                .backdropClick()
                .pipe(takeUntilDestroyed())
                .subscribe(() => {
                    if (!this.isDirty()) {
                        this.dialogRef.close();
                    } else {
                        store.dispatch(
                            ErrorActions.snackBarError({
                                error: "You have unsaved changes. Please apply the change or click 'Cancel' to discard."
                            })
                        );
                    }
                });
        }
```

Maybe we can improve upon the message in the snackbar. I am not thrilled with it as is...

scottyaslan: Let's just update the UX to match legacy nifi.

scottyaslan (approved):
LGTM

PR #8817, opened 2024-05-10
[NIFI-13127] - Add branching support to registry dialogs

rfellows:
[NIFI-13127](https://issues.apache.org/jira/browse/NIFI-13127)


bbende (approved):
Verified this from a functional perspective, works awesome! Able to select the branch on save/import for GitHub client that support branching, and verified NiFi Registry clients hiding the branch selection. Thanks for adding this!

exceptionfactory (approved):
Thanks for the improvement @rfellows, and thanks for the review @bbende, looks good! +1 merging

PR #8820, opened 2024-05-13
[NIFI-13220] - only allow buckets that users have write access to in the bucket dropdown when starting version control

rfellows:
[NIFI-13220](https://issues.apache.org/jira/browse/NIFI-13220)


exceptionfactory (approved):
Thanks for the improvement @rfellows! +1 merging

PR #8830, opened 2024-05-14
[NIFI-13229] - Include branch when submitting to save version control endpoint

rfellows:
[NIFI-13229](https://issues.apache.org/jira/browse/NIFI-13229)


bbende (approved):
Looks good and verified this fixes the issue, thanks!

PR #8833, opened 2024-05-15
[NIFI-13241] - extend CloseOnEscpaeDialog in dialogs previously missing it

rfellows:
[NIFI-13241](https://issues.apache.org/jira/browse/NIFI-13241)

There were a number of dialogs missed in the original pass at supporting escape to close them. This PR adds it to those dialogs.


scottyaslan:
reviewing...

scottyaslan (changes requested):
Thanks for the updates. I found a few more dialogs that present the issue:

- create port
- create PG
- create RPG

scottyaslan (approved):
LGTM will merge once CI completes

PR #8838, opened 2024-05-15
[NIFI-13243] - support cancel response in CloseOnEscapeDialog

rfellows:
[NIFI-13243](https://issues.apache.org/jira/browse/NIFI-13243)


scottyaslan (approved):
Works well! Thanks @rfellows !

PR #8842, opened 2024-05-16
NIFI-12670 JoltTransform processors incorrectly encode/decode text on Windows

jrsteinebrey:
NIFI-12670 JoltTransform processors incorrectly encode/decode text on the Jolt Specification on Windows

# Summary

[NIFI-12670](https://issues.apache.org/jira/browse/NIFI-12670)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-12670) issue created

### Pull Request Tracking

- [X] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-12670`
- [X] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-12670`

### Pull Request Formatting

# Verification

Add ran the processors in the UI and confirmed they worked correctly with linux and Windows encoded files.
I also added unit tests.

### Build

  - [X] JDK 21

### UI Contributions

No UI changes

### Licensing

New new dependencies

### Documentation

No documentation changes.

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/components/resource/ResourceReference.java
+++ b/nifi-api/src/main/java/org/apache/nifi/components/resource/ResourceReference.java
@@ -20,6 +20,7 @@ package org.apache.nifi.components.resource;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.net.URL;
 
 /**
@@ -43,6 +44,15 @@ public interface ResourceReference {
      * @throws IOException if unable to obtain an InputStream from the resource
      */
     InputStream read() throws IOException;
+
+    /**
+     * @return an InputStreamReader to read the contents of the resource
+     *
+     * @throws IOException if unable to obtain an InputStream from the resource
+     */
+    default InputStreamReader createReader() throws IOException {
+        return new InputStreamReader(this.read());
+    }
 
     /**
      * Indicates whether or not the resource is accessible. What it means for the resource to be accessible depends on the type of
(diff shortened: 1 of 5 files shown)


nifi-api/src/main/java/org/apache/nifi/components/resource/ResourceReference.java
```diff
@@ -44,6 +45,15 @@ public interface ResourceReference {
...
     InputStream read() throws IOException;
 
+    /**
+     * @return an InputStreamReader to read the contents of the resource
+     *
+     * @throws IOException if unable to obtain an InputStream from the resource
+     */
+    default InputStreamReader createReader() throws IOException {
+        return new InputStreamReader(this.read());
+    }
```
exceptionfactory, line 55: We should avoid introducing this default method because it defaults to the platform character encoding.

nifi-api/src/main/java/org/apache/nifi/components/resource/Utf8TextResource.java
```diff
@@ -46,6 +47,11 @@ public InputStream read() throws IOException {
         return new ByteArrayInputStream(text.getBytes(StandardCharsets.UTF_8));
     }
 
+    @Override
+    public InputStreamReader createReader() throws IOException {
+        return new InputStreamReader(this.read(), StandardCharsets.UTF_8);
+    }
```
exceptionfactory, line 53: Although this is convenient, it is best to avoid introducing it as an API-level method.

exceptionfactory (changes requested):
Thanks for proposing this correction and improvement @jrsteinebrey.

The changes in the `AbstractJoltTransform` class make sense, in terms of using UTF-8, however, we should avoid making changes to `nifi-api` interfaces. In particular, the createReader() method without a character set specified could easily lead to incorrect usage, falling back to platform-default character encoding.

For that reason, I recommend constructing the Reader with UTF-8 in the `AbstractJoltTransform` class as a way forward.

exceptionfactory (approved):
Thanks for adjusting the approach @jrsteinebrey, the latest version looks good. +1 merging

PR #8853, opened 2024-05-20
NIFI-13154 Display parameter reference when used as Sensitive property value

jrsteinebrey:
A parameter reference (containing a parameter name) is safe to display to users because the sensitive info is stored in the parameter value.

# Summary

[NIFI-13154](https://issues.apache.org/jira/browse/NIFI-13154)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-13154) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

I tested the current and new ui with different types of values for sensitive properties. When a parameter reference 
#{parameter} is in the sensitive property value then it is displayed to the user. If anything else if the sensitive property value, then the value is obfuscated from the user with text "Sensitive value set"

### Build

  - [X] JDK 21

### UI Contributions

Implemented in both the current and new UI.

- [X] NiFi is modernizing its UI. Any contributions that update the [current UI](https://github.com/apache/nifi/tree/main/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-ui) also need to be implemented in the [new UI](https://github.com/apache/nifi/tree/main/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi).  

### Licensing

No dependency changes.

### Documentation

No documentation changes.

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
+++ b/nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
@@ -34,7 +34,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
-import java.util.regex.Pattern;
 
 /**
  * An immutable object for holding information about a type of component
--- a/nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
+++ b/nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
@@ -720,15 +720,6 @@ public final class PropertyDescriptor implements Comparable<PropertyDescriptor> 
         return sensitive;
     }
 
-    // Pattern to match a parameter reference i.e. "#{anything}"
-    private static final Pattern PARAMETER_REFERENCE = Pattern.compile("^#\\{.*}$");
-
-    public static boolean isSensitiveValueSafeToDisplay(String value) {
-        // If the value is a parameter reference, then it is safe to display the parameter name.
-        // A parameter name is safe to display to users because the sensitive info is stored in the parameter value.
-        return value != null && PARAMETER_REFERENCE.matcher(value).matches();
-    }
-
     public boolean isDynamic() {
         return dynamic;
     }
--- a/nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
+++ b/nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
@@ -34,6 +34,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 /**
  * An immutable object for holding information about a type of component
@@ -717,6 +718,15 @@ public final class PropertyDescriptor implements Comparable<PropertyDescriptor> 
 
     public boolean isSensitive() {
         return sensitive;
+    }
+
+    // Pattern to match a parameter reference i.e. "#{anything}"
+    private static final Pattern PARAMETER_REFERENCE = Pattern.compile("^#\\{.*}$");
+
+    public static boolean isSensitiveValueSafeToDisplay(String value) {
+        // If the value is a parameter reference, then it is safe to display the parameter name.
+        // A parameter name is safe to display to users because the sensitive info is stored in the parameter value.
+        return value != null && PARAMETER_REFERENCE.matcher(value).matches();
     }
 
     public boolean isDynamic() {
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
@@ -4155,9 +4155,9 @@ public final class DtoFactory {
                // store the property descriptor
                dto.getDescriptors().put(descriptor.getName(), createPropertyDescriptorDto(descriptor, procNode.getProcessGroup().getIdentifier()));
 
-               // determine the property value - don't include sensitive properties
+               // determine the property value - don't include sensitive properties values (unless they are safe to display)
                String propertyValue = entry.getValue();
-               if (propertyValue != null && descriptor.isSensitive()) {
+               if (propertyValue != null && descriptor.isSensitive() && !PropertyDescriptor.isSensitiveValueSafeToDisplay(propertyValue)) {
                    propertyValue = SENSITIVE_VALUE_MASK;
                } else if (propertyValue == null && descriptor.getDefaultValue() != null) {
                    propertyValue = descriptor.getDefaultValue();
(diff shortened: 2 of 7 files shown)


nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
```diff
@@ -719,6 +720,15 @@ public boolean isSensitive() {
...
     }
 
+    // Pattern to match a parameter reference i.e. "#{anything}"
+    private static final Pattern PARAMETER_REFERENCE = Pattern.compile("^#\\{.*}$");
+
+    public static boolean isSensitiveValueSafeToDisplay(String value) {
+        // If the value is a parameter reference, then it is safe to display the parameter name.
+        // A parameter name is safe to display to users because the sensitive info is stored in the parameter value.
+        return value != null && PARAMETER_REFERENCE.matcher(value).matches();
+    }
```
exceptionfactory, line 730: This information should not be part of the `nifi-api` because the persistence strategy is specific to the framework implementation.

exceptionfactory (changes requested):
Thanks for working on this improvement @jrsteinebrey. The general concept makes sense, but the implementation needs to be handled carefully given the importance of protecting sensitive property values.

Of particular note, any methods to check for parameter references should not be placed in `nifi-api`, because that is not a public API concern. The persistence of sensitive values is a framework concern, meaning that it could be changed without impacting the public `nifi-api` contract.

This also raises a question about whether the frontend should have any part of the decision of what to display. Duplicating the pattern checking on the frontend seems problematic from a code synchronization point of view. It seems like it would make more sense to evaluate the compute the appropriate value on the server-side. Tagging @mcgilman for additional input on this particular question.

nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
jrsteinebrey: Or if you prefer to make the change to show ******* instead of Sensitive value set in both the new AND current UIs, then that is fine with me also.

mcgilman: I think the new UI should also be showing the same `Sensitive value set` messaging. I don't think moving away from that behavior was intended. I'd need to revisit it to be sure but I would assume we'd want to retain that behavior, when not referencing a Parameter that is.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
```diff
@@ -4116,7 +4117,16 @@ private List<ThreadDumpDTO> createThreadDumpDtos(final ProcessorNode procNode) {
        return threadDumps;
    }
 
-   /**
+    // Pattern to match a parameter reference i.e. "#{anything}"
+    private static final Pattern PARAMETER_REFERENCE = Pattern.compile("^#\\{.*}$");
```
joewitt, line 4121: This seems like a fairly dangerous direction as-is.  What we're trying to assess isn't simply a string pattern which tells us the value is safe.  What we want to know is if a user specifically set a given property to be a parameter reference and only a parameter reference.  If it was then we can show the parameter reference.  If it was not a parameter reference or not meant to be then we should not show it.

String comparison seems like a problematic path here.

It is extremely important a feature like this is done carefully.  We should be able to improve the user experience but we cannot do so at the expense of exposing anything unintentionally.

joewitt: It also seems like we are over rotating on the concept of 'safe to display'.  These responses would impact what is avialable through the REST api generally, no?

exceptionfactory, line 4121: This regular-expression-based approach should be replaced with parsing capabilities from the `nifi-parameter` module.

exceptionfactory (changes requested):
Thanks for removing the `nifi-api` changes @jrsteinebrey.

The `nifi-parameter` module has classes for parsing parameter references, and that is the general approach that should be used. That would avoid introducing low-level parameter format handling outside of the `nifi-parameter` module.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
jrsteinebrey: I understand and will change it to use nifi-parameter method(s). Thanks for pointing that out.

jrsteinebrey: Yes, this PR changes the results of one REST API call for all callers.
The code could add a query parameter which if omitted will work as before to always return obfuscated. If the new query param is specified , then it would use the new concept of returning parameter references in readable form instead of obfuscated. Would that be a good interface?

I can change to use nifi-parameter functions to validate it is meant to be a parameter. Earlier comments raises these questions:

1) Does a parameter reference need to match an existing parameter for us to show it as the sensitive property value?

2) Does the logged in user have to have permission to read the associated parameter for the associated parameter reference to be displayed as a sensitive property value?

nifi-api/src/main/java/org/apache/nifi/components/PropertyDescriptor.java
jrsteinebrey: I get your point. Thanks for the clarification that ***** in the dialog is not intended, @mcgilman.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
joewitt: Yeah these are a couple good questions that came to mind for me as well.  My opinion, and i'm curious what others would say, is that

1. Yes it should match an existing parameter for us to reflect that parameter name as the value in the case of a sensitive field whose value is a parameter.   I'm thinking we should be able to know they selected a parameter by more than simply matching the string but i'm not positive about that.
2. If the user in question does not have read access to the associated parameter then they should not get to know that parameter is in use for a sensitive field.

exceptionfactory:
@jrsteinebrey based on the feedback thus far, it seems like there are several substantive issues to consider for this to go forward. Now that the legacy Web UI has been removed, there are also some additional adjustments required after rebasing. Are you still planning on revisiting this soon? If not, it may be best to close the pull request and prepare a new one when ready.

jrsteinebrey:
@exceptionfactory Thanks for the suggestion. Given the raised issues and removal of the legacy UI, I am closing this PR and will prepare a new one when ready.

exceptionfactory:
Sounds good, thanks @jrsteinebrey!

PR #8859, opened 2024-05-21
NIFI-12967: Introducing Back action

mcgilman:
NIFI-12967:
- Adding support for navigating back to the users previous location in certain circumstances like going to a Controller Service, managing access policies, going to parameters, viewing documentation, opening advanced UIs, and viewing data provenance.
- Cleaning unnecessary instances of postUpdateNavigation.

The scenarios mentioned in the commit message above have a slightly different UX from the existing application. In the old UI, these actions would open in a large modal on top of the current component. When the user was done they could return to where they were by closing the modal. In the new UI, the user currently needs to use their browser back button as the application did not directly offer a back option. This PR introduces that back option but only in the cases identified above. Scenarios outside of this (other links in the application like Referencing Components) are not included here but could be considered in the future as a further improvement over the old UI.

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/app.component.ts
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/app.component.ts
@@ -24,7 +24,7 @@ import { MatDialog } from '@angular/material/dialog';
 import { NiFiState } from './state';
 import { Store } from '@ngrx/store';
 import { BackNavigation } from './state/navigation';
-import { popBackNavigation, pushBackNavigation } from './state/navigation/navigation.actions';
+import { popBackNavigation, pushBackNavigation, resetBackNavigation } from './state/navigation/navigation.actions';
 import { filter, map, tap } from 'rxjs';
 import { concatLatestFrom } from '@ngrx/operators';
 import { selectBackNavigation } from './state/navigation/navigation.selectors';
@@ -73,9 +73,10 @@ export class AppComponent {
                         })
                     );
                 } else if (previousBackNavigation) {
-                    if (!event.url.startsWith(previousBackNavigation.routeBoundary.join('/'))) {
-                        console.log('popping...');
+                    if (extras?.state?.['executeBackNavigation']) {
                         this.store.dispatch(popBackNavigation());
+                    } else if (!event.url.startsWith(previousBackNavigation.routeBoundary.join('/'))) {
+                        this.store.dispatch(resetBackNavigation());
                     }
                 }
             });
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/app.component.ts
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/app.component.ts
@@ -16,11 +16,18 @@
  */
 
 import { Component } from '@angular/core';
-import { GuardsCheckEnd, GuardsCheckStart, NavigationCancel, NavigationStart, Router } from '@angular/router';
+import { GuardsCheckEnd, GuardsCheckStart, NavigationCancel, NavigationStart, NavigationEnd, Router } from '@angular/router';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { Storage } from './service/storage.service';
 import { ThemingService } from './service/theming.service';
 import { MatDialog } from '@angular/material/dialog';
+import { NiFiState } from './state';
+import { Store } from '@ngrx/store';
+import { BackNavigation } from './state/navigation';
+import { popBackNavigation, pushBackNavigation } from './state/navigation/navigation.actions';
+import { filter, map, tap } from 'rxjs';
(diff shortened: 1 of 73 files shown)


rfellows:
reviewing...

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/app.component.ts
```diff
@@ -33,16 +40,40 @@ export class AppComponent {
...
+                    this.store.dispatch(
+                        pushBackNavigation({
+                            backNavigation
+                        })
+                    );
+                } else if (previousBackNavigation) {
+                    if (!event.url.startsWith(previousBackNavigation.routeBoundary.join('/'))) {
+                        console.log('popping...');
+                        this.store.dispatch(popBackNavigation());
+                    }
```
rfellows, line 74: This logic doesn't consider the fact that there might be multiple `backNavigation`'s in the store. If the user is more than one deep and navigates to another page (link from referencing components, main menu, ...) only the last 1 will be popped off and leaving the user with a potentially confusing back navigation option.
Here is an example:

Kapture 2024-05-22 at 13 35 37

If i had navigated to the referencing process group for the parameter context, the same thing would happen, but the back navigation would show over the canvas which is odd.
Screenshot 2024-05-22 at 1 38 00 PM


_Also, the `console.log` should go away_

rfellows (changes requested):

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/controller-services/controller-services.effects.ts
```diff
@@ -436,9 +514,24 @@ export class ControllerServicesEffects {
...
+                                        processGroupId,
+                                        'controller-services',
+                                        response.id,
+                                        'edit'
+                                    ],
+                                    routeBoundary: response.postUpdateNavigationBoundary,
+                                    context: 'Controller Service'
+                                } as BackNavigation
+                            }
+                        });
```
rfellows, line 534: This condition needs to fire regardless of `response.postUpdateNavigationBoundary` existing.
The scenario that demonstrates this is editing a CS then clicking a referencing component. when it asks you to save before navigating, choose Yes. The CS gets saved, but this condition fails and the user is not routed to the referencing component.

rfellows (approved):
This looks good. Thanks @mcgilman 👍

scottyaslan (changes requested):
I logged https://issues.apache.org/jira/browse/NIFI-13290 to address the browser history issues I mentioned below. I also found a slight bug that should be fixed here:

Kapture 2024-05-23 at 16 42 17

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
```diff
@@ -1392,14 +1478,7 @@ export class FlowEffects {
...
-                                        uri: request.uri,
-                                        type: request.type,
-                                        payload: updateProcessorRequest.payload,
-                                        errorStrategy: 'banner',
-                                        postUpdateNavigation: updateProcessorRequest.postUpdateNavigation
-                                    }
+                                    request: updateProcessorRequest
                                 })
                             );
                         });
```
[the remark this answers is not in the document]
scottyaslan: https://issues.apache.org/jira/browse/NIFI-13290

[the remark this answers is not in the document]
mcgilman: @scottyaslan Thanks for filing the JIRA. Agreed that is an issue with our dialog close handling when the user navigates forward/back with their browsers. The other scenario looks like a bug in the logic when a new back navigation is pushed that is introduced in this PR. I think simply avoid pushing a duplicate back navigation would address this case. Great idea. Let me prep another commit.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/state/navigation/navigation.reducer.ts
```diff
@@ -28,7 +28,14 @@ export const navigationReducer = createReducer(
     initialState,
     on(pushBackNavigation, (state, { backNavigation }) => {
         return produce(state, (draftState) => {
-            draftState.backNavigations.push(backNavigation);
+            if (draftState.backNavigations.length > 0) {
```
scottyaslan, line 31: Looks like there is still a bug in here:

Kapture 2024-05-23 at 21 59 20

scottyaslan (approved):
@mcgilman thank you for this contribution. It is really helpful to be able to quickly return to the component context. One potential improvement would be great if the router state was reapplied during back navigation after a full page refresh but I don't really see that as a common use case.

PR #8861, opened 2024-05-21
[NIFI-13277] - Make the main menu more compact in terms of spacing.

rfellows:
[NIFI-13277](https://issues.apache.org/jira/browse/NIFI-13277)

Before:
Screenshot 2024-05-21 at 1 09 45 PM

After:
Screenshot 2024-05-21 at 1 08 48 PM

With the changes, the menu (as of now) displays without the need for scrollbars until the displayable browser area gets below 570 pixels in height. Previously it needed more than 740px. I guess I'm running non-clustered... but in clustered mode there would be an extra menu item, so these numbers would be slightly bigger there.


mcgilman:
Will review...

mcgilman (approved):

PR #8863, opened 2024-05-22
NIFI-13268 Fixed missing grpc dependencies in GCP module

turcsanyip:
# Summary

[NIFI-13268](https://issues.apache.org/jira/browse/NIFI-13268)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### UI Contributions

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-services-api/pom.xml
+++ b/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-services-api/pom.xml
@@ -31,5 +31,23 @@
             <groupId>com.google.auth</groupId>
             <artifactId>google-auth-library-oauth2-http</artifactId>
         </dependency>
+        <!-- please do not remove grpc-* dependencies because they are used by GCP processors at runtime,
+             see also https://github.com/apache/nifi/pull/8491#issuecomment-1992542785 -->
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-context</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-util</artifactId>
+        </dependency>
     </dependencies>
 </project>


nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-services-api/pom.xml
```diff
@@ -31,5 +31,23 @@
             <groupId>com.google.auth</groupId>
             <artifactId>google-auth-library-oauth2-http</artifactId>
         </dependency>
+        <!-- please do not remove grpc-* dependencies because they are used by GCP processors at runtime,
+             see also https://github.com/apache/nifi/pull/8491#issuecomment-1992542785 -->
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-api</artifactId>
```
joewitt, line 38: scope for these four should be 'runtime' if they're not necessary for compile but needed for tests/execution and makes their necessity explicit.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-services-api/pom.xml
```diff
@@ -31,5 +31,23 @@
             <groupId>com.google.auth</groupId>
             <artifactId>google-auth-library-oauth2-http</artifactId>
         </dependency>
+        <!-- please do not remove grpc-* dependencies because they are used by GCP processors at runtime,
```
joewitt, line 34: can ditch this comment in favor of listing the intended scope with the dependencies

joewitt:
Unpacking these relationships more

org.apache.nifi:nifi-gcp-services-api
-> com.google.auth:google-auth-library-oauth2-http
 -> com.google.http-client:google-http-client
  -> io.grpc:grpc-context
   -> io.grpc:grpc-api
  -> io.opencensus:opencensus-api
   -> io.grpc:grpc-context

joewitt:
Ok i'm getting closer to following what is going on.  Let's keep the convo here instead of JIRA to be in one place.

So the controller service uses com.google.auth:google-auth-library-oauth2-http which has a bunch of dependencies.

You wanted to exclude all those in the nifi-gcp-nar so they're not duplicated.  You added two more that were not part of the dependency chain of com.google.auth:google-auth-library-oauth2-http such as 'grpc-core' and 'grpc-util' because you discovered that 'grpc-api' has a reflective dependency on these libraries (which is wild).  However, since it is loaded via reflection the classloader should be the child classloader and it seems to be should not be necessary to move them up to the top area.  

Did it fail for you otherwise or you chose to keep them together for consistency?  I ask because if you look in the nifi-gcp-nar dependency tree there are many other 'io.grpc' dependencies that we do allow to remain in that child nar.

This is a long winded way of saying I get why you're excluding all the items in the nifi-gcp-nar.  Though I wonder why not simply mark the com.google.auth:google-auth-library-oauth2-http dependency as provided which then should address any dependencies it has naturally.

To be clear we can just keep it as-is in the PR and move on.  My concern is just to ensure we do the minimal pom magic and let the dependency management do its work as best we can.  So before we throw in the towel and keep as-is i'd like to make sure we're there.

joewitt:
Ok I backed out your PR and all the exclusions then reviewed the resulting nar content.  

In short I now see why you took the path you did and I think it is the best course.  These dependencies in the nifi-gcp-services-api-nar are used *all over the place* in the child nifi-gcp-nar so either use excludes in the two cases and you can keep version references off or use provided overrides and you may end up having to then care about versions which could create new/different problems.

I am inclined to merge your PR as-is.

turcsanyip:
> since it is loaded via reflection the classloader should be the child classloader and it seems to be should not be necessary to move them up to the top area.

I believe if `grpc-api` is loaded by the parent class loader, it cannot load classes from the child class loader, even via reflection.

> why not simply mark the com.google.auth:google-auth-library-oauth2-http dependency as provided which then should address any dependencies it has naturally.

There are other google dependencies in the child nar with `compile` scope and transitively bringing the same jars as `google-auth-library-oauth2-http`.

joewitt:
"I believe if grpc-api is loaded by the parent class loader, it cannot load classes from the child class loader, even via reflection. But need to double check, maybe missed something."
IF the parent class loader was used to load a class which then itself depends on another class definition during that load you're correct.  IF though the secondary loading is done via reflection loadClass then it starts over again at the context class loader so it should be fine in theory.

I'm currently trying to understand how this was working before and what changed for us to now need this.  Looking at m2 i dont see any of these things but still reviewing

joewitt:
- Ah in nifi 2.0 m2 things worked.
- For m3 you updated in https://github.com/apache/nifi/pull/8491
- For m3 just after that I updated with the massive dep cleanup in https://github.com/apache/nifi/pull/8677 and I removed those nifi-gcp-services-api entries and should not have.
- This change restores those four.
I'm merging this.

PR #8870, opened 2024-05-24
NIFI-13289 Add tooltip to NewCanvas item

zeevo:
# Summary

[NIFI-13289](https://issues.apache.org/jira/browse/NIFI-13289)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### UI Contributions

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/ui/canvas/header/new-canvas-item/new-canvas-item.component.ts
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/ui/canvas/header/new-canvas-item/new-canvas-item.component.ts
@@ -25,12 +25,15 @@ import { selectDragging } from '../../../../state/flow/flow.selectors';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { ComponentType } from '../../../../../../state/shared';
 import { CanvasView } from '../../../../service/canvas-view.service';
+import { NifiTooltipDirective } from '../../../../../../ui/common/tooltips/nifi-tooltip.directive';
+import { TextTip } from '../../../../../../ui/common/tooltips/text-tip/text-tip.component';
+import { ConnectedPosition } from '@angular/cdk/overlay';
 
 @Component({
     selector: 'new-canvas-item',
     standalone: true,
     templateUrl: './new-canvas-item.component.html',
-    imports: [CdkDrag],
+    imports: [CdkDrag, NifiTooltipDirective],
     styleUrls: ['./new-canvas-item.component.scss']
 })
 export class NewCanvasItem {
@@ -38,6 +41,17 @@ export class NewCanvasItem {
     @Input() iconClass = '';
     @Input() iconHoverClass = '';
     @Input() disabled = false;
+    @Input() tooltip = '';
+
+    protected readonly TextTip = TextTip;
+
+    tooltipPosition: ConnectedPosition = {
+        originX: 'center',
+        originY: 'bottom',
+        overlayX: 'center',
+        overlayY: 'top',
+        offsetY: 8
+    };
 
     dragging = false;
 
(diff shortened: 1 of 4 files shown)


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/ui/canvas/header/new-canvas-item/new-canvas-item.component.ts
```diff
@@ -25,19 +25,33 @@ import { selectDragging } from '../../../../state/flow/flow.selectors';
...
     @Input() type!: ComponentType;
     @Input() iconClass = '';
     @Input() iconHoverClass = '';
     @Input() disabled = false;
+    @Input() tooltip = '';
+
+    protected readonly TextTip = TextTip;
+
+    tooltipPosition: ConnectedPosition = {
+        originX: 'center', // maybe start?
```
zeevo, line 49: I need to remove this comment

rfellows (changes requested):
@zeevo, Thanks for the contribution.

There is a minor issue with this implementation that I think we can resolve. The main problem is that the tooltip doesn't go away while dragging and tooltips are shown as you drag over other component types in the header.

Kapture 2024-05-28 at 13 05 13


One possible option could be to implement a new `@Input()` for `nifiTooltipDisbled` (borrowing from how angular material does it with `matTooltipDisabled`). When disabled, the overlay wouldn't get rendered.
Then, in the template that uses the nifiTooltip set it to be true while dragging. There is already a member variable for this.

You will most likely need to short circuit the `mouseEnter` method if disabled so it doesn't attach the overlay. Probably also need to consider if the overlay has already been attached and the mouse is still over the element. It should get detached in this case. You can probably track this in a new mouseMove method and consider the attached state of the overlay and if it is disabled or not.

zeevo:
@rfellows Thanks for the review and in-depth suggestions. It really helped.

My latest commit has the following new behaviors:
- Tooltip dismisses on drag.
- Tooltip reappears on mouse up if within a canvas item.

I decided to short circuit within `attach()` as it made things easier in my opinion. Let me know if you disagree with that implementation, and I'll try doing it within the `HostListener` event methods.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/tooltips/nifi-tooltip.directive.ts
```diff
@@ -58,11 +59,29 @@ export class NifiTooltipDirective<T> implements OnDestroy {
         }
     }
 
+    @HostListener('mousemove')
+    mouseMove() {
+        if (this.overlayRef?.hasAttached() && this.tooltipDisabled) {
+            this.overlayRef?.detach();
+        } else {
+            this.attach();
```
rfellows, line 67: I don't think this attach is needed. In fact it causes console errors...

Screenshot 2024-05-29 at 8 49 10 AM

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/tooltips/nifi-tooltip.directive.ts
```diff
@@ -58,11 +59,29 @@ export class NifiTooltipDirective<T> implements OnDestroy {
...
+        }
+    }
+
+    @HostListener('mouseup')
+    mouseup() {
+        this.attach();
```
rfellows, line 73: Nice touch but i think this should also be guarded. It currently can cause an error (similar to the above) if you mousedown outside of the target object and then mouseup over it. The tooltip will render based on the mouseenter and then try to attach another on the mouseup.
```suggestion
        if (!this.overlayRef?.hasAttached()) {
            this.attach();
        }
```

rfellows (changes requested):
@zeevo This is looking pretty good. Just a couple of things to clean up and then I think we can get it merged.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/ui/common/tooltips/nifi-tooltip.directive.ts
zeevo: Added guard

rfellows (approved):
This looks good 👍 . Thanks for the contribution, @zeevo.
I'll get this merged once CI completes.

PR #8871, opened 2024-05-24
NIFI-13290: Manually closing dialogs with proper result in Navigation lifecycle to ensure back/forward browser button works

mcgilman:
NIFI-13290:
- Manually closing dialogs with proper result in Navigation lifecycle to ensure back/forward browser button works correctly. This change also handles any scenario when a user routes away from a dialog via a link or go to action so we were able to remove the dialog close action form those places.


scottyaslan (approved):
Thanks for this improvement @mcgilman I have verified that routing away from a dialog via a browser back/forward navigation that the history is 'correctly' maintained as we expect!

PR #8872, opened 2024-05-24
NIFI-13286: Conditionally reloading the child PG or current PG when emptying all queues

mcgilman:
NIFI-13286:
- Conditionally reloading the child PG or current PG when emptying all queues based on the context of the request.


rfellows (approved):
Looks good. Thanks @mcgilman 👍

PR #8874, opened 2024-05-24
NIFI-13285: If there is no transform to restore for the current viewport, executing a zoom fit

mcgilman:
NIFI-13285:
- If there is no transform to restore for the current viewport, executing a zoom fit.


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/src/app/pages/flow-designer/service/canvas-view.service.ts
```diff
@@ -638,11 +638,11 @@ export class CanvasView {
             newScale = 1;
 
             // since the entire graph will fit on the canvas, offset origin appropriately
-            graphLeft -= 100;
-            graphTop -= 50;
+            graphLeft -= 313;
+            graphTop -= 25;
```
rfellows, line 642: This logic could probably be improved. If the canvas is just slightly bigger than the graph, the offset used here could push some content out of the viewable area.
We might consider offsetting by the difference between the canvas and graph size divided by 2 (essentially center in the canvas). If that is too much, we could only do it in the case where the difference in the sizes is smaller than the hardcoded offsets chose here.

rfellows (changes requested):
Just one minor improvement to consider.

rfellows (approved):
LGTM 👍 
Thanks @mcgilman

PR #8877, opened 2024-05-24
NIFI-13270: Persisting the select tab index

mcgilman:
NIFI-13270:
- Persisting the select tab index.


rfellows (approved):
Good work. This is a nice addition. 👍

PR #8888, opened 2024-05-29
NIFI-13309 Lookup compatible bundles even if previous flow was empty

ferencerdei:
# Summary

[NIFI-13309](https://issues.apache.org/jira/browse/NIFI-13309)

This change modifies the VersionedFlowSyncronizer to call mapCompatibleBundles even if the original flow was empty. 
An additional null check was added to the proposed dataFlow instead to prevent exception during NiFi startup.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### UI Contributions

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizer.java
```diff
@@ -257,6 +257,10 @@ private void mapCompatibleBundles(final DataFlow proposedFlow, final ExtensionMa
         final Set<String> missingComponentIds = proposedFlow.getMissingComponents();
         final VersionedDataflow dataflow = proposedFlow.getVersionedDataflow();
 
+        if (dataflow == null) {
```
simonbence, line 260: I propose the usage of the `isFlowEmpty`method here, which covers more possible variations for an empty `dataFlow`. I tested it via standard NiFi and it worked properly during initial startup.

simonbence (commented):
Thanks for your work! It looks good to me, I built and started NiFi and it is still capable to run without prior flow definition. I used standard NiFi for this, also I left a suggestion. Other than that it is a +1 from me.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizer.java
ferencerdei: changed it to use the recommended method

briansolo1985:
Thanks for submitting the change and incorporating the suggestions. +1 from my side

PR #8890, opened 2024-05-29
NIFI-13231 Added Private key authentication for GitHubFlowRegistryClient

maybevanshh:
# Summary

[NIFI-13231](https://issues.apache.org/jira/browse/NIFI-13231)
This PR adds function to Generate Jwt Token in GitHubFlowRegistryClient and PrivateKey Authentication support.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### UI Contributions

### Licensing

### Documentation


nifi-extension-bundles/nifi-github-bundle/nifi-github-extensions/src/main/java/org/apache/nifi/github/GitHubRepositoryClient.java
```diff
@@ -379,6 +392,41 @@ public GHContent deleteContent(final String filePath, final String commitMessage
...
+        Date now = new Date(nowMillis);
+        Date exp = new Date(expMillis);
+        PEMParser pemParser = new PEMParser(new StringReader(pemString));
+        Object object = pemParser.readObject();
+        pemParser.close();
+
+        if (object instanceof PEMKeyPair) {
+            PEMKeyPair keyPair = (PEMKeyPair) object;
+            RSAPrivateKey rsaPrivateKey = RSAPrivateKey.getInstance(keyPair.getPrivateKeyInfo().parsePrivateKey());
+            PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(rsaPrivateKey.getEncoded());
```
exceptionfactory, line 415: We should use the utility from the github-api library, even though it is limited to PKCS 8. This avoids the use of the Bouncy Castle library, and also avoids the custom methods for token generation.

exceptionfactory (changes requested):
Thanks for putting this together @maybevanshh.

I provided several recommendations on property naming and description.

Although there may be some value in supporting PKCS 1 private keys, the dependency on Bouncy Castle is not ideal from a general maintainability point of view.

I recommend changing the approach to use the utility from github-api library, which encapsulates the JWT token generation.

nifi-extension-bundles/nifi-github-bundle/nifi-github-extensions/src/main/java/org/apache/nifi/github/GitHubRepositoryClient.java
maybevanshh: I'll commit the changes you requested and will use the utility function for authentication.

maybevanshh: I removed the seperate function for jwt and used JWTTokenProvider for token generation.

nifi-extension-bundles/nifi-github-bundle/nifi-github-extensions/src/main/java/org/apache/nifi/github/GitHubRepositoryClient.java
```diff
@@ -19,7 +19,12 @@
 
 package org.apache.nifi.github;
 
+import io.jsonwebtoken.Jwts;
+import io.jsonwebtoken.SignatureAlgorithm;
 import org.apache.nifi.registry.flow.FlowRegistryException;
+import org.bouncycastle.asn1.pkcs.RSAPrivateKey;
+import org.bouncycastle.openssl.PEMKeyPair;
+import org.bouncycastle.openssl.PEMParser;
```
exceptionfactory, line 27: These imports can be removed.

nifi-extension-bundles/nifi-github-bundle/nifi-github-extensions/src/main/java/org/apache/nifi/github/GitHubRepositoryClient.java
```diff
@@ -74,6 +81,15 @@ private GitHubRepositoryClient(final Builder builder) throws IOException, FlowRe
         switch (authenticationType) {
             case PERSONAL_ACCESS_TOKEN -> gitHubBuilder.withOAuthToken(builder.personalAccessToken);
             case APP_INSTALLATION_TOKEN -> gitHubBuilder.withAppInstallationToken(builder.appInstallationToken);
+            case APP_INSTALLATION_ID_AND_PRIVATE_KEY -> {
+                try {
+                    JWTTokenProvider jwtTokenProvider = new JWTTokenProvider(builder().appId, builder().privateKey);
+                    String token = jwtTokenProvider.getEncodedAuthorization();
+                    gitHubBuilder.withJwtToken(token);
+                } catch (Exception e) {
+                    throw new FlowRegistryException(e.getMessage());
```
exceptionfactory, line 90: A custom message should be set, and the exception passed to the constructor to preserve the stack trace.
```suggestion
                    throw new FlowRegistryException("Failed to generate JWT from App ID and Private Key", e);
```

exceptionfactory (changes requested):
Thanks for making the changes @maybevanshh. Unused imports should be removed and the asterisk-based import should be updated to use explicit classes. I also noted a change for the exception thrown.

If you can make those adjustments, I can follow up soon with adjustments for reading the private key.

maybevanshh:
Sure
Making these changes

maybevanshh:
For some reason JwtTokenProvider throws Bad Credentials , can you please check into it?

exceptionfactory:
> For some reason JwtTokenProvider throws Bad Credentials , can you please check into it?

Thanks for the updates. It sounds like an error coming from the GitHub REST API. I will run some local tests as part of adding the key parsing and follow up.

exceptionfactory (commented):
@maybevanshh I pushed a commit that implements support for reading PKCS1 RSA Private Keys, noting the the [GitHub documentation](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/managing-private-keys-for-github-apps#generating-private-keys) generates private keys in that format.

[JDK-8023980](https://bugs.openjdk.org/browse/JDK-8023980) Added support for parsing PKCS1 RSA Private Keys in Java 17, so the `StandardPrivateKeyReader` implementation makes use of that feature without the need for Bouncy Castle libraries.

In addition to that change, I initially observed the same `Bad Credentials` error you reported. That error is the result of using the App Token instead of the App Installation Token. The `getAppInstallationAuthorizationProvider` method constructs the required components, but in essence, the initial GitHub Authorization Provider authenticates, gets a GitHub Installation, and then the `AppInstallationAuthorizationProvider` is able to get an App Installation Token that can be used for the rest of the operations. Using these classes from the `github-api` library handles the token generation and refresh handling.

With these changes, the feature should be ready to go pending successful automated builds.

exceptionfactory (approved):
Thanks again for getting this started @maybevanshh! +1 merging

maybevanshh:
Hey @exceptionfactory Thanks a lot for resolving this and teaching me the workaround! Your solution was really helpful and I appreciate the time you took to explain it. I'm looking forward to contributing more to NiFi.

PR #8891, opened 2024-05-29
NIFI-13284: Only saving canvas routes that are not edit or history

mcgilman:
NIFI-13284:
- Only saving canvas routes that are not edit or history.


rfellows:
Reviewing...

rfellows (approved):
LGTM 👍

PR #8892, opened 2024-05-29
NIFI-13311: Turn DB adapters into controller services

takraj:
# Summary

[NIFI-13311](https://issues.apache.org/jira/browse/NIFI-13311)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### UI Contributions

### Licensing

### Documentation


nifi-assembly/pom.xml
```diff
@@ -350,6 +350,42 @@ language governing permissions and limitations under the License. -->
             <version>2.0.0-SNAPSHOT</version>
             <type>nar</type>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-generic-db-adapter-provider-nar</artifactId>
+            <version>2.0.0-SNAPSHOT</version>
+            <type>nar</type>
```
joewitt, line 357: Can you share more about the benefit of each of these being their own nar?  If i understand each nar correctly it does not contain provider specific things like mssql/etc..  So it seems like they can all be together.  If something had bespoke requirements splitting them is good.

Can you share more on what you had in mind?

nifi-extension-bundles/nifi-extension-utils/nifi-database-adapters/nifi-database-adapter-api/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+        <artifactId>nifi-database-adapters</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>nifi-database-adapter-api</artifactId>
+    <packaging>jar</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-utils</artifactId>
```
joewitt, line 28: Can we dump this since commons-lang is there?   it seems nifi-utils is being used for a simple string utils class

nifi-assembly/pom.xml
```diff
@@ -350,6 +350,42 @@ language governing permissions and limitations under the License. -->
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-oracle-db-adapter-provider-nar</artifactId>
```
joewitt, line 361: Is the word 'provider' helpful in these?  

It seems to me we have database adapter api and database adapters that implement the api.  Ie we have

nifi-database-adapter-api is the api module.

nifi-oracle-db-adapter-nar implements the api specific to oracle column/types.

Provider seems not needed.  Can you share more if that is helpful?

nifi-assembly/pom.xml
takraj: I was not sure about what granularity would be accaptable. I'm fine with combining them all together in a single controller service. The user could then select the needed dialect from a dropdown.

nifi-extension-bundles/nifi-extension-utils/nifi-database-adapters/nifi-database-adapter-api/pom.xml
takraj: OK I'll check if I can remove this dependency. I used DBCP CS as a template, while creating these.

nifi-assembly/pom.xml
takraj: I'm fine with removing the word 'provider' from the controller services, but I'm not sure if it's a good practice, since we have actual 'adapter' classes, that are instantiated by the 'provider' classes. If you feel like you have better wording for this concept, then I'm OK with renaming.

Or did you mean this 'adapter' - 'provider' abstraction does not seem useful? IMHO it adds some flexibility to how we organize, or combine various adapters, because for example a single controller service for MSSQL can provide adapters for multiple versions of that dialect.

nifi-assembly/pom.xml
takraj: @exceptionfactory Sounds reasonable. I'll try to apply this concept to my PR soon.

nifi-assembly/pom.xml
joewitt: I dont think I realized we had a 'database-adapter-api' and a 'database-adapter-provider-api'. 

The word 'adapter' is generic/broad as is 'provider' so it just looks odd having them both side by side.

I dont know enough about why we need these two separate APIs so I guess I'll back out of this comment for now.

takraj:
Hey @joewitt & @exceptionfactory . I've tried to rework my PR based on your comments. Please revisit.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/DatabaseParameterProvider.java
```diff
@@ -38,37 +26,34 @@
...
-                .name("db-type")
-                .displayName("Database Type")
-                .description("The type/flavor of database, used for generating database-specific code. In many cases the Generic type "
-                        + "should suffice, but some databases (such as Oracle) require custom SQL clauses. ")
-                .allowableValues(dbAdapterValues.toArray(new AllowableValue[dbAdapterValues.size()]))
-                .defaultValue("Generic")
-                .required(true)
-                .build();
-    }
+    // TODO: There was no migration facility for Parameter Providers by the time of introducing this property,
```
joewitt, line 48: better to have todo not in the code and instead just file the JIRA for this topic if necessary

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/DatabaseParameterProvider.java
```diff
@@ -38,37 +26,34 @@
...
+    //  and removing the old 'db-type'.
+    public static final PropertyDescriptor DATABASE_ADAPTER_PROVIDER = new Builder()
```
joewitt, line 50: Yeah so i'm back to my provider vs not provider questions here then.  So here we have a processor property which references a controller service.  That will be a listing of controller services that implement the 'DatabaseAdapter' api.  Then we should just call this the 'database adapter' and the user selects a particular concrete impl.

We need to have less terms.  It is also somewhat confusing given the todo references a parameter provider and here we're talking about a database adapter provider.  I really don't think we need so many words to describe these concepts.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/DatabaseParameterProvider.java
```diff
@@ -38,37 +26,34 @@
...
+            .name("db-adapter-provider")
+            .displayName("Database Adapter Provider")
```
joewitt, line 52: we dont tend to want these both populated where one is a computer type name and the other is the human readable form.  @exceptionfactory what is the guidance now?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/DatabaseParameterProvider.java
```diff
@@ -233,7 +218,7 @@ private void validateValueNotNull(final String value, final String columnName) {
     }
 
     String getQuery(final ConfigurationContext context, final String tableName, final List<String> columns, final String whereClause) {
-        final DatabaseAdapter dbAdapter = dbAdapters.get(context.getProperty(DB_TYPE).getValue());
+        final DatabaseAdapter dbAdapter = context.getProperty(DATABASE_ADAPTER_PROVIDER).asControllerService(DatabaseAdapter.class);
```
joewitt, line 221: perfect example where we do not need the word adapter anywhere.

DatabaseAdapter.class is an interface.  There are concrete impls.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/DatabaseParameterProvider.java
exceptionfactory: Any new properties should use `name()` only and should use a human-readable label for the property. So in this case, it should be `name("Database Adapter Provider")` and the `displayName()` can be removed.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/DatabaseParameterProvider.java
takraj: Alright, I've filed a jira for this: https://issues.apache.org/jira/browse/NIFI-13327

takraj: What would be the good strategy, given that there is no way to apply migration on Parameter Providers?

1. Have the fixed list of Db adapters in the {{DatabaseParameterProvider}}?
2. Replace the fixed list with the new controller service concept, and require a manual migration step from the user after upgrade?
3. Block this commit until the issue is resolved?

takraj: @joewitt , @exceptionfactory 

Do you have any thoughts regarding to this?

takraj: @joewitt , @exceptionfactory 

I have a draft PR that attempts to add the missing, property migration feature to the Parameter Provider component. Currently it is only half-complete, becuase I can't really find where is the proper place to call the `migrateProperties()` method. The place where I've now inserted it, seems to not get called. I assume it is because Parameter Providers do not belong to Process Groups.

Could you please have a look?

[Pull Request #9080](https://github.com/apache/nifi/pull/9080)

exceptionfactory (changes requested):
@takraj Thanks for your patience on this pull request. I have not had the opportunity to review all of the details, but I noted a large number of import reordering changes, which distract from the substantive changes.

If you could rebase this pull reques to resolve conflicts, and revert all of the import order changes, that would help streamline the change set for review.

exceptionfactory:
@takraj Given the number of conflicts and comments outstanding, I am going to close this for now. If you can address the current merge conflicts and address the comments, feel free to reopen when ready.

PR #8896, opened 2024-05-29
NIFI-13265 Removed the instantiation of Object arrays for arguments in ComponentLog log and org.slf4j.Logger statements.

dan-s1:
# Summary

[NIFI-13265](https://issues.apache.org/jira/browse/NIFI-13265)
In addition to removing the unnecessary instantiation of Object arrays in logging statements this PR also aimed to remove parameters meant for exceptions which had for example `due to...` syntax as this was an artifact of legacy behavior that did not properly handle Throwable arguments. In addition this PR fixed some of the logging arguments as well e.g. not calling toString on an argument.
If possible, please port these changes to both the 1.x and 2.x branches. 

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### UI Contributions

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ReplaceTextWithMapping.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ReplaceTextWithMapping.java
@@ -207,7 +207,7 @@ public class ReplaceTextWithMapping extends AbstractProcessor {
 
             flowFile = session.write(flowFile, new ReplaceTextCallback(context, flowFile, maxBufferSize));
 
-            logger.info("Transferred {} to 'success'", new Object[]{flowFile});
+            logger.info("Transferred {} to 'success'", flowFile);
             session.getProvenanceReporter().modifyContent(flowFile, stopWatch.getElapsed(TimeUnit.MILLISECONDS));
             session.transfer(flowFile, REL_SUCCESS);
         }
@@ -254,21 +254,21 @@ public class ReplaceTextWithMapping extends AbstractProcessor {
                         if (file.lastModified() > lastModified.get()) {
                             lastModified.getAndSet(file.lastModified());
                             try (FileInputStream is = new FileInputStream(file)) {
-                                logger.info("Reloading mapping file: {}", new Object[]{fileName});
+                                logger.info("Reloading mapping file: {}", fileName);
 
                                 final Map<String, String> mapping = loadMappingFile(is);
                                 final ConfigurationState newState = new ConfigurationState(mapping);
                                 configurationStateRef.set(newState);
                             } catch (IOException e) {
-                                logger.error("Error reading mapping file: {}", new Object[]{e.getMessage()});
+                                logger.error("Error reading mapping file: ", e);
                             }
                         }
                     } else {
-                        logger.error("Mapping file does not exist or is not readable: {}", new Object[]{fileName});
+                        logger.error("Mapping file does not exist or is not readable: {}", fileName);
                     }
                 }
             } catch (Exception e) {
-                logger.error("Error loading mapping file: {}", new Object[]{e.getMessage()});
+                logger.error("Error loading mapping file: ", e);
             } finally {
                 processorLock.unlock();
             }
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ReplaceTextWithMapping.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ReplaceTextWithMapping.java
@@ -260,7 +260,7 @@ public class ReplaceTextWithMapping extends AbstractProcessor {
                                 final ConfigurationState newState = new ConfigurationState(mapping);
                                 configurationStateRef.set(newState);
                             } catch (IOException e) {
-                                logger.error("Error reading mapping file: ", e);
+                                logger.error("Error reading mapping file", e);
                             }
                         }
                     } else {
@@ -268,7 +268,7 @@ public class ReplaceTextWithMapping extends AbstractProcessor {
                     }
                 }
             } catch (Exception e) {
-                logger.error("Error loading mapping file: ", e);
+                logger.error("Error loading mapping file", e);
             } finally {
                 processorLock.unlock();
             }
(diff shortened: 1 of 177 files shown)


dan-s1:
@exceptionfactory Can you please advise on is there something I checked in that caused the following exceptions
```
Error:    TestExecuteProcess.testNotRedirectErrorStream:241 If redirect error stream is false, the output should be logged as a warning so that user can notice on bulletin. ==> expected: <1> but was: <0>
Error:    TestExecuteProcess.testRedirectErrorStream:268 expected: <1> but was: <0>
```

Those tests on my computer are working and they are working on the MacOS. What more can I do for this?

exceptionfactory:
> @exceptionfactory Can you please advise on iif s there something I checked in that caused the following exceptions
> 
> ```
> Error:    TestExecuteProcess.testNotRedirectErrorStream:241 If redirect error stream is false, the output should be logged as a warning so that user can notice on bulletin. ==> expected: <1> but was: <0>
> Error:    TestExecuteProcess.testRedirectErrorStream:268 expected: <1> but was: <0>
> ```
> 
> Those tests on my computer are working and they are working on the MacOS. What more can I do for this?

@dan-s1 The `TestExecuteProcess` class has some design issues, but it appears that this particular issue could be related to the `isCommandFailed()` method in `TestExecuteProcess`. That method looks for an error message containing the words `due to` on line 308. Removing the `due to` portion of the string should align with the changes you have implemented.

dan-s1:
@exceptionfactory Thanks for the help on that! Looking at the javadoc on the method `isCommandFailed` 
```
/**
     * On some environment, the test command immediately fail with an IOException
     * because of the native UnixProcess.init method implementation difference.
     *
     * @return true, if the command fails
     */
```

it would seems to indicate that the JUnit 5 `@EnabledOnOs` annotation should be used to run those unit tests this method is used in to only run on a Unix environment.

exceptionfactory:
> @exceptionfactory Thanks for the help on that! Looking at the javadoc on the method `isCommandFailed`
> 
> ```
> /**
>      * On some environment, the test command immediately fail with an IOException
>      * because of the native UnixProcess.init method implementation difference.
>      *
>      * @return true, if the command fails
>      */
> ```
> 
> it would seems to indicate that the JUnit 5 `@EnabledOnOs` annotation should be used to run those unit tests this method is used in to only run on a Unix environment.

Yes, that is just one of several issues with the test class as it stands. There are System.out.println() statements that should be removed, System properties that should be removed, and other issues. It would be worth rewriting from scratch, with the goal of modernizing the implementation and providing better coverage of some scenarios. Not something for this Jira issue, but something to consider separately.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ReplaceTextWithMapping.java
```diff
@@ -254,21 +254,21 @@ private void updateMapping(final ProcessContext context) {
...
                             try (FileInputStream is = new FileInputStream(file)) {
-                                logger.info("Reloading mapping file: {}", new Object[]{fileName});
+                                logger.info("Reloading mapping file: {}", fileName);
 
                                 final Map<String, String> mapping = loadMappingFile(is);
                                 final ConfigurationState newState = new ConfigurationState(mapping);
                                 configurationStateRef.set(newState);
                             } catch (IOException e) {
-                                logger.error("Error reading mapping file: {}", new Object[]{e.getMessage()});
+                                logger.error("Error reading mapping file: ", e);
```
exceptionfactory, line 263:
```suggestion
                                logger.error("Error reading mapping file", e);
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/ReplaceTextWithMapping.java
```diff
@@ -254,21 +254,21 @@ private void updateMapping(final ProcessContext context) {
...
                             }
                         }
                     } else {
-                        logger.error("Mapping file does not exist or is not readable: {}", new Object[]{fileName});
+                        logger.error("Mapping file does not exist or is not readable: {}", fileName);
                     }
                 }
             } catch (Exception e) {
-                logger.error("Error loading mapping file: {}", new Object[]{e.getMessage()});
+                logger.error("Error loading mapping file: ", e);
```
exceptionfactory, line 271:
```suggestion
                logger.error("Error loading mapping file", e);
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
```diff
@@ -1260,20 +1260,20 @@ private FileChannel createReader(final File file, final long position) {
         try {
             reader = FileChannel.open(file.toPath(), StandardOpenOption.READ);
         } catch (final IOException ioe) {
-            getLogger().warn("Unable to open file {}; will attempt to access file again after the configured Yield Duration has elapsed: {}", new Object[]{file, ioe});
+            getLogger().warn("Unable to open file {}; will attempt to access file again after the configured Yield Duration has elapsed: ", file, ioe);
```
exceptionfactory, line 1263:
```suggestion
            getLogger().warn("Unable to open file {}; will attempt to access file again after the configured Yield Duration has elapsed", file, ioe);
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestListenHTTP.java
```diff
@@ -651,10 +651,10 @@ private int startWebServer() {
             try {
                 socket.connect(socketAddress, SOCKET_CONNECT_TIMEOUT);
                 connected = true;
-                runner.getLogger().debug("Server Socket Connected after {} ms", new Object[]{elapsed});
+                runner.getLogger().debug("Server Socket Connected after {} ms", elapsed);
                 socket.close();
             } catch (final Exception e) {
-                runner.getLogger().debug("Server Socket Connect Failed: [{}] {}", new Object[]{e.getClass(), e.getMessage()});
+                runner.getLogger().debug("Server Socket Connect Failed:", e);
```
exceptionfactory, line 657:
```suggestion
                runner.getLogger().debug("Server Socket Connect Failed", e);
```

exceptionfactory (changes requested):
Thanks for update @dan-s1, the tests look good.

I went through the changes and the vast majority are ready to go.

I noted a handful of instances where there was a trailing character that no longer makes sense, such as a semicolon. If you could address those items, this should be ready to go.

exceptionfactory (approved):
Thanks again for the cleanup @dan-s1, the latest version looks good! +1 merging

exceptionfactory:
@dan-s1 This commit does not apply cleanly to the support branch given the number of other changes, so if you are interested in having it applied, it would require a separate pull request. I also recommend handling it in a different Jira issue, since there are likely a number of other classes, now removed from main, where this would otherwise apply.

dan-s1:
@exceptionfactory Can I just then modify NIFI-13323 to include other places besides the HashAttribute and PostHttp Processors?

exceptionfactory:
> @exceptionfactory Can I just then modify NIFI-13323 to include other places besides the HashAttribute and PostHttp Processors?

Sure, that sounds like a good plan.

PR #8900, opened 2024-05-30
[NIFI-13246] move actions from details columns into menu

scottyaslan:


mcgilman:
Will review...

mcgilman (commented):
Thanks for the PR @scottyaslan! Works great but I would suggest that we move the `View Documentation` menu items lower in the menu. If we compare it to the context menu on the canvas it should probably sit just above the `Manage Access Policies` menu item.

mcgilman (approved):

PR #8914, opened 2024-06-03
NIFI-13159 PutFTP/PutSFTP change when delete happens on REPLACE

mosermw:
# Summary

[NIFI-13159](https://issues.apache.org/jira/browse/NIFI-13159)

For PutFTP/PutSFTP conflict resolution strategy REPLACE, when DOT_RENAME or temp filename is used, delay the deletion of remote file until after temp file transfer completes.

# Tracking

### Issue Tracking

### Pull Request Tracking

- [ x Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-00000`

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### UI Contributions

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/PutFileTransfer.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/PutFileTransfer.java
@@ -225,7 +225,6 @@ public abstract class PutFileTransfer<T extends FileTransfer> extends AbstractPr
                 logger.warn("Resolving conflict by rejecting {} due to conflicting filename with a directory or file already on remote server", flowFile);
                 break;
             case FileTransfer.CONFLICT_RESOLUTION_REPLACE:
-                transfer.deleteFile(flowFile, path, fileName);
                 destinationRelationship = REL_SUCCESS;
                 transferFile = true;
                 penalizeFile = false;
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/FTPTransfer.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/FTPTransfer.java
@@ -475,6 +475,18 @@ public class FTPTransfer implements FileTransfer {
         }
 
         if (!filename.equals(tempFilename)) {
+            // if destination fullPath already exists, delete it before rename
+            try {
+                int reply = client.mlst(fullPath);
+                if (FTPReply.isPositiveCompletion(reply)) {
+                    if (!client.deleteFile(fullPath)) {
+                        throw new IOException("Failed to remove file " + fullPath + " due to " + client.getReplyString());
+                    }
+                }
+            } catch (final IOException e) {
+                throw new IOException("Failed to replace remote file " + fullPath, e);
+            }
+
             try {
                 logger.debug("Renaming remote path from {} to {} for {}", tempFilename, filename, flowFile);
                 final boolean renameSuccessful = client.rename(tempFilename, filename);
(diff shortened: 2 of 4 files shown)


exceptionfactory (changes requested):
@mosermw can you provide some additional context for the reasons behind this proposed change in behavior?

Reviewing the code, this change introduces an additional `mlist()` command for FTP and `stat()` command for SFTP, for each file transferred through the corresponding Processors. Those commands require both file access and network communication, which could have an impact on high volume flows. Is there a particular reason for adding those calls prior to calling `delete()`? It seems like that should not be necessary.

mosermw:
Thanks for looking at this PR, @exceptionfactory 

> @mosermw can you provide some additional context for the reasons behind this proposed change in behavior?

As a use case, let's say I have a server running software that has an anxiety attack if a certain file doesn't exist, and it checks for that file every 5 seconds.  I have a requirement to update that file periodically.  I use the power of NiFi to generate the file contents and then PutSFTP the file into place.  Currently, my server software sometimes panics because it checks for the file while I am updating it.  This is because PutSFTP deletes the file first, transfers the file as ".filename" then renames to "filename".  As my file gets larger, it can take more than 5 seconds to transfer.

After the change in this PR, the file will only not exist in the short period of time between an SFTP delete then rename.

I asked the question on Slack a while back and got positive feedback that this change would be useful.  I should have put the link into the Jira ticket.  https://apachenifi.slack.com/archives/C0L9S92JY/p1713799005100369

> Reviewing the code, this change introduces an additional `mlist()` command for FTP and `stat()` command for SFTP, for each file transferred through the corresponding Processors. Those commands require both file access and network communication, which could have an impact on high volume flows. Is there a particular reason for adding those calls prior to calling `delete()`? It seems like that should not be necessary.

You're absolutely right and I can do better.  It doesn't hurt to call delete whether the destination file exists or not.  I will modify the PR to remove those additional commands and test again.

exceptionfactory:
> Thanks for looking at this PR, @exceptionfactory
> 
> > @mosermw can you provide some additional context for the reasons behind this proposed change in behavior?
> 
> As a use case, let's say I have a server running software that has an anxiety attack if a certain file doesn't exist, and it checks for that file every 5 seconds. I have a requirement to update that file periodically. I use the power of NiFi to generate the file contents and then PutSFTP the file into place. Currently, my server software sometimes panics because it checks for the file while I am updating it. This is because PutSFTP deletes the file first, transfers the file as ".filename" then renames to "filename". As my file gets larger, it can take more than 5 seconds to transfer.
> 
> After the change in this PR, the file will only not exist in the short period of time between an SFTP delete then rename.
> 
> I asked the question on Slack a while back and got positive feedback that this change would be useful. I should have put the link into the Jira ticket. https://apachenifi.slack.com/archives/C0L9S92JY/p1713799005100369
> 
> > Reviewing the code, this change introduces an additional `mlist()` command for FTP and `stat()` command for SFTP, for each file transferred through the corresponding Processors. Those commands require both file access and network communication, which could have an impact on high volume flows. Is there a particular reason for adding those calls prior to calling `delete()`? It seems like that should not be necessary.
> 
> You're absolutely right and I can do better. It doesn't hurt to call delete whether the destination file exists or not. I will modify the PR to remove those additional commands and test again.

Thanks for the reply and additional background @mosermw, that is helpful.

At the core, I agree the amount of time between uploading and renaming should be minimal. If you are able to rework the approach and avoid introducing the additional status calls, it seems like a viable improvement.

It is worth noting, however, that having some other independent process checking for the existence of a file is still bound to result in a race condition. This change may reduce the chances, but it does not sound like a complete solution.

So with that said, I will take another look when you have posted some updates.

mosermw:
> It is worth noting, however, that having some other independent process checking for the existence of a file is still bound to result in a race condition. This change may reduce the chances, but it does not sound like a complete solution.

Oh I totally agree, but sometimes we must do the best with the hand we are dealt.  I made the changes and the PR did get much simpler.

exceptionfactory:
Thanks for the update @mosermw, this looks like a good approach as the put method already considers the temporary file when deciding whether to run the rename command.

If the rename failed, it would throw an exception, which could occur if the delete failed for some reason. For troubleshooting, I believe it would be helpful to log the remote exception at the debug level, and include the remote filename that could not be deleted. That way, if the rename operation fails, there is the opportunity to enable additional logging that could explain more about what failed. Otherwise, I think this approach looks good.

mosermw:
Added debug logging if the delete command throws an exception.  This is a good suggestion also because some static analysis tools do not like empty catch blocks.

exceptionfactory (approved):
Thanks for making the adjustments @mosermw, the latest version looks good. +1 merging

PR #8915, opened 2024-06-03
NIFI-13337: Adjust column widths of the queue listing table

mcgilman:
NIFI-13337:
- Adjust column widths of the queue listing table.


scottyaslan:
Reviewing...

scottyaslan (approved):
LGTM! +1

PR #8926, opened 2024-06-04
[NIFI13360] rename nifi theme to supplemental theme

scottyaslan:


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-frontend/src/main/nifi/apps/nifi/src/assets/styles/_app.scss
```diff
@@ -408,39 +420,43 @@
     // NOTE: nifi surface palette lighter/darker is purposefully omitted here. Any application of nifi surface colors should be as a background-color to a surface DOM element. The surface color should switch from light to dark depending on the mode for the theme.
```
rfellows, line 420: Should we update the comment here?
```suggestion
    // NOTE: supplemental surface palette lighter/darker is purposefully omitted here. Any application of supplemental surface colors should be as a background-color to a surface DOM element. The surface color should switch from light to dark depending on the mode for the theme.
```

PR #8927, opened 2024-06-04
[NIFI-13361] determine extension description height base on $body-2 l…

scottyaslan:
…ine-height configuration


rfellows (approved):
Very elegant solution. Thanks @scottyaslan!

PR #8931, opened 2024-06-05
NIFI-13368: Allowing tooltip mouse listeners to be destroyed when necessary

mcgilman:
NIFI-13368:
- Allowing tooltip mouse listeners to be destroyed when necessary.
- Ensuring connection source/destination run status and validation errors are updated when deleted.


rfellows:
reviewing...

rfellows (approved):
Nice fix @mcgilman 👍

PR #8932, opened 2024-06-05
NIFI-13364: Remove UI NiFi properties

mcgilman:
NIFI-13364:
- Removing ui auto refresh interval property.
- Removing ui banner text property.


exceptionfactory (approved):
Thanks for the thorough cleanup @mcgilman, looks good! +1 merging

PR #8936, opened 2024-06-06
NIFI-13354: Prepping codebase for colocating UIs

mcgilman:
NIFI-13354:
- Moving NiFi front end source into a top level maven module. This prepares for the introduction of other UIs that NiFi offers (like Custom UIs, Data Viewers, etc) to be colocated and share common components, styles, and dependencies.
- The nifi-web-frontend module which produces the war that is included in the server nar now no longer contains any source. It simply depends on the new nifi-frontend artifact that bundles all built UIs and unpacks its contents into the resulting war.
- Renaming nifi-web-frontend to nifi-ui. Now nifi-frontend at the top level will hold all frontend apps. In this commit the nifi app in nifi-frontend is bundled into a war called nifi-ui.


exceptionfactory (approved):

rfellows (approved):
Looks good 👍

PR #8951, opened 2024-06-10
NIFI-13340: Fixed a bug in which an Output Port can leave a Process G…

markap14:
…roup's DataValve open for output, but then the last FlowFile is terminated instead of going to an Output Port, ultimately resulting in the DataValve remaining open indefinitely. Now, this will be detected and the valve will be closed.

# Summary

[NIFI-13340](https://issues.apache.org/jira/browse/NIFI-13340)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardDataValve.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardDataValve.java
@@ -46,8 +46,11 @@ public class StandardDataValve implements DataValve {
     private final ProcessGroup processGroup;
     private final StateManager stateManager;
 
-    private Set<String> groupsWithDataFlowingIn = new HashSet<>();
-    private Set<String> groupsWithDataFlowingOut = new HashSet<>();
+    private final Set<String> groupsWithDataFlowingIn = new HashSet<>();
+    private final Set<String> groupsWithDataFlowingOut = new HashSet<>();
+
+    private boolean leftOpenDueToDataQueued = false;
+
 
     public StandardDataValve(final ProcessGroup processGroup, final StateManager stateManager) {
         this.processGroup = processGroup;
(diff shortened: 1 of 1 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardDataValve.java
```diff
@@ -67,30 +70,46 @@ public synchronized boolean tryOpenFlowIntoGroup(final ProcessGroup destinationG
...
+        // - The FlowFile is split into two FlowFiles
+        // - One of the FlowFiles is routed to the Output Port, while the other is routed elsewhere
+        // - The Output Port is triggered. It opens the valve to allow data to flow out of the group.
+        // - The Output Port goes to close flow out of the group. However, the group is not empty, so the valve is not closed.
+        // - The other FlowFile is never routed to an output port. Instead, it is auto-terminated by some processor.
+        // Now, the valve has been left open.
+        // In this case, though, when the Output Port failed to close the valve, this.leftOpenDueToDataQueued was set to true. If that is the case,
+        // we can go ahead and close the valve now, if there's no more data queued.
+        if (reasonForNotAllowing == FlowInForbiddenReason.OPEN_FOR_OUTPUT && leftOpenDueToDataQueued && !destinationGroup.isDataQueued()) {
+            groupsWithDataFlowingOut.remove(destinationGroup.getIdentifier());
```
jrsteinebrey, line 88: Does reasonForNotAllowing need to be set equal to null here so that the we can execute past 
if (reasonForNotAllowing != null) { which causes an early return exit.
If not set to null then code will not execute:
groupsWithDataFlowingIn.add(destinationGroup.getIdentifier());
which seems to be the main goal of this method.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardDataValve.java
```diff
@@ -67,30 +70,46 @@ public synchronized boolean tryOpenFlowIntoGroup(final ProcessGroup destinationG
...
+        }
+
         if (reasonForNotAllowing != null) {
             // Since there is a reason not to allow it, return false. The reason has already been logged at a DEBUG level.
             return false;
         }
 
-
         logger.debug("Opening valve to allow data to flow into {}", destinationGroup);
         groupsWithDataFlowingIn.add(destinationGroup.getIdentifier());
```
jrsteinebrey, line 97: This is the part of the code which my first comment above is referring.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardDataValve.java
jrsteinebrey: I withdraw my above comment because I now that the code will work correctly as is because this tryOpenFlowIntoGroup will get called again in the future and then on the second call, it will open the flow.

exceptionfactory (approved):
Thanks for tracking this down @markap14! +1 merging

PR #8957, opened 2024-06-11
NIFI-13382: Showing a full page error when the front end can no longer communicate with the back end

mcgilman:
NIFI-13382:
- Showing a full page error when the front end can no longer communicate with the back end.


rfellows (approved):
LGTM 👍 
Thanks, @mcgilman

PR #8958, opened 2024-06-11
NIFI-13390 MergeRecord: Skip setting an attribute if autoTerminated

jrsteinebrey:
NIFI-13390 MergeRecord: Skip setting an attribute on flow file which is being sent to autoTerminated original relationship

# Summary

[NIFI-13390](https://issues.apache.org/jira/browse/NIFI-13390)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-13390) issue created

### Pull Request Tracking

- [X] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-13390`
- [X] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-13390`

### Pull Request Formatting

# Verification

Running automated tests was the verification performed.

### Build

  - [X] JDK 21

### Licensing

No new files or dependencies

### Documentation

No documentation changes

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/merge/RecordBin.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/merge/RecordBin.java
@@ -400,7 +400,9 @@ public class RecordBin {
             attributes.put(MergeRecord.MERGE_BIN_AGE_ATTRIBUTE, Long.toString(getBinAge()));
 
             merged = session.putAllAttributes(merged, attributes);
-            flowFiles.forEach(ff -> session.putAttribute(ff, MergeRecord.MERGE_UUID_ATTRIBUTE, merged.getAttribute(CoreAttributes.UUID.key())));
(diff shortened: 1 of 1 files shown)


exceptionfactory (changes requested):
Thanks for proposing this change @jrsteinebrey.

Although this is a minor change, I am not inclined to move it forward. Although there is some cost associated with updating an attribute, this seems to be moving in the direction of introducing complexity for minimal gains. Avoiding content-related changes is worth skipping, but this is more minimal. I think we should avoid going into this level of adjustment because introducing conditionals in various places adds otherwise unnecessary complexity.

If others think differently, I'm not absolutely opposed to this specific change, but I'm raising a flag on this particular setting as an example of something where additional effort does not seem worth making the changes.

MikeThomsen (commented):
I'm inclined to agree with @exceptionfactory here. If this were on `SplitRecord` and had the potential to write the same attribute 10k times as it auto-terminates, that'd be something to avoid. This is a single flowfile's attribute, so its impact won't be meaningful on the JVM.

jrsteinebrey:
@exceptionfactory @MikeThomsen Thanks for your feedback. I see your point. I am closing this PR.

PR #8960, opened 2024-06-12
NIFI-13391: Setting up different environment files to configure ngrx store

mcgilman:
NIFI-13391:
- Setting up different environment files to configure ngrx store.


scottyaslan (approved):
LGTM

PR #8961, opened 2024-06-13
NIFI-13394: Check candidate virtual env for python command.

bobpaulin:
# Summary

[NIFI-13394](https://issues.apache.org/jira/browse/NIFI-13394)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

Built and started up on windows with python modules.

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcess.java
+++ b/nifi-extension-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcess.java
@@ -302,7 +302,7 @@ public class PythonProcess {
         } else if (virtualEnvDirectories.length == 1) {
             commandExecutableDirectory = virtualEnvDirectories[0].getName();
         } else {
-            // We should get at most 2 directories.  Check for python command. 
+            // We should get at most 2 directories.  Check for python command.
             if (virtualEnvDirectories[0].list((file, name) -> name.startsWith(pythonCmd)).length >= 1) {
                 commandExecutableDirectory = virtualEnvDirectories[0].getName();
             } else if (virtualEnvDirectories[1].list((file, name) -> name.startsWith(pythonCmd)).length >= 1) {
@@ -310,7 +310,7 @@ public class PythonProcess {
             } else {
                 throw new IOException("Failed to find pythond command: " + pythonCmd);
             }
-        } 
+        }
 
         final File pythonCommandFile = new File(virtualEnvHome, commandExecutableDirectory + File.separator + pythonCmd);
         return pythonCommandFile.getAbsolutePath();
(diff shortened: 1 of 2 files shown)


nifi-extension-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcess.java
```diff
@@ -302,8 +302,14 @@ String resolvePythonCommand() throws IOException {
         } else if (virtualEnvDirectories.length == 1) {
             commandExecutableDirectory = virtualEnvDirectories[0].getName();
         } else {
-            // Default to bin directory for macOS and Linux
-            commandExecutableDirectory = "bin";
+            // We should get at most 2 directories.  Check for python command.
+            if (virtualEnvDirectories[0].list((file, name) -> name.startsWith(pythonCmd)).length >= 1) {
```
exceptionfactory, line 306: The `list()` method can return `null` under some circumstances, so adjusting the approach seems necessary. What about moving this block into a separate method named something like `findExecutableDirectory`? That method could iterate over the `virtualEnvDirectories` array, and not necessary assume that there will be at most two directories.

nifi-extension-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcess.java
```diff
@@ -302,8 +302,14 @@ String resolvePythonCommand() throws IOException {
...
+                commandExecutableDirectory = virtualEnvDirectories[0].getName();
+            } else if (virtualEnvDirectories[1].list((file, name) -> name.startsWith(pythonCmd)).length >= 1) {
+                commandExecutableDirectory = virtualEnvDirectories[1].getName();
+            } else {
+                throw new IOException("Failed to find pythond command: " + pythonCmd);
```
exceptionfactory, line 311: Minor spelling and string construction recommendation:
```suggestion
                throw new IOException("Failed to find Python command [%s]".formatted(pythonCmd));
```

exceptionfactory (changes requested):
Thanks for working on this @bobpaulin. The concept makes sense, but for null-safety, I recommend reworking the command resolution a bit.

nifi-extension-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcess.java
bobpaulin: Yes I agree it is safer if I make it null safe and cleaner with a new method.  The challenge was working with the virtualEnvDirectories array in a for loop and ensuring commandExecutableDirectory is only set exactly once.  If I break I don't guarantee it's set if I loop without breaking it may be set multiple.  So I've replaced this with a stream which is not the most efficient but  does make the code readable and keeps the final guarantee or throws an exception.

nifi-extension-bundles/nifi-py4j-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/py4j/PythonProcess.java
bobpaulin: Thank you for catching that and for the additional feedback of course.

exceptionfactory (changes requested):
Thanks for the updates @bobpaulin. The new method looks cleaner, but the unit tests are now failing, so it looks like some additional changes are required.

exceptionfactory (approved):
Thanks for working through the feedback @bobpaulin, the latest version looks good! +1 merging

PR #8963, opened 2024-06-13
NIFI-13396 Include python processors' allowable values in manifest

briansolo1985:
# Summary

[NIFI-13396](https://issues.apache.org/jira/browse/NIFI-13396)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/manifest/StandardRuntimeManifestService.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/manifest/StandardRuntimeManifestService.java
@@ -20,7 +20,6 @@ import org.apache.nifi.bundle.Bundle;
 import org.apache.nifi.bundle.BundleDetails;
 import org.apache.nifi.c2.protocol.component.api.BuildInfo;
 import org.apache.nifi.c2.protocol.component.api.RuntimeManifest;
-import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.extension.manifest.AllowableValue;
 import org.apache.nifi.extension.manifest.ControllerServiceDefinition;
 import org.apache.nifi.extension.manifest.ExpressionLanguageScope;
@@ -212,8 +211,8 @@ public class StandardRuntimeManifestService implements RuntimeManifestService {
         property.setRequired(propertyDescription.isRequired());
         property.setSensitive(propertyDescription.isSensitive());
 
-        // TODO: Handle Allowable Values
         property.setControllerServiceDefinition(getManifestControllerServiceDefinition(propertyDescription.getControllerServiceDefinition()));
+        property.setAllowableValues(getAllowableValues(propertyDescription));
 
         return property;
     }
@@ -228,22 +227,17 @@ public class StandardRuntimeManifestService implements RuntimeManifestService {
         return definition;
     }
 
-    private static List<AllowableValue> getManifestAllowableValues(final PropertyDescriptor propertyDescriptor) {
-        if (propertyDescriptor.getAllowableValues() == null) {
-            return List.of();
-        }
-
-        final List<AllowableValue> allowableValues = new ArrayList<>();
-        for (final org.apache.nifi.components.AllowableValue allowableValue : propertyDescriptor.getAllowableValues()) {
-            final AllowableValue manifestAllowableValue = new AllowableValue();
-            manifestAllowableValue.setDescription(allowableValue.getDescription());
-            manifestAllowableValue.setValue(allowableValue.getValue());
-            manifestAllowableValue.setDisplayName(allowableValue.getDisplayName());
-            allowableValues.add(manifestAllowableValue);
-        }
-        return allowableValues;
+    private static List<AllowableValue> getAllowableValues(final PropertyDescription propertyDescription) {
+        return Optional.ofNullable(propertyDescription.getAllowableValues()).orElse(List.of())
+            .stream()
+            .map(value -> {
+                AllowableValue allowableValue = new AllowableValue();
+                allowableValue.setValue(value);
+                allowableValue.setDisplayName(value);
+                return allowableValue;
+            })
+            .toList();
     }
-
 
     private static List<UseCase> getUseCases(final PythonProcessorDetails pythonProcessorDetails) {
         final List<UseCase> useCases = new ArrayList<>();
--- a/nifi-extension-bundles/nifi-py4j-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/documentation.py
+++ b/nifi-extension-bundles/nifi-py4j-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/documentation.py
@@ -107,7 +107,8 @@ class PropertyDescription:
                  sensitive: bool = True,
                  default_value: str = None,
                  expression_language_scope: str = None,
-                 controller_service_definition: str = None):
+                 controller_service_definition: str = None,
+                 allowable_values: list[str] = None):
         self.name = name
         self.description = description
         self.display_name = display_name
@@ -116,6 +117,7 @@ class PropertyDescription:
         self.default_value = default_value
         self.expression_language_scope = expression_language_scope
         self.controller_service_definition = controller_service_definition
+        self.allowable_values = allowable_values if allowable_values is not None else []
 
     def getName(self):
         return self.name
@@ -140,3 +142,6 @@ class PropertyDescription:
 
     def getControllerServiceDefinition(self):
         return self.controller_service_definition
+
+    def getAllowableValues(self):
+        return ArrayList(self.allowable_values)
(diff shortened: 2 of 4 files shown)


briansolo1985:
Currently there are some properties which are not populated in the manifest for python processors, like allowable values, relationships, property dependencies. This PR is intended to be the first one in the series of address the above mentioned features.
If the approach is satisfactory, I would use it in the upcoming PRs well.

minifi/minifi-commons/minifi-commons-api/src/main/java/org/apache/nifi/minifi/commons/api/MiNiFiProperties.java
```diff
@@ -81,9 +81,8 @@ public enum MiNiFiProperties {
     C2_KEEP_ALIVE_DURATION("c2.rest.keepAliveDuration", "5 min", false, true, TIME_PERIOD_VALIDATOR),
     C2_REST_HTTP_HEADERS("c2.rest.http.headers", "Accept:application/json", false, true, VALID),
     C2_CONFIG_DIRECTORY("c2.config.directory", "./conf", false, true, VALID),
-    C2_RUNTIME_MANIFEST_IDENTIFIER("c2.runtime.manifest.identifier", "", false, true, VALID),
-    C2_RUNTIME_TYPE("c2.runtime.type", "", false, true, VALID),
-    C2_ASSET_DIRECTORY("c2.asset.directory", "./asset", false, true, VALID),
+    C2_RUNTIME_MANIFEST_IDENTIFIER("c2.runtime.manifest.identifier", "minifi", false, true, VALID),
+    C2_RUNTIME_TYPE("c2.runtime.type", "minifi-java", false, true, VALID),    C2_ASSET_DIRECTORY("c2.asset.directory", "./asset", false, true, VALID),
```
exceptionfactory, line 85: This change seems unrelated, I recommend reverting it and making it in another pull request if required.

briansolo1985: Reverted and created https://github.com/apache/nifi/pull/8970

exceptionfactory: Thanks!

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/manifest/StandardRuntimeManifestService.java
```diff
@@ -228,23 +227,18 @@ private static ControllerServiceDefinition getManifestControllerServiceDefinitio
...
-        for (final org.apache.nifi.components.AllowableValue allowableValue : propertyDescriptor.getAllowableValues()) {
-            final AllowableValue manifestAllowableValue = new AllowableValue();
-            manifestAllowableValue.setDescription(allowableValue.getDescription());
-            manifestAllowableValue.setValue(allowableValue.getValue());
-            manifestAllowableValue.setDisplayName(allowableValue.getDisplayName());
-            allowableValues.add(manifestAllowableValue);
-        }
-        return allowableValues;
+    private static List<AllowableValue> getAllowableValues(final PropertyDescription propertyDescription) {
+        return Optional.ofNullable(propertyDescription.getAllowableValues()).orElse(List.of())
```
exceptionfactory, line 231: This should use `Collections.emptyList()` to avoid unnecessary creation:
```suggestion
        return Optional.ofNullable(propertyDescription.getAllowableValues()).orElse(Collections.emptyList())
```

exceptionfactory (commented):
Thanks reorganizing things @briansolo1985. The basic approach looks good, just noted one minor improvement on initial review.

exceptionfactory (commented):
@briansolo1985 on closer review and separate conversation with @markap14, the requirement to use string constants from the same file seems concerning. Although it makes sense in Processors where the Property Descriptor is defined in the same class, the approach doesn't work when the Property Descriptor is defined separately and reused. For this reason, it looks like there is more detailed work required to evaluate the best way to support an alternative way to determine allowable values.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/manifest/StandardRuntimeManifestService.java
briansolo1985: List.of() does not create a new instance, it returns a static constant immutable list just like Collections.emptyList(). I prefer the former due it's shortness and conciseness.

briansolo1985:
Thanks for the feedback @exceptionfactory.

I was aware of that limitation. The current solution works for all of the Python processors currently present  in the NiFi codebase, except for ChunkDocument LANGUAGE property's allowable values where None is returned.
My feeling was that it is somewhat allowed / tolerated in the Python processor world, as the comment says on the related Java interface:
```
    /**
     * A list of property descriptions for the known properties. Note that unlikely Java Processors, Python-based Processors are
     * more dynamic, and the properties may not all be discoverable
     *
     * @return a list of descriptions for known properties
     */
    List<PropertyDescription> getPropertyDescriptions();
```

Regarding a more generic solution, it might be more complex and demanding.
I'm not a Python expert, and maybe there are already answers, but I came across the following difficulties:
* Retrieving values with AST gets complicated when the assignment is an ast.Call (eg object creation) or an ast.Name (eg assignment by name). For the former we need to implement the creation logic for each object type, for the latter we need to create a lookup map containing the name and the references to get the actual object not just the name. Creating the lookup map has the same difficulties
* Python is super permissive - as the Javadoc comment says correctly. For example we can define constants outside and inside classes. Another example is the syntax.
Given we defined a class:
```
class Descriptor():
   def __init__(self, name, description):
        self.name = name
        self.description = description
```
We can instantiate the object in three different ways:
```
Descriptor("name1",  "desc1")
Descriptor("name2",  description="desc2")
Descriptor(name="name3",  description="desc3")
```
From AST perspective these are all different cases, which had to be handled from code. Usually classes are more complex which will result in a rather complex code, or specific to each class.
* Evaluating functions: assignments can be more complex than assigning a constant value or instantiating new object. For example it can be a list comprehension just like in ChunkDocument. Processing expression like this is possible with AST, although it's rather complex, and hard to make it generic.
* Definitions spanning across multiple Python modules. An AST tree is built for single Python module, and does aware of it's dependencies. Although it seems to be possible to process import statement while traversing the AST tree, it will again lead to complex and hard to maintain codebase.
A bigger problem is: at manifest build time not all Python dependencies are guaranteed to be there, as Python processors are async loaded, and are downloading their dependencies after the processor being started as part of a flow. Instead a manifest is created during startup, where a flow is not necessarily present.

In summary my intention was to provide a good enough solution to make "allowable values" present in the manifest, while considering the current limitations we have.

For the dependency download issue https://issues.apache.org/jira/browse/NIFI-12959 can be an answer. If the dependencies are self contained, it should be possible to also contain a extension-manifest.xml in the NAR which could be the just picked up by StandardRuntimeManifestService, making the whole parsing logic unnecessary.

Still, for Python processors without self encapsulated NARs would require a way to provide information for the manifest. 
Besides "allowable values", "relationships" and "property dependencies" are also missing function
Given the above reasons, I'm not confident that AST is the best tool to achieve this.

What do you think? Is there already a plan or a suggested approach here? I'm open to any suggestion / solutions, and happy to participate in implementing them.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/manifest/StandardRuntimeManifestService.java
exceptionfactory: Thanks for noting that important detail, I agree, leaving it as `List.of()` sounds good.

ferencerdei:
Hello @exceptionfactory @markap14,
It would be great to follow up on this topic. The solution would affect the PropertyDependency and Custom relationship values as well. This information is missing from doc generation as well. I found a related JIRA already: https://issues.apache.org/jira/browse/NIFI-13432  
Based on what @briansolo1985 said, the current approach to how NiFi interact with the Python libraries and AST has a limitation. I wonder if you had a chance to look into this in more detail or have an alternative solution/suggestion.

exceptionfactory (commented):
Thanks for the thoughtful reply @briansolo1985, and thanks the additional comments @ferencerdei.

I agree that the nature of Python and the current ability to load dependencies at runtime complicate potential approaches. I don't have an immediate alternative suggestion.

I'm somewhat inclined to go forward with the current solution, as it covers a number of cases, and seems like it would be better than nothing.

Tagging @markap14 for additional input on other ideas.

markap14:
Thanks for pinging @exceptionfactory. Truthfully I would prefer not implementing anything until we provide a full solution. But it's not a particularly strong preference. I wouldn't necessarily oppose the addition as proposed here.

exceptionfactory (approved):
Thanks for the perspective @markap14, and thanks again for the work on this @briansolo1985. Although there are still cases that this doesn't support, this approach moves things forward for now, so I will proceed to merge.

PR #8969, opened 2024-06-15
NIFI-13406: enable netty native transports

sullis:
# Summary

[NIFI-13406](https://issues.apache.org/jira/browse/NIFI-13406)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/test/java/org/apache/nifi/event/transport/netty/NettyTransportsTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class NettyTransportsTest {
+  @Test
+  public void nioIsAvailable() {
+    assertTrue(NettyTransports.NIO.isAvailable());
+    assertTrue(NettyTransports.AVAILABLE_TRANSPORT_NAMES.contains("nio"));
+  }
```
exceptionfactory, line 35: This class should be changed to use four spaces instead of two spaces for indentation.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/EventLoopGroupFactory.java
```diff
@@ -35,6 +34,16 @@ class EventLoopGroupFactory {
...
 
+    private NettyTransports.NettyTransport nettyTransport;
+
+    public EventLoopGroupFactory() {
+        this(NettyTransports.getDefaultNettyTransport());
+    }
+
+    public EventLoopGroupFactory(NettyTransports.NettyTransport nettyTransport) {
+        this.nettyTransport = nettyTransport;
+    }
```
exceptionfactory, line 45: I recommend avoiding the addition of the parameterized constructor and instead use the availability determination logic in the get default method. Callers should not need to select a particular implementation, thus keeping the selection details private to this factory class.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyTransports.java
```diff
@@ -0,0 +1,92 @@
...
+      try {
+        Constructor<? extends EventLoopGroup> c = eventLoopGroupClass.getConstructor(Integer.TYPE, ThreadFactory.class);
+        return c.newInstance(workerThreads, threadFactory);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+    }
+  }
+
+  public static final NettyTransport EPOLL = new NettyTransport("epoll", Epoll.isAvailable(),
```
exceptionfactory, line 60: With the approach to keep the selection logic internal to this module, it looks like most of the members of this class could have `private` or package-level visibility.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyEventServerFactory.java
```diff
@@ -202,14 +200,15 @@ private void setChannelOptions(final AbstractBootstrap<?, ?> bootstrap) {
     }
 
     private AbstractBootstrap<?, ?> getBootstrap() {
+        NettyTransports.NettyTransport nettyTransport = this.getNettyTransport();
```
exceptionfactory, line 203: Instead of interacting with the `NettyTransport` record, it seems like this could be changed to have a `getDatagramChannelClass()` and `getServerSocketClass()` method that avoids exposing the record class itself.

exceptionfactory (commented):
Thanks for putting this together @sullis! The general approach looks good, the additional dependencies do not appear very large from initial review.

I raised a few points to consider regarding the implementation details, as far as what to expose. I think as an initial approach, keeping the native options hidden behind automatic availability selection is ideal. That way, calling components do not need additional logic. If we get to the point where the implementation needs to be selected, that's an option, but keeping it internal to the module seems like the best initial approach.

I can take a look at some runtime tests soon, but the general concept looks like it will be a straightforward and helpful improvement to several components.

joewitt:
@exceptionfactory I was thinking we need to avoid having what gets included be a function of the build selector.  We dont do a different build per runtime environment so that seems tough as-is.  Was thinking perhaps a controller service based model would be an option but I didnt really check meaningfully as was on the phone,.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/EventLoopGroupFactory.java
sullis: Agreed.  I'll remove the parameterized constructor.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyEventServerFactory.java
sullis: Agreed. I'll remove the getNettyTransport() method.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyTransports.java
sullis: Agreed.  I think package-level visibility is reasonable.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/test/java/org/apache/nifi/event/transport/netty/NettyTransportsTest.java
sullis: Fixed.

sullis:
> 
> I think as an initial approach, keeping the native options hidden behind automatic availability selection is ideal. That way, calling components do not need additional logic. If we get to the point where the implementation needs to be selected, that's an option, but keeping it internal to the module seems like the best initial approach.

Agreed.  I have incorporated your feedback and pushed changes to this PR.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/EventLoopGroupFactory.java
```diff
@@ -54,7 +63,11 @@ public void setWorkerThreads(final int workerThreads) {
     }
 
     protected EventLoopGroup getEventLoopGroup() {
```
sullis, line 64: perhaps this method should be renamed from 'getEventLoopGroup' to 'createEventLoopGroup'

exceptionfactory:
> @exceptionfactory I was thinking we need to avoid having what gets included be a function of the build selector. We dont do a different build per runtime environment so that seems tough as-is. Was thinking perhaps a controller service based model would be an option but I didnt really check meaningfully as was on the phone,.

@joewitt The `classifier` property on the dependency selects the file to include. The current version of the pull request includes support for Linux and macOS on x86_64 and aarch64, which should cover the most common use cases. This doesn't make the build specific to the runtime, but instead makes the native dependencies available to those runtime platforms. If the runtime OS and architecture match one of those supported libraries, that support should be enabled automatically according to this initial implementation.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/EventLoopGroupFactory.java
exceptionfactory: That could be worth considering, given the behavior, although I think it is fine to leave the current name for now.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyTransports.java
```diff
@@ -0,0 +1,88 @@
...
+import io.netty.channel.socket.nio.NioDatagramChannel;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.channel.socket.nio.NioSocketChannel;
+import java.lang.reflect.Constructor;
+import java.util.List;
+import java.util.concurrent.ThreadFactory;
+
+
+public class NettyTransports {
+    public record NettyTransport(String name,
```
exceptionfactory, line 43: Instead of having this as a `record`, it could be changed to an `enum`. That would also remove the need for the separate `name` property.

exceptionfactory (changes requested):
Thanks for making the adjustments @sullis, the changes look good.

On subsequent review, it seems like it would be helpful to change the `NettyTransport` from a `record` to an `enum`, which would remove the need for the separate string `name` property, and also make it easier to compare expected values in the test classes.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/EventLoopGroupFactory.java
sullis: I changed the method name to 'createEventLoopGroup'

sullis:
@exceptionfactory  would you like me to manually squash the commits or will you use GitHub to squash the PR commits?

joewitt:
> > @exceptionfactory I was thinking we need to avoid having what gets included be a function of the build selector. We dont do a different build per runtime environment so that seems tough as-is. Was thinking perhaps a controller service based model would be an option but I didnt really check meaningfully as was on the phone,.
> 
> @joewitt The `classifier` property on the dependency selects the file to include. The current version of the pull request includes support for Linux and macOS on x86_64 and aarch64, which should cover the most common use cases. This doesn't make the build specific to the runtime, but instead makes the native dependencies available to those runtime platforms. If the runtime OS and architecture match one of those supported libraries, that support should be enabled automatically according to this initial implementation.

ahhh yeah i see now.  Was thinking that controlled what gets selected :).  Doh. 

I'm very curious what the practical performance gain here is.  @sullis any benchmarking?

sullis:
> I'm very curious what the practical performance gain here is. @sullis any benchmarking?

I have not benchmarked Apache NiFi with Netty epoll.

This presentation talks about the motivation for the epoll transport:

https://youtu.be/DKJ0w30M0vg?t=986

image

The book [Netty for Action](https://www.manning.com/books/netty-in-action) describes the epoll transport as follows:

image

joewitt:
i got ya.  Motivation makes sense - was just curious if we had numbers to back it up

sullis:
I added Apache NiFi to the Netty adopters page:

https://github.com/netty/netty/wiki/Adopters

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyTransport.java
```diff
@@ -0,0 +1,120 @@
...
+
+public enum NettyTransport {
+    NIO("nio", true, NioEventLoopGroup.class, NioSocketChannel.class, NioServerSocketChannel.class, NioDatagramChannel.class),
+    EPOLL("epoll", Epoll.isAvailable(), EpollEventLoopGroup.class, EpollSocketChannel.class, EpollServerSocketChannel.class, EpollDatagramChannel.class),
+    KQUEUE("kqueue", KQueue.isAvailable(), KQueueEventLoopGroup.class, KQueueSocketChannel.class, KQueueServerSocketChannel.class, KQueueDatagramChannel.class);
+
+    private static final NettyTransport[] ALL_VALUES = values();
+
+    private final String nettyName;
+    private final boolean isAvailable;
```
exceptionfactory, line 52: Recommend changing this to `available`.
```suggestion
    private final boolean available;
```

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyTransport.java
```diff
@@ -0,0 +1,120 @@
...
+
+
+public enum NettyTransport {
+    NIO("nio", true, NioEventLoopGroup.class, NioSocketChannel.class, NioServerSocketChannel.class, NioDatagramChannel.class),
+    EPOLL("epoll", Epoll.isAvailable(), EpollEventLoopGroup.class, EpollSocketChannel.class, EpollServerSocketChannel.class, EpollDatagramChannel.class),
+    KQUEUE("kqueue", KQueue.isAvailable(), KQueueEventLoopGroup.class, KQueueSocketChannel.class, KQueueServerSocketChannel.class, KQueueDatagramChannel.class);
+
+    private static final NettyTransport[] ALL_VALUES = values();
+
+    private final String nettyName;
```
exceptionfactory, line 51: Is this property required? It seems unnecessary given the name of the `enum` instance itself.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/test/java/org/apache/nifi/event/transport/netty/NettyTransportsTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @Test
+    public void nioIsAvailable() {
+        assertTrue(NettyTransport.NIO.isAvailable());
+    }
+
+    @Test
+    @EnabledOnOs(value = OS.LINUX)
+    public void epollIsAvailableOnLinux() {
+        Epoll.ensureAvailability();
+        assertTrue(Epoll.isAvailable());
```
exceptionfactory, line 40: This can be removed because `ensureAvailability` will throw an exception if it is not available.
```suggestion
```

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/test/java/org/apache/nifi/event/transport/netty/NettyTransportsTest.java
```diff
@@ -0,0 +1,64 @@
...
+        assertTrue(NettyTransport.EPOLL.isAvailable());
+    }
+
+    @Test
+    @EnabledOnOs(value = OS.MAC)
+    public void kqueueIsAvailableOnMac() {
+        KQueue.ensureAvailability();
+        assertTrue(KQueue.isAvailable());
```
exceptionfactory, line 48:
```suggestion
```

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/test/java/org/apache/nifi/event/transport/netty/NettyTransportsTest.java
```diff
@@ -0,0 +1,64 @@
...
+        assertTrue(NettyTransport.KQUEUE.isAvailable());
+    }
+
+    @Test
+    public void getDefaultTransport() {
```
exceptionfactory, line 53:
```suggestion
    public void testGetDefaultTransport() {
```

exceptionfactory (changes requested):
The code is looking good, just a few more minor notes.

However, when evaluating the runtime behavior, the native libraries are not working as designed. Have you tested these changes at runtime with any particular Processors?

Testing the `ListenTCP` Processor on macOS with ARM64, the `KQueue` transport was not reported as available when evaluating with a connected debugger. The reported reason from Netty was:

```
java.lang.NoClassDefFoundError: io/netty/channel/kqueue/KQueueStaticallyReferencedJniMethods
```

This is probably due to the location of the Netty native transport libraries in the `nifi-standard-nar` as opposed to the `nifi-standard-shared-nar`. The `shared-nar` provides the standard `netty-transport` library to multiple inheriting NAR bundles, so the native transport libraries also need to be included at that level, and set as provided in standard shared BOM.

@sullis, If you are not familiar with those details, I could take a look at making the appropriate adjustments and pushing a change to this branch. This is certainly one of those changes that requires runtime testing and careful evaluation, since unit tests do not have the same class loader structure as runtime deployments.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyTransport.java
sullis: Fixed.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/test/java/org/apache/nifi/event/transport/netty/NettyTransportsTest.java
sullis: fixed

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/test/java/org/apache/nifi/event/transport/netty/NettyTransportsTest.java
sullis: fixed

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/test/java/org/apache/nifi/event/transport/netty/NettyTransportsTest.java
sullis: Fixed.

sullis:
> @sullis, If you are not familiar with those details, I could take a look at making the appropriate adjustments and pushing a change to this branch. This is certainly one of those changes that requires runtime testing and careful evaluation, since unit tests do not have the same class loader structure as runtime deployments.

Please go ahead and make any adjustments that you need.  Feel free to push your updates to this PR.

nifi-extension-bundles/nifi-extension-utils/nifi-event-transport/src/main/java/org/apache/nifi/event/transport/netty/NettyTransport.java
sullis: I was thinking about using the nettyName in a log message.   That's the only use case I was considering.

exceptionfactory: Thanks, in that case, I think the `enum` `name()` is sufficient.

sullis: Cool.  Feel free to make any adjustments to this PR.

exceptionfactory (commented):
@sullis I pushed a commit adding the native transport libraries to the `nifi-standard-shared-nar` and marking them as provided in the `nifi-standard-shared-bom`. This corrects the runtime class loading so that Processors such as `ListenTCP` and `PutTCP` discover the libraries as expected.

I ran some simple tests with a basic flow, sending a small payload from `PutTCP` to `ListenTCP` on the same machine, macOS with ARM64, which used the KQueue implementation. For the basic flow using a single concurrent task, I did not observe any noticeable performance differences between NiFi 2.0.0-M3 using the NIO transport and this branch using the KQueue transport. That may be due to other factors. I also plan to try the comparison on Linux soon.

Some commentary on socket implementations note that the JDK on Linux uses epoll and macOS uses kqueue under the hood.

Although this change is narrowly, it does bring in four additional libraries. For that reason, I think it would be helpful to provide some concrete examples of circumstances under which this provides noticeable improvements, so we have that as a point of reference.

joewitt:
@exceptionfactory From my admittedly limited reading the impression I had is that these native implementations become substantially more efficient at high rates/high scale/many concurrent connections/etc..  I suspect we'd not see it in routine testing but rather in deliberately pushing the agenda.  

With that in mind I am more inclined to suggest this only be activated via a profile allowing evaluation over time but avoiding inclusion in the normal case for now.  

Just my two cents.  I'm cool with a different approach too.

exceptionfactory:
> @exceptionfactory From my admittedly limited reading the impression I had is that these native implementations become substantially more efficient at high rates/high scale/many concurrent connections/etc.. I suspect we'd not see it in routine testing but rather in deliberately pushing the agenda.
> 
> With that in mind I am more inclined to suggest this only be activated via a profile allowing evaluation over time but avoiding inclusion in the normal case for now.
> 
> Just my two cents. I'm cool with a different approach too.

Thanks for the feedback @joewitt. I agree from reading some reference that the performance benefits should be more noticeable around large numbers of connections, so that's just where I would like to see a bit more comparison between Java 21 NIO and the native transports. With JDK improvements, perhaps the differences are not as noticeable, or only noticeable in certain circumstances. That's where having more recent and substantive examples would be helpful.

I'm not favor of a profile-based approach, because that would make the code more complicated, not being able to compile as it stands, if the libraries were not included. This could certainly be implemented using reflection, but as a non-default profile, I expect it would be minimally used.

To be clear, I think this could still be valuable, I would just like to see some type of scenario in NiFi, or benchmarks from other systems on Java 21, that illustrate the benefits.

exceptionfactory:
@sullis I went through some additional runtime testing on Ubuntu Linux to evaluate any notable differences. Based on some general reading, it seems like the Netty epoll implementation on Linux can be helpful in certain scenarios with large numbers of short-lived connections, but not necessarily with long-lived connections. I also noted that the OpenJDK NIO implementation uses epoll, but the triggering mechanism is different.

With this background, I'm not discounting the value of Netty native transports in certain scenarios, but it also seems like the value is only apparently under certain conditions.

I appreciate the background work on this, but I think it would be better to avoid a general change and instead consider particular components and use cases where it could provide value. As mentioned previously, although the libraries are fairly small, there should be a compelling reason to begin incorporating them in standard builds.

If you are interested in evaluating some particular Processors or use cases that highlight the value of Netty native libraries, that would be beneficial. Without that though, I'm not in favor of moving forward right now.

sullis:
Let's close this for now. I might revisit this when Netty 4.2 becomes available. (Version 4.2 will include the io_uring transport)

exceptionfactory:
Thanks @sullis, that sounds good, closing for now.

sullis:
For posterity: these are two of my other Netty libraries

(WIP) https://github.com/nettyplus/netty4-transports

(WIP) https://github.com/nettyplus/netty5-transports

PR #8978, opened 2024-06-18
NIFI-13414 Remove Property Protection Modules and Encrypt Config

exceptionfactory:
# Summary

[NIFI-13414](https://issues.apache.org/jira/browse/NIFI-13414) Removes the `nifi-property-protection-api` and associated implementation modules as well as the `nifi-toolkit-encrypt-config` and `minifi-toolkit-encrypt-config` modules. These modules provided encryption and decryption capabilities for application properties in `nifi.properties` and similar configuration files in MiNiFi and NiFi Registry. These changes do not impact the behavior encryption and decryption for sensitive values in flow configuration components.

[NIFI-1831](https://issues.apache.org/jira/browse/NIFI-1831) Introduced initial support for application property encryption in NiFi 1.0.0, with acknowledged issues related to having the encryption key stored in a file next to the configuration properties themselves. Although subsequent versions of NiFi introduced support for externalized secrets management, integration with different service providers required large dependency trees and configuration that was specific to each provider. Although the implementation provided some abstraction between interface and implementation, the API was not considered a public extension point, and thus required significant and specialized internal maintenance.

With this background, changes for NIFI-13414 remove all references to local and service-based encryption for application properties. This avoids the inherent confusion related to encryption of properties in one file while having the plaintext key in another file in the same directory. Removing these capabilities across NiFi, NiFi Registry, and MiNiFi as part of version 2.0.0 serves to clear the source repository for potential new approaches that avoid the security and implementation concerns of the current approach.

These changes remove optional bootstrap configuration files and the `encrypt-config.sh` command from the NiFi Toolkit. The dependency removals reduce the size of the NiFi binary distribution by over 60 MB, removing over 130 JAR files from the `properties` library directory.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


joewitt:
full clean build with contrib check looks good.  Full run of all integration tests look good.  Scanning through the code changes looks good.

This makes me think of our sensitive properties key construct in general.  We should formalize that the purpose of that is simply to help protect accidental distribution of sensitive values outside the bounds of a given nifi cluster.  The better answer and one we didn't have then that we do now is the usage of real vault stored parameters which thanks to parameter contexts and now be handled correctly.  It seems worthwhile to formalize and push to that answer and reduce usage of otherwise locally protected sensitive values.

I suppose we cache locally such vault values today.  We should not.  They should be retrieved on-deman/in memory only and we keep the protection of them as a function of the vault and the applications permission to access them.  Nothing to do with our sensitive key.

This also feels like it helps us make big progress https://issues.apache.org/jira/browse/NIFI-13080 on this one at some point.

ANyway long way of saying this is really wonderful to get this cleaned up.  A good step on continuing to make nifi more secure, more maintainable, and allows us to amplify and mature features like vault based parameter providers that we did not have when we made decisions that led to this stuff.  

Thanks!

I'm +1 but lets see if other input arrives.

PR #8981, opened 2024-06-19
NIFI-13304 Added SplitExcel capability

dan-s1:
# Summary

[NIFI-13304](https://issues.apache.org/jira/browse/NIFI-13304)

This PR introduces the SplitExcel processor which allows users to split an Excel spreadsheet consisting of many sheets into multiple spreadsheet flow files each containing a single sheet. The code for this PR is for the 2.x branch but this feature is desired by customers (see ticket for more details) on the 1.x branch to assist in their migration from 1.x to 2.x. There would have to be some minor changes made to `SplitExcel` in order to include it on the 1.x branch.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+            .build();
+
+    public static final PropertyDescriptor AVOID_TEMP_FILES = new PropertyDescriptor
+            .Builder().name("Avoid Temp Files")
+            .displayName("Avoid Temp Files")
+            .description("Enables a mode where the code tries to avoid creating temp files. By default, temp files are used to avoid holding onto too much data in memory.")
+            .required(true)
+            .defaultValue(Boolean.FALSE.toString())
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
```
exceptionfactory, line 103: I recommend removing this property since it should always be enabled for general processing.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+                        .rowCacheSize(100)
+                        .bufferSize(4096)
+                        .password(password)
+                        .setAvoidTempFiles(avoidTempFiles)
+                        .setReadHyperlinks(true) // NOTE: Needed for copying rows.
+                        .setReadSharedFormulas(true) // NOTE: If not set to true, then data with shared formulas fail.
+                        .open(in);
+
+                int index = 0;
+                for (Sheet originalSheet : originalWorkbook) {
```
exceptionfactory, line 167:
```suggestion
                for (final Sheet originalSheet : originalWorkbook) {
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+        @WritesAttribute(attribute = "fragment.index", description = "A one-up number that indicates the ordering of the split Excel FlowFiles that were created from a single parent Excel FlowFile"),
+        @WritesAttribute(attribute = "fragment.count", description = "The number of split Excel FlowFiles generated from the parent Excel FlowFile"),
+        @WritesAttribute(attribute = "segment.original.filename", description = "The filename of the parent Excel FlowFile"),
+        @WritesAttribute(attribute = SplitExcel.SHEET_NAME, description = "The name of the Excel sheet from the original spreadsheet."),
+        @WritesAttribute(attribute = SplitExcel.TOTAL_ROWS, description = "The number of rows in the Excel sheet from the original spreadsheet.")})
+public class SplitExcel extends AbstractProcessor {
+    public static final String FRAGMENT_ID = FragmentAttributes.FRAGMENT_ID.key();
+    public static final String FRAGMENT_INDEX = FragmentAttributes.FRAGMENT_INDEX.key();
+    public static final String FRAGMENT_COUNT = FragmentAttributes.FRAGMENT_COUNT.key();
+    public static final String SEGMENT_ORIGINAL_FILENAME = FragmentAttributes.SEGMENT_ORIGINAL_FILENAME.key();
```
exceptionfactory, line 73: I recommend removing these values and instead using static imports for the enums.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+    public static final String SHEET_NAME = "sheetname";
+    public static final String TOTAL_ROWS = "total.rows";
+
+    public static final PropertyDescriptor PROTECTION_TYPE = new PropertyDescriptor
+            .Builder().name("Protection Type")
+            .displayName("Protection Type")
```
exceptionfactory, line 79:
```suggestion
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+            .description("Specifies whether an Excel spreadsheet is protected by a password or not.")
+            .required(true)
+            .allowableValues(ProtectionType.class)
+            .defaultValue(ProtectionType.UNPROTECTED)
+            .build();
+
+    public static final PropertyDescriptor PASSWORD = new PropertyDescriptor
+            .Builder().name("Password")
+            .displayName("Password")
```
exceptionfactory, line 88:
```suggestion
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+public class SplitExcel extends AbstractProcessor {
+    public static final String FRAGMENT_ID = FragmentAttributes.FRAGMENT_ID.key();
+    public static final String FRAGMENT_INDEX = FragmentAttributes.FRAGMENT_INDEX.key();
+    public static final String FRAGMENT_COUNT = FragmentAttributes.FRAGMENT_COUNT.key();
+    public static final String SEGMENT_ORIGINAL_FILENAME = FragmentAttributes.SEGMENT_ORIGINAL_FILENAME.key();
+    public static final String SHEET_NAME = "sheetname";
+    public static final String TOTAL_ROWS = "total.rows";
+
+    public static final PropertyDescriptor PROTECTION_TYPE = new PropertyDescriptor
+            .Builder().name("Protection Type")
```
exceptionfactory, line 78:
```suggestion
    public static final PropertyDescriptor PROTECTION_TYPE = new PropertyDescriptor.Builder()
            .name("Protection Type")
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+            .displayName("Protection Type")
+            .description("Specifies whether an Excel spreadsheet is protected by a password or not.")
+            .required(true)
+            .allowableValues(ProtectionType.class)
+            .defaultValue(ProtectionType.UNPROTECTED)
+            .build();
+
+    public static final PropertyDescriptor PASSWORD = new PropertyDescriptor
+            .Builder().name("Password")
```
exceptionfactory, line 87:
```suggestion
    public static final PropertyDescriptor PASSWORD = new PropertyDescriptor.Builder()
            .name("Password")
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+            originalFileNameExtension = originalFileName.substring(extensionIndex);
+        }
+
+        final Map<String, String> attributes = new HashMap<>();
+        attributes.put(FRAGMENT_COUNT, String.valueOf(workbookSplits.size()));
+        attributes.put(FRAGMENT_ID, fragmentId);
+        attributes.put(SEGMENT_ORIGINAL_FILENAME, originalFileName);
+
+        for (WorkbookSplit split : workbookSplits) {
+            attributes.put(CoreAttributes.FILENAME.key(), originalFileNameWithoutExtension + "-" + split.index() + originalFileNameExtension);
```
exceptionfactory, line 217: Recommend using `String.format()` for readability.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -0,0 +1,239 @@
...
+        final List<FlowFile> flowFileSplits = workbookSplits.stream()
+                .map(WorkbookSplit::content)
+                .toList();
+
+        if (flowFileSplits.size() > 1) {
+            session.transfer(flowFileSplits, REL_SPLIT);
+        } else {
+            getLogger().warn("Original file {} did not have more than one sheet, nothing to split", originalFlowFile);
+            session.remove(flowFileSplits);
+        }
```
exceptionfactory, line 234: I'm not sure this check is necessary, I think it is better to always write the split files, even if there is only one, and avoid the warning.

exceptionfactory (changes requested):
Thanks for the work on this @dan-s1. The general approach looks straightforward, just a few implementation notes.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
dan-s1: I need it for `testPasswordProtected` in `TestSplitExcel`. Should I remove that test?

exceptionfactory: If that is the only way to exercise that test, then yes, I would remove the property and the test method. I know this was an issue for the previous implementation of the password-protection in ExcelReader, but exposing a Processor property just for testing is less than ideal, so removing it sounds like the best option under the circumstances.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
dan-s1: I cut and paste this from `ExcelReader`. Is `displayName` no longer needed?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-processors/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
exceptionfactory: That's correct, if the `displayName()` is the same as `name()`, then `displayName()` is not needed.

exceptionfactory (approved):
Thanks for making the adjustments @dan-s1. I noticed that the test class still creates a temporary password-protected stream, but I will push a commit removing it from the test, and then merge. Thanks for putting this together!

exceptionfactory (changes requested):
@dan-s1 Sorry for the last-minute change, but actually we need another approach. The addition of the `nifi-poi-processors-nar` introduces duplicates of the POI libraries, which are not small. Although Processors and Services should generally be separated, given that this is a narrowly focused NAR, with not other API NAR dependencies, they should be co-located.

Can you make the adjustments to keep both components in the existing `nifi-poi-nar`?

dan-s1:
@exceptionfactory Sure!

exceptionfactory (approved):
Thanks for streamlining the NAR bundling @dan-s1, the latest version looks good. +1 merging

dan-s1:
@exceptionfactory Thanks for merging. Two points.

1. I noticed that the `poi-utils` module is really not necessary. The one enum which it has `ProtectionType` was only moved to this new module if the `SplitExcel` and `ExcelReader` were going to be in two different modules so that both could utilize this for password protected Excel files. Now that they are all under one module `nifi-poi-services`, `ProtectionType` can be moved back there although still remain in its own file.
2. If I would like this change in the 1.x, I assume I have to make a separate PR, right?

exceptionfactory:
@dan-s1 Thanks for following up. I agree with you that `poi-utils` is not really necessary in retrospect, so if you would fold that back in to the `nifi-poi-services` module, that would be a useful separate Jira and PR.

With that change in place, then a separate new PR for potential backporting would be in order.

exceptionfactory:
I would not move the `ExcelReader` since that would break existing flow definitions.

I would either leave `ProtectionType` under `org.apache.nifi.excel`, or just move it to `org.apache.nifi.excel.property`.

PR #8984, opened 2024-06-19
NIFI-13403 Removed the use of toString in logging statements

dan-s1:
# Summary

[NIFI-13403](https://issues.apache.org/jira/browse/NIFI-13403)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/MiNiFi.java
+++ b/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/MiNiFi.java
@@ -75,8 +75,7 @@ public class MiNiFi {
         }
 
         Thread.setDefaultUncaughtExceptionHandler((t, e) -> {
-            logger.error("An Unknown Error Occurred in Thread {}: {}", t, e.toString());
-            logger.error("", e);
+            logger.error("An Unknown Error Occurred in Thread {}", t, e);
         });
 
         // register the shutdown hook
--- a/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/bootstrap/BootstrapListener.java
+++ b/minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/bootstrap/BootstrapListener.java
@@ -257,7 +257,7 @@ public class BootstrapListener implements BootstrapCommunicator {
                 try {
                     socket.close();
                 } catch (IOException ioe) {
-                    logger.warn("Failed to close socket to Bootstrap due to {}", ioe.toString());
+                    logger.warn("Failed to close socket to Bootstrap", ioe);
                 }
             }
         }
--- a/nifi-commons/nifi-site-to-site-client/src/main/java/org/apache/nifi/remote/client/http/HttpClient.java
+++ b/nifi-commons/nifi-site-to-site-client/src/main/java/org/apache/nifi/remote/client/http/HttpClient.java
@@ -181,7 +181,7 @@ public class HttpClient extends AbstractSiteToSiteClient implements PeerStatusPr
                 commSession.setUserDn(apiClient.getTrustedPeerDn());
             } catch (final Exception e) {
                 apiClient.close();
-                logger.warn("Penalizing peer {}", peer, e);
+                logger.warn("Penalizing a peer {} due to {}", peer, e.toString());
                 peerSelector.penalize(peer, penaltyMillis);
 
                 // Following exceptions will be thrown even if we tried other peers, so throw it.
--- a/nifi-commons/nifi-site-to-site-client/src/main/java/org/apache/nifi/remote/client/http/HttpClient.java
+++ b/nifi-commons/nifi-site-to-site-client/src/main/java/org/apache/nifi/remote/client/http/HttpClient.java
@@ -181,7 +181,7 @@ public class HttpClient extends AbstractSiteToSiteClient implements PeerStatusPr
                 commSession.setUserDn(apiClient.getTrustedPeerDn());
             } catch (final Exception e) {
                 apiClient.close();
-                logger.warn("Penalizing a peer {} due to {}", peer, e.toString());
+                logger.warn("Penalizing peer {}", peer, e);
                 peerSelector.penalize(peer, penaltyMillis);
 
                 // Following exceptions will be thrown even if we tried other peers, so throw it.
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/WriteJsonResult.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/WriteJsonResult.java
@@ -250,7 +250,7 @@ public class WriteJsonResult extends AbstractRecordSetWriter implements RecordSe
 
             endTask.apply(generator);
         } catch (final Exception e) {
-            logger.error("Failed to write {} with reader schema {} and writer schema {} as a JSON Object due to {}", record, record.getSchema(), writeSchema, e.toString(), e);
+            logger.error("Failed to write {} with reader schema {} and writer schema {} as a JSON Object", record, record.getSchema(), writeSchema, e);
             throw e;
         }
     }
--- a/nifi-extension-bundles/nifi-groovyx-bundle/nifi-groovyx-processors/src/main/java/org/apache/nifi/processors/groovyx/ExecuteGroovyScript.java
+++ b/nifi-extension-bundles/nifi-groovyx-bundle/nifi-groovyx-processors/src/main/java/org/apache/nifi/processors/groovyx/ExecuteGroovyScript.java
@@ -483,7 +483,7 @@ public class ExecuteGroovyScript extends AbstractProcessor {
             onCommitSQL(SQL);
             session.commitAsync();
         } catch (Throwable t) {
-            getLogger().error(t.toString(), t);
+            getLogger().error("", t);
             onFailSQL(SQL);
             if (toFailureOnError) {
                 //transfer all received to failure with two new attributes: ERROR_MESSAGE and ERROR_STACKTRACE.
--- a/nifi-extension-bundles/nifi-groovyx-bundle/nifi-groovyx-processors/src/main/java/org/apache/nifi/processors/groovyx/ExecuteGroovyScript.java
+++ b/nifi-extension-bundles/nifi-groovyx-bundle/nifi-groovyx-processors/src/main/java/org/apache/nifi/processors/groovyx/ExecuteGroovyScript.java
@@ -483,7 +483,7 @@ public class ExecuteGroovyScript extends AbstractProcessor {
             onCommitSQL(SQL);
             session.commitAsync();
         } catch (Throwable t) {
-            getLogger().error("", t);
+            getLogger().error(t.toString(), t);
             onFailSQL(SQL);
             if (toFailureOnError) {
                 //transfer all received to failure with two new attributes: ERROR_MESSAGE and ERROR_STACKTRACE.
--- a/nifi-extension-bundles/nifi-kudu-bundle/nifi-kudu-processors/src/main/java/org/apache/nifi/processors/kudu/PutKudu.java
+++ b/nifi-extension-bundles/nifi-kudu-bundle/nifi-kudu-processors/src/main/java/org/apache/nifi/processors/kudu/PutKudu.java
@@ -587,7 +587,7 @@ public class PutKudu extends AbstractKuduProcessor {
             final List<RowError> rowErrors = putKuduResult.getRowErrorsForFlowFile(flowFile);
 
             if (rowErrors != null && !rowErrors.isEmpty()) {
-                rowErrors.forEach(rowError -> getLogger().error("Failed to write due to {}", rowError));
+                rowErrors.forEach(rowError -> getLogger().error("Failed to write due to {}", rowError.toString()));
                 flowFile = session.putAttribute(flowFile, RECORD_COUNT_ATTR, Integer.toString(count - rowErrors.size()));
                 totalCount -= rowErrors.size(); // Don't include error rows in the counter.
                 session.transfer(flowFile, REL_FAILURE);
--- a/nifi-extension-bundles/nifi-kudu-bundle/nifi-kudu-processors/src/main/java/org/apache/nifi/processors/kudu/PutKudu.java
+++ b/nifi-extension-bundles/nifi-kudu-bundle/nifi-kudu-processors/src/main/java/org/apache/nifi/processors/kudu/PutKudu.java
@@ -587,7 +587,7 @@ public class PutKudu extends AbstractKuduProcessor {
             final List<RowError> rowErrors = putKuduResult.getRowErrorsForFlowFile(flowFile);
 
             if (rowErrors != null && !rowErrors.isEmpty()) {
-                rowErrors.forEach(rowError -> getLogger().error("Failed to write due to {}", rowError.toString()));
+                rowErrors.forEach(rowError -> getLogger().error("Failed to write due to {}", rowError));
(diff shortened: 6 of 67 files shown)


minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/MiNiFi.java
```diff
@@ -75,8 +75,7 @@ public MiNiFi(final NiFiProperties properties, ClassLoader rootClassLoader) thro
         }
 
         Thread.setDefaultUncaughtExceptionHandler((t, e) -> {
-            logger.error("An Unknown Error Occurred in Thread {}: {}", t, e.toString());
-            logger.error("", e);
+            logger.error("An Unknown Error Occurred in Thread {}", t, e);
```
joewitt, line 78: @dan-s1 I have not tested this but my concern with this change as is would be does it actually change the behavior?  Specifically calling toString on the exception as I recall grabs pretty much the message.  Whereas just passing the exception to the logger will result in the full stack trace.  We don't always want/both.  Please confirm whether this is changing behavior

joewitt:
We should not be changing behavior with this change.  The intent to avoid calling toString as we are so that we can benefit from it being invoked when needed by the logger depending on levels is good.  But changing behavior is not.

If I follow correctly many of these changes would mean instead of giving direct exception messages in cases where it can be useful we'd instead follow the log output with a big stack trace.  In my experience this results in far less useful log output.

I'm not saying we can't improve things but I am saying a blanket removal of 'toString' calls isn't the answer when it comes to error handling vs logging

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/MiNiFi.java
dan-s1: @exceptionfactory had suggested I do this in PRs for [NIFI-13265](https://issues.apache.org/jira/browse/NIFI-13265) and [NIFI-13266](https://issues.apache.org/jira/browse/NIFI-13266). If that is not the desired behavior I certainly can remove this.

joewitt:
I'd be curious to hear what @exceptionfactory has in mind.

But i will say from a user point of view (which was the point of the 'due to - message') mechanism we do want to surface/provide detailed log information and often times these toString outputs for exceptions are more valuable to a user than the exception stack trace.  We generally provided both but notably the message was on the same line as the main log output.  Stacktraces are in a series of lines that follow.  

We need to ensure what we do here is rooted in what gives value to the user as that is more important than the performance savings of 'toString' being invoked or not.

If you can show a before/after that would be helpful.  But please let's make sure a change that is rooted in being about avoiding calls isn't also changing behavior without us considering what that change means for the user.

Thanks

nifi-extension-bundles/nifi-groovyx-bundle/nifi-groovyx-processors/src/main/java/org/apache/nifi/processors/groovyx/ExecuteGroovyScript.java
```diff
@@ -483,7 +483,7 @@ public void onTrigger(final ProcessContext context, final ProcessSession _sessio
             onCommitSQL(SQL);
             session.commitAsync();
         } catch (Throwable t) {
-            getLogger().error(t.toString(), t);
+            getLogger().error("", t);
```
exceptionfactory, line 486: This should be reverted to preserve existing behavior, as opposed to logging an empty string as the message.

nifi-extension-bundles/nifi-kudu-bundle/nifi-kudu-processors/src/main/java/org/apache/nifi/processors/kudu/PutKudu.java
```diff
@@ -587,7 +587,7 @@ private void transferFlowFiles(final List<FlowFile> flowFiles,
             final List<RowError> rowErrors = putKuduResult.getRowErrorsForFlowFile(flowFile);
 
             if (rowErrors != null && !rowErrors.isEmpty()) {
-                rowErrors.forEach(rowError -> getLogger().error("Failed to write due to {}", rowError.toString()));
+                rowErrors.forEach(rowError -> getLogger().error("Failed to write due to {}", rowError));
```
exceptionfactory, line 590: This should be reverted since it could generate a large number of stack traces.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/service/StandardControllerServiceNode.java
```diff
@@ -759,7 +759,7 @@ private void invokeDisable(ConfigurationContext configContext) {
             final Throwable cause = e instanceof InvocationTargetException ? e.getCause() : e;
             final ComponentLog componentLog = new SimpleProcessLogger(getIdentifier(), StandardControllerServiceNode.this, new StandardLoggingContext(StandardControllerServiceNode.this));
             componentLog.error("Failed to invoke @OnDisabled method due to {}", cause);
```
exceptionfactory, line 761: This line should also be changed:
```suggestion
            componentLog.error("Failed to invoke @OnDisabled method", cause);
```

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/NiFiRegistrySecurityConfig.java
```diff
@@ -186,7 +185,7 @@ public void handle(final HttpServletRequest request, final HttpServletResponse r
             if (accessDeniedException instanceof CsrfException) {
                 status = HttpServletResponse.SC_FORBIDDEN;
                 message = "Access Denied: CSRF Header and Cookie not matched";
-                logger.info("Access Denied: CSRF Header [{}] not matched: {}", CsrfCookieName.REQUEST_TOKEN.getCookieName(), accessDeniedException.toString());
+                logger.info("Access Denied: CSRF Header [{}] not matched", CsrfCookieName.REQUEST_TOKEN.getCookieName(), accessDeniedException);
```
exceptionfactory, line 188: This specific change should be reverted since the stack trace is not necessarily useful for this type of error.

exceptionfactory (changes requested):
I went through an initial review and flagged several specific changes that should be reverted.

The general goal of removing unnecessary `toString()` calls is good, but as @joewitt notes, there are a couple places that the specific behavior is desired.

The set of changes cover a couple different issues. I think the majority of the changes are good, but there are several different goals. There are several categories are things changed:

1. Unnecessary `toString()` calls that should be removed so that Logback can perform the operation if needed
2. Unnecessary inclusion of `due to e.getMessage()` when the stack trace is already logged and will already contain the message
3. Unnecessary repetition of the exception message in one call, followed by an empty string and the stack trace at the same level
4. Questionable logging of the exception message, and conditional logging of the stack trace at the debug level
5. Valid logging of `e.getMessage()` without the stack trace included

The first three should be changed, instances of number 4 need careful evaluation, and instances of number 5 should not be changed.

I think the vast majority of the initial change set falls under 1 through 4, but there was at least one instance of 5.

In my initial review, I called out some changes to be reverted, but it would require a second pass to sort out everything. With the general framing, keeping this PR scoped to the first three types of changes would be best. Instances of number 4 need evaluation on a case-by-case basis. Some of the reasons may be historical, and should be changed now that Bulletins include the messages for each exception in the stack trace. Other instances should be retained because excessive stack trace repetition is not useful.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/NiFiRegistrySecurityConfig.java
joewitt: @exceptionfactory those five levels are excellent framing.

Levels 2 and 3 are where I'm focused actually.  The user experience will change if we alter this and part of the reason it is commonly like this is users can 'grep' the logs and see the main log line and the heart of the message.  Exceptions are on different lines and there is no guarantee the exception will print in a subsequent line particularly in busy logs.  Now maybe we're less noise now so perhaps this is ok.

But I want to make sure we're considering the best/most appropriate use case.  Logging the same exception over and over is terrible.  But having the core message of the exception available in context with where the error happened/ and was logged is helpful.

dan-s1:
@exceptionfactory @joewitt 

So with the given categories I tried to break it down per class which places fit into which category.  I tried to identify category 5 places but I would like confirmation they are indeed category 5. Please at your earliest convenience let me know what should be reverted. Thanks!

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">

<html>
<head></head>
<body>

Class | Category | Notes | Remove
-- | -- | -- | --
FlowStatusRunner | 1 |   |  
BootstrapListener | 5 | Is this a category 5? |  
MiNIFI | 3 |   |  
ElasticSearchClientServiceImpl | 1 | Both places |  
AbstractHeartbeatMonitor | 3 |   |  
StandardAsyncClusterResponse | 1 |   |  
ThreadPoolRequestReplicator | 1 and 3 | First place 1 and second place 3 |  
AbstractFlowAnalysisRuleNode | 5 | Is this a category 5? |  
StandardProcessSession | 1 and 2 | First place 1 and second place 2 |  
StandardLogRepository | 5 | Is this a category 5? |  
StandardRemoteProcessGroup | 4 |   |  
NioAsyncLoadBalanceClient | 2 |   |  
SwappablePriorityQueue | 3 | All three places |  
FileSystemRepository | 1, 3 and 4 | First place 3, second place 1, third-ninth places 4 and tenth place 3 |  
StandardProcessScheduler | 3 |   |  
VersionedFlowSynchronizer | 1 |   |  
WriteAheadLocalStateProvider | 3 |   |  
ConnectableTask | 2 and 3 | First place 2 and second place 3 |  
ExpireFlowFiles | 2 |   |  
ExtensionBuilder | 2 |   |  
FlowController | 4 |   |  
SystemDiagnosticsFactory | 4 | All three places |  
StandardExtensionDiscoveringManager | 1 |   |  
CuratorLeaderElectionManager | 4 | Both places |  
AbstractGetGcpVisionAnnotateOperationStatus | 1 |   |  
InMemoryGraphClient | 1 |   |  
DeleteHBaseCells | 1 |   |  
PutHbaseJSON | 2 |   |  
AbstractJsonRowRecordReader | 2 |   |  
WriteJsonResult | 2 |   |  
TinkerpopClientService | 1 | Both places |  
LuceneEventIndex | 3 |   |  
StandardIndexManager | 2 | Both places |  
WriteAheadStorePartition | 5 | Is this a category 5? |  
IndexConfiguration | 3 |   |  
LockedClient | 1 |   |  
TableWriterBasedInsertRowContext | 1 |   |  
LoggingRecordSink | 1 |   |  
GitFlowMetaData | 1 |   |  
LoggingEventHookProvider | 1 |   |  
FileAccessPolicyProvider | 1 |   |  
FileUserGroupProvider | 1 |   |  
NiFiRegistrySecurityConfig | 4 |   |  
Nifi | 2 and 5 | First place is 2 but is second place a 5? |  
SocketRemoteSiteListener | 4 and 5 | First-fourth and seventh places 4, fifth and sixth places 5 |  
StandardRemoteGroupPort | 4 |   |  
HttpClient | 5 | Is this a category 5? |  
PeerSelector | 1 |   |  
FTPTransfer | 2 |   |  
SFTPTransfer | 2 | Both places |  
ConvertJSONToSQL | 2 | Both places |  
FetchFile | 2 |   |  
ListFile | 1 and 2 | First place 1 and second place 2 |  
ReplaceText | 2 |   |  
TailFile | 2 |   |  
ValidateRecord | 1 |   |  
ControllerStatusReportingTask | 1 | All three places |  
RunStatelessFlow | 1 |   |  
LogComponentStatuses | 1 | Both places |  
StatelessProcessSession | 2 |   |  
DtoFactoryTest | 1 | Both places |  
OcspCertificateValidator | 1 | All four places |  
X509AuthenticationProvider | 1 |   |  
LengthDelimitedJournal | 2 |   |  


</body>

</html>

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/MiNiFi.java
exceptionfactory: In this particular case, the behavior is not changed because the stack trace was already logged in a separate line, so this change looks good.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/bootstrap/BootstrapListener.java
```diff
@@ -257,7 +257,7 @@ private void handleBootstrapRequest(Socket socket) {
                 try {
                     socket.close();
                 } catch (IOException ioe) {
-                    logger.warn("Failed to close socket to Bootstrap due to {}", ioe.toString());
+                    logger.warn("Failed to close socket to Bootstrap", ioe);
```
exceptionfactory, line 260: This is a slight behavior change, but I believe it is reasonable since closing the socket and failing can have different causes that would be useful to know.

nifi-commons/nifi-site-to-site-client/src/main/java/org/apache/nifi/remote/client/http/HttpClient.java
```diff
@@ -181,7 +181,7 @@ public Transaction createTransaction(final TransferDirection direction) throws I
                 commSession.setUserDn(apiClient.getTrustedPeerDn());
             } catch (final Exception e) {
                 apiClient.close();
-                logger.warn("Penalizing a peer {} due to {}", peer, e.toString());
+                logger.warn("Penalizing peer {}", peer, e);
```
exceptionfactory, line 184: This particular change should be reverted due to the potential for excessive stack trace logging.

nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/WriteJsonResult.java
```diff
@@ -250,7 +250,7 @@ private void writeRecord(final Record record, final RecordSchema writeSchema, fi
 
             endTask.apply(generator);
         } catch (final Exception e) {
-            logger.error("Failed to write {} with reader schema {} and writer schema {} as a JSON Object due to {}", record, record.getSchema(), writeSchema, e.toString(), e);
+            logger.error("Failed to write {} with reader schema {} and writer schema {} as a JSON Object", record, record.getSchema(), writeSchema, e);
```
exceptionfactory, line 253: This is a legitimate change since the stack trace was already being logged.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/NiFiRegistrySecurityConfig.java
```diff
@@ -166,8 +166,7 @@ private AuthenticationEntryPoint http401AuthenticationEntryPoint() {
 
             // return a 401 response
             final int status = HttpServletResponse.SC_UNAUTHORIZED;
-            logger.info("Client could not be authenticated due to: {} Returning 401 response.", authenticationException.toString());
```
exceptionfactory, line 169: Recommend reverting this change, but adding a message to the `debug` call along the lines of `debug("HTTP 401 Unauthorized", authenticationException)`

exceptionfactory (changes requested):
@dan-s1, I reviewed the changes, thanks for the summary. I recommended a couple places where the changes should be reverted or modified. Otherwise, I think the current approach moves things in the right direction.

exceptionfactory (approved):
Thanks for the updates @dan-s1, looks good! +1 merging

PR #8987, opened 2024-06-20
NIFI-13416: Inherited Parameter Table Improvements

mcgilman:
NIFI-13416:
- Allowing new parameters to override inherited parameters.
- Adding a new Override action to prepopulate the parameter name and sensitivity.
- Showing an indicator when a parameter in the listing is inherited.


rfellows:
Reviewing...

rfellows (approved):
This works well. Thanks @mcgilman 👍

PR #8992, opened 2024-06-21
NIFI-13430 Added CopyS3Object and ExistsS3Object

MikeThomsen:
# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for proposing these new Processors @MikeThomsen. The capabilities look useful.

On an initial review, I noticed that these Processors are using version 1 of the AWS SDK. There has been significant effort to move to AWS SDK 2, and I believe the S3 Processors are the remaining components to be migrated.

With that in mind, I do not think we should introduce new code that depends on AWS SDK 1, so if you can rework them using SDK 2, that would be helpful for moving things forward.

MikeThomsen:
@exceptionfactory Alright. I'll open a PR for this functionality on support/nifi-1.x and then start refactoring this PR to use AWS 2.X APIs. Does that work for you?

exceptionfactory:
> @exceptionfactory Alright. I'll open a PR for this functionality on support/nifi-1.x and then start refactoring this PR to use AWS 2.X APIs. Does that work for you?

I recommend refactoring first, as the AWS SDK 2 libraries are also available on the support branch. Working through the changes on the main branch first should limit the amount of work necessary to consider back-porting.

MikeThomsen:
Ok. Will do. I will probably have to massively expand the scope of this PR but it is what it is...

MikeThomsen:
@exceptionfactory I did some experimentation with this today, and there is going to be a lot of refactoring work in both the abstract processors module and the aws processors module to get the S3 stuff migrated. I'd rather add these with the V1 API since they're thin wrappers over that API and then do a bigger ticket afterward to try to get everything on the V2 API.

exceptionfactory:
Thanks for the follow up @MikeThomsen.

Incorporating new Processors and refactoring in the same pull request for S3 doesn't sound ideal, and I agree that the scope of these Processors is fairly narrow in terms of S3 API changes. With that in mind, considering these Processors as they stand seems like a reasonable way forward.

To that end, there are a few things I recommend evaluating. The Copy Processor seems straightforward, and I can take a closer look at potential changes there. I would note that new properties should just use a human-readable `name()` instead of a different `displayName()` going forward.

On the Exists Processor, I think it would be better to have it function as a `GetS3ObjectMetadata` Processor. The `doesObjectExist()` method requests the metadata, but ignores the response. I think having the metadata would present other potential uses beyond just whether the object exists, while still supporting that particular use case.

MikeThomsen:
I'll take a look into that @joewitt @exceptionfactory but it might be right to have 3 processors instead. I recall something about how doesObjectExists uses a HEAD which is faster for S3 than trying to fetch the metadata when your only goal is to ask whether the object exists. I'll look into it and adjust accordingly.

exceptionfactory:
> I'll take a look into that @joewitt @exceptionfactory but it might be right to have 3 processors instead. I recall something about how doesObjectExists uses a HEAD which is faster for S3 than trying to fetch the metadata when your only goal is to ask whether the object exists. I'll look into it and adjust accordingly.

@MikeThomsen The `doesObjectExist()` method calls `getObjectMetadata()`, so there should be no difference in performance. That's the reason I suggested going that direction, thus I would not recommend a separate processor for checking object existence.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.nifi.processor.exception.ProcessException;
+
+import java.util.List;
+
+import static org.apache.nifi.processors.aws.util.RegionUtilV1.S3_REGION;
+
+@Tags({"Amazon", "S3", "AWS", "Archive", "Copy"})
+@InputRequirement(InputRequirement.Requirement.INPUT_REQUIRED)
+@CapabilityDescription("Copies a file from one bucket and key to another in AWS S3")
+@SeeAlso({PutS3Object.class, DeleteS3Object.class, ListS3.class, TagS3Object.class, DeleteS3Object.class, FetchS3Object.class})
```
jrsteinebrey, line 40: I suggest you look at the @WritesAttribute annotations of other S3Object processors (especially ListS3 and PutS3Object) and see if any of them would be worthwhile to add CopyS3Object or ExistsS3Object or the new GetS3ObjectMetadata that has been talked about.  I think the exception related ones are worth adding, but there might be some others that are worth adding.

jrsteinebrey (changes requested):

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,145 @@
...
+            if (cannedAccessControlList != null) {
+                request.setCannedAccessControlList(cannedAccessControlList);
+            }
+
+            s3.copyObject(request);
+            session.getProvenanceReporter().send(flowFile, getTransitUrl(targetBucket, targetKey));
+
+            session.transfer(flowFile, REL_SUCCESS);
+        } catch (Exception ex) {
+            getLogger().error("Copy S3 Object Request failed with error:", ex);
```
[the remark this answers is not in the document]
joewitt: Are the exception details you mention a stack trace or a generally user beneficial string describing what went wrong?  We need to avoid large attributes being carried around - they're not designed for that.

If we're going to put error related attributes on we also need to be thoughtful about clearing them out when the success path occurs for the same component that would have put them on.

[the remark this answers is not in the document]
jrsteinebrey: Thanks, @joewitt for raising valid points to consider. I feel the attributes are beneficial and relatively small with no stack traces as you can inspect in the code below.
FYI, the method I suggested is used in all 5 of the existing S3Object related processors. 
@joewitt Are you OK with the new processors using the extractExceptionDetails() like the existing S3 processors do?

This is the method that adds exception related attributes.
`protected FlowFile extractExceptionDetails(final Exception e, final ProcessSession session, FlowFile flowFile) {
        flowFile = session.putAttribute(flowFile, "s3.exception", e.getClass().getName());
        if (e instanceof AmazonS3Exception) {
            flowFile = putAttribute(session, flowFile, "s3.additionalDetails", ((AmazonS3Exception) e).getAdditionalDetails());
        }
        if (e instanceof final AmazonServiceException ase) {
            flowFile = putAttribute(session, flowFile, "s3.statusCode", ase.getStatusCode());
            flowFile = putAttribute(session, flowFile, "s3.errorCode", ase.getErrorCode());
            flowFile = putAttribute(session, flowFile, "s3.errorMessage", ase.getErrorMessage());
        }
        return flowFile;
    }
`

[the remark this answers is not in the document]
jrsteinebrey: @joewitt I will add a new ticket for AbstractS3Processor class to clear out exception related attributes that it writes in its extractExceptionDetails() method to resolve your point about "when the success path occurs for the same component that would have put them on."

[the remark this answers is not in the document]
joewitt: thanks.  In general we have uses of attributes that are increasingly specific to a point in time in the flow/a particular component.  While this pattern can be useful for giving users some added context and not having dive into logs to find it we must also then be equally motivated to clean these values up when they're no longer relevant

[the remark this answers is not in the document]
joewitt: As for being ok with 'extract exception details' I am not familiar with precisely what is in the output.  Provided what it is would be a rather short and meant for more like end user reading then sure.  If it is really developer level info then this is often not helpful to end users and/or can be far too large.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
MikeThomsen: @jrsteinebrey good catch. I just pushed some updates for changes requested above, but will add this too.

MikeThomsen:
@jrsteinebrey @exceptionfactory we should be GTG now on the changes.

exceptionfactory:
> @jrsteinebrey @exceptionfactory we should be GTG now on the changes.

Thanks for the updates @MikeThomsen, I will take a closer look soon.

MikeThomsen:
@exceptionfactory any idea when you can get to this?

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,147 @@
...
+    public static final PropertyDescriptor SOURCE_BUCKET = new PropertyDescriptor.Builder()
+            .fromPropertyDescriptor(BUCKET_WITH_DEFAULT_VALUE)
+            .name("copy-s3-object-source-bucket")
+            .displayName("Source Bucket")
+            .description("The bucket that contains the file to be copied.")
+            .build();
+    public static final PropertyDescriptor TARGET_BUCKET = new PropertyDescriptor.Builder()
+            .fromPropertyDescriptor(BUCKET_WITHOUT_DEFAULT_VALUE)
+            .name("copy-s3-object-target-bucket")
+            .displayName("Target Bucket")
```
exceptionfactory, line 52: Recommend using `Destination` instead of `Target`:
```suggestion
    public static final PropertyDescriptor DESTINATION_BUCKET = new PropertyDescriptor.Builder()
            .fromPropertyDescriptor(BUCKET_WITHOUT_DEFAULT_VALUE)
            .name("Destination Bucket")
            .displayName("Destination Bucket")
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,147 @@
...
+            .description("The bucket that will receive the copy.")
+            .build();
+
+    public static final PropertyDescriptor SOURCE_KEY = new PropertyDescriptor.Builder()
+            .fromPropertyDescriptor(KEY)
+            .name("copy-s3-source-key")
+            .displayName("Source Key")
```
exceptionfactory, line 59:
```suggestion
            .name("Source Key")
            .displayName("Source Key")
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,147 @@
...
+            .description("The source key in the source bucket")
+            .build();
+
+    public static final PropertyDescriptor TARGET_KEY = new PropertyDescriptor.Builder()
+            .fromPropertyDescriptor(KEY)
+            .name("copy-s3-target-key")
+            .displayName("Target Key")
```
exceptionfactory, line 66:
```suggestion
    public static final PropertyDescriptor DESTINATION_KEY = new PropertyDescriptor.Builder()
            .fromPropertyDescriptor(KEY)
            .name("Destination Key")
            .displayName("Destination Key")
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,147 @@
...
+                request.setCannedAccessControlList(cannedAccessControlList);
+            }
+
+            s3.copyObject(request);
+            session.getProvenanceReporter().send(flowFile, getTransitUrl(targetBucket, targetKey));
+
+            session.transfer(flowFile, REL_SUCCESS);
+        } catch (AmazonClientException ex) {
+            flowFile = extractExceptionDetails(ex, session, flowFile);
+            getLogger().error("Copy S3 Object Request failed with error:", ex);
```
exceptionfactory, line 138:
```suggestion
            getLogger().error("Failed to copy S3 object", e);
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,147 @@
...
+
+            if (cannedAccessControlList != null) {
+                request.setCannedAccessControlList(cannedAccessControlList);
+            }
+
+            s3.copyObject(request);
+            session.getProvenanceReporter().send(flowFile, getTransitUrl(targetBucket, targetKey));
+
+            session.transfer(flowFile, REL_SUCCESS);
+        } catch (AmazonClientException ex) {
```
exceptionfactory, line 136:
```suggestion
        } catch (final AmazonClientException e) {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+@CapabilityDescription("Check for the existence of a file in S3 without attempting to download it. This processor can be " +
+        "used as a router for work flows that need to check on a file in S3 before proceeding with data processing")
+@SeeAlso({PutS3Object.class, DeleteS3Object.class, ListS3.class, TagS3Object.class, DeleteS3Object.class, FetchS3Object.class})
+public class GetS3ObjectMetadata extends AbstractS3Processor {
+    public static final AllowableValue MODE_FETCH_METADATA = new AllowableValue("fetch", "Fetch Metadata",
+            "This is the default mode. It will fetch the metadata and write it to either the flowfile body or an " +
+                    "attribute");
+    public static final AllowableValue MODE_ROUTER = new AllowableValue("router", "Router", "When selected," +
+            "this mode will skip writing the metadata and just send the flowfile to the found or not-found relationship. It should be used " +
+            "when the goal is to just route flowfiles based on whether or not a key is present in S3.");
```
exceptionfactory, line 60: Recommend replacing these allowable values with an `enum` that implements `DescribedValue`.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+import static org.apache.nifi.processors.aws.util.RegionUtilV1.S3_REGION;
+
+@Tags({"Amazon", "S3", "AWS", "Archive", "Exists"})
+@InputRequirement(InputRequirement.Requirement.INPUT_REQUIRED)
+@CapabilityDescription("Check for the existence of a file in S3 without attempting to download it. This processor can be " +
+        "used as a router for work flows that need to check on a file in S3 before proceeding with data processing")
+@SeeAlso({PutS3Object.class, DeleteS3Object.class, ListS3.class, TagS3Object.class, DeleteS3Object.class, FetchS3Object.class})
+public class GetS3ObjectMetadata extends AbstractS3Processor {
+    public static final AllowableValue MODE_FETCH_METADATA = new AllowableValue("fetch", "Fetch Metadata",
+            "This is the default mode. It will fetch the metadata and write it to either the flowfile body or an " +
```
exceptionfactory, line 56: The description should not mention whether it is the default as that fact is indicated through the PropertyDescriptor.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+    public static final PropertyDescriptor METADATA_ATTRIBUTE = new PropertyDescriptor.Builder()
+            .name("Metadata Attribute")
+            .description("The attribute where the metadata will be written")
+            .addValidator(StandardValidators.NON_EMPTY_EL_VALIDATOR)
+            .defaultValue("s3.object.metadata")
+            .dependsOn(METADATA_TARGET, TARGET_ATTRIBUTE)
+            .required(true)
+            .build();
+
+    public static final ObjectMapper MAPPER = new ObjectMapper();
```
exceptionfactory, line 94:
```suggestion
    private static final ObjectMapper MAPPER = new ObjectMapper();
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+                    route = REL_NOT_FOUND;
+                    flowFile = extractExceptionDetails(e, session, flowFile);
+                } else {
+                    throw e;
+                }
+            }
+
+            session.transfer(flowFile, route);
+        } catch (IOException | AmazonClientException ex) {
+            getLogger().error("There was a problem checking for " + String.format("s3://%s%s", bucket, key), ex);
```
exceptionfactory, line 187: Log statements should use placeholders instead of concatenation:
```suggestion
            getLogger().error("Failed to get S3 Object Metadata from s3://{}{} ", bucket, key, e);
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+                if (e.getStatusCode() == 404) {
+                    route = REL_NOT_FOUND;
+                    flowFile = extractExceptionDetails(e, session, flowFile);
+                } else {
+                    throw e;
+                }
+            }
+
+            session.transfer(flowFile, route);
+        } catch (IOException | AmazonClientException ex) {
```
exceptionfactory, line 186:
```suggestion
        } catch (final IOException | AmazonClientException e) {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+                combinedMetadata.putAll(metadata.getUserMetadata());
+
+                String metadataJson = MAPPER.writeValueAsString(combinedMetadata);
+
+                if (!isRouter && context.getProperty(METADATA_TARGET).getValue().equals(TARGET_ATTRIBUTE.getValue())) {
+                    String attribute = context.getProperty(METADATA_ATTRIBUTE).getValue();
+                    flowFile = session.putAttribute(flowFile, attribute, metadataJson);
+                } else if (!isRouter && context.getProperty(METADATA_TARGET).getValue().equals(TARGET_FLOWFILE_BODY.getValue())) {
+                    flowFile = session.write(flowFile, os -> os.write(metadataJson.getBytes(StandardCharsets.UTF_8)));
+                }
```
exceptionfactory, line 173: Rather than having the `Mode` property, it seems like a more descriptive approach would be something like `Metadata Output Strategy` with either `Attributes` or `Content`.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+
+        try {
+            Relationship route;
+
+            try {
+                ObjectMetadata metadata = s3.getObjectMetadata(bucket, key);
+                Map<String, Object> combinedMetadata = new HashMap<>(metadata.getRawMetadata());
+                combinedMetadata.putAll(metadata.getUserMetadata());
+
+                String metadataJson = MAPPER.writeValueAsString(combinedMetadata);
```
exceptionfactory, line 166: Rather than turning the entire metadata map into a JSON string, what about setting each metadata key as the attribute key, and then just turning the attribute value into a String. We could consider an Attributes JSON strategy as well, but mapping the metadata directly to attributes has some benefits.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+        "used as a router for work flows that need to check on a file in S3 before proceeding with data processing")
+@SeeAlso({PutS3Object.class, DeleteS3Object.class, ListS3.class, TagS3Object.class, DeleteS3Object.class, FetchS3Object.class})
+public class GetS3ObjectMetadata extends AbstractS3Processor {
+    public static final AllowableValue MODE_FETCH_METADATA = new AllowableValue("fetch", "Fetch Metadata",
+            "This is the default mode. It will fetch the metadata and write it to either the flowfile body or an " +
+                    "attribute");
+    public static final AllowableValue MODE_ROUTER = new AllowableValue("router", "Router", "When selected," +
+            "this mode will skip writing the metadata and just send the flowfile to the found or not-found relationship. It should be used " +
+            "when the goal is to just route flowfiles based on whether or not a key is present in S3.");
+    public static final PropertyDescriptor MODE = new PropertyDescriptor.Builder()
```
exceptionfactory, line 61: As noted in the implementation details, rather than this type of `Mode` property, I recommend a `Metadata Output Strategy` property.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/s3/TestCopyS3Object.java
```diff
@@ -0,0 +1,104 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class TestCopyS3Object {
+    private TestRunner runner = null;
+    private CopyS3Object mockCopyS3Object = null;
+    private AmazonS3Client mockS3Client = null;
```
exceptionfactory, line 48: These `null` initial values are not necessary.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/s3/TestCopyS3Object.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @DisplayName("Validate that S3 errors cleanly route to failure")
+    @Test
+    public void testS3ErrorHandling() {
+        var ex = new AmazonS3Exception("Manually triggered error");
+        ex.setStatusCode(503);
+        when(mockS3Client.copyObject(any(CopyObjectRequest.class)))
+                .thenThrow(ex);
+
+        runner.enqueue("".getBytes(StandardCharsets.UTF_8), setupRun());
```
exceptionfactory, line 87:
```suggestion
        runner.enqueue(new byte[]{}, setupRun());
```

exceptionfactory (changes requested):
@MikeThomsen Thanks for introducing the new processors and providing the `GetS3ObjectMetadata` Processor. I provided some detailed feedback, and recommended a different approach on how to name the operational strategy for the Processor.

MikeThomsen:
Thanks @exceptionfactory. I'll try to get a fast turn around on this

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/s3/TestCopyS3Object.java
MikeThomsen: Done

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
MikeThomsen: Done

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
MikeThomsen: I like this approach, but I'm going to go with 3 values: router, write to attribute, write to content because I'd like to give the option to have this function as a light weight router that does absolutely nothing to the flowfile except route it based on whether a file made it into S3.

exceptionfactory (changes requested):
@MikeThomsen the automated builds are failing since just accepting the changes in GitHub does not update all references.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+            flowFile = session.penalize(flowFile);
+            session.transfer(flowFile, REL_FAILURE);
+            return;
+        }
+
+        final String bucket = context.getProperty(BUCKET_WITH_DEFAULT_VALUE).evaluateAttributeExpressions(flowFile).getValue();
+        final String key = context.getProperty(KEY).evaluateAttributeExpressions(flowFile).getValue();
+
+        try {
+            Relationship route;
```
exceptionfactory, line 159: This should be named `relationship`.
```suggestion
            Relationship relationship;
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
jrsteinebrey: @MikeThomsen I totally agree with @exceptionfactory that it is more useful to store the metadata in separate attributes.

Here is a sample S3 metadata JSON that I found.
```
{
    "DocumentId": "S3 document ID, the S3 path to doc",
    "Attributes": {
        "_category": "document category",
        "_created_at": "ISO 8601 encoded string",
        "_last_updated_at": "ISO 8601 encoded string",
        "_source_uri": "document URI",
        "_version": "file version",
        "_view_count": number of times document has been viewed,
        "custom attribute key": "custom attribute value",
        additional custom attributes
    },
    "AccessControlList": [
         {
             "Name": "user name",
             "Type": "GROUP | USER",
             "Access": "ALLOW | DENY"
         }
    ],
    "Title": "document title",
    "ContentType": "For example HTML | PDF. For supported content types, see Types of documents."
}
```

jrsteinebrey: I think each of the fields under Attributes would be useful to store as separate attributes.
I mean Attributes._category (or just _category if you prefer and if you do not think the subfield names will collide with top level names DocumentId, Title, etc) for the attribute name and so on.
I would store the AccessControlList as a single attribute with its JSON array as the attribute value because I predict that info will be used less often.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,192 @@
...
+            .defaultValue(MODE_FETCH_METADATA.getValue())
+            .required(true)
+            .description("Configure the mode of operation for this processor")
+            .addValidator(Validator.VALID)
+            .build();
+
+    public static final AllowableValue TARGET_ATTRIBUTE = new AllowableValue("attribute", "Attribute", "When " +
+            "selected, the metadata will be written to a user-supplied attribute");
+    public static final AllowableValue TARGET_FLOWFILE_BODY = new AllowableValue("flowfile-body", "Flowfile Body", "Write " +
+            "the metadata to the flowfile's body");
```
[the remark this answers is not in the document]
MikeThomsen: Done

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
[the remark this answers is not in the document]
MikeThomsen: Done

MikeThomsen:
@exceptionfactory  @jrsteinebrey Changes should be good now. I changed the "attribute" configuration option to be an "attribute prefix" option. I don't think this version of the AWS Java SDK supports retrieving the ACL from the object metadata, but both raw and user-provided metadata is attached at the user-specified prefix.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,196 @@
...
+            .name("Mode")
+            .allowableValues(MODE_FETCH_METADATA, MODE_ROUTER)
+            .defaultValue(MODE_FETCH_METADATA.getValue())
+            .required(true)
+            .description("Configure the mode of operation for this processor")
+            .addValidator(Validator.VALID)
+            .build();
+
+    public static final AllowableValue TARGET_ATTRIBUTE = new AllowableValue("attribute", "Attribute", "When " +
+            "selected, the metadata will be written to a user-supplied attribute");
```
jrsteinebrey, line 72: @MikeThomsen Thanks again for all your work on these.
Could you please make these plural attributes because that will communicate it more clearly.
TARGET_ATTRIBUTE   change to  TARGET_ATTRIBUTES
AllowableValue("attribute", "Attribute",   change to    AllowableValue("attributes", "Attributes",

OLD:  metadata will be written to a user-supplied attribute   would be clearer to change to 
NEW: metadata will be written to attributes beginning with the user-supplied prefix

jrsteinebrey: @MikeThomsen That is fine is ACL is not returned in that version of AWS Java SDK. That was just the sample I found.
Could you post in here an example of the Metadata you do get back from that AWS Java SDK?
Also could you post the separate FF attributes names and values that would be created for the sample given how the code is implemented now?

MikeThomsen: I just ran a test with it on my employer's account and these are attributes that were added:

* Accept-Ranges
* Content-Length
* Content-Type
* ETag
* Last-Modified
* x-amz-server-side-encryption
* x-amz-version-id

Then the metadata I added manually to the upload:
* a = b
* x = y

Each of those was added to the flowfile as an attribute w/ a prefix of `s3.`

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,196 @@
...
+    public static final AllowableValue TARGET_FLOWFILE_BODY = new AllowableValue("flowfile-content", "FlowFile Body", "Write " +
```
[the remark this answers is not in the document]
MikeThomsen: Should be good now.

MikeThomsen:
@exceptionfactory @jrsteinebrey should be good now.

jrsteinebrey:
@MikeThomsen Thanks again for these new processors. I am fine with them being merged if the workflow actions succeed (they are still running now) and if @exceptionfactory (or any committer or higher) approves of it being merged.

exceptionfactory (changes requested):
Thanks for the updates @MikeThomsen, and thanks for the additional feedback @jrsteinebrey.

This looks close to completion, but I still have a question about the `Mode` property. The property name itself is not the most clear, but taking a step back, is the goal to be able to check for the existence of the object in S3 and route, without adding any FlowFile attributes? If this is the case, what do you think about removing mode, and instead having a `Metadata Include Pattern` property? This could default to `.*`, to include all Metadata keys, but could be adjusted as need to limit the FlowFile attributes produced.

As it stands, the `not found` route adds details as attributes from S3, so it seems like the `found` route should also some attributes.

MikeThomsen:
@exceptionfactory I played around a bit with that and liked the idea. Should be GTG now.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -166,14 +168,22 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
...
+                if (context.getProperty(METADATA_TARGET).getValue().equals(TARGET_ATTRIBUTES.getValue())) {
                     String attributePrefix = context.getProperty(METADATA_ATTRIBUTE_PREFIX).getValue();
                     Map<String, String> newAttributes = combinedMetadata
                             .entrySet().stream()
+                            .filter(e -> {
+                                if (attributePattern == null) {
+                                    return true;
+                                } else {
+                                    return attributePattern.matcher(attributePrefix + e.getKey())
+                                            .find();
```
exceptionfactory, line 180: It seems like the match should only consider `e.getKey()`, and the prefix should be added after, as it is now.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -94,12 +80,23 @@ public class GetS3ObjectMetadata extends AbstractS3Processor {
             .required(true)
             .build();
 
+    public static final PropertyDescriptor ATTRIBUTE_INCLUDE_PATTERN = new PropertyDescriptor.Builder()
+            .name("Attribute Include Pattern")
```
exceptionfactory, line 84: For consistency, naming this `Metadata Include Pattern` would help indicate that it used to match against to the S3 Metadata Key, as opposed to the FlowFile Attribute. Or to cover both bases, `Metadata Attribute Include Pattern`.

exceptionfactory (changes requested):
Thanks for making the adjustments @MikeThomsen, the updated approach looks good!

One more detail, it appears that the filter should focus on the metadata key itself, without the attribute prefix.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
jrsteinebrey: @MikeThomsen That looks good. Thanks for sharing that.

MikeThomsen:
@exceptionfactory should be good now.

MikeThomsen:
@exceptionfactory do you think you'll have time to look at this today?

exceptionfactory:
Thanks for the updates @MikeThomsen, I should have time to review again within the next few days. At a quick glance, the latest code changes look good, so it looks like it should be close to completion following some runtime testing.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,147 @@
...
+            getLogger().error("Failed to initialize S3 client", e);
+            flowFile = session.penalize(flowFile);
+            session.transfer(flowFile, REL_FAILURE);
+            return;
+        }
+
+        final String sourceBucket = context.getProperty(SOURCE_BUCKET).evaluateAttributeExpressions(flowFile).getValue();
+        final String sourceKey = context.getProperty(SOURCE_KEY).evaluateAttributeExpressions(flowFile).getValue();
+        final String targetBucket = context.getProperty(DESTINATION_BUCKET).evaluateAttributeExpressions(flowFile).getValue();
+        final String targetKey = context.getProperty(DESTINATION_KEY).evaluateAttributeExpressions(flowFile).getValue();
```
exceptionfactory, line 118:
```suggestion
        final String destinationBucket = context.getProperty(DESTINATION_BUCKET).evaluateAttributeExpressions(flowFile).getValue();
        final String destinationKey = context.getProperty(DESTINATION_KEY).evaluateAttributeExpressions(flowFile).getValue();
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,147 @@
...
+
+            session.transfer(flowFile, REL_SUCCESS);
+        } catch (final AmazonClientException e) {
+            flowFile = extractExceptionDetails(e, session, flowFile);
+            getLogger().error("Failed to copy S3 object", e);
+            session.transfer(flowFile, REL_FAILURE);
+        }
+    }
+
+    private String getTransitUrl(String targetBucket, String targetKey) {
```
exceptionfactory, line 143:
```suggestion
    private String getTransitUrl(String destinationBucket, String destinationKey) {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/CopyS3Object.java
```diff
@@ -0,0 +1,147 @@
...
+                request.setCannedAccessControlList(cannedAccessControlList);
+            }
+
+            s3.copyObject(request);
+            session.getProvenanceReporter().send(flowFile, getTransitUrl(targetBucket, targetKey));
+
+            session.transfer(flowFile, REL_SUCCESS);
+        } catch (final AmazonClientException e) {
+            flowFile = extractExceptionDetails(e, session, flowFile);
+            getLogger().error("Failed to copy S3 object", e);
```
exceptionfactory, line 138: Recommend including the source bucket and key:
```suggestion
            getLogger().error("Failed to copy S3 object from Bucket [{}] Key [{}]", sourceBucket, sourceKey, e);
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,208 @@
...
+                            .entrySet().stream()
+                            .filter(e -> {
+                                if (attributePattern == null) {
+                                    return true;
+                                } else {
+                                    return attributePattern.matcher(e.getKey())
+                                            .find();
+                                }
+                            })
+                            .collect(Collectors.toMap(e -> attributePrefix + e.getKey(), e -> e.getValue().toString()));
```
exceptionfactory, line 183: The `toString()` approach works for most cases, but it is not optimal for `Date` objects due to the behavior of `java.util.Date.toString()`. Recommend checking for `Date` instances and using the time in seconds.
```suggestion
                            .collect(Collectors.toMap(e -> attributePrefix + e.getKey(), e -> {
                                final Object value = e.getValue();
                                final String attributeValue;
                                if (value instanceof Date dateValue) {
                                    attributeValue = Long.toString(dateValue.getTime());
                                } else {
                                    attributeValue = value.toString();
                                }
                                return attributeValue;
                            }));
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
```diff
@@ -0,0 +1,208 @@
...
+            .build();
+
+    public static final PropertyDescriptor METADATA_ATTRIBUTE_PREFIX = new PropertyDescriptor.Builder()
+            .name("Metadata Attribute Prefix")
+            .description("The prefix for FlowFile attributes generated from the S3 object metadata.")
+            .addValidator(StandardValidators.NON_EMPTY_EL_VALIDATOR)
+            .defaultValue("s3.")
+            .dependsOn(METADATA_TARGET, TARGET_ATTRIBUTES)
+            .required(true)
+            .build();
```
exceptionfactory, line 81: Based on the fact that `ListS3` always uses the `s3.` prefix, this seems unnecessary as a configurable property.

exceptionfactory (changes requested):
Runtime testing of the latest version looks good.

I noted one more implementation detail, to specifically call out `Date` metadata values and convert those to milliseconds.

I also noted that the Attribute Prefix seems unnecessary in general, since `s3.` is a prefix we are already using in ListS3. It might be useful in some scenarios, but this seems like a case where less is more in terms of configuration.

Otherwise, this looks just about ready to go.

MikeThomsen:
Thanks @exceptionfactory. I'll try to get this completed tonight.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/s3/GetS3ObjectMetadata.java
MikeThomsen: Done.

MikeThomsen:
@exceptionfactory done and build started.

exceptionfactory (approved):
Thanks for working through the feedback @MikeThomsen, the latest version looks good! +1 merging

PR #9008, opened 2024-06-26
NIFI-13454: Renaming FileParameterProvider to KubernetesSecretParamet…

gresockj:
…erProvider

# Summary

[NIFI-13454](https://issues.apache.org/jira/browse/NIFI-13454)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/KubernetesSecretParameterProvider.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/KubernetesSecretParameterProvider.java
@@ -34,6 +34,7 @@ import org.apache.nifi.processor.DataUnit;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.apache.nifi.stream.io.LimitingInputStream;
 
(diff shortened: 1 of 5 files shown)


exceptionfactory (changes requested):
Thanks for the update @gresockj. Although renaming may help, it would also break existing deployments.

Instead of renaming the class, I recommend limiting the change to updating the Capability Description, which provides helpful background on the intended purpose.

joewitt:
I share the view that this is a wise move as it is not otherwise clear this is specifically intended for K8S based deployment scenarios.  I think we should ensure this ends up in the migration guidance but I think this is far more intentional about its purpose.

exceptionfactory:
Thanks for the additional perspective @joewitt. I'm fine with renaming the class, we can just note the change in a migration guide step.

gresockj:
> Thanks for the additional perspective @joewitt. I'm fine with renaming the class, we can just note the change in a migration guide step.

Thanks, @exceptionfactory.  Are there any additional changes needed on this?

exceptionfactory (approved):
Thanks @gresockj, the current version looks good! +1 merging, and I will add a note to the migration guide when I prepare the 2.0.0-M4 release candidate build.

PR #9015, opened 2024-06-27
NIFI-13461 Add processor DeleteFile (1.x backport)

EndzeitBegins:
# Summary

Backport of [NIFI-12880](https://issues.apache.org/jira/browse/NIFI-12880) / PR #8489 for 1.x.

See [NIFI-13461](https://issues.apache.org/jira/browse/NIFI-13461)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

- [y] Pull Request refers to a feature branch with one commit containing changes

# Verification

### Build

  - [x] JDK ~21~ 8

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/DeleteFile.java
+++ b/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/DeleteFile.java
@@ -1,0 +1,183 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.standard;
+
+import org.apache.nifi.annotation.behavior.DefaultRunDuration;
+import org.apache.nifi.annotation.behavior.InputRequirement;
+import org.apache.nifi.annotation.behavior.Restricted;
+import org.apache.nifi.annotation.behavior.Restriction;
+import org.apache.nifi.annotation.behavior.SupportsBatching;
+import org.apache.nifi.annotation.behavior.WritesAttribute;
+import org.apache.nifi.annotation.behavior.WritesAttributes;
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.RequiredPermission;
+import org.apache.nifi.expression.ExpressionLanguageScope;
+import org.apache.nifi.flowfile.FlowFile;
+import org.apache.nifi.flowfile.attributes.CoreAttributes;
+import org.apache.nifi.processor.AbstractProcessor;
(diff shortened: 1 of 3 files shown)


nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/DeleteFile.java
```diff
@@ -0,0 +1,183 @@
...
+@InputRequirement(InputRequirement.Requirement.INPUT_REQUIRED)
+@Tags({"file", "remove", "delete", "local", "files", "filesystem"})
+@CapabilityDescription("Deletes a file from the filesystem.")
+@WritesAttributes({
+        @WritesAttribute(
+                attribute = DeleteFile.ATTRIBUTE_FAILURE_REASON,
+                description = "Human-readable reason of failure. Only available if FlowFile is routed to relationship 'failure'."),
+        @WritesAttribute(
+                attribute = DeleteFile.ATTRIBUTE_EXCEPTION_CLASS,
+                description = "The class name of the exception thrown during processor execution. Only available if an exception caused the FlowFile to be routed to relationship 'failure'."),
```
joewitt, line 62: We do not want to be placing exception class names (class names in general) into flowfile attributes and/or any exception generated messages without context.  These processors should be doing the work of figuring out what went wrong on a low level and exposing that to the user in a human usable form.  That it routes to failure has two important distinctions to the user

1. did it fail for reasons I should retry and it might work later like the volume wasn't accessible temporarily or a conflicting file was there or i dont have write permissions
2 . did it fail because the file specified doesnt even exist and this would never work/etc..

We should not be putting exception messages in raw form without context or the exception class names.  Thanks

EndzeitBegins: Thank you for the feedback. 

I don't have any strong feelings about those attributes. Actually, I added them only [based on this feedback](https://github.com/apache/nifi/pull/8489#discussion_r1522123570) on the original PR #8489. 

As this is just a backport of [NIFI-12880](https://issues.apache.org/jira/browse/NIFI-12880) / PR #8489, which included those attributes. If we decide to remove these here, we should adjust the implementation on `main` as well. 

The case that the file is not found is handled explicitly by routing to a different relationship. Do you think it makes sense to handle additional cases, such as "permission.denied" similar to how `FetchFile` does and summarize all remaining cases under "failure"?

joewitt: I'm not saying there should be a hard rule to never do this but I think it is a mistake for this to become a standard.  A NiFi flow is not meant for only a Java developer to understand.  Instead the java dev/extension developer's role is to do the work necessary to surface key decision logic and context to the user in a user beneficial way.  Dumping exception class names and raw messages in there is generally not the answer.  Yes I do think it should not be on main either (and i didn't notice it was so thanks for flagging that).

marcus41495:
Is there any downside in adding the ability to remove directories? Maybe not for this ticket, but is it something that should be considered in the future? Currently there isn't any good way to do this inherently in NiFi, so it's common to end up with empty directories after using FetchFile or DeleteFile.

joewitt:
It can be an expensive check.  Checking each time if the file you just fetched is the last file in a location depending on OS/number of files in the directory could get costly.  It could be optional to do so though.

marcus41495:
That's a good point. I imagine it would have to be very intentional and not something the processor should do out of convenience when possible. ie, if the 'Filename' property is empty, and a 'Delete Directory' property is set to 'true', then it will check if it's empty and delete it (I imagine it can be dangerous to allow deleting directories that aren't empty).

exceptionfactory:
Regarding the consideration of deleting a directory, that would be something to consider in a separate Jira issue. Although deleting an individual file aligns with certain list-fetch processing patterns, deleting an entire directory would be a different use case, so that should be evaluated before introducing a potential change.

exceptionfactory (approved):
Concurring with the comments that adding an exception class as an attribute value is not a good general practice, I will proceed with merging this pull request as it aligns with the current behavior on the main branch.

It would be helpful to open a new pull request to remove the exception class attributes.

exceptionfactory:
Merged in https://github.com/apache/nifi/commit/089fa476b3003f785bf39fec45978c96c29974b1

PR #9017, opened 2024-06-28
NIFI-13344 Implement backend for uploading and managing custom NARs

bbende:
# Summary

This PR includes the work for [NIFI-13344](https://issues.apache.org/jira/browse/NIFI-13344) which introduces an API and framework components for uploading custom NARs and then managing the NARs that have been uploaded. This covers the "push model" described in the [Feature Proposal for Custom NAR Improvements](https://cwiki.apache.org/confluence/display/NIFI/Custom+NAR+Improvements).

The REST API introduced includes the following end-points:

```
Upload a NAR:
POST - /controller/nar-manager/upload (application/octet-stream)

List uploaded NARs:
GET -  /controller/nar-manager/nars

Get summary for a single NAR:
GET -  /controller/nar-manager/nars/{id}

Get more details for a single NAR like the extensions contained in it:
GET -  /controller/nar-manager/nars/{id}/detalls

Download a NAR:
GET -  /controller/nar-manager/nars/{id}/download

Delete a previously uploaded NAR:
DELETE -  /controller/nar-manager/nars/{id}
```
Some rules and restrictions of the current functionality...

If you upload a NAR with the same coordinate, it will replace the existing NAR, which includes:
- Stopping and ghosting any components instantiated from the NAR
- Unloading the previous NAR
- Loading the new NAR
- Unghosting the previously ghosted components
- Restoring the components state if needed

You can upload NARs in any order and if a NAR is missing it's dependency it will remain unloaded in a "Missing Dependency" state until the dependency is uploaded, at which it automatically becomes loaded/installed. Similar to how dropping NARs into `./extensions` works.

You can't upload a NAR with the same coordinate as a NAR that is not part of the "NAR Manager", meaning one of the NARs that was provided from a lib directory or a NarProvider. 

On delete, the `force` argument can be passed which will stop and ghost any components to allow delete to proceed. 

One other restriction is not being allowed to delete or replace a NAR that another NAR has a dependency on.

In a cluster, when a node joins the cluster it attempts to synchronize its NARs with the coordinator. This requires access to the REST APIs described above, which means the node identities will need READ on `/controller`.

CLI commands are implemented for all of the above end-points:
```
#> nifi upload-nar -nar nifi-example-processors-nar-1.0.nar
#> nifi upload-nar -nar nifi-example-service-api-nar-1.0.nar
#> nifi upload-nar -nar nifi-example-service-nar-1.1.nar

#> nifi list-nars

#   Id                                     Group             Artifact                       Version   State
-   ------------------------------------   ---------------   ----------------------------   -------   ---------
1   1c6575c8-e488-33d3-a226-7803bfc30004   org.apache.nifi   nifi-example-processors-nar    1.0       Installed
2   afd8c799-de79-33ad-9ad0-413bd5beff3b   org.apache.nifi   nifi-example-service-api-nar   1.0       Installed
3   5f63686a-b59f-30f6-ad2b-9b372678398e   org.apache.nifi   nifi-example-service-nar       1.1       Installed

#> nifi list-nar-component-types -nid 1c6575c8-e488-33d3-a226-7803bfc30004

#   Type                                             Kind
-   ----------------------------------------------   ---------
1   org.apache.nifi.processors.example.MyProcessor   PROCESSOR
```

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

- [XX] Pull Request refers to a feature branch with one commit containing changes

# Verification

There are two system tests that exercise the APIs, one in a standalone instance, and and one with a two node cluster to exercise the replication of uploads.

In addition, have been running a two node cluster locally and testing various API calls and scenarios.

### Build

  - [X] JDK 21

### UI Contributions

### Licensing

### Documentation


gresockj (changes requested):
Nice work, @bbende!  This looks super useful.  I've provided some initial feedback below, and I've tested out each of the endpoints and the CLI commands, including some of the error scenarios.

One scenario I hit is that after removing a NAR, you can get into a state where the UI still has access to add component types from the NAR, but the NAR has already been removed.  This is resolved by reloading the page, so not a huge inconvenience, but I wanted to make sure you were aware.
image

I also think some documentation would be appropriate to add to the User Guide.  Perhaps also something in the admin guide about the NAR directory.

I hope to test this out in clustered mode, but will submit this initial feedback first.

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarManifest.java
```diff
@@ -0,0 +1,260 @@
...
+        return new BundleCoordinate(dependencyGroup, dependencyId, dependencyVersion);
+    }
+
+    public static NarManifest fromFile(final File narFile) throws IOException {
+        try (final InputStream inputStream = new FileInputStream(narFile)) {
+            return fromInputStream(inputStream);
+        }
+    }
+
+    public static NarManifest fromInputStream(final InputStream narInputStream) throws IOException {
```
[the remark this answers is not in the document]
bbende: `fromFile` is the only one that is used from real code, but it was easier to test by creating a `Manifest` object, so I'll make `fromInputStream` and `fromManifest` have package visibility.

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarProperties.java
```diff
@@ -0,0 +1,214 @@
...
+    public Instant getInstalled() {
+        return installed;
+    }
+
+    public BundleCoordinate getCoordinate() {
+        return new BundleCoordinate(narGroup, narId, narVersion);
+    }
+
+    public Optional<BundleCoordinate> getDependencyCoordinate() {
+        if (narDependencyGroup == null) {
```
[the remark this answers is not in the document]
bbende: There is existing code somewhere else in Bundle or BundleDetails that has the same logic so I had continued with that, but I'll add a check that enforces there must be all 3 dependency values or none.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/NarInstallTask.java
```diff
@@ -0,0 +1,179 @@
...
+ * Task for installing a NAR that was submitted to the {@link NarManager}.
+ */
+public class NarInstallTask implements Runnable {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NarInstallTask.class);
+
+    private static final Set<Class<?>> ALLOWED_EXTENSION_TYPES = Set.of(
+            Processor.class,
+            ControllerService.class,
+            ReportingTask.class,
```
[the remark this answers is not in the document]
bbende: Good catch, it was meant to be

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,492 @@
...
+        this.narComponentManager = narComponentManager;
+        this.narLoader = narLoader;
+        this.webClientService = webClientService;
+        this.sslContext = sslContext;
+        this.installExecutorService = Executors.newSingleThreadExecutor();
+        this.deleteExecutorService = Executors.newSingleThreadExecutor();
+    }
+
+    // This serves two purposes...
+    // 1. Any previously stored NARs need to have their extensions loaded and made available for use during start up since they won't be in any of the standard NAR directories
```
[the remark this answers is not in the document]
bbende: `afterPropertiesSet` is a Spring interface method, similar to `@PostConstruct`, are you saying to make `afterPropertiesSet` call two smaller methods?

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarPersistenceProvider.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Called after instantiation to initialize the provider.
+     *
+     * @param initializationContext the context containing and properties and configuration required to initialize the provider
+     */
+    void initialize(NarPersistenceProviderInitializationContext initializationContext) throws IOException;
+
+    /**
+     * Creates a temporary file containing the contents of the given input stream.
+     * If the resulting file contains a valid NAR, it will be passed to the save method in order to move to its final destination.
```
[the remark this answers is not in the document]
bbende: Removing the second sentence

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,492 @@
...
+                final String coordinatorNarDigest = narSummaryDTO.getDigest();
+                final NarNode matchingNar = narNodesById.get(coordinatorNarId);
+                if (matchingNar == null) {
+                    LOGGER.info("Coordinator has NAR [{}] which does not exist locally, will download", coordinatorNarId);
+                    downloadNar(narRestApiClient, narSummaryDTO);
+                } else if (!coordinatorNarDigest.equals(matchingNar.getNarFileHexDigest())) {
+                    LOGGER.info("Coordinator has NAR [{}] which exists locally with a different digest, will download", coordinatorNarId);
+                    downloadNar(narRestApiClient, narSummaryDTO);
+                } else {
+                    LOGGER.info("Coordinator has NAR [{}] which exists locally with a matching digest, will not download", coordinatorNarId);
```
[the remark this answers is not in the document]
bbende: I think it is helpful to see all of the NARs returned from the coordinator, and also it si very infrequent only when joining the cluster.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/nifi/NiFiCommandGroup.java
```diff
@@ -202,6 +207,11 @@ protected List<Command> createCommands() {
         commands.add(new GetControllerConfiguration());
         commands.add(new UpdateControllerConfiguration());
         commands.add(new ChangeVersionProcessor());
+        commands.add(new UploadNar());
```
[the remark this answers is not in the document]
bbende: I think I'd prefer to delete the static list of commands from the toolkit guide since it is unmaintainable :) but I will appease.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ControllerResource.java
```diff
@@ -2411,6 +2429,289 @@ public Date getEndDate() {
...
+        final ClusterCoordinator clusterCoordinator = getClusterCoordinator();
+        if (clusterCoordinator != null) {
+            final Set<NodeIdentifier> disconnectedNodes = clusterCoordinator.getNodeIdentifiers(NodeConnectionState.CONNECTING, NodeConnectionState.DISCONNECTED, NodeConnectionState.DISCONNECTING);
+            if (!disconnectedNodes.isEmpty()) {
+                throw new IllegalStateException("Cannot upload NAR because the following %s nodes are not currently connected: %s".formatted(disconnectedNodes.size(), disconnectedNodes));
+            }
+        }
+
+        final long startTime = System.currentTimeMillis();
+        final InputStream maxLengthInputStream = new MaxLengthInputStream(inputStream, (long) DataUnit.GB.toB(1));
```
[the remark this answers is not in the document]
bbende: Made an update that does what I think you suggested. Was debating some other ideas like allowing passing in a custom message to the constructor of the input stream, or throwing a sub-class of IOException so caller could catch just that one and then throw a more specific message.

bbende:
@gresockj thanks for the thorough review! I pushed up a commit to address some changes, a few things to still address:
- user guide
- some stuff around the MaxLengthInputStream
- the afterProperties method question

Regarding the comment about needing to refresh the UI... Currently the UI retrieves the lists of processor and controller service types one time when the canvas is loaded. We can file something to try and improve it, but I consider it a separate effort since it already impacts stuff like dropping a NAR into ./extenions, or a NAR being pulled from a NarProvider and download into ./extensions.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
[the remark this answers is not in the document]
gresockj: Yes.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/nifi/NiFiCommandGroup.java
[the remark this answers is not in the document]
gresockj: Yeah, I'd support deleting the static list, too!  Perhaps in a different PR.

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarManifest.java
```diff
@@ -132,27 +132,47 @@ public BundleCoordinate getDependencyCoordinate() {
...
+     */
+    static NarManifest fromManifest(final Manifest manifest) {
         if (manifest == null) {
             throw new IllegalArgumentException("NAR content is missing required META-INF/MANIFEST entry");
         }
 
         final Attributes attributes = manifest.getMainAttributes();
 
-        return NarManifest.builder()
+        return org.apache.nifi.nar.NarManifest.builder()
```
gresockj, line 175: Is this now an unnecessary qualification?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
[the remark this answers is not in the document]
bbende: Updated

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarManifest.java
bbende: Updated

nifi-commons/nifi-utils/src/main/java/org/apache/nifi/stream/io/MaxLengthInputStream.java
```diff
@@ -0,0 +1,128 @@
...
+    public long skip(final long n) throws IOException {
+        final long toSkip = Math.min(n, limit - bytesRead);
+        final long skipped = in.skip(toSkip);
+        bytesRead += skipped;
+        return skipped;
+    }
+
+    @Override
+    public int available() throws IOException {
+        return in.available();
```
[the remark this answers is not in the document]
bbende: I'm leaning towards leaving this unchanged since we want the caller to keep reading and encounter the error if the available data exceeds the limit. There is a similar class in nifi-utils called `LimitingInputStream` which was a similar concept, except it just keeps returning `-1` once reaching the limit, and that one seems to also not alter the available logic.

nifi-commons/nifi-utils/src/test/java/org/apache/nifi/stream/io/MaxLengthInputStreamTest.java
```diff
@@ -0,0 +1,60 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class MaxLengthInputStreamTest {
+
```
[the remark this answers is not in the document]
bbende: I added some more tests, prob still more that could be done, but maybe good enough for now.

nifi-commons/nifi-utils/src/main/java/org/apache/nifi/stream/io/MaxLengthInputStream.java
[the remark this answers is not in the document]
gresockj: Got it, that makes sense.

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarManifestEntry.java
```diff
@@ -0,0 +1,48 @@
...
+    NAR_DEPENDENCY_VERSION("Nar-Dependency-Version"),
+    BUILD_TAG("Build-Tag"),
+    BUILD_REVISION("Build-Revision"),
+    BUILD_BRANCH("Build-Branch"),
+    BUILD_TIMESTAMP("Build-Timestamp"),
+    BUILD_JDK("Build-Jdk"),
+    BUILT_BY("Built-By"),
+    CREATED_BY("Created-By");
+
+    final String manifestName;
```
exceptionfactory, line 38: Should this be named `entryName` instead of `manifestName`?

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarPersistenceInfo.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.io.File;
+import java.util.Objects;
+
+/**
+ * The information about a NAR persisted by a {@link NarPersistenceProvider}.
+ */
+public class NarPersistenceInfo {
+
+    private final File narFile;
```
exceptionfactory, line 28: What do you think about using `Path` instead of `File` for a bit more abstraction of this member variable? Although they are mostly interchangeable, `Path` provides a bit more generalization in terms of methods available on the object.

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarPersistenceProvider.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @return the info for all NARs persisted by this persistence provider
+     */
+    Set<NarPersistenceInfo> getAllNarInfo() throws IOException;
+
+    /**
+     * Shuts down the provider and cleans up any resources held by it. Once this method has returned, the
+     * provider may be initialized once again via the {@link #initialize(NarPersistenceProviderInitializationContext)} method.
+     */
+    void shutdown();
```
exceptionfactory, line 107: What do you think about calling this `close()` and extending the `Closeable` interface. The only downside is the potential for throwing an `IOException`, but that may be useful in an implementation.

nifi-framework-api/src/test/java/org/apache/nifi/nar/NarManifestTest.java
```diff
@@ -0,0 +1,104 @@
...
+    public void testFromManifestFull() throws IOException {
+        final String manifestPath = "nar/MANIFEST-FULL.MF";
+        final Manifest manifest = loadManifest(manifestPath);
+        final NarManifest narManifest = NarManifest.fromManifest(manifest);
+        assertNotNull(narManifest);
+
+        assertEquals("nifi-framework-nar", narManifest.getId());
+        assertEquals("org.apache.nifi", narManifest.getGroup());
+        assertEquals("2.0.0-SNAPSHOT", narManifest.getVersion());
+
```
exceptionfactory, line 46: Recommend defining static Strings for these identifier and version values so they can be reused in building objects and in assertions.

nifi-framework-api/src/test/java/org/apache/nifi/nar/NarPropertiesTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private static final String PROPERTIES_FILE_SUFFIX = ".properties";
+
+    @Test
+    public void testNarProperties(@TempDir final Path tempDir) throws IOException {
+        final NarProperties narProperties = NarProperties.builder()
+                .sourceType(NarSource.UPLOAD.name())
+                .narGroup("org.apache.nifi")
+                .narId("nifi-standard-nar")
+                .narVersion("2.0.0")
```
exceptionfactory, line 46: Recommend defining static values and reusing across test methods.

nifi-framework-api/src/test/resources/nar/MANIFEST-FULL.MF
```diff
@@ -0,0 +1,13 @@
+Manifest-Version: 1.0
```
exceptionfactory, line 1: Do these manifest files support comments? If so, it would be possible to add the license header and avoid the need to exclude them from the license checking.

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/replication/StandardUploadRequestReplicator.java
```diff
@@ -0,0 +1,173 @@
...
+
+    public StandardUploadRequestReplicator(final ClusterCoordinator clusterCoordinator, final WebClientService webClientService) {
+        this.clusterCoordinator = Objects.requireNonNull(clusterCoordinator, "Cluster Coordinator is required");
+        this.webClientService = Objects.requireNonNull(webClientService, "Web Client Service is required");
+    }
+
+    @Override
+    public <T> T upload(final UploadRequest<T> uploadRequest) throws IOException {
+        final String filename = uploadRequest.getFilename();
+        final File tempFile = Files.createTempFile("nifi", "upload").toFile();
```
exceptionfactory, line 70: It seems like NiFi should use a directory under `work` instead of the JVM temporary directory, what do you think?

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/replication/StandardUploadRequestReplicator.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.OptionalLong;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+
+/**
+ * Implementation of {@link UploadRequestReplicator} that uses OkHttp Client.
```
exceptionfactory, line 52: Looks like the comment should be updated:
```suggestion
 * Implementation of {@link UploadRequestReplicator} that uses the nifi-web-client-api.
```

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/replication/StandardUploadRequestReplicator.java
```diff
@@ -0,0 +1,173 @@
...
+                    .header(RequestReplicator.REQUEST_EXECUTION_HTTP_HEADER, "true")
+                    .header(RequestReplicator.REPLICATION_INDICATOR_HEADER, "true")
+                    .header(ProxiedEntitiesUtils.PROXY_ENTITIES_CHAIN, ProxiedEntitiesUtils.buildProxiedEntitiesChainString(user))
+                    .header(ProxiedEntitiesUtils.PROXY_ENTITY_GROUPS, ProxiedEntitiesUtils.buildProxiedEntityGroupsString(user.getIdentityProviderGroups()));
+
+            logger.debug("Replicating upload request for {} to {}", filename, nodeId);
+
+            try (final HttpResponseEntity response = request.retrieve()) {
+                final int statusCode = response.statusCode();
+                if (!HttpResponseStatus.isSuccessful(statusCode)) {
```
exceptionfactory, line 162: Although this is convenient, it seems like it should be checking for a specific status code, given that the REST API should return a single expected status.

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/replication/StandardUploadRequestReplicator.java
```diff
@@ -0,0 +1,173 @@
...
+                    final String responseMessage = IOUtils.toString(response.body(), StandardCharsets.UTF_8);
+                    throw new UploadRequestReplicationException("Failed to replicate upload request to " + nodeId + " - " + responseMessage, statusCode);
```
exceptionfactory, line 164: Recommend minor adjustment to message formatting:
```suggestion
                    throw new UploadRequestReplicationException("Failed to replicate upload request to [%s] %s".formatted(nodeId, responseMessage), statusCode);
```

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/replication/StandardUploadRequestReplicator.java
```diff
@@ -0,0 +1,173 @@
...
+                }
+                final InputStream responseBody = response.body();
+                return objectMapper.readValue(responseBody, uploadRequest.getResponseClass());
+            }
+        }
+
+
```
exceptionfactory, line 171:
```suggestion
```

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/replication/UploadRequestReplicator.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.IOException;
+
+/**
+ * A request replicator for replicating file uploads to all nodes in the cluster.
+ */
+public interface UploadRequestReplicator {
+
+    String FILENAME_HEADER = "Filename";
+    String CONTENT_TYPE_HEADER = "Content-Type";
+    String UPLOAD_CONTENT_TYPE = "application/octet-stream";
```
exceptionfactory, line 29: I recommend avoiding member variables on an interface. Instead, recommend placing them in an `enum`, or on the implementation class.

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/test/java/org/apache/nifi/cluster/manager/NarSummariesMergerTest.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class NarSummariesMergerTest {
+
+    @Test
+    public void testMergeNarSummaries() {
+        final NarSummaryEntity narSummary1 = createNarSummaryEntity("org.apache.nifi", "nifi-nar-1", "1.0.0");
+        final NarSummaryEntity narSummary2 = createNarSummaryEntity("org.apache.nifi", "nifi-nar-2", "1.0.0");
+        final NarSummaryEntity narSummary3 = createNarSummaryEntity("org.apache.nifi", "nifi-nar-3", "1.0.0");
+        final NarSummaryEntity narSummary4 = createNarSummaryEntity("org.apache.nifi", "nifi-nar-4", "1.0.0");
```
exceptionfactory, line 40: Recommend defining and reusing host, version, and port variables across this test method.

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/test/java/org/apache/nifi/cluster/manager/NarSummaryDtoMergerTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class NarSummaryDtoMergerTest {
+
+    @Test
+    public void testMergeNarSummaryDtosInstallingIntoInstalled() {
+        final String narId = UUID.randomUUID().toString();
+
+        final NarSummaryDTO summary1 = new NarSummaryDTO();
+        summary1.setIdentifier(narId);
+        summary1.setCoordinate(new NarCoordinateDTO("org.apache.nifi", "nifi-nar-1", "1.0.0"));
```
exceptionfactory, line 37: Recommend defining and reusing the version number and other properties.

nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/nar/NarNode.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Objects;
+
+/**
+ * Represents a NAR that exists in the NAR Manager.
+ */
+public class NarNode {
+
+    private final String identifier;
+    private final File narFile;
+    private final String narFileHexDigest;
```
exceptionfactory, line 30: I recommend removing `Hex` from the name unless there is a particular reason to keep it.
```suggestion
    private final String narFileDigest;
```

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
```diff
@@ -1365,6 +1376,97 @@ private ProvenanceRepository createProvenanceRepository(final NiFiProperties pro
         }
     }
 
+    private NarPersistenceProvider createNarPersistenceProvider(final NiFiProperties nifiProperties) {
```
exceptionfactory, line 1379: Is there a reason for creating the implementation in this class? It seems cleaner to define it in the Spring configuration, and then pass the `NarPersistenceProvider` to the constructor, or through a setter.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/NarRestApiClient.java
```diff
@@ -0,0 +1,138 @@
...
+            try {
+                responseMessage = IOUtils.toString(response.body(), StandardCharsets.UTF_8);
+            } catch (IOException e) {
+                throw new NarRestApiRetryableException("Error reading response from " + requestUri + " - " + statusCode, e);
+            }
+            if (statusCode == HttpResponseStatus.CONFLICT.getCode()) {
+                throw new NarRestApiRetryableException("Error calling " + requestUri + " - " + statusCode + " - " + responseMessage);
+            } else {
+                throw new IllegalStateException("Error calling " + requestUri + " - " + statusCode + " - " + responseMessage);
+            }
```
exceptionfactory, line 135: Recommend using `.formatted()` to construct the error message strings for readability.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/NarRestApiClient.java
```diff
@@ -0,0 +1,138 @@
...
+        LOGGER.info("Downloading NAR [{}] from {}", identifier, requestUri);
+
+        try {
+            final HttpResponseEntity response = webClientService.get()
+                    .uri(requestUri)
+                    .header("Accept", "application/octet-stream")
+                    .retrieve();
+            return getResponseBody(requestUri, response);
+        } catch (final WebClientServiceException e) {
+            throw new NarRestApiRetryableException(e.getMessage(), e);
```
exceptionfactory, line 116: Recommend providing the identifier in the message:
```suggestion
            throw new NarRestApiRetryableException("Download NAR [%s] failed".formatted(identifier), e);
```

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/NarRestApiClient.java
```diff
@@ -0,0 +1,138 @@
...
+
+        final HttpRequestBodySpec requestBodySpec = webClientService.get()
+                .uri(requestUri)
+                .header("Accept", "application/json");
+
+        try (final HttpResponseEntity response = requestBodySpec.retrieve()) {
+            final InputStream responseBody = getResponseBody(requestUri, response);
+            return objectMapper.readValue(responseBody, NarSummariesEntity.class);
+        } catch (final WebClientServiceException | IOException e) {
+            throw new NarRestApiRetryableException(e.getMessage(), e);
```
exceptionfactory, line 91:
```suggestion
            throw new NarRestApiRetryableException("List NAR summaries failed", e);
```

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarComponentManager.java
```diff
@@ -0,0 +1,251 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+public class StandardNarComponentManager implements NarComponentManager {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(StandardNarComponentManager.class);
```
exceptionfactory, line 49: Recommend using lowercase for loggers.
```suggestion
    private static final Logger logger = LoggerFactory.getLogger(StandardNarComponentManager.class);
```

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,504 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.stream.Collectors;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+public class StandardNarManager implements NarManager, InitializingBean, DisposableBean {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(StandardNarManager.class);
```
exceptionfactory, line 61:
```suggestion
    private static final Logger logger = LoggerFactory.getLogger(StandardNarManager.class);
```

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,504 @@
...
+        this.narLoader = narLoader;
+        this.webClientService = webClientService;
+        this.sslContext = sslContext;
+        this.installExecutorService = Executors.newSingleThreadExecutor();
+        this.deleteExecutorService = Executors.newSingleThreadExecutor();
+    }
+
+    // This serves two purposes...
+    // 1. Any previously stored NARs need to have their extensions loaded and made available for use during start up since they won't be in any of the standard NAR directories
+    // 2. NarLoader keeps track of NARs that were missing dependencies to consider them on future loads, so this restores state that may have been lost on a restart
```
exceptionfactory, line 100: Recommend changing this to a method-style comment for consistency.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,504 @@
...
+    @Override
+    public void afterPropertiesSet() throws IOException {
+        final Collection<NarPersistenceInfo> narInfos = loadExistingNars();
+        restoreState(narInfos);
+    }
+
+    private Collection<NarPersistenceInfo> loadExistingNars() throws IOException {
```
exceptionfactory, line 107: Recommend moving all private methods after all public methods in this class.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,504 @@
...
+                .narFileHexDigest(narDigest)
+                .manifest(manifest)
+                .source(NarSource.valueOf(narInfo.getNarProperties().getSourceType()))
+                .sourceIdentifier(narInfo.getNarProperties().getSourceId())
+                .state(state)
+                .build();
+    }
+
+    @Override
+    public void destroy() {
```
exceptionfactory, line 151: Instead of implementing `DisposableBean`, this could implement `Closeable` and use the `close()` method.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,504 @@
...
+
+        final Bundle existingBundle = extensionManager.getBundle(coordinate);
+        if (existingBundle != null && !persistenceProvider.exists(coordinate)) {
+            throw new IllegalStateException("A NAR is already registered with the same group, id, and version, " +
+                    "and can not be replaced because it is not part of the NAR Manager");
+        }
+
+        final Set<Bundle> bundlesWithMatchingDependency = extensionManager.getDependentBundles(coordinate);
+        if (!bundlesWithMatchingDependency.isEmpty()) {
+            throw new IllegalStateException("Unable to replace NAR because it is a dependency of other NARs");
```
exceptionfactory, line 199: The coordinate should be included.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,504 @@
...
+            return persistenceProvider.readNar(coordinate);
+        } catch (final FileNotFoundException e) {
+            throw new NarNotFoundException(coordinate);
+        }
+    }
+
+    @Override
+    public synchronized void syncWithClusterCoordinator() {
+        if (clusterCoordinator == null) {
+            LOGGER.info("Cluster coordinator is null, will not sync NARs");
```
exceptionfactory, line 327: It seems like this should be changed to `debug`. Also recommend rewording:
```suggestion
            LOGGER.info("Cluster coordinator is configured: NAR synchronization disabled");
```

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,504 @@
...
+            return;
+        }
+
+        // This sync method is called from the method that loads the flow from a connection response, which means there must already be a cluster coordinator to
+        // have gotten a response from, but during testing there were cases where calling clusterCoordinator.getElectedActiveCoordinatorNode() was still null, so
+        // the helper method here will keep checking for the identifier up to a certain threshold to avoid slight timing issues
+        final NodeIdentifier coordinatorNodeId = getElectedActiveCoordinatorNode();
+        if (coordinatorNodeId == null) {
+            LOGGER.warn("Unable to obtain the node identifier for the cluster coordinator, will not sync NARs");
```
exceptionfactory, line 336:
```suggestion
            LOGGER.warn("Unable to obtain the node identifier for the cluster coordinator: NAR synchronization disabled");
```

nifi-system-tests/nifi-system-test-suite/src/test/resources/conf/clustered/node2/nifi.properties
```diff
@@ -125,6 +125,9 @@ nifi.components.status.repository.implementation=org.apache.nifi.controller.stat
 nifi.components.status.repository.buffer.size=1440
 nifi.components.status.snapshot.frequency=1 min
 
+# NAR Persistence Provider Properties
+nifi.nar.persistence.provider.properties.directory=./nar_storage
```
exceptionfactory, line 129:
```suggestion
nifi.nar.persistence.provider.properties.directory=./bundles
```

nifi-system-tests/nifi-system-test-suite/src/test/resources/conf/default/nifi.properties
```diff
@@ -126,6 +126,9 @@ nifi.components.status.repository.implementation=org.apache.nifi.controller.stat
 nifi.components.status.repository.buffer.size=1440
 nifi.components.status.snapshot.frequency=1 min
 
+# NAR Persistence Provider Properties
+nifi.nar.persistence.provider.properties.directory=./nar_storage
```
exceptionfactory, line 130:
```suggestion
nifi.nar.persistence.provider.properties.directory=./bundles
```

nifi-system-tests/nifi-system-test-suite/src/test/resources/conf/pythonic/nifi.properties
```diff
@@ -130,6 +130,9 @@ nifi.components.status.repository.implementation=org.apache.nifi.controller.stat
 nifi.components.status.repository.buffer.size=1440
 nifi.components.status.snapshot.frequency=1 min
 
+# NAR Persistence Provider Properties
+nifi.nar.persistence.provider.properties.directory=./nar_storage
```
exceptionfactory, line 134:
```suggestion
nifi.nar.persistence.provider.properties.directory=./bundles
```

exceptionfactory (changes requested):
Great work on this new feature @bbende! Thanks for the detailed feedback thus far @gresockj.

The overall approach and implementation looks good. I pointed out a number of naming and message-related comments, most of them are implementation details.

nifi-commons/nifi-web-client-api/src/main/java/org/apache/nifi/web/client/api/HttpResponseStatus.java
```diff
@@ -55,4 +57,12 @@ public enum HttpResponseStatus {
...
         return code;
     }
+
+    public boolean isSuccessful() {
+        return isSuccessful(code);
+    }
+
+    public static boolean isSuccessful(final int code) {
+        return code >= 200 && code < 300;
+    }
```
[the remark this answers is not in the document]
bbende: I will probably just remove both then since the member method isn't really helping since there is no way to get an instance of `HttpResonseStatus` from a code.

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarPersistenceProvider.java
bbende: It is never really intended to be closed except when shutting down the application and seemed more consistent with other framework extensions like `StateProvider` and repositories which have methods like `shutdown()` or `cleanup()`. I lean towards leaving this as is.

nifi-framework-api/src/test/java/org/apache/nifi/nar/NarManifestTest.java
bbende: I can move them to constants, but in this case the values are coming from the actual MANIFEST files in `src/test/resources`, so it is not helping for consistency between building vs assertions.

nifi-framework-api/src/test/resources/nar/MANIFEST-FULL.MF
bbende: As far as I know it does not support it, I tried using property file comment syntax which didn't work, and this seems to suggest it doesn't support it:
https://stackoverflow.com/questions/2025872/comment-lines-in-manifest-of-a-jar

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarPersistenceInfo.java
bbende: I can update it, but it seems like not much benefit because most of the existing NAR code already works off `File`, so this new code will be passing around `Path` just to then call `toFile()` to be able to call stuff like `NarLoadResult load(Collection<File> narFiles)`

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarPersistenceProvider.java
exceptionfactory: Thanks, that makes sense, the current approach looks good.

nifi-framework-api/src/main/java/org/apache/nifi/nar/NarPersistenceInfo.java
exceptionfactory: That's a good point given the other references. Thanks.

nifi-framework-api/src/test/java/org/apache/nifi/nar/NarManifestTest.java
exceptionfactory: Yea, the main value is just defining it once in the class.

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/replication/StandardUploadRequestReplicator.java
bbende: I will add an expected code to the UploadRequest so the caller can specify what is considered success

nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/replication/UploadRequestReplicator.java
bbende: I'm going to allow headers to be passed on the `UploadRequest` and then move these constants to `ControllerResource`. The replicator will then add any headers from the request to the replicated request.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
bbende: Good point, it really doens't need to be in FlowController. I will extract the creation to a separate factory bean that can be called from the new Java Spring config.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarManager.java
```diff
@@ -0,0 +1,504 @@
...
+    }
+
+    private NarSummariesEntity listNarSummaries(final NarRestApiClient narRestApiClient) {
+        try {
+            return narRestApiClient.listNarSummaries();
+        } catch (final NarRestApiRetryableException e) {
+            LOGGER.warn("[{}], will retry", e.getMessage());
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("", e);
+            }
```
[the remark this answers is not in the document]
bbende: The reason I did this was because it is normal that the listing will need to be retied a couple of times while possibly waiting for the coordinator to be ready to serve requests, so printing a huge stack trace each time makes it look like a major problem when it most likely isn't. I'll probably remove this since it is likely not needed.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/StandardNarPersistenceProviderInitializationContext.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.nifi.nar;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class StandardNarPersistenceProviderInitializationContext implements NarPersistenceProviderInitializationContext {
+
+    private final Map<String, String> properties;
+
+    public StandardNarPersistenceProviderInitializationContext(final Map<String, String> properties) {
```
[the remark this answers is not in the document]
bbende: I think I was originally thinking that some implementation may have no properties, but the framework would still pass in empty map, so it makes sense to require non null.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ControllerResource.java
```diff
@@ -2411,6 +2429,289 @@ public Date getEndDate() {
...
     }
 
+    // ------------
+    // NARs
+    // ------------
+
+    @POST
+    @Consumes(MediaType.APPLICATION_OCTET_STREAM)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("nar-manager/upload")
```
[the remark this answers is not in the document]
bbende: The reason I added `/upload` is because when we implement the second phase of this with the Extension Registry Clients, I picture there being another similar end-point, but instead of uploading it would be `/controller/nar-manager/install` where the body specifies a registry client id and coordinates to indicate to install the given NAR from the given client.

So we can do one of the following:

```
/nar-manager/upload
/nar-manager/install
```

```
/nar-manager/nars/upload
/nar-manager/nars/install
```

```
/nar-manager/nars  (this would be the upload without any specific path)
/nar-manager/nars/install
```
Thoughts?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ControllerResource.java
```diff
@@ -2411,6 +2429,289 @@ public Date getEndDate() {
...
+        }
+
+        final NarDetailsEntity detailsEntity = serviceFacade.getNarDetails(id);
+        return generateOkResponse(detailsEntity).build();
+    }
+
+    @GET
+    @Consumes(MediaType.WILDCARD)
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    @Path("/nar-manager/nars/{id}/download")
```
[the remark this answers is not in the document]
bbende: We already have a `GET` for `nar-manager/nars/{id}` which returns the summary for the NAR with the given id.

I guess then can be differentiated by the requested content type? 

We also have similar examples like downloading a flow definition vs retrieving JSON for a process process group:
```
/process-groups/{id}/download
/process-groups/{id}
```

nifi-framework-bundle/nifi-framework/nifi-resources/pom.xml
```diff
@@ -116,6 +116,10 @@
         <nifi.status.repository.questdb.persist.component.days>3</nifi.status.repository.questdb.persist.component.days>
         <nifi.status.repository.questdb.persist.location>./status_repository</nifi.status.repository.questdb.persist.location>
 
+        <!-- NAR Persistence properties -->
+        <nifi.nar.persistence.provider.implementation>org.apache.nifi.nar.StandardNarPersistenceProvider</nifi.nar.persistence.provider.implementation>
+        <nifi.nar.persistence.provider.properties.directory>./nar_storage</nifi.nar.persistence.provider.properties.directory>
```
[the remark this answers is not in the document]
bbende: I'm leaning towards `nar_repository` to be consistent with other repositories we have. I like something more specific than `nars` or `bundles` since it a special directory managed by the provider, as opposed to just a directory to drop NARs in.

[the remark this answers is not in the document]
exceptionfactory: Yes, `nar_repository` sounds like a good option, following existing examples.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ControllerResource.java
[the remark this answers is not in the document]
bbende: Decided to use `content` on the upload and download end-points, so it will be:
```
POST /nar-manager/nars/content
GET /nar-manager/nars/{id}/content
```

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/NarRestApiClient.java
bbende: I wanted to leave this one as is because the message from the retryable exception will be logged in the retry loop in `StandardNarManager`, so it looks like this currently:
```
2024-07-03 16:46:27,810 INFO [main] org.apache.nifi.nar.StandardNarManager Synchronizing NARs with cluster coordinator
2024-07-03 16:46:27,914 WARN [main] org.apache.nifi.nar.StandardNarManager Error calling https://localhost:8443/nifi-api/controller/nar-manager/nars - 409 - The Flow Controller is initializing the Data Flow.: retrying
2024-07-03 16:46:27,914 INFO [main] org.apache.nifi.nar.StandardNarManager Unable to retrieve NAR summaries from cluster coordinator: retrying until [2024-07-03T20:51:27.811Z]
```
If we changed it as suggested, then the second line there would just say `List NAR summaries failed` with no info about why.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/NarRestApiClient.java
bbende: Same reasoning as next comment

bbende:
@exceptionfactory thanks for the review! I pushed a couple of commits that should address most of your comments, let me know of anything else.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/nar/NarRestApiClient.java
exceptionfactory: Thanks for providing more context, that makes sense in light of the example described.

bbende:
@exceptionfactory @gresockj I pushed up another commit that adds support for handling of NARs containing Python processors. This required some changes on the Python side to track the coordinate of the NAR that the processor came from because this wasn't available on the Java side before, and all Python components on the Java side are represented as coming from `org.apache.nifi:python-extensions` with the version coming from the Processor's version.

exceptionfactory (approved):
Thanks for the great work on this new feature @bbende, and thanks for the feedback and testing @gresockj! +1 merging

PR #9020, opened 2024-07-01
[NIFI-13471] make theme file path configurable during maven build

scottyaslan:


nifi-frontend/pom.xml
```diff
@@ -33,10 +33,9 @@
         <frontend.working.dir>${project.build.directory}/frontend-working-directory</frontend.working.dir>
         <copy-rename-maven-plugin.version>1.0.1</copy-rename-maven-plugin.version>
         <!--
-            String name of the theme file to be used.
-            Ex: 'purple' // will overwrite the 'material' theme file with the 'purple' theme.
+            Path to the theme file to be used.
         -->
-        <frontend.theme>material</frontend.theme>
+        <frontend.theme>${frontend.working.dir}/libs/shared/src/assets/themes/material.scss</frontend.theme>
```
rfellows, line 38: It might be better to offer 2 variables. one for the directory of the themes and the other for the name of the theme or theme file. That way, if you want to override it on the command line you can still just provide the name of the theme you want (material, purple, ...) and not have to know the path to it.

```
<frontend.theme.dir>${frontend.working.dir}/libs/shared/src/assets/themes</frontend.theme.dir>
<frontend.theme.name>material</frontend.theme.name>
```

then the usage below could be
```
<sourceFile>${frontend.theme.dir}/${frontend.theme.name}.scss</sourceFile>
```

and on the command line we can override one or the other or both using `-Dfrontend.theme.dir=/Users/username/myNifiThemes/` and/or `-Dfrontend.theme.name=myTheme`

rfellows:
@scottyaslan can you rebase this PR please? it seems the failing `dependency-check` step was actually removed from the github workflow and should resolve this failure for us.

PR #9030, opened 2024-07-03
[NIFI-13483] use primary contrast color for active thread count icon …

scottyaslan:
…and text on primary surfaces


mcgilman (approved):
Thanks @scottyaslan!

PR #9056, opened 2024-07-05
NIFI-13523: Refreshing the controller service listing after creation or deletion

mcgilman:
NIFI-13523:
- Refreshing the controller service listing after creation or deletion to ensure allowable values of other services is accurate.


scottyaslan:
Reviewing...

scottyaslan (approved):
LGTM! Thanks @mcgilman

PR #9058, opened 2024-07-07
NIFI-13510 Remove KerberosCredentialsService and implementations

exceptionfactory:
# Summary

[NIFI-13510](https://issues.apache.org/jira/browse/NIFI-13510) Removes the `KerberosCredentialsService` Controller Service interface and associated implementations in favor of the `KerberosUserService` interface introduced in NiFi 1.15.0.

Additional removals include individual component properties for Kerberos Principal and Password values.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


joewitt:
I believe the env variable trick we came up with many moons ago can be removed/deleted now

Searching for 'ALLOW_EXPLICIT_KEYTAB' shows this is in nifi-env.sh and nifi-stateless-env.sh.  I believe all such references can be removed from those files now.

I also see various kerberos related values in the nifi.properties.  With the removal of the spnego bits and this update are those entries still necessary/desirable?  krb5.conf and so on?  I was wondering if maybe components that dont let us specify much context would still need it.

exceptionfactory:
> I believe the env variable trick we came up with many moons ago can be removed/deleted now
> 
> Searching for 'ALLOW_EXPLICIT_KEYTAB' shows this is in nifi-env.sh and nifi-stateless-env.sh. I believe all such references can be removed from those files now.
> 
> I also see various kerberos related values in the nifi.properties. With the removal of the spnego bits and this update are those entries still necessary/desirable? krb5.conf and so on? I was wondering if maybe components that dont let us specify much context would still need it.

Thanks for the pointer on the environment variable, I pushed a change to remove it from the default `nifi-env.sh`.

I also removed some unnecessary references to NiFi Properties from the test classes.

The Kerberos properties still have some references when creating the ZooKeeper client and when configuring Controller Services that could request Kerberos properties. However, those `KerberosContext` methods are limited to scripted components. That could be worth evaluating for removal in a separate issue, since it is public in `nifi-api` as opposed to extension components targeted in this pull request.

turcsanyip:
@exceptionfactory Thanks for cleaning up the Kerberos config in these components!
I think it would also be useful to add migration code for helping the move from the old controller service (or processor level properties) to `KerberosUserService`. Similar to the [proxy propery migraton](https://github.com/apache/nifi/blob/f5476ca3b4eea6be10ba9130456c274e98093710/nifi-extension-bundles/nifi-extension-utils/nifi-migration-utils/src/main/java/org/apache/nifi/migration/ProxyServiceMigration.java#L47-L68). Do you have plans to do it in this jira or in a separate one?

exceptionfactory:
> @exceptionfactory Thanks for cleaning up the Kerberos config in these components! I think it would also be useful to add migration code for helping the move from the old controller service (or processor level properties) to `KerberosUserService`. Similar to the [proxy propery migraton](https://github.com/apache/nifi/blob/f5476ca3b4eea6be10ba9130456c274e98093710/nifi-extension-bundles/nifi-extension-utils/nifi-migration-utils/src/main/java/org/apache/nifi/migration/ProxyServiceMigration.java#L47-L68). Do you have plans to do it in this jira or in a separate one?

Thanks for asking @turcsanyip. I added the `migrateProperties()` methods to remove existing references to the legacy properties as an incremental step forward. I wasn't planning to include a full migration in this pull request, but I think it could be considered a follow-on effort. With these components supporting both direct properties and the `KerberosCredentialsService`, a complete solution would have to consider each scenario, and there may be some opportunity for reuse in some components. As mentioned in the pull request, the `KerberosUserService` has been around since NiFi 1.15.0, so the migration seems less necessary than others, but if you are inclined to add it, it seems like a good scope for a subsequent Jira issue.

joewitt:
Ok thanks for adding the removal/cleanup related to the old workaround/hack thing we had with environment variables.  Thrilled to get this out.

Also sounds good that the migrate mechanism can be added to should there be challenges for existing configs. 

+1 merging

PR #9059, opened 2024-07-07
NIFI-13526 searched for mostly unused statics, unused variables, and …

joewitt:
…cleaned up a few dead classes and removed refactored left behind code in flowfilerepo

removed nifi-socket-utils and an abstract class with no usage.  Moved remaining usage of socket utils to only place usage remains

# Summary

[NIFI-13526](https://issues.apache.org/jira/browse/NIFI-13526)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


joewitt:
Could not find any apparent usage of this class https://github.com/apache/nifi/pull/9059/files#diff-421025e1497a4c2a1e9580f75bb9fa5ea41de556e1c31e332e8eab61143e22a3

CuratorNode... seemed dead. But I am not certain

nifi-extension-bundles/nifi-pgp-bundle/nifi-pgp-processors/src/main/java/org/apache/nifi/processors/pgp/attributes/CompressionAlgorithm.java
```diff
@@ -26,8 +26,6 @@ public enum CompressionAlgorithm {
 
     ZIP(CompressionAlgorithmTags.ZIP),
 
-    ZLIB(CompressionAlgorithmTags.ZLIB),
-
```
exceptionfactory, line 30: This is a valid option and the change should be reverted.

nifi-framework-api/src/main/java/org/apache/nifi/cluster/ConnectionState.java
```diff
@@ -24,41 +24,6 @@ public enum ConnectionState {
...
-    /**
-     * Instances has completed offloading
-     */
-    OFFLOADED,
-
-    /**
-     * Instance has been removed from the cluster
-     */
-    REMOVED,
-
```
exceptionfactory, line 61: This looks valid and seems like it should be reverted.

nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/groups/BundleUpdateStrategy.java
```diff
@@ -22,11 +22,7 @@ public enum BundleUpdateStrategy {
     /**
      * If a proposed flow has a different bundle than the current flow, the bundle is not updated
      */
-    IGNORE_BUNDLE,
```
exceptionfactory, line 25: This could be valid in some circumstances and should be reverted.

nifi-extension-bundles/nifi-provenance-repository-bundle/nifi-persistent-provenance-repository/src/main/java/org/apache/nifi/provenance/index/lucene/CommitPreference.java
```diff
@@ -19,6 +19,5 @@
 
 public enum CommitPreference {
     FORCE_COMMIT,
-    PREVENT_COMMIT,
-    NO_PREFERENCE;
+    PREVENT_COMMIT;
```
exceptionfactory, line 22: It looks like this could be valid, and should be reverted.

nifi-framework-bundle/nifi-framework/nifi-administration/src/main/java/org/apache/nifi/admin/service/entity/ConfigureDetailsEntity.java
```diff
@@ -20,8 +20,6 @@
  * Enumeration of Action Configure Details properties stored as Entity objects
  */
 public enum ConfigureDetailsEntity implements EntityProperty {
-    ACTION("action"),
-
```
exceptionfactory, line 24: This is valid and should be reverted.

exceptionfactory (changes requested):
Thanks for cleaning up these items @joewitt, the pruning and relocating of some socket classes also looks good.

I noted several `enum` element removals that should be reverted, as the unused values are still valid values in some circumstances.

Otherwise, this should be ready to go following those adjustments.

nifi-framework-api/src/main/java/org/apache/nifi/cluster/ConnectionState.java
joewitt: There is now a NodeConnectionState that has those.  These 'ConnectionState' entries look defunct at least from code scanning.     How are you finding they are valid?

nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/groups/BundleUpdateStrategy.java
joewitt: Can you share more about how this would be valid but never referenced?

joewitt:
Went ahead and pushed a change restoring all the enums. Thanks

nifi-framework-api/src/main/java/org/apache/nifi/cluster/ConnectionState.java
exceptionfactory: Yes, in this case, the `StandardClusterDetailsFactory` reads the `NodeConnectionState` and then converts that string name to this `ConnectionState` value.

nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/groups/BundleUpdateStrategy.java
exceptionfactory: On closer review, this appears to be limited to explicit usage in method arguments, so it seems like it could be removed. However, as a framework-core-api element, it may be better considered in the context of other framework changes, keeping it for the moment seems best.

PR #9061, opened 2024-07-08
NIFI-13528: Python processor's customValidate function does not work

mark-bathori:
# Summary

[NIFI-13528](https://issues.apache.org/jira/browse/NIFI-13528)

Example for customValidate usage with the current changes:

```
  def customValidate(self, context):
      validationResult = ValidationResult()
      validationResult.add_validation_result("Test subject 1", "Test validation 1", False, "Input 1")
      validationResult.add_validation_result("Test subject 2", "Test validation 2", False, "Input 2")

      return validationResult.build_validation_result_list()
```

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-py4j-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/properties.py
+++ b/nifi-extension-bundles/nifi-py4j-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/properties.py
@@ -292,30 +292,11 @@ class PropertyDescriptor:
         builder.identifiesExternalResource(cardinality, types[0], types[1:])
 
 
-class ProcessContext:
+class PropertyContext:
     __trivial_attribute_reference__ = re.compile(r"\$\{([^${}\[\],:;/*\' \t\r\n\\d][^${}\[\],:;/*\' \t\r\n]*)}")
     __escaped_attribute_reference__ = re.compile(r"\$\{'([^${}\[\],:;/*\' \t\r\n\\d][^${}\[\],:;/*\'\t\r\n]*)'}")
 
-    def __init__(self, java_context):
-        self.java_context = java_context
-
-        descriptors = java_context.getProperties().keySet()
-        self.name = java_context.getName()
-        self.property_values = {}
-        self.descriptor_value_map = {}
-
-        for descriptor in descriptors:
-            property_value = java_context.getProperty(descriptor.getName())
-            string_value = property_value.getValue()
-
-            property_value = self.__create_python_property_value(descriptor.isExpressionLanguageSupported(), property_value, string_value)
-            self.property_values[descriptor.getName()] = property_value
-
-            python_descriptor = PropertyDescriptor.from_java_descriptor(descriptor)
-            self.descriptor_value_map[python_descriptor] = string_value
-
-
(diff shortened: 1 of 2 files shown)


markap14:
Hey @mark-bathori thanks for the update! I agree with the problem and the general approach. The only concern that I have here is with the new `ValidationResult` class. Specifically, I don't think it makes sense to create a `ValidationResult` and then call `add_validation_result` on that object. I would also like to avoid having to create a `ValidationResult` just to call `build_validation_result_list()` just in order to return it.

Rather, I think we should define `ValidationResult` to have an `init` that takes those 4 parameters (`subject`, `valid`, `explanation`, input`) and just return a `list` of them from the method. For example, we should be able to implement the above as:

```
  def customValidate(self, context):
      results = []
      results.append(new ValidationResult(subject="Test subject 1", explanation="Test validation 1", valid=False, input="Input 1")
      results.append(new ValidationResult(subject="Test subject 2", explanation="Test validation 2", valid=False, input="Input 1")
      return results
```

Or, alternatively,
```
  def customValidate(self, context):
      return [
            new ValidationResult(subject="Test subject 1", explanation="Test validation 1", valid=False, input="Input 1"),
            new ValidationResult(subject="Test subject 2", explanation="Test validation 2", valid=False, input="Input 1")]
```

The `PythonProcessorAdapter`, then, should be responsible for taking that Python `list` that is returned and converting it into the `ArrayList` needed by the framework.

mark-bathori:
Thanks @markap14 for the review, I reworked the PR according to your comments.

pgyori (commented):
Thank you @mark-bathori !
The Python classes you created/modified all have corresponding Java interfaces. I understand that the Python classes do not need to implement all the methods from the corresponding Java interface. However, I found it a little confusing at first that there is a slight mismatch between the Python class methods and the Java interface method declarations. I would like to suggest the following changes to sync the Python side with the Java side:
- PropertyContext in Java has a method 'getAllProperties()', please consider renaming the Python PropertyContext.getProperties() method to that.
- ProcessContext and ValidationContext in Java both have methods called 'getProperties()' and 'newPropertyValue()'. Please consider moving the corresponding Python methods from PropertyContext to the two subclasses. (getProperties() in the sublclasses could call the supercass' getAllProperties() method). I understand that this would cause code duplication, I only recommend it so that the Python code matches the Java code.

I believe these small changes would make it a little more straightforward to understand the code should someone in the future continue extending the Python classes based on the Java interfaces.

mark-bathori:
Thanks @pgyori  for the review. I think that it is a good approach to offer the same functionality both in Python and Java side, however I feel it is unnecessary to add methods that are currently unused. Over time we can further extend the Python side with additional methods when new functionalities will require them.

- Renaming 'getProperties()' to 'getAllProperties()' wouldn't be good because they have different return type in Java side
- At this point I would also avoid code duplication, since both contexts use the same approach for 'getProperties()' and 'newPropertyValue()' so I would keep it as it is for now.

pgyori:
Thank you @mark-bathori , I understand and accept your reasoning.
+1 merging.

PR #9064, opened 2024-07-08
NIFI-12491 Implemented Header strategy for ExcelReader

dan-s1:
# Summary

[NIFI-12491](https://issues.apache.org/jira/browse/NIFI-12491)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.excel;
+
+import org.apache.commons.lang3.math.NumberUtils;
+import org.apache.nifi.components.AllowableValue;
+import org.apache.nifi.context.PropertyContext;
+import org.apache.nifi.logging.ComponentLog;
+import org.apache.nifi.schema.access.SchemaAccessStrategy;
+import org.apache.nifi.schema.access.SchemaField;
+import org.apache.nifi.schema.access.SchemaNotFoundException;
+import org.apache.nifi.schema.inference.FieldTypeInference;
+import org.apache.nifi.schema.inference.TimeValueInference;
+import org.apache.nifi.serialization.SimpleRecordSchema;
+import org.apache.nifi.serialization.record.DataType;
+import org.apache.nifi.serialization.record.RecordField;
+import org.apache.nifi.serialization.record.RecordSchema;
+import org.apache.nifi.util.SchemaInferenceUtil;
+import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.DataFormatter;
+import org.apache.poi.ss.usermodel.Row;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class ExcelHeaderSchemaStrategy implements SchemaAccessStrategy {
+    private static final Set<SchemaField> schemaFields = EnumSet.noneOf(SchemaField.class);
+    static final int NUM_ROWS_TO_DETERMINE_TYPES = 10; // NOTE: This number is arbitrary.
+    static final AllowableValue HEADER_DERIVED = new AllowableValue("Use String Fields From Header", "Use String Fields From Header",
+            "The configured first row of the Excel file is a header line that contains the names of the columns. The schema will be derived by using the "
+                    + "column names in the header and the following " + NUM_ROWS_TO_DETERMINE_TYPES + " rows to determine the type(s) of each column");
+
+    private final PropertyContext context;
+    private final ComponentLog logger;
+    private final TimeValueInference timeValueInference;
+    private final DataFormatter dataFormatter;
+
+    public ExcelHeaderSchemaStrategy(PropertyContext context, ComponentLog logger, TimeValueInference timeValueInference, Locale locale) {
+        this.context = context;
+        this.logger = logger;
+        this.timeValueInference = timeValueInference;
+        this.dataFormatter = locale == null ? new DataFormatter() : new DataFormatter(locale);
+    }
+
+    @Override
+    public RecordSchema getSchema(Map<String, String> variables, InputStream contentStream, RecordSchema readSchema) throws SchemaNotFoundException, IOException {
+        if (this.context == null) {
+            throw new SchemaNotFoundException("Schema Access Strategy intended only for validation purposes and cannot obtain schema");
+        }
+
+        final String requiredSheetsDelimited = context.getProperty(ExcelReader.REQUIRED_SHEETS).evaluateAttributeExpressions(variables).getValue();
+        final List<String> requiredSheets = ExcelReader.getRequiredSheets(requiredSheetsDelimited);
+        final Integer rawFirstRow = context.getProperty(ExcelReader.STARTING_ROW).evaluateAttributeExpressions(variables).asInteger();
+        final int firstRow = rawFirstRow == null ? NumberUtils.toInt(ExcelReader.STARTING_ROW.getDefaultValue()) : rawFirstRow;
+        final int zeroBasedFirstRow = ExcelReader.getZeroBasedIndex(firstRow);
+        final String password = context.getProperty(ExcelReader.PASSWORD).getValue();
+        final ExcelRecordReaderConfiguration configuration = new ExcelRecordReaderConfiguration.Builder()
+                .withRequiredSheets(requiredSheets)
+                .withFirstRow(zeroBasedFirstRow)
+                .withPassword(password)
+                .build();
+
+        final RowIterator rowIterator = new RowIterator(contentStream, configuration, logger);
+        final Map<String, FieldTypeInference> typeMap = new LinkedHashMap<>();
+        List<String> fieldNames = null;
+        int index = 0;
+
+        while (rowIterator.hasNext()) {
+           Row row = rowIterator.next();
+           if (index == 0) {
+               fieldNames = getFieldNames(firstRow, row);
+           } else if (index <= NUM_ROWS_TO_DETERMINE_TYPES) {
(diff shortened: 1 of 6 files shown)


nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class ExcelHeaderSchemaStrategy implements SchemaAccessStrategy {
+    private static final Set<SchemaField> schemaFields = EnumSet.noneOf(SchemaField.class);
+    static final int NUM_ROWS_TO_DETERMINE_TYPES = 10; // NOTE: This number is arbitrary.
+    static final AllowableValue HEADER_DERIVED = new AllowableValue("Use String Fields From Header", "Use String Fields From Header",
```
exceptionfactory, line 52: The reference to `String` seems unnecessary. What do you think about naming this `First Row Fields` or similar?

exceptionfactory (commented):
Thanks for working on this improvement @dan-s1! I plan to take a closer look soon and run some tests. I raised one question about the name of the strategy.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
dan-s1: I would rather not use the word "first" as per the code it technically does not have to be the very first row rather its the first configured row. How about
`"Use Field Names from Row Header"` or `"Use Column Names from Row Header"`

exceptionfactory: Thanks for the background link, that is helpful. With that background, I think simply calling it `Use Starting Row` is clear and concise. That would work with the default setting where the Starting Row is `1`, while also being clear that it could use a subsequent row if the `Starting Row` is configured to a different number.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+
+            fieldNames.add(fieldName);
+        }
+
+        return fieldNames;
+    }
+
+    private void inferSchema(final Row row, final List<String> fieldNames, final Map<String, FieldTypeInference> typeMap) throws SchemaNotFoundException {
+        if (!ExcelUtils.hasCells(row)) {
+            throw new SchemaNotFoundException(String.format("Row %s a required row for inferring field types has no cells", row.getRowNum()));
```
exceptionfactory, line 129: This check effectively requires a minimum of 10 rows, which should not be necessary.

Instead of requiring 10 rows and requiring a matching number of cells, it seems better to make this method more permissive. That would allow less than 10 rows, and it would allow some empty or partial rows, but that seems acceptable because the type inference is a best-effort approach.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+            throw new SchemaNotFoundException(String.format("Field names could not be determined from configured header row %s, as this row has no cells with data", firstRowIndex));
+        }
+
+        final List<String> fieldNames = new ArrayList<>();
+        for (int index = 0; index < row.getLastCellNum(); index++) {
+            final Cell cell = row.getCell(index);
+            final String fieldName = dataFormatter.formatCellValue(cell);
+
+            if (fieldName == null || fieldName.isEmpty()) {
+                throw new SchemaNotFoundException(String.format("Configured header row %s at cell %s is empty which is not suitable for a field name", row.getRowNum(), index));
```
exceptionfactory, line 118: This check requires more strict enforcement of a header row, but in some cases, an individual cell could be empty. Rather than throwing an error, it seems better to use a generated field name based on the cell number, such as `column_0` like the `ExcelSchemaInference` class.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class ExcelHeaderSchemaStrategy implements SchemaAccessStrategy {
+    private static final Set<SchemaField> schemaFields = EnumSet.noneOf(SchemaField.class);
+    static final int NUM_ROWS_TO_DETERMINE_TYPES = 10; // NOTE: This number is arbitrary.
+    static final AllowableValue HEADER_DERIVED = new AllowableValue("Use Starting Row", "Use Starting Row",
```
exceptionfactory, line 52: Recommend changing the variable name to reflect the value:
```suggestion
    static final AllowableValue USE_STARTING_ROW = new AllowableValue("Use Starting Row", "Use Starting Row",
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.EnumSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class ExcelHeaderSchemaStrategy implements SchemaAccessStrategy {
```
exceptionfactory, line 49: I recommend changing the `getDefaultSchemaAccessStrategy` to return this value as it is likely to be preferred over other alternatives.

exceptionfactory (changes requested):
Thanks for the working on this @dan-s1. I tested the implementation and noted a few places where the strategy should be a bit more lenient when inferring schema fields and value types.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
dan-s1: Per the logic on lines 91-99 this header strategy can work with less than ten lines if the data is less than 10 lines long. Furthermore my logic allows for empty cells but the total number of cells cannot exceed the initial size of the header (lines 132-134) which I thought makes sense as I need to align each field to a known column. If there all of a sudden is a new column that would throw things off.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
exceptionfactory: Thanks! Although using this strategy would seem to imply that spreadsheet has a fixed and complete header, I have seen cases where the first cell or the first row is empty, but the rest of the row is filled. Compensating for that would allow integrating components to handle specific fields. For example, a `QueryRecord` Processor could select only those fields that are required for subsequent evaluation, so missing a particular header field would not necessarily present a problem. For now, this seems like the best approach. If necessary, we could introduce a more strict strategy in the future.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
exceptionfactory: Thanks for the reply. Yes, I think encapsulating the logic to ignore empty rows would accomplish the end goal.

exceptionfactory (changes requested):
Thanks for the updates @dan-s1, this looks close to completion, just noted to more minor recommendations.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,162 @@
...
+        for (int index = 0; index < row.getLastCellNum(); index++) {
+            final Cell cell = row.getCell(index);
+            final String fieldName = dataFormatter.formatCellValue(cell);
+
+            // NOTE: This accounts for column(s) which may be empty in the configured starting row.
+            if (fieldName != null && !fieldName.isEmpty()) {
+                fieldNames.add(fieldName);
+            } else {
+                fieldNames.add(ExcelUtils.FIELD_NAME_PREFIX + index);
+            }
```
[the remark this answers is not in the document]
dan-s1: Always good to see other points of view as I actually thought the opposite.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -0,0 +1,162 @@
...
+           } else {
+               break;
+           }
+
+           index++;
+        }
+
+        if (typeMap.isEmpty()) {
+            throw new SchemaNotFoundException("Could not infer the schema from the first " + NUM_ROWS_TO_DETERMINE_TYPES +
+                    " rows as they were all blank");
```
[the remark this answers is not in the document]
dan-s1: FYI with a `%n`, Intellij complained with `Too many arguments for format string (found: 1, expected: 0) ` so I switched it to `%s`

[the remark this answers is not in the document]
exceptionfactory: Sorry about that, it should be `%d` for numbers.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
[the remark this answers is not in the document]
exceptionfactory: Thanks, admittedly a stylistic point of view, I find it easier to read conditional without negations, but I appreciate the emphasis on more expected versus less expected cases.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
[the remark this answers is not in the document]
dan-s1: Is it okay if I keep `%s`? I just checked it in that way before seeing this message.

exceptionfactory (approved):
Thanks again for the new feature @dan-s1, the latest version looks good! +1 merging

PR #9065, opened 2024-07-08
NIFI-13399 Implemented a different function that allows for Unicode

sashsuji:
NIFI-13399 Implemented a different function that allows for Unicode characters in the encoding of strings to base64

# Summary

[NIFI-13399](https://issues.apache.org/jira/browse/NIFI-13399)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
+++ b/nifi-registry/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
@@ -34,6 +34,23 @@ var config = {
     }
 };
 
+var utils = {
+    /**
+     * Encode strings to base-64 with support for Unicode characters
+     *
+     * @param {string} str     The string to be encoded.
+     * @returns {*}
+     */
+    b64UnicodeEncode: function (str) {
+        return btoa(encodeURIComponent(str).replace(
+            /%([0-9A-F]{2})/g,
+            function toSolidBytes(match, p1) {
+                return String.fromCharCode('0x' + p1);
+            }
+        ));
+    }
+};
+
 /**
  * NfRegistryApi constructor.
  *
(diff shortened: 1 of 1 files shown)


exceptionfactory (commented):
@sashsuji Have you tested this approach with non-ASCII characters authenticating to Registry? The Registry BasicAuth implementation is capable of decoding Base64 bytes into UTF-8, but this approach appears to encode the username and password using URI encoding before passing the bytes to the `btoa()` function.

nightgryphon:
@exceptionfactory the URI encoding is the intermediate step and escaped characters are than converted back to 'char' with regexp replace(). I didn't test this patch at the moment but looks like it should work.

sashsuji:
Full transparency, I've been trying to test this but I can't figure out how to get to the login page. I've only been able to run the registry site. How can I properly test this?

exceptionfactory:
@sashsuji Thanks for following up, and thanks for the comment @nightgryphon. This type of change to character encoding could have a serious impact if not implemented properly. It is very important to test changes as part of the pull request process. Reviewers are also responsible for evaluating and testing, but it is important for authors to do the initial work.

The [NiFi Registry Admin Guide](https://nifi.apache.org/docs/nifi-registry-docs/html/administration-guide.html#user_authentication) provides details on configuring security integrated with LDAP as one option. Unfortunately, NiFi Registry does not support the single-user style of authentication that NiFi provides, so I understand it requires more work to setup a test scenario.

With that background, I am closing this pull request for now, but feel free to reopen after running some tests with different types of inputs.

mohamedmeqlad99 (commented):
You replaced the plain btoa with utils.b64UnicodeEncode, which provides support for encoding Unicode characters (e.g., if a username or password contains special characters).
This change is beneficial in scenarios where credentials might include non-ASCII characters, as btoa alone would fail in those cases.

PR #9066, opened 2024-07-09
NIFI-13529: Update the Calcite Connection properties to indicate that…

markap14:
… timezone is UTC timezone. We do this because we already take care to ensure that timestamps, etc. are normalized to UTC timezone before passing them in. We do not want the Avatica ResultSet then treating Timestamp objects as local timestamps. We want them treated as UTC timestamps.

# Summary

[NIFI-13529](https://issues.apache.org/jira/browse/NIFI-13529)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-calcite-utils/src/main/java/org/apache/nifi/sql/CalciteDatabase.java
+++ b/nifi-commons/nifi-calcite-utils/src/main/java/org/apache/nifi/sql/CalciteDatabase.java
@@ -193,6 +193,10 @@ public class CalciteDatabase implements Closeable {
             calciteProperties = properties;
         }
 
+        // If not explicitly set, default timezone to UTC. We ensure that when we provide timestamps, we convert them to UTC. We don't want
+        // Calcite trying to convert them again.
+        calciteProperties.putIfAbsent("timeZone", "UTC");
+
         final Connection sqlConnection = DriverManager.getConnection("jdbc:calcite:", calciteProperties);
         final CalciteConnection connection = sqlConnection.unwrap(CalciteConnection.class);
         connection.getRootSchema().setCacheEnabled(false);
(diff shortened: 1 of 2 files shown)


Kapkiai:
+1
I have verified that the changes fixes bug. Thanks @markap14 for the quick fix!

exceptionfactory (changes requested):
I'm concerned about this type of change, we should not override local time zone settings as it could have other impacts when handling localized dates. We should consider other approaches rather than hard-coding UTC.

exceptionfactory:
@markap14 Can you provide a reference to where we are already converting input timestamps to UTC?

markap14:
Hey @exceptionfactory I suppose you're right, we are not necessarily converting to UTC before providing the object to Calcite. But what we really want is for Calcite not to mess with the Timestamp object we give it at all. When you call `ResultSet.getTimestamp(1)`, it should be returning the same object as if you call `Result.getObject(1)` but Avatica doesn't it see it that way. Instead, when you call `ResultSet.getTimestamp(1)` with Avatica it attempts to localize the Timestamp according to its configured TimeZone. If not specified, it uses the local timezone as that configured TimeZone. In essence what I'm doing with this PR is telling it to use UTC as the configured TImeZone so that it doesn't do any manipulation / localization.

So that's why I wrote the unit test as I did - when I give it a timestamp, and then get back a timezone, it should be:
(a) the same value that I gave to it, and
(b) the same result regardless of whether I call `ResultSet.getObject` vs. `ResultSet.getTimestamp`

exceptionfactory (approved):
Thanks for the explanation @markap14. Reviewing some of the Calcite code, I see places where it changes the returned timestamp, so setting the Calcite timeZone to UTC makes sense as the offset is `0`, avoiding any changes.

With that clarification, the tests are helpful, and this looks good! +1 merging

PR #9073, opened 2024-07-11
NIFI-13541 [MiNiFi] Validate flow during startup

briansolo1985:
# Summary

[NIFI-13541](https://issues.apache.org/jira/browse/NIFI-13541)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-headless-server/src/main/java/org/apache/nifi/headless/HeadlessNiFiServer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-headless-server/src/main/java/org/apache/nifi/headless/HeadlessNiFiServer.java
@@ -169,6 +169,7 @@ public class HeadlessNiFiServer implements NiFiServer {
             flowService.start();
             flowService.load(null);
             flowController.onFlowInitialized(true);
+            validateFlow();
             FlowManager flowManager = flowController.getFlowManager();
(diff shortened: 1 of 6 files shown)


nifi-framework-bundle/nifi-framework/nifi-headless-server/src/main/java/org/apache/nifi/headless/HeadlessNiFiServer.java
```diff
@@ -195,6 +196,10 @@ public void preDestruction() throws AuthorizerDestructionException {
         }
     }
 
+    protected void validateFlow() {
+        logger.info("Flow validated successfully");
```
bejancsaba, line 200: Shouldn't it be better to just leave this empty or provide a more generic message stating "Flow Validation was not implemented" or something similar. Just asking because if this is not overridden then no validation is actually executed. What do you think?

briansolo1985: Thanks for checking the PR. Yes it makes sense, updated the log message

bejancsaba (approved):
Thanks for the change, all looks good. +1 from my side as soon as the automation finishes I will merge

PR #9075, opened 2024-07-11
NIFI-13477: Fix flickering tooltip

mcgilman:
NIFI-13477:
- Refreshing the controller service listing after creation or deletion to ensure allowable values of other services is accurate.


scottyaslan (approved):
These changes work well! Thanks @mcgilman!

PR #9080, opened 2024-07-15
NIFI-13327: Allow migrating properties of Parameter Providers

takraj:
# Summary

[NIFI-13327](https://issues.apache.org/jira/browse/NIFI-13327)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -265,6 +265,8 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
                     service.migrateConfiguration(originalPropertyValues, serviceFactory);
                 } else if (extension instanceof final ReportingTaskNode task) {
                     task.migrateConfiguration(originalPropertyValues, serviceFactory);
+                } else if (extension instanceof final ParameterProviderNode parameterProvider) {
+                    parameterProvider.migrateConfiguration(originalPropertyValues, serviceFactory);
                 }
             }
         } finally {
--- a/nifi-api/src/main/java/org/apache/nifi/parameter/ParameterProvider.java
+++ b/nifi-api/src/main/java/org/apache/nifi/parameter/ParameterProvider.java
@@ -16,13 +16,13 @@
  */
 package org.apache.nifi.parameter;
 
+import java.io.IOException;
+import java.util.List;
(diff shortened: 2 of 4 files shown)


takraj:
As of 4e6f79d, I need some help to continue integrating this feature into NiFi. As far as I understand, Parameter Providers are special in the sense they do not belong to Process Groups, but rather they are 'Controller Level' components. The current implementation does not work, as it seems like the `migrateProperties` function does not get called, even though there is a Parameter Provider defined in the `flow.json.gz`, which was copied from a previous version of NiFi.

I'm pretty much stuck. Please help me with some hints.

exceptionfactory:
@takraj Thanks for starting work on this approach. Given that the feature does not work as it stands, I recommend closing the pull request for now and continuing the discussion on the associated Jira issue. I am planning on taking evaluating some other Parameter Provider changes, but unfortunately I'm not in a position to provide a concrete way forward right now.

Referencing your feature branch in the Jira issue would provide an opportunity for some consideration prior to reopening the pull request, after confirming that the basic functionality works.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -265,6 +265,8 @@ public void synchronize(final ProcessGroup group, final VersionedExternalFlow ve
                     service.migrateConfiguration(originalPropertyValues, serviceFactory);
                 } else if (extension instanceof final ReportingTaskNode task) {
                     task.migrateConfiguration(originalPropertyValues, serviceFactory);
+                } else if (extension instanceof final ParameterProviderNode parameterProvider) {
+                    parameterProvider.migrateConfiguration(originalPropertyValues, serviceFactory);
```
exceptionfactory, line 269: @takraj This conditional never appears to be fired because Parameter Providers are never added to Created Extensions.

exceptionfactory (changes requested):
@takraj I highlighted a particular issue in the Synchronzier where the logic appears to be off. There is more research necessary to track down the right place to implement the changes to call the migration at the appropriate time. If you are not familiar with the changes that need to be made, it may be better to revisit this after further evaluation.

exceptionfactory:
@takraj Revisiting this pull request in light of the Database Adapter migration, I think it is best to close this for now and consider revisiting in the future.

Migrating Parameter Provider properties seems like less of a concern, since it is a more specialized feature, and the database-backed implementation is also a less-common use case. The change for Database Adapters could move forward independent of this change, so I recommend focusing there before coming back to this, and also getting something to a working state before reopening.

PR #9086, opened 2024-07-16
NIFI-13554 GetSlackReaction processor to fetch reactions for Slack message

krisztina-zsihovszki:
# Summary

[NIFI-13554](https://issues.apache.org/jira/browse/NIFI-13554)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-slack-bundle/nifi-slack-processors/src/main/java/org/apache/nifi/processors/slack/GetSlackReaction.java
+++ b/nifi-extension-bundles/nifi-slack-bundle/nifi-slack-processors/src/main/java/org/apache/nifi/processors/slack/GetSlackReaction.java
@@ -1,0 +1,524 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


nifi-extension-bundles/nifi-slack-bundle/nifi-slack-processors/src/main/java/org/apache/nifi/processors/slack/GetSlackReaction.java
```diff
@@ -0,0 +1,524 @@
...
+             + "and in case of '" + MessageIdentifierStrategy.JSON_PATH.getValue() + "' additional JSON path values needs to be specified to fetch channel id "
+             + "and message timestamp from incoming JSON content.")
+            .required(true)
+            .allowableValues(MessageIdentifierStrategy.class)
+            .defaultValue(MessageIdentifierStrategy.ATTRIBUTES.getValue())
+            .build();
+
+    static final PropertyDescriptor CHANNEL_ID_JSON_PATH = new PropertyDescriptor.Builder()
+            .name("Channel ID JSON Path")
+            .description("The JSON Path which identifies the channel ID in the incoming JSON.")
```
mattyb149, line 145: Should we require input to be in JSON format? It might be more flexible (and facilitate a future ConsumeSlackRecord processor for example) to use Record Path and a Record Reader?

exceptionfactory (changes requested):
I'm a bit concerned about the use of JSON Paths for several properties. If the input format is from other Processors, then making this configurable seems unnecessary.

Beyond that point, I'm more concerned about the scalability of this the check and wait strategy. The idea of the wait period property seems like it could easily lead to FlowFiles backed up in a queue, waiting for reactions that may never arrive. This presents some fundamental flow design concerns. I understand that retrieving reactions is separate from messages themselves, however, it seems that a more scalable approach would require either changes to ConsumeSlack, or a new ConsumeSlackReactions Processor. A feed of reactions allows for eventual merging with messages, as opposed to a Processor designed to loop and wait with a large number of polling requests.

PR #9087, opened 2024-07-17
[NIFI-13535] - Fix: Issues selecting GitHub branch on version control screens in new UI

rfellows:
[NIFI-13535](https://issues.apache.org/jira/browse/NIFI-13535)

Properly include the branch query param in API calls to the backend registry API when provided.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
@@ -212,7 +212,8 @@ export class ImportFromRegistry extends CloseOnEscapeDialog implements OnInit {
     flowChanged(flowId: string): void {
         const registryId = this.importFromRegistryForm.get('registry')?.value;
         const bucketId = this.importFromRegistryForm.get('bucket')?.value;
-        this.loadVersions(registryId, bucketId, flowId);
+        const branch = this.importFromRegistryForm.get('branch')?.value;
+        this.loadVersions(registryId, bucketId, flowId, branch);
     }
 
     loadBranches(registryId: string): void {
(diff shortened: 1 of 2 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
```diff
@@ -212,7 +212,8 @@ export class ImportFromRegistry extends CloseOnEscapeDialog implements OnInit {
     flowChanged(flowId: string): void {
         const registryId = this.importFromRegistryForm.get('registry')?.value;
         const bucketId = this.importFromRegistryForm.get('bucket')?.value;
-        this.loadVersions(registryId, bucketId, flowId);
+        const branch = this.importFromRegistryForm.get('branch')?.value;
+        this.loadVersions(registryId, bucketId, flowId, branch);
```
bbende, line 216: It looks like there is a similar function for `bucketChanged` which then calls `loadFlows(registryId, bucketId)` and I think that would need to include the branch as well right?

bbende:
Also, we need the same behavior on the "Start Version Version Control" screen. I am not sure if that is backed by `save-version-dialog.component.ts`, but in there I see we have `branchChanged` which will reload the buckets based on the branch, but we don't have anything for `bucketChanged` to reload flows, or `flowChanged` to reload versions. It is possible I am looking at the wrong thing.

rfellows:
> Also, we need the same behavior on the "Start Version Version Control" screen. I am not sure if that is backed by `save-version-dialog.component.ts`, but in there I see we have `branchChanged` which will reload the buckets based on the branch, but we don't have anything for `bucketChanged` to reload flows, or `flowChanged` to reload versions. It is possible I am looking at the wrong thing.

Start version control doesn't load flows or flow versions since we are adding a new flow to the repository. I'm not sure I understand what needs changed there.

bbende (approved):
Sorry for the confusion, you are right about the Start dialog. Latest changes here look good and are working correctly. +1

mcgilman:
Will review...

mcgilman (approved):
Thanks @rfellows!

PR #9112, opened 2024-07-25
NIFI-13575 Improve Jetty Server handling of Web App Exceptions

exceptionfactory:
# Summary

[NIFI-13575](https://issues.apache.org/jira/browse/NIFI-13575) Enables the `throwUnavailableOnStartupException` property for configured Web Applications in the framework Jetty Server. This property instructs Jetty to throw exceptions caught when starting a configured Web Applications, instead of just capturing and storing the exception. This approach removes the need for iterating over configured Jetty Server Handlers, and avoids potential issues related to nested handlers, ensuring that a Web Application exception will cause the Jetty Server to shutdown instead of continuing the startup process.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/JettyServer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/JettyServer.java
@@ -202,7 +202,7 @@ public class JettyServer implements NiFiServer, ExtensionUiLoader {
     private UiExtensionMapping componentUiExtensions;
     private Collection<WebAppContext> componentUiExtensionWebContexts;
 
-    private Map<BundleCoordinate, List<App>> appsByBundleCoordinate = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


bbende (commented):
Code looks good and verified the original scenario of failing to fetch parameters from a parameter provider now fails start up.

I noticed there is still a code block for extension UIs that is checking the context for the unavailable exception:
https://github.com/apache/nifi/blob/main/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/JettyServer.java#L375-L382

It seems like the previous intent of the code was to not fail start up for extension UIs that failed to start, and instead just log a warning, but now since all of them go through the `loadWar` helper method, I think a failed extension UI would also throw the exception and fail start up. Just wanted to get your thoughts on that before proceeding.

exceptionfactory:
Thanks for the review @bbende.

That's a very good point regarding failure handling for extension applications. I pushed a changed enumerating the required contexts paths, and adjusted setting of the throw exceptions property based on whether the context is a required framework path.

bbende (approved):
The latest updated looks good, I'll wait for the builds to complete and then merge if they pass

PR #9114, opened 2024-07-25
NIFI-13584 Remove unused included libraries from nifi-assembly

exceptionfactory:
# Summary

[NIFI-13584](https://issues.apache.org/jira/browse/NIFI-13584) Removes unused library includes from the `nifi-assembly` descriptor configuration to resolve warnings in the build related to dependencies that are no longer specified.

Additional changes include removing the `README.md` in `nifi-assembly` and instead packaging the main project `README.md` from the parent module.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


nifi-assembly/README.md
```diff
@@ -1,117 +0,0 @@
-<!--
```
bbende, line 1: I'm not strongly against this, but do we care that this READ for the assembly was more focused on the nifi-assembly, where as the one in the root of the source tree is more general about building source and also has sections about MiNiFi and NiFi Registry which would not be relevant for someone looking at nifi-assembly?

exceptionfactory: Thanks @bbende, I considered that fact, but I think the duplication was a greater concern than the inclusion of extra information. We probably should streamline the main project README.md as well. However, if you think this separate README is worth maintaining, I can back out the change.

bbende: I personally lean towards leaving the existing assembly README because we also have similar distribution READMEs for MiNiFi and NiFi Registry, and the one in the root of the source tree feels more like a general README for the source repo, and not really for the assembly binary. Maybe another ticket to consider options for streamlining all of the READMEs?

bbende (approved):
Looks good, thanks!

PR #9115, opened 2024-07-26
NIFI-13409 Enforcing lexiographical order of the identity mappings to match documentation

simonbence:
# Summary

Referred documentation: https://nifi.apache.org/docs/nifi-docs/html/administration-guide.html#identity-mapping-properties

[NIFI-13409](https://issues.apache.org/jira/browse/NIFI-13409)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-security-identity/src/main/java/org/apache/nifi/authorization/util/IdentityMappingUtil.java
+++ b/nifi-commons/nifi-security-identity/src/main/java/org/apache/nifi/authorization/util/IdentityMappingUtil.java
@@ -26,6 +26,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -141,6 +143,28 @@ public class IdentityMappingUtil {
      * @return the mapped identity, or the same identity if no mappings matched
      */
     public static String mapIdentity(final String identity, List<IdentityMapping> mappings) {
+        final TreeMap<String, IdentityMapping> mappingsInOrderByKey = new TreeMap<>();
+        mappings.forEach(m -> mappingsInOrderByKey.put(m.getKey(), m));
+
+        for (final Map.Entry<String, IdentityMapping> mappingEntry : mappingsInOrderByKey.entrySet()) {
+            final IdentityMapping mapping = mappingEntry.getValue();
+            final Matcher m = mapping.getPattern().matcher(identity);
+            if (m.matches()) {
+                final String pattern = mapping.getPattern().pattern();
+                final String replacementValue = escapeLiteralBackReferences(mapping.getReplacementValue(), m.groupCount());
+                final String replacement = identity.replaceAll(pattern, replacementValue);
+
+                if (Transform.UPPER.equals(mapping.getTransform())) {
+                    return replacement.toUpperCase();
+                } else if (Transform.LOWER.equals(mapping.getTransform())) {
+                    return replacement.toLowerCase();
+                } else {
+                    return replacement;
+                }
+            }
+        }
+
+
         final List<String> keys = mappings.stream().map(m -> m.getKey()).collect(Collectors.toList());
         Collections.sort(keys);
 
--- a/nifi-commons/nifi-security-identity/src/main/java/org/apache/nifi/authorization/util/IdentityMappingUtil.java
+++ b/nifi-commons/nifi-security-identity/src/main/java/org/apache/nifi/authorization/util/IdentityMappingUtil.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static org.apache.nifi.util.NiFiProperties.SECURITY_GROUP_MAPPING_PATTERN_PREFIX;
 import static org.apache.nifi.util.NiFiProperties.SECURITY_GROUP_MAPPING_TRANSFORM_PREFIX;
@@ -140,8 +141,12 @@ public class IdentityMappingUtil {
      * @return the mapped identity, or the same identity if no mappings matched
      */
     public static String mapIdentity(final String identity, List<IdentityMapping> mappings) {
-        for (IdentityMapping mapping : mappings) {
-            Matcher m = mapping.getPattern().matcher(identity);
+        final List<String> keys = mappings.stream().map(m -> m.getKey()).collect(Collectors.toList());
+        Collections.sort(keys);
+
+        for (final String key : keys) {
+            final IdentityMapping mapping = mappings.stream().filter(m -> m.getKey().equals(key)).findFirst().orElseThrow();
+            final Matcher m = mapping.getPattern().matcher(identity);
             if (m.matches()) {
                 final String pattern = mapping.getPattern().pattern();
                 final String replacementValue = escapeLiteralBackReferences(mapping.getReplacementValue(), m.groupCount());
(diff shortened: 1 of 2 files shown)


nifi-commons/nifi-security-identity/src/main/java/org/apache/nifi/authorization/util/IdentityMappingUtil.java
```diff
@@ -140,8 +141,12 @@ public int compare(IdentityMapping m1, IdentityMapping m2) {
      * @return the mapped identity, or the same identity if no mappings matched
      */
     public static String mapIdentity(final String identity, List<IdentityMapping> mappings) {
-        for (IdentityMapping mapping : mappings) {
-            Matcher m = mapping.getPattern().matcher(identity);
+        final List<String> keys = mappings.stream().map(m -> m.getKey()).collect(Collectors.toList());
+        Collections.sort(keys);
+
+        for (final String key : keys) {
+            final IdentityMapping mapping = mappings.stream().filter(m -> m.getKey().equals(key)).findFirst().orElseThrow();
```
exceptionfactory, line 148: Instead of iterating over the keys and then searching for the matching IdentityMapping, did you consider creating a `TreeMap` of the key and the `IdentityMapping` object? That would avoid the need for this stream, filter, and findFirst approach.

nifi-commons/nifi-security-identity/src/test/java/org/apache/nifi/authorization/util/IdentityMappingUtilTest.java
```diff
@@ -0,0 +1,156 @@
...
+        final List<IdentityMapping> identityMappings = IdentityMappingUtil.getIdentityMappings(niFiProperties);
+
+        final Optional<IdentityMapping> dnMapping = identityMappings.stream().filter(im -> im.getKey().equals("dn")).findFirst();
+        final Optional<IdentityMapping> ldapMapping = identityMappings.stream().filter(im -> im.getKey().equals("ldap")).findFirst();
+
+        Assertions.assertTrue(dnMapping.isPresent());
+        Assertions.assertEquals("NONE", dnMapping.get().getTransform().name());
+
+        Assertions.assertTrue(ldapMapping.isPresent());
+        Assertions.assertEquals("UPPER", ldapMapping.get().getTransform().name());
```
exceptionfactory, line 45: Assertions should use static imports for readability.

nifi-commons/nifi-security-identity/src/test/java/org/apache/nifi/authorization/util/IdentityMappingUtilTest.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    // NIFI-13409 TC1
```
exceptionfactory, line 48: Although it can be useful to know the Jira issue that prompted this change, that information is already available in the commit message itself, so I recommend removing the comments that reference the Jira issue.

nifi-commons/nifi-security-identity/src/test/java/org/apache/nifi/authorization/util/IdentityMappingUtilTest.java
```diff
@@ -0,0 +1,156 @@
...
+    @Test
+    public void testMappingIdentity() {
+        final Map<String, String> properties = new HashMap<>();
+        properties.putAll(getDnProperties());
+        properties.putAll(getLdapProperties("UPPER"));
+        final NiFiProperties niFiProperties = new NiFiProperties(properties);
+
+        final List<IdentityMapping> identityMappings = IdentityMappingUtil.getIdentityMappings(niFiProperties);
+        Assertions.assertEquals("nifi-node1", IdentityMappingUtil.mapIdentity("CN=nifi-node1, ST=MD, C=US", identityMappings));
+        Assertions.assertEquals("NIFIADMIN", IdentityMappingUtil.mapIdentity("nifiadmin", identityMappings));
```
exceptionfactory, line 58: As the identity and mapped identity values are repeated across multiple methods, it would be helpful to create static variables that can be reused across test methods.

exceptionfactory (changes requested):
Thanks for working on this issue @simonbence. The general change makes sense, having expected ordering is helpful for these mappings. I noted one recommendation on the implementation, and several small adjustments to the test class.

exceptionfactory:
@simonbence Just following up on this, glad to help bring this to completion, so let me know if you have any questions about the comments provided earlier.

nifi-commons/nifi-security-identity/src/main/java/org/apache/nifi/authorization/util/IdentityMappingUtil.java
simonbence: I did not think of this, but a good idea. For the whole picture, I was hesitant with the change because while the `TreeMap` avoids the constructs above (which in this case as we are not talking about bigger data structures, might not be have as much affect on performance, using an `entrySet` [1], where the contract would allow too much freedom on the order of the elements (even if the implementation provides further restrictions) seems like a questionable move forward. In the end, after weighting the implications, I decided to go with the `TreeMap`. I will commit the changes soonish.

[1] using the `keySet` would result the same, and the `forEach` looks to be a bit convoluted here.

nifi-commons/nifi-security-identity/src/test/java/org/apache/nifi/authorization/util/IdentityMappingUtilTest.java
simonbence: The ticket number itself is kind of a duplication of the commit message. The primary peace of information is the "TC1" (2...5) which supposed to connect the given test method with a given test case from the ticket, as there is a 1to1 mapping between them.

This is not a mandatory information but I found it useful for later usage. Adding only "TC1" to the comment would not suggest any relation to anything outside the test class, so I decided to keep it a bit verbosive.

Please share your view on this, thanks!

exceptionfactory: Thanks for the additional note regarding the test case scenario. That's helpful, so in that case, leaving the comment sounds good.

simonbence:
Hi @exceptionfactory !

For full disclosure, it looks like there is already an ordering within the mapping process: [here](https://github.com/apache/nifi/blob/7ade77ac22fc2e5bbd38151e32c96667020f30a8/nifi-commons/nifi-security-identity/src/main/java/org/apache/nifi/authorization/util/IdentityMappingUtil.java#L125)

I take a looks of the usages and it seems to me that not all usages of the method is based (directly) on the result of the `getMapping` ([example](https://github.com/apache/nifi/blob/83b701a25ea2820d20768ca4946f5698249beb0a/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/IdentityAuthenticationProvider.java#L92), so I could not safely say that the original ordering is enough.

The question comes however if with this addition we still need the original ordering or not. I would not disturb it within this ticket but later on it might be a reasonable move to remove that (after considering the possible implications)

exceptionfactory:
Thanks for following up @simonbence. That's a good point about values already being sorted. The other reference you noted is for NiFi Registry, which is a different class. With the existing sorting, have you traced down where the sorting is not being applied on NiFi itself? It would seem to render the current set of changes unnecessary.

exceptionfactory (changes requested):
@simonbence On further review, this change does not seem like the right approach. References to `mapIdentity()` occur in association with calls to get identity mappings from this same class, which already implements sorting. The Jira issue does not provide the precise code path for the issue, so further evaluation of the problem is necessary. The `mapIdentity()` method should not perform sorting when the list should already be sorted, so tracking down the source of unexpected sorting seems to be the next step.

I'm closing this pull request for now, but feel free to revisit after confirming the code path that exhibits the behavior described in the Jira issue.

PR #9116, opened 2024-07-26
NIFI-13587 remove RestrictedSSLContextService interface in favor of t…

markobean:
…he identical SSLContextService

           rename StandardRestrictedSSLContextService to more concise name RestrictedSSLContextService

# Summary

This PR may be somewhat controversial since it renames a heavily used controller service which is clearly a breaking change. However, as we near the release of 2.0.0, now is an appropriate time to make such a change.

It has always seemed odd that there is a service named StandardRestrictedSSLContextService with both "standard" and "restricted" in its name. Historically, there is logic to this naming. There is an interface, RestrictedSSLContextService, which at one time was distinctly different from the SSLContextService interface. However, the two are now identical.

This PR proposes to consolidate naming conventions. The RestrictedSSLContextService interface has been removed. All references to it have been updated to SSLContextService. Removing the interface also allows the implementation name to become RestrictedSSLContextService rather than StandardRestrictedSSLContextService.

There are no functional changes here. Rather this only updates the naming of interfaces and corresponding implementations.

[NIFI-13587](https://issues.apache.org/jira/browse/NIFI-13587)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


joewitt:
This is not workable.  The disruption this causes is not met by gain. 

I am not fond of the current naming either but the breakage of basically every flow isn't great.  We've had some cases of breakage that are specific to certain components and that creates enough challenge.

This one is just too much.  

Assuming there is agreement on what the names should be the next question is how best to transition people. This needs to be automated to be worthwhile.

markobean:
I understand (and expected) the argument against this. Would a more appropriate approach? Rename the interface (internal, non-user facing change) and create a new controller service "RestrictedSSLContextService" while marking the current StandardRestrictedSSLContextService as deprecated - yet keep it fully functional.

This approach does beg the question of when then would it be appropriate to remove the deprecated service - only when we get to a future 3.0?

joewitt:
I haven't looked into the codebase fully myself.  

If we have something now on the 2.x line which is deprecated then yeah we gotta get rid of that.

If you're saying we want to deprecate something that isn't now then that means deprecate on 1.x and remove/change on 2.x.  We've done that pattern with a number of things.  But for something like this we really want to make migration as automatic as possible.

I have a couple ideas on how to do that but dont want to blurt them out as I've not thought deeply on this.  But if this change is happening - we need options explored.

markobean:
I'll cancel this PR and give it some more thought. There's the possibility of introducing a migration script to be run on flows when upgrading to 2.x which would translate the implementation name. That's at least half the problem solved, but still leaves an issue for saved flows in Registry or flow definition files. Perhaps there's a way to internally handle this translation that happens - as you say - automatically. Open to thoughts, if you have some already. If so, feel free to update the Jira issue.

PR #9117, opened 2024-07-26
NIFI-13017: Replace JoltTransformJSON custom UI

mcgilman:
NIFI-13017:
- Remove old JoltTransformJSON custom UI.
- Update build to use new JoltTransformJSON custom UI.
- Fix revision query parameter.


scottyaslan (commented):
A bit unrelated but I noticed that there is an error message the flashes on initial load of the Jolt Custom UI. This is a separate issue and I logged https://issues.apache.org/jira/browse/NIFI-13589 to address that.

rfellows (approved):
Excellent work @mcgilman! Nice to finally get rid of AngularJS!

scottyaslan (approved):

scottyaslan:
I verified that purple mode works well and that the browser back button navigation is fixed. Nice work @mcgilman!

PR #9126, opened 2024-07-30
NIFI-13601 Add code-coverage workflow

exceptionfactory:
# Summary

[NIFI-13601](https://issues.apache.org/jira/browse/NIFI-13601) Adds a `code-coverage` workflow to the set of GitHub Action workflows to handle running both unit tests and integration tests on a daily schedule. The `code-coverage` workflow also handles publishing code coverage metrics to Codecov, removing this step for the `ci-workflow`.

Additional changes include removing the scheduled settings for other test workflows as this workflow covers all types of tests and provides a daily status.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,0 +1,70 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+name: code-coverage
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+  pull_request:
+    paths:
+      - '.github/workflows/code-coverage.yml'
+
+env:
+  DEFAULT_MAVEN_OPTS: >-
+    -Xms6g
+    -Xmx6g
+    -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN
+
+permissions:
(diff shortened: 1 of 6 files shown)


.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          python-version: '3.10'
+      - name: Maven Build
+        env:
+          MAVEN_OPTS: >-
+            ${{ env.DEFAULT_MAVEN_OPTS }}
+        run: >
+          ./mvnw --fail-fast --no-snapshot-updates --no-transfer-progress --show-version
+          -D include-python-integration-tests=true
+          -P integration-tests
```
ChrisSamo632, line 61: Should this include the `nifi-registry-integration-tests` profile as well so that we capture associated code coverage?

.github/workflows/docker-tests.yml
```diff
@@ -16,8 +16,6 @@
 name: docker-tests
 
 on:
-  schedule:
-    - cron: "0 2 * * *"
```
ChrisSamo632, line 20: For consideration (maybe not a blocker) - the docker tests will now only trigger if the docker files themselves or application assemblies are changed.

While this will assure any such changes don't break things, we'll miss the possibility of assuring that changes within other modules haven't broken the assembly and running of the applications, e.g. if a change is made to `nifi.properties` but not catered for in the docker build and that subsequently prevents the application from running

I can't say for certain that such a thing would be a pertinent, and I agree that reducing the number of GitHub Actions is a good thing, I wonder whether a reduction in frequency (e.g. weekly) rather than removal of the cron schedule would be more sensible in order to retain more of the existing QA

.github/workflows/code-coverage.yml
exceptionfactory: Yes, that's a good suggestion!

.github/workflows/docker-tests.yml
exceptionfactory: That's a good question. At this point, the Docker integration test script is a minimal check that the REST API is accessible. I think changing this to a weekly schedule is a good change for now.

exceptionfactory:
Thanks for the feedback @ChrisSamo632! I pushed a commit adding the NiFi Registry integration tests profile to the code-coverage workflow and adjusted the docker-tests workflow to run once a week.

ChrisSamo632 (approved):
LGTM 👍 

Thanks for the change @exceptionfactory , will merge

PR #9130, opened 2024-07-31
NIFI-13606: Ensuring that the custom ui uri contains the proxy path when appropriate

mcgilman:
NIFI-13606:
- Ensuring that the custom ui uri contains the proxy path when appropriate.


exceptionfactory (approved):
Thanks for the correction @mcgilman, this works as expected. +1 merging

PR #9132, opened 2024-07-31
NIFI-13610: Fix conditions controlling the availability of new Controller Service (in flow), new Parameter Context, and Move to Parent Group

mcgilman:
NIFI-13610:
- Fix conditions controlling the availability of new Controller Service (in flow), new Parameter Context, and Move to Parent Group.


rfellows:
reviewing

rfellows (approved):
These are good changes to get in soon! Great work @mcgilman.

PR #9133, opened 2024-07-31
[NIFI-13621] - UI - handle backend errors when hitting the search api

rfellows:
[NIFI-13621](https://issues.apache.org/jira/browse/NIFI-13621)

If there is an error from the search API, handle it by showing the error in a snackbar or fullscreen depending on the type.

Also fixed an issue with escape closing the search results and then not being able to open them back up again.


mcgilman:
Will review...

mcgilman (approved):

PR #9136, opened 2024-08-01
NIFI-13614: populate failure cause in ack response

KalmanJantner:
# Summary

[NIFI-13614](https://issues.apache.org/jira/browse/NIFI-13614)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/c2/c2-client-bundle/c2-client-service/src/main/java/org/apache/nifi/c2/client/service/operation/C2OperationHandler.java
+++ b/c2/c2-client-bundle/c2-client-service/src/main/java/org/apache/nifi/c2/client/service/operation/C2OperationHandler.java
@@ -20,14 +20,18 @@ package org.apache.nifi.c2.client.service.operation;
 import static java.util.Optional.ofNullable;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.nifi.c2.protocol.api.C2Operation;
 import org.apache.nifi.c2.protocol.api.C2OperationAck;
 import org.apache.nifi.c2.protocol.api.C2OperationState;
+import org.apache.nifi.c2.protocol.api.FailureCause;
 import org.apache.nifi.c2.protocol.api.OperandType;
 import org.apache.nifi.c2.protocol.api.OperationType;
 import org.apache.nifi.c2.serializer.C2Serializer;
+import org.apache.nifi.minifi.validator.ValidationException;
 
 /**
  * Handler interface for the different operation types
@@ -79,11 +83,36 @@ public interface C2OperationHandler {
      * @param details        additional status info to detail the state
      * @return the created state
      */
-    default C2OperationState operationState(C2OperationState.OperationState operationState, String details) {
+    default C2OperationState operationState(C2OperationState.OperationState operationState, String details, Exception e) {
         C2OperationState state = new C2OperationState();
         state.setState(operationState);
         state.setDetails(details);
+        ofNullable(e).map(this::toFailureCause).ifPresent(state::setFailureCause);
         return state;
+    }
+
+    private FailureCause toFailureCause(Exception exception) {
+        FailureCause failureCause = new FailureCause();
+        failureCause.setExceptionMessage(exception.getMessage());
+        List<String> causeList = new LinkedList<>();
+        populateCausedChain(ofNullable(exception.getCause()), causeList);
+        failureCause.setCausedByMessages(causeList);
+        if (exception instanceof ValidationException validationException) {
+            failureCause.setValidationResults(validationException.getValidationResults());
+        }
+        return failureCause;
+    }
+
+    private List<String> populateCausedChain(Optional<Throwable> cause, List<String> causeList) {
+        cause.ifPresent(c -> {
+            causeList.add(c.getMessage());
+            populateCausedChain(cause.map(Throwable::getCause), causeList);
+        });
+        return causeList;
+    }
+
+    default C2OperationState operationState(C2OperationState.OperationState operationState, String details) {
+        return operationState(operationState, details, null);
     }
 
     /**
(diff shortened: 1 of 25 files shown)


minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/main/java/org/apache/nifi/minifi/c2/command/DefaultUpdateConfigurationStrategy.java
```diff
@@ -148,7 +150,7 @@ private void reloadFlow(Set<String> proposedConnectionIds) throws IOException {
         List<ValidationResult> validationErrors = validate(flowController.getFlowManager());
         if (!validationErrors.isEmpty()) {
             LOGGER.error("Validation errors found when reloading the flow: {}", validationErrors);
-            throw new IllegalStateException("Unable to start flow due to validation errors");
+            throw new ValidationException(validationErrors);
```
briansolo1985, line 153: I would keep the error message here as well.
Otherwise when creating the operation response, the exception message will be null. It is more informative for the user if they get a message that the flow did not start due to a validation error.

For the the ValidationException class also needs to be modified:

```
public class ValidationException extends IllegalStateException {
    private List<ValidationResult> validationResults;

    public ValidationException(String message, List<ValidationResult> details) {
        super(message);
        this.validationResults = details;
    }
```

c2/c2-protocol/c2-protocol-api/src/main/java/org/apache/nifi/c2/protocol/api/FailureCause.java
```diff
@@ -0,0 +1,65 @@
...
+package org.apache.nifi.c2.protocol.api;
+
+import java.util.List;
+import java.util.Objects;
+import org.apache.nifi.components.ValidationResult;
+
+public class FailureCause {
+    private List<ValidationResult> validationResults;
+    private String exceptionMessage;
+    private String causedByMessage;
```
briansolo1985, line 27: The causedByMessage should be a collection instead of a string.
In many cases exception are recursively chained through Throwable.cause, and often the real cause is the exception on the bottom layer.
We need to process the cause exception recursively and store them into a list to keep all the information.
The list is fortunately an ordered collection so the order of the exception will also be preserved in this way

briansolo1985 (changes requested):
Thank you for the pull request. This is really an useful feature, which will provide more insight of the status of the C2 operation when an error happens.

I left some comments, please check them.
Also two general comments:
* we should also apply this approach for all the the handlers, not just UpdateConfigurationOperationHandler. To use this approach properly, potentially some refactoring will be needed on handler level and below. For example, in UpdateAssetOperation handler the exception is logged, an optional is returned only. We should move away from the optional approach in favor of exception managed approach like in UpdateConfigurationOperationHandler. 
* Please squash the commits. Before the final review, there should only be one commit

c2/c2-client-bundle/c2-client-service/src/main/java/org/apache/nifi/c2/client/service/operation/C2OperationHandler.java
```diff
@@ -79,13 +81,26 @@ default boolean requiresRestart() {
      * @param details        additional status info to detail the state
      * @return the created state
      */
-    default C2OperationState operationState(C2OperationState.OperationState operationState, String details) {
+    default C2OperationState operationState(C2OperationState.OperationState operationState, String details, Exception e) {
```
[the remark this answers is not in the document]
KalmanJantner: Fixed, thanks.

c2/c2-client-bundle/c2-client-service/src/main/java/org/apache/nifi/minifi/validator/ValidationException.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public ValidationException(List<ValidationResult> details) {
+        this.validationResults = details;
+    }
+
+    public List<ValidationResult> getValidationResults() {
+        return validationResults;
+    }
+
+    public void setValidationResults(List<ValidationResult> validationResults) {
```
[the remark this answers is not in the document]
KalmanJantner: Fixed, thanks.

c2/c2-protocol/c2-protocol-api/src/main/java/org/apache/nifi/c2/protocol/api/C2OperationState.java
```diff
@@ -42,6 +42,17 @@ public class C2OperationState implements Serializable {
     @Schema(description = "Additional details about the state")
     private String details;
 
+    @Schema(description = "Additional details cause of the failure")
```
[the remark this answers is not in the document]
KalmanJantner: Fixed, thanks.

minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-framework-core/src/main/java/org/apache/nifi/minifi/c2/command/DefaultUpdateConfigurationStrategy.java
KalmanJantner: Fixed, thanks.

c2/c2-protocol/c2-protocol-api/src/main/java/org/apache/nifi/c2/protocol/api/FailureCause.java
KalmanJantner: Fixed, thanks.

KalmanJantner:
> Thank you for the pull request. This is really an useful feature, which will provide more insight of the status of the C2 operation when an error happens.
> 
> I left some comments, please check them. Also two general comments:
> 
> * we should also apply this approach for all the the handlers, not just UpdateConfigurationOperationHandler. To use this approach properly, potentially some refactoring will be needed on handler level and below. For example, in UpdateAssetOperation handler the exception is logged, an optional is returned only. We should move away from the optional approach in favor of exception managed approach like in UpdateConfigurationOperationHandler.
> * Please squash the commits. Before the final review, there should only be one commit

Thanks you for the review. I applied the recommended changes and squashed the commits.

briansolo1985 (approved):
Thanks for incorporating the suggested changes. LGTM
Will proceed with merging

PR #9138, opened 2024-08-01
NIFI-12898 Allow uploading asset and referencing via Parameter

bbende:
# Summary

This pull request adds the ability to upload assets to parameter contexts and then reference them from parameters in the context. This feature is described in [NIFI-12898](https://issues.apache.org/jira/browse/NIFI-12898) and the [feature proposal](https://cwiki.apache.org/confluence/display/NIFI/Asset+Management).

A new framework extension - `AssetManager` -  is responsible for storing/retrieving assets on the file system. 

The following REST end-points are added:

Create/upload an asset:
`POST /parameter-contexts/{id}/assets`  

List assets:
`GET  /parameter-contexts/{id}/assets`

Get asset content:
`GET  /parameter-contexts/{id}/assets/{assetId}`

The last two end-points are primarily intended to be used when a node joins the cluster for it to compare it's current assets with the coordinator and synchronize it's assets to match.

To reference an asset it is done through a normal parameter context update, where a parameter will send no value and instead provide an asset reference:
```
referencedAssets: [
  {
     "id": "assetId1"
  },
 {
     "id": "assetId2"
  },
]
```

The back-end will automatically make the value of the parameter become the concatenated paths of the referenced assets.

The following CLI commands were added:

```
nifi create-asset -pcid <param-context-id> -af <full-path-to-a-file>

nifi list-assets -pcid <param-context-id>

nifi add-asset-reference -pcid <param-context-id> -pn <param-name> -aid <asset-id>

nifi remove-asset-reference -pcid <param-context-id> -pn <param-name> -aid <asset-id>
```

When a parameter context is updated, any asset that was previously referenced by a parameter and no longer referenced after the update will be deleted.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


exceptionfactory (commented):
Thanks for putting together this new feature @bbende! I plan to take a closer look soon. 

GitHub CodeQL flagged several issues related to using input parameters for file names, so that will require some closer evaluation.

bbende:
Thanks @exceptionfactory ! I noticed the CodeQL issues as well...

I did sanitize the Filename header value before passing it to any code that would use it to write a File. 

If that isn't enough then I suppose a fall back option would be to write the file names with the generated UUID and store some kind of mapping file along side the asset that contains the original filename. 

The problem with this is that the value of the parameter that references the asset is going to be the concatenated paths of the assets and it's going to contain file names that the user has no idea what they are. Example, upload PostgresDriver.jar and it generates id 123, then reference this asset in a parameter named db_driver, and the value of the parameter will be '/path/to/assets/123'. So some looking at the parameter has no idea what file that is.

exceptionfactory:
Thanks for the reply @bbende. Several classes mitigate the path traversal problem using Java Path objects with normalization and checking that the resolved child path starts with the selected parent path.

https://github.com/apache/nifi/blob/83d2078a4f878ba742f8adba2da382fb117ad6c0/nifi-system-tests/nifi-system-test-extensions-bundle/nifi-system-test-extensions/src/main/java/org/apache/nifi/flow/registry/FileSystemFlowRegistryClient.java#L434

It appears that the parent Parameter Context ID also needs to be checked. Instead of using the user-provided value, recommend looking up the Parameter Context and using the ID returned.

bbende:
@exceptionfactory thanks for the pointers! I made a couple of improvements and seems to have resolved the CodeQL issues. I think most of them were actually somewhat false positives because the sanitize method was returning the input variable if null or blank, but that was causing CodeQL to think the unmodified user input was being potentially used later on.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ParameterContextResource.java
```diff
@@ -323,6 +356,204 @@ public Response updateParameterContext(
...
+        });
+
+        // If we need to replicate the request, we do so using the Upload Request Replicator, rather than the typical replicate() method.
+        // This is because Upload Request Replication works differently in that it needs to be able to replicate the InputStream multiple times,
+        // so it must create a file on disk to do so and then use the file's content to replicate the request. It also bypasses the two-phase
+        // commit process that is used for other requests because doing so would result in uploading the file twice to each node or providing a
+        // different request for each of the two phases.
+
+        final long startTime = System.currentTimeMillis();
+        final InputStream maxLengthInputStream = new MaxLengthInputStream(assetContents, (long) DataUnit.GB.toB(1));
```
exceptionfactory, line 432: For tracking purposes, it would be helpful to promote the maximum length value to a static value. Although this could be made configurable, it seems unlikely to change, and 1 GB seems like it should be large enough to handle anything of reasonable size.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ParameterContextResource.java
```diff
@@ -323,6 +356,204 @@ public Response updateParameterContext(
...
+                    .header(FILENAME_HEADER, assetName)
+                    .header(CONTENT_TYPE_HEADER, UPLOAD_CONTENT_TYPE)
+                    .exampleRequestUri(getAbsolutePath())
+                    .responseClass(AssetEntity.class)
+                    .successfulResponseStatus(HttpResponseStatus.OK.getCode())
+                    .build();
+            assetEntity = uploadRequestReplicator.upload(uploadRequest);
+        } else {
+            final String existingContextId = contextEntity.getId();
+            final String sanitizedAssetName = FileUtils.sanitizeFilename(assetName);
```
exceptionfactory, line 450: Instead of simply sanitizing the filename, should an input filename be rejected if it contains unexpected characters?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
```diff
@@ -1509,6 +1512,26 @@ private ParameterProviderConfigurationEntity createParameterProviderConfiguratio
...
+         return entity;
+   }
+
+   public AssetDTO createAssetDto(final Asset asset) {
+       final File assetFile = asset.getFile();
+       final AssetDTO dto = new AssetDTO();
+       dto.setId(asset.getIdentifier());
+       dto.setName(asset.getName());
+       dto.setDigest(asset.getDigest().orElse(null));
+       dto.setMissingContent(!assetFile.exists());
```
exceptionfactory, line 1527: What do you think about naming this property `contentNotFound` instead of `missingContent`?

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/client/nifi/ParamContextClient.java
```diff
@@ -40,4 +43,10 @@ public interface ParamContextClient {
 
     ParameterContextUpdateRequestEntity deleteParamContextUpdateRequest(String contextId, String updateRequestId) throws NiFiClientException, IOException;
 
+    AssetEntity createAsset(String contextId, String assetName, File file) throws NiFiClientException, IOException;
+
+    AssetsEntity getAssets(String contextId) throws NiFiClientException, IOException;
+
+    File getAssetContent(String contextId, String assetId, File outputDirectory) throws NiFiClientException, IOException;
```
exceptionfactory, line 50: What do you think about returning a `Path` instead of `File` for this method?

exceptionfactory (commented):
Great work on this new feature @bbende! The majority of the code changes look good, and the initial changes to address the CodeQL path concerns also look good. I plan to do some additional testing, but on initial review, I just noted a couple minor things.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ParameterContextResource.java
bbende: That probably does make sense, will play around with that.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
bbende: I don't have a strong preference, it seemed like we used "missing" for other stuff like components that are ghosted.

bbende:
@exceptionfactory I pushed a commit that addresses the comments you had, and also another commit that adds a delete end-point to be able to clean up assets that were uploaded and maybe never referenced, let me know of anything else

markobean:
Is this new feature accessible only via the API and not in the GUI?

bbende:
@markobean at the moment yes, if someone picks up a UI JIRA to add it then it may be in the UI later on

exceptionfactory (approved):
Thanks for building out these new features @bbende, the latest version looks good! +1 merging

PR #9149, opened 2024-08-05
[NIFI-13631] update purple theme to reference M2 color palettes

scottyaslan:
https://issues.apache.org/jira/browse/NIFI-13631


mcgilman (approved):
Thanks @scottyaslan!

PR #9151, opened 2024-08-06
NIFI-13630 Handle map Avro type in BigQuery writing

juldrixx:
# Summary

[NIFI-13630](https://issues.apache.org/jira/browse/NIFI-13630)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
+++ b/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
@@ -24,10 +24,8 @@ import com.google.protobuf.Descriptors;
 import com.google.protobuf.DynamicMessage;
 
 import java.math.BigDecimal;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -52,14 +50,12 @@ public class ProtoUtils {
                    if (value instanceof Object[] arrayValue) {
                        valueMaps = Arrays.stream(arrayValue)
                                .map(item -> (Map<String, Object>) item).toList();
-                   } else if (value instanceof HashMap<?, ?> hashMapValue) {
-                       valueMaps = new ArrayList<>();
-                       for (Map.Entry<?, ?> entry : hashMapValue.entrySet()) {
-                           Map<String, Object> map = new HashMap<>();
-                           map.put("key", entry.getKey());
-                           map.put("value", entry.getValue());
-                           valueMaps.add(map);
-                       }
+                   } else if (value instanceof Map<?, ?> mapValue) {
+                       valueMaps = mapValue.entrySet().stream()
+                               .map(entry -> Map.of(
+                                       "key", entry.getKey(),
+                                       "value", entry.getValue()
+                               )).toList();
                    } else {
                        valueMaps = (Collection<Map<String, Object>>) value;
                    }
--- a/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
+++ b/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
@@ -24,8 +24,10 @@ import com.google.protobuf.Descriptors;
 import com.google.protobuf.DynamicMessage;
 
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -46,7 +48,21 @@ public class ProtoUtils {
            switch (field.getType()) {
            case MESSAGE:
                if (field.isRepeated()) {
-                   Collection collection = value.getClass().isArray() ? Arrays.asList((Object[]) value) : (Collection) value;
+                   Collection collection = null;
+                   if (value.getClass().isArray()) {
+                       collection = Arrays.asList((Object[]) value);
+                   } else if (value instanceof HashMap) {
+                       collection = new ArrayList();
+                       for (Object entryObj : ((HashMap<?, ?>) value).entrySet()) {
+                           Map.Entry<?, ?> entry = (Map.Entry<?, ?>) entryObj;
+                           Map<String, Object> map = new HashMap<>();
+                           map.put("key", entry.getKey());
+                           map.put("value", entry.getValue());
+                           collection.add(map);
+                       }
+                   } else {
+                       collection = (Collection) value;
+                   }
                    collection.forEach(act -> builder.addRepeatedField(field, createMessage(field.getMessageType(), (Map<String, Object>) act, tableSchema)));
                } else {
                    builder.setField(field, createMessage(field.getMessageType(), (Map<String, Object>) value, tableSchema));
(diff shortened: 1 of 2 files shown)


nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
```diff
@@ -46,7 +48,21 @@ public static DynamicMessage createMessage(Descriptors.Descriptor descriptor, Ma
            switch (field.getType()) {
            case MESSAGE:
                if (field.isRepeated()) {
-                   Collection collection = value.getClass().isArray() ? Arrays.asList((Object[]) value) : (Collection) value;
+                   Collection collection = null;
```
EndzeitBegins, line 51: As all possible code paths provide a value for this, it can be declared `final` and need not to be initialised.

And while you didn't choose the name `collection` here I'd like to use this change as an opportunity to provide a more meaningful name for this variable.

All code paths should lead to a collection of `Map<String, Object>` objects, thus we may as well define the type here instead of using a raw collection.

```suggestion
                   final Collection<Map<String, Object>> valueMaps;
```

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
```diff
@@ -46,7 +48,21 @@ public static DynamicMessage createMessage(Descriptors.Descriptor descriptor, Ma
...
+                   if (value.getClass().isArray()) {
+                       collection = Arrays.asList((Object[]) value);
```
EndzeitBegins, line 53: The original code casts all types of array into an `Object[]`. This only works for arrays with non-primitive items and thus the explicit `isArray` check can be replaced with a simpler (and more useful) `instanceof`.

This way we can use pattern-matching available in more modern Java and avoid an explicit additional cast.

```suggestion +1
                   if (value instanceof Object[] arrayValue) {
                       valueMaps = Arrays.stream(arrayValue)
                               .map(item -> (Map<String, Object>) item).toList();
```

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
```diff
@@ -46,7 +48,21 @@ public static DynamicMessage createMessage(Descriptors.Descriptor descriptor, Ma
...
+                           Map.Entry<?, ?> entry = (Map.Entry<?, ?>) entryObj;
+                           Map<String, Object> map = new HashMap<>();
+                           map.put("key", entry.getKey());
+                           map.put("value", entry.getValue());
+                           collection.add(map);
+                       }
+                   } else {
+                       collection = (Collection) value;
+                   }
                    collection.forEach(act -> builder.addRepeatedField(field, createMessage(field.getMessageType(), (Map<String, Object>) act, tableSchema)));
```
EndzeitBegins, line 66: As I've changed the type of the former `collection` field, the unchecked cast must be altered slightly.

However, we can drop the cast inside the `foreach` this way.

```suggestion
                   } else {
                       valueMaps = (Collection<Map<String, Object>>) value;
                   }

                   valueMaps.forEach(act -> builder.addRepeatedField(field, createMessage(field.getMessageType(), act, tableSchema)));
```

EndzeitBegins (changes requested):
Thank you for tackling this @juldrixx.

I've got some recommendations based on code style solely but overall the changes look good to me.
Extra thanks for adding a reproducing test case, that's always appreciated and helpful.

juldrixx:
Thanks @EndzeitBegins for the review, should be good now. Let me know if there are any other changes to make.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
```diff
@@ -46,8 +48,22 @@ public static DynamicMessage createMessage(Descriptors.Descriptor descriptor, Ma
            switch (field.getType()) {
            case MESSAGE:
                if (field.isRepeated()) {
-                   Collection collection = value.getClass().isArray() ? Arrays.asList((Object[]) value) : (Collection) value;
-                   collection.forEach(act -> builder.addRepeatedField(field, createMessage(field.getMessageType(), (Map<String, Object>) act, tableSchema)));
+                   final Collection<Map<String, Object>> valueMaps;
+                   if (value instanceof Object[] arrayValue) {
+                       valueMaps = Arrays.stream(arrayValue)
+                               .map(item -> (Map<String, Object>) item).toList();
+                   } else if (value instanceof HashMap<?, ?> hashMapValue) {
```
dan-s1, line 55: If value is a Map, is it always guaranteed to be specifically a `HashMap`?

EndzeitBegins: Good catch @dan-s1 . 
I also wanted to address that but didn't find the time to answer until now. 

I don't think we should assume that. Also there's no benefit of doing so, as we only ever use methods of `Map<?, ?>` anyways. 

```suggestion
                   } else if (value instanceof Map<?, ?> mapValue) {
```

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/bigquery/proto/ProtoUtils.java
```diff
@@ -46,8 +48,22 @@ public static DynamicMessage createMessage(Descriptors.Descriptor descriptor, Ma
...
+                       valueMaps = new ArrayList<>();
+                       for (Map.Entry<?, ?> entry : hashMapValue.entrySet()) {
+                           Map<String, Object> map = new HashMap<>();
+                           map.put("key", entry.getKey());
+                           map.put("value", entry.getValue());
+                           valueMaps.add(map);
+                       }
```
EndzeitBegins, line 62: This code is totally fine, but we could simplify it further by using `Map.of` as introduced in Java 9.

```suggestion
                       valueMaps = new ArrayList<>();
                       for (Map.Entry<?, ?> entry : hashMapValue.entrySet()) {
                           valueMaps.add(Map.of(
                                   "key", entry.getKey(),
                                   "value", entry.getValue()
                           ));
                       }
```

Alternatively we could replace the imperative `for` loop with a more functional approach, but that's just down to personal preference I assume.

```suggestion
                       valueMaps = mapValue.entrySet().stream()
                               .map(entry -> Map.of(
                                       "key", entry.getKey(),
                                       "value", entry.getValue()
                               )).toList();
```

EndzeitBegins (changes requested):
Thank you for incorporating the feedback so quickly, @juldrixx.

@dan-s1 and I found a small point for improvement regarding the ìnstanceof` expression that we should resolve before this gets merged.

Also I've added a proposal for a small stylistic change, but that's up to you.

juldrixx:
Made the changes @EndzeitBegins  & @dan-s1. Thanks for the reviews.

EndzeitBegins (approved):
Looks good to me. Thank you for the contribution @juldrixx. 👍🏻

exceptionfactory (changes requested):
Thanks for working on this improvement @juldrixx, and thanks for the reviews @EndzeitBegins.

The only area for improvement I noticed was the string concatenation for the JSON examples. With current Java support for [multi-line strings](https://www.baeldung.com/java-multiline-string) reformatting the JSON examples would make it easier to read.

With that adjusted, the implementation itself looks close to completion.

exceptionfactory (approved):
@juldrixx Given that the suggestion was just reformatting in a test class, I pushed a commit with the change and will proceed to merge. Thanks again for your contribution!

PR #9153, opened 2024-08-06
NIFI-13636: Fixing back navigation when navigating to manage access p…

mcgilman:
…olicies.


rfellows (approved):
Thanks @mcgilman! This is a good fix.

PR #9154, opened 2024-08-06
[NIFI-13637] add padding to Jolt UI expansion panels

scottyaslan:
Screenshot 2024-08-06 at 2 12 56 PM


rfellows (approved):
Thanks for keeping things consistent, @scottyaslan.

PR #9165, opened 2024-08-07
[NIFI-13645] property table cells clickable even when value is set to…

scottyaslan:
https://issues.apache.org/jira/browse/NIFI-13645


mcgilman:
Reviewing...

mcgilman (approved):
Thanks @scottyaslan!

PR #9171, opened 2024-08-11
NIFI-12855: Add previous event IDs to provenance events to facilitate full graph traversal

mattyb149:
# Summary

[NIFI-12855](https://issues.apache.org/jira/browse/NIFI-12855) This PR adds "previousEventIds" to provenance events. This can help graph-based applications/databases provide correct lineage more quickly (namely, without having to sort the events by timestamp)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/provenance/ProvenanceEventRecord.java
+++ b/nifi-api/src/main/java/org/apache/nifi/provenance/ProvenanceEventRecord.java
@@ -18,6 +18,7 @@ package org.apache.nifi.provenance;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Describes an event that happened to a FlowFile.
@@ -33,6 +34,12 @@ public interface ProvenanceEventRecord {
      * added to the {@link ProvenanceEventRepository}
      */
     long getEventId();
+
+    /**
+     * @return a unique ID for the "parent" Provenance Event, namely the one that came directly before this event
+     * for the given FlowFile. For source events such as CREATE, this value should be set to -1
+     */
+    Set<Long> getPreviousEventIds();
 
     /**
      * @return the time at which this Provenance Event was created, as the
@@ -149,14 +156,14 @@ public interface ProvenanceEventRecord {
     /**
      * @return the UUID's of all Parent FlowFiles. This is applicable only when
      * the {@link ProvenanceEventType} is of type
-     * {@link ProvenanceEventType#SPAWN SPAWN}
+     * {@link ProvenanceEventType#JOIN JOIN}
      */
     List<String> getParentUuids();
 
     /**
      * @return the UUID's of all Child FlowFiles. This is applicable only when
      * the {@link ProvenanceEventType} is of type
-     * {@link ProvenanceEventType#SPAWN SPAWN}
+     * {@link ProvenanceEventType#FORK FORK}
      */
     List<String> getChildUuids();
 
--- a/nifi-api/src/main/java/org/apache/nifi/provenance/ProvenanceEventRepository.java
+++ b/nifi-api/src/main/java/org/apache/nifi/provenance/ProvenanceEventRepository.java
@@ -18,6 +18,7 @@ package org.apache.nifi.provenance;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Set;
 
 /**
  * This Repository houses Provenance Events. The repository is responsible for
@@ -93,5 +94,18 @@ public interface ProvenanceEventRepository {
      */
     void close() throws IOException;
 
+    /**
+     * Returns the previous provenance event IDs for the given FlowFile
+     * @param flowFileUUID the UUID of the FlowFile
+     * @return the previous event IDs for the given FlowFile
+     */
+    Set<Long> getPreviousEventIds(String flowFileUUID);
 
+    /**
+     * Updates the previous provenance event IDs for the given event
+     *
+     * @param record The record for which to update the previous event IDs
+     * @param previousIds the list of previous event IDs to set for the record, or null to remove
+     */
+    void updatePreviousEventIds(ProvenanceEventRecord record, Set<Long> previousIds);
 }
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-database-utils/src/main/java/org/apache/nifi/util/db/JdbcCommon.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-database-utils/src/main/java/org/apache/nifi/util/db/JdbcCommon.java
@@ -122,6 +122,8 @@ public class JdbcCommon {
 
     public static final Pattern LONG_PATTERN = Pattern.compile("^-?\\d{1,19}$");
     public static final Pattern SQL_TYPE_ATTRIBUTE_PATTERN = Pattern.compile("sql\\.args\\.(\\d+)\\.type");
+
+    public static final Pattern JDBC_ARRAY_ELEMENT_CLASS_TYPE_PATTERN = Pattern.compile("JavaType\\(class (.*)\\) ARRAY");
     public static final Pattern NUMBER_PATTERN = Pattern.compile("-?\\d+");
 
     public static final String MIME_TYPE_AVRO_BINARY = "application/avro-binary";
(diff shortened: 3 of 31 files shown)


joewitt:
Is this PR intended for Apache NiFI?  It appears to fail the build due to usage of internal/fork versions of NiFi components.

It appears to be very similar to the second of two PRs previously offered for this JIRA specifically https://github.com/apache/nifi/pull/8498

In that previous PR (8498) a concern was raised and appears to remain which relates to 'previousEventIds'.  That comment can be found here https://github.com/apache/nifi/pull/8498#pullrequestreview-1943250066.  I echo that concern.  

I left comments/questions on the previous PR as well.  I don't see responses on that PR or the JIRA or any comments in this PR to address those concerns/questions.  Possibly I'm just missing where that dialogue happened and if so please direct where it is.

mattyb149:
My mistake on the version, I pushed the fix. I believe the discussion took place on Slack but it's too old to retrieve from the Apache NiFi Slack as we don't have the pro version. Basically the idea is to provide a sort of "linked-list adjacency" that you see implemented fully as "index-free adjacency" in graph databases. Currently the lineage computation involves getting every event for a FlowFile, then sorting by timestamp, then repeating for all parent and child FlowFiles. This is too slow for determining an entire provenance graph. Consider GenerateFlowFile -> UpdateAttribute -> UpdateAttribute -> <whatever>, the UUID for the FlowFile remains the same but to tell which UpdateAttribute came first, you have to sort them by time. previousEventIds is used to enable a sort of "on-the-fly" streaming behavior to stream new provenance events into a graph database without having to re-link any of the previous nodes, just adding new nodes and links from the previous event(s) to the new one.

nifi-api/src/main/java/org/apache/nifi/provenance/ProvenanceEventRecord.java
```diff
@@ -34,6 +35,12 @@ public interface ProvenanceEventRecord {
      */
     long getEventId();
 
+    /**
+     * @return a unique ID for the "parent" Provenance Event, namely the one that came directly before this event
+     * for the given FlowFile. For source events such as CREATE, this value should be set to -1
+     */
+    Set<Long> getPreviousEventIds();
```
joewitt, line 42: Why choose '-1' as a value to indicate this is the first in a chain rather than an empty set which would also be consistent with other methods?

What scenario could result in more than one previous event identifier?   In the case this event represents a JOIN then there will have been one to many parent flow files and we already have those captured as parent flow file references. That seems correct as the mechanism then to follow the provenance event trail of those flowfiles already referenced here.  Capturing both the event ids and the parent flowfile references seems unnecessary.  If there is no previousEventIdentifier it should suggest you check the parentFlowFiles to iterate and follow their provenance path.

nifi-api/src/main/java/org/apache/nifi/provenance/ProvenanceEventRepository.java
```diff
@@ -93,5 +94,18 @@ public interface ProvenanceEventRepository {
...
+     */
+    Set<Long> getPreviousEventIds(String flowFileUUID);
 
+    /**
+     * Updates the previous provenance event IDs for the given event
+     *
+     * @param record The record for which to update the previous event IDs
+     * @param previousIds the list of previous event IDs to set for the record, or null to remove
+     */
+    void updatePreviousEventIds(ProvenanceEventRecord record, Set<Long> previousIds);
```
joewitt, line 110: why would they need to be updated rather than just set?

nifi-extension-bundles/nifi-extension-utils/nifi-database-utils/src/main/java/org/apache/nifi/util/db/JdbcCommon.java
```diff
@@ -122,6 +122,8 @@ public class JdbcCommon {
 
     public static final Pattern LONG_PATTERN = Pattern.compile("^-?\\d{1,19}$");
     public static final Pattern SQL_TYPE_ATTRIBUTE_PATTERN = Pattern.compile("sql\\.args\\.(\\d+)\\.type");
+
+    public static final Pattern JDBC_ARRAY_ELEMENT_CLASS_TYPE_PATTERN = Pattern.compile("JavaType\\(class (.*)\\) ARRAY");
```
joewitt, line 126: is this JDBC utility change necessary for the provenance change being proposed in this PR?

nifi-extension-bundles/nifi-provenance-repository-bundle/nifi-persistent-provenance-repository/pom.xml
```diff
@@ -51,5 +51,11 @@
             <artifactId>lucene-backward-codecs</artifactId>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-core-api</artifactId>
```
joewitt, line 56: What is happening that is requiring a new dependency on the nifi-framework-core-api to be introduced here?  And the reference to compile - is it intended this API is copied into the NAR?

nifi-extension-bundles/nifi-provenance-repository-bundle/nifi-persistent-provenance-repository/src/main/java/org/apache/nifi/provenance/schema/LookupTableEventRecordFields.java
```diff
@@ -33,6 +33,7 @@ public class LookupTableEventRecordFields {
 
     // General Event fields.
     public static final RecordField RECORD_IDENTIFIER_OFFSET = new SimpleRecordField(EventFieldNames.EVENT_IDENTIFIER, FieldType.INT, EXACTLY_ONE);
+    public static final RecordField PREVIOUS_EVENT_IDENTIFIERS = new SimpleRecordField(EventFieldNames.PREVIOUS_EVENT_IDENTIFIERS, FieldType.LONG, ZERO_OR_MORE);
```
joewitt, line 36: seems like it would be 'one or more' given the '-1' mechanism.  But zero or more does seem more correct.   Left previous comment on that

nifi-extension-bundles/nifi-provenance-repository-bundle/nifi-persistent-provenance-repository/src/main/java/org/apache/nifi/provenance/util/StorageSummaryEvent.java
```diff
@@ -19,11 +19,14 @@
...
 import java.util.Map;
+import java.util.Set;
+
 import org.apache.nifi.provenance.ProvenanceEventRecord;
 import org.apache.nifi.provenance.ProvenanceEventType;
+import org.apache.nifi.provenance.UpdateableProvenanceEventRecord;
 import org.apache.nifi.provenance.serialization.StorageSummary;
 
-public class StorageSummaryEvent implements ProvenanceEventRecord {
+public class StorageSummaryEvent implements UpdateableProvenanceEventRecord {
```
joewitt, line 29: This change to an 'UpdateableProvenanceEventRecord' abstraction and addition of setters stands out.  What is it about this new concept (previous event identifiers) that is causing a change in abstraction ?

joewitt:
Understood on Slack.  It is well understood about that limitation.  Please put the necessary context in the PR or JIRA so reviewers can participate and/or know their comments are being factored in.  You can copy/paste from slack.

I better understand the intent based on your recent comment.  Effectively while the key information is provided in our current provenance event/flowfile model it puts too much strain on the destination database to determine what the lineage graph looked like in terms of ordering.  This is because we provide the necessary relationships and timestamps but it would require the downstream database to reconstruct them once all events are completed as timestamp is currently a key indicator of order of operations for a given set of events associated with a single flow file.

I can see how this could be beneficial to resolve.

What other solutions were considered to address this?

The proposed change creates a link by the long identifier of all possible 'previous' events for a given provenance event and adds an alternative linkage over the current model which is purely based on the relationship of a flow file, the type of event, and any parent flowfiles from which it stems.

The core issue is ordering within the lifespan of a single flowfile.  That seems solvable with a single 'previous event identifier'.  In the case the previous event identifier is not established this is the start of this flowfile's life.  That could mean it was a created flowfile or it was the result of various different sources being combined into a new child object.  We already know and capture the parent flowfile identifier so this chain can be followed by the downstream database.

I left a couple comments as it relates to changes in the abstractions as well (Abstract repository, UpdateableEvent...). It isn't clear to me why such changes are desirable.

exceptionfactory (changes requested):
@mattyb149 Thanks for providing a bit of additional context around the difficulties related to streaming Provenance Records into a graph database as they stand. It is helpful to understand the intent in addition to the code itself.

With that being said, I am a -1 on the current approach. Making changes to the public `nifi-api` for Provenance Records, with the associated impact on implementation components, does not seem like the right solution. I have several concerns at the implementation level as well, but I'm skipping over those for now.

As I highlighted in previous pull requests, and as @joewitt noted, the parent FlowFile link already exists in the Provenance Record.

I would also be interested to know what other approaches were considered. The `ProvenanceRepository` in a framework-level extension, which is one potential alternative for supporting other solutions.

Although the use case is worth consideration, this is an important area to evaluate possible options. We should give careful thought to introducing public API changes for something that is outside of a standard framework concern.

mattyb149:
I will investigate a single previous event ID, hopefully that will be good enough for lineage purposes.

PR #9173, opened 2024-08-12
[NIFI-13641] - Remove parameter value from parameter context update request if there are referenced assets

rfellows:
_This should be merged along with #9138. When that allows adding of referenced assets to a parameter via the REST API the UI needs to be aware as to not overwrite this value if someone edits the parameter description from the UI._ 

[NIFI-13641](https://issues.apache.org/jira/browse/NIFI-13641)

If a parameter has referenced assets, the textual value (the path to the file) should be removed (if unchanged in the UI) before submitting it in an update request. This really only applies when a user might change the description of a parameter that has a referenced asset. However, if the user decides to change the value of the parameter in the UI, it should still respect that.


mcgilman:
Will review....

mcgilman (approved):
Thanks @rfellows!

PR #9182, opened 2024-08-16
NIFI-13660 Python source processors yield when they return with no Fl…

pgyori:
…owFiles

# Summary

[NIFI-13660](https://issues.apache.org/jira/browse/NIFI-13660)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-api/src/main/java/org/apache/nifi/processor/ProcessContext.java
+++ b/nifi-api/src/main/java/org/apache/nifi/processor/ProcessContext.java
@@ -74,15 +74,6 @@ public interface ProcessContext extends PropertyContext, ClusterContext {
     void yield();
 
     /**
-     * <p>
-     * An alternative name for the yield() method so that it can be called from Python.
-     * </p>
-     */
-    default void yieldResources() {
-        this.yield();
-    }
-
-    /**
      * @return the maximum number of threads that may be executing this
      * processor's code at any given time
      */
--- a/nifi-api/src/main/java/org/apache/nifi/processor/ProcessContext.java
+++ b/nifi-api/src/main/java/org/apache/nifi/processor/ProcessContext.java
@@ -74,6 +74,15 @@ public interface ProcessContext extends PropertyContext, ClusterContext {
     void yield();
 
     /**
+     * <p>
+     * An alternative name for the yield() method so that it can be called from Python.
+     * </p>
+     */
+    default void yieldResources() {
+        this.yield();
+    }
+
+    /**
      * @return the maximum number of threads that may be executing this
      * processor's code at any given time
      */
--- a/nifi-docs/src/main/asciidoc/python-developer-guide.adoc
+++ b/nifi-docs/src/main/asciidoc/python-developer-guide.adoc
@@ -327,7 +327,7 @@ In case there is no useful information to return from the `create` method, `retu
 empty `FlowFileSourceResult`. When `create()` returns with `None`, the processor does not produce any output.
 When there is nothing to return, it might be useful to yield the processor's resources and not schedule the processor
 to run for the period of time defined by the processor's Yield Duration. This can be achieved by calling
-`context.yieldResources()` from the processor's `create` method right before returning `None`.
+`context.yield_resources()` from the processor's `create` method right before returning `None`.
 
 
 [[property-descriptors]]
--- a/nifi-docs/src/main/asciidoc/python-developer-guide.adoc
+++ b/nifi-docs/src/main/asciidoc/python-developer-guide.adoc
@@ -324,7 +324,8 @@ created in the Processor's Python code. `attributes` and `contents` are both opt
 the FlowFile will still have the usual `filename`, `path` and `uuid` attributes, but no additional ones.
 If `contents` is not provided, a FlowFile with no contents (only attributes) will be created.
 In case there is no useful information to return from the `create` method, `return None` can be used instead of returning an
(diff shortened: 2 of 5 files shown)


EndzeitBegins:
Changes look straightforward and good to me. 

What do you think about a test case to verify this behaviour, to ensure it's not broken in a future refactor?

pgyori:
Thank you for your review @EndzeitBegins !
It turns out this might not be what we want after all. This yields every time None is returned and we might want to be able to yield from the Python code instead. So I'm reiterating the idea.
As for unit tests: I could not find a way to insert a ProcessContext mock to check for yield() call, apparently it isn't tested anywhere, so I figured it's ok to leave it out.

nifi-api/src/main/java/org/apache/nifi/processor/ProcessContext.java
```diff
@@ -73,6 +73,15 @@ public interface ProcessContext extends PropertyContext, ClusterContext {
      */
     void yield();
 
+    /**
+     * <p>
+     * An alternative name for the yield() method so that it can be called from Python.
+     * </p>
+     */
+    default void yieldResources() {
```
joewitt, line 81: Changing the nifi-api is not in scope for this JIRA or PR.  Other solutions should be identified.

arpadboda: To me it seems like a straight forward _extension_ of the api, which provides an alternative name for a given function without breaking anything. What better alternatives do you see?

EndzeitBegins: Introducing an alias only required by the Python side into the Java codebase seems not a good enough reason to me personally. 
The surface of such a core API should be as narrow as possible I think. 

From what I've gather from other issues, the mailing list and Slack is that the Python API is intentionally kept simplistic. I'm not certain the Python API should allow yielding explicitly at all. 

I personally thought the original commit was a better approach by making the yield implicit and thus enforcing a good practice used widely by `ListX` style Java processors.

joewitt: It does not extend the API.  It creates a method with a slightly different name then calls the other.

I am not presenting alternatives but seeking alternatives.  My role in this is as a reviewer considering whether this change is warranted and my assertion at this point is that we should seek alternatives.

arpadboda: I agree with the approach of keeping the python API simple, on the other hand, we are in the perfect time to modify/extend it based on our experience with it. It's not final in M2, but a very cool thing to have, so people started using it and naturally hit limitations. Before going GA we have the chance to adjust it to our needs, later on our hands will be much more tied because of the compatibility we have to respect. 
The "Handling of the contents of the failure result in the python FlowFileTransform API"  thread on the dev mailing list was I think one of the first occasions where the limits of the current API was hit and people don't completely agree on the general approach moving forward. From this point most of the questions become a tradeoff:
-do we want to allow the users to implement python controller services for eg?
-how much feature-parity we are going to provide with the java api when creating processors?
-how simple we would like to keep the python api?
-how "intrusive" we let python related code changes to be on the java side?

I feel like we all have different answers to these questions above. 
My personal preference is the feature parity way, but I'm pretty sure there is no consensus on that.

This is way off the scope of this PR, but I think we should find a general approach and common goals for the python API and make sure these are clear for everyone and kept in my before submitting changes, otherwise we will end up discussing things on PRs that are more like principles and not implementation details.

szaszm: Using py4j [get_method](https://www.py4j.org/py4j_java_gateway.html#py4j.java_gateway.get_method) might allow us to retrieve a method reference to yield without the keyword conflict, since it takes the method name as a string.

joewitt: If you are proposing that there are more general/fundamental things to nail down prior to the 2.0 release as it relates to the Python API please initiate a discussion thread or wiki with a proposal.

For the specific issue I flagged with this PR we need other options that do not leak such a concern into the nifi-api.

markap14: I would agree that we should avoid changing the nifi-api for this. I think the suggestion provided by @szaszm is very reasonable. The Python side already provides a wrapper around the ProcessContext proxy, so we can certainly add a `yield_resources` method to that Python-only ProcessContext, and have it use  `get_method`. 

Alternatively, we could take the approach of updating the return Object (`FlowFileSourceResult`) to provide an `isYielded()` method, and have the Python side always return an object, rather than `null`, if `yieldResources` were called.

That said, I do think that @szaszm's solution of just retrieving the method by name and invoking it directly is a better solution. Just thinking through alternatives.

pgyori:
Thank you to every one of you for your comments.
@szaszm , your get_method idea works nicely.
I pushed a commit that enables the yield functionality without modifying the NiFi API. May I ask some of you to have a look at it?

szaszm (approved):
looks good to me

arpadboda (approved):

lordgamez (approved):
LGTM

PR #9184, opened 2024-08-19
NIFI-13663: Only centering when viewing stats history when component …

mcgilman:
…is offscreen.

- Ran prettier.


scottyaslan:
Will review...

PR #9187, opened 2024-08-20
NIFI-12992: When updating Execution Engine of a Process Group, reset …

markap14:
…the validation state for all processors/controller services so that they will re-validate

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
@markap14 The change makes sense on its own, but it appears to result in timeout failures in one of the systems, across multiple attempts.

markap14:
Thanks for noting the failure @exceptionfactory . I pushed a new update. The code was causing processor validation to be reset ('VALIDATING') but the processor was already running. Since it was running, it would never re-validate. As a result, the system test was failing because it was waiting for a 'VALID' state. So I updated it such that when `resetValidationState` is called, if the component is not in a state that would allow for validation (i.e., RUNNING) it will simply log a debug message and return. When the component is stopped, the validation will automatically run again anyway.

exceptionfactory (approved):
Thanks for tracking this down @markap14! The latest tests look good, along with the change! +1 merging

PR #9196, opened 2024-08-23
NIFI-13579 Ensured whenever a timezone is specified in the timestamp pattern that it is taken into account when reading and writing.

dan-s1:
# Summary

[NIFI-13579](https://issues.apache.org/jira/browse/NIFI-13579)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for working on this issue @dan-s1.

I plan to take a closer look soon, but it would be helpful to add a new unit test class for the `ObjectZonedDateTimeConverter` covering a variety of cases. In addition, the unit test for the Timestamp Field Converter should be expanded to include timestamps with zones.

The fundamental issue is the lack of zone information in `java.sql.Timestamp`, making it inherently local to the runtime system zone. Using `ZoneId.systemDefault()` reflects that fact, so that aspect looks correct.

Something to take a closer look at is [DateTimeFormatter.getResolverFields()](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#getResolverFields--) as a way to determine whether a zone is present in the format. The regular expression pattern match would fail if the time zone is not at the end. That is the most common case, but it is a not necessarily always true, so that is another point of review in a way forward.

dan-s1:
@exceptionfactory Per the Javadocs on [DateTimeFormatter.getResolverFields()](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#getResolverFields--)

> This returns the resolver fields, used during the second phase of parsing when fields are resolved into dates and times. 
> By default, a formatter has no resolver fields, and thus returns null

Trying to use this before any parsing returns null which is where I need to use it in `ObjectZonedDateTimeConverter`

exceptionfactory:
> @exceptionfactory Per the Javadocs on [DateTimeFormatter.getResolverFields()](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#getResolverFields--)
> 
> > This returns the resolver fields, used during the second phase of parsing when fields are resolved into dates and times.
> > By default, a formatter has no resolver fields, and thus returns null
> 
> Trying to use this before any parsing returns null which is where I need to use it in `ObjectZonedDateTimeConverter`

Thanks for the reply, good point, sounds like another approach is needed.

dan-s1:
@exceptionfactory Could I just relax the regular expression to look for timezone characters any place within the timestamp pattern?

exceptionfactory:
> @exceptionfactory Could I just relax the regular expression to look for timezone characters any place within the timestamp pattern?

That's possible, although the characters could be escaped with single quotes, so the pattern would need some additional precision.

dan-s1:
@exceptionfactory The regular expression is only looking for any of those single characters and I could relax that it has to be specifically at the end.

dan-s1:
@exceptionfactory Any ideas why the unit test I added failed on the JP and FR boxes? I suspect it has to do with the Timezone although I am not sure why my resetting of the Timezone did not work.

exceptionfactory:
> @exceptionfactory Any ideas why the unit test I added failed on the JP and FR boxes? I suspect it has to do with the Timezone although I am not sure why my resetting of the Timezone did not work.

Yes, overriding the time zone in a particular unit test method is not a reliable solution for testing, and the failures point to a different approach needed. This is part of the reason for each workflow running in different time zones, as it surfaces time zone specific assumptions.

Instead of attempting to set the time zone, the expected values need to be constructed in a way that the resolved values matched the expected values.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestValidateRecord.java
```diff
@@ -593,6 +594,47 @@ public void testValidateMaps() throws IOException, InitializationException, Malf
...
+        final String content = "apache_date◆apache_ip_source◆apache_method◆apache_path◆apache_query_string◆apache_response_code◆apache_referer◆apache_user_agent\n" +
+                "Wed, 24 Jul 2024 15:04:23+0200◆10.4.3.20◆GET◆/path◆?test=toto◆200◆-◆";
+
+        runner.enqueue(content);
+        runner.run();
+
+        runner.assertTransferCount(ValidateRecord.REL_VALID, 1);
+        final MockFlowFile validFlowFile = runner.getFlowFilesForRelationship(ValidateRecord.REL_VALID).get(0);
+        //Validate timezone is taken into account
+        assertTrue(validFlowFile.getContent().contains("Wed, 24 Jul 2024 13:04:23+0000"));
```
exceptionfactory, line 634: This value should match the input value to confirm that no unexpected conversion is occurring. Recommend assigning a variable for the date stamp and using it for both input and expected values.

dan-s1: Actually this tests that the output is different than the input as the input was 
`Wed, 24 Jul 2024 15:04:23+0200◆10.4.3.20` and the assertion is the timezone was taken into account  and hence an offset in the time to `13:04:23`

exceptionfactory: Thanks for the explanation. I thought that might be the case, however, a goal of the test should be to assert the same values. As a timestamp without a zone is inherently local, the value should not be recalculated, so I recommend designing the test with that in mind.

dan-s1: I am a little confused. Shouldn't I be testing when a timezone is specified in the pattern that it is taken into account for the timestamp?

exceptionfactory: Sorry for the confusion. The reported problem is the inclusion of the time zone in the format, but that conflicts with the current conversion approach where `java.sql.Timestamp` does not understand a time zone.

I'm thinking this may need to be addressed at a more fundamental level, but I have not evaluated all the options thus far.

exceptionfactory: If you can get the tests passing on all platforms, that could be helpful as a basis for consideration. However, it may require several iterations, so it might be better to work on that in your fork with GitHub actions first, and then reconsider the pull request at that point. I would like to take a closer look at the issue soon, and I can share additional thoughts in Jira after addressing some other things.

dan-s1: @exceptionfactory You were %100 right it took a few iterations to get this right. But after this [search](https://github.com/search?q=repo%3Aapache%2Fnifi++Excel&type=pullrequests&p=2) I remembered I had locale issues when developing the Excel Record Reader in #7194. When I set the locales I realized I could not use this `"Wed, 24 Jul 2024 15:04:23+0200"` so I ended up removing the three character day of the week, the three day month and the `+0200` timezone.  I also tweeked the timezone regular expression to cover all cases.
Please let me know what else needs to be done. Thanks!

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestValidateRecord.java
```diff
@@ -593,6 +599,47 @@ public void testValidateMaps() throws IOException, InitializationException, Malf
...
+        final String content = "apache_date◆apache_ip_source◆apache_method◆apache_path◆apache_query_string◆apache_response_code◆apache_referer◆apache_user_agent\n" +
+                timestampWithTimezone + "◆10.4.3.20◆GET◆/path◆?test=toto◆200◆-◆";
+
+        runner.enqueue(content);
+        runner.run();
+
+        runner.assertTransferCount(ValidateRecord.REL_VALID, 1);
+        final MockFlowFile validFlowFile = runner.getFlowFilesForRelationship(ValidateRecord.REL_VALID).getFirst();
+        //Validate timezone is taken into account
+        assertFalse(validFlowFile.getContent().contains(timestampWithTimezone));
```
exceptionfactory, line 639: The comment does not seem to align with the assertion. This seems to indicate that the timestamp is _not_ expected to be found?

exceptionfactory (changes requested):
Thanks for the updates @dan-s1. Although I think a more fundamental solution would be to move away from `java.sql.Timestamp` in records, that would be more involved, so this approach may be sufficient.

It is good to see the passing unit tests, but overriding the default `Locale` should not be necessary, since that occurs at the system level for the different GitHub workflows. If you remove the local settings, does it work? I also raised a question about the assertion for the timestamp itself.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestValidateRecord.java
dan-s1: I see how its not clear but ultimately I want to assert that the `15` in `15:04:23` is no longer there which should demonstrate the timezone is taken into account.

exceptionfactory: In other words, the test is expecting the timestamp to change?

dan-s1: Yes

exceptionfactory: Thanks for confirming. That highlights the core concern. The current behavior of the `ObjectTimestampFieldConverter` expects the input and output values to be the same, which aligns with the understanding the `java.sql.Timestamp` does not have a zone offset. When converting from a "zoned" timestamp as in this test, it raises the question whether the output should reflect that offset when computing output timestamp.

Perhaps one way to make the test both more specific, and workable across zones, is to build input string based on the system default zone offset. That way, the input and output hours should remain unchanged.

dan-s1: I guess I do not fully understand why  the `ObjectTimestampFieldConverter` expects the input and output values to be the same if the format pattern has timezone information in it.

dan-s1: BTW I do not have it working as I thought I did. Actually the time remains the same. This though does not address the first concern mentioned in the ticket 

> When reading, the Z is accepted in the format but doesn't seem to be taken into account. Indeed, if we have a timestamp with +0200 (plus 2 hours), on output we have the timestamp on the NiFi timezone but without the time offset. For example, with 15h in +0200 I'd have 15h in +0000 instead of 13h in +0000.

dan-s1: Please ignore previous comment as I have it working now to be able to change the hour.

dan-s1: @exceptionfactory On a separate note, I noticed that `ObjectLocalDateTimeFieldConverter` has special logic for converting when the field is a `Number`.  There are many overloaded `toLocalDateTime` methods to handle the different scenarios. I do not see similar logic in any of the converters for `Number`. Can you please give me insight as to what is the reason this logic is only located in `ObjectLocalDateTimeFieldConverter`?

exceptionfactory: @dan-s1 Yes, that is an important point, as noted in [NIFI-12710](https://issues.apache.org/jira/browse/NIFI-12710) handling `Double` and `Float` support fractional seconds.

dan-s1: Thanks for your prompt reply. Why isn't that handled in any of the converters?

exceptionfactory: Since the record `TIMESTAMP` type maps to a `java.sql.Timestamp` the change was focused to particular converter. It may be worth implementing in other converters, but not necessarily. It primarily depends on what provides the conversion for the Record `TIMESTAMP` Type.

exceptionfactory: Refactoring it out may not be necessary at the moment, as other date and time types do not have fractional seconds.

dan-s1: I started refactoring it to use it in  `ObjectZonedDateTimeConverter`.  I am trying to clarify if you still have an issue with `ObjectZonedDateTimeConverter` would keep the same behavior as `ObjectLocalDateTimeFieldConverter` if there is no timezone information specified in the format?

exceptionfactory: Thanks for the update. Given that the real focus of the issue is `java.sql.Timestamp`, I recommend building the behavior into `ObjectTimestampFieldConverter` itself. There may be other uses for a `ZonedDateTime` converter in the future, but as we are considering the specialized handling of time zones for `java.sql.Timestamp`, it would be best to implement this in the Timestamp Converter itself.

dan-s1: I was able to take most of the tests from `TestObjectLocalDateTimeFieldConverter` and ran them in `ObjectZonedDateTimeConverterTest`.  I replaced `LocalDateTime` with `ZonedDateTime`. The only test I could not get working is testDoubleAsEpochSecondsAsString. To be honest I am not sure what exactly the issue is.

dan-s1: @exceptionfactory I see you have been very busy the last couple of days. If you have time can you please review my changes and see if this is what you had in mind? Thanks!

exceptionfactory: Thanks @dan-s1, I plan to get back to this soon.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverter.java
```diff
@@ -39,7 +57,80 @@ class ObjectTimestampFieldConverter implements FieldConverter<Object, Timestamp>
      */
     @Override
     public Timestamp convertField(final Object field, final Optional<String> pattern, final String name) {
```
exceptionfactory, line 59: The multiple `return` statements in this method make it much more difficult to follow. Refactoring to have a single return would be helpful.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/FractionalSecondsUtils.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.time.Instant;
+
+public class FractionalSecondsUtils {
+    private static final long YEAR_TEN_THOUSAND = 253_402_300_800_000L;
+
+    private FractionalSecondsUtils() {
+    }
+
+    public static Instant tryParseAsNumber(final String value) {
```
exceptionfactory, line 27: Recommend renaming this method to `toInstant()` as well, following the pattern of the other methods.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/FractionalSecondsUtils.java
```diff
@@ -0,0 +1,57 @@
...
+        // If decimal, treat as a double and convert to seconds and nanoseconds.
+        if (value.contains(".")) {
+            final double number = Double.parseDouble(value);
+            return toInstant(number);
+        }
+
+        // attempt to parse as a long value
+        final long number = Long.parseLong(value);
```
exceptionfactory, line 35: Promoting this parsing to a public method moves the exception handling requirements to the caller, which could be unexpected for referencing classes. I recommend catching the NumberFormatException in this class and throwing the FieldConversionException.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverterTest.java
```diff
@@ -44,70 +50,108 @@ public class ObjectTimestampFieldConverterTest {
...
+        final String actualString = actual.toString();
+
+        //Validate timezone is taken into account as the hour in the timestamp should be different per timezone offset
+        final String partialTimestampWithOriginalHour = " %s:".formatted(originalTimestampHour);
+        assertFalse(actualString.contains(partialTimestampWithOriginalHour));
+    }
+
+    @ParameterizedTest
+    @MethodSource("getPatterns")
+    void testTimeZonePattern(String pattern, boolean expected) {
```
exceptionfactory, line 136:
```suggestion
    void testTimeZonePattern(String pattern, boolean matchExpected) {
```

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverterTest.java
```diff
@@ -44,70 +50,108 @@ public class ObjectTimestampFieldConverterTest {
...
+        final String originalTimestampHour = "12";
+        //NOTE: Antarctica/Casey is the timezone offset chosen in timestamp below
+        final String originalTimestamp = "2000-01-01 " + originalTimestampHour + ":00:00+0800";
+        final Optional<String> timezonePattern = Optional.of("yyyy-MM-dd HH:mm:ssZ");
+        final Timestamp actual = CONVERTER.convertField(originalTimestamp, timezonePattern, FIELD_NAME);
+        final String actualString = actual.toString();
+
+        //Validate timezone is taken into account as the hour in the timestamp should be different per timezone offset
+        final String partialTimestampWithOriginalHour = " %s:".formatted(originalTimestampHour);
+        assertFalse(actualString.contains(partialTimestampWithOriginalHour));
```
exceptionfactory, line 131: Asserting false does not provide a high level of validation for this behavior. Instead, it would be better to declare the expected value and assert equals.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverter.java
```diff
@@ -39,7 +57,80 @@ class ObjectTimestampFieldConverter implements FieldConverter<Object, Timestamp>
...
+                        if (matcher.find()) {
+                            zonedDateTime = ZonedDateTime.parse(stringTrimmed, formatter);
+                        } else {
+                            final LocalDateTime localDateTime = LocalDateTime.parse(stringTrimmed, formatter);
+                            zonedDateTime = ZonedDateTime.of(localDateTime, ZoneId.systemDefault());
+                        }
+                        instant = zonedDateTime.toInstant();
+                    } catch (final DateTimeParseException e) {
+                        return tryParseAsNumber(stringTrimmed, name);
+                    }
```
exceptionfactory, line 113: This logic is complex enough it looks like it warrants a separate method.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverter.java
```diff
@@ -39,7 +57,80 @@ class ObjectTimestampFieldConverter implements FieldConverter<Object, Timestamp>
...
+                        // no timezone information. Hence, a regular expression is used to determine whether it is necessary
+                        // to parse with ZoneDateTime or not.
+                        final Matcher matcher = TIMEZONE_PATTERN.matcher(patternString);
+                        final ZonedDateTime zonedDateTime;
+
+                        if (matcher.find()) {
+                            zonedDateTime = ZonedDateTime.parse(stringTrimmed, formatter);
+                        } else {
+                            final LocalDateTime localDateTime = LocalDateTime.parse(stringTrimmed, formatter);
+                            zonedDateTime = ZonedDateTime.of(localDateTime, ZoneId.systemDefault());
```
exceptionfactory, line 108: This intermediate conversion to `ZonedDateTime` and then `Instant` is not necessary when the destination type is `Timestamp`. It seems like the general approach should be refactored to set a Timestamp value instead of using `Instant` as the intermediate type.

exceptionfactory (changes requested):
Thanks for your continued work here @dan-s1. I noted a few more recommendations.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/FractionalSecondsUtils.java
dan-s1: Currently `FieldConversionException` is protected, should I make both constructors public?

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverter.java
dan-s1: I followed the pattern I saw in `ObjectLocalDateTimeFieldConverter` which has multiple returns.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverterTest.java
dan-s1: The expected value will vary from timezone to timezone and in the case of the build there are three different time zones making it hard to provide a fixed value. Hence I chose to use `assertFalse` knowing that it cannot be the hour originally used.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/FractionalSecondsUtils.java
exceptionfactory: I would just move the utility class into the same `field` package in this case.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverter.java
exceptionfactory: Good point, scratch that recommendation.

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverter.java
dan-s1: I thought I needed this step in order to keep track of the right timezone. I added similar logic in `ObjectStringFieldConverter` although I did not use an `Instant` as there is only formatting to a string there.

dan-s1: @exceptionfactory So I refactored all that code now into method `toInstant` on lines 113-130. Should I change the signature `toTimestamp` and return a `Timestamp`?

exceptionfactory: @dan-s1 Thanks for the update. I'm taking a look at an alternative approach to this fundamental time zone problem, and I hope to follow up soon with more details.

dan-s1: @exceptionfactory I am just curios if the alternative approach you had was successful or not. Thanks!

exceptionfactory: Thanks for your patience @dan-s1. I made some progress, but had several other things to address this week. I hope to have something more substantive before too long.

exceptionfactory:
Thanks again for your work and patience on this issue @dan-s1.

After taking a closer look at the issue and implementation considerations, I submitted PR #9367. The timestamp to string conversion follows the same approach you had implemented, but the string to timestamp parsing avoids the regular expression approach in favor of using the `DateTimeFormatter.parseBest()` method with `ZonedDateTime` and `LocalDateTime`. The PR also includes tests to assert expected values, but include zone offset evaluation based on the JVM.

dan-s1:
My pleasure. Should I close this PR?

exceptionfactory:
> My pleasure. Should I close this PR?

Thanks, yes, closing this PR sounds good.

PR #9214, opened 2024-08-29
[NIFI-13624] - Allow View Status History of the current process group.

rfellows:
[NIFI-13624](https://issues.apache.org/jira/browse/NIFI-13624)

Added support in the context menu to view status history of the current process group.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
@@ -1132,13 +1132,7 @@ export class FlowEffects {
                 ofType(FlowActions.navigateToViewStatusHistoryForCurrentProcessGroup),
                 concatLatestFrom(() => this.store.select(selectCurrentProcessGroupId)),
                 tap(([, processGroupId]) => {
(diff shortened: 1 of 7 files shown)


mcgilman:
Will review...

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
```diff
@@ -1126,6 +1126,24 @@ export class FlowEffects {
...
+                ofType(FlowActions.navigateToViewStatusHistoryForCurrentProcessGroup),
+                concatLatestFrom(() => this.store.select(selectCurrentProcessGroupId)),
+                tap(([, processGroupId]) => {
+                    this.router.navigate([
+                        '/process-groups',
+                        processGroupId,
+                        ComponentType.ProcessGroup,
+                        processGroupId,
+                        'history'
+                    ]);
```
mcgilman, line 1141: The assumption for this route is that the second id is a child component of the Process Group specified by first id. I think adding a route for `/process-groups/{id}/history` would be more correct for opening the status history for the current Process Group.

mcgilman (approved):
Thanks for the updates @rfellows! Looks great... will merge once CI completes.

PR #9218, opened 2024-08-29
[NIFI-13683] - Refactor Status History dialog to not support manual r…

rfellows:
…esizing via a dialog drag handle, but favor a maximize/minimize button approach familiar to other windowed applications. When maximized, the dialog will resize as the browser window resizes. This solution eliminates the need for the unsupported (in Firefox) computedStyleMap that was causing issues.

[NIFI-13683](https://issues.apache.org/jira/browse/NIFI-13683)


scottyaslan:
Reviewing...

scottyaslan (approved):
Works well! +1

PR #9221, opened 2024-08-29
NIFI-13632: Updating content access.

mcgilman:
- Removing existing content viewer application.
- Introduced new front end based content viewer applicaiton.
- Introduced new bundled content viewers for hex and images.
- Deleted previous image content viewer.
- Migrated existing standard content viewer (which handles json, xml, yaml, csv, and text) to utilize new content access interface.
- Moved standard content viewer into its own NAR.
- Moved and renamed custom ui utils and content viewer utils into nifi-common.
- Added mime type to FlowFileSummary response payload to help drive the initially opened content viewer.


rfellows:
Reviewing from a front-end perspective...

exceptionfactory:
Thanks for the substantial work @mcgilman! I will review the server-side implementation soon.

I did notice a build failure related to a missing reference in `nifi-code-coverage`, so that should be removed for successful CI builds.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/content-viewer/feature/content-viewer.component.html
```diff
@@ -0,0 +1,49 @@
...
+
+<div class="content-viewer-background h-screen flex flex-col gap-y-4 p-4">
+    <div class="flex justify-between items-center">
+        <div>
+            <img ngSrc="assets/icons/nifi-logo-about.svg" priority width="150" height="64" alt="NiFi Logo" />
+        </div>
+        <form [formGroup]="viewerForm">
+            <mat-form-field subscriptSizing="dynamic" class="w-72">
+                <mat-label>View</mat-label>
+                <mat-select formControlName="viewAs" (selectionChange)="viewAsChanged($event)">
```
rfellows, line 26: Consider conditionally setting the panelWidth of this select between `auto` and `null` depending on if there are any types of viewers with multiple options.
> If set to auto, the panel will match the trigger width. If set to null or an empty string, the panel will grow to match the longest option's text.

If the case where there are multiple content viewers for the same display name, the displayed value is really long and wraps. It makes it challenging to read.

Screenshot 2024-08-30 at 1 34 41 PM

rfellows (changes requested):
This is really good work, @mcgilman. I did leave some comments inline. Feel free to reach out if you want to discuss them further.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/content-viewer/feature/content-viewer.component.ts
```diff
@@ -0,0 +1,273 @@
...
+})
+export class ContentViewerComponent implements OnInit, OnDestroy {
+    viewerForm: FormGroup;
+    viewAsOptions: SelectGroup[] = [];
+
+    private supportedMimeTypeId = 0;
+    private supportedMimeTypeLookup: Map<number, SupportedMimeTypes> = new Map<number, SupportedMimeTypes>();
+    private supportedMimeTypeContentViewerLookup: Map<number, ContentViewer> = new Map<number, ContentViewer>();
+    private mimeTypeDisplayNameLookup: Map<number, string> = new Map<number, string>();
+    private mimeTypeIdsSupportedByBundledUis: Set<number> = new Set<number>();
```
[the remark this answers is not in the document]
mcgilman: Sure makes sense but given how we check if a given viewer is embedded/bundled that likely makes sense to keep as a separate collection.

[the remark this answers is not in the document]
rfellows: If you put that flag in the object as I suggested, you can use that when we need to know if it is embedded or bundled.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/content-viewer/feature/content-viewer.component.html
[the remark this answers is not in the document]
mcgilman: This implementation replaces a JSP/Servlet based approach where the markup is generated server side. In the previous implementation we had access to these details when building the response. The new client side implementation (which is invoked from a queue listing or a provenance event) does not have these details and these is no endpoint available to retrieve details about `content`. 

We do know the filename from the flowfile in the queue or from the provenance event. We could get the filename from those source resources and include in a query parameter just for the new content viewer to display. Having a query parameter whose sole purpose was to render a string on the page didn't seem like a great option.

nifi-frontend/src/main/frontend/apps/standard-content-viewer/src/app/pages/standard-content-viewer/feature/standard-content-viewer.component.ts
```diff
@@ -0,0 +1,135 @@
...
+
+                    this.loadContent();
+                }
+            });
+    }
+
+    loadContent(): void {
+        if (this.ref && this.mimeTypeDisplayName) {
+            this.setMode(this.mimeTypeDisplayName);
+
```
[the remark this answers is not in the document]
mcgilman: I wasn't sure if we wanted the skeleton loader to flash when switching between Original/Formatted when more often than not it will not take a lengthy time for the request to complete. But I can make this update as it's more correct.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/content-viewer/feature/content-viewer.component.html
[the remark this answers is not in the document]
rfellows: The trouble is when there isn't a content viewer available for your flowfile. You might not know what the content type is to make an informed guess as to which of the viewers _might_ work or at least look the best. The filename (extension) also could give hints here.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/content-viewer/feature/content-viewer.component.html
```diff
@@ -0,0 +1,49 @@
...
+                    }
+                </mat-select>
+            </mat-form-field>
+        </form>
+    </div>
+    <div class="p-2 flex-1 flex border overflow-y-auto">
+        @if (viewerSelected) {
+            <router-outlet></router-outlet>
+        } @else {
+            <div class="unset">No data reference specified</div>
```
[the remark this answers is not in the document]
mcgilman: I think we should be falling back to the `hex` viewer when encountering an unsupported mime type. If the data ref has been specified we should do our best to load it with the correct viewer, but if we cannot we should be falling back to `hex`. Let me track down why that resulted in the `No data reference specified` message.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/pom.xml
```diff
@@ -18,9 +26,8 @@
     <artifactId>nifi-standard-content-viewer</artifactId>
     <packaging>war</packaging>
     <properties>
-        <frontend.dependency.configs>${basedir}/src/main/frontend</frontend.dependency.configs>
-        <frontend.working.dir>${project.build.directory}/frontend-working-directory</frontend.working.dir>
-        <frontend.assets>${project.build.directory}/${project.build.finalName}/assets</frontend.assets>
+        <spring.boot.version>3.3.2</spring.boot.version>
```
exceptionfactory, line 29: The latest version of Spring Boot is now 3.3.3 as used in NiFi Registry.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/src/main/resources/application.yml
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+server:
+  servlet:
+    application-display-name: Standard Content Viewer
```
exceptionfactory, line 19: Is this configuration file used?

exceptionfactory (commented):
This looks solid from a server implementation perspective, just noted two questions regarding the Spring Boot version and the configuration file.

joewitt:
Testing this out a bit.  Making great progress.

I tried a scenario with GenerateFlowFile to a python script.  The python script writes out flowfile content that is just "1.25.0".  The flowfile has only filename, path, and uuid attributes set.  I then route to updateAttribute that is stopped.  I then list queue and view content for one of these.  It shows up by default in a hex editor.  That looks good.  Then when I select view-text it shows a blank/white screen.  If I go back to hex editor it looks good.  Then back to text and it looks good.

So something funky there but perhaps related to flowfiles with minimal attributes.

Also notable in that same panel if I click around to csv it will sometimes give the white screen and sometimes render.

nifi-frontend/src/main/frontend/apps/standard-content-viewer/src/app/pages/standard-content-viewer/feature/standard-content-viewer.component.scss
```diff
@@ -0,0 +1,26 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+.standard-content-viewer {
+    h3 {
+        margin-bottom: 0;
```
scottyaslan, line 20: This seems a little unnecessary. We rely on the angular typography to style this header and we should probably not be changing that. What is the issue here? I see you also added a `mr-2` under the 'Original|Formatted' buttons. I think if you remove that margin and remove this margin on the `h3` then the layout should be correct.

nifi-frontend/src/main/frontend/apps/standard-content-viewer/src/app/pages/standard-content-viewer/feature/standard-content-viewer.component.scss
```diff
@@ -0,0 +1,26 @@
...
+
+        .mime-type-display-name {
+            text-transform: uppercase;
```
scottyaslan, line 23: This seems like we should be using the tailwind 'uppercase' class. Is there a reason this h3 needs to be all upper case? It seems all of our other page header `h3`'s are not.

scottyaslan: Disregard this question. I see now that you are doing this for text like CSV or other mime type. I guess you could use angular in the template to conditionally apply the tailwind class for these strings.

nifi-frontend/src/main/frontend/apps/standard-content-viewer/src/app/pages/standard-content-viewer/feature/standard-content-viewer.component.html
```diff
@@ -0,0 +1,44 @@
...
+  ~  See the License for the specific language governing permissions and
+  ~  limitations under the License.
+  -->
+
+<div class="standard-content-viewer p-5 flex h-screen">
+    <form [formGroup]="contentFormGroup" class="w-full flex-1 flex flex-col gap-y-2">
+        <div class="flex justify-between items-center">
+            <h3 class="primary-color">
+                @if (mimeTypeDisplayName) {
+                    <span class="mime-type-display-name">{{ mimeTypeDisplayName }}</span
```
scottyaslan, line 23:
```suggestion
                    <span class="uppercase">{{ mimeTypeDisplayName }}</span
```

nifi-frontend/src/main/frontend/apps/standard-content-viewer/src/app/pages/standard-content-viewer/feature/standard-content-viewer.component.scss
mcgilman: Where did you see an `mr-2` in the standard content viewer UI?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/src/main/resources/application.yml
mcgilman: Good call. As-is this can be removed.

nifi-frontend/src/main/frontend/apps/standard-content-viewer/src/app/pages/standard-content-viewer/feature/standard-content-viewer.component.scss
mcgilman: I've applied your suggestion below.

joewitt:
additional testing after a hard refresh the white screen i described still happens but it is very rare. in a local docker deployment it seems a bit easier to reproduce.  in local mode without docker it is very hard to reproduce.  it happened once in maybe 50 attempts.  

So in short - i dont think it is worth bothering with at this point.  might be a local issue/hard refresh certainly seems to have made the issue less likely.

thanks

nifi-frontend/src/main/frontend/apps/standard-content-viewer/src/app/pages/standard-content-viewer/feature/standard-content-viewer.component.html
```diff
@@ -33,7 +33,7 @@ <h3 class="primary-color">
             @if (error) {
                 <div>{{ error }}</div>
             } @else {
-                <div class="flex-1 overflow-auto">
+                <div class="flex-1 overflow-auto min-h-[125px]">
```
scottyaslan, line 36: The codemirror base style sheet already sets a height:

```
.CodeMirror {
  /* Set height, width, borders, and global font properties here */
  font-family: monospace;
  height: 300px;
  color: black;
  direction: ltr;
}
```
and the nifi codemirror theme already adds a min-height: 108px;

https://github.com/apache/nifi/blob/main/nifi-frontend/src/main/frontend/libs/shared/src/assets/styles/_codemirror-theme.scss#L65

rfellows (approved):
This is looking great from a UI-perspective. Thanks for the hard work @mcgilman.

scottyaslan (approved):
The latest changes work well. Thanks @mcgilman!

exceptionfactory (approved):
Thanks again for the work on this @mcgilman, looks good!

PR #9222, opened 2024-08-30
NIFI-13695: Appending a trailing slash when opening a custom ui and i…

mcgilman:
…n the query string to fragment filter to avoid a jetty based redirect that could lead to losing the proxy path.


exceptionfactory (approved):
Thanks @mcgilman, verified this works as designed for both direct and proxied access. +1 merging

PR #9232, opened 2024-09-05
NIFI-13620: Resolved MaxWaitTime Issue in QueryCassandra

Rahul937303:
- Resolved the problem where QueryCassandra was not respecting the configured Max Wait Time.
- Updated the logic to ensure that the Max Wait Time is correctly applied to all query executions.


exceptionfactory (approved):
Thanks for the correction and updating the corresponding test class @Rahul937303! +1 merging

PR #9236, opened 2024-09-06
NIFI-13720: Component is not reloaded when the isolation key depends on service property

mark-bathori:
# Summary
To be able to create a system test for the PR the system-test-extension had to be restructured because when a component is annotated with  **@RequiresInstanceClassLoading** and contains a controller service property then the framework does not allow the component to be in the same bundle as the service API.

[NIFI-13720](https://issues.apache.org/jira/browse/NIFI-13720)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/controller/AbstractComponentNode.java
```diff
@@ -784,13 +784,14 @@ public synchronized void reloadAdditionalResourcesIfNecessary() {
...
-        if (dynamicallyModifiesClasspath) {
+        if (dynamicallyModifiesClasspath || isolationKey != null) {
             final Set<URL> additionalUrls = this.getAdditionalClasspathResources(descriptors, this::getEffectivePropertyValueWithDefault);
 
-            final String newFingerprint = ClassLoaderUtils.generateAdditionalUrlsFingerprint(additionalUrls, determineClasloaderIsolationKey());
-            if (!StringUtils.equals(additionalResourcesFingerprint, newFingerprint)) {
-                setAdditionalResourcesFingerprint(newFingerprint);
+            final String newFingerprint = ClassLoaderUtils.generateAdditionalUrlsFingerprint(additionalUrls, isolationKey);
+            if (!StringUtils.equals(componentFingerprint, newFingerprint)) {
+                setComponentFingerprint(newFingerprint);
```
turcsanyip, line 794: I suggest not renaming the variable/method from `additionalResourcesFingerprint` to `componentFingerprint` but leaving it as it is now.
To be fully consistent, the interface methods (`isReloadAdditionalResourcesNecessary()`, `reloadAdditionalResourcesIfNecessary()`) should be renamed too. However, I don't think it is worth changing these API methods.
Also, the isolation key was added in the "additional resources fingerprint" earlier so it was already a bit inconsistent name. It will be no better or worse due to the change in this PR.

turcsanyip (approved):
Thanks for the fix @mark-bathori!

+1 LGTM

@markap14 Do you have any other idea to overcome the issue mentioned in the PR summary? (other than restructuring the system tests module)

> To be able to create a system test for the PR the system-test-extension had to be restructured because when a component is annotated with @RequiresInstanceClassLoading and contains a controller service property then the framework does not allow the component to be in the same bundle as the service API.

turcsanyip:
Merging to main.

@markap14 Feel free to add your notes on this system tests module change later on and we can adjust it in a follow-up jira. Thanks

PR #9255, opened 2024-09-10
[NIFI-13735] update codemirror to use height 100% and fix jolt advanc…

scottyaslan:
…ed ui scrollbar issue


mcgilman:
Will review...

mcgilman (approved):
Thanks @scottyaslan! The changes here look good. When reviewing a discovered a separate issue that current exists on `main` when attempting to edit FlowFile Input in the Jolt Advanced UI. I'll file a separate JIRA to address that issue.

PR #9262, opened 2024-09-12
NIFI-13743: Ensuring the mime type is set when navigating to the cont…

mcgilman:
…ent viewer from Provenance.

- Adding a trailing slash when loading the external viewer to eliminate an unneeded redirection.


rfellows:
reviewing...

rfellows (approved):
Thanks for these updates @mcgilman. This makes viewing content much nicer from the provenance view.

Will merge once CI completes

PR #9264, opened 2024-09-13
NIFI-13644 Add GitLab Flow Registry Client

bbende:
- Refactoring to create base class for other git-based flow registry clients
- Implement GitLabRepositoryClient
- Implement GitLabFlowRegistryClient

# Summary

[NIFI-13644](https://issues.apache.org/jira/browse/NIFI-13644)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
@@ -237,11 +237,11 @@ public abstract class AbstractGitFlowRegistryClient extends AbstractFlowRegistry
         final String branch = flowLocation.getBranch();
         final String filePath = getSnapshotFilePath(flowLocation);
         final String commitMessage = DEREGISTER_FLOW_MESSAGE_FORMAT.formatted(flowLocation.getFlowId());
-        final InputStream deletedSnapshotContent = repositoryClient.deleteContent(filePath, commitMessage, branch);
-
-        final RegisteredFlowSnapshot deletedSnapshot = getSnapshot(deletedSnapshotContent);
-        updateBucketReferences(repositoryClient, deletedSnapshot, flowLocation.getBucketId());
-        return deletedSnapshot.getFlow();
+        try (final InputStream deletedSnapshotContent = repositoryClient.deleteContent(filePath, commitMessage, branch)) {
+            final RegisteredFlowSnapshot deletedSnapshot = getSnapshot(deletedSnapshotContent);
+            updateBucketReferences(repositoryClient, deletedSnapshot, flowLocation.getBucketId());
+            return deletedSnapshot.getFlow();
+        }
     }
 
     @Override
@@ -415,9 +415,9 @@ public abstract class AbstractGitFlowRegistryClient extends AbstractFlowRegistry
 
     private FlowRegistryBucket createFlowRegistryBucket(final GitRepositoryClient repositoryClient, final String name) {
         final FlowRegistryPermissions bucketPermissions = new FlowRegistryPermissions();
-        bucketPermissions.setCanRead(repositoryClient.getCanRead());
-        bucketPermissions.setCanWrite(repositoryClient.getCanWrite());
-        bucketPermissions.setCanDelete(repositoryClient.getCanWrite());
+        bucketPermissions.setCanRead(repositoryClient.hasReadPermission());
+        bucketPermissions.setCanWrite(repositoryClient.hasWritePermission());
+        bucketPermissions.setCanDelete(repositoryClient.hasWritePermission());
 
         final FlowRegistryBucket bucket = new FlowRegistryBucket();
         bucket.setIdentifier(name);
@@ -551,13 +551,13 @@ public abstract class AbstractGitFlowRegistryClient extends AbstractFlowRegistry
     }
 
     private void verifyWritePermissions(final GitRepositoryClient repositoryClient) throws AuthorizationException {
-        if (!repositoryClient.getCanWrite()) {
+        if (!repositoryClient.hasWritePermission()) {
             throw new AuthorizationException("Client does not have write access to the repository");
         }
     }
 
     private void verifyReadPermissions(final GitRepositoryClient repositoryClient) throws AuthorizationException {
-        if (!repositoryClient.getCanRead()) {
+        if (!repositoryClient.hasReadPermission()) {
             throw new AuthorizationException("Client does not have read access to the repository");
         }
     }
@@ -587,15 +587,15 @@ public abstract class AbstractGitFlowRegistryClient extends AbstractFlowRegistry
     // If the client has write permissions to the repo, then ensure the directory for the default bucket is present and if not create it,
     // otherwise the client can only be used to import flows from the repo and won't be able to set up the default bucket
     private void initializeDefaultBucket(final FlowRegistryClientConfigurationContext context) throws IOException, FlowRegistryException {
-        if (!repositoryClient.getCanWrite()) {
-            getLogger().info("Repository client does not have write permissions to the repository, skipping setup of default bucket");
+        if (!repositoryClient.hasWritePermission()) {
+            getLogger().info("Repository client [{}] does not have write permissions to the repository, skipping setup of default bucket", getIdentifier());
             return;
         }
 
         final String branch = context.getProperty(REPOSITORY_BRANCH).getValue();
         final Set<String> bucketDirectoryNames = repositoryClient.getTopLevelDirectoryNames(branch);
         if (!bucketDirectoryNames.isEmpty()) {
-            getLogger().info("Found existing buckets, skipping setup of default bucket");
+            getLogger().debug("Found {} existing buckets, skipping setup of default bucket", bucketDirectoryNames.size());
             return;
         }
 
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
@@ -77,7 +77,7 @@ public abstract class AbstractGitFlowRegistryClient extends AbstractFlowRegistry
 
     public static final PropertyDescriptor REPOSITORY_PATH = new PropertyDescriptor.Builder()
             .name("Repository Path")
-            .description("The path with in the repository that this client will use to store all data. " +
+            .description("The path in the repository that this client will use to store all data. " +
                     "If left blank, then the root of the repository will be used.")
             .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
             .required(false)
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
@@ -401,7 +401,7 @@ public abstract class AbstractGitFlowRegistryClient extends AbstractFlowRegistry
         final String filePath = getSnapshotFilePath(flowLocation);
 
         final List<GitCommit> commits = repositoryClient.getCommits(filePath, branch);
-        final String latestVersion = commits.isEmpty() ? null : commits.getFirst().getId();
+        final String latestVersion = commits.isEmpty() ? null : commits.getFirst().id();
         return Optional.ofNullable(latestVersion);
     }
 
@@ -431,10 +431,10 @@ public abstract class AbstractGitFlowRegistryClient extends AbstractFlowRegistry
         snapshotMetadata.setBranch(flowLocation.getBranch());
         snapshotMetadata.setBucketIdentifier(flowLocation.getBucketId());
         snapshotMetadata.setFlowIdentifier(flowLocation.getFlowId());
-        snapshotMetadata.setVersion(commit.getId());
-        snapshotMetadata.setAuthor(commit.getAuthor());
-        snapshotMetadata.setComments(commit.getMessage());
-        snapshotMetadata.setTimestamp(commit.getCommitDate().getTime());
+        snapshotMetadata.setVersion(commit.id());
+        snapshotMetadata.setAuthor(commit.author());
+        snapshotMetadata.setComments(commit.message());
+        snapshotMetadata.setTimestamp(commit.commitDate().toEpochMilli());
         return snapshotMetadata;
     }
 
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
@@ -1,0 +1,644 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.nifi.registry.flow.git;
+
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.ValidationContext;
+import org.apache.nifi.components.ValidationResult;
+import org.apache.nifi.flow.ConnectableComponent;
+import org.apache.nifi.flow.Position;
+import org.apache.nifi.flow.VersionedComponent;
+import org.apache.nifi.flow.VersionedConnection;
+import org.apache.nifi.flow.VersionedFlowCoordinates;
+import org.apache.nifi.flow.VersionedProcessGroup;
+import org.apache.nifi.processor.util.StandardValidators;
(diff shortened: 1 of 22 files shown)


nifi-extension-bundles/nifi-github-bundle/nifi-github-extensions/src/main/java/org/apache/nifi/github/GitHubRepositoryClient.java
```diff
@@ -255,20 +258,27 @@ public InputStream getContentFromCommit(final String path, final String commitSh
...
         return execute(() -> {
             try {
                 final GHRef branchGhRef = repository.getRef(branchRef);
-                return repository.queryCommits()
+                final List<GHCommit> ghCommits = repository.queryCommits()
                         .path(resolvedPath)
                         .from(branchGhRef.getObject().getSha())
                         .pageSize(COMMIT_PAGE_SIZE)
                         .list()
                         .toList();
```
EndzeitBegins, line 275: As we're iterating / mapping over the result immediately after, we don't need to collect it into a `List` at this point.

```suggestion
                final PagedIterable<GHCommit> ghCommits = repository.queryCommits()
                        .path(resolvedPath)
                        .from(branchGhRef.getObject().getSha())
                        .pageSize(COMMIT_PAGE_SIZE)
                        .list();
```

nifi-extension-bundles/nifi-gitlab-bundle/nifi-gitlab-extensions/src/main/java/org/apache/nifi/gitlab/GitLabAuthenticationType.java
```diff
@@ -0,0 +1,22 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.nifi.gitlab;
+
+public enum GitLabAuthenticationType {
+    ACCESS_TOKEN
+}
```
EndzeitBegins, line 22: Does it make sense to turn this into a `DescribedValue` similar to the existing `GitHubAuthenticationType`? 

```suggestion
public enum GitLabAuthenticationType implements DescribedValue {
    ACCESS_TOKEN("Access Token", "Group, Project, or Personal Access Token");

    private final String displayName;
    private final String description;

    GitLabAuthenticationType(final String displayName, final String description) {
        this.displayName = displayName;
        this.description = description;
    }

    @Override
    public String getValue() {
        return name();
    }

    @Override
    public String getDisplayName() {
        return displayName;
    }

    @Override
    public String getDescription() {
        return description;
    }
}
```

If so, we could provide users with more readable names / descriptions and apply minor simplifications in property declaration, see other comments.

nifi-extension-bundles/nifi-gitlab-bundle/nifi-gitlab-extensions/src/main/java/org/apache/nifi/gitlab/GitLabFlowRegistryClient.java
```diff
@@ -0,0 +1,145 @@
...
+            .description("The name of the repository")
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .required(true)
+            .build();
+
+    static final PropertyDescriptor AUTHENTICATION_TYPE = new PropertyDescriptor.Builder()
+            .name("Authentication Type")
+            .description("The type of authentication to use for accessing GitLan")
+            .allowableValues(GitLabAuthenticationType.class)
+            .defaultValue(GitLabAuthenticationType.ACCESS_TOKEN.name())
```
EndzeitBegins, line 71: If `GitLabAuthenticationType` was a `DescribedValue` this could be simplified minimally to:

```suggestion
            .defaultValue(GitLabAuthenticationType.ACCESS_TOKEN)
```

nifi-extension-bundles/nifi-gitlab-bundle/nifi-gitlab-extensions/src/main/java/org/apache/nifi/gitlab/GitLabFlowRegistryClient.java
```diff
@@ -0,0 +1,145 @@
...
+            .required(true)
+            .build();
+
+    static final PropertyDescriptor ACCESS_TOKEN = new PropertyDescriptor.Builder()
+            .name("Access Token")
+            .description("The access token to use for authentication")
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .required(true)
+            .sensitive(true)
+            .dependsOn(AUTHENTICATION_TYPE, GitLabAuthenticationType.ACCESS_TOKEN.name())
```
EndzeitBegins, line 81: If `GitLabAuthenticationType` was a `DescribedValue` this could be simplified minimally to:
```suggestion
            .dependsOn(AUTHENTICATION_TYPE, GitLabAuthenticationType.ACCESS_TOKEN)
```

EndzeitBegins (approved):
As written in Jira before, thank you for introducing support for GitLab @bbende. 

I've looked over the code changes which look good overall. I've added some remarks, none of which are deal-breaker from my point of view, just some suggestions.
 
However, I hadn't had the chance to test the new client against a running GitLab instance yet.

nifi-extension-bundles/nifi-github-bundle/nifi-github-extensions/src/main/java/org/apache/nifi/github/GitHubRepositoryClient.java
bbende: The `toList()` call here is from GH client's `PagedIterable` and its a way to eagerly fetch the whole list, which we want since the NiFi UI currently doesn't have a way to page through the versions of the flows. 

It could be possible to try and return the converted list in one big statement by adding:
```
.stream()
.map(this::toGitCommit)
.toList();
```
But the `toGitCommit` method throws `IOException` because of one of the model classes throwing `IOException`, and then you can't have the checked exception inside the stream statements, so that is why I ended up with the second block of code to convert it to the other list.

nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/client/GitCommit.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.Date;
+import java.util.Objects;
+
+public class GitCommit {
+
+    private final String id;
+    private final String author;
+    private final String message;
+    private final Date commitDate;
```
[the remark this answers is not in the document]
bbende: Both of the underlying libraries (`github-api` and `gitlab4j`) both return `Date` in their commit object, so I was mostly just passing the same thing through, but would be easy to use `Instant`, I'll take a look.

nifi-extension-bundles/nifi-gitlab-bundle/nifi-gitlab-extensions/src/main/java/org/apache/nifi/gitlab/GitLabAuthenticationType.java
bbende: Yes good idea, this makes sense, I'll make these changes related to use of `DescribedValue`.

nifi-extension-bundles/nifi-github-bundle/nifi-github-extensions/src/main/java/org/apache/nifi/github/GitHubRepositoryClient.java
EndzeitBegins: Yes, I just thought the call to `.toList()` after `.list()` is not necessary as we immediately iterate over all elements to create our own `List` anyway. 
I assumed the iteration could run over the `PagedIterable` directly instead of creating an intermediate `List` object. 

Didn't notice `toGitCommit` throwing an `IOException` making it less suited for the `Stream` api. Thanks for pointing that out.

bbende: If we don't call `toList()` then `list()` returns one page, and we'd have use the results of the `PagedIterable` to keep making more calls to `repository.queryCommits()` for the rest of the pages, which is exactly what `toList()` is doing for us.

bbende:
@EndzeitBegins thanks for the review! pushed a commit addressing most of the suggestions

nifi-extension-bundles/nifi-github-bundle/nifi-github-extensions/src/main/java/org/apache/nifi/github/GitHubRepositoryClient.java
EndzeitBegins: I see, thanks for clarifying.

nifi-extension-bundles/nifi-gitlab-bundle/nifi-gitlab-extensions/src/main/java/org/apache/nifi/gitlab/PersonalAccessToken.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public boolean isActive() {
+        return active;
+    }
+
+    public void setActive(final boolean active) {
+        this.active = active;
+    }
+}
```
EndzeitBegins, line 111: Can we use a `record` for this DTO? From what I know newer versions of Jackson should be able so serialise JSON into record classes.

Additionally, do we want to stick to camelCase for field names and provide a explicit `@JsonProperty` for fields that use snake_case?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardFlowRegistryDAO.java
```diff
@@ -116,12 +116,19 @@ public Set<FlowRegistryBranch> getBranchesForUser(final FlowRegistryClientUserCo
...
                 throw new IllegalArgumentException("The specified registry id is unknown to this NiFi.");
             }
+            final FlowRegistryBranch defaultBranch = flowRegistry.getDefaultBranch(context);
             final Set<FlowRegistryBranch> branches = flowRegistry.getBranches(context);
-            final Set<FlowRegistryBranch> sortedBranches = new TreeSet<>(Comparator.comparing(FlowRegistryBranch::getName));
+            // Sort the default branch first, this allows main to always be the first
+            final Set<FlowRegistryBranch> sortedBranches = new TreeSet<>((branch1, branch2) -> {
+                if (branch1.getName().equals(defaultBranch.getName())) {
+                    return -1;
+                }
```
EndzeitBegins, line 125: Is it enough to handle `branch1` being the default branch here?
Or should we handle the case that the default branch is provided as `branch2` as well?

```suggestion
                if (branch1.getName().equals(defaultBranch.getName())) {
                    return -1;
                } else if (branch2.getName().equals(defaultBranch.getName())) {
                    return 1;
                } 
```

nifi-extension-bundles/nifi-gitlab-bundle/nifi-gitlab-extensions/src/main/java/org/apache/nifi/gitlab/GitLabFlowRegistryClient.java
```diff
@@ -0,0 +1,145 @@
...
+    static final PropertyDescriptor REPOSITORY_NAME = new PropertyDescriptor.Builder()
+            .name("Repository Name")
+            .description("The name of the repository")
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .required(true)
+            .build();
+
+    static final PropertyDescriptor AUTHENTICATION_TYPE = new PropertyDescriptor.Builder()
+            .name("Authentication Type")
+            .description("The type of authentication to use for accessing GitLan")
```
EndzeitBegins, line 69:
```suggestion
            .description("The type of authentication to use for accessing GitLab")
```

EndzeitBegins (approved):
Thank you for addressing my feedback and your overall work on this new feature @bbende.

I've found a few smaller nitpicks, you might (or not) want to consider before merging..

bbende:
Pushed another update to address the additional feedback

nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
```diff
@@ -0,0 +1,644 @@
...
+
+    @Override
+    public RegisteredFlow deregisterFlow(final FlowRegistryClientConfigurationContext context, final FlowLocation flowLocation) throws FlowRegistryException, IOException {
+        final GitRepositoryClient repositoryClient = getRepositoryClient(context);
+        verifyWritePermissions(repositoryClient);
+
+        final String branch = flowLocation.getBranch();
+        final String filePath = getSnapshotFilePath(flowLocation);
+        final String commitMessage = DEREGISTER_FLOW_MESSAGE_FORMAT.formatted(flowLocation.getFlowId());
+        final InputStream deletedSnapshotContent = repositoryClient.deleteContent(filePath, commitMessage, branch);
```
exceptionfactory, line 240: Is there a reason for not calling this in a try-with-resources?

nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
```diff
@@ -0,0 +1,644 @@
...
+            }
+        }
+        repositoryClient = null;
+    }
+
+    // If the client has write permissions to the repo, then ensure the directory for the default bucket is present and if not create it,
+    // otherwise the client can only be used to import flows from the repo and won't be able to set up the default bucket
+    private void initializeDefaultBucket(final FlowRegistryClientConfigurationContext context) throws IOException, FlowRegistryException {
+        if (!repositoryClient.getCanWrite()) {
+            getLogger().info("Repository client does not have write permissions to the repository, skipping setup of default bucket");
```
exceptionfactory, line 591: It would be helpful to include the client identifier or some other reference to this instance for troubleshooting.

nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
```diff
@@ -0,0 +1,644 @@
...
+            return;
+        }
+
+        final String branch = context.getProperty(REPOSITORY_BRANCH).getValue();
+        final Set<String> bucketDirectoryNames = repositoryClient.getTopLevelDirectoryNames(branch);
+        if (!bucketDirectoryNames.isEmpty()) {
+            getLogger().info("Found existing buckets, skipping setup of default bucket");
```
exceptionfactory, line 598: Is this necessary to be an `info` message, or should it be a `debug`? Does it make sense to include the bucketDirectoryNames? Perhaps just the number of directories would be helpful.

nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/client/GitRepositoryClient.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Client for a git repository.
+ */
+public interface GitRepositoryClient {
+
+    /**
+     * @return true if the client can read from the configured repository, false otherwise
+     */
+    boolean getCanRead();
```
exceptionfactory, line 36: Should this be named `hasReadPermission`?

nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/client/GitRepositoryClient.java
```diff
@@ -0,0 +1,147 @@
...
+     * Deletes the file at the given path on the given branch.
+     *
+     * @param filePath the path of the file
+     * @param commitMessage the commit message
+     * @param branch the branch
+     * @return the input stream to the deleted content
+     * @throws IOException if an I/O error occurs
+     * @throws FlowRegistryException if a non-I/O error occurs
+     */
+    InputStream deleteContent(String filePath, String commitMessage, String branch) throws FlowRegistryException, IOException;
```
exceptionfactory, line 137: It seems like the caller is responsible to close the stream? That would be worth noting one way or the other.

nifi-extension-bundles/nifi-gitlab-bundle/nifi-gitlab-extensions/src/main/java/org/apache/nifi/gitlab/GitLabRepositoryClient.java
```diff
@@ -0,0 +1,496 @@
...
+        final int responseCode;
+        final String responseContent;
+        final HttpURLConnection connection;
+        try {
+            connection = createConnection(PERSONAL_ACCESS_TOKENS_SELF_PATH);
+            responseCode = connection.getResponseCode();
+            responseContent = IOUtils.toString(connection.getInputStream(), StandardCharsets.UTF_8);
+        } catch (final Exception e) {
+            throw new FlowRegistryException("Unable to retrieve personal access token details", e);
+        }
```
exceptionfactory, line 342: Recommend adding a `finally` block to call `connection.disconnect()`

nifi-extension-bundles/nifi-gitlab-bundle/nifi-gitlab-extensions/src/main/java/org/apache/nifi/gitlab/GitLabRepositoryClient.java
```diff
@@ -0,0 +1,496 @@
...
+
+    private static final String PERSONAL_ACCESS_TOKENS_SELF_PATH = "/personal_access_tokens/self";
+    private static final String PRIVATE_TOKEN_HEADER  = "PRIVATE-TOKEN";
+    private static final String READ_API_SCOPE = "read_api";
+    private static final String WRITE_API_SCOPE = "api";
+    private static final String DIRECTORY_MODE = "040000";
+
+    private static final int DEFAULT_ITEMS_PER_PAGE = 100;
+    private static final int UNAUTHORIZED = 403;
+    private static final int NOT_FOUND = 404;
```
exceptionfactory, line 75: The HttpURLConnection static values could be used.

exceptionfactory (commented):
Great work on this new Flow Registry Client @bbende! I noted a couple questions, particular around stream and connection handling. The abstracted Git Flow Client looks very useful for current and future implementations.

bbende:
Thanks @exceptionfactory , pushed an update to address your suggestions

exceptionfactory (approved):
Great work on this integration with GitLab @bbende, and thanks for the review @EndzeitBegins! The latest version looks good! +1 merging

PR #9273, opened 2024-09-17
NIFI-13755 On NiFi restart, re-enable as many previously enabled controller services as possible 

jrsteinebrey:
(meaning all the ones which are not dependent on another disabled service).
Only skip trying to re-enable a previously disabled controller service IF it depends on another service which is disabled.

# Summary

[NIFI-13755](https://issues.apache.org/jira/browse/NIFI-13755)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-13755) issue created

### Pull Request Tracking

- [X] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-13755`
- [X] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-13755`

### Pull Request Formatting

# Verification

1) I added and updated automated tests to exercise the new behavior
2) I tested with running NiFi server for several disabled and enabled state services/combinations and it worked as ticket describes it should.

### Build

  - [X] JDK 21

### Licensing

No new dependencies

### Documentation

No documentation changes

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/service/StandardControllerServiceProvider.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/service/StandardControllerServiceProvider.java
@@ -45,7 +45,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/service/StandardControllerServiceProvider.java
```diff
@@ -253,38 +252,44 @@ public CompletableFuture<Void> enableControllerService(final ControllerServiceNo
...
             }
+            if (isStartable) {
+                startableServiceNodes.add(serviceNode);
+            }
+        }
+        if (startableServiceNodes.size() < serviceNodes.size()) {
+            // If any service was removed, then recheck all remaining services because the removed one might be required by another service.
+            startableServiceNodes = filterStartableControllerServices(startableServiceNodes);
         }
+        return startableServiceNodes;
```
exceptionfactory, line 292: This approach to build a `HashSet` for every invocation and recursively call this method seems like it could get expensive for flows with larger numbers of Controller Services.

exceptionfactory (changes requested):
@jrsteinebrey The Jira issue provides some motivation behind this change, but based on that example, the current approach seems questionable given the need to create multiple HashSets for services.

More to the point, however, the scenario described changes the effective status of Controller Services that may have been set as disabled. Is it correct to read the change as unconditionally enabling disabled services, provided they have no dependencies? If so, this should not go forward.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/service/StandardControllerServiceProvider.java
jrsteinebrey: @exceptionfactory Thanks for reviewing this and giving feedback.

This method being changed: 
StandardControllerServiceProvider.enableControllerServices(
        final Collection<ControllerServiceNode> **serviceNodesIn**)

In the existing code and in the new code, the controller services in the **serviceNodes** parameter
contains only controller services which are supposed to be enabled. 
The caller decides which services it intends to enable.
If a controller service is not supposed to be started on Nifi restart (i.e. scheduledState is disabled), then it should not be (and is not) put in **serviceNodes** by the existing calling code.

To recap, the method I am changing receives a list of only the services that the caller wants to enable (based on the caller's decision logic). It does **not**  get a list of all controller services when NiFi is being restarted.

jrsteinebrey: So this change does **not** enable any controller services which the caller of this method did **not** ask to enable. And does not introduce any new side effect of enabled unwanted controller services.

jrsteinebrey: @exceptionfactory I refactored the code based on your comments on its structure.

mattyb149:
+1 LGTM, verified the expected behavior. @exceptionfactory do you want another look at this? If not I'll merge

exceptionfactory:
Thanks @mattyb149, I will take another look and follow up soon.

jrsteinebrey:
@exceptionfactory Can you please let me know what you think of this updated PR  when you have a chance?

mattyb149:
@exceptionfactory Do you think you'll be able to finish your review this week? Or if you're satisfied I can do the merge, thanks!

exceptionfactory:
> @exceptionfactory Do you think you'll be able to finish your review this week? Or if you're satisfied I can do the merge, thanks!

Thanks for the reminder @mattyb149, yes, I should be able to finish a review this week.

exceptionfactory (commented):
Thanks for your patience with this @jrsteinebrey and @mattyb149.

The updated approach that eliminates the duplicative HashSet creation looks much better.

The only outstanding question is whether the existing behavior that treats the input Controller Services as a unit of work needs should be altered in all cases. Flagging with @markap14 for a final review.

markap14:
Thanks for pinging @exceptionfactory. I do think the change is reasonable. Looking at how the method gets used, I don't see anywhere that there is a strict requirement for atomicity. So I'm fine with making these changes. I do see @exceptionfactory that Github still shows '1 change requested' from you, which I assume is no longer relevant. But I'll hold off on clicking the "Squash and merge" button to make sure. But I'm a +1. Thanks all!

exceptionfactory (approved):
Thanks for the confirmation @markap14, and thanks again for the work on this @jrsteinebrey! +1 merging

PR #9290, opened 2024-09-20
NIFI-13744 2.x Resolved ExcelReader time conversion issues by ensuring the same string value used for determining the record field type is the same value used for reading the record value.

dan-s1:
# Summary

[NIFI-13744](https://issues.apache.org/jira/browse/NIFI-13744)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
```diff
@@ -45,8 +46,14 @@ public class ExcelRecordReader implements RecordReader {
     private final String dateFormat;
     private final String timeFormat;
     private final String timestampFormat;
+    private final DataFormatter dataFormatter;
 
     public ExcelRecordReader(ExcelRecordReaderConfiguration configuration, InputStream inputStream, ComponentLog logger) throws MalformedRecordException {
+        this(configuration, inputStream, logger, null);
+    }
+
+    public ExcelRecordReader(ExcelRecordReaderConfiguration configuration, InputStream inputStream, ComponentLog logger, Locale locale) throws MalformedRecordException {
```
exceptionfactory, line 55: Instead of adding the new constructor, recommend changing all references to the single constructor with the `Locale` required, defaulting to `Locale.getDefault()`

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
```diff
@@ -127,17 +134,6 @@ private Map<String, Object> getCurrentRowValues(Row currentRow, boolean coerceTy
         return currentRowValues;
     }
 
-    private static Object getCellValue(Cell cell) {
-        if (cell != null) {
-            return switch (cell.getCellType()) {
-                case _NONE, BLANK, ERROR, FORMULA, STRING -> cell.getStringCellValue();
-                case NUMERIC -> DateUtil.isCellDateFormatted(cell) ? cell.getDateCellValue() : cell.getNumericCellValue();
-                case BOOLEAN -> cell.getBooleanCellValue();
```
exceptionfactory, line 135: Does the `DataFormatter` sufficiently handle all of these cases now?

exceptionfactory (changes requested):
Thanks for the work on this issue @dan-s1. The changes look good in general. I just noted one recommendation on the constructor arguments, and one question regarding the switch to `DataFormatter` for cell value handling.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
dan-s1: I believe it does but I will add a unit test to demonstrate it does.

dan-s1: @exceptionfactory I am glad you brought up this point since as a result I discovered there was bug in formula results as they were not being displayed rather the actual formula was being retrieved. I have fixed that with the latest changes and I have expanded the ExcelReader unit tests to exercise all types.

dan-s1: @exceptionfactory I see the build is failing on the Windows box which has the French Locale and in France the separator is a comma and not a period. The stacktrace I see is

```
Caused by: java.lang.NumberFormatException: For input string: "18,36"
	at java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054)
	at java.base/jdk.internal.math.FloatingDecimal.parseFloat(FloatingDecimal.java:122)
	at java.base/java.lang.Float.parseFloat(Float.java:556)
	at org.apache.nifi.serialization.record.util.DataTypeUtils.toFloat(DataTypeUtils.java:1330)
	at org.apache.nifi.serialization.record.util.DataTypeUtils.convertType(DataTypeUtils.java:209)
	at org.apache.nifi.serialization.record.util.DataTypeUtils.convertType(DataTypeUtils.java:174)
	at org.apache.nifi.excel.ExcelRecordReader.convert(ExcelRecordReader.java:134)
	at org.apache.nifi.excel.ExcelRecordReader.lambda$getCurrentRowValues$0(ExcelRecordReader.java:119)
	at java.base/java.util.stream.Streams$RangeIntSpliterator.forEachRemaining(Streams.java:104)
	at java.base/java.util.stream.IntPipeline$Head.forEach(IntPipeline.java:617)
	at org.apache.nifi.excel.ExcelRecordReader.getCurrentRowValues(ExcelRecordReader.java:106)
```

Can this be a bug in `DataTypeUtils` which does not take into account the `Locale` when parsing a string to a number?

exceptionfactory: I think the strategy of using the `DataFormatter` in all cases is not great because of the string localization. The `getNumericValue()` method seems like a better option if that works consistently.

dan-s1: The problem I see is I have to eventually pass it as a string to `SchemaInferenceUtil.getDataType` . I have to ensure the `toString` method of `Double` does not take the locale into account.

dan-s1: Seems like it wasn't an issue as the build passed.

dan-s1:
@exceptionfactory I know you are extremely busy but if you have some cycles can you please review my recent changes to see if they meet what you had in mind? Thanks!

exceptionfactory:
Thanks @dan-s1, I hoping to get back to this issue during this week.

dan-s1:
Any chance this can get to 1.28 (i.e. I would backport these changes)?  My team could personally use these changes. Thanks!

briansolo1985:
Sure @dan-s1 , it can get in. I plan to start the release process next week.

dan-s1:
@briansolo1985 Thanks I appreciate it. But this one still needs to be approved before I back port it.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelReader.java
```diff
@@ -0,0 +1,160 @@
...
+        excelReader = new ExcelReader();
+        properties = new HashMap<>();
+        new ExcelReader().getSupportedPropertyDescriptors().forEach(prop -> properties.put(prop, prop.getDefaultValue()));
+    }
+
+    @Test
+    void testReadWithInferredSchema() throws Exception {
+        properties.put(DateTimeUtils.DATE_FORMAT, "M/dd/yy");
+        properties.put(DateTimeUtils.TIME_FORMAT, "HH:mm");
+        properties.put(DateTimeUtils.TIMESTAMP_FORMAT, "M/dd/yyyy HH:mm:ss");
```
exceptionfactory, line 97: It would be helpful to declare static values for the date, time, and timestamp formats, which can be reused across this class.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
```diff
@@ -141,15 +140,15 @@ public void testMultipleRecordsSingleSheet() throws MalformedRecordException {
...
+        List<Record> records = getRecords(recordReader, false);
 
         assertEquals(9, records.size());
     }
 
     private RecordSchema getDataFormattingSchema() {
         final List<RecordField> fields = Arrays.asList(
                 new RecordField("Numbers", RecordFieldType.DOUBLE.getDataType()),
-                new RecordField("Timestamps", RecordFieldType.DATE.getDataType()),
+                new RecordField("Timestamps", RecordFieldType.STRING.getDataType()),
```
exceptionfactory, line 151: Is this type change just to simplify the test, with the understanding that `TestExcelReader` covers the various field types?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelUtils.java
```diff
@@ -16,15 +16,45 @@
...
+    public static String getFormattedCellValue(Cell cell) {
+        if (cell != null) {
+            /* DataFormatter should be used in most cases unless an exception will be thrown or
+              string localization may occur which could result in later exceptions. A cell which has a formula cannot
+              format a cell which has an error value and an exception will be thrown. In addition, a cell whose type
+              is numeric when formatted will take the current locale into consideration which could prevent later parsing
+              as a float/double.*/
+            if (cell.getCellType().equals(CellType.FORMULA) && cell.getCachedFormulaResultType().equals(CellType.ERROR)) {
+                return cell.getStringCellValue();
+            } else if (cell.getCellType().equals(CellType.NUMERIC) && !DateUtil.isCellDateFormatted(cell)) {
```
exceptionfactory, line 51: It looks like it would be helpful to declare a variable for the `cellType`, making these conditionals a bit more readable than calling `getCellType()` in both blocks.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelUtils.java
```diff
@@ -16,15 +16,45 @@
...
+                final double numericCellValue = cell.getNumericCellValue();
+                return numericCellValue % 1 == 0 ? Long.toString((long) numericCellValue) : Double.toString(numericCellValue);
```
exceptionfactory, line 53: Is there any other more optimized way to determine whether the value is a `long` or a `double` with having to do the modulo operation?

exceptionfactory (changes requested):
Thanks for your patience on this @dan-s1.

Reviewing the changes in more detail, I am a bit concerned about the implications of converting all cell values to `String` as opposed to keeping the native type. Did you consider an approach that changes the schema inference approach, as opposed changing the reading everything as a string?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
dan-s1: Yes

exceptionfactory:
@dan-s1, I'm taking a closer look at the implementation, and I believe more significant work is necessary. The SchemaInferenceUtil is not very helpful for the ExcelReader, because it requires a `String` input as you noted. I've made some initial progress on an alternative approach, and plan to follow up when I have more details.

dan-s1:
@exceptionfactory Can I give it a shot to refactor these?

exceptionfactory:
> @exceptionfactory Can I give it a shot to refactor these?

Thanks @dan-s1, I had already made good progress on a refactored approach, so I put together PR #9379. The changes there remove the `SchemaInferenceUtil` usage and make direct use of Cell Type and Cell Style information. I highlighted some of the timestamp challenges in the description, but it provided better handling of date and time values from the test spreadsheet on the Jira issue. However, it would be great to get some additional testing of the changes, so we can collaborate on the approach.

exceptionfactory:
Closing in favor of #9379. Thanks again for the initial work here @dan-s1, it was helpful for initial evaluation of the issues. Let's continue collaborating on the other PR.

PR #9293, opened 2024-09-20
NIFI-4146 Limit count of split FlowFiles from SplitRecord for large record sets

EndzeitBegins:
# Summary

Adresses [NIFI-4146](https://issues.apache.org/jira/browse/NIFI-4146) by limiting the count of FlowFiles created in a single run of `SplitRecord` to a configurable value. I've set the default to `10000` in accordance with the default FlowFile queue backpressure object count.

If more than the configured value of split FlowFiles are needed, the original FlowFile is pushed back to the upstream queue with status information on already processed records added in attributes. This allows splitting large record sets into single or few records per FlowFile iteratively while using standard back pressure mechanisms to limit the count of FlowFiles in the Flow at once. 

While "flushing in-flight sessions with lots of flowfile references to disk" as suggested in a comment on the Jira issue would address the memory problem as well, it would still result in lots of FlowFiles in the downstream queue which users might not desire.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/SplitRecord.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/SplitRecord.java
@@ -35,7 +35,6 @@ import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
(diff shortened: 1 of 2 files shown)


exceptionfactory (changes requested):
Thanks for proposing a solution to this issue with `SplitRecord` @EndzeitBegins. On initial review, however, solving the problem in the Processor itself does not seem like the right approach. It is worth noting that generating a very large number of FlowFiles is an anti-pattern, so trying to accommodate the use case of generating 100,000 split FlowFiles is problematic. I think it would be better to reconsider the options in the Jira discussion before proceeding.

PR #9315, opened 2024-09-26
[NIFI-13806] finalize color hues

scottyaslan:
The secondary darker color for both light and dark mode has been updated to use the appropriate hue. 

The primary palette in dark mode has been update to match the secondary palette. 

Shared palettes have been encapsulated to ensure these colors do not get update in either light or dark mode. Only the neutral and neutral variants should be the only real differences between light and dark mode palettes.


mcgilman (approved):

PR #9317, opened 2024-09-26
NIFI-13803 Improve loading of flow when clustered to only sync into t…

bbende:
…he coordinator after getting cluster response

# Summary

[NIFI-13803](https://issues.apache.org/jira/browse/NIFI-13803)

Previously when clustered, the node would first load the flow from disk and  sync FlowController with the local flow, then attempt to connect to the cluster, and if a response was received it would then re-sync the controller with the flow from the cluster.

This changes the logic so that the node doesn't bother syncing the local flow into the controller anymore. It will send the local flow to the coordinator in the connect request, and only when receiving the connected response will it take the flow from the response and sync it into the controller. In addition, if the initial connect request does not succeed, it will not long call initialize on the flow controller, it will just wait until it has been tried again and receives a successful response.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/StandardFlowService.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/StandardFlowService.java
@@ -16,7 +16,6 @@
  */
 package org.apache.nifi.controller;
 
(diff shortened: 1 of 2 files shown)


exceptionfactory (approved):
Thanks for this improvement @bbende, verified on a local cluster, looks good! +1 merging

PR #9328, opened 2024-10-01
[NIFI-13821] add Max Concurrent Tasks and Stateless Flow Timeout form…

scottyaslan:
… fields when PG Execute Engine is STATELESS
Screenshot 2024-10-01 at 11 26 07 AM


mcgilman (approved):
Thanks @scottyaslan!

PR #9330, opened 2024-10-01
[NIFI-13778] markdown syntax highlighting and h1-h5, tables, ul, and ol

scottyaslan:
Screenshot 2024-09-27 at 5 55 16 PM


nifi-frontend/src/main/frontend/libs/shared/src/assets/styles/_prism-theme.scss
```diff
@@ -0,0 +1,187 @@
...
+        line-height: 1.5;
+
+        -moz-tab-size: 4;
+        -o-tab-size: 4;
+        tab-size: 4;
+
+        -webkit-hyphens: none;
+        -moz-hyphens: none;
+        -ms-hyphens: none;
+        hyphens: none;
```
sardell, line 61: Are these vendor prefixes needed? `hyphens` has been supported in all major browsers since last year, and Nx leverages autoprefixer to include vendor prefixes in builds.

scottyaslan: I think we should probably keep them for now. I started this file based on the provided themes from prismjs in `node_modules/prismjs/themes`. Each of those examples also contain these rules so I think the nifi prismjs theme should too.

sardell: Okay, no problem. I just wanted to mention because our CSS output from the build should generate all of these prefixes for the nifi prismjs theme, which means less code for us to write/maintain.

rfellows (approved):
This is a great new feature in the documentation. Excellent work, @scottyaslan!

PR #9335, opened 2024-10-02
NIFI-13822: Ensuring the parameter value is always included when crea…

mcgilman:
…ting new even if it is unchanged from the initial property value.


rfellows (approved):
Nice fix, @mcgilman.

PR #9337, opened 2024-10-02
NIFI-13823: Using startsWith testing content viewer compatibility to …

mcgilman:
…better support content types that might have additional parameters.


rfellows (approved):
Looks good to me. 👍

PR #9338, opened 2024-10-02
NIFI-13831: Adding inheritance to versioned component synchronizer pa…

gresockj:
…rameter context synchronization when considering referencing components to restart

# Summary

[NIFI-13831](https://issues.apache.org/jira/browse/NIFI-13831)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -1685,10 +1685,19 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
 
     protected Set<String> getUpdatedParameterNames(final ParameterContext parameterContext, final VersionedParameterContext proposed) {
         final Map<String, ParameterValueAndReferences> originalValues = new HashMap<>();
-        parameterContext.getParameters().values().forEach(param -> originalValues.put(param.getDescriptor().getName(), getValueAndReferences(param)));
+        parameterContext.getEffectiveParameters().forEach((pd, param) -> originalValues.put(pd.getName(), getValueAndReferences(param)));
 
         final Map<String, ParameterValueAndReferences> proposedValues = new HashMap<>();
         if (proposed != null) {
+            if (proposed.getInheritedParameterContexts() != null) {
+                for (int i = proposed.getInheritedParameterContexts().size() - 1; i >= 0; i--) {
+                    final String name = proposed.getInheritedParameterContexts().get(i);
+                    final ParameterContext inheritedContext = getParameterContextByName(name);
+                    if (inheritedContext != null) {
+                        inheritedContext.getEffectiveParameters().forEach((pd, param) -> proposedValues.put(pd.getName(), getValueAndReferences(param)));
+                    }
+                }
+            }
             proposed.getParameters().forEach(versionedParam -> proposedValues.put(versionedParam.getName(), getValueAndReferences(versionedParam)));
         }
 
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -1683,8 +1683,14 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
         }
     }
 
+    private void collectValueAndReferences(final ParameterContext parameterContext, final Map<String, ParameterValueAndReferences> valueAndRef) {
+        parameterContext.getEffectiveParameters()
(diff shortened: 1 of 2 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -1685,10 +1685,19 @@ public void synchronize(final ParameterContext parameterContext, final Versioned
 
     protected Set<String> getUpdatedParameterNames(final ParameterContext parameterContext, final VersionedParameterContext proposed) {
         final Map<String, ParameterValueAndReferences> originalValues = new HashMap<>();
-        parameterContext.getParameters().values().forEach(param -> originalValues.put(param.getDescriptor().getName(), getValueAndReferences(param)));
+        parameterContext.getEffectiveParameters().forEach((pd, param) -> originalValues.put(pd.getName(), getValueAndReferences(param)));
```
bejancsaba, line 1688: As this code part is essentially reused what do you think about just extracting it to its own helper function like:
```
    private void collectValueAndReferences(final ParameterContext parameterContext, final Map<String, ParameterValueAndReferences> valueAndRef) {
        parameterContext.getEffectiveParameters()
                .forEach((pd, param) -> valueAndRef.put(pd.getName(), getValueAndReferences(param)));
    }
```

You could go one step further and collecting / returning the map from the function (not expecting it as parameter) but that would mean some flatMap in the logic below and I'm not sure what is the appetite for that.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -1685,10 +1685,19 @@ public void synchronize(final ParameterContext parameterContext, final Versioned
...
 
         final Map<String, ParameterValueAndReferences> proposedValues = new HashMap<>();
         if (proposed != null) {
+            if (proposed.getInheritedParameterContexts() != null) {
```
bejancsaba, line 1692: With the changes above this could be made somewhat more functional:
```
            Optional.ofNullable(proposed.getInheritedParameterContexts())
                    .map(List::reversed)
                    .ifPresent(x -> x.forEach(name -> Optional.ofNullable(getParameterContextByName(name))
                            .ifPresent(inheritedContext -> collectValueAndReferences(inheritedContext, proposedValues))));
```
What do you think? Just an idea on the other hand list reverting won't apply cleanly to the 1.x branch so maybe it is cleaner to leave it as is. I will let you decide please let me know. Otherwise I'm +1, thanks for the test explaining the functionality.

bejancsaba (commented):
I had two smaller comments please take a look otherwise I'm ok to merge. The added test helps a lot thanks for that, what do you think about creating dedicated test for that functionality I understand it will result in some duplication but would be cleaner when it fails.

joewitt:
"Just an idea on the other hand list reverting won't apply cleanly to the 1.x branch so maybe it is cleaner to leave it as is. I will let you decide please let me know."

It is not ok to do anything for NiFi 2.x code to make it clean/easy/etc.. to backport.  NiFi 2.x is and needs to be able to breath/grow/transform on its own.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
markap14: We should very much avoid trying to make things functional. We should always prefer Imperative unless there is a strong reason to use functional. I.e., in cases where the imperative is verbose but functional is trivial, such as:
```
final List<String> filtered = values.stream().filter(s -> s.contains("hello")).toList();
```
can be used over:
```
final List<String> filtered = new ArrayList<>();
for (final String value : values) {
  if (value.contains("hello") {
     filtered.add(value);
  }
}
```
Because it is simpler to understand the code and the intent. However, this code as-is, in its imperative form, is much simpler than the functional form.

gresockj: I will leave it as is, in keeping with @markap14's suggestion.

bejancsaba: Understood, it is ok from my side. I find functional code to be much easier to read and understand as it shows "what" is happening in contrast of imperative where it is more about "how" we achive our goal. I also agree that there are cases where functional approach could be more involved. In the end it is subjective I suppose. If the community prefers imperative then code consistency is more valuable.

bejancsaba (approved):
+1 from my side. Thanks for the small adjustment. My other comment was discussed I'm ok to merge but will leave a little time for others to chime in just in case there are concerns.

PR #9344, opened 2024-10-03
NIFI-13534: Only reloading the flow when the browser tab is visible. …

mcgilman:
…This prevents attempts to re-render the SVG when the browser tab is hidden.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/app.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/app.component.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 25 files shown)


mcgilman:
While troubleshooting the issues addressed in this PR I discovered that the application still is affected by [1]. Once the fix has been released and we'll want to upgrade to it as well.

[1] https://github.com/angular/components/issues/29696

rfellows:
will review...

rfellows (approved):
Nice work, @mcgilman. This really helps limit/eliminate the pointless polling for canvas changes when the user isn't even viewing the NiFi tab.
Also cleans up lots of things that were hanging around longer than they needed to be.

PR #9349, opened 2024-10-04
NIFI-13790: Report SEND provenance event in PublishKafka

mattyb149:
# Summary

[NIFI-13790](https://issues.apache.org/jira/browse/NIFI-13790) This PR reports a SEND event for each FlowFile successfully published and transferred to success. I was hesitant to add a `getTransitUri()` method to the KafkaConnectionService API but I certainly can in order to give a more informative URI, and could add other methods if need be to generate an informative description for the provenance event. I could also wrap the new code in a `try-catch` catching and logging an error versus rolling back / failing because the FlowFiles will have already been transferred.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/PublishKafka.java
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/PublishKafka.java
@@ -485,6 +485,9 @@ public class PublishKafka extends AbstractProcessor implements KafkaPublishCompo
 
             final Relationship relationship = flowFileResult.getExceptions().isEmpty() ? REL_SUCCESS : REL_FAILURE;
             session.transfer(flowFile, relationship);
+            final int numTopics = flowFileResult.getSentPerTopic().keySet().size();
+            final String topicList = String.join(",", flowFileResult.getSentPerTopic().keySet().toArray(new String[numTopics]));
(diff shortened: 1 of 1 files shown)


nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/PublishKafka.java
```diff
@@ -485,6 +485,9 @@ private void routeResults(final ProcessSession session, final List<FlowFileResul
 
             final Relationship relationship = flowFileResult.getExceptions().isEmpty() ? REL_SUCCESS : REL_FAILURE;
             session.transfer(flowFile, relationship);
+            final int numTopics = flowFileResult.getSentPerTopic().keySet().size();
+            final String topicList = String.join(",", flowFileResult.getSentPerTopic().keySet().toArray(new String[numTopics]));
```
MikeThomsen, line 489: The `toArray` call is unnecessary because it'll take an `Iterable`. I'll fix and commit.

MikeThomsen (approved):

exceptionfactory (commented):
Just noting for reference, this should be revisited. Having a transit URI without the remote system, and having a list of topics, does not provide sufficient granularity to be useful.

joewitt:
We definitely want a weak send event rather than no send event.  But yeah we should revisit.

PR #9363, opened 2024-10-08
NIFI-13674 Added merge.reason attribute for MergeRecord.

dan-s1:
# Summary

[NIFI-13674](https://issues.apache.org/jira/browse/NIFI-13674)

I copied the description of the `WritesAttribute`  annotation from `MergeContent` `merge.reason`. Please confirm whether that is okay.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/MergeRecord.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/MergeRecord.java
@@ -96,7 +96,7 @@ import java.util.stream.Collectors;
     @WritesAttribute(attribute = "merge.bin.age", description = "The age of the bin, in milliseconds, when it was merged and output. Effectively "
         + "this is the greatest amount of time that any FlowFile in this bundle remained waiting in this processor before it was output"),
     @WritesAttribute(attribute = "merge.uuid", description = "UUID of the merged FlowFile that will be added to the original FlowFiles attributes"),
-    @WritesAttribute(attribute = MergeRecord.REASON_FOR_MERGING, description = "This processor allows for several thresholds to be configured for merging FlowFiles. "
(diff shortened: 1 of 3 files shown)


exceptionfactory (changes requested):
Thanks for adding this attribute to `MergeRecord` @dan-s1. For more programmatic-focused handling, I recommend creating an `enum` and using the enum name for the reason, as opposed to a string explanation. That would following the pattern of the `EvictionReason` in `MergeContent`.

dan-s1:
Just to clarify you want method signatures of 
`RecordBinManager.handleCompletedBins(final Predicate<RecordBin> completionTest, final String completionReason)`
and
`RecordBin.complete(final String completionReason)`

to now take an enum value instead of a string?

Also the values for completionReason I see are 

1. Bin has reached Max Bin Age
2. Bin is full enough
3. Bin is completely full

Is there a way we could use the existing `EvictionReason` enum to map the above values to values in the enum (i.e. use the specific values in `EvictionReason` )?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/MergeRecord.java
```diff
@@ -171,6 +174,7 @@ public class MergeRecord extends AbstractSessionFactoryProcessor {
     public static final String MERGE_COUNT_ATTRIBUTE = "merge.count";
     public static final String MERGE_BIN_AGE_ATTRIBUTE = "merge.bin.age";
     public static final String MERGE_UUID_ATTRIBUTE = "merge.uuid";
+    public static final String REASON_FOR_MERGING = "merge.reason";
```
exceptionfactory, line 177: Recommend adjusting the naming to provide some differentiation from `MergeContent`, which has a different set of defined values.
```suggestion
    public static final String MERGE_COMPLETION_REASON = "merge.completion.reason";
```

exceptionfactory (approved):
Thanks for making the adjustments @dan-s1, the latest version looks good! +1 merging

PR #9365, opened 2024-10-09
NIFI-13856: Capture build scans on ge.apache.org to benefit from deep build insights

clayburn:
# Summary

[NIFI-13856](https://issues.apache.org/jira/browse/NIFI-13856)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -21,7 +21,7 @@ env:
   MAVEN_COMMAND: ./mvnw
   MAVEN_COMMAND_WINDOWS: ./mvnw.cmd
   DEFAULT_MAVEN_OPTS: >-
-    -Xmx4g
+    -Xmx5g
     -XX:ReservedCodeCacheSize=1g
     -XX:+UseG1GC
     -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN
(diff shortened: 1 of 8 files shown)


exceptionfactory (commented):
Thanks for putting together the changes to enable integration with the Apache instance of [Develocity](https://ge.apache.org) @clayburn! This looks like it should provide some helpful insights into the build process.

exceptionfactory (changes requested):
@clayburn Unfortunately a number of the builds failed with heap memory errors:

```
Error: Exception in thread "JNA Cleaner" java.lang.OutOfMemoryError: Java heap space
Error:  Java heap space -> [Help 1]
```

Are there minimum heap requirements for running the Develocity extension?

exceptionfactory:
In order to focus on the heap usage concerns, it would be helpful to scope the changes to just the `ci-workflow` for standard builds. Can that be handled through an explicit environment variable added to the workflow?

clayburn:
@exceptionfactory - unfortunately there is no good way to disable the extension via an env var. Being in the extensions.xml, it will always get picked up. I suppose we could add a step that deletes that file, but it would be an awkward way to do it.

For the moment, I tried adding max heap to each job to see if that gives some breathing room here. We can discuss whether that is the right change or not.

exceptionfactory (changes requested):
Thanks again for working on this integration with Develocity @clayburn.

Given the failures after increasing the maximum heap size, it looks like some closer evaluation of the memory impact is necessary. Perhaps another incremental increase is worth a try, but as this appears to be reproducible on a local machine when setting a maximum heap size, it may only be worth revisiting after confirming a successful local build.

I'll leave this open for a few days for continued discussion. If no immediate solution is apparent, we can collaborate offline and circle back on a new PR after considering the options.

clayburn:
@exceptionfactory - I still need to do some experimenting, but don't see me achieving that this week given other obligations. I'll close this for now and reopen once I have more conclusive data.

PR #9370, opened 2024-10-09
[NIFI-13859] use tertiary color for spinners

scottyaslan:
Screenshot 2024-10-09 at 5 42 27 PM


rfellows (approved):
Thanks @scottyaslan. This is a much better color for the spinners.

PR #9374, opened 2024-10-10
[NIFI-13861] - Update to Nx 19.8.4 and Angular 18.2.8

rfellows:
[NIFI-13861](https://issues.apache.org/jira/browse/NIFI-13861)


mcgilman (approved):
Thanks @rfellows!

PR #9376, opened 2024-10-10
NIFI-13855: Ensuring that tooltips get closed when the host component…

mcgilman:
… is clicked.


rfellows (approved):
Nice fix. This aligns with how normal html title's work. They close if you click while still over the target element.
Thanks @mcgilman

PR #9379, opened 2024-10-11
NIFI-13744 Correct Excel Reader Cell Type Inferencing

exceptionfactory:
# Summary

[NIFI-13744](https://issues.apache.org/jira/browse/NIFI-13744) Corrects Excel Reader type inferencing with Record Data Type detection based on spreadsheet cell properties.

The `CellType` provides the general shape of the value, but requires additional evaluation of `NUMERIC` and `STRING` types to determine whether the field could representation a timestamp, based on Cell Style.

The new `CellFieldTypeReader` replaces existing use of the `SchemaInferenceUtil`, which is limited to `String` input values, requiring value formatting. The standard implementation avoids cell value formatting in most cases, instead relying type detection. The POI library can return timestamp values in several forms, including `java.util.Date` and `java.time.LocalDateTime`, but it does not differentiate between values that contain only dates or only times. For these scenarios, using a defined schema for the Excel Reader provides better resolution.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
@@ -37,7 +37,6 @@ import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
@@ -55,11 +55,11 @@ public class ExcelHeaderSchemaStrategy implements SchemaAccessStrategy {
     private final CellFieldTypeReader cellFieldTypeReader;
     private final DataFormatter dataFormatter;
 
-    public ExcelHeaderSchemaStrategy(PropertyContext context, ComponentLog logger, TimeValueInference timeValueInference, Locale locale) {
+    public ExcelHeaderSchemaStrategy(PropertyContext context, ComponentLog logger, TimeValueInference timeValueInference) {
         this.context = context;
         this.logger = logger;
         this.cellFieldTypeReader = new StandardCellFieldTypeReader(timeValueInference);
-        this.dataFormatter = locale == null ? new DataFormatter() : new DataFormatter(locale);
+        this.dataFormatter = new DataFormatter();
     }
 
     @Override
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
@@ -26,15 +26,12 @@ import org.apache.nifi.schema.access.SchemaNotFoundException;
 import org.apache.nifi.schema.inference.FieldTypeInference;
 import org.apache.nifi.schema.inference.TimeValueInference;
 import org.apache.nifi.serialization.SimpleRecordSchema;
-import org.apache.nifi.serialization.record.DataType;
 import org.apache.nifi.serialization.record.RecordField;
 import org.apache.nifi.serialization.record.RecordSchema;
-import org.apache.nifi.util.SchemaInferenceUtil;
 import org.apache.poi.ss.usermodel.Cell;
 import org.apache.poi.ss.usermodel.DataFormatter;
 import org.apache.poi.ss.usermodel.Row;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.EnumSet;
@@ -55,18 +52,18 @@ public class ExcelHeaderSchemaStrategy implements SchemaAccessStrategy {
 
     private final PropertyContext context;
     private final ComponentLog logger;
-    private final TimeValueInference timeValueInference;
+    private final CellFieldTypeReader cellFieldTypeReader;
     private final DataFormatter dataFormatter;
 
     public ExcelHeaderSchemaStrategy(PropertyContext context, ComponentLog logger, TimeValueInference timeValueInference, Locale locale) {
         this.context = context;
         this.logger = logger;
-        this.timeValueInference = timeValueInference;
+        this.cellFieldTypeReader = new StandardCellFieldTypeReader(timeValueInference);
         this.dataFormatter = locale == null ? new DataFormatter() : new DataFormatter(locale);
     }
 
     @Override
-    public RecordSchema getSchema(Map<String, String> variables, InputStream contentStream, RecordSchema readSchema) throws SchemaNotFoundException, IOException {
+    public RecordSchema getSchema(Map<String, String> variables, InputStream contentStream, RecordSchema readSchema) throws SchemaNotFoundException {
         if (this.context == null) {
             throw new SchemaNotFoundException("Schema Access Strategy intended only for validation purposes and cannot obtain schema");
         }
@@ -140,10 +137,7 @@ public class ExcelHeaderSchemaStrategy implements SchemaAccessStrategy {
                     .forEach(index -> {
                         final Cell cell = row.getCell(index);
                         final String fieldName = fieldNames.get(index);
-                        final FieldTypeInference typeInference = typeMap.computeIfAbsent(fieldName, key -> new FieldTypeInference());
-                        final String formattedCellValue = dataFormatter.formatCellValue(cell);
-                        final DataType dataType = SchemaInferenceUtil.getDataType(formattedCellValue, timeValueInference);
-                        typeInference.addPossibleDataType(dataType);
+                        cellFieldTypeReader.inferCellFieldType(cell, fieldName, typeMap);
                     });
         }
     }
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
@@ -26,6 +26,7 @@ import org.apache.nifi.serialization.record.RecordField;
 import org.apache.nifi.serialization.record.RecordSchema;
 import org.apache.nifi.serialization.record.util.DataTypeUtils;
 import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.CellType;
 import org.apache.poi.ss.usermodel.DateUtil;
 import org.apache.poi.ss.usermodel.Row;
 
@@ -133,15 +134,32 @@ public class ExcelRecordReader implements RecordReader {
         return currentRowValues;
     }
 
-    private static Object getCellValue(Cell cell) {
-        if (cell != null) {
-            return switch (cell.getCellType()) {
-                case _NONE, BLANK, ERROR, FORMULA, STRING -> cell.getStringCellValue();
+    private static Object getCellValue(final Cell cell) {
+        final Object cellValue;
+
+        if (cell == null) {
+            cellValue = null;
+        } else {
+            final CellType cellType = cell.getCellType();
+            cellValue = switch (cellType) {
+                case _NONE, BLANK, ERROR, STRING -> cell.getStringCellValue();
                 case NUMERIC -> DateUtil.isCellDateFormatted(cell) ? cell.getDateCellValue() : cell.getNumericCellValue();
                 case BOOLEAN -> cell.getBooleanCellValue();
+                case FORMULA -> getFormulaCellValue(cell);
             };
         }
-        return null;
+
+        return cellValue;
+    }
+
+    private static Object getFormulaCellValue(final Cell cell) {
+        final CellType formulaResultType = cell.getCachedFormulaResultType();
+        return switch (formulaResultType) {
+            case BOOLEAN -> cell.getBooleanCellValue();
+            case STRING -> cell.getStringCellValue();
+            case NUMERIC -> DateUtil.isCellDateFormatted(cell) ? cell.getDateCellValue() : cell.getNumericCellValue();
+            default -> null;
+        };
     }
 
     private Object convert(final Object value, final DataType dataType, final String fieldName) {
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
@@ -156,7 +156,7 @@ public class ExcelRecordReader implements RecordReader {
         final CellType formulaResultType = cell.getCachedFormulaResultType();
         return switch (formulaResultType) {
             case BOOLEAN -> cell.getBooleanCellValue();
-            case STRING -> cell.getStringCellValue();
+            case STRING, ERROR -> cell.getStringCellValue();
             case NUMERIC -> DateUtil.isCellDateFormatted(cell) ? cell.getDateCellValue() : cell.getNumericCellValue();
             default -> null;
         };
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.excel;
+
+import org.apache.nifi.schema.inference.FieldTypeInference;
+import org.apache.nifi.schema.inference.TimeValueInference;
+import org.apache.nifi.serialization.record.DataType;
+import org.apache.nifi.serialization.record.RecordFieldType;
+import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.CellType;
+import org.apache.poi.ss.usermodel.DateUtil;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Standard implementation of Cell Field Type Reader uses Cell Type and Cell Value information with inference based on Timestamp formats
+ */
+class StandardCellFieldTypeReader implements CellFieldTypeReader {
+    private final TimeValueInference timeValueInference;
+
+    StandardCellFieldTypeReader(final TimeValueInference timeValueInference) {
+        this.timeValueInference = Objects.requireNonNull(timeValueInference, "Time Value Inference required");
+    }
+
+    /**
+     * Infer Cell Field Type and append possible Data Types to mapped Field Type Inference information
+     *
+     * @param cell Spreadsheet Cell required
+     * @param fieldName Cell field name for tracking in Field Types
+     * @param fieldTypes Map of Field Name to Field Type Inference information
+     */
+    @Override
+    public void inferCellFieldType(final Cell cell, final String fieldName, final Map<String, FieldTypeInference> fieldTypes) {
+        Objects.requireNonNull(cell, "Cell required");
+        Objects.requireNonNull(fieldName, "Field Name required");
+        Objects.requireNonNull(fieldTypes, "Field Types required");
+
+        final FieldTypeInference fieldTypeInference = fieldTypes.computeIfAbsent(fieldName, key -> new FieldTypeInference());
+        final DataType dataType = getCellDataType(cell);
+        fieldTypeInference.addPossibleDataType(dataType);
+    }
+
+    /**
+     * Get Record Data Type from Spreadsheet Cell Type and additional resolution of NUMERIC and STRING types
+     *
+     * @param cell Spreadsheet Cell required
+     * @return Record Data Type
+     */
+    @Override
+    public DataType getCellDataType(final Cell cell) {
+        Objects.requireNonNull(cell, "Cell required");
+
+        final CellType cellType = cell.getCellType();
+
+        final DataType dataType;
+
+        if (CellType.NUMERIC == cellType) {
+            // Date Formatting check limited to NUMERIC Cell Types
+            if (DateUtil.isCellDateFormatted(cell)) {
+                // Default to TIMESTAMP for date formatted values using standard Date Time Pattern
+                dataType = RecordFieldType.TIMESTAMP.getDataType();
+            } else {
+                final double numericCellValue = cell.getNumericCellValue();
+                final long roundedCellValue = (long) numericCellValue;
+                if (roundedCellValue == numericCellValue) {
+                    dataType = RecordFieldType.LONG.getDataType();
+                } else {
+                    // Default to DOUBLE for NUMERIC values following cell.getNumericCellValue()
+                    dataType = RecordFieldType.DOUBLE.getDataType();
+                }
+            }
+        } else if (CellType.BOOLEAN == cellType) {
+            dataType = RecordFieldType.BOOLEAN.getDataType();
+        } else if (CellType.STRING == cellType) {
+            final String cellValue = cell.getRichStringCellValue().getString();
+            // Attempt Time Value inference for STRING cell values
+            final Optional<DataType> timeDataType = timeValueInference.getDataType(cellValue);
+            dataType = timeDataType.orElse(RecordFieldType.STRING.getDataType());
+        } else if (CellType.FORMULA == cellType) {
+            dataType = RecordFieldType.STRING.getDataType();
+        } else {
+            // Default to null for known and unknown Cell Types: BLANK, ERROR
+            dataType = null;
+        }
+
+        return dataType;
+    }
+}
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
@@ -79,8 +79,7 @@ class StandardCellFieldTypeReader implements CellFieldTypeReader {
         if (CellType.NUMERIC == cellType) {
             // Date Formatting check limited to NUMERIC Cell Types
             if (DateUtil.isCellDateFormatted(cell)) {
-                // Default to TIMESTAMP for date formatted values using standard Date Time Pattern
-                dataType = RecordFieldType.TIMESTAMP.getDataType();
+                dataType = getDateTimeDataType(cell);
             } else {
                 final double numericCellValue = cell.getNumericCellValue();
                 final long roundedCellValue = (long) numericCellValue;
@@ -103,6 +102,24 @@ class StandardCellFieldTypeReader implements CellFieldTypeReader {
         } else {
             // Default to null for known and unknown Cell Types
             dataType = null;
+        }
+
+        return dataType;
+    }
+
+    private DataType getDateTimeDataType(final Cell cell) {
+        final DataType dataType;
+
+        final double numericCellValue = cell.getNumericCellValue();
+        final long roundedCellValue = (long) numericCellValue;
+        if (roundedCellValue == numericCellValue) {
+            // Numbers without decimal fractions indicate Dates without Times
+            dataType = RecordFieldType.DATE.getDataType();
+        } else if (numericCellValue < 1) {
+            // Decimal fractions indicate Times without Dates
+            dataType = RecordFieldType.TIME.getDataType();
+        } else {
+            dataType = RecordFieldType.TIMESTAMP.getDataType();
         }
 
         return dataType;
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
@@ -78,13 +78,11 @@ class StandardCellFieldTypeReader implements CellFieldTypeReader {
 
         if (CellType.NUMERIC == cellType) {
             // Date Formatting check limited to NUMERIC Cell Types
+            final double numericCellValue = cell.getNumericCellValue();
             if (DateUtil.isCellDateFormatted(cell)) {
-                final double numericCellValue = cell.getNumericCellValue();
                 dataType = getDateTimeDataType(numericCellValue);
             } else {
-                final double numericCellValue = cell.getNumericCellValue();
-                final long roundedCellValue = (long) numericCellValue;
-                if (roundedCellValue == numericCellValue) {
+                if (isWholeNumber(numericCellValue)) {
                     dataType = RecordFieldType.LONG.getDataType();
                 } else {
                     // Default to DOUBLE for NUMERIC values following cell.getNumericCellValue()
@@ -111,8 +109,7 @@ class StandardCellFieldTypeReader implements CellFieldTypeReader {
     private DataType getDateTimeDataType(final double numericCellValue) {
         final DataType dataType;
 
-        final long roundedCellValue = (long) numericCellValue;
-        if (roundedCellValue == numericCellValue) {
+        if (isWholeNumber(numericCellValue)) {
             // Numbers without decimal fractions indicate Dates without Times
             dataType = RecordFieldType.DATE.getDataType();
         } else if (numericCellValue < 1) {
@@ -148,5 +145,10 @@ class StandardCellFieldTypeReader implements CellFieldTypeReader {
         }
 
         return dataType;
+    }
+
+    private boolean isWholeNumber(final double numericCellValue) {
+        final long roundedCellValue = (long) numericCellValue;
+        return roundedCellValue == numericCellValue;
     }
 }
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
@@ -88,7 +88,7 @@ class StandardCellFieldTypeReader implements CellFieldTypeReader {
         } else if (CellType.BOOLEAN == cellType) {
             dataType = RecordFieldType.BOOLEAN.getDataType();
         } else if (CellType.STRING == cellType) {
-            final String cellValue = cell.getRichStringCellValue().getString();
+            final String cellValue = cell.getStringCellValue();
             // Attempt Time Value inference for STRING cell values
             final Optional<DataType> timeDataType = timeValueInference.getDataType(cellValue);
             dataType = timeDataType.orElse(RecordFieldType.STRING.getDataType());
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
@@ -34,6 +34,11 @@ import java.util.Optional;
 class StandardCellFieldTypeReader implements CellFieldTypeReader {
     private final TimeValueInference timeValueInference;
 
+    /**
+     * Standard Cell Field Type Reader constructor with Time Value Inference for handling STRING Cell Types that may contain values with Timestamps
+     *
+     * @param timeValueInference Time Value Inference required for STRING Cell Type evaluation
+     */
     StandardCellFieldTypeReader(final TimeValueInference timeValueInference) {
         this.timeValueInference = Objects.requireNonNull(timeValueInference, "Time Value Inference required");
     }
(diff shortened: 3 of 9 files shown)


nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,104 @@
...
+                    dataType = RecordFieldType.LONG.getDataType();
+                } else {
+                    // Default to DOUBLE for NUMERIC values following cell.getNumericCellValue()
+                    dataType = RecordFieldType.DOUBLE.getDataType();
+                }
+            }
+        } else if (CellType.BOOLEAN == cellType) {
+            dataType = RecordFieldType.BOOLEAN.getDataType();
+        } else if (CellType.STRING == cellType) {
+            final String cellValue = cell.getRichStringCellValue().getString();
```
dan-s1, line 91: Are you sure you want to use `getRichStringCellValue` ? From the following [post](https://stackoverflow.com/questions/16170275/what-is-the-difference-between-getrichstringcellvalue-and-getstringcellvalue) it seems its better to use `getStringCellValue`?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,104 @@
...
+            // Attempt Time Value inference for STRING cell values
+            final Optional<DataType> timeDataType = timeValueInference.getDataType(cellValue);
+            dataType = timeDataType.orElse(RecordFieldType.STRING.getDataType());
+        } else if (CellType.FORMULA == cellType) {
+            dataType = RecordFieldType.STRING.getDataType();
+        } else {
```
dan-s1, line 97: I am not sure this is correct because the formula has to be evaluated and can evaluated to a number or a boolean. The text of the formula is a string but I assume most end user's are not interested in the formula text but rather what the formula evaluates to.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,104 @@
...
+                dataType = RecordFieldType.TIMESTAMP.getDataType();
+            } else {
+                final double numericCellValue = cell.getNumericCellValue();
+                final long roundedCellValue = (long) numericCellValue;
+                if (roundedCellValue == numericCellValue) {
+                    dataType = RecordFieldType.LONG.getDataType();
+                } else {
+                    // Default to DOUBLE for NUMERIC values following cell.getNumericCellValue()
+                    dataType = RecordFieldType.DOUBLE.getDataType();
+                }
```
dan-s1, line 86: Do we want to determine whether the data types are `INTEGER` or `FLOAT` or just keep at `LONG`/`DOUBLE`?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
exceptionfactory: Good recommendation, thanks for the reference. I started with `getRichStringCellValue()` based on the behavior of the POI `DataFormatter`, but for Cell Type inference purposes, just using `getStringCellValue()` makes more sense.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
exceptionfactory: That's a good point about the evaluated value of the cell based on the formula. The current `ExcelRecordReader.getCellValue()` method calls `Cell.getStringCellValue()` for the `FORMULA` Cell Type, so this behavior aligns with that approach. This could be changed and improved, but that might make sense as a separate change. What do you think?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
exceptionfactory: I considered that as an option, however, inferring `LONG` or `DOUBLE` provides greater type safety. For example, if the maximum numeric value for the evaluated rows fits into an `int`, but subsequent rows exceed the maximum value of an `int`, then processing would fail.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
dan-s1: In my PR I was trying to correct that logic as I had not realized `Cell.getStringCellValue()` get the text of the formula. That is why I had added to ExcelUtils

 ```
private static final DataFormatter DATA_FORMATTER = new DataFormatter(Locale.getDefault());

    static {
        DATA_FORMATTER.setUseCachedValuesForFormulaCells(true);
    }
```

as DataFormatter can obtain the formula values after calling `setUseCachedValuesForFormulaCells(true)`

Comment (exceptionfactory) [reply nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java]: I pushed another update to improve the handling of Formula Cells, based on the Cached Formula Result Type. This follows the same general behavior of the `DataFormatter` when it is configured to use cached values from formula cells.

File: nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java [hunk 4]
```diff
@@ -0,0 +1,129 @@
...
+                // Default to TIMESTAMP for date formatted values using standard Date Time Pattern
+                dataType = RecordFieldType.TIMESTAMP.getDataType();
+            } else {
+                // Default to DOUBLE for NUMERIC values following cell.getNumericCellValue()
+                dataType = RecordFieldType.DOUBLE.getDataType();
+            }
+        } else {
+            // Default to null for known and unknown Formula Result Cell Types: BLANK, ERROR
+            dataType = null;
+        }
```
dan-s1, line 125: I didn't realize null type is an option. Doesn't everything need to have an associated type?
Furthermore it is possible to to have a string value of the error which end users may or may not be interested in.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,129 @@
...
+                    // Default to DOUBLE for NUMERIC values following cell.getNumericCellValue()
+                    dataType = RecordFieldType.DOUBLE.getDataType();
+                }
+            }
+        } else if (CellType.BOOLEAN == cellType) {
+            dataType = RecordFieldType.BOOLEAN.getDataType();
+        } else if (CellType.STRING == cellType) {
+            final String cellValue = cell.getStringCellValue();
+            // Attempt Time Value inference for STRING cell values
+            final Optional<DataType> timeDataType = timeValueInference.getDataType(cellValue);
```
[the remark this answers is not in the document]
exceptionfactory: That's a good question. It depends how flexible we should be for timestamp handling. In other words, it is possible for a `STRING` cell to contain something that looks like a timestamp. This would effectively remove the need for any time or date format properties in the `ExcelReader`, but it would require times and dates to be stored as `NUMERIC` types. I'm inclined to leave it in, given the existence of those format properties, but it might be helpful to keep it for now to enable wider time and date detection from some spreadsheets.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
exceptionfactory: The `null` return is valid, and the Field Type Inference class simply ignores it. In that, case, the inferred type is effectively "unknown", however, the standard behavior of record type inferencing falls back to `STRING`. Since `null` is valid as a return type, this effectively indicates that the type is unknown or unable to be determined by the inferencing process.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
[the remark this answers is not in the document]
dan-s1: I just stored a date in an Excel spreadsheet as a text e.g. 5/21/2024 so I see what you mean.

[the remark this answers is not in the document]
dan-s1: I think though this piece should be documented that any time or date format properties  specified by the user (stored in `TimeValueInference`) will **_only_** be used to parse those cells whose cell type is string and not whose already is date.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelSchemaInference.java
```diff
@@ -50,69 +45,54 @@ public class TestExcelSchemaInference {
...
         final Map<PropertyDescriptor, String> properties = new HashMap<>();
         new ExcelReader().getSupportedPropertyDescriptors().forEach(prop -> properties.put(prop, prop.getDefaultValue()));
         final PropertyContext context = new MockConfigurationContext(properties, null, null);
 
         final RecordSchema schema;
-        try (final InputStream inputStream = getResourceStream("/excel/simpleDataFormatting.xlsx")) {
+        try (final InputStream inputStream = getResourceStream()) {
             final InferSchemaAccessStrategy<?> accessStrategy = new InferSchemaAccessStrategy<>(
                     (variables, content) -> new ExcelRecordSource(content, context, variables, logger),
                     new ExcelSchemaInference(timestampInference), Mockito.mock(ComponentLog.class));
```
dan-s1, line 66: No need to call `Mockito.mock(ComponentLog.class)` as we already have an instance of ComponentLog annotated with the `@Mock` annotation.
```suggestion
                    new ExcelSchemaInference(timestampInference), logger);
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,129 @@
...
+            dataType = timeDataType.orElse(RecordFieldType.STRING.getDataType());
+        } else if (CellType.FORMULA == cellType) {
+            dataType = getFormulaResultDataType(cell);
+        } else {
+            // Default to null for known and unknown Cell Types: BLANK, ERROR
```
dan-s1, line 98: Per `getCachedFormulaResultType` [javadocs](https://poi.apache.org/apidocs/5.0/org/apache/poi/ss/usermodel/Cell.html#getCachedFormulaResultType--) there is no BLANK cell type for formulas
```suggestion
            // Default to null for known and unknown Cell Type:  ERROR
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelHeaderSchemaStrategy.java
```diff
@@ -55,18 +52,18 @@ public class ExcelHeaderSchemaStrategy implements SchemaAccessStrategy {
...
-    private final TimeValueInference timeValueInference;
+    private final CellFieldTypeReader cellFieldTypeReader;
     private final DataFormatter dataFormatter;
 
     public ExcelHeaderSchemaStrategy(PropertyContext context, ComponentLog logger, TimeValueInference timeValueInference, Locale locale) {
         this.context = context;
         this.logger = logger;
-        this.timeValueInference = timeValueInference;
+        this.cellFieldTypeReader = new StandardCellFieldTypeReader(timeValueInference);
         this.dataFormatter = locale == null ? new DataFormatter() : new DataFormatter(locale);
```
dan-s1, line 62: I do not think the Locale argument is needed any longer as we can instantiate DataFormatter as follows:
```suggestion
        this.dataFormatter = new DataFormatter(Locale.getDefault());
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
```diff
@@ -133,15 +134,32 @@ private Map<String, Object> getCurrentRowValues(Row currentRow, boolean coerceTy
...
+        return cellValue;
+    }
+
+    private static Object getFormulaCellValue(final Cell cell) {
+        final CellType formulaResultType = cell.getCachedFormulaResultType();
+        return switch (formulaResultType) {
+            case BOOLEAN -> cell.getBooleanCellValue();
+            case STRING -> cell.getStringCellValue();
+            case NUMERIC -> DateUtil.isCellDateFormatted(cell) ? cell.getDateCellValue() : cell.getNumericCellValue();
+            default -> null;
```
dan-s1, line 161: This will return null for those formulas which have errors such as `Err:511` and `#NAME?`. End users may want to see these in order to correct their Excel worksheet.
```suggestion
        return switch (formulaResultType) {
            case BOOLEAN -> cell.getBooleanCellValue();
            case ERROR, STRING -> cell.getStringCellValue();
            case NUMERIC -> DateUtil.isCellDateFormatted(cell) ? cell.getDateCellValue() : cell.getNumericCellValue();
```

exceptionfactory: Good catch, that is worth retaining.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * Infer Cell Field Type and append possible Data Types to mapped Field Type Inference information
+     *
+     * @param cell Spreadsheet Cell required
+     * @param fieldName Cell field name for tracking in Field Types
+     * @param fieldTypes Map of Field Name to Field Type Inference information
+     */
+    @Override
+    public void inferCellFieldType(final Cell cell, final String fieldName, final Map<String, FieldTypeInference> fieldTypes) {
+        Objects.requireNonNull(cell, "Cell required");
```
dan-s1, line 50: I do not think there should be this check as there can be cells which have no data in it and `Cell `will be null. I uncovered this running the TestEsxcelReader which I had created for my PR.
```suggestion
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,129 @@
...
+        Objects.requireNonNull(fieldName, "Field Name required");
+        Objects.requireNonNull(fieldTypes, "Field Types required");
+
+        final FieldTypeInference fieldTypeInference = fieldTypes.computeIfAbsent(fieldName, key -> new FieldTypeInference());
+        final DataType dataType = getCellDataType(cell);
+        fieldTypeInference.addPossibleDataType(dataType);
```
dan-s1, line 56: I do not think there should be this check as there can be cells which have no data in it and `Cell `will be null. I uncovered this running the `TestExcelReader` which I had created for my PR.
```suggestion
        Objects.requireNonNull(fieldName, "Field Name required");
        Objects.requireNonNull(fieldTypes, "Field Types required");

        final FieldTypeInference fieldTypeInference = fieldTypes.computeIfAbsent(fieldName, key -> new FieldTypeInference());
        final DataType dataType = cell != null ? getCellDataType(cell) : null;
        fieldTypeInference.addPossibleDataType(dataType);
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,129 @@
...
+
+        final CellType cellType = cell.getCellType();
+
+        final DataType dataType;
+
+        if (CellType.NUMERIC == cellType) {
+            // Date Formatting check limited to NUMERIC Cell Types
+            if (DateUtil.isCellDateFormatted(cell)) {
+                // Default to TIMESTAMP for date formatted values using standard Date Time Pattern
+                dataType = RecordFieldType.TIMESTAMP.getDataType();
```
dan-s1, line 77: I do not think defaulting to `TIMESTAMP` is right for every instance. In the `TestExcelReader` I have in my PR I have three columns

1. ` Date_Custom` which has a date value 07/25/2024 (displayed as `07/25/24`)
2.  `Time_Custom` which has a time value of 04:00:00 PM (displayed as `16:00`)
3.  `Timestamp` which has a timestamp 07/26/2024 16:00:00

When running through your code they are parsed and look like the following when calling `toString` on the `java.sql.Timestamp`

1. 2024-07-05 00:00:00.0
2. 1899-12-31 00:00:00.0
3. 2024-07-26 16:00:00.0

For dates and timestamp 1 and 3 are okay but for those fields which only represent time, I do not think users are interested in the default date Excel has for time 1899-12-31. I think there has to a more fine grained way of determining whether the date we have is representing a date or only time.

exceptionfactory: Thanks for highlighting these details @dan-s1. This is an important issue of the way that Excel stores dates, and the way the POI library handles dates.

For all date and time values, the POI library always returns a `java.sql.Timestamp` or ` java.time.LocalDateTime`. The library interprets date values with zero for hours, minutes, and seconds. For time values, the POI library uses a default starting date, that is the reason for these inferred values.

When using a defined schema, the time and date values will be handled as expected. When using an inferred schema, the Schema Inference logic is unable to differentiate between Timestamp, Date, and Time, so the inferred Schema handles everything as a Timestamp. It might be possible to evaluate some Timestamp values and infer that they contain only a Date or only Time, but such inference would be prone to error. For this scenario, the inference process is necessarily general, and is unable to narrow down values to either or Date or Time. It might be possible to try something based on the date format from Excel, if specified, but this is also prone to error based on the variety of values, which are different than Java date formatting.

dan-s1: I thought though the user entered properties can/should assist the inference logic in help determining whether a field is a date, time or timestamp.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRecordReader.java
dan-s1: Seems like the `default -> null;` needs to still be there otherwise at least in Intellij there is a warning that the `switch` statement does not cover all cases.

dan-s1:
Also I noticed even with an explicit schema if we do not get the formatted form there is an exception. The case in point I see is when the schema defines a boolean and the column returns a boolean from a formula. Without the formatted form 1.0 is returned which will not be interpreted as a boolean in Java With the formatted form TRUE is returned which can interpreted as a boolean in Java.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelSchemaInference.java
```diff
@@ -50,69 +45,54 @@ public class TestExcelSchemaInference {
     private static final String EXPECTED_SECOND_FIELD_NAME = ExcelUtils.FIELD_NAME_PREFIX + "1";
     private static final String EXPECTED_THIRD_FIELD_NAME = ExcelUtils.FIELD_NAME_PREFIX + "2";
     private static final String EXPECTED_FOURTH_FIELD_NAME = ExcelUtils.FIELD_NAME_PREFIX + "3";
+
+    private static final String SIMPLE_FORMATTING_PATH = "/excel/simpleDataFormatting.xlsx";
+
     private final TimeValueInference timestampInference = new TimeValueInference("MM/dd/yyyy", "HH:mm:ss.SSS", "yyyy/MM/dd/ HH:mm");
```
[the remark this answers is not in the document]
exceptionfactory: Thanks, I changed the existing `TimeValueInference` to a mock, but added a new test method with a `TimeValueInference` that handles a `STRING` field as a timestamp.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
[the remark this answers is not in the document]
exceptionfactory: Thanks, I added some comments around the constructor.

exceptionfactory:
> Also I noticed even with an explicit schema if we do not get the formatted form, an exception can be thrown when reading with `ExcelRecordReader`. The case in point I see is when the schema defines a boolean and the column returns a boolean from a formula. Without the formatted form, `1.0` is returned which will not be interpreted as a boolean in Java. With the formatted form, `TRUE` is returned which can interpreted as a boolean in Java.

This sounds like a case where the cached formula result type is not returning `BOOLEAN` and is instead returning `NUMERIC` or `STRING`.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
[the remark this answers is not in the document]
dan-s1: I actually meant end user documentation as this is a change in functionality from what has been up until now.

exceptionfactory:
Thanks for the thorough review @dan-s1.

I pushed an update to address the comments. I updated comments in several places, adding handling for `null` Cells, and added a new unit test method to exercise the Timestamp inference behavior for `STRING` Cell Types.

I marked several threads as resolved, but if there are any where you think additional feedback is needed, feel free to unresolve and continue the conversation. Otherwise, hopefully we can wrap this up soon.

dan-s1:
> > Also I noticed even with an explicit schema if we do not get the formatted form, an exception can be thrown when reading with `ExcelRecordReader`. The case in point I see is when the schema defines a boolean and the column returns a boolean from a formula. Without the formatted form, `1.0` is returned which will not be interpreted as a boolean in Java. With the formatted form, `TRUE` is returned which can interpreted as a boolean in Java.
> 
> This sounds like a case where the cached formula result type is not returning `BOOLEAN` and is instead returning `NUMERIC` or `STRING`.

Is there anything to do for this case or leave it as is?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,135 @@
...
+        final DataType dataType = getCellDataType(cell);
+        fieldTypeInference.addPossibleDataType(dataType);
+    }
+
+    /**
+     * Get Record Data Type from Spreadsheet Cell Type and additional resolution of NUMERIC and STRING types
+     *
+     * @param cell Spreadsheet Cell can be null
+     * @return Record Data Type or null
+     */
```
dan-s1, line 68: Is it necessary to duplicate the parameter javadoc information from the interface? I see the descriptions are different as they add information which I can understand why they are there (unless it would make sense to pull this information to the interface also).

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,135 @@
...
+            dataType = RecordFieldType.BOOLEAN.getDataType();
+        } else if (CellType.STRING == cellType) {
+            final String cellValue = cell.getStringCellValue();
+            // Attempt Time Value inference for STRING cell values
+            final Optional<DataType> timeDataType = timeValueInference.getDataType(cellValue);
+            dataType = timeDataType.orElse(RecordFieldType.STRING.getDataType());
+        } else if (CellType.FORMULA == cellType) {
+            dataType = getFormulaResultDataType(cell);
+        } else {
+            // Default to null for known and unknown Cell Types such as ERROR
```
dan-s1, line 104:
```suggestion
            // Default to null for known and unknown Cell Types such as BLANK, ERROR
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,135 @@
...
+            // Date Formatting check limited to NUMERIC Cell Types without Conditional Formatting Evaluator
+            if (DateUtil.isCellDateFormatted(cell)) {
+                // Default to TIMESTAMP for date formatted values using standard Date Time Pattern
+                dataType = RecordFieldType.TIMESTAMP.getDataType();
+            } else {
+                // Default to DOUBLE for NUMERIC values following cell.getNumericCellValue()
+                dataType = RecordFieldType.DOUBLE.getDataType();
+            }
+        } else {
+            // Default to null for known and unknown Formula Result Cell Types: BLANK, ERROR
```
dan-s1, line 129:
```suggestion
            // Default to null for known and unknown Formula Result Cell Types: ERROR
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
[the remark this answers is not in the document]
dan-s1: Also it might be worth documenting that when inferring a schema, the schema will be limited to the following specific types BOOLEAN, DOUBLE,  LONG, STRING, TIMESTAMP and perhaps DATE and TIME.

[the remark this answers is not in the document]
exceptionfactory: Reviewing the current `additionalDetails.md` documentation, the description actually seems to reflect the behavior accurately:

> If the "Time Format," "Timestamp Format," or "Date Format" properties are configured, any value that would otherwise be considered a STRING type is first checked against the configured formats to see if it matches any of them. If the value matches the Timestamp Format, the value is considered a Timestamp field. If it matches the Date Format, it is considered a Date field. If it matches the Time Format, it is considered a Time field. In the unlikely event that the value matches more than one of the configured formats, they will be matched in the order: Timestamp, Date, Time. I.e., if a value matched both the Timestamp Format and the Date Format, the type that is inferred will be Timestamp. Because parsing dates and times can be expensive, it is advisable not to configure these formats if dates, times, and timestamps are not expected, or if processing the data as a STRING is acceptable. For use cases when this is important, though, the inference engine is intelligent enough to optimize the parsing by first checking several very cheap conditions. For example, the string's length is examined to see if it is too long or too short to match the pattern. This results in far more efficient processing than would result if attempting to parse each string value as a timestamp.

I'm not sure how valuable it is to go into some additional implementation details at the moment, if future revisions are necessary. However, if you have a suggestion for an additional bullet point, glad to include.

dan-s1:
@exceptionfactory I just found this helpful [tutorial](https://www.myonlinetraininghub.com/excel-date-and-time#:~:text=When%20you%20look%20at%20a,decimal%20portion%20is%20the%20time.) which explains how dates and times are stored in Excel. Based on this information I believe we can provide better inferences when the cell type is numeric and is date formatted. If its a whole number we know its date, if its a fraction we know its a time and if its a whole  number with a fraction then its a timestamp. Hence lines 81-84 in `StandardCellFieldTypeReader` can be expanded to incorporate this logic to identify the most precise type.

exceptionfactory:
> > > Also I noticed even with an explicit schema if we do not get the formatted form, an exception can be thrown when reading with `ExcelRecordReader`. The case in point I see is when the schema defines a boolean and the column returns a boolean from a formula. Without the formatted form, `1.0` is returned which will not be interpreted as a boolean in Java. With the formatted form, `TRUE` is returned which can interpreted as a boolean in Java.
> > 
> > 
> > This sounds like a case where the cached formula result type is not returning `BOOLEAN` and is instead returning `NUMERIC` or `STRING`.
> 
> Is there anything to do for this case or leave it as is?

In this case, it sounds like an explicit schema should be configured with a `STRING` type instead of a `BOOLEAN`. The behavior of converting number `1.0` to a string `TRUE` seems a bit difficult to generalize unless there is some other fact we can use to handle the value as a boolean.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
exceptionfactory: It isn't absolutely necessary, but if there is a comment, it should include the parameters. I kept them for completeness, since the method comment is different.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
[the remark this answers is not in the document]
dan-s1: > any value that would otherwise be considered a STRING type
That part is bothering me as until now that was everything but now any column whose Excel format is date is not included in this. I am not sure users will understand that.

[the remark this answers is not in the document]
exceptionfactory: I think part of the challenge is that Excel provides a narrow set of stored type values, and a slightly wider set of supported rendered type values. When combined with type inferencing, there is a lot of potential for confusion. That being said, I think the above paragraph is still accurate. At least the difference between a number of string should be clear enough, so that distinction remains, and is now more strongly followed.

exceptionfactory:
> @exceptionfactory I just found this helpful [tutorial](https://www.myonlinetraininghub.com/excel-date-and-time#:~:text=When%20you%20look%20at%20a,decimal%20portion%20is%20the%20time.) which explains how dates and times are stored in Excel. Based on this information I believe we can provide better inferences when the cell type is numeric and is date formatted. If its a whole number we know its date, if its a fraction we know its a time and if its a whole number with a fraction then its a timestamp. Hence lines 81-84 in `StandardCellFieldTypeReader` can be expanded to incorporate this logic to identify the most precise type (DATE, TIME or TIMESTAMP).
> 
> e.g. (Which probably would look better incorporated in its own method)
> 
> ```
> if (DateUtil.isCellDateFormatted(cell)) {
>              final double numericCellValue = cell.getNumericCellValue();
>              if (numericCellValue > 0 && numericCellValue < 1;) {
>                   dataType = RecordFieldType.TIME.getDataType();
>              } else {
>                   final long roundedCellValue = (long) numericCellValue;
>                   if (roundedCellValue == numericCellValue) {
>                     dataType = RecordFieldType.DATE.getDataType();
>                   } else {
>                     dataType = RecordFieldType.TIMESTAMP.getDataType();
>                 }
>            }
>      }
> ```

Thanks for the background and example @dan-s1, that is very helpful. I will incorporate that approach and include you as a co-author on the commit.

dan-s1:
@exceptionfactory With that change should we also then improve what is returned by ExcelRecordReader on line 146?
Currently we a a `java.util.Date` with `cell.getDateCellValue()` but I think we should employ similar logic to determine if its a DATE/TIME then call `cell.getLocalDateTimeCellValue()` to return a `java.time.LocalDateTime`.

exceptionfactory:
> @exceptionfactory With that change should we should also then improve what is returned by ExcelRecordReader on line 146. Currently we return a `java.util.Date` with `cell.getDateCellValue()` but I think we should employ similar logic to determine if its a DATE/TIME then call `cell.getLocalDateTimeCellValue()` to return a `java.time.LocalDateTime`.
> 
> e.g.
> 
> ```
> case NUMERIC -> {
>                     if (DateUtil.isCellDateFormatted(cell)) {
>                         final double numericCellValue = cell.getNumericCellValue();
>                         final long roundedCellValue = (long) numericCellValue;
>                         if ((numericCellValue > 0 && numericCellValue < 1) || (roundedCellValue == numericCellValue)) {
>                             yield cell.getLocalDateTimeCellValue();
>                         } else {
>                             yield cell.getDateCellValue();
>                         }
>                     } else {
>                         yield cell.getNumericCellValue();
>                     }
>                 }
> ```

The `java.util.Date` includes both Date and Time, so that is why it is functionally equivalent to the return value of `getLocalDateTimeCellValue()` in this scenario.

exceptionfactory:
@dan-s1 I pushed a change with the additional `Date` and `Time` inference handling:

https://github.com/apache/nifi/pull/9379/commits/bab137b35f3dbdf249397d24bf75ec824226b292

dan-s1:
@exceptionfactory I built the code without the most recent change but I am not sure that will make a difference. Part of the user's issue in this ticket was for his time custom column the final format is a negative number `-2209017600000` which is still occurring with the current changes. I assume this needs to be corrected.

exceptionfactory:
> @exceptionfactory I built the code without the most recent change but I am not sure that will make a difference. Part of the user's issue in this ticket was for his time custom column the final format is a negative number `-2209017600000` which is still occurring with the current changes. I assume this needs to be corrected. As for his issue of epoch strings I do not think that is an issue if the writer is configured with a timestamp pattern. Let me take that back when I configure a timestamp pattern my resulting CSV is
> 
> ```
> Date_Standard,Date_Custom,Time_Standard,Time_Custom,Timestamp
> 2024-07-24 00:00:00,2024-07-25 00:00:00,2024-07-24 16:00:00,1899-12-31 16:00:00,2024-07-26 16:00:00
> ```
> 
> I am not sure though users are going to expect `1899-12-31 16:00:00`
> 
> Is there a way we can only have the time?

Have you configured a schema for the CSV Writer that indicates the Time field?

dan-s1:
I assume I need your latest changes to have the TIME type in order for the CSVWriter to pick that up with the time pattern.

dan-s1:
Now its working as expected after I set on date, time and timestamp patterns on the on the `CSVRecordSetWriter`

```
Date_Standard,Date_Custom,Time_Standard,Time_Custom,Timestamp
2024-07-24,2024-07-25,2024-07-24 16:00:00,16:00:00,2024-07-26 16:00:00
```

dan-s1:
> > Also I noticed even with an explicit schema if we do not get the formatted form, an exception can be thrown when reading with `ExcelRecordReader`. The case in point I see is when the schema defines a boolean and the column returns a boolean from a formula. Without the formatted form, `1.0` is returned which will not be interpreted as a boolean in Java. With the formatted form, `TRUE` is returned which can interpreted as a boolean in Java.
> 
> This sounds like a case where the cached formula result type is not returning `BOOLEAN` and is instead returning `NUMERIC` or `STRING`.

Its actually returning a NUMERIC a double to be precise. From your latest changes the corresponding field in the schema is producing
`{"name":"Boolean_Function","type":["double","null"]}`

dan-s1:
@exceptionfactory Thank you for making all the requested changes. Also thank you for including me as a coauthor Everything looks good now and seems to be producing the correct results. LGTM+1.  I plan to merge as soon as the CI build completes.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
```diff
@@ -0,0 +1,152 @@
...
+        final CellType formulaResultType = cell.getCachedFormulaResultType();
+        if (CellType.BOOLEAN == formulaResultType) {
+            dataType = RecordFieldType.BOOLEAN.getDataType();
+        } else if (CellType.STRING == formulaResultType) {
+            dataType = RecordFieldType.STRING.getDataType();
+        } else if (CellType.NUMERIC == formulaResultType) {
+            // Date Formatting check limited to NUMERIC Cell Types without Conditional Formatting Evaluator
+            if (DateUtil.isCellDateFormatted(cell)) {
+                // Default to TIMESTAMP for date formatted values using standard Date Time Pattern
+                dataType = RecordFieldType.TIMESTAMP.getDataType();
```
dan-s1, line 140: I do not see any methods which can get the cached formula result but do we leave this as `TIMESTAMP` or somehow try to get the result and determine whether its a `DATE` or `TIME` like we did above?

dan-s1: I am not seeing anything directly, its probably best to leave as is.

exceptionfactory: Sounds good, we can revisit in the future based on potential issues if needed.

exceptionfactory:
> @exceptionfactory Thank you for making all the requested changes. Also thank you for including me as a coauthor. Everything looks good now and seems to be producing the correct results. LGTM+1.
> 
> I plan to merge as soon as the CI build completes.

Thanks @dan-s1, great collaborative effort on refining the behavior on the Excel Reader!

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StandardCellFieldTypeReader.java
dan-s1: Looking at the DataFormatter code, specifically lines 513-522

 ```
synchronized(dateFormat) {
                if (dateFormat instanceof ExcelStyleDateFormatter) {
                    ((ExcelStyleDateFormatter)dateFormat).setDateToBeFormatted(cell.getNumericCellValue());
                }

                Date d = cell.getDateCellValue();
                return this.performDateFormatting(d, (Format)dateFormat);
            }
```

it seems it gets the cell's numeric or date value. If that is the case then I assume you can reuse the `getDateTimeDataType` above to determine the precise date type.

exceptionfactory: Thanks for the additional pointer. I ran a test with a formula cell in XLSX, confirmed the numeric cell value was available, and I was able to reuse the same `getDateTimeDataType()` method to narrow the Data Type detection, as already used with numeric types. This also aligns with the behavior of `getFormulaCellValue()` in `ExcelRecordReader`.

dan-s1: Thanks for pointing out that this aligns to `getFormulaCellValue()`

dan-s1 (commented):
@exceptionfactory I know we are really close. I just have two more suggestions to take care of. Thanks!

exceptionfactory:
Thanks @dan-s1, I pushed an update with those recommended changes.

dan-s1:
I plan to merge this as soon as the CI build finishes (for real this time :) ). In regards to to 1.x would you like me to make a separate PR to have these changes merged into the 1.x branch?

exceptionfactory:
> I plan to merge this as soon as the CI build finishes (for real this time :) ). In regards to to 1.x would you like me to make a separate PR to have these changes merged into the 1.x branch?

Thanks! I'll put together a separate PR for the support branch today.

dan-s1:
@exceptionfactory Thank you again for making all the requested changes. Also thank you again for including me as a coauthor. 
LGTM+1.

exceptionfactory:
Thanks @dan-s1! The automated builds for the backported pull request #9397 are also close to completion.

dan-s1:
Yeah I saw. One failed so I restarted it.

PR #9386, opened 2024-10-14
NIFI-12061: add SECRET_NAME parameter to avoid listing all secrets

grishick:
# Summary

[NIFI-12061](https://issues.apache.org/jira/browse/NIFI-12061)
Add `Secret Name` parameter to `AwsSecretsManagerParameterProvider` to allow the provider to work without `ListSecrets` permission. Original behavior where the provider lists all secrets and matches them to `Secret Name Pattern` is preserved by leaving `Secret Name` parameter undefined. The new behavior will skip listing secrets and ignore `Secret Name Pattern` only if `Secret Name` is provided.  I made this change in [Anetac's Nifi fork](https://github.com/Eng-Anetac/nifi) months ago and we are using it in production.

Related PR to main branch: https://github.com/apache/nifi/pull/9483
# Tracking

### Issue Tracking

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-12061`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-12061`

### Pull Request Formatting

- [x] Pull Request based on current revision of the `support/nifi-1.x` branch

# Verification

### Build

  - [x] JDK 1.8
  - [x] JDK 11
  - [x] JDK 17

### Licensing

### Documentation

### DIFF
--- a/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
+++ b/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
@@ -39,6 +39,7 @@ import org.apache.nifi.annotation.documentation.Tags;
 import org.apache.nifi.components.AllowableValue;
 import org.apache.nifi.components.ConfigVerificationResult;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.PropertyValue;
 import org.apache.nifi.controller.ConfigurationContext;
 import org.apache.nifi.logging.ComponentLog;
 import org.apache.nifi.parameter.AbstractParameterProvider;
@@ -71,11 +72,20 @@ import java.util.regex.Pattern;
         "key/value pairs in the secret mapping to Parameters in the group.")
 public class AwsSecretsManagerParameterProvider extends AbstractParameterProvider implements VerifiableParameterProvider {
 
+    public static final PropertyDescriptor SECRET_NAME = new PropertyDescriptor.Builder()
+            .name("secret-name")
+            .displayName("Secret Name")
+            .description("The name of the secret to fetch parameters from. Will be used instead of the Secret Name Pattern property if set." +
+                    "Use this parameter instead of Secret Name Pattern if your AWS Account does not have permission to list secrets.")
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .required(false)
+            .build();
     public static final PropertyDescriptor SECRET_NAME_PATTERN = new PropertyDescriptor.Builder()
             .name("secret-name-pattern")
             .displayName("Secret Name Pattern")
             .description("A Regular Expression matching on Secret Name that identifies Secrets whose parameters should be fetched. " +
-                    "Any secrets whose names do not match this pattern will not be fetched.")
+                    "Any secrets whose names do not match this pattern will not be fetched. Using this parameter requires the ListSecrets permission." +
+                    "This parameter is ignored if the Secret Name parameter is set.")
             .addValidator(StandardValidators.REGULAR_EXPRESSION_VALIDATOR)
             .required(true)
             .defaultValue(".*")
@@ -120,6 +130,7 @@ public class AwsSecretsManagerParameterProvider extends AbstractParameterProvide
     private static final String DEFAULT_USER_AGENT = "NiFi";
     private static final Protocol DEFAULT_PROTOCOL = Protocol.HTTPS;
     private static final List<PropertyDescriptor> PROPERTIES = Collections.unmodifiableList(Arrays.asList(
+            SECRET_NAME,
             SECRET_NAME_PATTERN,
             REGION,
             AWS_CREDENTIALS_PROVIDER_SERVICE,
@@ -139,19 +150,33 @@ public class AwsSecretsManagerParameterProvider extends AbstractParameterProvide
         AWSSecretsManager secretsManager = this.configureClient(context);
 
         final List<ParameterGroup> groups = new ArrayList<>();
-        ListSecretsRequest listSecretsRequest = new ListSecretsRequest();
-        ListSecretsResult listSecretsResult = secretsManager.listSecrets(listSecretsRequest);
-        while(!listSecretsResult.getSecretList().isEmpty()) {
-            for (final SecretListEntry entry : listSecretsResult.getSecretList()) {
-                groups.addAll(fetchSecret(secretsManager, context, entry.getName()));
-            }
-            final String nextToken = listSecretsResult.getNextToken();
-            if (nextToken == null) {
-                break;
-            }
 
-            listSecretsRequest.setNextToken(nextToken);
-            listSecretsResult = secretsManager.listSecrets(listSecretsRequest);
+        // Check if secret-name is set and return the single secret if it is
+        final PropertyValue secretNameProp = context.getProperty(SECRET_NAME);
+        if (secretNameProp.isSet()) {
+            groups.addAll(fetchSecret(secretsManager, secretNameProp.getValue()));
+        } else {
+            // if secret-name is not set, list all secrets and fetch the ones that match the pattern
+            final Pattern secretNamePattern = Pattern.compile(context.getProperty(SECRET_NAME_PATTERN).getValue());
+            ListSecretsRequest listSecretsRequest = new ListSecretsRequest();
+            ListSecretsResult listSecretsResult = secretsManager.listSecrets(listSecretsRequest);
+            while(!listSecretsResult.getSecretList().isEmpty()) {
+                for (final SecretListEntry entry : listSecretsResult.getSecretList()) {
+                    String secretName = entry.getName();
+                    if (!secretNamePattern.matcher(secretName).matches()) {
+                        getLogger().debug("Secret [{}] does not match the secret name pattern {}", secretName, secretNamePattern);
+                        continue;
+                    }
+                    groups.addAll(fetchSecret(secretsManager, secretName));
+                }
+                final String nextToken = listSecretsResult.getNextToken();
+                if (nextToken == null) {
+                    break;
+                }
+
+                listSecretsRequest.setNextToken(nextToken);
+                listSecretsResult = secretsManager.listSecrets(listSecretsRequest);
+            }
         }
 
         return groups;
@@ -184,18 +209,13 @@ public class AwsSecretsManagerParameterProvider extends AbstractParameterProvide
         return results;
     }
 
-    private List<ParameterGroup> fetchSecret(final AWSSecretsManager secretsManager, final ConfigurationContext context, final String secretName) {
+    private List<ParameterGroup> fetchSecret(final AWSSecretsManager secretsManager, final String secretName) {
         final List<ParameterGroup> groups = new ArrayList<>();
-        final Pattern secretNamePattern = Pattern.compile(context.getProperty(SECRET_NAME_PATTERN).getValue());
 
         final List<Parameter> parameters = new ArrayList<>();
 
-        if (!secretNamePattern.matcher(secretName).matches()) {
-            getLogger().debug("Secret [{}] does not match the secret name pattern {}", secretName, secretNamePattern);
-            return groups;
-        }
-
         final GetSecretValueRequest getSecretValueRequest = new GetSecretValueRequest().withSecretId(secretName);
+
         try {
             final GetSecretValueResult getSecretValueResult = secretsManager.getSecretValue(getSecretValueRequest);
 
--- a/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
+++ b/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
@@ -74,31 +74,27 @@ import java.util.regex.Pattern;
 public class AwsSecretsManagerParameterProvider extends AbstractParameterProvider implements VerifiableParameterProvider {
     enum ListingStrategy implements DescribedValue {
         Enumeration(
-                "Enumeration",
                 "Enumerate secret names",
                 "Provides a set of secret names to fetch, separated by a comma delimiter ','. " +
                         "This strategy requires the GetSecretValue permission only."
         ),
         Pattern(
-                "Pattern",
                 "Use regular expression",
                 "Provides a regular expression to match keys of attributes to filter for. " +
                         "This strategy requires both ListSecrets and GetSecretValue permissions."
         );
 
-        private final String value;
         private final String displayName;
         private final String description;
 
-        ListingStrategy(final String value, final String displayName, final String description) {
+        ListingStrategy(final String displayName, final String description) {
             this.displayName = displayName;
-            this.value = value;
             this.description = description;
         }
 
         @Override
         public String getValue() {
-            return this.value;
+            return name();
         }
 
         @Override
--- a/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
+++ b/nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
@@ -71,15 +71,20 @@ import java.util.regex.Pattern;
 @CapabilityDescription("Fetches parameters from AWS SecretsManager.  Each secret becomes a Parameter group, which can map to a Parameter Context, with " +
         "key/value pairs in the secret mapping to Parameters in the group.")
 public class AwsSecretsManagerParameterProvider extends AbstractParameterProvider implements VerifiableParameterProvider {
-
-    public static final PropertyDescriptor SECRET_NAME = new PropertyDescriptor.Builder()
-            .name("secret-name")
-            .displayName("Secret Name")
-            .description("The name of the secret to fetch parameters from. Will be used instead of the Secret Name Pattern property if set." +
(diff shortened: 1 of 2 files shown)


EndzeitBegins:
Thanks for your contribution @grishick.

Instead of one property overruling the other and `SECRET_NAME_PATTERN` remaining required even though it may not have any effect, what are your thoughts on introducing a "fetch mode" (or whatever name you see fit)?
With the current rather implicit approach, users might provide a `SECRET_NAME_PATTERN` of e.g. `FOO_*` and a `SECRET_NAME` of `BAR` simultaneously. 
On first look it might not be obvious what the behaviour is; which property overrules which or are the results from both combined?

I'm not familiar with this component, but to me it reads like introducing a different "mode" of fetching secrets.
Personally, I'd think first having to select a "fetch mode" with allowed values "Secret by Name" and "All Secrets by Pattern" (or something along those lines) and only then have the corresponding properties available would be more intuitive. 

What are your thoughts on this?

exceptionfactory (changes requested):
Thanks for the contribution @grishick, and thanks for the thoughtful feedback @EndzeitBegins.

I concur with the recommendation from @EndzeitBegins. Having an alternative retrieval strategy that avoids having to list all secret names seems useful, but it is important to make the configuration understandable.

With that in mind, a new property named something like `Secret Listing Strategy` with values of `Pattern` and `Enumerated` would be a good way forward. The default value would be `Pattern`, and the `Secret Name Pattern` property would depend on having set that strategy.

For the new property, listing only one Secret Name is rather limiting, so instead `Secret Names`, supporting one or more comma-separated values could be dependent on the `Enumerated` strategy.

grishick:
> Thanks for the contribution @grishick, and thanks for the thoughtful feedback @EndzeitBegins.
> 
> I concur with the recommendation from @EndzeitBegins. Having an alternative retrieval strategy that avoids having to list all secret names seems useful, but it is important to make the configuration understandable.
> 
> With that in mind, a new property named something like `Secret Listing Strategy` with values of `Pattern` and `Enumerated` would be a good way forward. The default value would be `Pattern`, and the `Secret Name Pattern` property would depend on having set that strategy.
> 
> For the new property, listing only one Secret Name is rather limiting, so instead `Secret Names`, supporting one or more comma-separated values could be dependent on the `Enumerated` strategy.

I got to agree, this is a better design. Will try to find time to rewrite it.

grishick:
> > Thanks for the contribution @grishick, and thanks for the thoughtful feedback @EndzeitBegins.
> > I concur with the recommendation from @EndzeitBegins. Having an alternative retrieval strategy that avoids having to list all secret names seems useful, but it is important to make the configuration understandable.
> > With that in mind, a new property named something like `Secret Listing Strategy` with values of `Pattern` and `Enumerated` would be a good way forward. The default value would be `Pattern`, and the `Secret Name Pattern` property would depend on having set that strategy.
> > For the new property, listing only one Secret Name is rather limiting, so instead `Secret Names`, supporting one or more comma-separated values could be dependent on the `Enumerated` strategy.
> 
> I got to agree, this is a better design. Will try to find time to rewrite it.

I think with the addition of `Secret Listing Strategy` there is no need for additional `Secret Names` parameter. Instead, the original `Secret Name Patter` parameter will be used either as a coma separated list or as a regular expression depending on the listing strategy.

exceptionfactory:
> > > Thanks for the contribution @grishick, and thanks for the thoughtful feedback @EndzeitBegins.
> > > I concur with the recommendation from @EndzeitBegins. Having an alternative retrieval strategy that avoids having to list all secret names seems useful, but it is important to make the configuration understandable.
> > > With that in mind, a new property named something like `Secret Listing Strategy` with values of `Pattern` and `Enumerated` would be a good way forward. The default value would be `Pattern`, and the `Secret Name Pattern` property would depend on having set that strategy.
> > > For the new property, listing only one Secret Name is rather limiting, so instead `Secret Names`, supporting one or more comma-separated values could be dependent on the `Enumerated` strategy.
> > 
> > 
> > I got to agree, this is a better design. Will try to find time to rewrite it.
> 
> I think with the addition of `Secret Listing Strategy` there is no need for additional `Secret Names` parameter. Instead, the original `Secret Name Patter` parameter will be used either as a coma separated list or as a regular expression depending on the listing strategy.

That could probably work, although it may involve changing the name and description of the property. With the `dependsOn()` builder method, the properties are shown or hidden based on the selected value, so in this case, it may be cleaner to have different properties to avoid any confusion.

grishick:
> > > > Thanks for the contribution @grishick, and thanks for the thoughtful feedback @EndzeitBegins.
> > > > I concur with the recommendation from @EndzeitBegins. Having an alternative retrieval strategy that avoids having to list all secret names seems useful, but it is important to make the configuration understandable.
> > > > With that in mind, a new property named something like `Secret Listing Strategy` with values of `Pattern` and `Enumerated` would be a good way forward. The default value would be `Pattern`, and the `Secret Name Pattern` property would depend on having set that strategy.
> > > > For the new property, listing only one Secret Name is rather limiting, so instead `Secret Names`, supporting one or more comma-separated values could be dependent on the `Enumerated` strategy.
> > > 
> > > 
> > > I got to agree, this is a better design. Will try to find time to rewrite it.
> > 
> > 
> > I think with the addition of `Secret Listing Strategy` there is no need for additional `Secret Names` parameter. Instead, the original `Secret Name Patter` parameter will be used either as a coma separated list or as a regular expression depending on the listing strategy.
> 
> That could probably work, although it may involve changing the name and description of the property. With the `dependsOn()` builder method, the properties are shown or hidden based on the selected value, so in this case, it may be cleaner to have different properties to avoid any confusion.

If I add `Secret Names` in addition to `Secret Listing Strategy` - would `Secret Names` and `Secret Pattern` have `required` set to `true` or to `false`?

exceptionfactory:
> If I add `Secret Names` in addition to `Secret Listing Strategy` - would `Secret Names` and `Secret Pattern` have `required` set to `true` or to `false`?

They can both be set to `required(true)`, but the framework ensures that only the one associated with the related strategy value will be shown and required.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
```diff
@@ -70,12 +71,26 @@
...
 public class AwsSecretsManagerParameterProvider extends AbstractParameterProvider implements VerifiableParameterProvider {
+    public static final String ENUMERATED_STRATEGY = "Enumerated";
+    public static final String PATTERN_STRATEGY = "Pattern";
+    public static final PropertyDescriptor SECRET_LISTING_STRATEGY = new PropertyDescriptor.Builder()
+            .name("secret-listing-strategy")
+            .displayName("Secret Listing Strategy")
+            .description("Strategy to use when listing secrets. 'Pattern' strategy treats Secret Name Pattern as a regular expression and fetches  " +
+                    "all secrets whose names match the pattern. 'Pattern' strategy requires ListSecrets and GetSecretValue permissions. "+
+                    "'Enumerated' strategy treats Secret Name Pattern as a coma-separated list and fetches all secrets whose names are in the list. " +
+                    "'Enumerated' strategy requires only GetSecretValue permission.")
```
exceptionfactory, line 82: Instead of describing the option in the property description, each value should be an instance of `AllowableValue` and include a description property.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
```diff
@@ -70,12 +71,26 @@
...
+            .defaultValue(PATTERN_STRATEGY)
+            .build();
 
     public static final PropertyDescriptor SECRET_NAME_PATTERN = new PropertyDescriptor.Builder()
             .name("secret-name-pattern")
             .displayName("Secret Name Pattern")
             .description("A Regular Expression matching on Secret Name that identifies Secrets whose parameters should be fetched. " +
-                    "Any secrets whose names do not match this pattern will not be fetched.")
+                    "Any secrets whose names do not match this pattern will not be fetched. Using this parameter requires the ListSecrets permission." +
+                    "This parameter is ignored if the Secret Name parameter is set.")
```
exceptionfactory, line 93: To avoid confusion, I think it is clearer to keep this property description unchanged, and add a new property with the a corresponding description.

exceptionfactory (changes requested):
Thanks for the initial updates @grishick. Based on the new strategy property, adding a new `Secret Names` property looks like it would be cleaner and easier to understand than overloading the existing `Secret Name Pattern` property.

As with other pull requests, please also submit the pull request to the `main` branch first, and then it can be considered for backporting in a follow-on PR.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
EndzeitBegins: I agree with David to move the description of the options to each option itself respectively.

An alternative approach to using `AllowableValue` is the declaration of an `enum` that implements the `DescribedValue` interface. An example can be found [here](https://github.com/apache/nifi/blob/92f571940edbb65c423d47b6881db638572fe466/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FilterAttribute.java#L268-L300). 
Due to improved type safety at the use site, this approach allows for exhaustive switch expressions, as can be seen [here](https://github.com/apache/nifi/blob/92f571940edbb65c423d47b6881db638572fe466/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FilterAttribute.java#L214-L218) and [here](https://github.com/apache/nifi/blob/92f571940edbb65c423d47b6881db638572fe466/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/FilterAttribute.java#L181-L185) for example. 

The advantage I see in this approach is that the compiler enforces to handle all options on all use sites, especially as new options are added. 

Regardless of the approach you take, moving the descriptions to the options is sensible.

grishick:
OK. I'll try to make the changes asap. It's a small change, but integrating it back to Anetac's fork is the tricky part.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
grishick: I added `dependsOn` to `SECRET_NAME_PATTERN` and added a new property called `SECRET_NAMES` also with `dependsOn` option. The new property `SECRET_NAMES` is expected to be a coma-separated list of AWS secret names. I've also added ` testFetchTwoSecrets` unit test to verify the behavior when `SECRET_NAMES` is a list.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
```diff
@@ -70,16 +70,38 @@
...
             .addValidator(StandardValidators.REGULAR_EXPRESSION_VALIDATOR)
-            .required(true)
+            .dependsOn(SECRET_LISTING_STRATEGY, PATTERN_STRATEGY)
             .defaultValue(".*")
             .build();
+
+    public static final PropertyDescriptor SECRET_NAMES = new PropertyDescriptor.Builder()
+            .name("secret-names")
+            .displayName("Secret Names")
+            .description("Comma-separated list of secret names to fetch. This parameter is ignored if the Secret Listing Strategy parameter is set to 'Pattern'.")
```
EndzeitBegins, line 101: As outlined before, no need to document this due to `dependsOn`.

```suggestion
            .description("Comma-separated list of secret names to fetch.")
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
```diff
@@ -70,16 +70,38 @@
...
+            .dependsOn(SECRET_LISTING_STRATEGY, ENUMERATED_STRATEGY)
```
EndzeitBegins, line 102: Akin to `SECRET_NAME_PATTERN`, this needs to be set, when the `SECRET_LISTING_STRATEGY` is set to `ListingStrategy.ENUMERATION`, thus we must keep the required(true) here.

```suggestion
            .dependsOn(SECRET_LISTING_STRATEGY, ListingStrategy.ENUMERATION)
            .required(true)
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
```diff
@@ -70,16 +70,38 @@
...
+    public static final String PATTERN_STRATEGY = "Pattern";
+    public static final PropertyDescriptor SECRET_LISTING_STRATEGY = new PropertyDescriptor.Builder()
+            .name("secret-listing-strategy")
+            .displayName("Secret Listing Strategy")
+            .description("Strategy to use when listing secrets. 'Pattern' strategy uses Secret Name Pattern property and fetches  " +
+                    "all secrets whose names match the pattern. 'Pattern' strategy requires ListSecrets and GetSecretValue permissions. "+
+                    "'Enumerated' strategy uses 'Secret Names' property and fetches all secrets in the list. " +
+                    "'Enumerated' strategy requires only GetSecretValue permission.")
+            .required(true)
+            .allowableValues(PATTERN_STRATEGY, ENUMERATED_STRATEGY)
```
EndzeitBegins, line 83: In case you're opting for using an enum, this can be simplified to:

```suggestion
            .allowableValues(ListingStrategy.class)
```

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
```diff
@@ -139,19 +163,35 @@ public List<ParameterGroup> fetchParameters(final ConfigurationContext context)
...
+                        continue;
+                    }
+                    groups.addAll(fetchSecret(secretsManager, secretName));
+                }
+                final String nextToken = listSecretsResult.getNextToken();
+                if (nextToken == null) {
+                    break;
+                }
+                listSecretsRequest.setNextToken(nextToken);
+                listSecretsResult = secretsManager.listSecrets(listSecretsRequest);
```
EndzeitBegins, line 193: You could untangle determining the set of secrets to fetch from the actual fetching in here, turning one intertwined action into two separate steps.

That is, first collecting a `Set<String>` of secretNames (based on either the provided enumeration or by fetching all available secret names and then filtering for those matching the regex) and then using that set to fetch the corresponding secrets.

EndzeitBegins (changes requested):
Thank you for your continued work on this @grishick.

There may have been a mistake due to the force-push, but some of the previous review feedback hasn't been addressed in the latest code version.
I've added some remarks and suggestions.

Additionally, as NiFi 2.0 is almost there, pull-requests should be addressing the `main` branch first before back porting changes to the `support/nifi-1.x` branch.

greg-anetac:
> Thank you for your continued work on this @grishick.
> 
> There may have been a mistake due to the force-push, but some of the previous review feedback hasn't been addressed in the latest code version. I've added some remarks and suggestions.
> 
> Additionally, as NiFi 2.0 is almost there, pull-requests should be addressing the `main` branch first before back porting changes to the `support/nifi-1.x` branch.

Roger that. I'll fix the merge. I was waiting to create a PR for main until discussions on this PR settle on implementation approach.

nifi-nar-bundles/nifi-aws-bundle/nifi-aws-parameter-providers/src/main/java/org/apache/nifi/parameter/aws/AwsSecretsManagerParameterProvider.java
grishick: Good point. Done.

grishick:
> > Thank you for your continued work on this @grishick.
> > There may have been a mistake due to the force-push, but some of the previous review feedback hasn't been addressed in the latest code version. I've added some remarks and suggestions.
> > Additionally, as NiFi 2.0 is almost there, pull-requests should be addressing the `main` branch first before back porting changes to the `support/nifi-1.x` branch.
> 
> Roger that. I'll fix the merge. I was waiting to create a PR for main until discussions on this PR settle on implementation approach.

done

grishick:
Here is a PR for `main` branch with the same change: https://github.com/apache/nifi/pull/9483

EndzeitBegins (changes requested):
Most of [David's review comments](https://github.com/apache/nifi/pull/9483#pullrequestreview-2414564111) on #9483 apply here too. 

The only things that isn't available for `1.x` that comes to my mind is `asAllowableValue()`. 
`List.of` isn't available for Java 8, so the `PROPERTIES` declaration can stay as is.

exceptionfactory:
Thanks again for your work on this improvement @grishick. I have merged the corresponding pull request for the main branch and closed the Jira issue.

Voting is [currently in progress](https://lists.apache.org/thread/4wgrlytm3xgsc67l32m83rl2g77dsqvs) to declare the NiFi version 1 series as End of Support in a few weeks. We are moving to wrap up a 1.28.1 patch release version soon, and we are not considering new features, so I'm closing this pull request for the support branch.

PR #9387, opened 2024-10-14
NIFI-13868: add `pg-delete` command to Nifi Toolkit

grishick:
# Summary

[NIFI-13868](https://issues.apache.org/jira/browse/NIFI-13868)

Adding `pg-delete` command to Nifi Toolkit. The command works the same way as deleting a process group via UI. If there are messages in PG queue, deletion will fail. I ported this change from [Anetac's Nifi fork](https://github.com/Eng-Anetac/nifi) where I made it a while ago. We have been using this feature in production at Anetac for months now.

PR with the same feature for `main` branch: https://github.com/apache/nifi/pull/9388
The change for support/nifi-1.x branch is larger than for main, because the support branch does not have changes for NIFI-12898

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

- [x] Pull Request based on current revision of the `support/nifi-1.x` branch

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/client/nifi/ProcessGroupClient.java
+++ b/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/client/nifi/ProcessGroupClient.java
@@ -75,4 +75,6 @@ public interface ProcessGroupClient {
         throws NiFiClientException, IOException;
 
     FlowComparisonEntity getLocalModifications(String processGroupId) throws NiFiClientException, IOException;
+
+    void deleteProcessGroup(ProcessGroupEntity entity) throws NiFiClientException, IOException;
 }
--- a/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/client/nifi/ProcessGroupClient.java
+++ b/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/client/nifi/ProcessGroupClient.java
@@ -76,5 +76,5 @@ public interface ProcessGroupClient {
 
     FlowComparisonEntity getLocalModifications(String processGroupId) throws NiFiClientException, IOException;
 
-    void deleteProcessGroup(ProcessGroupEntity entity) throws NiFiClientException, IOException;
+    ProcessGroupEntity deleteProcessGroup(ProcessGroupEntity entity) throws NiFiClientException, IOException;
 }
--- a/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/nifi/pg/PGDelete.java
+++ b/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/nifi/pg/PGDelete.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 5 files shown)


nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/client/nifi/ProcessGroupClient.java
```diff
@@ -75,4 +75,6 @@ FlowEntity copySnippet(String processGroupId, CopySnippetRequestEntity copySnipp
         throws NiFiClientException, IOException;
 
     FlowComparisonEntity getLocalModifications(String processGroupId) throws NiFiClientException, IOException;
+
+    void deleteProcessGroup(ProcessGroupEntity entity) throws NiFiClientException, IOException;
```
exceptionfactory, line 79: As other delete commands return the deleted entity, that looks like a good pattern to follow in this scenario.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/nifi/pg/PGDelete.java
```diff
@@ -0,0 +1,67 @@
...
+
+        @Override
+        public VoidResult doExecute(final NiFiClient client, final Properties properties)
+                throws NiFiClientException, IOException, MissingOptionException {
+
+            final String pgId = getRequiredArg(properties, CommandOption.PG_ID);
+
+            final ProcessGroupClient pgClient = client.getProcessGroupClient();
+            final ProcessGroupEntity pgEntity = pgClient.getProcessGroup(pgId);
+            if(pgEntity == null) {
```
exceptionfactory, line 60: Spacing:
```suggestion
            if (pgEntity == null) {
```

exceptionfactory (changes requested):
Thanks for the contribution @grishick. The general implementation looks good, just one functional note on returning the deleted entity instead of void.

Also, please submit new features to the main branch first. Once that is approved and merged, then submitting a separate pull request for the support branch would be helpful if it is desirable to have this functionality on the support branch.

nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/client/nifi/ProcessGroupClient.java
grishick: done

grishick:
> Thanks for the contribution @grishick. The general implementation looks good, just one functional note on returning the deleted entity instead of void.
> 
> Also, please submit new features to the main branch first. Once that is approved and merged, then submitting a separate pull request for the support branch would be helpful if it is desirable to have this functionality on the support branch.

Got it. Here is the PR for main branch: https://github.com/apache/nifi/pull/9388

pvillard31 (commented):
Thanks for the improvement. I'd recommend adding more details in the helper text for this command. For this command to be successful, three actions would need to be taken before:
- stop the processors in the process group (`pg-stop`)
- disable the controller services in the process group (`pg-disable-services`)
- empty all queues in the process group (we don't have a CLI command for that now, would certainly be a nice addition)

I didn't test the current PR but would be good to confirm that everything works fine especially when having nested process groups with controller services at multiple levels, etc.

grishick:
Sounds good. I'll try to make the changes asap.

exceptionfactory (approved):
Thanks @grishick, I corrected formatting on `PGDelete` as in the PR for the main branch. The rest looks good. +1 merging

PR #9395, opened 2024-10-16
NIFI-13877: JdbcCatalog connections are not closed properly in PutIceberg

mark-bathori:
# Summary

[NIFI-13877](https://issues.apache.org/jira/browse/NIFI-13877)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (commented):
Thanks for addressing this issue @mark-bathori. Although this change looks straightforward, making use of `BaseMetastoreCatalog` as a `Closeable` in try-with-resources, what do you think about simply checking if the `Catalog` instance implements `Closeable` and handling that in a `finally` block in `loadTable`? That would minimize the scope of changes.

mark-bathori:
Thanks for the review @exceptionfactory. Unfortunately the catalog can't be closed within the `loadTable` method because the jdbc connection is used in later part of the processor when the commit happens to the table (this is used via the table object). The finally block could be added to the lower try-catch block so the `Catalog` type could be kept but the `loadCatalog` extraction would be still needed to make the catalog object accessible.

Currently every supported `Catalog` implementation extended from `BaseMetastoreCatalog` but if there will be other catalog support added in the future that is not extended from that class e.g. RESTCatalog, then your suggested solution will be needed (checking if the `Catalog` instance implements `Closeable` and handling that in a finally block)
Let me know what you think.

exceptionfactory:
> Thanks for the review @exceptionfactory. Unfortunately the catalog can't be closed within the `loadTable` method because the jdbc connection is used in later part of the processor when the commit happens to the table (this is used via the table object). The finally block could be added to the lower try-catch block so the `Catalog` type could be kept but the `loadCatalog` extraction would be still needed to make the catalog object accessible.
> 
> Currently every supported `Catalog` implementation extended from `BaseMetastoreCatalog` but if there will be other catalog support added in the future that is not extended from that class e.g. RESTCatalog, then your suggested solution will be needed (checking if the `Catalog` instance implements `Closeable` and handling that in a finally block) Let me know what you think.

Thanks for the reply @mark-bathori. With the growing popularity of REST Catalog implementations, particularly Apache Polaris, I expect this feature to be requested in the future. It may also require more significant refactoring, but I think keeping it in mind, in this particular change, is worth doing, so I recommend avoiding the switch to `BaseMetastoreCatalog`.

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-common/src/main/java/org/apache/nifi/processors/iceberg/catalog/ClosableCatalog.java
```diff
@@ -0,0 +1,196 @@
...
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Wrapper class for Iceberg catalogs
+ */
+public class ClosableCatalog implements Catalog, Closeable {
```
exceptionfactory, line 36:
```suggestion
public class CloseableCatalog implements Catalog, Closeable {
```

exceptionfactory (changes requested):
Thanks for the updates @mark-bathori, however, I think the `ClosableCatalog` wrapper is a step in a different direction. The wrapper interface will require additional maintenance as new methods may be added in future Iceberg versions. Did you consider defining `Catalog` outside of the `try` block and simply do an `instanceof` check in a new `finally` block?

exceptionfactory:
Thanks for the latest update @mark-bathori, this approach looks good. Planning to merge pending successful CI builds.

exceptionfactory (approved):
+1 merging, thanks @mark-bathori!

PR #9396, opened 2024-10-16
[NIFI-13876] - Adding context aware error banners

rfellows:
[NIFI-13876](https://issues.apache.org/jira/browse/NIFI-13876)

Now, instead of only having a single spot in the store for errors to land that drive banners... we have support for named/contextual spots. This helps support having 2 components on the screen with different errors in a banner at the same time.

`<context-error-banner>` is now the preferred way of adding error banners to a screen/dialog. It only takes the ContextErrorKey it is interested in as input. The component handles watching the store for errors of that key and clearing the store when it is destroyed and when the user hits the dismiss button too.


scottyaslan (commented):
@mcgilman are we ok removing the disc on the ul/li elements? Are we good with the padding at the bottom with the "DISMISS" button off on it's own row?

Previously:

scottyaslan (commented):
@mcgilman Rob didn't change anything do with with when the error messages are cleared in this PR but I had a question as to the value of keeping old error messages on subsequent submits. Theses error messages can pile up and make the dialog become unusable:

Screenshot 2024-10-16 at 4 04 03 PM

If we don't want to clear the old error messages on subsequent submits another option could be to set a max-height and make the errors scrollable.

scottyaslan (approved):
The contextual error banners functionally work well so I am a +1. I left a few questions/comments for possible future improvements.

scottyaslan:
I am +1. If we decide the padding at the bottom is too much that is an easy fix.

PR #9428, opened 2024-10-21
[NIFI-13911] - force change detection after setting local variables i…

rfellows:
…ntended to drive the visibility of the overlay to prevent slow response when opening nf-editor of a dependent property

[NIFI-13911](https://issues.apache.org/jira/browse/NIFI-13911)


scottyaslan:
Reviewing...

scottyaslan:
LGTM!

scottyaslan (approved):
Works well! Good solution.

PR #9429, opened 2024-10-21
NIFI-13824 Install processor dependencies in one go

fgerlits:
# Summary

[NIFI-13824](https://issues.apache.org/jira/browse/NIFI-13824)

If a Python processor defines dependencies both inline and in a requirements.txt file, then we need to install the two groups of dependencies in a single `pip install` command, otherwise `pip` is not able to resolve the web of dependencies correctly.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/main/python/framework/ExtensionManager.py
```diff
@@ -285,37 +285,31 @@ def import_external_dependencies(self, processor_details, work_dir):
             logger.info("All dependencies have already been imported for {0}".format(class_name))
             return True
 
-        python_cmd = os.getenv("PYTHON_CMD")
+        packages_to_install = []
```
exceptionfactory, line 288: Recommend renaming this variable as it includes both packages and `requirements.txt` when found.
```suggestion
        dependency_references = []
```

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/test/python/framework/TestDownloadDependencies.py
```diff
@@ -0,0 +1,90 @@
...
+            python_command = os.environ.get("PYTHON_CMD")
+            processor_packages_dir = os.path.join(packages_dir, 'extensions', 'ProcessorWithDependencies', '0.0.1')
+            pip_list_result = subprocess.run([python_command, '-m', 'pip', 'list', '--path', processor_packages_dir], capture_output=True)
+            self.assertEqual(pip_list_result.returncode, 0)
+            pip_list_lines = [line.decode() for line in pip_list_result.stdout.splitlines()]
+
+            grpcio_version = get_version(pip_list_lines, 'grpcio')
+            self.assertIsNotNone(grpcio_version)
+            grpcio_status_version = get_version(pip_list_lines, 'grpcio-status')
+            self.assertEqual(grpcio_version, grpcio_status_version)
```
exceptionfactory, line 87: With the assertion based on the `grpcio` library, I recommend changing the dependencies to list it explicitly, instead of depending on it as a transitive dependency.

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/test/resources/python/framework/processor_with_dependencies/requirements.txt
```diff
@@ -0,0 +1 @@
+google-cloud-vision==3.7.4
```
exceptionfactory, line 1: This file is missing the license header.

exceptionfactory (changes requested):
Thanks for working on this improvement @fgerlits.

I noted a couple code-level recommendations, but also noticed that the new Python unit test is failing on both macOS and Ubuntu Linux. It looks like an issue with the expected availability of a directory.

```
Traceback (most recent call last):
  File "/home/runner/work/nifi/nifi/nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/test/python/framework/TestDownloadDependencies.py", line 76, in test_import_external_dependencies
    extension_manager.import_external_dependencies(details, packages_dir)
  File "/home/runner/work/nifi/nifi/nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/main/python/framework/ExtensionManager.py", line 305, in import_external_dependencies
    result = subprocess.run(args)
  File "/usr/lib/python3.10/subprocess.py", line 503, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1738, in _execute_child
    and os.path.dirname(executable)
  File "/usr/lib/python3.10/posixpath.py", line 152, in dirname
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/test/python/framework/TestDownloadDependencies.py
fgerlits: The point of this test is that both packages depend on `grpcio`, but with different version bounds, so if we install one after the other, we'll end up with inconsistent versions. If I add a specific `grpcio` version to both places, then the old code will work, as well, so the test will no longer show what was fixed.

Is it OK if I add a comment here in the test file with an explanation of why we are checking the `grpcio` versions?

exceptionfactory: Thanks for clarifying, that is helpful. Yes, I think simply adding a comment above these assertions is sufficient.

fgerlits:
> the new Python unit test is failing on both macOS and Ubuntu Linux

This was because we need to set `PYTHON_CMD` for some of the Python code to work; fixed in 3889d12beb6162172ca5ee70c7ec9a47a9450210

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/main/python/framework/ExtensionManager.py
fgerlits: I have renamed it in c32d56bfb247e80fb072c062ea6b1e32d1a4f3d2

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/test/python/framework/TestDownloadDependencies.py
fgerlits: I have add the comments in bc87377435b02c67a599d8bdc6a0868831148639

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-framework/src/test/resources/python/framework/processor_with_dependencies/requirements.txt
fgerlits: I have added the header in 6d97284da4104a0695a86c23ecd80aa5ac104629

exceptionfactory (changes requested):
Thanks for the updates @fgerlits.

The Maven Wrapper script should not be changed to resolve the Python command environment variable, since it comes from the Apache Maven Wrapper project, and is focused solely on running Maven itself.

As mentioned in one comment, using `sys.executable` in the test class is one potential option for resolving the Python command to use for running the test method.

mvnw
```diff
@@ -105,6 +106,21 @@ trim() {
...
+      export PYTHON_CMD="${PYTHON_PATH}"
+      break
+    fi
+  done
+
+  if [ -z "${PYTHON_CMD-}" ]; then
+    echo "Python executable not found; please set PYTHON_CMD" >&2
+    return 1
+  fi
+fi
```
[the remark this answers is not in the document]
fgerlits: fixed in 7af4f4ce001d5f3c6fb22b208c10535d109d3fc3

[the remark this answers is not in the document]
exceptionfactory: Thanks for the adjusted approach @fgerlits.

Reviewing the test itself, it seems like this is more suitable as an integration test under the `nifi-system-test-suite` module. The `system-tests` GitHub workflow already installs Python, and it is a better place to run longer types of tests, such as this one that require installing dependency sets. What do you think about refactoring this test into the system test module?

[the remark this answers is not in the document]
fgerlits: You are right, a unit test should not download stuff from the internet (and take 20+ seconds).  In e3ab1c7c53742fe22cddcc8384018e3a55e93f81, I have mocked out the `subprocess.run` call, and now only test that it gets called once instead of twice, as it used to be.

I also agree it would be a good idea to add an integration test which does the package installations and version checks; can we do that separately, in another pull request?

exceptionfactory (changes requested):
Thanks for the updates @fgerlits.

Although not wanting to extend this review unnecessarily, I'm now wondering about the value of the `TestDownloadDependencies.py` class. By mocking the executable call, it is no longer exercising the behavior in a way that confirms the expected results.

If you would like defer implementing a system test for a separate PR, that is acceptable, and I would be willing to have this go forward without a system test. However, if that is the direction, then I would remove the existing test code, and remove the changes from the `ci-workflow`. Then we can rely on runtime verification for these changes.

.github/workflows/ci-workflow.yml
```diff
@@ -162,6 +162,10 @@ jobs:
           distribution: 'corretto'
           java-version: '21'
           cache: 'maven'
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
```
fgerlits, line 168: This is a bugfix: we were already using Python in unit tests, and without this step, the CI jobs were using whatever Python version was available by default in the GitHub Action image (which happened to be a homebrew-installed Python 3.13 in the case of the MacOS image).  With the explicit setup step, now the CI jobs are using Python 3.12 in a predictable and reproducible way.

Also, with system tests using Python 3.10 and unit tests using Python 3.12, we can verify that the Python components in NiFi work with both, which is a good thing.

mvnw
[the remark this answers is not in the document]
fgerlits: This unit test has value: it loads the ExtensionManager class, so if there are e.g. syntax errors, they get caught earlier; and it validates the change in the PR in the sense that if you run the test with the ExtensionManager code on `main`, it will fail: we are now calling `pip install` once instead of twice.

I have renamed and refactored it to make it clearer that this is just a unit test for a single method in the ExtensionManager class, and also to make it easier to add unit tests for the rest of the methods later.

.github/workflows/ci-workflow.yml
exceptionfactory: Thanks for clarifying, that makes sense, and I agree that installing a specific version is the right approach.

exceptionfactory (approved):
Thanks again for putting this together and working through the feedback @fgerlits, the latest version looks good! +1 merging

PR #9435, opened 2024-10-22
NIFI-13921: Routing away from the edit path when closing the EditProc…

mcgilman:
…essGroup dialog.


rfellows (approved):
LGTM 👍 
Thanks, @mcgilman

PR #9437, opened 2024-10-22
NIFI-13904 - Fix CLI commands to set/remove parameter context inheritance

pvillard31:
# Summary

[NIFI-13904](https://issues.apache.org/jira/browse/NIFI-13904) - Linking Parameter Contexts using NiFi Toolkit causes sensitive parameters to become invalid

The current implementation of the CLI for setting / removing the parameter context inheritance is broken when the parameter context to update contains sensitive parameters. The reason is that, currently, the update parameter request is adding all of the parameter entities in the request. However, when we retrieve sensitive parameters, we never really retrieve the sensitive values on the client side, we only get `*****` strings. Because of this, the current update request is actually updating the values of the sensitive parameters to this  `*****` value and this is what makes referencing components invalid. The update request needs to be updated so that the value of the sensitive parameters is explicitly set to `null` so that the backend knows that the sensitive parameters are not changed by this update request.

This pull request also slightly improves the display output of parameters retrieved via a parameter provider. I'm happy to put this into a separate PR if we think this is better. While reproducing the issue, I noticed that the command to fetch parameters from a parameter provider is not giving information about the created parameter contexts so this additional information can be useful and removes the need for listing parameter contexts.

Example:

````
./bin/cli.sh nifi fetch-params -p nifi-cli.properties -ppid b56a9960-0192-1000-b4f7-33908ea4027f --sensitiveParamPattern ".*ensitive.*" --applyParameters

Property Name                 Property Value
---------------------------   ------------------------------------------------------
parameter-group-directories   /Users/pierre/dev/nifi_use_cases/NIFI-13904/parameters
parameter-value-byte-limit    256 B
parameter-value-encoding      plaintext


Parameter Group   Parameter Context Id                   Parameter Context Name   Fetched Parameter Name
---------------   ------------------------------------   ----------------------   ----------------------
parameters        b56e3ce7-0192-1000-65fd-1ce89d329bbe   parameters               myParameter
parameters        b56e3ce7-0192-1000-65fd-1ce89d329bbe   parameters               mySensitiveParameter
````

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/nifi/params/FetchParams.java
+++ b/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/nifi/params/FetchParams.java
@@ -23,11 +23,13 @@ import org.apache.nifi.toolkit.cli.api.CommandException;
 import org.apache.nifi.toolkit.cli.api.Context;
 import org.apache.nifi.toolkit.cli.impl.client.nifi.NiFiClient;
 import org.apache.nifi.toolkit.cli.impl.client.nifi.NiFiClientException;
+import org.apache.nifi.toolkit.cli.impl.client.nifi.ParamContextClient;
 import org.apache.nifi.toolkit.cli.impl.client.nifi.ParamProviderClient;
 import org.apache.nifi.toolkit.cli.impl.command.CommandOption;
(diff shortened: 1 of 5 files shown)


exceptionfactory (approved):
Thanks for fixing the issue with sensitive parameter value handling in the CLI @pvillard31. The background description and comments are helpful. +1 merging

PR #9442, opened 2024-10-23
[NIFI-13917] update sidenav font color to match applicaiton font color

scottyaslan:
Previously:
Screenshot 2024-10-23 at 12 20 50 PM

With this change applied:
Screenshot 2024-10-23 at 12 20 18 PM

Now the font color of the PG name matches the operation palette.


mcgilman:
Reviewing...

mcgilman (approved):
Thanks @scottyaslan!

PR #9443, opened 2024-10-23
NIFI-13926: Removed used of default branch name in favor of null in I…

mcgilman:
…mport From Registry and Start Version Control.

- Reset form fields as necessary when queries to load buckets, branches, etc fail.


rfellows:
reviewing...

rfellows (approved):
Nice fix, @mcgilman. Thanks for jumping in to fix this so fast.

PR #9445, opened 2024-10-24
NIFI-13927 - Refactor PublishGCPPubSub to fix stuck thread issue when using records

pvillard31:
# Summary

[NIFI-13927](https://issues.apache.org/jira/browse/NIFI-13927) - PublishGCPPubSub processor stop working and is stucked when using Record Oriented mode

In `org.apache.nifi.processors.gcp.pubsub.publish.FlowFileResult.reconcile()` (https://github.com/apache/nifi/blob/main/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/pubsub/publish/FlowFileResult.java#L60), we can be in a case where 
````java
while (futures.size() > (successes.size() + failures.size()))
````
will never be true.

It means that, for a reason I didn't identify, `TrackedApiFutureCallback` is not always going into the callback `onSuccess`  or that the `successes` list is not updated as it should. For example, when I'm stuck in the while loop, I have `futures.size() = 2500`, `successes.size() = 2495`, `failures.size() = 0`, and I can confirm that I'm able to consume 2500 messages so all messages have been properly published while the flowfile is stuck in the Publish processor.

The feeling is that the approach can be refactored to not rely on the callback but instead directly check the futures once they all completed and check one by one what is the status (ie. is there a message ID or a throwable?).

````java
try {
    ApiFutures.allAsList(futures).get();
} catch (InterruptedException | ExecutionException e) {
    logger.error("Failed to reconcile PubSub send operation status", e);
}
// we now know that all futures have completed
// we go over all futures to see if we have the message ID (which means this is a success)
// and we can see if some futures threw exception to deal with failures
````

This PR refactors the code by taking this approach. Some of the code changes have been made to support existing unit tests without too much refactoring in the unit tests. Besides, the changes have been tested with the flow definition provided in the JIRA as well as in more advanced scenarios (different batch sizes, concurrent tasks, etc) and everything seems to be working well.

The logging has also been slightly changed to properly report bulletins at processor level when a flow file is sent into the failure and/or retry relationships.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


greyp9:
> @pvillard31 Thanks for your analysis on the issue! The missing piece of the puzzle is that the `successes` list is not synchronized and the callback threads write the list in parallel in `TrackedApiFutureCallback.onSuccess()` which caused lost items (the same is true for the `failures` list). So the issue could be fixed via adding some synchronization stuff around the lists (e.g. `synchronized` block, `Collections.synchronizedList()` or `CopyOnWriteArrayList`). I have not checked your solution in detail and I have no strong opinion at the moment. Just wanted to clarify the root cause and to mention other possible solutions to it.

Thanks @turcsanyip - this is my analysis as well.  Once I had things set up, this issue was fairly easy to reproduce.  After adding synchronization to the `successes` and `failures` lists, the logic ran cleanly, even when loaded.

pvillard31:
Thanks @turcsanyip, that makes total sense and thanks @greyp9 for testing and discussions on the Slack thread. If we prefer to just use a synchronized list instead, I'm totally OK with this approach but, as I said before, I don't really see the value of maintaining successes and failures lists as we're not really doing anything with it (like logging number of failures vs number of successes, etc).

If we prefer to maintain the current approach and change to a synchronized list, I do think a bit of reworking is still needed to expose the logs from FlowFileResult as processor level bulletins - right now, routing flow files to failure or retry would not trigger any bulletin.

Happy to file another PR if that's the consensus.

greyp9:
> Thanks @turcsanyip, that makes total sense and thanks @greyp9 for testing and discussions on the Slack thread. If we prefer to just use a synchronized list instead, I'm totally OK with this approach but, as I said before, I don't really see the value of maintaining successes and failures lists as we're not really doing anything with it (like logging number of failures vs number of successes, etc).
> 
> If we prefer to maintain the current approach and change to a synchronized list, I do think a bit of reworking is still needed to expose the logs from FlowFileResult as processor level bulletins - right now, routing flow files to failure or retry would not trigger any bulletin.
> 
> Happy to file another PR if that's the consensus.

FlowFileResult / RelationshipMapper acquires the success messageID from the first success, and iterates through the failures to determine whether the FlowFile should be considered retryable.  Other than that, it is true that the lists are mostly unused.

It seems like a good idea that failure messages might be logged somehow, or that non-zero successCount / failureCount could be added as FlowFile attributes.  I wish I had considered this during the original implementation.  It is also a great idea to expose the FlowFileResult as processor level bulletins.

Given the timing of this (pending releases), and depending on your availability, would you consider a two-phase approach here?  I could put up a simple PR to synchronize the lists, which would prevent the hang.  Given additional time, your reworked PR could implement additional logging and tracking to improve the user experience.  Then again, a reworked PR might end up being simple enough to quickly turn around by itself.

I'm open to your thoughts, and set up to test whatever you think best.

pvillard31:
Given upcoming releases, I've filed #9449 to fix the immediate issue and we can discuss further on nice improvements for the processor. Will close this one for now to avoid confusion.

PR #9446, opened 2024-10-24
NIFI-13929: Removed Provenance Repository from the stateless Reposito…

markap14:
…ryContextFactory and added it to the DataflowTriggerContext. This was necessary because the previous design overlooked the possibility of many threads concurrently running the same dataflow. They all shared the same StatelessProvenanceRepository, but the code was designed as if only a single thread would be using the repository. As a result, the events that were registered with the stateless prov repo were being copied many times into NiFi's underlying provenance repository. This refactoring also led to the discovery of some old Java 8 syntax that could be cleaned up, and it led to the discovery of some methods that were no longer being used and could be cleaned up. Finally, in testing, I found that when a Stateless Group was scheduled, it scheduled the triggering of the stateless group before marking the state as RUNNING; as a result, the second thread could run, determine that the state is STARTING instead of RUNNING, and return without triggering the stateless group. This was addressed by ensuring that we set the state to RUNNING before triggering the stateless group to be triggered.

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/flow/NopProvenanceEventRepository.java
+++ b/nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/flow/NopProvenanceEventRepository.java
@@ -1,0 +1,370 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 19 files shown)


nifi-stateless/nifi-stateless-api/src/main/java/org/apache/nifi/stateless/flow/NopProvenanceEventRepository.java
```diff
@@ -0,0 +1,370 @@
...
+import org.apache.nifi.processor.Relationship;
+import org.apache.nifi.provenance.ProvenanceEventBuilder;
+import org.apache.nifi.provenance.ProvenanceEventRecord;
+import org.apache.nifi.provenance.ProvenanceEventRepository;
+import org.apache.nifi.provenance.ProvenanceEventType;
+
+import java.util.List;
+import java.util.Map;
+
+public class NopProvenanceEventRepository implements ProvenanceEventRepository {
```
exceptionfactory, line 30: As this class is in `nifi-stateless-api`, but seems to have internal use, should the visibility be changed to package-private?

exceptionfactory (commented):
Thanks for refactoring the provenance handling for stateless to correct this issue @markap14. I just noted one question about the visibility of the `NopProvenanceRepository` class, but not a blocker. I plan to take a second look soon.

exceptionfactory (approved):
Thanks again for resolving this issue and making the other improvements to stateless handling @markap14, the latest version looks good. +1 merging

PR #9465, opened 2024-10-28
NIFI-13944: Upgrade Redis to 5.2.0

mattyb149:
# Summary

[NIFI-13944](https://issues.apache.org/jira/browse/NIFI-13944) This PR upgrades the Redis library to 5.2.0 along with the corresponding Spring Redis version, and excludes the banned org.json:json dependency in favor of Ted Dunning's version.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
+++ b/nifi-extension-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
@@ -54,10 +54,6 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-beans</artifactId>
-        </dependency>
-        <dependency>
             <groupId>redis.clients</groupId>
             <artifactId>jedis</artifactId>
             <version>${jedis.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -853,7 +853,7 @@
                                         <!-- JUnit 5 is the preferred test framework -->
                                         <exclude>org.testng:testng</exclude>
                                         <!-- Cat-X Deps -->
-                                        <exclude>org.json:json:*:*:compile</exclude>
+                                        <exclude>org.json:json:[,20240303]:compile</exclude>
                                         <exclude>c3p0:c3p0:*:*:compile</exclude>
                                         <!-- Versions of JSR305 before 3.0.1 are not allowed https://github.com/findbugsproject/findbugs/issues/128 -->
                                         <exclude>com.google.code.findbugs:jsr305:[,3.0.0]:compile</exclude>
--- a/pom.xml
+++ b/pom.xml
@@ -853,7 +853,7 @@
                                         <!-- JUnit 5 is the preferred test framework -->
                                         <exclude>org.testng:testng</exclude>
                                         <!-- Cat-X Deps -->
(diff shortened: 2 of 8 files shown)


joewitt:
Seems to build.   But the bundled dependencies of Spring seem considerable and likely avoidable.  The LICENSE/NOTICE is unchanged but clearly does need to change.

Seems like at least a few modules also included that are likely in a parent class loader.  Reviewing further but this is a good example of a dependency where maintenance jumps (major lines no less) are much more than bumping versions.

exceptionfactory (approved):
Thanks for the upgrade @mattyb149, this looks good from my perspective.

@joewitt, reviewing the current Redis NAR bundles, they contain roughly the same set of Spring Framework dependencies, most of which are transitive through Spring Data Redis. From that perspective, I think the upgrade works as it stands, although it does highlight the potential need for further review in a separate issue. It is worth noting, however, that only the Framework NAR contains the Spring libraries, so these would not be inherited from the `nifi-standard-shared-nar`. I will defer to you for further review.

exceptionfactory:
Although for clarification, it looks like the `NOTICE` files are outdated, so I agree they should be updated to reflect current contents and versions.

nifi-extension-bundles/nifi-redis-bundle/nifi-redis-extensions/pom.xml
```diff
@@ -62,6 +62,17 @@
             <artifactId>jedis</artifactId>
             <version>${jedis.version}</version>
             <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.json</groupId>
+                    <artifactId>json</artifactId>
```
exceptionfactory, line 68: This version of the `json` librar [20240303](https://central.sonatype.com/artifact/org.json/json/20240303) is actually licensed under the Public Domain, not the historically banned license. For that reason, the exclusion is not necessary and the banned dependency settings can be adjusted.

exceptionfactory (changes requested):
As discussed in an offline conversation, the license and notice files contain more information than needed as dependencies with the `provided` scope should not be referenced.

exceptionfactory (changes requested):
This looks closer to completion, but the information in `NOTICE` looks like it reflects what is in `nifi-redis-service-api-nar`, not what is in `nifi-redis-nar`, so both NOTICE files need to be updated.

pom.xml
```diff
@@ -853,7 +853,7 @@
                                         <!-- JUnit 5 is the preferred test framework -->
                                         <exclude>org.testng:testng</exclude>
                                         <!-- Cat-X Deps -->
-                                        <exclude>org.json:json:*:*:compile</exclude>
+                                        <exclude>org.json:json:[,20240303]:compile</exclude>
```
exceptionfactory, line 856: Reviewing the [Maven Version Ranges](https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html), it looks like this should use parentheses to indicate `x < 20240303`:
```suggestion
                                        <exclude>org.json:json:(,20240303):compile</exclude>
```

exceptionfactory (approved):
Thanks for the updates @mattyb149. On further review, I noticed several additional updates needed, which I pushed in the latest commit. +1 merging

PR #9476, opened 2024-10-31
NIFI-13952 - Flow Analysis Rule to restrict backpressure configuration

pvillard31:
# Summary

[NIFI-13952](https://issues.apache.org/jira/browse/NIFI-13952) - Flow Analysis Rule to restrict backpressure configuration

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private enum BackpressureViolationType {
+
+        BP_COUNT_THRESHOLD_BELOW_LIMIT("BackpressureCountThresholdTooLow"),
+        BP_COUNT_THRESHOLD_ABOVE_LIMIT("BackpressureCountThresholdTooHigh"),
+        BP_SIZE_THRESHOLD_BELOW_LIMIT("BackpressureSizeThresholdTooLow"),
+        BP_SIZE_THRESHOLD_ABOVE_LIMIT("BackpressureSizeThresholdTooHigh");
+
+        private String id;
```
mosermw, line 231: Recommend making this variable final.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Tags({"connection", "backpressure"})
+@CapabilityDescription("This rule will generate a violation if backpressure settings of a connection exceed configured thresholds. "
+        + "Improper configuration of backpressure settings can lead to decreased performances because of excessive swapping as well as "
+        + "to filled up content repository with too much in-flight data in NiFi.")
```
mosermw, line 46: Looks like a grammar typo here.  Recommend the second sentence read
"Improper configuration of backpressure settings can lead to decreased performance because of excessive swapping and can fill up the content repository with too much in-flight data."

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
```diff
@@ -0,0 +1,242 @@
...
+                                BackpressureViolationType.BP_SIZE_THRESHOLD_ABOVE_LIMIT,
+                                getViolationMessage(BackpressureViolationType.BP_SIZE_THRESHOLD_ABOVE_LIMIT, connection.getBackPressureDataSizeThreshold(), context.getProperty(SIZE_MAX).getValue())));
+                    }
+                }
+            );
+
+        return results;
+    }
+
+    private GroupAnalysisResult buildViolation(final VersionedConnection connection, final VersionedComponent source,
```
mosermw, line 181: This method chooses whether to assign the violation to the connection or nearest real Processor.  Ideally, backpressure violations would always be assigned to the connection.  Was this done as a workaround for NIFI-13953 because the UI does not showing a "Go To" link for connection violations?

I'm not sure we should keep a workaround like this.  If you feel that we need the workaround, then a comment in the code to that effect should be here, stating why it works like this and when it should be replaced.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
```diff
@@ -0,0 +1,242 @@
...
+    private String getLocationMessage(final VersionedConnection connection, final VersionedComponent source, final VersionedComponent destination) {
+        if (source == null || destination == null) {
+            return "The connection [" + connection.getIdentifier() + "] is violating the rule for backpressure settings. ";
+        }
+        return "The connection [" + connection.getIdentifier() + "] connecting " + source.getName() + " [" + source.getIdentifier() + "] to "
+                + destination.getName() + " [" + destination.getIdentifier() + "] is violating the rule for backpressure settings. ";
+    }
+
+    private String getViolationMessage(final BackpressureViolationType backpressureViolationType, final String configured, final String limit) {
+        switch (backpressureViolationType) {
```
mosermw, line 210: This switch statement looks like a candidate for using enhanced switch and removing the default case that isn't needed for switching over an Enum.
```
return switch (backpressureViolationType) {
  case BP_COUNT_THRESHOLD_ABOVE_LIMIT ->
    "blah";
  case BP_COUNT_THRESHOLD_BELOW_LIMIT ->
    "blah";
  case BP_SIZE_THRESHOLD_ABOVE_LIMIT ->
    "blah";
  case BP_SIZE_THRESHOLD_BELOW_LIMIT ->
    "blah";
};
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
```diff
@@ -0,0 +1,242 @@
...
+        final long maxCount = context.getProperty(COUNT_MAX).asLong();
+        final double minSize = context.getProperty(SIZE_MIN).asDataSize(DataUnit.B);
+        final double maxSize = context.getProperty(SIZE_MAX).asDataSize(DataUnit.B);
+
+        // Map of all id/components to generate more human readable violations
+        final Map<String, VersionedComponent> idComponent = Stream.of(
+                    pg.getFunnels().stream(),
+                    pg.getProcessors().stream(),
+                    pg.getInputPorts().stream(),
+                    pg.getOutputPorts().stream()
```
mosermw, line 140: I haven't thought it through fully, but is there a reason pg.getProcessGroups() and pg.getRemoteProcessGroups() are not included here?

mosermw: After looking at this more (while building a new rule myself), do you need this map?  It seems that a source and destination ConnectableComponent, available from the VersionedConnection itself, has the component's id, name and type available.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
pvillard31: Correct - I chose this approach because of [NIFI-13953](https://issues.apache.org/jira/browse/NIFI-13953) as the "Go To" is a very important aspect when dealing with very large flows. I'll add a TODO and explanation in the code to revisit this.

pvillard31: My bad, it's been a bit too long since I did this. The reason why we want the violation to be on the processor when we have one (either as source or destination) is because in case the rule is "enforced" we want the corresponding component to be invalid.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
pvillard31: I don't have PGs and RPGs because connections are not connecting to PGs/RPGs but to the ports.

I started making the changes as you suggested (just use the ConnectableComponent objects) but the problem is with `GroupAnalysisResult.forComponent` that expects a VersionedComponent. After a quick look, I don't think we would want to change this logic. If we were only building the violations on the connection itself, what you suggest would work but if we can make a component invalid when the rule has an enforced policy, we should do it.

pvillard31:
Thanks for the review @mosermw ! I have addressed/answered your comments and rebased on main to have the latest checkstyle/PMD rules.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
mosermw: Thanks for the explanation @pvillard31, this makes sense.

While working on another FlowAnalysisRule, I may spend some time trying to optimize this to avoid creating this map until a violation is detected.  With many ProcessGroups to analyze, I would expect the vast majority to not have any violations.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
mosermw: Understood, thanks!

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictBackpressureSettings.java
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    private String getLocationMessage(final VersionedConnection connection, final VersionedComponent source, final VersionedComponent destination) {
+        if (source == null || destination == null) {
+            return "The connection [" + connection.getIdentifier() + "] is violating the rule for backpressure settings. ";
+        }
+        return "The connection [" + connection.getIdentifier() + "] connecting " + source.getName() + " [" + source.getIdentifier() + "] to "
+                + destination.getName() + " [" + destination.getIdentifier() + "] is violating the rule for backpressure settings. ";
+    }
```
mosermw, line 212: I feel like this can be improved, because there are cases where one source/destination exists and the other is null.  I'm thinking one of:
- "The connection [] is violating the rule"
- "The connection [] from source [] is violating the rule"
- "The connection [] to destination [] is violating the rule"
- "The connection [] from source [] to destination [] is violating the rule"

I have also found that when the source or destination is a Funnel and the other is a Processor, that the Funnel is referred to as "null", such as "The connection [] connecting UpdateAttribute [] to null [] is violating the rule".  We can probably detect that the VersionedComponent is a funnel (that has no name) and replace the name with the word "Funnel".

mosermw: If you would like, we can accept this PR as it is, and I can improve this message while I'm working on a new FlowAnalysisRule.

mosermw: @pvillard31 Not sure if you have seen this latest question.  I could make this improvement as I work on NIFI-14132 if you like.

pvillard31: Hey @mosermw - sorry for the late reply, I meant to answer your comment and got caught up with other developments. I'm not sure I can get to this during this week so if you're up to it, that would be greatly appreciated!

mosermw:
+1 merging.  Looking forward to using this rule and future Flow Analysis Rules.

pvillard31:
Thanks @mosermw !

PR #9477, opened 2024-10-31
NIFI-13955 - Filter out .directories in Git Registry clients

pvillard31:
# Summary

[NIFI-13955](https://issues.apache.org/jira/browse/NIFI-13955) - Filter out .directories in Git Registry clients

If a Git Registry Client (GitHub/GitLab) is configured without a repo path (looking at the root of the repo), we should filter out directories with names starting with a dot (as we may have directories like `.github`), otherwise they will be showing as potential buckets.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
@@ -60,6 +60,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


EndzeitBegins:
Thank you for the contribution and providing the motivation behind this change, @pvillard31. 

I think it makes a lot of sense to be able to exclude directories from being detected as buckets. 

Does it make sense to allow users to define a pattern of paths to exempt as buckets, instead of hard-coding to ignore any "hidden", dot-prefixed directories? 
The default value still could be `[.].*` but users would be able to exclude other directories as well, e.g. `ci` or `docs`.

pvillard31:
Hey @EndzeitBegins - thanks for the feedback! I don't think it's really needed here but I'm definitely not against adding a property if you feel like this is better (I just tried to keep configuration super easy). The reason why I don't think this is needed is because, if the repo is used only for flows, then it would look like:

````
- .github
- bucket1
- bucket2
````
But if the repo is going to be for more than just flows, then the registry client would be configured with a repo path (`flows`) and it would be something like:

````
- .github
- flows
  - bucket1
  - bucket2
- docs
````

Having said that... are you aware of CI/CD stacks that would create/expect a non hidden directory at the root level?

pvillard31:
I've not been using Jenkins in ages but if it expects a directory like that, I can add a property and we just save ourselves from another change in the future. Regarding potential files next to the flow definitions, I'm currently designing/working on some flow testing capabilities so I'll definitely catch any issue in case the clients are not happy with this scenario.

I'll update the current PR.

EndzeitBegins:
Changes look good to me. 👍🏻 Thank you for the adjustments, adding a test case and especially the feedback.

bbende (approved):
Looks good, thanks for the improvement!

PR #9480, opened 2024-11-01
[NIFI-13482] add filtered extension count

scottyaslan:
Screenshot 2024-11-01 at 12 52 53 PM
Screenshot 2024-11-01 at 12 53 15 PM


mcgilman:
Reviewing....

nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/extension-creation/extension-creation.component.html
```diff
@@ -15,23 +15,26 @@
...
-                    placeholder="Filter types..." />
-            </mat-form-field>
-        </div>
-    </div>
+    <h2 mat-dialog-title>Add {{ componentType }}</h2>
     <mat-dialog-content>
-        <div class="flex flex-col dialog-content">
+        <div class="flex flex-col dialog-content gap-y-2">
+            <div class="flex flex-col">
+                <div class="flex items-center gap-x-2">
```
mcgilman, line 22:
```suggestion
                <div class="flex items-center">
```

There is only a single child element so the `gap` isn't needed here.

nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/extension-creation/extension-creation.component.html
```diff
@@ -15,23 +15,26 @@
...
+                        <input
+                            matInput
+                            class="px-1.5 py-1 w-64"
+                            type="text"
+                            (keyup)="filterTypes($event)"
+                            tabindex="0"
+                            placeholder="Filter types..." />
+                    </mat-form-field>
+                </div>
+                <div class="tertiary-color font-medium mt-2">
```
mcgilman, line 34: Can we use `gap-y-2` on this elements parent instead of `mt-2` here?

mcgilman (commented):
Thanks @scottyaslan!  Just noted a few things below.

mcgilman (approved):

PR #9485, opened 2024-11-04
NIFI-13963 Ensured unknown fields are not included in outputted schema for 1.x.

dan-s1:
# Summary

[NIFI-13963](https://issues.apache.org/jira/browse/NIFI-13963)

This is a back port of  the fix made in #9347 for 2.x.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK ~21~ 8

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/RecordReader.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/RecordReader.java
@@ -47,7 +47,7 @@ public interface RecordReader extends Closeable {
      * @throws SchemaValidationException if a Record contains a field that violates the schema and cannot be coerced into the appropriate field type.
      */
     default Record nextRecord() throws IOException, MalformedRecordException {
-        return nextRecord(true, false);
+        return nextRecord(true, true);
     }
 
     /**
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/RecordReader.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/RecordReader.java
@@ -47,7 +47,7 @@ public interface RecordReader extends Closeable {
      * @throws SchemaValidationException if a Record contains a field that violates the schema and cannot be coerced into the appropriate field type.
      */
     default Record nextRecord() throws IOException, MalformedRecordException {
-        return nextRecord(true, true);
+        return nextRecord(true, false);
     }
 
     /**
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/MapRecord.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/MapRecord.java
@@ -514,9 +514,6 @@ public class MapRecord implements Record {
     private Optional<RecordField> setValueAndGetField(final String fieldName, final Object value) {
         final Optional<RecordField> field = getSchema().getField(fieldName);
         if (!field.isPresent()) {
-            if (dropUnknownFields) {
-                return field;
-            }
 
             final Object previousValue = values.put(fieldName, value);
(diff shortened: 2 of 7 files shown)


pvillard31 (commented):
As said on the JIRA, this can be considered as a breaking change and, as such, I don't think this should be applied on 1.x. I'm no longer a user of the 1.x branch so I won't be a -1 but if this change is merged, I believe this should be made very explicit to users.

exceptionfactory:
It seems worth correcting the behavior, but I question whether this is the right solution on the support branch. As @pvillard31, the default value of `nextRecord()` has remained unchanged, so it seems like we need a different solution for the support branch if this is to be corrected.

I'm not to the point of voting this down, but I'm leaning in that direction.

@dan-s1 or @mosermw, have you evaluated approaches other than this one to correct the behavior for the support branch?

dan-s1:
@exceptionfactory I had an idea to limit the scope of this change to only `JsonTreeReader` and I would like to have your feedback. My idea is to keep @pvillard31 changes in JsonTreeRowRecordReader.java for trimming the schema but change lines 129-133
from 
```
@Override
    protected Record convertJsonNodeToRecord(final JsonNode jsonNode, final RecordSchema schema, final boolean coerceTypes, final boolean dropUnknownFields)
            throws IOException, MalformedRecordException {
        return convertJsonNodeToRecord(jsonNode, schema, coerceTypes, dropUnknownFields, null);
    }
```
to
```
@Override
    protected Record convertJsonNodeToRecord(final JsonNode jsonNode, final RecordSchema schema, final boolean coerceTypes, final boolean dropUnknownFields)
            throws IOException, MalformedRecordException {
        return convertJsonNodeToRecord(jsonNode, schema, coerceTypes, !this.schemaInferred || dropUnknownFields, null);
    }
```
where `schemaInferred` is a boolean calculated in `JsonTreeReader` in method `storePropertyValues` with
       ` this.schemaInferred = INFER_SCHEMA.getValue().equals(context.getProperty(SCHEMA_ACCESS_STRATEGY).getValue());`

and it is passed to `JsonTreeRowRecordReader` by expanding its constructor to take this flag.

This would thereby ensure whenever a schema is specified unknown fields are not kept.
Please let me know your thoughts.

dan-s1:
Hmm this may not work as 3 tests in `TestJsonTreeRowRecordReader` whenever `reader.nextRecord(false, false)` is used are failing.

dan-s1:
@pvillard31 In PR #9347 you had suggested

> If we agree on the above, then it may be easier to juste override nextRecord() in AbstractJsonRowRecordReader and default to nextRecord(true, true).

and both @exceptionfactory and @markap14 agreed to that change. How come instead of that the change was made to

`nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/RecordReader.java` ?

```
default Record nextRecord() throws IOException, MalformedRecordException {
        return nextRecord(true, true);
    }
```

dan-s1:
@exceptionfactory I limited the scope of the changes as the original suggestion in #9347 to `AbstractJsonRowRecordReader`

exceptionfactory:
Thanks for evaluating different options @dan-s1, this is helpful.

I think limiting the scope of the `nextRecord()` changes to the `AbstractJsonRowRecordReader` could provide an optimal solution.

I will follow up again soon and look for additional input from @pvillard31 on this approach.

dan-s1:
@exceptionfactory @pvillard31 I am following up on this PR to see if this fix can get into the 1.28.1 release as this is a fix my team needs. Thanks!

exceptionfactory (approved):
Thanks for making the adjustments @dan-s1, the latest change scoping `nextRecord(true, true)` to the `AbstractJsonRowRecordReader` looks like a good approach. +1 merging to the support 1.x and 1.28 branches.

PR #9494, opened 2024-11-06
NIFI-13960: Using the ExtensionCreation component when adding a Regis…

mcgilman:
…try Client to better handle different bundles.


rfellows (approved):
Changes look good. I like how adding registry clients feels consistent with how we add other component types. 👍 @mcgilman

PR #9495, opened 2024-11-06
NIFI-13969: Rerendering labels after the viewport is restored to ensu…

mcgilman:
…re that calculations are performed after the transform is applied.


scottyaslan:
Reviewing...

scottyaslan (approved):
LGTM

PR #9511, opened 2024-11-11
NIFI-13988 Refactored number handling to account for a blank string and treat it as a null when the inferred type is a number.

dan-s1:
# Summary

[NIFI-13988](https://issues.apache.org/jira/browse/NIFI-13988)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
@@ -1203,7 +1203,7 @@ public class DataTypeUtils {
             }
             case String string -> {
                 try {
-                    return new BigInteger(string);
+                    return isBlank(string) ? null : new BigInteger(string);
                 } catch (NumberFormatException nfe) {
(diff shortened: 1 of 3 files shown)


nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
```diff
@@ -2078,4 +2078,8 @@ public static boolean isFittingNumberType(final Object value, final RecordFieldT
 
         return false;
     }
+
+    private static boolean isBlank(String string) {
+        return string == null || string.isBlank();
```
exceptionfactory, line 2083: Based on the usage, the `string == null` check appears to be unnecessary. With that, it looks like this method could be removed and `string.isBlank()` could be used directly.

exceptionfactory (changes requested):
Thanks for implementing this improvement @dan-s1. The general approach of treating an empty string as `null` for numbers looks good.

Based on the usage, it looks like the `string == null` check is unnecessary since it should never be null based on the existing uses.

PR #9514, opened 2024-11-13
[NIFI-14002] - Quick copy of IDs

rfellows:
[NIFI-14002](https://issues.apache.org/jira/browse/NIFI-14002)

Added a new directive `copy` that just takes in a string. It adds the copy icon at the end of the element, but only displayed when it is hovered. Click it to copy to the clipboard.

Example usage to copy an id (assuming `id` is a variable):
```
<div [copy]="id">{{ id }}</div>
```

Kapture 2024-11-13 at 08 37 43


scottyaslan (approved):
LGTM

PR #9516, opened 2024-11-13
NIFI-14005 - Support Custom Content from External File in GenerateFlowFile

pvillard31:
# Summary

[NIFI-14005](https://issues.apache.org/jira/browse/NIFI-14005) - Support Custom Content from External File in GenerateFlowFile

Add a property in GenerateFlowFile allowing the user to specify a local path to an external file. If specified the content of the external file will be used as content of the generated flow files.

This makes things easier compared to a GetFile for example. Besides, coupled with the Asset feature, it makes it easy to load a test dataset as content for GenerateFlowFile processor in a cloud native environment (container based).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
@@ -29,6 +29,8 @@ import org.apache.nifi.annotation.lifecycle.OnScheduled;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.ValidationContext;
 import org.apache.nifi.components.ValidationResult;
+import org.apache.nifi.components.resource.ResourceCardinality;
+import org.apache.nifi.components.resource.ResourceType;
 import org.apache.nifi.expression.AttributeExpression;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
@@ -41,8 +43,10 @@ import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.io.OutputStreamCallback;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.apache.nifi.scheduling.SchedulingStrategy;
+import org.apache.nifi.stream.io.StreamUtils;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -73,6 +77,7 @@ public class GenerateFlowFile extends AbstractProcessor {
 
     public static final String DATA_FORMAT_BINARY = "Binary";
     public static final String DATA_FORMAT_TEXT = "Text";
+    public static final String DATA_FORMAT_EXTERNAL_FILE = "External File";
 
     public static final PropertyDescriptor FILE_SIZE = new PropertyDescriptor.Builder()
             .name("File Size")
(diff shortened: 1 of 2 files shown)


exceptionfactory (commented):
Thanks for putting together this improvement @pvillard31.

This change has implications for the permissions required to use the `GenerateFlowFile` Processor, as other Processors with access to the file system require the `READ_FILESYSTEM` permission. Although it is easy to add, it may have implications for existing flow designs. What do you think?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
```diff
@@ -183,12 +200,22 @@ protected Collection<ValidationResult> customValidate(ValidationContext validati
...
 
         if (isCustom && (isUnique || !isText)) {
             results.add(new ValidationResult.Builder().subject("Custom Text").valid(false).explanation("If Custom Text is set, then Data Format must be "
                     + "text and Unique FlowFiles must be false.").build());
         }
+        if (isCustomContent && (isUnique || !isExternal)) {
+            results.add(new ValidationResult.Builder().subject("Custom Content").valid(false).explanation("If Custom Content is set, then Data Format must be "
+                    + "External File and Unique FlowFiles must be false.").build());
+        }
+        if (isCustomContent && isCustom) {
```
EndzeitBegins, line 215: Do we need this validation explicitly? The `dependsOn` conditionals should result in only on of `Custom Text` or `Custom Content` being available to the user I assume?

pvillard31:
That is an extremely valid point that I completely missed. Given that some users definitely want to restrict processors with access to local file system, it feels like we shouldn't change expectations around GenerateFlowFile and a dedicated processor for this specific usage would be better. What do you think?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
```diff
@@ -111,8 +116,19 @@ public class GenerateFlowFile extends AbstractProcessor {
...
             .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
+            .dependsOn(DATA_FORMAT, DATA_FORMAT_TEXT)
+            .dependsOn(UNIQUE_FLOWFILES, "false")
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
+    public static final PropertyDescriptor CUSTOM_CONTENT = new PropertyDescriptor.Builder()
+            .name("Custom Content")
+            .description("Path to a file. If specified, this file will be used as content of the generated FlowFiles and the File Size will be ignored.")
+            .required(false)
+            .identifiesExternalResource(ResourceCardinality.SINGLE, ResourceType.FILE, ResourceType.URL, ResourceType.TEXT)
```
EndzeitBegins, line 127: Doesn't this warrant declaring the processor to require `READ_FILESYSTEM` permission? Is that a problem in regards to backwards compatibility?

EndzeitBegins (commented):
Thank you for working on this Pierre.

This improvement will help with both manual and automated testing and allow to inject larger and non-text payloads more easily.

The overall changes look good to me. I've just left some remarks in regards to naming and declaration.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
EndzeitBegins: Looks like @exceptionfactory beat me to it.

exceptionfactory:
> That is an extremely valid point that I completely missed. Given that some users definitely want to restrict processors with access to local file system, it feels like we shouldn't change expectations around GenerateFlowFile and a dedicated processor for this specific usage would be better. What do you think?

Yes, under the circumstances, it seems like a separate Processor would be better. I think it would be good to scope out what it looks like on the Jira issue and then move to consider the implementation.

EndzeitBegins:
> a dedicated processor for this specific usage would be better. What do you think?

Thinking about it, a separate processor might make sense insofar that the processor is not actually "generating" content but rather injecting pre-defined content when using the contents of a file (although the same could be said about the existing "Custom Text")

PR #9519, opened 2024-11-14
NIFI-14005 - GetFileResource processor

pvillard31:
# Summary

[NIFI-14005](https://issues.apache.org/jira/browse/NIFI-14005) - GetFileResource processor

The goal of this new GetFileResource processor is to provide a way for a user to inject a File Resource as a FlowFile with custom attributes. It means that the processor would be able to specify a path to a file and its content would be used as FlowFile's content. By leveraging the asset feature, it makes it easy to load a test dataset as a FlowFile in a cloud native environment (container based).

Alternatives that have been considered:
- GenerateFlowFile with a "Custom File/Content" property. However, a processor being able to access the local file system needs to be associated with a specific set of permissions and that would be a significant breaking change for all users already using GenerateFlowFile processor where such permissions are enforced.
- GetFile with some specific improvements. However, longer term we would likely want to completely remove this processor and only have ListFile/FetchFile. Beside, its configuration is overly complex for the intended goal here.

Testing:
- tested with local file reference
- tested with file referenced via URL
- tested using Asset feature and Parameter reference

For information:

````
$ ./bin/cli.sh nifi create-asset -p nifi-cli.properties -pcid dc39db55-1eb2-3d26-9a8a-caaf097cfcc2 -af /Users/pierre/dev/myFile.txt

b099141f-2bc4-3073-8dc0-243dac470f08

$ ./bin/cli.sh nifi add-asset-reference -p nifi-cli.properties -aid b099141f-2bc4-3073-8dc0-243dac470f08 -pcid dc39db55-1eb2-3d26-9a8a-caaf097cfcc2 -pn MyAsset
````

The parameter `#{MyAsset}` can then be used in the processor (assuming proper Parameter Context binding on the Process Group).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
@@ -38,40 +38,37 @@ import org.apache.nifi.processor.AbstractProcessor;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
-import org.apache.nifi.processor.io.OutputStreamCallback;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.apache.nifi.scheduling.SchedulingStrategy;
 import org.apache.nifi.stream.io.StreamUtils;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-@Tags({"test", "file", "generate", "load"})
+@Tags({ "test", "file", "generate", "load" })
 @InputRequirement(Requirement.INPUT_FORBIDDEN)
-@CapabilityDescription("This processor creates FlowFiles with the content of the configured File Resource. GetFileResource "
-        + "is useful for load testing, configuration, and simulation.")
-@DynamicProperty(name = "Generated FlowFile attribute name", value = "Generated FlowFile attribute value",
-        expressionLanguageScope = ExpressionLanguageScope.ENVIRONMENT,
-        description = "Specifies an attribute on generated FlowFiles defined by the Dynamic Property's key and value." +
-        " If Expression Language is used, evaluation will be performed only once per batch of generated FlowFiles.")
-@WritesAttributes({
-        @WritesAttribute(attribute = "mime.type", description = "Sets the MIME type of the output if the 'Mime Type' property is set"),
-})
+@CapabilityDescription("""
+        This processor creates FlowFiles with the content of the configured File Resource. GetFileResource
+        is useful for load testing, configuration, and simulation.
+        """)
+@DynamicProperty(
+        name = "Generated FlowFile attribute name", value = "Generated FlowFile attribute value", expressionLanguageScope = ExpressionLanguageScope.ENVIRONMENT,
+        description = "Specifies an attribute on generated FlowFiles defined by the Dynamic Property's key and value."
+)
+@WritesAttributes(
+    {
+            @WritesAttribute(attribute = "mime.type", description = "Sets the MIME type of the output if the 'MIME Type' property is set"),
+    }
+)
 @DefaultSchedule(strategy = SchedulingStrategy.TIMER_DRIVEN, period = "1 min")
 @Restricted(
         restrictions = {
-                @Restriction(
-                        requiredPermission = RequiredPermission.READ_FILESYSTEM,
-                        explanation = "Provides operator the ability to read from any file that NiFi has access to."),
-                @Restriction(
-                        requiredPermission = RequiredPermission.REFERENCE_REMOTE_RESOURCES,
-                        explanation = "File Resource can reference resources over HTTP/HTTPS"
-                )
+                @Restriction(requiredPermission = RequiredPermission.READ_FILESYSTEM, explanation = "Provides operator the ability to read from any file that NiFi has access to."),
+                @Restriction(requiredPermission = RequiredPermission.REFERENCE_REMOTE_RESOURCES, explanation = "File Resource can reference resources over HTTP/HTTPS")
         }
 )
 public class GetFileResource extends AbstractProcessor {
@@ -84,8 +81,8 @@ public class GetFileResource extends AbstractProcessor {
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
     public static final PropertyDescriptor MIME_TYPE = new PropertyDescriptor.Builder()
-            .name("Mime Type")
-            .description("Specifies the value to set for the \"mime.type\" attribute.")
+            .name("MIME Type")
+            .description("Specifies the value to set for the [mime.type] attribute.")
             .required(false)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
@@ -122,12 +119,12 @@ public class GetFileResource extends AbstractProcessor {
 
     @Override
     public void onTrigger(final ProcessContext context, final ProcessSession session) {
-        Map<PropertyDescriptor, String> processorProperties = context.getProperties();
-        Map<String, String> generatedAttributes = new HashMap<String, String>();
+        final Map<PropertyDescriptor, String> processorProperties = context.getProperties();
+        final Map<String, String> generatedAttributes = new HashMap<String, String>();
         for (final Map.Entry<PropertyDescriptor, String> entry : processorProperties.entrySet()) {
-            PropertyDescriptor property = entry.getKey();
-            if (property.isDynamic() && property.isExpressionLanguageSupported()) {
-                String dynamicValue = context.getProperty(property).evaluateAttributeExpressions().getValue();
+            final PropertyDescriptor property = entry.getKey();
+            if (property.isDynamic()) {
+                final String dynamicValue = context.getProperty(property).evaluateAttributeExpressions().getValue();
                 generatedAttributes.put(property.getName(), dynamicValue);
             }
         }
@@ -139,14 +136,9 @@ public class GetFileResource extends AbstractProcessor {
         FlowFile flowFile = session.create();
 
         try (final InputStream inputStream = context.getProperty(FILE_RESOURCE).asResource().read()) {
-            flowFile = session.write(flowFile, new OutputStreamCallback() {
-                @Override
-                public void process(final OutputStream out) throws IOException {
-                    StreamUtils.copy(inputStream, out);
-                }
-            });
+            flowFile = session.write(flowFile, out -> StreamUtils.copy(inputStream, out));
         } catch (IOException e) {
-            getLogger().error("Could not create FlowFile with specified Custom Content", e);
+            getLogger().error("Could not create FlowFile from Resource [{}]", context.getProperty(FILE_RESOURCE).getValue(), e);
         }
 
         flowFile = session.putAllAttributes(flowFile, generatedAttributes);
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
@@ -62,6 +62,7 @@ import java.util.Set;
 @WritesAttributes(
     {
             @WritesAttribute(attribute = "mime.type", description = "Sets the MIME type of the output if the 'MIME Type' property is set"),
+            @WritesAttribute(attribute = "Dynamic property key", description = "Value for the corresponding dynamic property, if any is set")
     }
 )
 @DefaultSchedule(strategy = SchedulingStrategy.TIMER_DRIVEN, period = "1 min")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestGetFileResource.java [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+public class TestGetFileResource {
+
+    @Test
+    public void testGetFileResource() throws IOException {
+        final TestRunner runner = TestRunners.newTestRunner(new GetFileResource());
+
+        runner.setProperty(GetFileResource.FILE_RESOURCE, "");
+        runner.assertNotValid();
+
+        runner.setProperty(GetFileResource.FILE_RESOURCE, "src/test/resources/TestCountText/jabberwocky.txt");
```
exceptionfactory, line 40: Recommend declaring the resource path as a variable to avoid repeating it in the property value and the assertion.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.Set;
+
+@Tags({"test", "file", "generate", "load"})
+@InputRequirement(Requirement.INPUT_FORBIDDEN)
+@CapabilityDescription("This processor creates FlowFiles with the content of the configured File Resource. GetFileResource "
+        + "is useful for load testing, configuration, and simulation.")
+@DynamicProperty(name = "Generated FlowFile attribute name", value = "Generated FlowFile attribute value",
+        expressionLanguageScope = ExpressionLanguageScope.ENVIRONMENT,
+        description = "Specifies an attribute on generated FlowFiles defined by the Dynamic Property's key and value." +
+        " If Expression Language is used, evaluation will be performed only once per batch of generated FlowFiles.")
```
exceptionfactory, line 61: Instead of string concatenation, this could use a multiline string.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+@Tags({"test", "file", "generate", "load"})
+@InputRequirement(Requirement.INPUT_FORBIDDEN)
+@CapabilityDescription("This processor creates FlowFiles with the content of the configured File Resource. GetFileResource "
+        + "is useful for load testing, configuration, and simulation.")
```
exceptionfactory, line 57: This could use a multiline string.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+@DynamicProperty(name = "Generated FlowFile attribute name", value = "Generated FlowFile attribute value",
+        expressionLanguageScope = ExpressionLanguageScope.ENVIRONMENT,
+        description = "Specifies an attribute on generated FlowFiles defined by the Dynamic Property's key and value." +
+        " If Expression Language is used, evaluation will be performed only once per batch of generated FlowFiles.")
+@WritesAttributes({
+        @WritesAttribute(attribute = "mime.type", description = "Sets the MIME type of the output if the 'Mime Type' property is set"),
```
exceptionfactory, line 63:
```suggestion
        @WritesAttribute(attribute = "mime.type", description = "Sets the MIME type of the output if the 'MIME Type' property is set"),
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+
+    public static final PropertyDescriptor FILE_RESOURCE = new PropertyDescriptor.Builder()
+            .name("File Resource")
+            .description("Location of the File Resource (Local File or URL). This file will be used as content of the generated FlowFiles.")
+            .required(true)
+            .identifiesExternalResource(ResourceCardinality.SINGLE, ResourceType.FILE, ResourceType.URL)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .build();
+    public static final PropertyDescriptor MIME_TYPE = new PropertyDescriptor.Builder()
+            .name("Mime Type")
```
exceptionfactory, line 87: Recommand putting `MIME` in all caps since it is an acronym.
```suggestion
            .name("MIME Type")
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+            .description("Specifies the value to set for the \"mime.type\" attribute.")
```
exceptionfactory, line 88: Recommend using square brackets or single quotes to avoid having to escape the double quote character.
```suggestion
            .description("Specifies the value to set for the [mime.type] attribute.")
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Override
+    public Set<Relationship> getRelationships() {
+        return RELATIONSHIPS;
+    }
+
+    @Override
+    public void onTrigger(final ProcessContext context, final ProcessSession session) {
+        Map<PropertyDescriptor, String> processorProperties = context.getProperties();
+        Map<String, String> generatedAttributes = new HashMap<String, String>();
```
exceptionfactory, line 126:
```suggestion
        final Map<String, String> generatedAttributes = new HashMap<>();
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @Override
+    public Set<Relationship> getRelationships() {
+        return RELATIONSHIPS;
+    }
+
+    @Override
+    public void onTrigger(final ProcessContext context, final ProcessSession session) {
+        Map<PropertyDescriptor, String> processorProperties = context.getProperties();
```
exceptionfactory, line 125:
```suggestion
        final Map<PropertyDescriptor, String> processorProperties = context.getProperties();
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+        Map<String, String> generatedAttributes = new HashMap<String, String>();
+        for (final Map.Entry<PropertyDescriptor, String> entry : processorProperties.entrySet()) {
+            PropertyDescriptor property = entry.getKey();
+            if (property.isDynamic() && property.isExpressionLanguageSupported()) {
```
exceptionfactory, line 129: Is the `isExpressionLanguageSupported()` check needed?
```suggestion
            if (property.isDynamic()) {
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+
+        FlowFile flowFile = session.create();
+
+        try (final InputStream inputStream = context.getProperty(FILE_RESOURCE).asResource().read()) {
+            flowFile = session.write(flowFile, new OutputStreamCallback() {
+                @Override
+                public void process(final OutputStream out) throws IOException {
+                    StreamUtils.copy(inputStream, out);
+                }
+            });
```
exceptionfactory, line 147: It looks like this could be simplified using a lambda expression.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,156 @@
...
+        } catch (IOException e) {
+            getLogger().error("Could not create FlowFile with specified Custom Content", e);
```
exceptionfactory, line 149: Recommend declaring the Resource as a variable so that the location can be referenced in this error.
```suggestion
            getLogger().error("Could not create FlowFile from Resource [{}]", fileResource, e);
```

exceptionfactory (changes requested):
Thanks for putting this together @pvillard31. It looks like a good approach in general, I noted a handful of minor adjustments.

pvillard31:
Thanks for the review @exceptionfactory - I pushed a commit to address your comments

pvillard31:
@EndzeitBegins - you're right that people could use dynamic properties (assuming they read the docs and see that they can use dynamic properties). The fact that `mime.type` is "special" is that this attribute is used to automatically choose the right content viewer when looking at the content of the flow file in the UI. This is why this one is made available as a dedicated property in GenerateFlowFile and I took the same approach here.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,148 @@
...
+@InputRequirement(Requirement.INPUT_FORBIDDEN)
+@CapabilityDescription("""
+        This processor creates FlowFiles with the content of the configured File Resource. GetFileResource
+        is useful for load testing, configuration, and simulation.
+        """)
+@DynamicProperty(
+        name = "Generated FlowFile attribute name", value = "Generated FlowFile attribute value", expressionLanguageScope = ExpressionLanguageScope.ENVIRONMENT,
+        description = "Specifies an attribute on generated FlowFiles defined by the Dynamic Property's key and value."
+)
+@WritesAttributes(
```
EndzeitBegins, line 62: Does it make sense to add an `@WritesAttribute` for the dynamic property support here as well?

I know that's not declared in `GenerateFlowFile` either, but I assume that way, users might more easily recognise that specifying additional attributes by adding properties is possible.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GetFileResource.java
```diff
@@ -0,0 +1,148 @@
...
+    @Override
+    protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+        return PROPERTIES;
+    }
+
+    @Override
+    protected PropertyDescriptor getSupportedDynamicPropertyDescriptor(final String propertyDescriptorName) {
+        return new PropertyDescriptor.Builder()
+                .name(propertyDescriptorName)
+                .required(false)
```
EndzeitBegins, line 107: Is it helpful to add a dynamic description here?
Something along the lines of:

> Adds the attribute [NAME] with the properties value as the attribute's value

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestGetFileResource.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Unit tests for the GetFileResource processor.
+ */
+public class TestGetFileResource {
+
+    @Test
+    public void testGetFileResource() throws IOException {
```
EndzeitBegins, line 34: What do you think about splitting up this test into multiple test cases?
I know there are a lot of tests in the NiFi codebase in this style but I'm my opinion, they document intent rather poorly. 

From what I've gathered you basically have the following test cases:
- isInvalidWhenFileResourceIsNotSet
- isValidWhenFileResourceIsSet
- createsAFlowFileWithTheFileResourceAsItsContent
- supportsSpecifyingMimeType
- supportsSpecifyingAdditionalAttributes

EndzeitBegins (commented):
Thank you for clarifying your intent @pvillard31. 

I was looking at this processor from a different angle. 
When thinking about injecting FlowFiles for downstream processors, the FlowFile attributes and content should align with what's expected when they would get there regularly, outside of testing. In that case, `mime.type` is just one attribute among many, depending on the position the processor is used inside the Flow. 

Having said that, I understand where you're coming from as well and thus I'm fine with keeping the additional property.  

I've added some remarks that might be worth looking into, but nothing that hinders a merge from my point of view. 

Thank you for adding this useful functionality. 👍🏻

pvillard31:
Thanks @EndzeitBegins - I pushed a commit with your recommendations

EndzeitBegins (approved):
LGTM 👍🏻

PR #9523, opened 2024-11-14
Nifi 14000

HunterG6700:
# Summary

[NIFI-14000](https://issues.apache.org/jira/browse/NIFI-14000)
Updated the process group so that when the 'With external services' option is selected, '_with_services' is appended to the downloaded .json filename.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (commented):
Thanks for proposing this change @HunterG6700, however, I'm not sure it is warranted. 

Is it often that you save flow definitions both with and without services? As you noted in the Jira issue, the original file is not overwritten, although I agree the standard browser naming that appends a `(1)` or similar is not helpful for differentiating. files. That being said, this introduces a more concrete behavior in naming, which may not be optimal. It seems that the suffix could just as easily be applied to "without services".

For this reason, I'm inclined to recommend leaving the behavior as it stands. However, if you have more background on why this is important as a general behavior pattern, that could be worth considering.

PR #9526, opened 2024-11-15
NIFI-13682 - Versioning breaks in case of style change on labels

pvillard31:
# Summary

[NIFI-13682](https://issues.apache.org/jira/browse/NIFI-13682) - Versioning breaks in case of style change on labels

Style for labels is a map which can have up to two keys: `font-size` and `background-color`. If between two versions, this map has a different key set, the current code creates an issue where we would always block any version change / any revert local changes because the current code merges the changes of the two versions and it will show as local changes that we'll never be able to revert. This PR changes the code to not merge the styles, but strictly apply the style of the desired version.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (approved):
Thanks for tracking down this one @pvillard31, looks like it dates back to the initial version of the project! +1 merging

PR #9529, opened 2024-11-15
[NIFI-13918][NIFI-14015] - Flow analysis unauthorized root group does not display group id in the PG details section

rfellows:
[NIFI-13918](https://issues.apache.org/jira/browse/NIFI-13918) & [NIFI-14015](https://issues.apache.org/jira/browse/NIFI-14015)


scottyaslan (approved):
LGTM! Verified that the console error is no longer present and the UUID is displayed when the user does not have read permissions.

PR #9535, opened 2024-11-18
NIFI-13976: Adding endpoints to support copying and pasting flow cont…

mcgilman:
…ents.

- Including the revision for the current Process Group in the flow response.
- Introduce a method propose built for adding new versioned components to an existing Process Group.
- Updating the paste response payload to only include the entities for the components that were just added.
- Always generating new IDs when copying components.
- Creating an ID for the copy action.
- Code clean up.
- Mapping IDs on paste to ensure connections are created correctly.
- Authorizing any components whose instance IDs are specified.
- Copying any sensitive properties from copied instances when pasting.
- Fixing mapping of copy response to ensure that nested Controller Services can be referenced.
- Including external Controller Services in copy response to they can be resolved on paste.
- Excluding unreferenced services from copy response.
- Including property descriptors to allow external services to resolve when pasting.
- Including Parameters and Parameter Providers in copy response.
- Fixing issue returning unreferenced services in the current process group.
- Reverting changes building up external service references.
- Only including external services that are references by the copy selection.
- Adding some additional test coverage.
- Adding consideration for the current ids in the paste id mapping.
- Adding an endpoint response merger for the paste endpoint.
- Including version control information in copy response.
- Being more lenient when pasting unknown registry clients.
- Ensuring groups IDs are mapped correctly when and when not versioned.
- Including registry client id in vci of versioned process groups.
- Updating copy/paste logic in existing system tests to leverage new endpoints.
- Restoring originally proposed port name if possible.
- Allow ports to be pasted into the root group.
- Fixing authorization issue when pasting a flow containing Parameter Providers.
- Fixing issue when authorizing exporting a Flow that contains a Process Group bound to a Parameter Context.
- Introducing some Copy/Paste system tests.
- Adding system tests for verify various copy paste functionality.
- Fix checkstyle issue.
- Fixing checkstyle issues.
- Ensuring appropriate access around Parameter Providers following resolution based on locally available Parameter Providers.
- Ensuring user has access to controller services directly referenced.
- Fixing NPE in log message of Parameter Context.
- Adding more copy/paste system tests.
- Moving copy/paste version tests into the CopyPasteIT.
- Fixing unit test mocking.
- Include service authorization based on versioned component id since it may resolve during synchronization.
- Skipping properties with no values when resolving service identifiers.

This PR can be tested with the corresponding front end changes available in this PR [1].

[1] https://github.com/apache/nifi/pull/9536

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-client-dto/src/main/java/org/apache/nifi/web/api/entity/CopyResponseEntity.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-client-dto/src/main/java/org/apache/nifi/web/api/entity/CopyResponseEntity.java
@@ -1,0 +1,220 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.web.api.entity;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.xml.bind.annotation.XmlType;
+import org.apache.nifi.flow.ExternalControllerServiceReference;
+import org.apache.nifi.flow.ParameterProviderReference;
+import org.apache.nifi.flow.VersionedConnection;
+import org.apache.nifi.flow.VersionedFunnel;
+import org.apache.nifi.flow.VersionedLabel;
+import org.apache.nifi.flow.VersionedParameterContext;
+import org.apache.nifi.flow.VersionedPort;
+import org.apache.nifi.flow.VersionedProcessGroup;
+import org.apache.nifi.flow.VersionedProcessor;
+import org.apache.nifi.flow.VersionedRemoteProcessGroup;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A response to copy a portion of the flow.
+ */
+@XmlType(name = "copyResponseEntity")
+public class CopyResponseEntity extends Entity {
+
+    private String id;
+
+    private Map<String, ExternalControllerServiceReference> externalControllerServiceReferences;
+    private Map<String, VersionedParameterContext> parameterContexts;
+    private Map<String, ParameterProviderReference> parameterProviders;
+
+    private Set<VersionedProcessGroup> processGroups = new HashSet<>();
+    private Set<VersionedRemoteProcessGroup> remoteProcessGroups = new HashSet<>();
+    private Set<VersionedProcessor> processors = new HashSet<>();
+    private Set<VersionedPort> inputPorts = new HashSet<>();
+    private Set<VersionedPort> outputPorts = new HashSet<>();
+    private Set<VersionedConnection> connections = new HashSet<>();
+    private Set<VersionedLabel> labels = new HashSet<>();
+    private Set<VersionedFunnel> funnels = new HashSet<>();
+
+    /**
+     * The id for this copy action.
+     *
+     * @return The id
+     */
+    @Schema(description = "The id for this copy action."
+    )
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    /**
+     * The external controller service references.
+     *
+     * @return The external controller service reference
+     */
+    @Schema(description = "The external controller service references."
+    )
+    public Map<String, ExternalControllerServiceReference> getExternalControllerServiceReferences() {
+        return externalControllerServiceReferences;
+    }
+
+    public void setExternalControllerServiceReferences(Map<String, ExternalControllerServiceReference> externalControllerServiceReferences) {
+        this.externalControllerServiceReferences = externalControllerServiceReferences;
+    }
+
+    /**
+     * The referenced parameter contexts.
+     *
+     * @return The referenced parameter contexts
+     */
+    @Schema(description = "The referenced parameter contexts."
+    )
+    public Map<String, VersionedParameterContext> getParameterContexts() {
+        return parameterContexts;
+    }
+
+    public void setParameterContexts(Map<String, VersionedParameterContext> parameterContexts) {
+        this.parameterContexts = parameterContexts;
+    }
+
+    /**
+     * The referenced parameter providers.
+     *
+     * @return The referenced parameter providers
+     */
+    @Schema(description = "The referenced parameter providers."
+    )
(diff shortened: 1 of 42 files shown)


nifi-framework-bundle/nifi-framework/nifi-client-dto/src/main/java/org/apache/nifi/web/api/entity/CopyResponseEntity.java
```diff
@@ -0,0 +1,220 @@
...
+ * A response to copy a portion of the flow.
+ */
+@XmlType(name = "copyResponseEntity")
+public class CopyResponseEntity extends Entity {
+
+    private String id;
+
+    private Map<String, ExternalControllerServiceReference> externalControllerServiceReferences;
+    private Map<String, VersionedParameterContext> parameterContexts;
+    private Map<String, ParameterProviderReference> parameterProviders;
```
exceptionfactory, line 46: Is there a reason for declaring default `HashSet` values below but not declaring default `HashMap` values for these properties?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/ProcessGroupAuthorizable.java
```diff
@@ -32,13 +33,28 @@ public interface ProcessGroupAuthorizable extends AuthorizableHolder {
...
      * @return all encapsulated processors
      */
     Set<ComponentAuthorizable> getEncapsulatedProcessors();
 
+    /**
+     * The authorizables for all encapsulated processors that meet the specified predicate. Non null
+     *
+     * @return all encapsulated processors
+     */
+    Set<ComponentAuthorizable> getEncapsulatedProcessors(Predicate<org.apache.nifi.authorization.resource.ComponentAuthorizable> filter);
```
exceptionfactory, line 56: Since the method has a comment, it would be helpful to add a description for the `filter` parameter.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/ProcessGroupAuthorizable.java
```diff
@@ -32,13 +33,28 @@ public interface ProcessGroupAuthorizable extends AuthorizableHolder {
      */
     ProcessGroup getProcessGroup();
 
+    /**
+     * Returns the Parameter Context Authorizable. May be null if the underlying Process Group is not
+     * bound to a Parameter Context.
+     *
+     * @return the Parameter Context authorizable
+     */
+    Authorizable getParameterContextAuthorizable();
```
exceptionfactory, line 42: If this method can return `null`, would it be better to return it in an `Optional` wrapper?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/ProcessGroupAuthorizable.java
```diff
@@ -89,10 +105,17 @@ public interface ProcessGroupAuthorizable extends AuthorizableHolder {
...
+     * @return all encapsulated controller services
      */
     Set<ComponentAuthorizable> getEncapsulatedControllerServices();
 
+    /**
+     * The authorizables for all encapsulated controller services that meet the specified predicate. Non null
+     *
+     * @return all encapsulated controller services
+     */
+    Set<ComponentAuthorizable> getEncapsulatedControllerServices(Predicate<org.apache.nifi.authorization.resource.ComponentAuthorizable> filter);
```
exceptionfactory, line 119: See note above on `filter` parameter.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessGroupResource.java
```diff
@@ -326,6 +336,85 @@ public Response exportProcessGroup(
...
 
+    /**
+     * Generates a copy response for the given copy request.
+     *
+     * @param groupId The id of the process group
+     * @param copyRequestEntity The copy request
+     * @return A copyResponseEntity.
+     */
+    @POST
+    @Consumes(MediaType.WILDCARD)
```
exceptionfactory, line 347: Should this MediaType be `APPLICATION_JSON`?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessGroupResource.java
```diff
@@ -326,6 +336,85 @@ public Response exportProcessGroup(
...
+    public Response copy(
+            @Parameter(
+                    description = "The process group id.",
+                    required = true
+            )
+            @PathParam("id") final String groupId,
+            @Parameter(
+                    description = "The request including the components to be copied from the specified Process Group.",
+                    required = true
+            ) final CopyRequestEntity copyRequestEntity) {
```
exceptionfactory, line 375: Does this `copyRequestEntity` need to be null-checked?

exceptionfactory (commented):
Thanks for the detailed work on the support for copy-paste @mcgilman. I ran through an initial code review and noted a few minor questions and comments.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessGroupResource.java
```diff
@@ -2836,6 +2950,331 @@ public Response importProcessGroup(
...
+                            final VersionedPropertyDescriptor propertyDescriptor = propertyDescriptors.get(propertyEntry.getKey());
+                            if (propertyDescriptor != null && propertyDescriptor.getIdentifiesControllerService()) {
+                                return serviceIdMapping.containsKey(propertyEntry.getValue());
+                            }
+                        }
+
+                        return false;
+                    })
+                    .findFirst()
+                    .ifPresent(serviceEntry -> serviceEntry.setValue(serviceIdMapping.get(serviceEntry.getValue())));
```
markap14, line 3133: I think we want to use `.forEach` here instead of `.findFirst().ifPresent()` because we want to do this for all properties, right?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessGroupResource.java
```diff
@@ -2836,6 +2950,331 @@ public Response importProcessGroup(
...
```
markap14, line 3153: And a `.forEach` here as well

markap14:
Thanks @mcgilman I ran a bunch of tests, copying & pasting locally as well as copying from 1 instance and pasting to another. Tested simple processors/connections. Tested child groups. Tested with Controller Services. Tested with Parameter Contexts and references. Tested sensitive properties were removed, but not those that reference parameters. Tested Controller Service A referencing Controller Service B. Tested Processors with multiple Controller Services. Tested references to external Controller Services when (a) the external services doesn't exist in the destination instance (b) the external service does exist in the destination instance.

Everything worked flawlessly. Code review looks good to me as well. I'm a +1! Many thanks for adding this feature, I'm very excited to be able to do this!

markap14 (approved):

PR #9539, opened 2024-11-19
NIFI-14031: Ensuring all Parameter Contexts are added when Editing a …

mcgilman:
…Process Group. For contexts the user lacks access to will be added as disabled options.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/property-table/editors/combo-editor/combo-editor.component.spec.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/property-table/editors/combo-editor/combo-editor.component.spec.ts
@@ -115,6 +115,7 @@ describe('ComboEditor', () => {
             deleted: false,
             added: false,
             dirty: false,
+            savedValue: 'flowfile-attribute',
             type: 'required'
         };
     });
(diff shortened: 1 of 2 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/process-group/edit-process-group/edit-process-group.component.ts
```diff
@@ -71,6 +67,12 @@ export class EditProcessGroup extends TabbedDialog {
                     value: parameterContext.id,
                     description: parameterContext.component.description
                 });
+            } else {
+                this.parameterContextsOptions.push({
+                    text: parameterContext.id,
+                    value: parameterContext.id,
+                    disabled: true
+                });
```
rfellows, line 75: If the user doesn't have read access to the currently set parameter context we should not disable it as a selectable option in the dropdown. If they accidentally change the context in the dropdown, they have no way of setting it back to what it was before without cancelling the dialog and potentially losing other changes made in the dialog.

rfellows (changes requested):
There are unit test failures introduced with the latest changes. Other than that, this is looking good.

rfellows (approved):
Thanks, @mcgilman. This was an issue that was a bit more complicated that it first seemed but I think you came up with a solid solution that will make sense to users.

Will merge when CI is green.

PR #9541, opened 2024-11-19
[NIFI-14032] - Format Core Load Average in Cluster system table to 2 decimal places.

rfellows:
[NIFI-14032](https://issues.apache.org/jira/browse/NIFI-14032)


mcgilman:
Reviewing...

PR #9543, opened 2024-11-19
[NIFI-14033] - Disable links to referencing components while saving.

rfellows:
[NIFI-14033](https://issues.apache.org/jira/browse/NIFI-14033)


mcgilman (approved):
Thanks @rfellows!

PR #9547, opened 2024-11-20
NIFI-14029 - BitBucket Registry Client

pvillard31:
# Summary

[NIFI-14029](https://issues.apache.org/jira/browse/NIFI-14029) - BitBucket Registry Client

Adding an implementation of the Registry Client to integrate with BitBucket. For reference, APIs are described here:

https://developer.atlassian.com/cloud/bitbucket/rest/intro

In addition to the implementation itself, two bugs are fixed:

- NPE in case the initialization fails for some reason. In this case, we already called `clientInitialized.set(true)` and this caused to always have the directory exclusion filter being `null`. This is now fixed.
- NPE in the `deregisterFlow` method. This happens when we start version control and for some reason, something fails after the file has been committed. In that case, we delete the content in the repo. Because we didn't populate the snapshot metadata, this was causing a NPE.

````
java.lang.NullPointerException: Cannot invoke "org.apache.nifi.registry.flow.RegisteredFlowSnapshotMetadata.setBucketIdentifier(String)" because "snapshotMetadata" is null
	at org.apache.nifi.registry.flow.git.AbstractGitFlowRegistryClient.updateBucketReferences(AbstractGitFlowRegistryClient.java:548)
	at org.apache.nifi.registry.flow.git.AbstractGitFlowRegistryClient.deregisterFlow(AbstractGitFlowRegistryClient.java:255)
````

Regarding the implementation itself. Three authentication methods are supported:

- Basic Auth with Username and App Password
- OAuth 2.0 - I tested the Client Credentials Flow with OAuth Consumers. Note that the endpoint to configure in the controller service for OAuth is `https://bitbucket.org/site/oauth2/access_token`. It is important to also note that, when using the Client Credentials flow, the permissions/scopes set at the OAuth Consumer level in the BitBucket Admin UI are ignored and it will inherit the permissions of the **owner** of the OAuth Consumer. This is quite misleading...
- Access Token - I have not tested this option as this is a paid option of BitBucket.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### CODE REVIEW

File: nifi-extension-bundles/nifi-atlassian-bundle/nifi-atlassian-extensions/src/main/java/org/apache/nifi/atlassian/bitbucket/BitBucketFlowRegistryClient.java [hunk 1]
```diff
@@ -0,0 +1,163 @@
...
+    public boolean isStorageLocationApplicable(FlowRegistryClientConfigurationContext context, String location) {
+        // TODO Auto-generated method stub
+        return false;
+    }
+
+    @Override
+    protected String getStorageLocation(GitRepositoryClient repositoryClient) {
+        // TODO Auto-generated method stub
+        return null;
+    }
```
bbende, line 162: Technically it shouldn't cause any problems to not implement these, but the pattern we used for GitHub and GitLab was to construct a string that is similar to the value you use when cloning the repo:
```
git@github.com:<owner>/<repo>.git
```
And then `isStorageLocationApplicable` checks if it starts with the prefix `git@github.com:`.

pvillard31: Thanks @bbende - I completely forgot to get back to this when I switched to the other class and testing... Will address now

mbegerau:
First of all thanks @pvillard31 for initiating this PR. We are starting to migrate some projects from NiFi 1 to NiFi 2 and the new git registry client feature is actually one of the top reasons why we want to migrate, because it really reduces the complexity of the setup. Also my company is locked in to Atlassian tech stack which is why I'm looking forward to the Bitbucket implementation.

That being said, I checked out your branch today and tried to set it up. Here are my notes:
1. Naming: Doesn't really matter but the second "b" of Bitbucket is usually lowercase and we should probably keep it like that.
2. The atlassian bundle is missing in the [pom.xml of the nifi-assembly bundle](nifi-assembly/pom.xml). Without adding the dependency there the BitbucketFlowRegistryClient did not show up with the usual _mvnw install..._, because the nar file was not installed.
3. The Bitbucket access token flow would be our preferred authentication solution and it's not working. You said you can't test it due to being a paid service. Maybe we can debug this together. The reason for now is that the [endpoint](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-repositories/#api-user-permissions-repositories-get) you use in the [checkRepoPermissions](https://github.com/pvillard31/nifi/blob/838e21cdbf0d2368c94e57797f728f7e2efff077/nifi-extension-bundles/nifi-atlassian-bundle/nifi-atlassian-extensions/src/main/java/org/apache/nifi/atlassian/bitbucket/BitBucketRepositoryClient.java#L395) method always returns _{"type": "error", "error": {"message": "Token is invalid, expired, or not supported for this endpoint."}}_ when used with an access token. From what I read in the [Access Token usage documentation](https://developer.atlassian.com/cloud/bitbucket/rest/intro/#access-tokens) access tokens can't have the "account" scope. So all endpoints requiring this scope will not work. For example the [/user endpoint](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-users/#api-user-get) does not work as well. Also it says "Access Tokens are not listed in any repository or workspace permission API response." So the [/permissions-config/users endpoint](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-repositories/#api-repositories-workspace-repo-slug-permissions-config-users-get) actually works with access token, but it does not return any results. You can check for repository (read) and repository:write permissions indirectly, [Get a repository](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-repositories/#api-repositories-workspace-repo-slug-get) returns status code 200 with read permissions (404 otherwise) and [Create a commit by uploading a file](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-source/#api-repositories-workspace-repo-slug-src-post) without a request body returns status code 415 with write permissions (404 otherwise).
4. When I switched to oauth with client credentials it works. I tested creating a new flow, commiting a new version and returning to an old version.

mbegerau:
I have tested with the latest commit and I can confirm that **creating flows**, **commiting changes** and **changing versions** works with all three authentication methods. As far as I know there is no way yet to delete commits or entire flows with a git registry client.

Maybe you should consider applying the access token logic (if read, then assume read+write) also to the username/appPassword flow because with the current setup you need to give the app password "account:read" permission (same reason as for access token), otherwise "repository" and "repository:write" would be enough.

Since there is no "View Documentation" option for registry clients, maybe it could be helpful for the user to extend the "Authentication Type" descriptions. Here some bullet point ideas:
- Basic Auth: Username (not email) and App Password (link to [documentation](https://support.atlassian.com/bitbucket-cloud/docs/app-passwords/)). List required permissions: repository, repository:read, (possibly account:read)
- Access Token: (link to [documentation](https://support.atlassian.com/bitbucket-cloud/docs/access-tokens/)). List required permissions: repository, repository:read
- OAuth: Only works with client credentials flow which requires OAuth consumers (link to [documentation](https://support.atlassian.com/bitbucket-cloud/docs/use-oauth-on-bitbucket-cloud/)). OAuth consumer needs to be private, and callback url needs to be set, but can have any value.

pvillard31:
Thanks @mbegerau - I have made some additional changes based on your feedback. I applied the same approach for checking permissions in case of basic auth.

mbegerau:
Great! No further comments from my side. I ran `mvn clean install` and tested everything from scratch and for me all registry functions are working with all three authentication types.

Let's hope it gets merged before 2.4.0 :crossed_fingers: 

I'm not sure if any of the maintainers are using Bitbucket cloud or have access to it...

PR #9561, opened 2024-12-02
NIFI-13710: Rebuilding Update Attribute Custom UI.

mcgilman:


scottyaslan (commented):
If the user enters a name the 'Add' button becomes enabled but if the user attempts to click the 'Add' button they will receive the snackbar error. Do you think that the 'Add' button should be disabled in this case?

Screenshot 2024-12-10 at 11 43 00 AM

scottyaslan:
We probably need to update the docs for this processor too:

Screenshot 2024-12-10 at 4 06 31 PM

scottyaslan:
Do we need to consider the case where the user does not have modify permissions and handle that such that the advanced UI becomes read only?

scottyaslan:
How does the user edit the rule? It took me a minute to figure out that I could click on the row and expand it to edit the rule.

Screenshot 2024-12-10 at 4 15 32 PM

scottyaslan:
Maybe we should discuss the UX for the invalid state of the form. It is confusing to the user when creating a new rule that there is already an existing empty expression and attributes. If the user clicks to add an expression while still having an 'empty' expression the user receives an error even after deleting the empty expression:

Kapture 2024-12-10 at 16 21 25

Does the user have to include an attribute and value? Can those be empty strings? Should the user delete those before the add the rule?

mcgilman:
> Does the user have to include an attribute and value? Can those be empty strings? Should the user delete those before the add the rule?

I cannot tell what's happening in your screen capture. However, I do notice that the conditions are incorrectly labeled `Expressions` which may be causing confusion here. Let me update that and try to reproduce what is being shown here.

mcgilman:
> If the user enters a name the 'Add' button becomes enabled but if the user attempts to click the 'Add' button they will receive the snackbar error. Do you think that the 'Add' button should be disabled in this case?

The front end cannot fully validate the input. It's relies on server side logic where the expressions are parsed using the antlr grammar. We can introduce additional validation on the front end to ensure that the condition expression is not empty.

mcgilman:
> How does the user edit the rule? It took me a minute to figure out that I could click on the row and expand it to edit the rule.

What do you propose? I'm open to suggestions.

mcgilman:
> Do we need to consider the case where the user does not have modify permissions and handle that such that the advanced UI becomes read only?

These should be handled. Please let me know if you ran into a situation where the UI was read only but still allowed modification.

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.scss
```diff
@@ -0,0 +1,28 @@
...
+ * under the License.
+ */
+
+:host {
+    width: 100%;
+}
+
+.rule-listing {
+    .cdk-drag-disabled {
+        font-style: normal;
```
scottyaslan, line 26: Is this needed? What is the purpose?

scottyaslan (commented):
I spent a little more time reviewing this tonight. Overall great progress but a few things to work through.

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/action-table/action-table.component.html
```diff
@@ -0,0 +1,150 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<div class="action-table flex flex-col h-full gap-y-2">
+    <div class="flex justify-end items-center">
+        @if (!isDisabled) {
+            <button mat-icon-button class="primary-icon-button" type="button" (click)="newActionClicked()">
```
scottyaslan, line 21: Try adding 7 or 8 of these. Then save it. Some get saved and some do not. Do we want to allow the user to save 'empty' attribute/value pairs?

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/condition-table/condition-table.component.html
```diff
@@ -0,0 +1,111 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<div class="condition-table flex flex-col h-full gap-y-2">
+    <div class="flex justify-end items-center">
+        @if (!isDisabled) {
+            <button mat-icon-button class="primary-icon-button" type="button" (click)="newConditionClicked()">
```
scottyaslan, line 21: Try to create a new rule. There will be an initial 'empty' expression added for the user. Do not add any value and close the editor. Click the button to add another condition. Enter a valid value for this condition. Now try to save. It will fail. Now delete the 'empty' condition. Attempt to save again. Still you will get an error.

mcgilman:
@scottyaslan Thanks for the review... I believe I've addressed the comments here in my latest commit(s).

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.scss
mcgilman: Yes it is because there is an existing `font-style` applied to `cdk-drag-disabled` globally that results in affecting the edit rule for inside the expansion panel. The other instances of `cdk-drag-disable` is a simple listing and not applied to an expansion panel.

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.html
```diff
@@ -0,0 +1,187 @@
...
+                                [id]="rule.id"
+                                [name]="rule.name"
+                                [existingRuleNames]="getExistingRuleNames()"
+                                [comments]="rule.comments"
+                                [conditions]="rule.conditions"
+                                [actions]="rule.actions"
+                                [editable]="false"></edit-rule>
+                        </mat-expansion-panel>
+                    }
+                </mat-accordion>
```
scottyaslan, line 110: The drag and drop styles still don't match here and I still think we should not try to combine the expansion panel UX with the drag/drop UX. 

```suggestion
                <div
                    class="cdk-list cdk-drag-selected border"
                    cdkDropList
                    [cdkDropListDisabled]="!isEditable"
                    [cdkDropListData]="rulesList"
                    (cdkDropListDropped)="reorderRules($event)">
                    @for (rule of filteredRulesList; track rule.id) {
                        <div
                            class="border mat-elevation-z2 m-1 font-bold"
                            cdkDragPreviewContainer="parent"
                            cdkDrag>
                            <div class="flex w-full items-center">
                                <span class="grip -mt-1 pr-5"></span>
                                <div class="truncate" [title]="rule.name">
                                    {{ rule.name }}
                                </div>
                                @if (rule.comments) {
                                    <i
                                        class="pl-2 pr-2 fa fa-comment neutral-color"
                                        nifiTooltip
                                        [tooltipComponentType]="TextTip"
                                        [tooltipInputData]="rule.comments"></i>
                                }
                            </div>
                        </div>
                    }
                </div>
```

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.scss
```diff
@@ -0,0 +1,56 @@
...
+    }
+
```
scottyaslan, line 28: You will also need these styles in order to match the look for the drag/drop UX from prioritizers and parameter context inheritance.

```suggestion

.cdk-drag {
        height: 38px;
        padding-left: 6px;
        display: flex;
        flex-direction: row;
        align-items: center;
        justify-content: space-between;
        box-sizing: border-box;
    }
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/edit-parameter-dialog/edit-parameter-dialog.component.ts
```diff
@@ -17,7 +17,7 @@
 
 import { Component, EventEmitter, Inject, Input, Output } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogModule } from '@angular/material/dialog';
-import { EditParameterRequest, EditParameterResponse, Parameter } from '../../../state/shared';
+import { EditParameterRequest, EditParameterResponse } from '../../../state/shared';
```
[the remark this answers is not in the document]
mcgilman: No. `EditParameterRequest` and `EditParameterResponse` are not in `@nifi/shared`.

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.html
```diff
@@ -0,0 +1,175 @@
...
+            <div class="flex flex-col gap-y-1">
+                <div class="flex gap-x-1 items-center">
+                    <div>Use original FlowFile for matching rules</div>
+                    <form [formGroup]="flowFilePolicyForm">
+                        <mat-slide-toggle
+                            formControlName="useOriginalFlowFilePolicy"
+                            (change)="flowFilePolicyToggled($event)"></mat-slide-toggle>
+                    </form>
+                </div>
+                <div class="text-sm">If disabled rules will be executed using a clone of the original FlowFile</div>
```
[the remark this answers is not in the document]
mcgilman: This is incorrect. If disabled, ordering doesn't matter because each matching rule will use a clone of the original FlowFile. Ordering only matters when executing using the original FlowFile.

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.html
```diff
@@ -0,0 +1,187 @@
...
+                                    <div class="flex justify-between items-center">
+                                        <div class="flex-1 truncate">
+                                            {{ rule.comments }}
+                                        </div>
+                                        @if (isEditable) {
+                                            <button
+                                                mat-icon-button
+                                                type="button"
+                                                [matMenuTriggerFor]="actionMenu"
+                                                (click)="$event.stopPropagation()"
```
scottyaslan, line 132: If the user is using the keyboard and tabs to the mat-icon-button and presses the space or the enter key the menu should open not the expansion panel.

```suggestion
                                                (click)="$event.stopPropagation()"
                                                (keydown)="$event.stopPropagation()"
```

scottyaslan (approved):
Thanks for all the hard work on this one @mcgilman !!!

nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.scss
scottyaslan: It has been removed int the latest. Thanks!

PR #9566, opened 2024-12-04
NIFI-13016 Added groups mapping from OIDC token claim for Registry

hazmat345:
# Summary

[NIFI-13016](https://issues.apache.org/jira/browse/NIFI-13016)

This PR adds support for respecting groups returned by an external OIDC token to nifi-registry.

### History
The support was added to nifi proper in [NIFI-7823](https://issues.apache.org/jira/browse/NIFI-7823). I've tried to make the implementation in nifi-registry as similar as possible. However, a major overhaul to OIDC was done as part of [NIFI-4890](https://issues.apache.org/jira/browse/NIFI-4890), and those changes were not ported to nifi-registry.

### Questions / Thoughts
- I'm using Spring `SimpleGrantedAuthority` because that's what the current implementation of nifi uses, but as far as I can tell there's no real reason to use it - I'm just packing the group name into it and pulling it out later. It'd probably be cleaner to just pass around the group names.
- I haven't added any unit tests yet - as this is my first PR I'd like to get confirmation that the implementation direction makes sense before I add tests.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

- [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/nifi-registry-core/nifi-registry-security-api/src/main/java/org/apache/nifi/registry/security/authorization/UserGroupProvider.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-security-api/src/main/java/org/apache/nifi/registry/security/authorization/UserGroupProvider.java
@@ -76,6 +76,30 @@ public interface UserGroupProvider {
     Group getGroup(String identifier) throws AuthorizationAccessException;
 
     /**
+     * Retrieves a Group by name.
+     *
+     * @param name the name of the group to retrieve
+     * @return the Group with the given name, or null if no matching group was found
+     * @throws AuthorizationAccessException if there was an unexpected error performing the operation
+     */
+    default Group getGroupByName(String name) throws AuthorizationAccessException {
+        final Set<Group> allGroups = getGroups();
+        if (allGroups == null) {
+            return null;
+        }
+
+        Group matchingGroup = null;
+        for (Group group : allGroups) {
+            if (group.getName().equals(name)) {
+                matchingGroup = group;
+                break;
+            }
+        }
+
+        return matchingGroup;
+    }
+
+    /**
      * Gets a user and their groups. Must be non null. If the user is not known the UserAndGroups.getUser() and
      * UserAndGroups.getGroups() should return null
      *
--- a/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtIdentityProvider.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtIdentityProvider.java
@@ -34,6 +34,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 @Component
@@ -70,7 +71,8 @@ public class JwtIdentityProvider extends BearerAuthIdentityProvider implements I
 
         try {
             final String jwtPrincipal = jwtService.getUserIdentityFromToken(jwtAuthToken);
-            return new AuthenticationResponse(jwtPrincipal, jwtPrincipal, expiration, issuer);
+            final Set<String> groups = jwtService.getUserGroupsFromToken(jwtAuthToken);
+            return new AuthenticationResponse(jwtPrincipal, jwtPrincipal, expiration, issuer, groups);
         } catch (JwtException e) {
             throw new InvalidAuthenticationException(e.getMessage(), e);
         }
--- a/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
@@ -34,11 +34,13 @@ import org.apache.nifi.registry.security.key.Key;
 import org.apache.nifi.registry.security.key.KeyService;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.GrantedAuthority;
 import org.springframework.stereotype.Service;
 
 import java.nio.charset.StandardCharsets;
-import java.util.Calendar;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 @Service
 public class JwtService {
@@ -48,6 +50,7 @@ public class JwtService {
     private static final MacAlgorithm SIGNATURE_ALGORITHM = Jwts.SIG.HS256;
     private static final String KEY_ID_CLAIM = "kid";
     private static final String USERNAME_CLAIM = "preferred_username";
+    private static final String GROUPS_CLAIM = "groups";
 
     private final KeyService keyService;
 
@@ -76,6 +79,27 @@ public class JwtService {
             }
             return jws.getPayload().getSubject();
         } catch (JwtException e) {
+            final String errorMessage = "There was an error validating the JWT";
+            logger.error(errorMessage, e);
+            throw e;
+        }
+    }
+
+    public Set<String> getUserGroupsFromToken(final String base64EncodedToken) throws JwtException {
+        // The library representations of the JWT should be kept internal to this service.
+        try {
+            final Jws<Claims> jws = parseTokenFromBase64EncodedString(base64EncodedToken);
+
+            if (jws == null) {
+                throw new JwtException("Unable to parse token");
+            }
+
+            @SuppressWarnings("unchecked")
+            ArrayList<String> groupsString = jws.getPayload().get(GROUPS_CLAIM, ArrayList.class);
+
+            return new HashSet<>(groupsString);
+        } catch (JwtException e) {
+            logger.debug("The Base64 encoded JWT: " + base64EncodedToken);
             final String errorMessage = "There was an error validating the JWT";
             logger.error(errorMessage, e);
             throw e;
@@ -125,10 +149,15 @@ public class JwtService {
                 authenticationResponse.getUsername(),
                 authenticationResponse.getIssuer(),
                 authenticationResponse.getIssuer(),
-                authenticationResponse.getExpiration());
+                authenticationResponse.getExpiration(),
+                null);
     }
 
     public String generateSignedToken(String identity, String preferredUsername, String issuer, String audience, long expirationMillis) throws JwtException {
+        return this.generateSignedToken(identity, preferredUsername, issuer, audience, expirationMillis, null);
+    }
+
+    public String generateSignedToken(String identity, String preferredUsername, String issuer, String audience, long expirationMillis, Collection<? extends GrantedAuthority> authorities) throws JwtException {
 
         if (identity == null || StringUtils.isEmpty(identity)) {
             String errorMessage = "Cannot generate a JWT for a token with an empty identity";
@@ -155,6 +184,7 @@ public class JwtService {
                     .audience().add(audience).and()
                     .claim(USERNAME_CLAIM, preferredUsername)
                     .claim(KEY_ID_CLAIM, key.getId())
+                    .claim(GROUPS_CLAIM, authorities.stream().map(GrantedAuthority::getAuthority).collect(Collectors.toSet()))
                     .issuedAt(now.getTime())
                     .expiration(expiration.getTime())
                     .signWith(Keys.hmacShaKeyFor(keyBytes), SIGNATURE_ALGORITHM).compact();
--- a/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/oidc/StandardOidcIdentityProvider.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/oidc/StandardOidcIdentityProvider.java
@@ -21,9 +21,11 @@ import java.net.URI;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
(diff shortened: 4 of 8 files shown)


nifi-registry/nifi-registry-core/nifi-registry-security-api/src/main/java/org/apache/nifi/registry/security/authorization/UserGroupProvider.java
```diff
@@ -75,6 +75,30 @@ public interface UserGroupProvider {
...
     Group getGroup(String identifier) throws AuthorizationAccessException;
 
+    /**
+     * Retrieves a Group by name.
+     *
+     * @param name the name of the group to retrieve
+     * @return the Group with the given name, or null if no matching group was found
+     * @throws AuthorizationAccessException if there was an unexpected error performing the operation
+     */
+    default Group getGroupByName(String name) throws AuthorizationAccessException {
```
exceptionfactory, line 85: As this method appears to be used only in the `StandardManagedAuthorizer`, it does not seem like a good candidate for adding to the `UserGroupProvider` interface right now.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtIdentityProvider.java
```diff
@@ -70,7 +71,8 @@ public AuthenticationResponse authenticate(AuthenticationRequest authenticationR
 
         try {
             final String jwtPrincipal = jwtService.getUserIdentityFromToken(jwtAuthToken);
-            return new AuthenticationResponse(jwtPrincipal, jwtPrincipal, expiration, issuer);
+            final Set<String> groups = jwtService.getUserGroupsFromToken(jwtAuthToken);
```
exceptionfactory, line 74: This approach requires parsing the authentication token string twice, which is not very efficient. It probably requires some refactoring to do this once, and then extract the claims required.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
```diff
@@ -34,11 +34,13 @@
 import org.apache.nifi.registry.security.key.KeyService;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.GrantedAuthority;
 import org.springframework.stereotype.Service;
 
 import java.nio.charset.StandardCharsets;
-import java.util.Calendar;
+import java.util.*;
```
exceptionfactory, line 41: Coding conventions require explicit imports, so this change should be reverted.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
```diff
@@ -82,6 +85,27 @@ public String getUserIdentityFromToken(final String base64EncodedToken) throws J
         }
     }
 
+    public Set<String> getUserGroupsFromToken(final String base64EncodedToken) throws JwtException {
```
exceptionfactory, line 88: As mentioned on the calling reference, decoding, parsing, and verifying the token more than once is a concern because it happens on every HTTP request. Evaluating an alternative approach that parses and verifies the token, returning a structured set of claims, would be a better path forward.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
```diff
@@ -82,6 +85,27 @@ public String getUserIdentityFromToken(final String base64EncodedToken) throws J
...
+            if (jws == null) {
+                throw new JwtException("Unable to parse token");
+            }
+
+            @SuppressWarnings("unchecked")
+            ArrayList<String> groupsString = jws.getPayload().get(GROUPS_CLAIM, ArrayList.class);
+
+            return new HashSet<>(groupsString);
+        } catch (JwtException e) {
+            logger.debug("The Base64 encoded JWT: " + base64EncodedToken);
```
exceptionfactory, line 102: The token itself should never be logged because it provides access to the REST API, so this line should be removed.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
```diff
@@ -82,6 +85,27 @@ public String getUserIdentityFromToken(final String base64EncodedToken) throws J
...
+            final String errorMessage = "There was an error validating the JWT";
+            logger.error(errorMessage, e);
+            throw e;
```
exceptionfactory, line 105: Logging the exception as an error here should be unnecessary. If it is necessary to convey additional details, throwing a new exception that has the `JwtException` as a cause could work.

exceptionfactory (changes requested):
Thanks for your work on this improvement @hazmat345.

As you noted, NiFi itself has significantly refactored the OIDC implementation using Spring Security, making it more difficult to implement additional functionality for Registry. The changes proposed seem reasonably scoped to consider on their own for now.

I noted a few things in detailed comments. The primary issue is duplicative parsing of the encoded JWT. That reflects some of the shortcomings of the current implementation for NiFi Registry. If you are able to evaluate and propose a refactored approach that avoids multiple rounds of token parsing to retrieve the necessary user and group claims, that should provide a way forward.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
hazmat345: Sorry, IntelliJ insists on reverting this every time I update it. Fixed.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
hazmat345: 👍🏼

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
hazmat345: Makes sense. I copied the exception handling from the `getUserIdentityFromToken` method above, so I'll change this in both places.

nifi-registry/nifi-registry-core/nifi-registry-security-api/src/main/java/org/apache/nifi/registry/security/authorization/UserGroupProvider.java
hazmat345: Ahh, right. That was a changed I pulled from the nifi major OIDC rewrite but I agree it doesn't make sense to do here at this point. I've removed this method from the interface and moved the logic into the `StandardManagedAuthorizer`.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtIdentityProvider.java
hazmat345: Refactored to improve the efficiency of token parsing / validation. I did change the method signature of a couple public methods - if you'd rather me not do that just let me know and I'll just add new methods instead of replacing the existing ones.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
hazmat345: Refactored to improve the efficiency of token parsing / validation. I did change the method signature of a couple public methods - if you'd rather me not do that just let me know and I'll just add new methods instead of replacing the existing ones.

hazmat345:
@exceptionfactory thanks for the review! I believe I've address all of the feedback and this is ready for another round of review.

hazmat345:
@exceptionfactory would you mind giving this another look? Thanks!

exceptionfactory:
Thanks @hazmat345, I plan to follow up on this soon.

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtIdentityProvider.java
```diff
@@ -61,16 +64,19 @@ public AuthenticationResponse authenticate(AuthenticationRequest authenticationR
...
-
         if (credentials == null) {
             logger.info("JWT not found in authenticationRequest credentials, returning null.");
             return null;
         }
 
         try {
-            final String jwtPrincipal = jwtService.getUserIdentityFromToken(jwtAuthToken);
-            return new AuthenticationResponse(jwtPrincipal, jwtPrincipal, expiration, issuer);
+            String jwtAuthToken = credentials instanceof String ? (String) credentials : null;
```
exceptionfactory, line 73: With the null check for `credentials` on line 67, it looks like the intermediate `jwtAuthToken` can be declared as `credentials.toString()`

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
```diff
@@ -74,14 +79,24 @@ public String getUserIdentityFromToken(final String base64EncodedToken) throws J
...
         }
     }
 
+    public String getUserIdentityFromToken(final Jws<Claims> jws) throws JwtException {
+        return jws.getPayload().getSubject();
+    }
+
+    public Set<String> getUserGroupsFromToken(final Jws<Claims> jws) throws JwtException {
+        @SuppressWarnings("unchecked")
+        ArrayList<String> groupsString = jws.getPayload().get(GROUPS_CLAIM, ArrayList.class);
```
exceptionfactory, line 95: This should be declared as `List` instead of `ArrayList`:
```suggestion
        final List<String> groupsString = jws.getPayload().get(GROUPS_CLAIM, ArrayList.class);
```

exceptionfactory: It looks like this could return null, so the groups need to be null-checked before being passed to `new HashSet<>()`

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/oidc/StandardOidcIdentityProvider.java
```diff
@@ -401,6 +401,10 @@ private String convertOIDCTokenToNiFiToken(OIDCTokenResponse response) throws Ba
         String identityClaim = properties.getOidcClaimIdentifyingUser();
         String identity = claimsSet.getStringClaim(identityClaim);
 
+        // Attempt to extract groups from the configured claim; default is 'groups'
+        String groupsClaim = properties.getOidcClaimGroups();
+        List<String> groups = claimsSet.getStringListClaim(groupsClaim);
```
exceptionfactory, line 406:
```suggestion
        final String groupsClaim = properties.getOidcClaimGroups();
        final List<String> groups = claimsSet.getStringListClaim(groupsClaim);
```

exceptionfactory (changes requested):
Thanks for your patience on this @hazmat345, it looks close to completion. I noted a few more minor adjustments related to null handling, and then this should be ready to go.

One additional change, the new property should be added to the OpenID Connect section of the Admin Guide for Registry.

hazmat345:
@exceptionfactory thanks for the review! I think I've addressed everything but please let me know if there's anything else.

I realized that the property wasn't being added to nifi-registry.properties - I went ahead an added it but noticed there are a couple more properties that are listed in the admin guide but aren't in nifi-registry.properties file. Maybe a follow-up PR?

nifi-registry/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/jwt/JwtService.java
```diff
@@ -155,6 +176,7 @@ public String generateSignedToken(String identity, String preferredUsername, Str
                     .audience().add(audience).and()
                     .claim(USERNAME_CLAIM, preferredUsername)
                     .claim(KEY_ID_CLAIM, key.getId())
+                    .claim(GROUPS_CLAIM, groups)
```
exceptionfactory, line 179: This `groups` claim value can be null in some cases, such as authentication with LDAP. Instead of adding a `groups` claim with a `null` value, either an empty list should be set, or the claim should be omitted. As this is the JWT that NiFi Registry generates, it seems better to set an empty list when the input `groups` parameter is null.

exceptionfactory (changes requested):
Thanks for the updates @hazmat345, the property additions and documentation update look good.

Additional other properties could be addressed in a separate pull request.

I noticed one remaining place where the `groups` collection could be `null` when generating the application token.

exceptionfactory (approved):
Thanks for working through all of the feedback @hazmat345! I verified the latest version works as expected with and without the groups claim coming from an OIDC provider, looks good! +1 merging

hazmat345:
> Thanks for working through all of the feedback @hazmat345! I verified the latest version works as expected with and without the groups claim coming from an OIDC provider, looks good! +1 merging

Absolutely! Thank you for all the reviews and guidance. Feels good to make my first contribution! 🚀

joewitt:
> > Thanks for working through all of the feedback @hazmat345! I verified the latest version works as expected with and without the groups claim coming from an OIDC provider, looks good! +1 merging
> 
> Absolutely! Thank you for all the reviews and guidance. Feels good to make my first contribution! 🚀

Congrats Matt - very cool!

exceptionfactory:
> > Thanks for working through all of the feedback @hazmat345! I verified the latest version works as expected with and without the groups claim coming from an OIDC provider, looks good! +1 merging
> 
> Absolutely! Thank you for all the reviews and guidance. Feels good to make my first contribution! 🚀

Always great to have a committed contributor! You picked a challenging first issue, so thanks again for collaborating to get it to completion!

SamirZebbouche:
Hello,

Have you tried a property other than the default one? In my NiFi Registry configuration, I use the following setting:
```
nifi.registry.security.user.oidc.claim.groups=roles
```
However, it doesn't seem to be working.

Thank you.

PR #9569, opened 2024-12-06
NIFI-14066: Add Encrypt/Decrypt RecordPath functions

mattyb149:
# Summary

[NIFI-14066](https://issues.apache.org/jira/browse/NIFI-14066) This PR adds encrypt() and decrypt() RecordPath functions

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record-path/pom.xml
+++ b/nifi-commons/nifi-record-path/pom.xml
@@ -132,5 +132,11 @@
             <version>2.1.0-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-property-encryptor</artifactId>
+            <version>2.1.0-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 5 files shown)


mattyb149:
I still have to add the doc to the RecordPath guide but wanted to get some 👀 on it while I do that

exceptionfactory (changes requested):
Thanks for proposing this new feature @mattyb149.

Unfortunately I'm going to vote -1 for now, based on anything using the internal NiFi Property Encryptor.

I recommend closing the PR so we can work through the implementation details first in Jira.

Summarizing the issues for now, the Property Encryptor is something internal to NiFi, it is not intended for external transmission. As such, anything encrypted by these functions would not be readable outside of NiFi, rendering it less than useful.

In addition, we should avoid exposing these algorithm names as parameters. The values are holdovers from NiFi 1, and although the implementations are reasonable for internal use, we should not build what amounts to an encryption protocol on internal NiFi algorithm names. Anything new should be redesigned for compatibility and avoidance of custom algorithm names, preferably avoiding algorithm selection completely in favor of abstracted protocol versions.

Beyond the compatibility concerns, passing the key as a Record Path function also presents serious security concerns from a flow design perspective. It might be possible to mitigate this through Parameter references, but that should be enforced.

Glad to work through these details in the Jira issue, but given the importance of anything related to encryption, this may also be worth considering as a NiFi Improvement Proposal. For now, working through the details in Jira would be a good place to start.

PR #9583, opened 2024-12-17
NIFI-14082 Refactored the rest of the Processors to be uniform when creating properties and relationships

dan-s1:
# Summary

[NIFI-14082](https://issues.apache.org/jira/browse/NIFI-14082)
This PR focuses on a uniform manner of creating the list of PropertyDescriptor objects and Set of relationships each processor may have. In addition Intellij suggestions were made more or less to the various unit tests which exercised the changed processors and in some cases to the processors themselves.
Where  possible renamed all list of PropertyDescriptor to PROPERTIES and the set of RelationShip to RELATIONSHIPS ensured they were both `private static final` and ensured all properties and relationships are on individual lines for increased readability.
All processors should now either use `List.of` or `Stream.concat` (where concatenation is needed between multiple lists) for declaring their list of property descriptor and `Set.of` when declaring their relationships.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/GetMongoRecord.java
+++ b/nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/GetMongoRecord.java
@@ -50,7 +50,6 @@ import java.io.OutputStream;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Stream;
 
 @CapabilityDescription("A record-based version of GetMongo that uses the Record writers to write the MongoDB result set.")
 @Tags({"mongo", "mongodb", "get", "fetch", "record", "json"})
@@ -77,7 +76,7 @@ public class GetMongoRecord extends AbstractMongoQueryProcessor {
         .required(true)
         .build();
 
-    private static final List<PropertyDescriptor> PROPERTIES = Stream.of(
+    private static final List<PropertyDescriptor> PROPERTIES = List.of(
             CLIENT_SERVICE,
             WRITER_FACTORY,
             DATABASE_NAME,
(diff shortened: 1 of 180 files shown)


nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-processors/src/main/java/org/apache/nifi/processors/mongodb/GetMongoRecord.java
```diff
@@ -79,35 +77,29 @@ public class GetMongoRecord extends AbstractMongoQueryProcessor {
...
-
-        DESCRIPTORS = Collections.unmodifiableList(_temp);
-
-        Set<Relationship> _rels = new HashSet<>();
-        _rels.add(REL_SUCCESS);
-        _rels.add(REL_FAILURE);
-        _rels.add(REL_ORIGINAL);
-        RELATIONSHIPS = Collections.unmodifiableSet(_rels);
-    }
+    private static final List<PropertyDescriptor> PROPERTIES = Stream.of(
```
mosermw, line 80: Hi @dan-s1 I'm curious why you sometimes chose to use Stream.of() instead of List.of()?  I understand using Stream when an argument is a collection itself, but this instance and some others appear to be better suited to List.of().

dan-s1: @mosermw Excellent question. The `List.of() `method is overloaded and it seems the max number of arguments it can take is 10. Hence where there are more than 10 properties I has to use `Stream.of`.

joewitt: Yeah i'd say unless we're able to move from current practice to one clean model that offers the same semantics with cleaner code then it isn't worth changing.

exceptionfactory (approved):
Thanks for the cleanup and standardization @dan-s1. This helps bring some of the especially older approaches in line with current collection `of()` methods. The latest version looks good! +1 merging

PR #9584, opened 2024-12-17
NIFI-14090: Removing unused code now that the resize handle has been …

mcgilman:
…removed from the status history table.

- Remove references to computedStyleMap which is not supported in Firefox.


scottyaslan (commented):
Reviewing...

scottyaslan (approved):
Thanks @mcgilman! I was able to observe the error in the status history chart running in the latest firefox 133.0.3. After applying this PR the console error is resolved and the status history chart displays as expected. +1

PR #9589, opened 2024-12-20
NIFI-14100: Add option to sequential download of dependencies for Pyt…

Lehel44:
…hon processors

# Summary

[NIFI-14100](https://issues.apache.org/jira/browse/NIFI-14100)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
@Lehel44 Thanks for highlighting this issue, but I don't think the proposed approach is a good way forward.

This is not something an administrator should have to configure, and we should avoid introducing new configuration properties if we can implement better behavior in general.

There are inherent limitations with automatic dependency downloads for Python, which is one of the reasons for introducing Python-based NAR bundling. That should be the solution for resource-constrained environments.

We may be able to improve download behavior by replacing `pip` with `uv`, although that is a system requirement to consider.

As it stands, we should further consider the options before pursuing the current approach.

exceptionfactory:
@Lehel44, following up from an offline discussion, and based on the concerns with the current approach, I'm going to close this pull request for now. We should continue the discussion on the Jira issue and then determine the best way forward.

PR #9595, opened 2024-12-25
NIFI-14110 Support to limit content size in PackageFlowFile

EndzeitBegins:
Due to using a different API to retrieve the FlowFiles the behaviour when working with multiple queues is no longer unspecified.

I had an circular dependency problem when depending on `nifi-mock` from `nifi-utils`, which is why I use an anonymous implementation of `FlowFile` inside the tests instead of `MockFlowFile`.

# Summary

[NIFI-14110](https://issues.apache.org/jira/browse/NIFI-14110)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/processor/util/FlowFileFilters.java
+++ b/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/processor/util/FlowFileFilters.java
@@ -43,22 +43,20 @@ public class FlowFileFilters {
 
             @Override
             public FlowFileFilterResult filter(final FlowFile flowFile) {
-                if (count == 0) {
-                    count++;
-                    size += flowFile.getSize();
+                count += 1;
+                size += flowFile.getSize();
 
+                if (count == 1) {
+                    // first FlowFile is always accepted
                     return FlowFileFilterResult.ACCEPT_AND_CONTINUE;
                 }
 
-                if ((size + flowFile.getSize() > maxBytes) || (count + 1 > maxCount)) {
+                if (size > maxBytes || count > maxCount) {
                     return FlowFileFilterResult.REJECT_AND_TERMINATE;
                 }
 
-                count++;
-                size += flowFile.getSize();
                 return FlowFileFilterResult.ACCEPT_AND_CONTINUE;
             }
-
         };
     }
 
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PackageFlowFile.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PackageFlowFile.java
@@ -32,10 +32,13 @@ import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.flowfile.attributes.CoreAttributes;
 import org.apache.nifi.flowfile.attributes.StandardFlowFileMediaType;
 import org.apache.nifi.processor.AbstractProcessor;
+import org.apache.nifi.processor.DataUnit;
+import org.apache.nifi.processor.FlowFileFilter;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
+import org.apache.nifi.processor.util.FlowFileFilters;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.apache.nifi.util.FlowFilePackager;
 import org.apache.nifi.util.FlowFilePackagerV3;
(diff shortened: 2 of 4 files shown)


nifi-commons/nifi-utils/src/main/java/org/apache/nifi/processor/util/FlowFileFilters.java
```diff
@@ -43,23 +43,20 @@ public static FlowFileFilter newSizeBasedFilter(final double maxSize, final Data
...
+                count += 1;
+                size += flowFile.getSize();
 
+                if (count == 1) {
+                    // first FlowFile is always accepted
                     return FlowFileFilterResult.ACCEPT_AND_CONTINUE;
                 }
 
-                if ((size + flowFile.getSize() > maxBytes) || (count + 1 > maxCount)) {
+                if (size > maxBytes || count > maxCount) {
```
joewitt, line 54: In the previous code count and size were not changed for this case.  The case of reject and terminate is terminal and this FlowFileFilterResult will not be further used nor those values in anyway referenced?

EndzeitBegins: I'm not sure I correctly understood your comment. Please clarify if the following doesn't address your point.

My goal was to simplify the implementation by unifying the three points of computation (inside the first `if` statement, inside the predicate of the second `if`, and before the last return) into a single point.
I added test cases for a few selected scenarios before adjusting the implementation to ensure that the behavior remains the same.

The JavaDoc states:
>  Returns a new {@link FlowFileFilter} that will pull FlowFiles until the maximum file size has been reached, or the maximum FlowFile Count was been reached ...

Therefore, once either the maximum count or accumulated size limit is reached, all subsequent FlowFiles should be rejected.

Upon closer inspection, the old implementation contained a minor bug related to this behavior. If a FlowFile exceeding the size limit was initially rejected (with `REJECT_AND_TERMINATE`), and the caller subsequently passed an empty FlowFile to the filter, the filter would incorrectly return `ACCEPT_AND_CONTINUE` even though the limit had already been reached. While this behavior is technically incorrect, it's primarily a consequence of the caller not adhering to the `FlowFileFilterResult` contract and thus not a critical bug.
I enhanced one of the test cases to demonstrate this behavior.

Both the old and new implementations are susceptible to potential integer overflow issues for both `count` and `size`. I'm not sure if that's something worth addressing though.

I'm content with leaving the implementation of `newSizeBasedFilter` unchanged if that's prefered. I encountered the code while working on the desired changes for `PackageFlowFile` and deemed it worth some tests / minor refactor.

mosermw:
I recommend modifying a MultiProcessorUseCase or creating another UseCase in order to make the documentation for combining the two Batch Size properties clear.  It's very important to be clear that flowfiles will not be delayed in the input queue  waiting for a batch size to be reached.  It's also very important to support packaging exactly 1 flowfile.

While this improvement appears to be worthwhile, we should be very careful with configuration creep on PackageFlowFile.  It's only justification for existence is to be easier to use than MergeContent for a specific use case.  Too many features would ruin that justification.

EndzeitBegins:
Thank you for the useful feedback @mosermw. I've adjusted the documentation of the UseCases to clarify the batching behaviour of the processor.

`PackageFlowFile` in combination with `UnpackContent` is a useful pair of processors to transfer FlowFiles between NiFi clusters where the more robust approach using remote process groups is not applicable, e.g. due to network restrictions.
Packaging more than one FlowFile can improve efficiency both in storage and transmission.
In my opinion, when the content size of the FlowFiles to transfer can vary largely, being able to apply a soft constraint on the package size can be helpful. 

Personally I do not intent to add other properties to the processor at the moment.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PackageFlowFile.java
```diff
@@ -120,17 +122,34 @@
...
             .addValidator(StandardValidators.createLongValidator(1, 10_000, true))
             .build();
 
-    private static final List<PropertyDescriptor> PROPERTIES = List.of(BATCH_SIZE);
+    public static final PropertyDescriptor BATCH_CONTENT_SIZE = new PropertyDescriptor.Builder()
+            .name("Maximum Batch Content Size")
+            .description("Maximum combined content size of FlowFiles to package into one output FlowFile. " +
+                    "Note, that FlowFiles whose content exceeds this limit are packaged separately.")
+            .required(true)
+            .defaultValue("1 GB")
```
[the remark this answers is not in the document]
mosermw: After thinking about this more, and considering that this property enforces a soft upper limit on packaged file size, a large value here won't affect anything if Maximum Batch Size = 1.  So no changes needed.

EndzeitBegins:
Thank you for your review @mosermw. 

Apologies for the delayed response. I've now addressed the feedback.

The pull request has been rebased onto the latest main commit, resolving all conflicts.
Each `PropertyDescriptor` is now on its own line.
The redundant batching benefits description has been removed, and a newline added to the test file's end.

I'd appreciate an review of the updated changes.

mosermw:
Sorry it took so long to review this @EndzeitBegins.  I tested this again and the functionality and documentation looks good.  Code looks good.
+1

PR #9605, opened 2024-12-30
NIFI-10831 enable JWT realm authentication with Elasticsearch

ChrisSamo632:
# Summary

[NIFI-10831](https://issues.apache.org/jira/browse/NIFI-10831) enable JWT realm authentication with Elasticsearch.

Also allow for setting of `runas_user` in the Elasticsearch Controller Service, which may be of use to impersonate a user within Elasticsearch from NiFi, without having the user's credentials (but where the connecting NiFi user/token is granted permission to impersonate the alternate user).

Other changes include:
- removing unused dependencies from Elasticsearch component POMs

Testing:
- Attempted to verify changes using LogTo.io as an OIDC and JWT IdP, but it turns out that LogTo provides Access Token JWTs with a `typ` of `at+jwt`, which Elasticsearch cannot currently handle (https://github.com/elastic/elasticsearch/issues/119370)
- Verified using Keycloak instead (with a machine-to-machine/service-account only `Client` setup to provide Access Tokens)
- Configure Elasticsearch to allow [JWT realm](https://www.elastic.co/guide/en/elasticsearch/reference/current/jwt-auth-realm.html) authentication, an example might be like:

```yml
    jwt.keycloak_test_jwt:
      order: 3
      token_type: access_token
      client_authentication.type: shared_secret
      allowed_issuer: https://localhost:9443/realms/elasticsearch-test
      allowed_audiences: [account]
      allowed_subjects: [e61503d0-8fab-473e-9f70-95d91b085f52]
      allowed_signature_algorithms: [ES384]
      pkc_jwkset_path: https://keycloak:9443/realms/elasticsearch-test/protocol/openid-connect/certs
      claims.principal: preferred_username
      ssl.certificate_authorities:
        [/usr/share/elasticsearch/config/step/certs/root_ca.crt]
```

Note that NiFi requires the IdP to be running over HTTPS (for which SmallStep CA was used to generate certificates locally).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

- ~~[ ] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)~~
- ~~[ ] New dependencies are documented in applicable `LICENSE` and `NOTICE` files~~

### Documentation

- ~~[ ] Documentation formatting appears as expected in rendered files~~

### DIFF
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/pom.xml
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/pom.xml
@@ -36,9 +36,8 @@
             <artifactId>nifi-proxy-configuration-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-            <scope>compile</scope>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-oauth2-provider-api</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/pom.xml
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/pom.xml
@@ -36,6 +36,11 @@
             <artifactId>nifi-proxy-configuration-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-oauth2-provider-api</artifactId>
         </dependency>
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientService.java
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientService.java
@@ -69,9 +69,19 @@ public interface ElasticSearchClientService extends ControllerService, Verifiabl
             .displayName("OAuth2 Access Token Provider")
             .description("The OAuth2 Access Token Provider used to provide JWTs for Bearer Token Authorization with Elasticsearch.")
             .dependsOn(AUTHORIZATION_SCHEME, AuthorizationScheme.JWT)
-            .required(false)
+            .required(true)
             .identifiesControllerService(OAuth2AccessTokenProvider.class)
             .addValidator(Validator.VALID)
+            .build();
+
+    PropertyDescriptor JWT_SHARED_SECRET = new PropertyDescriptor.Builder()
+            .name("jwt-shared-secret")
+            .displayName("JWT Shared Secret")
+            .description("JWT realm Shared Secret.")
+            .dependsOn(AUTHORIZATION_SCHEME, AuthorizationScheme.JWT)
+            .required(true)
+            .sensitive(true)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
 
     PropertyDescriptor RUN_AS_USER = new PropertyDescriptor.Builder()
@@ -88,7 +98,7 @@ public interface ElasticSearchClientService extends ControllerService, Verifiabl
             .displayName("Username")
             .description("The username to use with XPack security.")
             .dependsOn(AUTHORIZATION_SCHEME, AuthorizationScheme.BASIC)
-            .required(false)
+            .required(true)
             .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
@@ -98,7 +108,7 @@ public interface ElasticSearchClientService extends ControllerService, Verifiabl
             .displayName("Password")
             .description("The password to use with XPack security.")
             .dependsOn(AUTHORIZATION_SCHEME, AuthorizationScheme.BASIC)
-            .required(false)
+            .required(true)
             .sensitive(true)
             .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
@@ -109,7 +119,7 @@ public interface ElasticSearchClientService extends ControllerService, Verifiabl
             .displayName("API Key ID")
             .description("Unique identifier of the API key.")
             .dependsOn(AUTHORIZATION_SCHEME, AuthorizationScheme.API_KEY)
-            .required(false)
+            .required(true)
             .sensitive(false)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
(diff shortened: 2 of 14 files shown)


nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientService.java
```diff
@@ -103,6 +124,16 @@ public interface ElasticSearchClientService extends ControllerService, Verifiabl
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
 
+    PropertyDescriptor JWT_SHARED_SECRET = new PropertyDescriptor.Builder()
+            .name("jwt-shared-secret")
+            .displayName("JWT Shared Secret")
+            .description("JWT realm Shared Secret.")
+            .dependsOn(AUTHORIZATION_SCHEME, AuthorizationScheme.JWT)
+            .required(false)
```
exceptionfactory, line 132: This can be changed to `required(true)` and the custom validation can be removed. Also recommend moving this directly under the `OAuth2 Access Token Provider` property.
```suggestion
            .required(true)
```

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service/pom.xml
```diff
@@ -83,23 +87,11 @@
...
             <scope>provided</scope>
         </dependency>
-        <dependency>
-            <groupId>commons-io</groupId>
-            <artifactId>commons-io</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-compress</artifactId>
-        </dependency>
```
exceptionfactory, line 93: Are these dependencies not used?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientServiceImpl.java
```diff
@@ -102,10 +104,12 @@ public class ElasticSearchClientServiceImpl extends AbstractControllerService im
...
     private ObjectMapper mapper;
 
-    private static final List<PropertyDescriptor> properties = List.of(HTTP_HOSTS, PATH_PREFIX, AUTHORIZATION_SCHEME, USERNAME, PASSWORD, API_KEY_ID, API_KEY,
-            PROP_SSL_CONTEXT_SERVICE, PROXY_CONFIGURATION_SERVICE, CONNECT_TIMEOUT, SOCKET_TIMEOUT, CHARSET,
-            SUPPRESS_NULLS, COMPRESSION, SEND_META_HEADER, STRICT_DEPRECATION, NODE_SELECTOR, SNIFF_CLUSTER_NODES,
-            SNIFFER_INTERVAL, SNIFFER_REQUEST_TIMEOUT, SNIFF_ON_FAILURE, SNIFFER_FAILURE_DELAY);
+    private static final List<PropertyDescriptor> properties = List.of(HTTP_HOSTS, PATH_PREFIX, AUTHORIZATION_SCHEME, USERNAME, PASSWORD,
+            API_KEY_ID, API_KEY, JWT_SHARED_SECRET, OAUTH2_ACCESS_TOKEN_PROVIDER, RUN_AS_USER, PROP_SSL_CONTEXT_SERVICE, PROXY_CONFIGURATION_SERVICE,
+            CONNECT_TIMEOUT, SOCKET_TIMEOUT, CHARSET, SUPPRESS_NULLS, COMPRESSION, SEND_META_HEADER, STRICT_DEPRECATION, NODE_SELECTOR,
+            SNIFF_CLUSTER_NODES, SNIFFER_INTERVAL, SNIFFER_REQUEST_TIMEOUT, SNIFF_ON_FAILURE, SNIFFER_FAILURE_DELAY);
```
exceptionfactory, line 110: This is a good opportunity to reformat these properties and list one per line for easier readability.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientServiceImpl.java
```diff
@@ -153,6 +160,23 @@ protected Collection<ValidationResult> customValidate(final ValidationContext va
...
+                );
+            }
+            if (!jwtSharedSecretSet) {
+                results.add(new ValidationResult.Builder().subject(JWT_SHARED_SECRET.getName()).valid(false)
+                        .explanation(String.format("if '%s' is '%s' then '%s' must be set.",
+                                AUTHORIZATION_SCHEME.getDisplayName(), authorizationScheme.getDisplayName(), JWT_SHARED_SECRET.getDisplayName())
+                        ).build()
+                );
+            }
+        }
```
exceptionfactory, line 178: This can be be removed given the use of dependent properties.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientService.java
```diff
@@ -62,6 +64,25 @@ public interface ElasticSearchClientService extends ControllerService, Verifiabl
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
 
+    PropertyDescriptor OAUTH2_ACCESS_TOKEN_PROVIDER = new PropertyDescriptor.Builder()
+            .name("el-cs-oauth2-token-provider")
+            .displayName("OAuth2 Access Token Provider")
+            .description("The OAuth2 Access Token Provider used to provide JWTs for Bearer Token Authorization with Elasticsearch.")
+            .dependsOn(AUTHORIZATION_SCHEME, AuthorizationScheme.JWT)
+            .required(false)
```
exceptionfactory, line 72:
```suggestion
            .required(true)
```

exceptionfactory (changes requested):
Thanks for introducing JWT support and describing the details @ChrisSamo632.

Of particular note, since the new properties depend on the `JWT` scheme, they can be marked as required, which removes the need for custom validation and conditional checking.

On a related note, with the introduction of Run As User support here, does that remove one of the potential use cases for custom HTTP request headers in the related pull request?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service/pom.xml
ChrisSamo632: IntelliJ reports that these dependencies are duplicated or not used here.

It looks like these are now inherited through the `nifi-elasticsearch-bundle`, which itself extends the `nifi-standard-shared-bom`, which provides these standard dependencies.

But this could be a mistake, I'm happy to add them back in if you think this might be an issue (I'm a little rusty on nested POMs) - things seem to be working OK when I run some tests though at least

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientServiceImpl.java
ChrisSamo632: I took the opportunity to update the `BASIC` and `API_KEY` auth types too - I think we can get rid of those custom validations now too (with the associated `dependsOn` settings for those)

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/pom.xml
```diff
@@ -36,9 +36,8 @@
             <artifactId>nifi-proxy-configuration-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
```
exceptionfactory, line 40: The `ObjectMapper` is referenced in `IndexOperationResponse`, so this dependency should be retained, but the scope should be `provided`.

exceptionfactory (commented):
Thanks for the updates @ChrisSamo632, just one additional note on the `jackson-databind` dependency, and then this looks ready to go.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/pom.xml
ChrisSamo632: Good catch!

Reverted the removal and changed the `scope` to `provided`

exceptionfactory (approved):
Thanks for making the adjustments and for adding this new feature @ChrisSamo632! +1 merging

PR #9606, opened 2024-12-31
NIFI-14104 allow setting of HTTP Request Headers for Elasticsearch requests

ChrisSamo632:
# Summary

[NIFI-14104](https://issues.apache.org/jira/browse/NIFI-14104) allow setting of HTTP Request Headers for Elasticsearch requests

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

- ~~[ ] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)~~
- ~~[ ] New dependencies are documented in applicable `LICENSE` and `NOTICE` files~~

### Documentation

- ~~[ ] Documentation formatting appears as expected in rendered files~~

### DIFF
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticsearchRequestOptions.java
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticsearchRequestOptions.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nifi.elasticsearch;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ElasticsearchRequestOptions {
+    private final Map<String, String> requestParameters;
+    private final Map<String, String> requestHeaders;
+
+    public ElasticsearchRequestOptions(final Map<String, String> requestParameters, final Map<String, String> requestHeaders) {
+        this.requestParameters = requestParameters == null ? new HashMap<>() : requestParameters;
+        this.requestHeaders = requestHeaders == null ? new HashMap<>() : requestHeaders;
+    }
+
+    public ElasticsearchRequestOptions() {
+        this(new HashMap<>(), new HashMap<>());
+    }
+
+    public Map<String, String> getRequestParameters() {
+        return requestParameters;
+    }
+
+    public Map<String, String> getRequestHeaders() {
+        return requestHeaders;
+    }
+}
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticsearchRequestOptions.java
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticsearchRequestOptions.java
@@ -17,7 +17,7 @@
 
 package org.apache.nifi.elasticsearch;
 
-import java.util.HashMap;
+import java.util.Collections;
 import java.util.Map;
 
 public class ElasticsearchRequestOptions {
@@ -25,12 +25,12 @@ public class ElasticsearchRequestOptions {
     private final Map<String, String> requestHeaders;
 
     public ElasticsearchRequestOptions(final Map<String, String> requestParameters, final Map<String, String> requestHeaders) {
-        this.requestParameters = requestParameters == null ? new HashMap<>() : requestParameters;
-        this.requestHeaders = requestHeaders == null ? new HashMap<>() : requestHeaders;
+        this.requestParameters = requestParameters == null ? Collections.emptyMap() : Collections.unmodifiableMap(requestParameters);
+        this.requestHeaders = requestHeaders == null ? Collections.emptyMap() : Collections.unmodifiableMap(requestHeaders);
     }
 
     public ElasticsearchRequestOptions() {
-        this(new HashMap<>(), new HashMap<>());
+        this(null, null);
     }
 
     public Map<String, String> getRequestParameters() {
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientServiceImpl.java
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientServiceImpl.java
@@ -823,9 +823,9 @@ public class ElasticSearchClientServiceImpl extends AbstractControllerService im
     public boolean documentExists(final String index, final String type, final String id, final ElasticsearchRequestOptions elasticsearchRequestOptions) {
         boolean exists = true;
         try {
-            final ElasticsearchRequestOptions existsRequestOptions = elasticsearchRequestOptions == null ? new ElasticsearchRequestOptions() : elasticsearchRequestOptions;
-            existsRequestOptions.getRequestParameters().putIfAbsent("_source", "false");
-            get(index, type, id, existsRequestOptions);
+            final Map<String, String> existsParameters = elasticsearchRequestOptions == null ? new HashMap<>() : new HashMap<>(elasticsearchRequestOptions.getRequestParameters());
+            existsParameters.putIfAbsent("_source", "false");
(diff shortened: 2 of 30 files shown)


exceptionfactory (commented):
Thanks for putting this together @ChrisSamo632.

The general approach seems fairly straightforward in terms of mapping prefixed dynamic properties from Processors through to Controller Services. It is notable that this impacts the method signatures of all methods, which raises a general question: should support for custom request headers be at the Processor level, or is it sufficient to be configured at the Controller Service level? In other words, do you expect the need to influence request headers for every call to Elasticsearch, or is this something that could be set on the Controller Service and applied for all requests? The Run As User example mentioned in the Jira seems like it could go either way depending on the flow design. However, if it is sufficient to support custom headers at the Controller Service level, that would reduce the complexity and scope significantly.

ChrisSamo632:
Thanks for the question @exceptionfactory.

I thought this was worth sticking with when thinking about a use case such as using NiFi as a webserver that can receive files/data through a `HandleHttpRequest` or `ListenHttp` processor where the user to be used in Elasticsearch for storage is presented as a request header/attribute that is extracted from the request to a FlowFile attribute, then used when sending the document to Elasticsearch.

A similar use case would be the same sort of thing but with a JWT presented per request that could then be used for authentication with Elasticsearch.

These use cases would need careful thought by the Flow designer so as to ensure that only valid/applicable users were passed in the request. An alternative may be that the Flow contains logic to change the "run as user" based upon some other runtime attributes or FlowFile content.

The problem with the Controller Service-only approach is that it's limited to a single setup per service instance, i.e. it can be configured to connect/run as one single user. The per-processor/FlowFile approach means that multiple users could be used through the same service instance.

There are potentially other request headers that one might want to control on a per-request basis to Elasticsearch, but "run as user" (or JWT auth) in the one that seemed like it might make this change worthwhile, alongside #9605 for NIFI-10831 that allows the simpler configuration of a single user/JWT provider for an alternative static authentication mechanism between NiFi and Elasticsearch.

ChrisSamo632:
@exceptionfactory what do you think about a change to the Elasticsearch Service's API to replace the existing `requestParams` and new `requestHeaders` maps with a more generic combined `requestOptions` object?

Thinking that, in the future, there may well be other per-request options that would be useful, and agreeing that it would be best to avoid wholesale changes to all methods on the Service API in future tickets, such a change here would help reduce the impact of future changes in the same area.

exceptionfactory:
> @exceptionfactory what do you think about a change to the Elasticsearch Service's API to replace the existing `requestParams` and new `requestHeaders` maps with a more generic combined `requestOptions` object?
> 
> Thinking that, in the future, there may well be other per-request options that would be useful, and agreeing that it would be best to avoid wholesale changes to all methods on the Service API in future tickets, such a change here would help reduce the impact of future changes in the same area.

That sounds like a good approach. That would support passing with parameters and headers for this set of changes, and make it easier to introduce any other changes down the road.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service/src/main/java/org/apache/nifi/elasticsearch/ElasticSearchClientServiceImpl.java
```diff
@@ -336,7 +335,9 @@ private void verifySniffer(final ConfigurationContext context, final RestClient
     }
 
     private List<String> getElasticsearchRoot(final RestClient verifyClient, final OAuth2AccessTokenProvider tokenProvider) throws IOException {
-        final Request request = addJWTAuthorizationHeader(new Request("GET", "/"), tokenProvider);
+        final Request request = new Request("GET", "/");
+        addJWTAuthorizationHeader(request.getOptions().toBuilder(), tokenProvider);
```
ChrisSamo632, line 339: Need to `setOptions` on the `request` once the request option builder has been updated by `addJWTAuthorizationHeader`

ChrisSamo632: Addressed mistake made after `rebase` from `main`

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-client-service-api/src/main/java/org/apache/nifi/elasticsearch/ElasticsearchRequestOptions.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class ElasticsearchRequestOptions {
+    private final Map<String, String> requestParameters;
+    private final Map<String, String> requestHeaders;
+
+    public ElasticsearchRequestOptions(final Map<String, String> requestParameters, final Map<String, String> requestHeaders) {
+        this.requestParameters = requestParameters == null ? new HashMap<>() : requestParameters;
+        this.requestHeaders = requestHeaders == null ? new HashMap<>() : requestHeaders;
```
exceptionfactory, line 29: Recommend wrapping the input Maps in `unmodifiableMap`.

exceptionfactory (commented):
@ChrisSamo632, returning to this pull request, does it still seem necessary in light of #9605 supporting the Run As User property?

The tests use `Accept` as an example, but that is actually a good example of potentially breaking changes that could be introduced through dynamic headers, making the request invalid. Dynamic properties provide a useful way to cover a large number of custom configuration scenarios without introducing many explicit properties. In the case of Elasticsearch, however, the HTTP request seems more narrowly defined in terms of possible options that Elasticsearch supports.

I think the general implementation approach looks good, but if there isn't a strong case for a variety of custom HTTP request headers, it seems better to avoid introducing this level of complexity for now.

ChrisSamo632:
@exceptionfactory thanks for continuing to look at this.

> I think the general implementation approach looks good, but if there isn't a strong case for a variety of custom HTTP request headers, it seems better to avoid introducing this level of complexity for now.

The use of request headers for Elasticsearch interaction is somewhat limited, but becomes a bit more relevant if people are using Elasticsearch-like services and/or proxies that require/allow extra request headers in-between.

You're right that the use of `Accept` headers in the tests does highlight the possibility of someone introducing a configuration that would break NiFi's interaction with Elasticsearch, but the same could be said for other combinations of properties across the Elasticearch components within NiFi.

The main use case I was raising this for is to allow for per-request (i.e. FlowFile of data coming through an Elasticsearch processor) setting of some request headers, namely the run-as-user. But I then chose not to put that in the tests as it may cause issues/confusion if not paired with other required headers that such an authentication mechanism needs within Elasticsearch.

There are Flows where one may wish to index data as different users, but wouldn't necessarily know those users upfront, nor wish to create an ElasticsearchClientService controller and matching *Elasticsearch processor combination for every user that will be used for data ingest. The main example of this kind of approach that I can think of would be if the NiFi Flow were fronted by a HandleHttpRequest/ListenHTTP processor that is acting as a RESTful API endpoint - a setup that I've wanted to create in the past, but have had to agree for an alternative workaround in the meantime (which can complicate RBAC/ABAC controls within Elasticsearch itself, rather than relying on per-user access configuration).

A more generic version of this would be if NiFi were to enable the use of FlowFile attributes at the Controller level, e.g. a property is evaluated at runtime from FlowFile attributes as they are within Processors. That, of course, is a much bigger framework-level change and comes with its own set of difficulties for things like this around connection pooling, etc.

exceptionfactory:
Thanks for the thoughtful reply @ChrisSamo632.

Supporting FlowFile-attributed-based parameters for something like a run-as user raises some potential security concerns, but the same could be said of certain other scenarios using InvokeHTTP.

Are there any other use cases besides that one that come to mind? The run-as user case does not seem to be a strong reason, and I'm not thinking of others right now. If there are others, I would be willing to go forward with these changes.

ChrisSamo632:
@exceptionfactory thanks for the thoughtful discussion. I understand the reticence around the change to the Elasticsearch processors and components within nifi of there's uncertainty of the level of benefit.

Along with the potential use for per-request user credentials such as [JWTs](https://www.elastic.co/guide/en/elasticsearch/reference/current/jwt-auth-realm.html#hmac-oidc-example-request-headers) (which do indeed come with the potential for security concerns, but as noted similar concerns exist elsewhere with other processors), Elasticsearch has a few general request headers that can be useful for audit and debug logging, along with functional use:

* [content type](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#_content_type_requirements)
  * although this isn't likely to be overly useful for nifi use cases, as the flowfile content type isn't likely to change within the flow between requests, and even if it did, all of the Elasticsearch components currently handle only (nd)json
* [request origin tracking](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#x-opaque-id)
  * could be used for general observability between nifi & elasticsearch
* [request tracing](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#traceparent)
  * could be used for application logging between components and, potentially, request auditing

The use case of nifi as an intermediary webserver between a client and elasticsearch where the operations are intended to be authenticate with the end user's identity are probably somewhat niche, but I've run into such environments a few times over the years, hence the proposed enhancement.

Probably more useful are the cross-system tracing capabilities for the increasingly relevant and popular observability. Use of these headers could be implemented a different way, but this generic approach would allow for each of these use cases, along with catering for environments that have a proxy in between system components potentially requiring additional request headers (although again, those setups aren't likely to need per-request differences in the headers for the most part).

There are likely some Elasticsearch endpoints (or equivalent provider systems that follow the same general endpoints as those provided by Elastic) that would make use of additional request headers, but I don't have any available to show here (I'm aware of some such systems I've used in the past, but not with nifi, and not that are generally available).

One thing included in this PR was a refactor of how request parameters are handled between the Elasticsearch components within nifi, making such changes simpler in the future. That said, in itself, it probably doesn't warrant the change if the functional benefits aren't thought sufficient.

exceptionfactory (approved):
Thanks for the helpful and thoughtful response @ChrisSamo632. The request tracing capabilities seem potentially useful, and the refactoring you mentioned is helpful across the board. I appreciate you taking the time to evaluate the approach. With that, I agree this should go forward. +1 merging

PR #9607, opened 2024-12-31
NIFI-14059 - add properties needed to use Kafka authentication mechan…

greyp9:
…ism SASL_SSL

Tested using Docker kerberos/kafka images and property "SELF_CONTAINED_KERBEROS_USER_SERVICE", but also speculatively adding property "KERBEROS_CREDENTIALS_SERVICE" from deprecated implementation.

As a running kerberos instance is needed to test this, I did not implement an IT, but I'm open to alternate means of providing test coverage of the changes.

# Summary

[NIFI-14059](https://issues.apache.org/jira/browse/NIFI-14059)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-service/src/main/java/org/apache/nifi/kafka/service/Kafka3ConnectionService.java
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-service/src/main/java/org/apache/nifi/kafka/service/Kafka3ConnectionService.java
@@ -53,7 +53,6 @@ import org.apache.nifi.kafka.shared.property.SaslMechanism;
 import org.apache.nifi.kafka.shared.property.provider.KafkaPropertyProvider;
 import org.apache.nifi.kafka.shared.property.provider.StandardKafkaPropertyProvider;
 import org.apache.nifi.kafka.shared.transaction.TransactionIdSupplier;
-import org.apache.nifi.kerberos.KerberosUserService;
 import org.apache.nifi.kerberos.SelfContainedKerberosUserService;
 import org.apache.nifi.logging.ComponentLog;
 import org.apache.nifi.processor.util.StandardValidators;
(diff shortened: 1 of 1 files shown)


nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-service/src/main/java/org/apache/nifi/kafka/service/Kafka3ConnectionService.java
```diff
@@ -146,6 +149,22 @@ public class Kafka3ConnectionService extends AbstractControllerService implement
             )
             .build();
 
+    public static final PropertyDescriptor KERBEROS_CREDENTIALS_SERVICE = new PropertyDescriptor.Builder()
+            .name("kerberos-credentials-service")
+            .displayName("Kerberos Credentials Service")
+            .description("Service supporting generalized credentials authentication with Kerberos")
+            .identifiesControllerService(KerberosUserService.class)
+            .required(false)
+            .build();
```
bbende, line 158: It has been a while since I worked on this, but I am not sure the Kafka processors can use the regular `KerberosUserService` because they expect to get the JAAS config string from the service and give it to the Kafka client which does the actual login, and this won't work for some of the implementations of `KerberosUserService`, this was the reason for limiting it to the `SelfContainedKerberosUserService`, see comment here:

https://github.com/apache/nifi/blob/main/nifi-extension-bundles/nifi-standard-services/nifi-kerberos-user-service-api/src/main/java/org/apache/nifi/kerberos/SelfContainedKerberosUserService.java#L20-L25

greyp9: Thanks for confirming.  All this lines up perfectly with my observations.  I didn't try `KERBEROS_CREDENTIALS_SERVICE`; I'll remove it.  I was able to get things working with `KERBEROS_USER_SERVICE`, which referenced `SelfContainedKerberosUserService`, by adding the custom property `sasl.jaas.config` as described in this Confluent documentation:

https://docs.confluent.io/platform/7.3/kafka/authentication_sasl/authentication_sasl_gssapi.html#kafka-sasl-auth-gssapi

Also assuming that `KERBEROS_SERVICE_NAME` is needed, as it would seem that NiFi would need a means to convey which service it wants to authenticate to.

mattyb149:
+1 LGTM, verified the changes allow the proper authentication to happen. Thanks for the fix! Merging to main

PR #9617, opened 2025-01-09
NIFI-14141 add variable size for FlowFiles created by GenerateFlowFile

markobean:
# Summary

[NIFI-14141](https://issues.apache.org/jira/browse/NIFI-14141)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-14141) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

Configured GenerateFlowFile with Variable File Size set to true. Noted validation failures if Unique FlowFiles is set to false or Maximum File Size is less than Minimum File Size.

Executed GenerateFlowFile with varying Batch Size values. All values lead to variable sizes of FlowFiles even within the same batch.

Added unit test and ran thousands of times to assure there were no failures given that the nature of this change involves a pseudo-random FlowFile size.

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
@@ -81,16 +81,12 @@ public class GenerateFlowFile extends AbstractProcessor {
     public static final PropertyDescriptor MINIMUM_FILE_SIZE = new PropertyDescriptor.Builder()
             .name("Minimum File Size")
             .description("The minimum size of the FlowFile that will be generated")
-            .required(true)
-            .defaultValue("0B")
             .addValidator(StandardValidators.DATA_SIZE_VALIDATOR)
             .dependsOn(VARIABLE_SIZE, "true")
             .build();
     public static final PropertyDescriptor MAXIMUM_FILE_SIZE = new PropertyDescriptor.Builder()
             .name("Maximum File Size")
             .description("The maximum size of the FlowFile that will be generated")
-            .required(true)
-            .defaultValue("1 KB")
             .addValidator(StandardValidators.DATA_SIZE_VALIDATOR)
             .dependsOn(VARIABLE_SIZE, "true")
             .build();
@@ -216,14 +212,33 @@ public class GenerateFlowFile extends AbstractProcessor {
         }
 
         if (isVariableSize) {
-            double minSize = validationContext.getProperty(MINIMUM_FILE_SIZE).asDataSize(DataUnit.B);
-            double maxSize = validationContext.getProperty(MAXIMUM_FILE_SIZE).asDataSize(DataUnit.B);
-            if (maxSize < minSize) {
+            boolean minMaxSet = true;
+            if (!validationContext.getProperty(MINIMUM_FILE_SIZE).isSet()) {
+                results.add(new ValidationResult.Builder()
+                        .subject("Minimum File Size")
+                        .valid(false)
+                        .explanation("Minimum File Size must be set when Variable File Size is set to true.")
+                        .build());
+                minMaxSet = false;
+            }
+            if (!validationContext.getProperty(MAXIMUM_FILE_SIZE).isSet()) {
                 results.add(new ValidationResult.Builder()
                         .subject("Maximum File Size")
                         .valid(false)
-                        .explanation("Maximum File Size must be greater than or equal to Minimum File Size")
+                        .explanation("Maximum File Size must be set when Variable File Size is set to true.")
                         .build());
+                minMaxSet = false;
+            }
+            if (minMaxSet) {
+                double minSize = validationContext.getProperty(MINIMUM_FILE_SIZE).asDataSize(DataUnit.B);
+                double maxSize = validationContext.getProperty(MAXIMUM_FILE_SIZE).asDataSize(DataUnit.B);
+                if (maxSize < minSize) {
+                    results.add(new ValidationResult.Builder()
+                            .subject("Maximum File Size")
+                            .valid(false)
+                            .explanation("Maximum File Size must be greater than or equal to Minimum File Size")
+                            .build());
+                }
             }
             if (!isUnique) {
                 results.add(new ValidationResult.Builder()
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
@@ -71,12 +71,36 @@ public class GenerateFlowFile extends AbstractProcessor {
     public static final String DATA_FORMAT_BINARY = "Binary";
     public static final String DATA_FORMAT_TEXT = "Text";
 
-    public static final PropertyDescriptor FILE_SIZE = new PropertyDescriptor.Builder()
-            .name("File Size")
-            .description("The size of the file that will be used")
+    public static final PropertyDescriptor VARIABLE_SIZE = new PropertyDescriptor.Builder()
+            .name("Variable File Size")
+            .description("The size of the FlowFile will be randomized between a minimum and maximum size value. 'Unique FlowFiles' must be set to true in order to achieve varying sizes.")
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .build();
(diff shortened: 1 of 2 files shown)


exceptionfactory (changes requested):
Thanks for proposing this change @markobean.

As the Jira issue does not provide much background, can you describe the general use case in mind for randomized FlowFile sizes?

From an implementation perspective, it seems like it would be simpler to support using Expression Language in the existing `File Size` property. That would allow for some amount of random size generation if desired, without introducing additional configuration properties.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
```diff
@@ -71,12 +71,36 @@ public class GenerateFlowFile extends AbstractProcessor {
     public static final String DATA_FORMAT_BINARY = "Binary";
     public static final String DATA_FORMAT_TEXT = "Text";
 
+    public static final PropertyDescriptor VARIABLE_SIZE = new PropertyDescriptor.Builder()
+            .name("Variable File Size")
+            .description("The size of the FlowFile will be randomized between a minimum and maximum size value. 'Unique FlowFiles' must be set to true in order to achieve varying sizes.")
+            .required(true)
```
joewitt, line 77: why make all these required?

markobean: I was keeping in style with the existing File Size attribute. I can make not required if you prefer - and perhaps add some custom validation to be sure min/max size is specified when Variable File Size is true.

markobean:
> Thanks for proposing this change @markobean.
> 
> As the Jira issue does not provide much background, can you describe the general use case in mind for randomized FlowFile sizes?
> 
> From an implementation perspective, it seems like it would be simpler to support using Expression Language in the existing `File Size` property. That would allow for some amount of random size generation if desired, without introducing additional configuration properties.

Expression Language will not work in this case as GenerateFlowFile does not accept an incoming connection. At best, an environment variable could be referenced, but again, this would be a fixed value.

This use case is to have FlowFiles of various sizes injected into a flow for further testing. Some flows are specifically designed to handle FlowFiles of certain sizes differently. For example, if files are "large", perhaps they are compressed. This is a trivial example, but the point is this feature supports a greater degree of flexibility of data while testing a flow.

exceptionfactory:
> > Thanks for proposing this change @markobean.
> > As the Jira issue does not provide much background, can you describe the general use case in mind for randomized FlowFile sizes?
> > From an implementation perspective, it seems like it would be simpler to support using Expression Language in the existing `File Size` property. That would allow for some amount of random size generation if desired, without introducing additional configuration properties.
> 
> Expression Language will not work in this case as GenerateFlowFile does not accept an incoming connection. At best, an environment variable could be referenced, but again, this would be a fixed value.
> 
> This use case is to have FlowFiles of various sizes injected into a flow for further testing. Some flows are specifically designed to handle FlowFiles of certain sizes differently. For example, if files are "large", perhaps they are compressed. This is a trivial example, but the point is this feature supports a greater degree of flexibility of data while testing a flow.

Expression Language can be used in absence of an input FlowFile, it would be evaluated as a self-contained expression. EL has a `random()` function that would support this kind of use case.

Thanks for the additional background on potential test scenarios. I can appreciate testing files of different sizes. However, if that is the case, it seems like it would be important to have deterministic inputs, versus random inputs. Either way, EL with some logic would actually provide more flexibility for some sort of pseudo-random size-based output files.

markobean:
> > > Thanks for proposing this change @markobean.
> > > As the Jira issue does not provide much background, can you describe the general use case in mind for randomized FlowFile sizes?
> > > From an implementation perspective, it seems like it would be simpler to support using Expression Language in the existing `File Size` property. That would allow for some amount of random size generation if desired, without introducing additional configuration properties.
> > 
> > 
> > Expression Language will not work in this case as GenerateFlowFile does not accept an incoming connection. At best, an environment variable could be referenced, but again, this would be a fixed value.
> > This use case is to have FlowFiles of various sizes injected into a flow for further testing. Some flows are specifically designed to handle FlowFiles of certain sizes differently. For example, if files are "large", perhaps they are compressed. This is a trivial example, but the point is this feature supports a greater degree of flexibility of data while testing a flow.
> 
> Expression Language can be used in absence of an input FlowFile, it would be evaluated as a self-contained expression. EL has a `random()` function that would support this kind of use case.
> 
> Thanks for the additional background on potential test scenarios. I can appreciate testing files of different sizes. However, if that is the case, it seems like it would be important to have deterministic inputs, versus random inputs. Either way, EL with some logic would actually provide more flexibility for some sort of pseudo-random size-based output files.

I added EL support and tried a couple EL statements. They do not seem to generate the required <size> <label> format, e.g. "10 B". The following do not work:

${random():mod(10):plus(1):append(" B")}
${random():mod(10):plus(1)} B

java.lang.IllegalArgumentException: Invalid data size: ${random():mod(10):plus(1):append(" B")}
java.lang.IllegalArgumentException: Invalid data size: ${random():mod(10):plus(1)} B

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
markobean: @joewitt I made the min/max file sizes non-required. It makes more work for the user to have to enter values. However, I'm inclined to think the default values would not be appropriate in a majority of the cases anyhow.

exceptionfactory:
> I added EL support and tried a couple EL statements. They do not seem to generate the required format, e.g. "10 B". The following do not work:
> 
> ${random():mod(10):plus(1):append(" B")} ${random():mod(10):plus(1)} B
> 
> java.lang.IllegalArgumentException: Invalid data size: ${random():mod(10):plus(1):append(" B")} java.lang.IllegalArgumentException: Invalid data size: ${random():mod(10):plus(1)} B

The validator needs to be changed in order to accept EL, in addition to directly-declared sizes.

markobean:
> > I added EL support and tried a couple EL statements. They do not seem to generate the required format, e.g. "10 B". The following do not work:
> > ${random():mod(10):plus(1):append(" B")} ${random():mod(10):plus(1)} B
> > java.lang.IllegalArgumentException: Invalid data size: ${random():mod(10):plus(1):append(" B")} java.lang.IllegalArgumentException: Invalid data size: ${random():mod(10):plus(1)} B
> 
> The validator needs to be changed in order to accept EL, in addition to directly-declared sizes.

Ok, I have made the changes and it seems to be working. I just want to add a more specific validator. I used the standard NON_EMPTY_VALIDATOR just to see if the EL would work. I'll have to add some custom logic to be sure the value properly evaluates to a <value> <label> for data size.

Having said that, while this EL approach can be made to work, I find it far more obscure than the provided implementation. From a user-friendly perspective, constructing the proper EL is more complex. I rarely agree with changing code just because a user wasn't aware of a feature or capability. In this case though, I find it much easier to understand.

I would like the submitted implementation to be considered over EL unless there is a strong argument against it. Why do you prefer EL?

exceptionfactory:
> I would like the submitted implementation to be considered over EL unless there is a strong argument against it. Why do you prefer EL?

I will also note that the current test case that runs the processor 1000 times in an effort to produce files of expected sizes highlights the problem with the non-deterministic nature of the current approach. Such a test is prone to inconsistent behavior. That is some that could be addressed separately, but it points back to another reason for going with EL.

markobean:
> > I would like the submitted implementation to be considered over EL unless there is a strong argument against it. Why do you prefer EL?
> 
> I will also note that the current test case that runs the processor 1000 times in an effort to produce files of expected sizes highlights the problem with the non-deterministic nature of the current approach. Such a test is prone to inconsistent behavior. That is some that could be addressed separately, but it points back to another reason for going with EL.

The same issue exists for the EL solution. Both do not necessarily require 1000 iterations. But because they ultimately utilize a pseudo-random number generator, I thought it wise to keep the number of iterations relatively large. It's not a compute-intensive processor, so iterating this many times is acceptable. On my system, an iteration of 1 takes ~150ms; 1,000 iterations 250ms. It's a small delta for a dramatic increase in data. This will ensure a very high likelihood that the full spectrum of allowed data sizes are covered. And, the asserts will continue to work even if the full range is not exercised. For example, say the range is intended to be 1-10 bytes, but the test only achieves 2-9 byte files. The absence of 1 and 10 will not cause the unit test to fail; it just never reached the edge cases. The only way to fail is if all 1,000 executions happen to generate the exact same-sized file. Not going to happen in our lifetime.. unless the randomizer is broken. ;)

markobean:
I am going to close this PR and open a new one with the EL solution.

PR #9618, opened 2025-01-09
NIFI-14136: Fixing issue where a new Parameters value would be cleare…

mcgilman:
…d if it were edited prior to submission.


scottyaslan (approved):
I have verified that if the user creates a new parameter with a value in a parameter context, then before submitting the user edits the new parameter and updates the description, that the correct value for the new parameter is in fact submitted. Thanks for this fix @mcgilman!!! +1

The unit tests are an awesome addition as well!

PR #9620, opened 2025-01-09
NIFI-14141 Accept Expression Language statement in GenerateFlowFile, …

markobean:
…File Size property allowing FlowFiles of variable sizes to be generated. Included some examples of EL statements in Additional Details for GenerateFlowFile.

# Summary

[NIFI-14141](https://issues.apache.org/jira/browse/NIFI-14141)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
@@ -76,7 +76,7 @@ public class GenerateFlowFile extends AbstractProcessor {
             .description("The size of the file that will be used")
             .required(true)
             .defaultValue("0B")
-            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .addValidator(StandardValidators.DATA_SIZE_VALIDATOR)
             .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
             .build();
     public static final PropertyDescriptor BATCH_SIZE = new PropertyDescriptor.Builder()
@@ -187,8 +187,6 @@ public class GenerateFlowFile extends AbstractProcessor {
             results.add(new ValidationResult.Builder().subject("Custom Text").valid(false).explanation("If Custom Text is set, then Data Format must be "
                     + "text and Unique FlowFiles must be false.").build());
         }
-
-        results.add(StandardValidators.DATA_SIZE_VALIDATOR.validate("File Size", validationContext.getProperty(FILE_SIZE).evaluateAttributeExpressions().getValue(), validationContext));
 
         return results;
     }
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
@@ -76,7 +76,8 @@ public class GenerateFlowFile extends AbstractProcessor {
             .description("The size of the file that will be used")
             .required(true)
             .defaultValue("0B")
-            .addValidator(StandardValidators.DATA_SIZE_VALIDATOR)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
             .build();
     public static final PropertyDescriptor BATCH_SIZE = new PropertyDescriptor.Builder()
             .name("Batch Size")
@@ -187,11 +188,13 @@ public class GenerateFlowFile extends AbstractProcessor {
                     + "text and Unique FlowFiles must be false.").build());
         }
 
+        results.add(StandardValidators.DATA_SIZE_VALIDATOR.validate("File Size", validationContext.getProperty(FILE_SIZE).evaluateAttributeExpressions().getValue(), validationContext));
+
         return results;
     }
 
     private byte[] generateData(final ProcessContext context) {
-        final int byteCount = context.getProperty(FILE_SIZE).asDataSize(DataUnit.B).intValue();
+        final int byteCount = context.getProperty(FILE_SIZE).evaluateAttributeExpressions().asDataSize(DataUnit.B).intValue();
 
         final Random random = new Random();
         final byte[] array = new byte[byteCount];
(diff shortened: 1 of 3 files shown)


nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
```diff
@@ -76,7 +76,8 @@ public class GenerateFlowFile extends AbstractProcessor {
             .description("The size of the file that will be used")
             .required(true)
             .defaultValue("0B")
-            .addValidator(StandardValidators.DATA_SIZE_VALIDATOR)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
```
exceptionfactory, line 79: The `DATA_SIZE_VALIDATOR` includes support for expression language, so is this change necessary?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestGenerateFlowFile.java
```diff
@@ -73,4 +78,30 @@ public void testDynamicPropertiesToAttributes() {
         generatedFlowFile.assertAttributeEquals("mime.type", "application/text");
     }
 
+    @Test
+    public void testExpressionLanguageSupport() {
+        TestRunner runner = TestRunners.newTestRunner(new GenerateFlowFile());
+        runner.setProperty(GenerateFlowFile.FILE_SIZE, "${random():mod(10):plus(1)}B");  // represents a range of 1-10 bytes, inclusive
```
exceptionfactory, line 84: Instead of building the unit test against the complexities around the random function, it would be better to use a simple deterministic expression, avoiding the need for multiple runs.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestGenerateFlowFile.java
```diff
@@ -73,4 +78,30 @@ public void testDynamicPropertiesToAttributes() {
...
+        flowFiles.forEach(ff -> {
+            long fileSize = ff.getSize();
+            fileSizeDistribution.put(fileSize, fileSizeDistribution.getOrDefault(fileSize, 0) + 1);
+        });
+        long minSize = fileSizeDistribution.keySet().stream().min(Long::compareTo).orElse(-1L);
+        long maxSize = fileSizeDistribution.keySet().stream().max(Long::compareTo).orElse(-1L);
+
+        // Assert all file sizes fall within the range and that there exists more than one unique file size value
+        Assertions.assertTrue(minSize > 0 && minSize < maxSize);
+        Assertions.assertTrue(maxSize <= 10);
```
exceptionfactory, line 103: The import static should be used for `assertTrue`.

exceptionfactory (changes requested):
Thanks for putting forward the expression language-based approach @markobean.

I noted a couple things on the validator configuration, and I also recommend simplifying the unit test to avoid using the `random()` function. The Additional Details documentation is great, and provides very useful examples, but it isn't necessary to introduce that complexity in the unit test because it is a feature of expression language itself.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
markobean: Without that validator within the property, the processor was invalid indicating:
"'File Size' validated against '${random():mod(101)}B' is invalid because 'File Size' is not a supported property or has no Validator associated with it"
As the property is in fact required, this is a reasonable validator to add.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestGenerateFlowFile.java
markobean: Modified to use `nextInt()` function. But still made 2 batch executions to confirm file size uniqueness (or not) based on `UNIQUE_FLOWFILES` setting. This is slightly out of scope, but the batch processing does not have any other unit test. 2 birds.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestGenerateFlowFile.java
markobean: Done.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
exceptionfactory: @markobean Reverting this change and running the unit test worked as expected. See `StandardValidators` line 370 where the validator handles the presence of expression language. Do you still see the validation error?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
```diff
@@ -187,11 +188,13 @@ protected Collection<ValidationResult> customValidate(ValidationContext validati
                     + "text and Unique FlowFiles must be false.").build());
         }
 
+        results.add(StandardValidators.DATA_SIZE_VALIDATOR.validate("File Size", validationContext.getProperty(FILE_SIZE).evaluateAttributeExpressions().getValue(), validationContext));
```
exceptionfactory, line 191: This line should also be removed after reverting the validator change.

exceptionfactory (changes requested):
Thanks for making the updates @markobean, the simplified unit test approach looks good.

I tested reverting the validator changes and the `DATA_SIZE_VALIDATOR` works with and without EL configured in the property, based on the unit tests included. Please review that behavior as it should be possible to revert that change. If you are still seeing an error, there could be something else going on.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
markobean: Thanks for pointing this out. I missed the fact that `DATA_SIZE_VALIDATOR` supports EL. The error I saw was only when no validator was provided at all.
I reverted back to `DATA_SIZE_VALIDATOR` and tested in a running NiFi. All good.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java
markobean: Removed.

exceptionfactory (approved):
Thanks for working through the feedback @markobean, the latest version looks good! +1 merging

PR #9623, opened 2025-01-10
 NIFI-14147 add support for JWT token to JettyWebSocketClient controller

POliveNokia:
# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-websocket-bundle/nifi-websocket-services-jetty/pom.xml
+++ b/nifi-extension-bundles/nifi-websocket-bundle/nifi-websocket-services-jetty/pom.xml
@@ -55,6 +55,21 @@
             <artifactId>jetty-ee10-servlet</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.jsonwebtoken</groupId>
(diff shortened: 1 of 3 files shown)


exceptionfactory (changes requested):
Thanks for proposing support for JWT authentication in the WebSocketClient @POliveNokia.

The design and implementation raise a number of security-relevant implementation questions. I highlighted some library choices that need to be revisited, but the main question relates to use case and design of the properties. The OAuth 2 Token Provider service supports a standard method for obtaining Access Tokens from an Identity Provider. The proposed changes with a static file-based JWT do not provide a scalable or standards-oriented approach.

There is definitely value in supporting token-based authentication, but this needs to be implemented in a way that follows published standards for JWT generation.

As this will require some significant rework, it would be much better to provide more substantive details in the associated Jira issue. I would be glad to work through some of the design considerations there, and then return to considering a new pull request after considering the approach.

PR #9635, opened 2025-01-15
NIFI-14163: Enhance GCP Credential to use delegate user

bobpaulin:
# Summary

[NIFI-14163](https://issues.apache.org/jira/browse/NIFI-14163)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

To verify this change you must use a service account credential that has been configured to use domain-wide delegation with the scopes that your processors are expecting to use (per https://developers.google.com/identity/protocols/oauth2/service-account#delegatingauthority).  Ex ListGoogleDrive specifically requires the https://www.googleapis.com/auth/drive.metadata.readonly scope.

To verify positve and negative cases:
Use a Google Drive that has the service account user added and configure the GCPCredentialControllerService to use  a Service Account JSON.
Expected Result: Drive Files should appear

Remove the same Service Account email from the drive permissions.
Expected Result: Drive Files will not appear.  (Service Account can't access them)

Configure GCPCredentialControllerService to use the delegation user of an email that currently has access to the drive.
Expected Result: Drive Files will appear.,

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/CredentialPropertyDescriptors.java
+++ b/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/CredentialPropertyDescriptors.java
@@ -91,4 +91,22 @@ public final class CredentialPropertyDescriptors {
             .sensitive(true)
             .build();
 
+    public static final PropertyDescriptor DELEGATION_STRATEGY = new PropertyDescriptor.Builder()
+            .name("Delegation Strategy")
+            .required(true)
+            .defaultValue(DelegationStrategy.SERVICE_ACCOUNT)
+            .allowableValues(DelegationStrategy.class)
+            .description("The Delegation Strategy determines which account is used when calls are made with the GCP Credential.")
+            .build();
+
+    public static final PropertyDescriptor DELEGATION_USER = new PropertyDescriptor.Builder()
+            .name("Delegation User")
+            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
+            .required(true)
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .dependsOn(DELEGATION_STRATEGY, DelegationStrategy.DELEGATED_ACCOUNT)
+            .description("This user will be impersonated by the service account for api calls. " +
+                    "API calls made using this credential will appear as if they are coming from delegate user with the delegate user's access. " +
+                    "Any scopes supplied from processors to this credential must have domain-wide delegation setup with the service account.")
+            .build();
 }
--- a/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/strategies/AbstractServiceAccountCredentialsStrategy.java
+++ b/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/strategies/AbstractServiceAccountCredentialsStrategy.java
@@ -19,6 +19,8 @@ package org.apache.nifi.processors.gcp.credentials.factory.strategies;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.auth.oauth2.GoogleCredentials;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.processors.gcp.credentials.factory.CredentialPropertyDescriptors;
+import org.apache.nifi.processors.gcp.credentials.factory.DelegationStrategy;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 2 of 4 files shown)


nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/CredentialPropertyDescriptors.java
```diff
@@ -91,4 +91,24 @@ private CredentialPropertyDescriptors() { }
             .sensitive(true)
             .build();
 
+    public static final PropertyDescriptor DELEGATION_STRATEGY = new PropertyDescriptor.Builder()
+            .name("delegation-strategy")
+            .displayName("Delegation Strategy")
```
pvillard31, line 96: We're trying to get away from `.displayName`, we should only use `.name` for new properties. I think it'd be worth for us to mark `displayName` as deprecated on the API side to make this more obvious.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/CredentialPropertyDescriptors.java
```diff
@@ -91,4 +91,24 @@ private CredentialPropertyDescriptors() { }
...
+            .required(false)
```
pvillard31, line 97: Since it's a property with allowable values, I believe the convention is to have it as required = true.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/CredentialPropertyDescriptors.java
```diff
@@ -91,4 +91,24 @@ private CredentialPropertyDescriptors() { }
...
+            .defaultValue(DelegationStrategy.SERVICE_ACCOUNT)
+            .allowableValues(DelegationStrategy.class)
+            .description("The Delegation Strategy determines which account is used when calls are made with the GCP Credential.")
+            .build();
+
+    public static final PropertyDescriptor DELEGATION_USER = new PropertyDescriptor.Builder()
+            .name("delegation-user")
+            .displayName("Delegation User")
```
pvillard31, line 105: Remove use of `displayName`

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/strategies/AbstractServiceAccountCredentialsStrategy.java
```diff
@@ -37,7 +39,13 @@ public AbstractServiceAccountCredentialsStrategy(String name, PropertyDescriptor
 
     @Override
     public GoogleCredentials getGoogleCredentials(Map<PropertyDescriptor, String> properties, HttpTransportFactory transportFactory) throws IOException {
+        final String delegationStrategy = properties.get(CredentialPropertyDescriptors.DELEGATION_STRATEGY);
+        if (delegationStrategy != null && delegationStrategy.equals(DelegationStrategy.DELEGATED_ACCOUNT.getValue())) {
+            final String delegationUser = properties.get(CredentialPropertyDescriptors.DELEGATION_USER);
+            return GoogleCredentials.fromStream(getServiceAccountJson(properties), transportFactory).createDelegated(delegationUser);
+        } else {
             return GoogleCredentials.fromStream(getServiceAccountJson(properties), transportFactory);
+        }
```
pvillard31, line 48: Given this is an enum and it'll never be null, can't we use the switch case return style here?

pvillard31 (changes requested):
Thanks @bobpaulin - left some code style related comments. Will give this a try!

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/strategies/AbstractServiceAccountCredentialsStrategy.java
bobpaulin: So that would be possible but the switch statement would be based on the string value since properties map doesn't return the enum directly.  So I'd still have a default case to handle that would essentially be the null check.  

```
switch (delegationStrategy) {
   case DelegationStrategy.DELEGATED_ACCOUNT.getValue():
   ....
   case DelegationStrategy.SERVICE_ACCOUNT.getValue()
   ...
   default:
      throw IllegalArgumentException("Delegation Strategy not defined");
}
```

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/CredentialPropertyDescriptors.java
pvillard31: I'd say just remove on the newly added properties and leave the existing ones as-is even if it's not consistent. We did the same in another PR we merged earlier today.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/credentials/factory/strategies/AbstractServiceAccountCredentialsStrategy.java
pvillard31: Yeah I thought that we would convert the string to the enum so that we anticipate the case where a value may be added to the enum to support another strategy and in that case the switch case would error as not all values would be covered. But I don't really mind to be honest, I'm ok with all options.

bobpaulin:
Thank you for the review @pvillard31 !  Updates made.

pvillard31 (approved):
Made some tests and it worked as expected. Assuming all checks are green, I'm a +1. Thanks for the improvement @bobpaulin

PR #9637, opened 2025-01-16
NIFI-14161: Fix QueryCassandra Processor Attribute Loss Issue

Rahul937303:
Resolved an issue in the QueryCassandra processor where attributes from the incoming FlowFile were occasionally lost in subsequent FlowFiles when the processor split the results due to a configured Fetch Size (e.g., 750). This fix ensures that all FlowFiles maintain the original attributes, even when multiple FlowFiles are generated from a single query result.


pvillard31:
Thanks for pull request! Any commit that should land on NiFi 1.x should be via a PR against the branch https://github.com/apache/nifi/tree/support/nifi-1.x. However, I don't think the community is willing to invest additional time in making new releases for the 1.x branch. It is, of course, up to any committer/PMC member to decide if such a release should happen though and if, collectively, enough people are willing to go through the release process which is quite time consuming.

joewitt:
As per the end of support notice on https://nifi.apache.org/download/ I dont think this change should proceed.

Please submit this change against the 2.x branch if appropriate.  Thanks

Rahul937303:
Hi joewitt
Currently, there is no 2.x branch available. Could you please clarify where this change should be submitted?

PR #9640, opened 2025-01-16
NIFI-14129 Add Database Dialect Service

exceptionfactory:
# Summary

[NIFI-14129](https://issues.apache.org/jira/browse/NIFI-14129) Adds a `DatabaseDialectService` Controller Service interface in a new `nifi-database-dialect-service-api` as a migration path the existing internal `DatabaseAdapter` interface and associated implementations.

The new `nifi-database-dialect-service-nar` includes a `StandardDatabaseDialectService` implementation that provides a generic implementation for the minimum Statement Types of ALTER, CREATE, and SELECT.

The introduction of the `DatabaseDialectService` applies to the following Processors:

- PutDatabaseRecord
- GenerateTableFetch
- QueryDatabaseTable
- QueryDatabaseTableRecord
- UpdateDatabaseTable

The service also applies to the `DatabaseParameterProvider`.

This pull request provides an interim transitional strategy for moving away from the `DatabaseAdapter` to the new `DatabaseDialectService`. The initial set of changes preserve existing `Database Type` properties that use `DatabaseAdapter` implementations and wraps interacts with the `DatabaseAdapter` in a new `DatabaseAdapterDatabaseDialectService`.

Existing `Database Type` properties include a new value named `Database Dialect Service` that enables the dependent `Database Dialect Service` property for Controller Service implementations.

This wrapping approach avoids the need to convert existing `DatabaseAdapter` implementations as part of this change, allowing existing Processor configurations to work without changes, while also support incremental migration of `DatabaseAdapter` implementations to new `DatabaseDialectService` implementations.

The `DatabaseDialectService` interface has a simplified contract method named `getStatement()` which takes a `StatementRequest` and returns a `StatementResponse` containing rendered SQL. The `StatementRequest` interface has standard implementations, and a `QueryStatementRequest` extended interface for describing queries. The `getSupportedStatementTypes()` provides a concise method for implementations to indicate supported statements.

Changes include modifications to existing unit tests that involved direct access to Processor methods using the `DatabaseAdapter`. Existing unit and integration tests against public methods and behaviors remain in place to validate implementation changes.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


pvillard31 (approved):
Huge PR, thanks @exceptionfactory !

I have been through extensive testing all day using Postgres and MySQL databases. I have tested many cases before and after this PR (INSERTs, UPDATEs, UPSERTs, DELETEs). Everything is working as expected and I didn't catch any regression / breaking change.

I also tried the new Standard Database Dialect controller service with basic statements and it worked well. I believe this new interface is covering all of the needs we may have in the future while still being very generic.

The approach you used to keep the "old" way and the "new" way is really cool and it is a great addition! Given the size of the PR, I'll wait until some day next week before merging in case someone else in the community wants to give it a try / to provide feedback. I'm a +1 as far as I'm concerned.

tpalfy:
Thanks for this update @exceptionfactory and @pvillard31 for the extensive testing.

I'm doing a review for the code and planning to test it on Oracle as well.
Would need a couple of days to finish if that's okay.

@exceptionfactory 
As I see the essence of the backward compatibility is duplicated in the processors:
```java
    protected DatabaseDialectService getDatabaseDialectService(final PropertyContext context) {
        final DatabaseDialectService databaseDialectService;
        final String databaseType = context.getProperty(DB_TYPE).getValue();
        if (DatabaseDialectServiceDatabaseAdapter.NAME.equals(databaseType)) {
            databaseDialectService = context.getProperty(DATABASE_DIALECT_SERVICE).asControllerService(DatabaseDialectService.class);
        } else {
            databaseDialectService = new DatabaseAdapterDatabaseDialectService(databaseType);
        }
        return databaseDialectService;
    }
```

Could we move the logic into the `DatabaseAdapterDescriptor` class or somewhere else in one place?

exceptionfactory:
> Could we move the logic into the `DatabaseAdapterDescriptor` class or somewhere else in one place?

Thanks for the initial feedback @tpalfy. The Database Type properties have different names depending on the referencing classes, which required the creation of unique Property Descriptors. I pushed an update to streamline most of the instantiation process, taking the resolved Database Type to avoid confusion related to the Database Type property descriptor.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/db/impl/DatabaseAdapterDatabaseDialectService.java
```diff
@@ -0,0 +1,218 @@
...
+                .filter(ColumnDefinition::primaryKey)
+                .map(ColumnDefinition::columnName)
+                .toList();
+        final List<ColumnDescription> columnDescriptions = getColumnDescriptions(tableDefinition);
+
+        final String sql;
+
+        if (StatementType.ALTER == statementType) {
+            final List<String> statements = databaseAdapter.getAlterTableStatements(tableDefinition.tableName(), columnDescriptions, true, true);
+            sql = statements.getFirst();
```
tpalfy, line 100: In theory a DatabaseAdapter may return a separate "ADD COLUMN" for each new individual column. The Derby adapter does this in practice (although exists only in tests).

Does this work in such cases?

If we rely on this feature being basically deprecated and support only a single ALTER statement, we might want to add some comments/docs to clarify this or change the interface altogether.

exceptionfactory: Thanks for highlighting this point @tpalfy.

To clear up the confusion, and since Apache Derby usage is focused on testing, I changed the `DatabaseAdapter` interface definition for alter tables to return a single string, aligning the approach with all other statements. This clarifies the behavior and works with existing Derby unit tests that exercise adding a single column.

nifi-extension-bundles/nifi-standard-services/nifi-database-dialect-service-api/src/main/java/org/apache/nifi/database/dialect/service/api/ColumnDefinition.java
```diff
@@ -0,0 +1,42 @@
...
+ * Database Table Column Definition
+ */
+public interface ColumnDefinition {
+    String columnName();
+
+    int dataType();
+
+    Nullable nullable();
+
+    Optional<String> defaultValue();
```
tpalfy, line 31: The `defaultValue()` doesn't seem to be called from anywhere. Is it intentional?

nifi-extension-bundles/nifi-standard-services/nifi-database-dialect-service-api/src/main/java/org/apache/nifi/database/dialect/service/api/StandardQueryStatementRequest.java
```diff
@@ -0,0 +1,48 @@
...
+ * Standard implementation of Query Statement Request with required properties
+ *
+ * @param statementType SQL Statement Type
+ * @param tableDefinition Database Table Definition
+ * @param derivedTable Derived Table Query or empty when not defined
+ * @param queryClauses Collection of Query Clauses can be empty
+ * @param pageRequest Page Request can be empty
+ */
+public record StandardQueryStatementRequest(
+        StatementType statementType,
```
tpalfy, line 33: Can the StatementType anything else but SELECT for a QueryStatementRequest?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-parameter-providers/src/main/java/org/apache/nifi/parameter/DatabaseParameterProvider.java
```diff
@@ -233,8 +223,23 @@ private void validateValueNotNull(final String value, final String columnName) {
...
+                .toList();
+        final TableDefinition tableDefinition = new TableDefinition(Optional.empty(), Optional.empty(), tableName, columnDefinitions);
+        final QueryStatementRequest queryStatementRequest = new StandardQueryStatementRequest(
+                StatementType.SELECT,
+                tableDefinition,
+                Optional.empty(),
+                List.of(new QueryClause(QueryClauseType.WHERE, whereClause)),
+                Optional.empty()
+        );
+        final StatementResponse statementResponse = databaseDialectService.getStatement(queryStatementRequest);
```
tpalfy, line 241: When the optional WHERE property is not set (null), the generated select statement will have a "WHERE null" clause, resulting to an exception (at least it did when testing on Oracle).

nifi-extension-bundles/nifi-standard-services/nifi-database-dialect-service-api/src/main/java/org/apache/nifi/database/dialect/service/api/StandardQueryStatementRequest.java
exceptionfactory: Not at this time, so it could be defaulted to `SELECT`. Other use cases would be theoretical right now, perhaps such as a `SELECT INTO` or something along those lines.

nifi-extension-bundles/nifi-standard-services/nifi-database-dialect-service-bundle/nifi-database-dialect-service/src/main/java/org/apache/nifi/database/dialect/service/StandardDatabaseDialectService.java
```diff
@@ -0,0 +1,326 @@
...
+import java.util.OptionalLong;
+import java.util.Set;
+
+@CapabilityDescription("""
+    Database Dialect Service supporting ANSI SQL.
+    Supported Statement Types: ALTER, CREATE, SELECT
+"""
+)
+@Tags({ "Relational", "Database", "JDBC", "SQL" })
+public class StandardDatabaseDialectService extends AbstractControllerService implements DatabaseDialectService {
```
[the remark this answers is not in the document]
exceptionfactory: I considered the `Generic` prefix for this service implementation, but decided to go with `Standard` as that is more common for other service implementations. I also considered other names like `ANSI`, but it seemed like `Standard` was clear, without necessarily exposing some of the previous particular naming conventions around the internal `DatabaseAdapter` interface.

nifi-extension-bundles/nifi-standard-services/nifi-database-dialect-service-api/src/main/java/org/apache/nifi/database/dialect/service/api/ColumnDefinition.java
exceptionfactory: That's a good point. This was carried over from the current structure, but as you noted, it is not used, so I will look at removing it.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/db/impl/DatabaseAdapterDatabaseDialectService.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Transitional implementation of Database Dialect Service bridging to existing Database Adapters
```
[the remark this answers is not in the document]
exceptionfactory: Based on the refactoring, this class does not need to public at all, as the instantiation is localized to the `DatabaseAdapterDescriptor`. I will update the comment to indicate that it is an internal implementation for clarity.

nifi-extension-bundles/nifi-standard-services/nifi-database-dialect-service-api/src/main/java/org/apache/nifi/database/dialect/service/api/QueryStatementRequest.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Collection;
+import java.util.Optional;
+
+/**
+ * Query extension of Statement Request with additional properties for SELECT statements
+ */
+public interface QueryStatementRequest extends StatementRequest {
+    Optional<String> derivedTable();
+
+    Collection<QueryClause> queryClauses();
```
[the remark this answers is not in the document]
exceptionfactory: Thanks for the feedback. In the initial stage of refactoring, I considered the possibility for other types of query clauses. As the implementation played out, I agree that switching from the collection to optional properties for whereClause and orderByClause would be easier to follow and maintain. I will make adjustments and push an updated version.

exceptionfactory:
Thanks for the helpful feedback @tpalfy, I pushed an update that addresses your comments. Replacing the QueryClause record with Optional values for the clauses and removing the defaultValue from ColumnDefinition helped simplify the implementation in several places.

tpalfy (approved):

tpalfy:
LGTM
Thank you for your work @exceptionfactory and @pvillard31 for the review.
Merged into main.

PR #9646, opened 2025-01-17
[NIFI-14172] - Support pasting a Flow Definition string on the canvas

rfellows:
[NIFI-14172](https://issues.apache.org/jira/browse/NIFI-14172)

Support pasting a Flow Definition string on the canvas. The Flow Definition format is the same format you would get if you download a flow from the canvas as well as the same format saved into a flow registry. Both "flavors" should be verified when reviewing this PR.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/canvas.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/canvas.component.ts
@@ -67,7 +67,7 @@ import { initialState } from '../../state/flow/flow.reducer';
 import { CanvasContextMenu } from '../../service/canvas-context-menu.service';
 import { getStatusHistoryAndOpenDialog } from '../../../../state/status-history/status-history.actions';
 import { concatLatestFrom } from '@ngrx/operators';
-import { ComponentType, isDefinedAndNotNull, selectUrl, Storage } from '@nifi/shared';
+import { ComponentType, isDefinedAndNotNull, NiFiCommon, selectUrl, Storage } from '@nifi/shared';
 import { CanvasUtils } from '../../service/canvas-utils.service';
 import { CanvasActionsService } from '../../service/canvas-actions.service';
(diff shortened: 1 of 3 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/canvas.component.ts
```diff
@@ -759,10 +760,45 @@ export class Canvas implements OnInit, OnDestroy {
...
             }
 
-            // attempting to paste something other than CopyResponseEntity
+            // check to see if this is a FlowDefinition
+            const maybeFlowDefinition: any = JSON.parse(json);
+            const isFlowDefinition = this.nifiCommon.isDefinedAndNotNull(maybeFlowDefinition.flowContents);
+            if (isFlowDefinition) {
+                // make sure the flow has some components
+                const flowHasCopiedContent = Object.entries(maybeFlowDefinition.flowContents).some((entry) => {
+                    return supportedKeys.includes(entry[0]) && Array.isArray(entry[1]) && entry[1].length > 0;
```
mcgilman, line 769:
```suggestion
                    return supportedKeys.includes(entry[0]) && Array.isArray(entry[1]);
```

Thoughts on relaxing this condition? With it as-is, someone could export an empty flow definition and be rejected when pasting calling it an `incompatible format`.

mcgilman (commented):
Thanks for the PR @rfellows! I've raised a minor item belong.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/canvas.component.ts
rfellows: Yeah, i did consider that originally but decided against it just to be consistent with the logic for verifying the CopyResponseEntity. However, these are a bit different use cases so I can make that change.

mcgilman (approved):
Thanks @rfellows!

PR #9651, opened 2025-01-20
NIFI-14176 Add runstate information to C2 FlowInfo

martinzink:
# Summary

[NIFI-14176](https://issues.apache.org/jira/browse/NIFI-14176)
Add runstate information to C2 FlowInfo

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/c2/c2-protocol/c2-protocol-api/src/main/java/org/apache/nifi/c2/protocol/api/ProcessorStatus.java
+++ b/c2/c2-protocol/c2-protocol-api/src/main/java/org/apache/nifi/c2/protocol/api/ProcessorStatus.java
@@ -37,6 +37,7 @@ public class ProcessorStatus implements Serializable {
     private long processingNanos;
     private int activeThreadCount;
     private int terminatedThreadCount;
+    private String runState;
 
     @Schema(description = "The id of the processor")
     public String getId() {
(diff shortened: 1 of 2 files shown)


c2/c2-protocol/c2-protocol-api/src/main/java/org/apache/nifi/c2/protocol/api/ProcessorStatus.java
```diff
@@ -37,6 +37,7 @@ public class ProcessorStatus implements Serializable {
     private long processingNanos;
     private int activeThreadCount;
     private int terminatedThreadCount;
+    private String runState;
```
exceptionfactory, line 40: Is there a reason for calling this `runState` as opposed to `runStatus` as named in the source Processor Status object?

martinzink: Not particularly, and I think we wanna go with runStatus instead. We did indeed implement it as runStatus in c++
https://github.com/apache/nifi-minifi-cpp/commit/07b9641ff033c1efb214d050079f4451df4e3eb9#diff-00076823b84623f3ab4d37455b5d6943cf6becc21da552be650affcae6344abdR39

To be honest, I only created this draft PR as a discussion starter, so we can see what would be the natural way to implement this in minifi java so we can implement the same thing in minifi c++.

But it turns out the C2 command to start stop components is only present in the C++ agents so this seems to be putting the cart before the horse in the case of the java agent. I will close this, and we can add this field when we implement the start/stop C2 command.

exceptionfactory: Thanks for the background @martinzink, sounds good for now.

PR #9669, opened 2025-01-27
[NIFI-14185] - upgrade angular 19, nx 20

rfellows:
# Summary

[NIFI-14185](https://issues.apache.org/jira/browse/NIFI-14185)
Migrated to Nx latest (20). This also upgraded to Angular 19. This PR is everything required for those changes to be complete.

In angular 19, the default component type is standalone. So if you do not declare it `standalone: false`, it will be true. This change had ramifications including needing to upgrade some 3rd party components to make them compatible as well as relaxing an eslint rule from error to warn when declaring a component as `standalone: false`.

There were a couple of new compiler errors that showed up as a result of the upgrade too. These were solved in this PR as well.

One particular change to mention, the mat-expand-panel had some behavior changes that made scrolling the contents of the documentation panel on load not work properly. You will see in this PR that we now initialize state of the desired expanded panel based on the current route and control its expanded state through properties on the class rather than assume the `openAll` method on the accordion will complete in time before the attempt to scroll happens.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting


nifi-frontend/src/main/frontend/.eslintrc.json
```diff
@@ -26,7 +26,8 @@
                     {
                         "suffixes": [""]
                     }
-                ]
+                ],
+                "@angular-eslint/prefer-standalone": "warn"
```
scottyaslan, line 30: Do we want to just turn this rule off?

rfellows: I considered turning it off but decided to make it a warning as it is something we should strive to resolve going forward. We should make all components standalone, these warning remind us of which ones still need converted.

scottyaslan (approved):
LGTM! Will merge once CI completes.

PR #9682, opened 2025-02-01
NIFI-14219 JoltTransform Enhancement for EL in path

yakshup:
# Summary

[NIFI-14219](https://issues.apache.org/jira/browse/NIFI-14219)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

- [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-jolt-bundle/nifi-jolt-processors/src/main/java/org/apache/nifi/processors/jolt/AbstractJoltTransform.java
+++ b/nifi-extension-bundles/nifi-jolt-bundle/nifi-jolt-processors/src/main/java/org/apache/nifi/processors/jolt/AbstractJoltTransform.java
@@ -44,6 +44,8 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


yakshup:
I have added a separate field for Jolt Spec File, as this is the standard practice from all the other modules.

exceptionfactory (changes requested):
Thanks for proposing this change @yakshup.

As it stands, however, this approach cannot go forward because it breaks the existing behavior of the `Jolt Specification` property, which supports files as a source.

In general for Property Descriptors, the Resource abstraction should be used instead of direct file access. This ensures that the framework manages file access.

The associated Jira is not quite clear in terms of the motivation for the new property. With the `Jolt Specification` property already supporting Expression Language, can you clarify the intended use case?

yakshup:
@exceptionfactory Thank you for the review and feedback.

### Clarification on the Motivation for the New Property
The primary reason for introducing a separate property for the filename is to resolve an existing issue with how `JOLT_SPEC ` handles Expression Language (EL).

Currently, the `JOLT_SPEC` property serves two distinct purposes:

- Accepting inline JOLT specifications as JSON.
- Accepting a file path (which may also contain EL).

The problem arises when the property value contains Expression Language referring to a file path. 
The current logic evaluates EL first and always assumes the result is a JSON specification, even if the value was intended to be a file path. This creates an edge case where valid file paths containing EL become invalid, as they are incorrectly parsed as JSON.

By introducing a dedicated property for the filename, we can:

- Clearly distinguish between inline JSON specs and file-based specs.
- Ensure correct handling of EL, whether used for JSON content or file paths.
- Improve maintainability by eliminating ambiguity in `JOLT_SPEC`.

**Use of the Resource Abstraction**

The existing implementation already utilizes the `readTransform` method from the Jolt abstraction layer, which ensures proper handling of file paths and JSON parsing. Additionally, necessary validations are already in place to verify that the file path is valid before processing.

Could you clarify if the concern is specifically about file access within `readTransform`, or if it is a general recommendation to adopt NiFi's Resource abstraction for consistency? If required, I can explore integrating `ResourceReference` to enhance how NiFi handles file-based properties.

exceptionfactory:
Thanks for the quick reply @yakshup.

On closer evaluation, the current behavior of `JoltTransformJSON` has an unnecessary conditional around the Expression Language evaluation. Instead of introducing a new property, removing the conditional and always evaluating EL before attempting to read the `Jolt Specification` as a Resource enables support for file paths with Expression Language.

Given the significant difference in the approach, I put together a new pull request #9684 implementing the changes. The PR includes a new unit test method that exercises the use case. If you could take a look at that approach to determine whether it enables the use case you described, that would be helpful.

yakshup:
I see the change which you have done @exceptionfactory. This was my first approach, but I had my apprehensions around keeping things in single field and design in other components of nifi. I will test your PR in my setup and get back to you.

Also, I see a lot of test cases modified and assert statements removed. Are you sure those are required to be removed.?

PR #9685, opened 2025-02-02
NIFI-14204 Add a Flow Analysis Rule for connection load balancing parameters

markobean:
# Summary

[NIFI-14204](https://issues.apache.org/jira/browse/NIFI-14204)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

Added unit tests to verify each of the load balance strategies, compression options and valid/invalid configuration of the Flow Analysis Rule.
Installed in a running NiFi and performed additional testing of various combinations of configurable options in a running flow.

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/pom.xml
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/pom.xml
@@ -23,7 +23,13 @@
     <artifactId>nifi-standard-rules</artifactId>
     <packaging>jar</packaging>
 
-    <dependencies />
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
     
     <build>
         <plugins>
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictLoadBalancing.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictLoadBalancing.java
@@ -1,0 +1,266 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.flowanalysis.rules;
+
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.ValidationContext;
+import org.apache.nifi.components.ValidationResult;
+import org.apache.nifi.controller.queue.LoadBalanceCompression;
+import org.apache.nifi.controller.queue.LoadBalanceStrategy;
+import org.apache.nifi.flow.VersionedProcessGroup;
+import org.apache.nifi.flowanalysis.AbstractFlowAnalysisRule;
+import org.apache.nifi.flowanalysis.FlowAnalysisRuleContext;
+import org.apache.nifi.flowanalysis.GroupAnalysisResult;
+import org.apache.nifi.flowanalysis.rules.util.ConnectionViolation;
+import org.apache.nifi.flowanalysis.rules.util.FlowAnalysisRuleUtils;
+import org.apache.nifi.flowanalysis.rules.util.ViolationType;
+import org.apache.nifi.processor.util.StandardValidators;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+@Tags({"connection", "load", "balance", "strategy", "compression"})
+@CapabilityDescription("This rule generates a violation when the Load Balance Strategy setting of a connection is set to a strategy which is not allowed. "
+        + "This rule may be useful for preventing certain strategy types or load-balancing compression. It may be implemented to assist in locating "
+        + "all connections which have load balancing enabled.")
+public class RestrictLoadBalancing extends AbstractFlowAnalysisRule {
+
+    static final String CONFIGURE_STRATEGY_ERROR_MESSAGE = "at least one load balance strategy property must be set to 'true'";
+    static final String COMPRESSION_CONFIGURATION_ERROR_MESSAGE =
+            "there is no compression configuration for content only without including attributes. If content compression is allowed, attribute compression must also be allowed.";
+
+    public static final PropertyDescriptor ALLOW_DO_NOT_LOAD_BALANCE = new PropertyDescriptor.Builder()
+            .name("Allow 'Do not load balance' Strategy")
+            .description("If set to true, connections may be configured with the load balancing strategy of 'Do not load balance'. This is the default "
+                    + "setting for new connections.")
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("true")
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor ALLOW_PARTITION = new PropertyDescriptor.Builder()
+            .name("Allow 'Partition by attribute' Strategy")
+            .description("If set to true, connections may be configured with the load balancing strategy of 'Partition by attribute'.")
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor ALLOW_ROUND_ROBIN = new PropertyDescriptor.Builder()
+            .name("Allow 'Round robin' Strategy")
+            .description("If set to true, connections may be configured with the load balancing strategy of 'Round robin'.")
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor ALLOW_SINGLE_NODE = new PropertyDescriptor.Builder()
+            .name("Allow 'Single node' Strategy")
+            .description("If set to true, connections may be configured with the load balancing strategy of 'Single node'.")
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor ALLOW_ATTRIBUTE_COMPRESSION = new PropertyDescriptor.Builder()
+            .name("Allow Attribute Compression")
+            .description("If set to true, connections which are configured to load balance may compress attributes. "
+                    + "This property is ignored if the only allowed Load Balance Strategy is 'Do not load balance'.")
+            .required(false)
+            .allowableValues("true", "false")
+            .defaultValue("true")
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
+    public static final PropertyDescriptor ALLOW_CONTENT_COMPRESSION = new PropertyDescriptor.Builder()
+            .name("Allow Content Compression")
+            .description("If set to true, connections which are configured to load balance may compress attributes and content. "
+                    + "This property is ignored if the only allowed Load Balance Strategy is 'Do not load balance'. "
+                    + "When set to true, 'Allow Attribute Compression' must also be set to true")
+            .required(false)
+            .allowableValues("true", "false")
+            .defaultValue("true")
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
+
+    private static final List<PropertyDescriptor> PROPERTY_DESCRIPTORS = List.of(
+            ALLOW_DO_NOT_LOAD_BALANCE,
+            ALLOW_PARTITION,
+            ALLOW_ROUND_ROBIN,
+            ALLOW_SINGLE_NODE,
+            ALLOW_ATTRIBUTE_COMPRESSION,
+            ALLOW_CONTENT_COMPRESSION
+    );
+
+    @Override
+    protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+        return PROPERTY_DESCRIPTORS;
+    }
+
+    @Override
+    protected Collection<ValidationResult> customValidate(ValidationContext validationContext) {
+        final List<ValidationResult> results = new ArrayList<>();
+
+        // For this flow analysis rule to be valid, at least one load balancing strategy must be allowed
+        if (!validationContext.getProperty(ALLOW_DO_NOT_LOAD_BALANCE).asBoolean()
+                && !isLoadBalancingAllowed(validationContext)) {
+            results.add(new ValidationResult.Builder()
+                    .subject("Load Balance Strategy Properties")
+                    .valid(false)
+                    .explanation(CONFIGURE_STRATEGY_ERROR_MESSAGE)
+                    .build());
+        }
+
+        // The only forms of compression are "attributes only" and "both attributes and content".
+        // Therefore, if content compression is allowed, attribute compression must also be allowed.
+        // It is also valid to have only attribute compression or no compression at all allowed.
+        // In addition, this is only relevant if a load balance strategy is allowed, i.e. if the only allowed strategy is "Do not load balance",
+        // then any combination of compression is allowed because it will be ignored.
+        if (isLoadBalancingAllowed(validationContext)
+                && validationContext.getProperty(ALLOW_CONTENT_COMPRESSION).isSet() && validationContext.getProperty(ALLOW_CONTENT_COMPRESSION).asBoolean()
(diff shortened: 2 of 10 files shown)


mosermw:
Thanks for working on this new Flow Analysis Rule @markobean.  I just started looking at it.  Would it be possible to combine the five src/test/resource files that include a violation into one file?  Then the unit test could configure the rule to only look for one type of violation.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/pom.xml
```diff
@@ -23,7 +23,13 @@
     <artifactId>nifi-standard-rules</artifactId>
     <packaging>jar</packaging>
 
-    <dependencies />
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-api</artifactId>
+            <scope>provided</scope>
```
mosermw, line 30: Don't need to specify the scope here, because nifi-bom declares it as provided.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictLoadBalancing.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        // The only forms of compression are "attributes only" and "both attributes and content".
+        // Therefore, if content compression is allowed, attribute compression must also be allowed.
+        // It is also valid to have only attribute compression or no compression at all allowed.
+        // In addition, this is only relevant if a load balance strategy is allowed, i.e. if the only allowed strategy is "Do not load balance",
+        // then any combination of compression is allowed because it will be ignored.
+        if (isLoadBalancingAllowed(validationContext)
+                && validationContext.getProperty(ALLOW_CONTENT_COMPRESSION).isSet() && validationContext.getProperty(ALLOW_CONTENT_COMPRESSION).asBoolean()
+                && validationContext.getProperty(ALLOW_ATTRIBUTE_COMPRESSION).isSet() && !validationContext.getProperty(ALLOW_ATTRIBUTE_COMPRESSION).asBoolean()) {
```
mosermw, line 141: Since both ALLOW_CONTENT_COMPRESSION and ALLOW_ATTRIBUTE_COMPRESSION have allowable values, they will always be set so I don't think you need the `isSet()` check.

markobean:
> Thanks for working on this new Flow Analysis Rule @markobean. I just started looking at it. Would it be possible to combine the five src/test/resource files that include a violation into one file? Then the unit test could configure the rule to only look for one type of violation.

I updated the unit tests to all use a single flow definition file. Due to the nature of combining load balancing strategy and compression, some unit tests have more than a single connection in violation. I did create static variables that are more descriptive of the component in violation. Hopefully, it is clear.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/pom.xml
markobean: Removed.

exceptionfactory (changes requested):
Thanks for proposing this rule @markobean.

I made some comments around the property definitions, as the number of Boolean values creates a variety of options.

The proposed description mentions locating connections with load balancing, but that should not be the primary purpose of analysis rules.

On initial review, the rule itself seems too broad. Connection Load Balancing makes sense at very particular parts of a flow, and does not make sense in others. One anti-pattern is load balancing at every Processor, as opposed to just an initial source. Having a general rule that checks for the presence of load balancing one way or the other doesn't provide enough granularity to determine whether the implementation is good or bad.

Are there other use cases where this generalized approach fits?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/pom.xml
mosermw: I looked into this, too, @exceptionfactory.  I found that the org.apache.nifi.controller.queue.LoadBalanceStrategy and org.apache.nifi.controller.queue.LoadBalanceCompression classes used in this rule are located in nifi-framework-api.

exceptionfactory: @mosermw, although those classes are available at that level, general usage is restricted to components that implement framework-level interfaces. This seems like a pattern to avoid for Flow Analysis Rules, as these implement `nifi-api` interfaces as opposed to `nifi-framework-api` interfaces.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictLoadBalancing.java
```diff
@@ -0,0 +1,266 @@
...
+        + "This rule may be useful for preventing certain strategy types or load-balancing compression. It may be implemented to assist in locating "
+        + "all connections which have load balancing enabled.")
+public class RestrictLoadBalancing extends AbstractFlowAnalysisRule {
+
+    static final String CONFIGURE_STRATEGY_ERROR_MESSAGE = "at least one load balance strategy property must be set to 'true'";
+    static final String COMPRESSION_CONFIGURATION_ERROR_MESSAGE =
+            "there is no compression configuration for content only without including attributes. If content compression is allowed, attribute compression must also be allowed.";
+
+    public static final PropertyDescriptor ALLOW_DO_NOT_LOAD_BALANCE = new PropertyDescriptor.Builder()
+            .name("Allow 'Do not load balance' Strategy")
```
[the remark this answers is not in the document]
markobean: Without the quotes, the property would read as:

Allow 'Do not load balance' Strategy

versus

Allow Do not load balance Strategy

The inconsistent capitalization offends me more than single quotes. Would it help if the quotes only appeared in `.displayName()` versus the actual name of the `PropertyDescriptor`?

exceptionfactory:
> > Thanks for proposing this rule @markobean.
> > I made some comments around the property definitions, as the number of Boolean values creates a variety of options.
> > The proposed description mentions locating connections with load balancing, but that should not be the primary purpose of analysis rules.
> > On initial review, the rule itself seems too broad. Connection Load Balancing makes sense at very particular parts of a flow, and does not make sense in others. One anti-pattern is load balancing at every Processor, as opposed to just an initial source. Having a general rule that checks for the presence of load balancing one way or the other doesn't provide enough granularity to determine whether the implementation is good or bad.
> > Are there other use cases where this generalized approach fits?
> 
> I envisioned this particular rule (and maybe others) as a way to spot check your flow. This is why it is great the rules have an enable/disable feature. It doesn't have to be left enabled all the time, but rather enable it to perform analysis and then disable it again. This may be more appropriate for the warning type violation, not enforcement.
> 
> I see your point though. One way around this would be to apply Flow Analysis Rules to specific Process Group(s). Doing so would be well beyond the scope of this ticket. I recommend this PR continue with the existing Rules framework - good and bad. (I already have another Jira issue to improve the way violations on a connection work.)
> 
> There is an immediately use case for using this rule to quickly identify all locations where load balancing is employed. As you say, this should be implemented judiciously for obvious performance reasons. This rule will aid in finding an over-abundance of load balanced connections.

Thanks for the additional background @markobean, that provides helpful context for consideration.

I agree with you that being able to scope rules analysis to the Process Group level would be useful in this scenario. Implementing it is not trivial as you noted, and some past work around this didn't go forward due to initial usability concerns, so it is an open area.

Although the Flow Analysis Rules space has fewer examples right now, this is an area where it is important to provide generally usable and understandable components, just like Processors. In other words, too many knobs can be counter-productive long term. For that reason, and for the use of the `nifi-framework-api`, I don't think this rule provides a good general pattern as it stands.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictLoadBalancing.java
```diff
@@ -0,0 +1,266 @@
...
+            .build();
+
+    public static final PropertyDescriptor ALLOW_SINGLE_NODE = new PropertyDescriptor.Builder()
+            .name("Allow 'Single node' Strategy")
+            .description("If set to true, connections may be configured with the load balancing strategy of 'Single node'.")
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
```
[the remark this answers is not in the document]
markobean: I do not like the approach of selecting only a single type of load balancing. The flow may dictate having different strategies in different locations of the flow. However, I will work towards replacing the first property (`Do not load balance`) with a generalized `Load Balancing Strategy Allowed`.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/pom.xml
mosermw: This code has to check the return value of connection.getLoadBalanceStrategy() against configured strategies to see if it would be allowed/disallowed.  An alternative to referencing nifi-framework-api is to hard-code those strings directly in this rule.

I suspect the LoadBalanceStrategy and LoadBalanceCompression classes belong in nifi-api.  But that's a separate discussion ;).

markobean:
Based on previous comments and a general theme that emerged, I did some refactoring. Now, there is a property `Load Balancing Policy`. Setting it to `Allowed` opens access to all the remaining properties. It effectively replaces the old property, `Allow 'Do not load balance' Strategy`.

I also optimized away the convoluted logic related compression of only attributes or attributes plus content.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/pom.xml
exceptionfactory: The reason is helpful, but the concern is not the dependency size, as it is provided, but with the logical distinction between framework components and extension components. Under the circumstances, using String values instead of adding the dependency would be the best way to go.

markobean: @exceptionfactory I removed the `nifi-framework-api` dependency.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictLoadBalancing.java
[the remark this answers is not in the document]
markobean: Refactor eliminated the quotes in the property  name.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictLoadBalancing.java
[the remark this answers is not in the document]
markobean: Refactor created a property `Load Balancing Policy`. It is the only property for the rule if set for `Disallowed`. If set for `Allowed`, additional properties for load balancing strategies and compression are available.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RestrictLoadBalancing.java
```diff
@@ -0,0 +1,266 @@
...
+
+        // For this flow analysis rule to be valid, at least one load balancing strategy must be allowed
+        if (!validationContext.getProperty(ALLOW_DO_NOT_LOAD_BALANCE).asBoolean()
+                && !isLoadBalancingAllowed(validationContext)) {
+            results.add(new ValidationResult.Builder()
+                    .subject("Load Balance Strategy Properties")
+                    .valid(false)
+                    .explanation(CONFIGURE_STRATEGY_ERROR_MESSAGE)
+                    .build());
+        }
```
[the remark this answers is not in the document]
markobean: As recommended, a policy property was added in the recent refactor.

exceptionfactory (changes requested):
Thanks for the updates @markobean.

The property naming looks better, and the removing the dependency on `nifi-framework-api` addresses that concern.

However, as mentioned earlier, the fundamental approach of the rule does not seem like something that should be included for ongoing maintenance in the project. Allowing or disallowing load balancing across the board does not provide a useful level of granularity, so as it stands, I'm a -1 on the overall addition at this time.

exceptionfactory:
Thanks for the reply @mosermw, see responses in line:
 
> I support users that temporarily add load balanced connections to help distribute a rare spike in flowfile load (think of a Split processor that normally has no trouble but will rarely output millions of flowfiles). They load balance a connection to work off a load, then 2 hours later want to remove that load balancing later. A Flow Analysis Rule such as this (in WARN enforcement mode) can ensure that this isn't forgotten and mistakenly left in place (which happens a lot).

In this scenario, is there a reason for not having load balancing enabled all the time? Having to turn it on and off seems questionable, but perhaps there is some other driving reason. In this scenario, is the desire that load balancing is never enabled anywhere in the flow?

> 
> Another use case is a NiFi system owner has given access to tenants and doesn't want those tenants to use certain load balanced connection configurations, or load balancing at all.

Disallowing all load balancing could be a reason for a streamlined version of this rule, but then the configuration properties could be simplified. Allowing only specific types of load balancing across all flow configurations seems to broad of a limitation, which is part of the fundamental concern.

> 
> Another use case can be the desire to require compression on attributes+content because the flow manager knows that their data is easily compressible and wants to ensure this efficiency is always enabled.

Would the current approach support this use case? The properties allow compression, but would not require it if I am following. Also, this would require the same compression setting for all load balanced connections.

> 
> With many people modifying a NiFi graph, these Flow Analysis Rules are a great tool for performing some quality checks.

I agree with you that performing quality checks is a value, the current problem is that the rule implementation doesn't provide enough fine-grained control. This could arguably be a limitation of the current framework implementation for evaluating all rules across all process groups, but that's where things stand.

markobean:
> Thanks for the reply @mosermw, see responses in line:
> 
> > I support users that temporarily add load balanced connections to help distribute a rare spike in flowfile load (think of a Split processor that normally has no trouble but will rarely output millions of flowfiles). They load balance a connection to work off a load, then 2 hours later want to remove that load balancing later. A Flow Analysis Rule such as this (in WARN enforcement mode) can ensure that this isn't forgotten and mistakenly left in place (which happens a lot).
> 
> In this scenario, is there a reason for not having load balancing enabled all the time? Having to turn it on and off seems questionable, but perhaps there is some other driving reason. In this scenario, is the desire that load balancing is never enabled anywhere in the flow?
> 
> > Another use case is a NiFi system owner has given access to tenants and doesn't want those tenants to use certain load balanced connection configurations, or load balancing at all.
> 
> Disallowing all load balancing could be a reason for a streamlined version of this rule, but then the configuration properties could be simplified. Allowing only specific types of load balancing across all flow configurations seems to broad of a limitation, which is part of the fundamental concern.
> 
> > Another use case can be the desire to require compression on attributes+content because the flow manager knows that their data is easily compressible and wants to ensure this efficiency is always enabled.
> 
> Would the current approach support this use case? The properties allow compression, but would not require it if I am following. Also, this would require the same compression setting for all load balanced connections.
> 
> > With many people modifying a NiFi graph, these Flow Analysis Rules are a great tool for performing some quality checks.
> 
> I agree with you that performing quality checks is a value, the current problem is that the rule implementation doesn't provide enough fine-grained control. This could arguably be a limitation of the current framework implementation for evaluating all rules across all process groups, but that's where things stand.

I agree with use cases that @mwmoser presented. @exceptionfactory you are correct that the submitted implementation and the compression use case outlined is will not work quite the way we'd like for that use case. For it to work, there would have to be a `Required` state in addition to `Allowed` and `Disallowed`. The best we could do would be negative testing; set compression for `Disallowed` and verify the list of violations covers all load balanced connections.

I could add an additional state if this use case is strong desired. But I won't spend the cycles if it still going to receive a "-1".

The original use case - which is still a powerful one that has been requested for years - is to use this rule as a mechanism to locate all load balanced connections. Currently, there is no way to do that short of manually parsing the flow.json.gz file. For this reason alone, I find this Flow Analysis Rule a significant improvement. For this reason alone, I find it unreasonable to reject even if not yet a perfect solution (with many of the arguments against being within framework, not the rule.)

exceptionfactory:
> I could add an additional state if this use case is strongly desired. But I won't spend the cycles if it is still going to receive a "-1".

Yes, it doesn't sound like making further changes at the point would be worth pursuing.

> 
> The original use case - which is still a powerful one that has been requested for years - is to use this rule as a mechanism to locate all load balanced connections. Currently, there is no way to do that short of manually parsing the flow.json.gz file. For this reason alone, I find this Flow Analysis Rule a significant improvement. For this reason alone, I find it unreasonable to reject even if not yet a perfect solution (with many of the arguments against being within framework, not the rule.)

Thanks for focusing the goal of the current rule. The purpose of Flow Analysis Rules should not be finding component configuration properties, but finding settings that are not aligned with flow design requirements or recommendations.

As an extension point, the Rule API can be used for any number of things, so as a custom implementation, this rule may meet that need. However, as a way to fill a gap in functionality, this approach seems to go outside the general purpose of Flow Analysis Rules. For these reasons, it does not appear to be a good fit for inclusion in the standard set of rules.

exceptionfactory:
> The original use case - which is still a powerful one that has been requested for years - is to use this rule as a mechanism to locate all load balanced connections. Currently, there is no way to do that short of manually parsing the flow.json.gz file. For this reason alone, I find this Flow Analysis Rule a significant improvement. For this reason alone, I find it unreasonable to reject even if not yet a perfect solution (with many of the arguments against being within framework, not the rule.)

Is there an open Jira issue for finding load-balanced connections? That would be a helpful point of reference and description for the desired capability.

markobean:
> > The original use case - which is still a powerful one that has been requested for years - is to use this rule as a mechanism to locate all load balanced connections. Currently, there is no way to do that short of manually parsing the flow.json.gz file. For this reason alone, I find this Flow Analysis Rule a significant improvement. For this reason alone, I find it unreasonable to reject even if not yet a perfect solution (with many of the arguments against being within framework, not the rule.)
> 
> Is there an open Jira issue for finding load-balanced connections? That would be a helpful point of reference and description for the desired capability.

I did not find a Jira issue, so I created a new one:
https://issues.apache.org/jira/browse/NIFI-14243

exceptionfactory:
Thanks for creating the new Jira issue @markobean! Closing this PR based on the discussion.

PR #9693, opened 2025-02-05
NIFI-14235: Additional updates following Angular 19 upgrade

mcgilman:
- Clean up unused imports.
- Fix nested sass rules.
- Disabling prefer-standalone.


rfellows (changes requested):
This is good, except there are still some runtime angular-sass warnings when you fire up `update-attribute` and `standard-content-viewer` custom ui's. We should get those updated here as well.

rfellows (approved):
Nice work @mcgilman 👍 . It is good to see those warnings disappear since the nx/angular upgrade.

PR #9695, opened 2025-02-05
NIFI-14238 - Add ConsumeBoxEvents processor

pvillard31:
# Summary

[NIFI-14238](https://issues.apache.org/jira/browse/NIFI-14238) - Add ConsumeBoxEvents processor

This pull request adds a new processor that will consume all Box events and output the content of the events as an array of JSON records. In addition to this processor, the pull request also adds another controller service to connect to the Box API using the developer token option. This option is used only for testing as this token expires after 60 minutes but it allows testing the Box components in NiFi without the need for a Business/Enterprise account.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
@@ -21,7 +21,6 @@ import com.box.sdk.BoxEvent;
 import com.box.sdk.EventListener;
 import com.box.sdk.EventStream;
 import com.eclipsesource.json.Json;
-import com.eclipsesource.json.JsonArray;
 import com.eclipsesource.json.JsonObject;
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
@@ -52,6 +51,7 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -61,17 +61,21 @@ import java.util.concurrent.LinkedBlockingQueue;
 @PrimaryNodeOnly
 @TriggerSerially
 @Tags({"box", "storage"})
-@CapabilityDescription("Captures all events from Box. This processor can be used to capture events such as uploads, modifications, deletions, etc. The content "
-        + "of the events is sent to the 'success' relationship as a JSON array.")
+@CapabilityDescription("""
+        Captures all events from Box. This processor can be used to capture events such as uploads, modifications, deletions, etc.
+        The content of the events is sent to the 'success' relationship as a JSON array.
+        """)
 @SeeAlso({ FetchBoxFile.class, PutBoxFile.class, ListBoxFile.class })
 @InputRequirement(Requirement.INPUT_FORBIDDEN)
 public class CaptureBoxEvents extends AbstractProcessor implements VerifiableProcessor {
 
     public static final PropertyDescriptor MAX_MESSAGE_QUEUE_SIZE = new PropertyDescriptor.Builder()
-            .name("Max Size of Message Queue")
-            .description("The maximum size of the internal queue used to buffer events being transferred from the underlying stream to the processor. " +
-                    "Setting this value higher allows more messages to be buffered in memory during surges of incoming messages, but increases the total " +
-                    "memory used by the processor during these surges.")
+            .name("Queue Capacity")
+            .description("""
+                    The maximum size of the internal queue used to buffer events being transferred from the underlying stream to the processor.
+                    Setting this value higher allows more messages to be buffered in memory during surges of incoming messages, but increases the total
+                    memory used by the processor during these surges.
+                    """)
             .addValidator(StandardValidators.POSITIVE_INTEGER_VALIDATOR)
             .defaultValue("10000")
             .required(true)
@@ -118,7 +122,7 @@ public class CaptureBoxEvents extends AbstractProcessor implements VerifiablePro
             @Override
             public void onEvent(BoxEvent event) {
                 if (!events.offer(event)) {
-                    getLogger().warn("Failed to add event to queue. Queue is full.");
+                    getLogger().warn("Failed to add event (ID = {}) to queue. Queue is full.", event.getID());
                 }
             }
 
@@ -159,7 +163,7 @@ public class CaptureBoxEvents extends AbstractProcessor implements VerifiablePro
                     .explanation("Successfully validated Box connection")
                     .build());
         } catch (Exception e) {
-            getLogger().warn("Failed to verify configuration.", e);
+            getLogger().warn("Failed to verify configuration", e);
             results.add(new ConfigVerificationResult.Builder()
                     .verificationStepName("Box API Connection")
                     .outcome(Outcome.FAILED)
@@ -178,30 +182,32 @@ public class CaptureBoxEvents extends AbstractProcessor implements VerifiablePro
         }
 
         final FlowFile flowFile = session.create();
-        final List<BoxEvent> eventList = new ArrayList<>();
-        final int nbEvents = events.drainTo(eventList);
+        final List<BoxEvent> boxEvents = new ArrayList<>();
+        final int recordCount = events.drainTo(boxEvents);
 
-        try {
-            // converting events to JSON using the library coming with Box SDK
-            final JsonArray jsonEvents = Json.array();
-            for (BoxEvent event : eventList) {
-                jsonEvents.add(toRecord(event));
+        try (final OutputStream out = session.write(flowFile)) {
+            final Writer writer = new OutputStreamWriter(out, StandardCharsets.UTF_8);
+            writer.write("[");
+            final Iterator<BoxEvent> iterator = boxEvents.iterator();
+            while (iterator.hasNext()) {
+                BoxEvent event = iterator.next();
+                JsonObject jsonEvent = toRecord(event);
+                jsonEvent.writeTo(writer);
+                if (iterator.hasNext()) {
+                    writer.write(",");
+                }
             }
-
-            try (final OutputStream out = session.write(flowFile)) {
-                final Writer writer = new OutputStreamWriter(out, StandardCharsets.UTF_8);
-                jsonEvents.writeTo(writer);
-                writer.flush();
-            }
+            writer.write("]");
+            writer.flush();
         } catch (Exception e) {
-            getLogger().error("Failed to write events to FlowFile due to {}; will re-queue events and try again", e.getMessage(), e);
-            eventList.forEach(events::offer);
+            getLogger().error("Failed to write events to FlowFile; will re-queue events and try again", e);
+            boxEvents.forEach(events::offer);
             session.remove(flowFile);
             context.yield();
             return;
         }
 
-        session.putAttribute(flowFile, "record.count", String.valueOf(nbEvents));
+        session.putAttribute(flowFile, "record.count", String.valueOf(recordCount));
         session.putAttribute(flowFile, CoreAttributes.MIME_TYPE.key(), "application/json");
         session.transfer(flowFile, REL_SUCCESS);
     }
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
@@ -57,6 +57,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicLong;
 
 @PrimaryNodeOnly
 @TriggerSerially
@@ -114,11 +115,12 @@ public class CaptureBoxEvents extends AbstractProcessor implements VerifiablePro
         eventStream = new EventStream(boxAPIConnection);
 
         final int eventsCapacity = context.getProperty(MAX_MESSAGE_QUEUE_SIZE).asInteger();
-        if (events == null) {
-            events = new LinkedBlockingQueue<>(eventsCapacity);
-        }
+        events = new LinkedBlockingQueue<>(eventsCapacity);
 
         eventStream.addListener(new EventListener() {
+
+            private final AtomicLong position = new AtomicLong(0);
+
             @Override
             public void onEvent(BoxEvent event) {
                 if (!events.offer(event)) {
@@ -127,13 +129,14 @@ public class CaptureBoxEvents extends AbstractProcessor implements VerifiablePro
             }
 
             @Override
-            public void onNextPosition(long position) {
+            public void onNextPosition(long pos) {
+                position.set(pos);
                 getLogger().debug("Next position: {}", position);
             }
 
             @Override
             public boolean onException(Throwable e) {
-                getLogger().warn("An error has been received from the stream.", e);
+                getLogger().warn("An error has been received from the stream. Last tracked position {}", position.get(), e);
                 return true;
             }
         });
@@ -145,6 +148,7 @@ public class CaptureBoxEvents extends AbstractProcessor implements VerifiablePro
     public void stopped() {
         if (eventStream != null && eventStream.isStarted()) {
             eventStream.stop();
+            events.clear();
         }
     }
 
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
@@ -1,0 +1,227 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.box;
+
+import com.box.sdk.BoxAPIConnection;
+import com.box.sdk.BoxEvent;
+import com.box.sdk.EventListener;
+import com.box.sdk.EventStream;
+import com.eclipsesource.json.Json;
+import com.eclipsesource.json.JsonArray;
+import com.eclipsesource.json.JsonObject;
+import org.apache.nifi.annotation.behavior.InputRequirement;
+import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
+import org.apache.nifi.annotation.behavior.PrimaryNodeOnly;
+import org.apache.nifi.annotation.behavior.TriggerSerially;
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.SeeAlso;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.annotation.lifecycle.OnScheduled;
+import org.apache.nifi.annotation.lifecycle.OnStopped;
+import org.apache.nifi.box.controllerservices.BoxClientService;
+import org.apache.nifi.components.ConfigVerificationResult;
+import org.apache.nifi.components.ConfigVerificationResult.Outcome;
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.flowfile.FlowFile;
+import org.apache.nifi.flowfile.attributes.CoreAttributes;
+import org.apache.nifi.logging.ComponentLog;
+import org.apache.nifi.processor.AbstractProcessor;
+import org.apache.nifi.processor.ProcessContext;
+import org.apache.nifi.processor.ProcessSession;
+import org.apache.nifi.processor.Relationship;
+import org.apache.nifi.processor.VerifiableProcessor;
+import org.apache.nifi.processor.exception.ProcessException;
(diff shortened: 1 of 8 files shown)


nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+
+    @Override
+    public void onTrigger(ProcessContext context, ProcessSession session) throws ProcessException {
+        if (events.isEmpty()) {
+            context.yield();
+            return;
+        }
+
+        final FlowFile flowFile = session.create();
+        final List<BoxEvent> eventList = new ArrayList<>();
```
exceptionfactory, line 181: Recommend naming this `boxEvents`.
```suggestion
        final List<BoxEvent> boxEvents = new ArrayList<>();
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+        final int nbEvents = events.drainTo(eventList);
+
+        try {
+            // converting events to JSON using the library coming with Box SDK
+            final JsonArray jsonEvents = Json.array();
+            for (BoxEvent event : eventList) {
+                jsonEvents.add(toRecord(event));
+            }
```
exceptionfactory, line 189: This approach ends up creating a new object for every `BoxEvent`, which has some memory impacts. Instead of this approach, it would be much more efficient to use a streaming JSON writer to add each event to the output stream.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+
+            try (final OutputStream out = session.write(flowFile)) {
+                final Writer writer = new OutputStreamWriter(out, StandardCharsets.UTF_8);
+                jsonEvents.writeTo(writer);
+                writer.flush();
+            }
+        } catch (Exception e) {
+            getLogger().error("Failed to write events to FlowFile due to {}; will re-queue events and try again", e.getMessage(), e);
```
exceptionfactory, line 197: Recommend avoiding the `due to {}` approach since the message is already included in the exception stack trace.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+    @Override
+    public void onTrigger(ProcessContext context, ProcessSession session) throws ProcessException {
+        if (events.isEmpty()) {
+            context.yield();
+            return;
+        }
+
+        final FlowFile flowFile = session.create();
+        final List<BoxEvent> eventList = new ArrayList<>();
+        final int nbEvents = events.drainTo(eventList);
```
exceptionfactory, line 182: Recommend `recordCount` to align with the attribute naming.
```suggestion
        final int recordCount = events.drainTo(eventList);
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+
+        try {
+            boxAPIConnection.refresh();
+            results.add(new ConfigVerificationResult.Builder()
+                    .verificationStepName("Box API Connection")
+                    .outcome(Outcome.SUCCESSFUL)
+                    .explanation("Successfully validated Box connection")
+                    .build());
+        } catch (Exception e) {
+            getLogger().warn("Failed to verify configuration.", e);
```
exceptionfactory, line 162: Recommend avoiding log messages ending with a period `.` character.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+        final int eventsCapacity = context.getProperty(MAX_MESSAGE_QUEUE_SIZE).asInteger();
+        if (events == null) {
+            events = new LinkedBlockingQueue<>(eventsCapacity);
+        }
+
+        eventStream.addListener(new EventListener() {
+            @Override
+            public void onEvent(BoxEvent event) {
+                if (!events.offer(event)) {
+                    getLogger().warn("Failed to add event to queue. Queue is full.");
```
exceptionfactory, line 121: Does the EventListener contract support throwing any kind of exception? It seems like this could end up generating a lot of messages in some cases. Also recommend including the event ID or some marker instead of the static message.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+            }
+
+            @Override
+            public void onNextPosition(long position) {
+                getLogger().debug("Next position: {}", position);
+            }
+
+            @Override
+            public boolean onException(Throwable e) {
+                getLogger().warn("An error has been received from the stream.", e);
```
exceptionfactory, line 132: Recommend tracking the `position` so that it could be logged.
```suggestion
                getLogger().warn("Event stream processing failed at position {}", position, e);
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @OnScheduled
+    public void onScheduled(final ProcessContext context) {
+        final BoxClientService boxClientService = context.getProperty(BoxClientService.BOX_CLIENT_SERVICE).asControllerService(BoxClientService.class);
+        boxAPIConnection = boxClientService.getBoxApiConnection();
+        eventStream = new EventStream(boxAPIConnection);
+
+        final int eventsCapacity = context.getProperty(MAX_MESSAGE_QUEUE_SIZE).asInteger();
+        if (events == null) {
```
exceptionfactory, line 113: Is the purpose of the null check to avoid clearing any queued events?

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.LinkedBlockingQueue;
+
+@PrimaryNodeOnly
+@TriggerSerially
+@Tags({"box", "storage"})
+@CapabilityDescription("Captures all events from Box. This processor can be used to capture events such as uploads, modifications, deletions, etc. The content "
+        + "of the events is sent to the 'success' relationship as a JSON array.")
```
exceptionfactory, line 65: Multiline strings can be used instead of concatenation.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+@SeeAlso({ FetchBoxFile.class, PutBoxFile.class, ListBoxFile.class })
+@InputRequirement(Requirement.INPUT_FORBIDDEN)
+public class CaptureBoxEvents extends AbstractProcessor implements VerifiableProcessor {
+
+    public static final PropertyDescriptor MAX_MESSAGE_QUEUE_SIZE = new PropertyDescriptor.Builder()
+            .name("Max Size of Message Queue")
+            .description("The maximum size of the internal queue used to buffer events being transferred from the underlying stream to the processor. " +
+                    "Setting this value higher allows more messages to be buffered in memory during surges of incoming messages, but increases the total " +
+                    "memory used by the processor during these surges.")
```
exceptionfactory, line 74: Multiline strings can be used.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,227 @@
...
+@TriggerSerially
+@Tags({"box", "storage"})
+@CapabilityDescription("Captures all events from Box. This processor can be used to capture events such as uploads, modifications, deletions, etc. The content "
+        + "of the events is sent to the 'success' relationship as a JSON array.")
+@SeeAlso({ FetchBoxFile.class, PutBoxFile.class, ListBoxFile.class })
+@InputRequirement(Requirement.INPUT_FORBIDDEN)
+public class CaptureBoxEvents extends AbstractProcessor implements VerifiableProcessor {
+
+    public static final PropertyDescriptor MAX_MESSAGE_QUEUE_SIZE = new PropertyDescriptor.Builder()
+            .name("Max Size of Message Queue")
```
exceptionfactory, line 71: Recommend `Queue Capacity` for a shorter name.
```suggestion
            .name("Queue Capacity")
```

exceptionfactory (changes requested):
Thanks for putting together this new Processor @pvillard31. The Box API seems fairly straightforward. I noted several recommendations on naming. On the general approach, I'm concerned about creating a new array of JSON objects in memory, instead of streaming the currently queued events.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
pvillard31: Yes, I figured that in case we stop the processor, we may not want to clear the queued events. Should I instead clear the queue and just drop what we already consumed? In any case we would have ignored the events while the processor was stopped...

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
pvillard31: We do not have the position when an exception is thrown. We have it in `onNextPosition` and it's already logged there. Or am I missing something?

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
pvillard31: I don't see anything unfortunately - if we throw an exception here, it looks like we would just drop all of the events that were pulled in the background. Will add the event ID in the log message.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
exceptionfactory: Right, the change would involve tracking the last position from `onNextPosition` in an AtomicLong, so it could be pulled and referenced in this warning. That could be helpful when debugging is not enabled.

pvillard31: Yep, understood. I looked a bit more into the Box code to see how this position is tracked and I see how it can be useful in certain scenarios.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
pvillard31: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,237 @@
...
+    @Override
+    public final Set<Relationship> getRelationships() {
+        return RELATIONSHIPS;
+    }
+
+    @OnScheduled
+    public void onScheduled(final ProcessContext context) {
+        final BoxClientService boxClientService = context.getProperty(BoxClientService.BOX_CLIENT_SERVICE).asControllerService(BoxClientService.class);
+        boxAPIConnection = boxClientService.getBoxApiConnection();
+        eventStream = new EventStream(boxAPIConnection);
```
exceptionfactory, line 115: This constructor starts processing events at the current time according to the [EventStream](https://github.com/box/box-java-sdk/blob/f6c37ce61ed02eb21ce6eb6ec5d23ffcabbb76d2/src/main/java/com/box/sdk/EventStream.java#L46) class constructor.

This seems like a reasonable initial implementation, but are there any concerns around picking up from the last position on start or stop using state information?

exceptionfactory (commented):
Thanks for making the adjustments @pvillard31, the latest version looks close to completion.

I raised one more comment about state tracking.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
pvillard31: I didn't even notice that it was possible to resume at a specific position, things make much more sense now. I've changed the code to not clear the queue if the processor is stopped and reuse the queue if restarted, this way stopping/starting the processor would not cause missing events. It could still happen in case of a restart though.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+        or if the queue capacity is exceeded. The last known position of the Box stream is stored in the processor state and is used to
+        resume the stream from the last known position when the processor is restarted.
+        """)
+@SeeAlso({ FetchBoxFile.class, PutBoxFile.class, ListBoxFile.class })
+@InputRequirement(Requirement.INPUT_FORBIDDEN)
+@Stateful(description = """
+        The last known position of the Box stream is stored in the processor state and is used to
+        resume the stream from the last known position when the processor is restarted.
+        """, scopes = { Scope.CLUSTER })
+public class CaptureBoxEvents extends AbstractProcessor implements VerifiableProcessor {
```
exceptionfactory, line 80: Reviewing this once more, it seems like the more appropriate name would be `ConsumeBoxEvents`, following the convention of other event processors.
```suggestion
public class ConsumeBoxEvents extends AbstractProcessor implements VerifiableProcessor {
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+
+    private final static String POSITION_KEY = "position";
+
+    public static final PropertyDescriptor MAX_MESSAGE_QUEUE_SIZE = new PropertyDescriptor.Builder()
```
exceptionfactory, line 84:
```suggestion
    public static final PropertyDescriptor QUEUE_CAPACITY = new PropertyDescriptor.Builder()
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+        boxAPIConnection = boxClientService.getBoxApiConnection();
+
+        try {
+            final String position = context.getStateManager().getState(Scope.CLUSTER).get(POSITION_KEY);
+            if (position != null) {
+                // we resume from the last known position
+                eventStream = new EventStream(boxAPIConnection, Long.parseLong(position));
+            } else {
+                eventStream = new EventStream(boxAPIConnection);
+            }
```
exceptionfactory, line 135: Recommend reversing the condition for readability:
```suggestion
            if (position == null) {
                eventStream = new EventStream(boxAPIConnection);
            } else {
                eventStream = new EventStream(boxAPIConnection, Long.parseLong(position));
            }
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+        } catch (Exception e) {
+            throw new ProcessException("Could not retrieve processor state", e);
```
exceptionfactory, line 137:
```suggestion
            throw new ProcessException("Could not retrieve last event position", e);
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        final int eventsCapacity = context.getProperty(MAX_MESSAGE_QUEUE_SIZE).asInteger();
```
exceptionfactory, line 140:
```suggestion
        final int queueCapacity = context.getProperty(MAX_MESSAGE_QUEUE_SIZE).asInteger();
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+            final LinkedBlockingQueue<BoxEvent> newQueue = new LinkedBlockingQueue<>(eventsCapacity);
+            newQueue.addAll(events);
+            events = newQueue;
+        }
+
+        eventStream.addListener(new EventListener() {
+
+            @Override
+            public void onEvent(BoxEvent event) {
+                if (!events.offer(event)) {
```
exceptionfactory, line 154: Reviewing the `EventStream` behavior, it seems like a better option would be to use `put()`, which would block the EventStream Thread until the events queue has capacity. This would avoid errors, and dropping data.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+
+            @Override
+            public void onNextPosition(long pos) {
+                try {
+                    context.getStateManager().setState(Map.of(POSITION_KEY, String.valueOf(pos)), Scope.CLUSTER);
+                    position.set(pos);
+                } catch (IOException e) {
+                    getLogger().warn("Failed to save position in processor state", e);
+                }
+                getLogger().debug("Next position: {}", position);
```
exceptionfactory, line 167: This seems unnecessary now that the position is tracked in state.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+                }
+            }
+
+            @Override
+            public void onNextPosition(long pos) {
+                try {
+                    context.getStateManager().setState(Map.of(POSITION_KEY, String.valueOf(pos)), Scope.CLUSTER);
+                    position.set(pos);
+                } catch (IOException e) {
+                    getLogger().warn("Failed to save position in processor state", e);
```
exceptionfactory, line 165:
```suggestion
                    getLogger().warn("Failed to save position [{}] in processor state", pos, e);
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
```diff
@@ -0,0 +1,270 @@
...
+            context.yield();
+            return;
+        }
+
+        session.putAttribute(flowFile, "record.count", String.valueOf(recordCount));
+        session.putAttribute(flowFile, CoreAttributes.MIME_TYPE.key(), "application/json");
+        session.transfer(flowFile, REL_SUCCESS);
+    }
+
+    private JsonObject toRecord(BoxEvent event) {
```
exceptionfactory, line 252: Is there a reason for creating the new `JsonObject` as opposed to using the `getSourceJson()` method on `BoxEvent`?

exceptionfactory (changes requested):
Thanks for implementing the state tracking @pvillard31! I noted a few more convention-related comments.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CaptureBoxEvents.java
pvillard31: The `getSourceJson()` method only returns a subset of the data contained in the BoxEvent object. We actually use it in this `toRecord` method:
````java
json.add("source", Objects.requireNonNullElse(event.getSourceJSON(), Json.NULL));
````

exceptionfactory (approved):
Thanks for working through the feedback and making the latest round of updates @pvillard31, the latest version looks good! +1 merging

PR #9696, opened 2025-02-05
[NIFI-14239] - Address Windows build failures of the frontend

rfellows:
# Summary

[NIFI-14239](https://issues.apache.org/jira/browse/NIFI-14239)

Regenerated the package-lock.json file: it now includes the missing dependencies for windows-specific builds as described in https://github.com/nrwl/nx/issues/29551.

Also ran prettier:fix and it found a minor thing in the doc component that is included here.


mcgilman (approved):
Thanks for the fix @rfellows!

PR #9697, opened 2025-02-05
[NIFI-14241] - Replace deprecated imports of setup-jest with new approach

rfellows:
# Summary

[NIFI-14241](https://issues.apache.org/jira/browse/NIFI-14241)
Replace deprecated imports of setup-jest with new approach

Migrated to use 
```
setupZoneTestEnv({
    errorOnUnknownElements: true,
    errorOnUnknownProperties: true
});
```

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/feature/users.component.spec.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/feature/users.component.spec.ts
@@ -26,6 +26,7 @@ import { MockComponent } from 'ng-mocks';
 import { Navigation } from '../../../ui/common/navigation/navigation.component';
 import { usersFeatureKey } from '../state';
 import { BannerText } from '../../../ui/common/banner-text/banner-text.component';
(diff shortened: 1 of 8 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/feature/users.component.spec.ts
```diff
@@ -34,7 +35,13 @@ describe('Users', () => {
     beforeEach(() => {
         TestBed.configureTestingModule({
             declarations: [Users, UserListing],
-            imports: [RouterModule, RouterTestingModule, MockComponent(BannerText), MockComponent(Navigation)],
+            imports: [
+                RouterModule,
+                RouterTestingModule,
+                MockComponent(BannerText),
+                MockComponent(Navigation),
+                MockComponent(NgxSkeletonLoaderComponent)
```
mcgilman: I don't think this component uses the skeleton loader so mocking here shouldn't be needed. I realize this PR been open for some time so it's possible the components changed since opening.

mcgilman (approved):
Thanks for the updates @rfellows! Will merge once CI complets.

PR #9701, opened 2025-02-07
[NIFI-14327][NIFI-14249] - Copy/Paste handling when clipboard is not available

rfellows:
# Summary

[NIFI-14237](https://issues.apache.org/jira/browse/NIFI-14237) - Disable copy when system clipboard is not available
[NIFI-14249](https://issues.apache.org/jira/browse/NIFI-14249) - Remove paste from operation panel as only keyboard shortcuts allow for paste.

Also removed the copy effect as it is not used and can not be used due to it making the copy asynchronous, a potential security issue prevented by browsers without explicit acceptance.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
@@ -956,7 +956,16 @@ export class CanvasUtils {
         return { x, y };
     }
 
+    public isClipboardAvailable(): boolean {
+        // system clipboard interaction requires the browser to be in a secured context.
+        return window.isSecureContext && Object.hasOwn(window, 'ClipboardItem');
+    }
+
     public isCopyable(selection: d3.Selection<any, any, any, any>): boolean {
+        if (!this.isClipboardAvailable()) {
+            return false;
+        }
+
         // if nothing is selected return
         if (selection.empty()) {
             return false;
@@ -1001,7 +1010,7 @@ export class CanvasUtils {
     }
 
     public isPastable(): boolean {
-        return this.canvasPermissions.canWrite;
+        return this.isClipboardAvailable() && this.canvasPermissions.canWrite;
     }
 
     /**
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
@@ -20,6 +20,14 @@ import { FlowService } from '../../service/flow.service';
 import { Actions, createEffect, ofType } from '@ngrx/effects';
 import { concatLatestFrom } from '@ngrx/operators';
 import * as FlowActions from './flow.actions';
+import {
+    disableComponent,
+    enableComponent,
+    setRegistryClients,
+    startComponent,
+    startPollingProcessorUntilStopped,
+    stopComponent
+} from './flow.actions';
 import * as StatusHistoryActions from '../../../../state/status-history/status-history.actions';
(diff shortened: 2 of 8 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
```diff
@@ -956,7 +956,16 @@ export class CanvasUtils {
         return { x, y };
     }
 
+    public isClipboardAvailable(): boolean {
+        // system clipboard interaction requires the browser to be in a secured context.
+        return window.isSecureContext && Object.hasOwn(window, 'ClipboardItem');
```
sfc-gh-aokrushko, line 961: (outside of the scope for this PR): usually window is wrapped with an injection token for a few reasons:
- testing
- ssr support (which might be needed in the future)

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
```diff
@@ -956,7 +956,16 @@ export class CanvasUtils {
         return { x, y };
     }
 
+    public isClipboardAvailable(): boolean {
```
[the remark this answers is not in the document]
rfellows: Understood, just added it for consistency with the rest of the file.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
alex-okrushko, line 961: (outside of the scope for this PR): usually window is wrapped with an injection token for a few reasons:

testing
ssr support (which might be needed in the future)

rfellows: Good idea to inject window in the future, but there are many places where it is used directly now. We should address it across the board in its own commit. I filed this [1] JIRA, feel free to add commentary to it.

[1] [NIFI-14250](https://issues.apache.org/jira/browse/)

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/graph-controls/operation-control/operation-control.component.html
```diff
@@ -99,14 +99,6 @@
                             (click)="copy(selection)">
                             <i class="fa fa-copy"></i>
                         </button>
-                        <button
-                            mat-icon-button
-                            class="primary-icon-button mr-2"
-                            type="button"
-                            [disabled]="!canPaste()"
-                            (click)="paste()">
```
mcgilman, line 107: We should be able to also remove `canPaste` and `paste` from the underlying component.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
```diff
@@ -2495,45 +2494,6 @@ export class FlowEffects {
         )
     );
 
-    copy$ = createEffect(() =>
```
mcgilman, line 2498: I think we can also remove the `copy` action since we're not removing the effect for it.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
mcgilman: Do we need similar logic in the `CopyDirective` so we don't show the copy button when copying it's supported?

mcgilman (approved):
Thanks @rfellows!

PR #9706, opened 2025-02-09
NIFI-14244 send original Records to errors output for PutElasticsearchRecord errors

ChrisSamo632:
# Summary

[NIFI-14244](https://issues.apache.org/jira/browse/NIFI-14244) send original Records to errors output for PutElasticsearchRecord errors

Additional changes:
- update the `MockProcessSession` to allow for `recursive read` of a FlowFile (similar to the `StandardProcessSession` allows)
- rationalise `retriable` to `retryable`

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-docs/src/main/asciidoc/user-guide.adoc
+++ b/nifi-docs/src/main/asciidoc/user-guide.adoc
@@ -695,9 +695,9 @@ In order for a Processor to be considered valid and able to run, each Relationsh
 
 ===== Automatically Retry
 Users can also configure whether or not FlowFiles routed to a given Relationship should be retried. If a FlowFile is routed to any Relationship that is configured to be retried,
-the FlowFile will be re-queued and the Processor will attempt to process it again. If the Processor routes the FlowFile to a retriable Relationship again (either the same Relationship
+the FlowFile will be re-queued and the Processor will attempt to process it again. If the Processor routes the FlowFile to a retryable Relationship again (either the same Relationship
 or another that is configured to be retried), it will be re-queued again, up to the number of
-times specified by the user. If the Processor routes the FlowFile to a retriable Relationship after the specified number of retries, the FlowFile will be transferred to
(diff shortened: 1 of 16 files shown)


exceptionfactory (commented):
Thanks for putting together these improvements for Elasticsearch Record processing @ChrisSamo632.

The changes to `nifi-mock` look good.

I'm concerned about the potential memory impact on building a `Map` of error records, as opposed to writing them directly. I haven't traced the details to consider the issues with going in that direction, did you consider an approach that avoids keeping error Records in memory?

ChrisSamo632:
> I'm concerned about the potential memory impact on building a `Map` of error records, as opposed to writing them directly. I haven't traced the details to consider the issues with going in that direction, did you consider an approach that avoids keeping error Records in memory?

Fair quesiton, and it's probably possible, provided careful handling of the additional Reader for the Input FlowFile.

I'd opted to separate such handling into a try-with-resources method in order to ensure we didn't accidentally leave any file handles open. Accepting, however, that we'd have a duplication of the Records in memory - they're already stored in the operation list that's being sent to Elasticsearch and iterated through to check for Errors. This means double the memory use (temporarily for each batch of data being processed), but greater confidence around the FlowFile reader handling.

I can certainly look again at reading each Record from the Input FlowFile individually and either using it immediately or throwing it away as we loop through the responses checking for errors, but being careful to ensure all Readers are closed at the end.

ChrisSamo632:
@exceptionfactory I refactored to hold no more than a single Record from the Input Flowfile when outputting errors, while only actually reading the flowfile if there were errors in the Elasticsearch response for the batch operation - otherwise, the input flowfile is opened yet not re-read.

exceptionfactory (approved):
Thanks for making the adjustments @ChrisSamo632, the updated approach looks good! +1 merging

PR #9707, opened 2025-02-10
NIFI-14245 Enable caching in DatabaseUserGroupProvider and DatabaseAccessPolicyProvider to improve performance

yoshiata:
# Summary

[NIFI-14245](https://issues.apache.org/jira/browse/NIFI-14245)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/database/DatabaseAccessPolicyHolder.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/database/DatabaseAccessPolicyHolder.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.registry.security.authorization.database;
+
+import org.apache.nifi.registry.security.authorization.AccessPolicy;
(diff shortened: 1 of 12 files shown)


yoshiata:
I’d really appreciate it if you could review this.

yoshiata:
I noticed the changes to the unit tests were incorrect, so I’ve fixed them again.
https://github.com/apache/nifi/pull/9707/commits/8c1f5d3becbba0d5a95cf4fad0d2f24986219bd0

This should allow us to test the same behavior as before the patch, meaning we can confirm that enabling the cache doesn’t actually change anything.

exceptionfactory (changes requested):
Thanks for proposing this improvement @yoshiata. I have not reviewed all of the details, but I'm concerned about the potential complexity introduced through method synchronization and calls to refresh the entire cache of information. The Jira issue description mentions improving the speed of listing users, but this by itself does not seem like sufficient reason to introduce the complexity of caching. If there are additional benefits, are there are some performance comparisons, that may help determine the relative value of introducing the complexity.

yoshiata:
@exceptionfactory Thanks for replying! And sorry I didn’t give enough info about the performance.
I left a comment on Jira issue with the performance comparison results.

https://issues.apache.org/jira/browse/NIFI-14245?focusedCommentId=17940552&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17940552

I implemented caching by referencing FileUserGroupProvider class and FileAccessPolicyProvider class as follows:

* On startup, retrieve all records and store them in the cache
* On updates, write to the database first, then retrieve all records again to rebuild the cache

exceptionfactory (commented):
Thanks for the reply and performance comparison details on the Jira issue @yoshiata! Although 1000 users is probably on the higher end, that is quite a performance hit.

The reference to the File-based implementations is helpful. I was considering recommending a different approach based on Caffeine Cache, but noting a similar existing approach is useful.

In that light, the strategy seems reasonable in general. I may be able to take a closer look at these changes if others aren't able to review them before me, but I have removed my current request for changes flag.

yoshiata:
Thanks for removing the request for changes @exceptionfactory! I’ll wait to see if anyone else has feedback.
And let me know if you think of anything else—happy to update!

exceptionfactory (commented):
Thanks for your patience on this @yoshiata. On subsequent review and comparison to the File-based Providers, the implementation looks like a reasonable approach to provide caching. There have been some more recent updates to trigger additional testing on changes for NiFi Registry, so if you can rebase this pull request on the current main branch, that should confirm compatibility and allow this to go forward.

yoshiata:
Thanks for the review, @exceptionfactory.
I’ve rebased onto the latest main branch, so please feel free to move this forward.

yoshiata:
Thanks for merging!

PR #9710, opened 2025-02-11
NIFI-14213 Fixed code for(Inherited Parameter Context Not Retained Wh…

2020555642rahul:
This PR addresses an issue where inherited parameter contexts were not properly attached when upgrading a versioned process group. Previously, upgrading a flow would create the new inherited parameter context but fail to link it to the parameter context, causing missing parameter references. This fix ensures that inherited parameter contexts are correctly retained and updated during flow upgrades.

NIFI Issue link: https://issues.apache.org/jira/browse/NIFI-14213


pvillard31:
Hi @2020555642rahul - why is the PR opened against NiFi 1.28? is the issue existing only in NiFi 1 or also in NiFi 2? If it exists on NiFi 2 as well, we would want the fix in the main branch first and then consider a cherry-pick into the support branch.

PR #9715, opened 2025-02-12
NIFI-14260: ParquetReader can't handle timestamps stored as INT96 type

mark-bathori:
# Summary

[NIFI-14260](https://issues.apache.org/jira/browse/NIFI-14260)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
@@ -16,8 +16,10 @@
  */
 package org.apache.nifi.serialization.record.field;
 
+import org.apache.nifi.serialization.record.util.DataTypeUtils;
 import org.apache.nifi.serialization.record.util.IllegalTypeConversionException;
 
+import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
@@ -29,12 +31,15 @@ import java.time.temporal.TemporalQueries;
 import java.time.temporal.TemporalQuery;
 import java.util.Date;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Convert Object to java.time.LocalDateTime using instanceof evaluation and optional format pattern for DateTimeFormatter
  */
 class ObjectLocalDateTimeFieldConverter implements FieldConverter<Object, LocalDateTime> {
     private static final long YEAR_TEN_THOUSAND = 253_402_300_800_000L;
+    private static final int JULIAN_EPOCH_OFFSET_DAYS = 2_440_588;
+    private static final byte INT96_TIMESTAMP_BYTES_LENGTH = 12;
 
     private static final TemporalQuery<LocalDateTime> LOCAL_DATE_TIME_TEMPORAL_QUERY = new LocalDateTimeQuery();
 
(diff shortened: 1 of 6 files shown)


nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateTimeFieldConverter.java
```diff
@@ -85,13 +90,31 @@ public LocalDateTime convertField(final Object field, final Optional<String> pat
                     return tryParseAsNumber(string, name);
                 }
             }
+            case Object[] objectArray -> {
+                if (objectArray.length != INT96_TIMESTAMP_BYTES_LENGTH) {
```
exceptionfactory, line 94: This is a generalized converter, so introducing Parquet-specific handling does not look like the right approach. The implementation should be localized to the ParquetReader.

nifi-extension-bundles/nifi-parquet-bundle/nifi-parquet-processors/src/main/java/org/apache/nifi/parquet/ParquetReader.java
```diff
@@ -40,16 +47,37 @@
         "The schema will come from the Parquet data itself.")
 public class ParquetReader extends AbstractControllerService implements RecordReaderFactory {
 
+    public static final PropertyDescriptor INT96_TIMESTAMP_FIELDS = new PropertyDescriptor.Builder()
+            .name("INT96 timestamp fields")
```
exceptionfactory, line 51: This property name does not follow standard conventions.

exceptionfactory (changes requested):
Thanks for flagging this issue @mark-bathori, however, there are several implementation concerns that need to be considered.

As background, the [INT96](https://parquet.apache.org/docs/file-format/implementationstatus/#physical-types) type is deprecated, although Parquet notes it is still commonly used. I question the value of introducing this support for a deprecated feature in the current ParquetReader. 

The Jira issue mentions Spark as the source, but is there a particular Spark version that still uses INT96? That would help determine whether this should be supported.

If it is something to consider supporting, it would be better to have an alternative ParquetReader implementation to avoid introducing this deprecated feature. In more depth on the implementation, the standard `ObjectLocalDateTimeFieldConverter` should not include any of this specialized conversion handling. I understand that it is referenced through the `AvroTypeUtil`, so decoupling and introducing an alternate implementation may be involved.

mark-bathori:
Thanks @exceptionfactory for the review. I've checked the Spark documentations and it seems like they are still using INT96 for timestamps in the latest 3.5.4 version: https://spark.apache.org/docs/3.5.4/sql-data-sources-parquet.html
I think the support of reading INT96 type would be useful for users migrating their data into something non-deprecated format but it is just unfortunate that the **parquet-avro** lib doesn't provide any option for it beside the current byte array format. I'm not really sure about introducing another ParquetReader just for this function. 
I'll think about a new approach which completely avoids the AvroTypeUtil and ObjectLocalDateTimeFieldConverter classes but due the complexity of supporting this I'm closing this PR for now.

exceptionfactory:
Thanks for the reply @mark-bathori.

I see some Spark options for controlling where and how to use [INT96 Types](https://spark.apache.org/docs/3.5.4/sql-data-sources-parquet.html#configuration), making it sound like something that can be configured on the Spark side.

Use of the `parquet-avro` library is convenient for NiFi Records, but it does raise the question about whether that level of indirection is worth revisiting. That's a larger question as it relates to reading Parquet, but perhaps worth considering separately.

Thanks for considering the options.

PR #9718, opened 2025-02-12
NIFI-14106 Fixed bug which led to loss of dates and times precision when using SplitExcel.

dan-s1:
# Summary

[NIFI-14106](https://issues.apache.org/jira/browse/NIFI-14106)
As detailed in the ticket , I reverted the change made in #9466 as the `org.apache.poi.ss.usermodel.CellCopyContext` can be used to prevent exceeding the cell styles. In addition this PR involved having to correct a POI bug which I reported in the following [thread](https://lists.apache.org/thread/71pxytfsn6g30b3q5k8p904mxdr50407)  and can be found [here](https://bz.apache.org/bugzilla/show_bug.cgi?id=69583). As a result I had to copy the following methods from the following classes in order to fix the bug and make use of `org.apache.poi.ss.usermodel.CellCopyContext`.

- `copyRows` from `org.apache.poi.xssf.usermodel.XSSFSheet`
-  `copyRowFrom` from `org.apache.poi.xssf.usermodel.XSSFRow` 
-  `copyCell` from `org.apache.poi.ss.util.CellUtil`

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -103,3 +103,12 @@ This includes derived works from Spring Framework available under Apache Softwar
       https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
     and can be found in
       nifi-api/src/main/java/org/apache/nifi/processor/util/URLValidator.java
+
+This includes derived works from Apache POI available under Apache Software License V2
+    Copyright 2003-2025 The Apache Software Foundation
+    The derived work is adapted from
+    https://svn.apache.org/repos/asf/poi/trunk/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFSheet (copyRows method)
+    https://svn.apache.org/repos/asf/poi/trunk/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFRow (copyRowFrom method)
+    https://svn.apache.org/repos/asf/poi/trunk/poi/src/main/java/org/apache/poi/ss/util/CellUtil.java (copyCell method).
+    and can be found in
+    nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java.SheetCopyStrategy
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
@@ -35,16 +35,29 @@ import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.poi.common.Duplicatable;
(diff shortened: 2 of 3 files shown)


exceptionfactory (commented):
@dan-s1 Can you include a link to the issue created for the POI project? This change includes copying a good bit of code, and the particular elements that required modification are not clear. On a cursory read, if we proceed with this approach, it seems like there are some aspects of the methods that are not needed.

dan-s1:
@exceptionfactory Is what I included now in the description enough?

exceptionfactory:
> @exceptionfactory Is what I included now in the description enough? Based on the included thread, I understood that the POI folks would make the necessary changes although I am not sure when that will be hence I included the necessary changes in `SplitExcel`.

Thanks for providing the additional background and link, that is helpful.

I'm a bit concerned about the level of complexity from these additional methods, but I haven't walked through the details closely. I will plan on reviewing soon.

exceptionfactory:
> When copying code from another project, one must include a note to that effect in the NOTICE file the NAR, correct?
> 
> However, I would be -1 on copying POI code into NiFi, when we could instead simply wait for a POI release that has the desired behavior.

I agree with you @mosermw, it looks like there has already been some progress on this issue in the POI project, so it seems better to wait for those changes to be completed and released.

exceptionfactory:
@dan-s1 It looks like the changes in POI in the following commit and several others related would obviate the changes in this pull request:

https://github.com/apache/poi/commit/f61dddea125824f6fb496adb6299e08625646c73

With the work there, it seems best to close this pull request and track the POI version changes when they are ready.

What do you think?

dan-s1:
@exceptionfactory Thanks for showing me that latest commit as that was the change I was waiting for to allow for reverting the changes of #9466. It seems the POI changes would solve all the issues I had so I am fine with closing this PR. Should I put something in a comment on the ticket so we know to watch for this?

exceptionfactory:
@dan-s1, Yes, a comment in the Jira, linking to the POI commit and mentioning that this should be resolved in the next version would be helpful.

dan-s1:
@exceptionfactory Okay I added that. Is there anyway to put the ticket on hold in Jira?

exceptionfactory:
I'm not seeing a blocked status option in Jira, but that's fine, the comment is sufficient for now.

dan-s1:
@dan-s1 I just wanted to add even after the latest POI changes, there is still a need to revert the changes made in #9466 to allow cell styles to be copied which would enable copying of date formats. I think this ticket could cover that. Is that okay or should another ticket be created?

exceptionfactory:
> @exceptionfactory I just wanted to add even after the latest POI changes, there is still a need to revert the changes made in #9466 to allow cell styles to be copied which would enable copying of date formats. I think this ticket could cover that. Is that okay or should another ticket be created?

Sure, this ticket can cover those changes.

dan-s1:
@exceptionfactory I encapsulated the copy rows logic in a private static class as you had suggested and I updated the NOTICE file accordingly.

exceptionfactory (commented):
Thanks for the updated approach @dan-s1. I plan to take a closer look soon.

dan-s1:
This PR is no longer necessary with the lates POI version 5.4.1 which will be incorporated in a PR for https://issues.apache.org/jira/browse/NIFI-14439

PR #9720, opened 2025-02-13
NIFI-14266 - Allow ExecuteSQL to not overwrite flowfile content when no result

pvillard31:
# Summary

[NIFI-14266](https://issues.apache.org/jira/browse/NIFI-14266) - Allow ExecuteSQL to not overwrite flowfile content when no result

The ExecuteSQL processor may be used in a flow definition to execute SQL statements such as TRUNCATE TABLE, DROP TABLE, etc. While there is the option to use pre/post queries, it might not always be an option to have everything executed within the same processor. In such a scenario, it might be needed to not change the content of the FlowFile.

This improvement adds a property allowing a user to specify that the FlowFile's content should remain untouched when the execution of the query does not generate any result and, of course, when the processor was triggered by an incoming FlowFile.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AbstractExecuteSQL.java
```diff
@@ -188,6 +189,17 @@ public abstract class AbstractExecuteSQL extends AbstractProcessor {
             .required(true)
             .build();
 
+    public static final PropertyDescriptor OVERWRITE_FLOW_FILE_CONTENT = new PropertyDescriptor.Builder()
+            .name("Overwrite FlowFile Content")
```
exceptionfactory, line 193: Rather than introducing a Boolean property, it seems like it would be better to call this something like `Content Output Strategy`, with values such as `Results` and `Ignored`. That would provide more potential for future flexibility.

exceptionfactory (changes requested):
Thanks for introducing this new property @pvillard31. The general concept looks good, but I recommend creating a strategy property, instead of Boolean property for the implementation.

exceptionfactory:
Completed and merged in PR #9748.

PR #9734, opened 2025-02-20
NIFI-14282: Ensuring that the pasted property descriptor is sensitive…

mcgilman:
… before applying a sensitive property from the copied instance.


exceptionfactory (approved):
Thanks for improving the behavior and adding a test case for this scenario @mcgilman, looks good! +1 merging

PR #9737, opened 2025-02-21
NIFI-14283 ConsumeKinesisStream non-static workerLock

dariuszseweryn:
# Summary

[NIFI-14283](https://issues.apache.org/jira/browse/NIFI-14283)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

- [x] (there are none) New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [x] (there are none) New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [x] (no changes) Documentation formatting appears as expected in rendered files

### DIFF
--- a/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
+++ b/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
@@ -345,7 +345,7 @@ public class ConsumeKinesisStream extends AbstractAwsAsyncProcessor<KinesisAsync
             "leaseManagementConfig.failoverTimeMillis", FAILOVER_TIMEOUT
     );
 
-    private static final Object WORKER_LOCK = new Object();
+    private final Object workerLock = new Object();
     private static final String SCHEDULER_THREAD_NAME_TEMPLATE = ConsumeKinesisStream.class.getSimpleName() + "-" + Scheduler.class.getSimpleName() + "-";
 
     private static final Set<Relationship> RELATIONSHIPS = Set.of(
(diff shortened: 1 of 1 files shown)


exceptionfactory (commented):
Thanks for looking at improving this behavior @dariuszseweryn.

Although moving from a static object to an instance variable is a step forward, it seems like this is worth some additional evaluation. For example, should the Scheduler thread be created in the onScheduled method, but only started in onTrigger? It may also be more efficient to use an atomic variable, as opposed to the synchronized lock object approach.

dariuszseweryn:
> For example, should the Scheduler thread be created in the onScheduled method, but only started in onTrigger?

This is possible but has a downside. Scheduler needs a record processor which uses a `ProcessSessionFactory` that is available only in `onTrigger()`. The processor won't get called before the thread the Scheduler runs on is started, yet it would introduce additional state in the record processor which is definitely a downside.

The question is: is the advantage bigger than the disadvantage? 
imho not given the creation is fairly infrequent

> It may also be more efficient to use an atomic variable, as opposed to the synchronized lock object approach.

I may lack experience with NiFi but it seems to me that start and stop events are fairly infrequent for processors — this indicates there will be little contention especially when the lock is inherent to a processor instance, no?

If so, having an AtomicBoolean or other Atomic* will only unnecessary complicate the code.

dariuszseweryn:
> That actually highlights a larger concern in that having the Processor hold on to the Session Factory in this way is generally poor practice and should be revisited.

Given the Scheduler produces records at arbitrary moments we can:
1. On each `onTrigger()` substitute the factory in the processor regularly
2. Buffer the records and submit them only `onTrigger()` — this would need to have manual checkpointing, not sure yet how cumbersome will it be

> Starting and stopping is less frequent, but the framework calls onTrigger() very frequently, thus inefficient nature of the synchronized block in a method that the framework calls repeatedly.

The synchronization is only used if there is no `Scheduler` created yet, it uses (searches notes) _double-checked locking_ to avoid the synchronization overhead

dariuszseweryn:
@exceptionfactory I have broken out the Session Factory issue into a separate [JIRA: NIFI-14292](https://issues.apache.org/jira/browse/NIFI-14292)

exceptionfactory (approved):
Thanks for discussing the options @dariuszseweryn. Tracking the more substantive changes in a separate Jira issue sounds good. As this pull request is a useful incremental adjustment, I will proceed to merge and we can consider the larger questions under the new Jira issue you created.

PR #9752, opened 2025-02-28
[NIFI-14299] - FIX: Apply button has no text on the Controller Settings -> General tab

rfellows:
# Summary

[NIFI-14299](https://issues.apache.org/jira/browse/NIFI-14299)

* Added missing import for the `NifiSpinnerDirective`. The button text is embedded within the element specifying this directive. Since the directive wasn't imported, it failed to render the text as a consequence.


mcgilman (approved):

PR #9753, opened 2025-02-28
NIFI-14303 Additional metadata attributes in ListGoogleDrive

turcsanyip:
Introduced optionally configurable metadata attributes in ListGoogleDrive Added drive.created.time and drive.modified.time in List/Fetch/PutGoogleDrive Fixed drive.timestamp in Fetch/PutGoogleDrive

# Summary

[NIFI-14303](https://issues.apache.org/jira/browse/NIFI-14303)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/drive/ListGoogleDrive.java
+++ b/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/drive/ListGoogleDrive.java
@@ -17,11 +17,11 @@
 package org.apache.nifi.processors.gcp.drive;
 
 import com.google.api.client.http.HttpTransport;
-import com.google.api.client.util.DateTime;
 import com.google.api.services.drive.Drive;
 import com.google.api.services.drive.DriveScopes;
 import com.google.api.services.drive.model.File;
 import com.google.api.services.drive.model.FileList;
+import com.google.api.services.drive.model.User;
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
 import org.apache.nifi.annotation.behavior.PrimaryNodeOnly;
@@ -48,7 +48,10 @@ import org.apache.nifi.processors.gcp.ProxyAwareTransportFactory;
 import org.apache.nifi.processors.gcp.util.GoogleUtils;
 import org.apache.nifi.proxy.ProxyConfiguration;
 import org.apache.nifi.scheduling.SchedulingStrategy;
+import org.apache.nifi.serialization.SimpleRecordSchema;
+import org.apache.nifi.serialization.record.RecordField;
 import org.apache.nifi.serialization.record.RecordSchema;
+import org.apache.nifi.util.StringUtils;
 
 import java.io.IOException;
 import java.time.Instant;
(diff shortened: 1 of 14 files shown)


nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/drive/ListGoogleDrive.java
```diff
@@ -150,6 +177,29 @@ public class ListGoogleDrive extends AbstractListProcessor<GoogleDriveFileInfo>
             .dependsOn(LISTING_STRATEGY, BY_ENTITIES)
             .build();
 
+    public static final PropertyDescriptor ENRICH_FILE_METADATA = new PropertyDescriptor.Builder()
```
exceptionfactory, line 180: Does this need to be a configurable property? It seems more straightforward to always include these available metadata properties, as opposed to optionally including them based on the property.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/drive/ListGoogleDrive.java
```diff
@@ -197,7 +266,37 @@ public void onScheduled(final ProcessContext context) throws IOException {
...
         HttpTransport httpTransport = new ProxyAwareTransportFactory(proxyConfiguration).create();
 
-        driveService = createDriveService(context, httpTransport, DriveScopes.DRIVE_METADATA_READONLY);
+        driveService = createDriveService(context, httpTransport, DriveScopes.DRIVE, DriveScopes.DRIVE_METADATA_READONLY);
+    }
+
+    @OnScheduled
+    @Override
+    // initListedEntityTracker() needs to be overridden because the super implementation calls getRecordSchema()
+    // so recordSchema must be initialized before that
```
exceptionfactory, line 275: Method-level comments should follow Javadoc conventions, and be set before method annotations.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/drive/ListGoogleDrive.java
```diff
@@ -197,7 +266,37 @@ public void onScheduled(final ProcessContext context) throws IOException {
...
+        Stream.of(
+                GoogleDriveFlowFileAttribute.ID,
+                GoogleDriveFlowFileAttribute.FILENAME,
+                GoogleDriveFlowFileAttribute.SIZE,
+                GoogleDriveFlowFileAttribute.SIZE_AVAILABLE,
+                GoogleDriveFlowFileAttribute.TIMESTAMP,
+                GoogleDriveFlowFileAttribute.CREATED_TIME,
+                GoogleDriveFlowFileAttribute.MODIFIED_TIME,
+                GoogleDriveFlowFileAttribute.MIME_TYPE
+        ).forEach(a -> recordFields.add(a.getRecordField()));
```
exceptionfactory, line 290: Creating a new `Stream` seems unnecessary, is it possible to simply iterative over these values, or if required, creating a static `List.of()`?

exceptionfactory (changes requested):
Thanks for working on these improvements to provide additional metadata from Google Drive @turcsanyip.

Including the metadata seems helpful, but I'm not sure about making a configurable property for which attributes should be included. It seems simpler to always include the metadata. Is there a particular reason for providing the fine-grained level of control over the output?

I'm also wondering about the Folder Separator property, as it seems potentially confusing based on the description. It sounds like it requires some background knowledge of what will be listed, which seems like a dangerous assumption for continuous flows. Instead of making this configurable, it seems that a better approach would be some type of encoding strategy. Using URL encoding seems like the most obvious choice, which would support a standard `/` separator, and encode that character when it is part of file or folder names. There may be other options, but that is one that comes to mind.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/drive/ListGoogleDrive.java
turcsanyip: @exceptionfactory Thanks for sharing your suggestions!

The original idea was to avoid impacting the current behavior and performance.

The extra metadata needs to be retrieved from the Google Drive service (same query but additional fields) which may affect response time. Then the metadata fields are added in the processors output as FlowFile attributes or record fields in content which increases its size.

I did some measurements and it seems the performance loss is not significant: 7-8% overhead in case of 1000 files in a folder, basically the same performance (or unmeasurable difference) in case of 10 or 100 files.

Regarding the FF size: it is roughly doubled with the new fields but the size is still small (261 vs 561 KB for 1000 files in record mode).

All in all, I'm inclined to remove the configurable property and add the new metadata fields in a static way. That makes the code much simpler too.

exceptionfactory: Thanks for the additional background @turcsanyip, that's helpful.

It sounds like the differences, although worth noting, are not that significant, so setting the new metadata fields statically sounds good.

nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/drive/ListGoogleDrive.java
turcsanyip: Not an issue in the reimplemented version.

PR #9755, opened 2025-02-28
NIFI-14297 Added support for Avro schemas with String default for type bytes and 'decimal' logicalType.

dan-s1:
# Summary

[NIFI-14297 ](https://issues.apache.org/jira/browse/NIFI-14297 )
This fix in this PR addresses the need to support Avro schema byes type and decimal logical type when a default is specified. As part of fix the unit test `testIsBigDecimalTypeCompatible` in `TestDataTypeUtils` was converted to a JUnit 5 Paramaeterized test and the particular use case of a decimal as bytes was added.
In addition a unit test was added to `TestConvertRecord` to exercise the use case of the user who submitted the ticket although this may not be necessary.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
@@ -1318,6 +1318,7 @@ public class DataTypeUtils {
             case null -> false;
             case Number ignored -> true;
             case String s -> stringPredicate.test(s);
+            case byte[] bytes -> stringPredicate.test(new String(bytes, StandardCharsets.UTF_8));
             default -> false;
         };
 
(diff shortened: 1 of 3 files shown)


exceptionfactory (commented):
Thanks for working on this issue @dan-s1. The implementation looks good, but I noticed that handling the default string value did not work as expected after adjusting the input content for the test. I pushed an update that expects content that includes the default value and plan to revisit soon.

exceptionfactory (approved):
Thanks again @dan-s1, +1 merging

dan-s1:
@exceptionfactory If you do not mind can you please explain why your changes were necessary? I am trying to wrap my head around this but I am not understanding. Thanks!

exceptionfactory:
> @exceptionfactory If you do not mind can you please explain why your changes were necessary? I am trying to wrap my head around this but I am not understanding. Thanks!

Thanks for asking! The additional case handling for a `byte[]` in `DataTypeUtils` was necessary to handle support for a default string value using the byte array type for the field. When reading the record, the Avro Schema returns the default value as a byte array. In order to treat it as the `BigDecimal` logical type, the conversion from a byte array to a BigDecimal is required.

If you remove the case statement in `DataTypeUtils`, you can see how the unit test fails when attempting to handle the default value.

PR #9765, opened 2025-03-04
NIFI-14315: Ensuring the bulletin level is included in the request pa…

mcgilman:
…yload when editing a Controller Service.


sfc-gh-jaslan:
Reviewing...

PR #9771, opened 2025-03-05
NIFI-14323 Add Support for Astral uv to Install Python Dependencies

exceptionfactory:
# Summary

[NIFI-14323](https://issues.apache.org/jira/browse/NIFI-14323) Adds optional support for using [Astral uv](https://docs.astral.sh/uv/) to install dependencies for Python Processors.

Astral uv is a modern, Apache 2-licensed, Python package manager with a [pip interface](https://docs.astral.sh/uv/pip/) that offers much better performance for loading Python dependencies.

The changes to the Python `ExtensionManager` use the `uv` command discovery to determine whether `uv` is installed. `ExtensionManager` prefers Astral uv when available and falls back to current Python pip behavior when uv is not found. This approach provides compatibility with existing installations, and supports opting in to Astral uv by installing it and making it available to the operating system user that runs NiFi.

Updates to the NiFi Docker image include [installing  uv](https://docs.astral.sh/uv/getting-started/installation/) using the standard method as part of the build, and removing the `python3-pip` package from the build.

Updates to the `system-tests` workflow add the [setup-uv](https://github.com/astral-sh/setup-uv) action to make it available for Python-based tests. Current tests include installing Python Processors with dependencies, so the system test workflow verifies working behavior with Astral uv.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -102,6 +102,11 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: '3.10'
+      - name: Set up Astral uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: '3.10'
+          enable-cache: false
 
       - name: Build Dependencies
         env:
--- a/nifi-docker/dockerhub/Dockerfile
+++ b/nifi-docker/dockerhub/Dockerfile
@@ -52,13 +52,14 @@ RUN groupadd -g ${GID} nifi || groupmod -n nifi `getent group ${GID} | cut -d: -
     && apt-get update \
     && apt-get install -y jq xmlstarlet procps unzip \
     && apt-get install -y python3 \
-    && apt-get install -y python3-pip \
     && apt-get install -y python3-venv \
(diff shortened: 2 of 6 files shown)


Lehel44:
Thanks @exceptionfactory for adding this. What do you think of making the uv cache configurable?

exceptionfactory:
> Thanks @exceptionfactory for adding this. What do you think of making the uv cache configurable?

Thanks for the feedback @Lehel44. Supporting optional caching seems worth considering, but there are tradeoffs as it can effectively increase the amount of disk space required for dependencies. On the other hand, with large numbers of Python Processors that share the same dependencies, it could be useful in some scenarios.

I think it would be better to consider the caching option in a separate PR, and make it applicable to both pip and uv. Something like a cache directory property, defaulting to unset, could be considered.

Lehel44 (approved):
Thanks @exceptionfactory for adding this. LGTM+1 merging.

PR #9774, opened 2025-03-06
NIFI-14332: Creating processor to list collaborators of Box Files

ncover21:
# Summary

[NIFI-14332](https://issues.apache.org/jira/browse/NIFI-14332)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/GetBoxFileCollaborators.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/GetBoxFileCollaborators.java
@@ -23,7 +23,6 @@ import com.box.sdk.BoxCollaboration;
 import com.box.sdk.BoxCollaborator;
 import com.box.sdk.BoxFile;
 import org.apache.nifi.annotation.behavior.InputRequirement;
-import org.apache.nifi.annotation.behavior.ReadsAttribute;
 import org.apache.nifi.annotation.behavior.WritesAttribute;
 import org.apache.nifi.annotation.behavior.WritesAttributes;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
@@ -40,13 +39,13 @@ import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
-import org.jetbrains.annotations.VisibleForTesting;
 
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static org.apache.nifi.processors.box.BoxFileAttributes.ERROR_CODE;
 import static org.apache.nifi.processors.box.BoxFileAttributes.ERROR_CODE_DESC;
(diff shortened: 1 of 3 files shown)


nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/GetBoxFileCollaborators.java
```diff
@@ -0,0 +1,222 @@
...
+import static org.apache.nifi.processors.box.BoxFileAttributes.ERROR_MESSAGE;
+import static org.apache.nifi.processors.box.BoxFileAttributes.ERROR_MESSAGE_DESC;
+import static org.apache.nifi.processors.box.BoxFileAttributes.ID;
+import static org.apache.nifi.processors.box.BoxFileAttributes.ID_DESC;
+
+@InputRequirement(InputRequirement.Requirement.INPUT_REQUIRED)
+@Tags({"box", "storage", "collaboration", "permissions", "sharing"})
+@CapabilityDescription("Retrieves all collaborators on a Box file and adds the collaboration information to the FlowFile's attributes.")
+@SeeAlso({FetchBoxFile.class, ListBoxFile.class})
+@ReadsAttribute(attribute = ID, description = ID_DESC)
```
pvillard31, line 62: Similar to comment on #9769 - should be removed as reading a flow file attribute may not be true (although, extremely likely as it would surely be used in combination with other Box processors)

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/GetBoxFileCollaborators.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    /**
+     * Creates a BoxFile instance for a given file ID.
+     *
+     * @param fileId the ID of the file
+     * @return BoxFile instance
+     */
+    @VisibleForTesting
```
pvillard31, line 166: Similar to comment on #9769 - recommend removing this annotation to avoid the implied dependency

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/GetBoxFileCollaborators.java
```diff
@@ -0,0 +1,222 @@
...
+
+        final Map<String, String> attributes = new HashMap<>();
+        attributes.put(ID, fileId);
+        attributes.put("box.collaborations.count", String.valueOf(count));
+
+        collabMap.forEach((key, value) -> {
+            if (!value.isEmpty()) {
+                addAttributeIfNotEmpty(attributes, "box.collaborations." + key + ".ids", value);
+            }
+        });
```
pvillard31, line 210: No need to check value twice.

```suggestion
        collabMap.forEach((key, value) -> {
            addAttributeIfNotEmpty(attributes, "box.collaborations." + key + ".ids", value);
        });
```

pvillard31 (changes requested):
Is there a concern around the size of the attributes we're creating here? I don't have a better solution/approach to suggest but given the attribute size limit, I'm wondering if we could be in case where we have too many users/groups and it won't fit.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/GetBoxFileCollaborators.java
ncover21: Removed.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/GetBoxFileCollaborators.java
ncover21: Removed

pvillard31 (approved):

PR #9776, opened 2025-03-06
NIFI-13738 Add Max Inbound Message Body Size Property for AMQP

hr-ireh:
setMaxInboundMessageBodySize property would be configurable as a NiFi UI property.


nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
```diff
@@ -138,7 +138,14 @@ abstract class AbstractAMQPProcessor<T extends AMQPWorker> extends AbstractProce
...
             .build();
-
+    public static final PropertyDescriptor MAX_INBOUND_MESSAGE_BODY_SIZE = new PropertyDescriptor.Builder()
+            .name("MaxInboundMessageBodySize")
+            .description("Maximum body size of inbound (received) messages in bytes.")
+            .required(false)
+            .defaultValue("67108864")
+            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .build();
```
pvillard31, line 148: Recommend using built-in data size validator:

```suggestion
    public static final PropertyDescriptor MAX_INBOUND_MESSAGE_BODY_SIZE = new PropertyDescriptor.Builder()
            .name("Max Inbound Message Body Size")
            .description("Maximum body size of inbound (received) messages.")
            .required(true)
            .defaultValue("64 MB")
            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
            .addValidator(StandardValidators.DATA_SIZE_VALIDATOR)
            .build();
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
```diff
@@ -303,7 +311,8 @@ protected Connection createConnection(ProcessContext context, ExecutorService ex
         final ConnectionFactory cf = new ConnectionFactory();
         cf.setUsername(context.getProperty(USER).evaluateAttributeExpressions().getValue());
         cf.setPassword(context.getProperty(PASSWORD).getValue());
-
+        cf.setMaxInboundMessageBodySize(Integer.parseInt(context.getProperty(MAX_INBOUND_MESSAGE_BODY_SIZE).evaluateAttributeExpressions().getValue()));
+
```
pvillard31, line 315:
```suggestion
        cf.setMaxInboundMessageBodySize(context.getProperty(MAX_INBOUND_MESSAGE_BODY_SIZE).evaluateAttributeExpressions().asDataSize(DataUnit.B).intValue());

```

pvillard31 (changes requested):

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
hr-ireh: MaxInboundMessageBodySize is an integer and we need to validate that the input is not out of range.

My suggestion is to use this:
StandardValidators.POSITIVE_INTEGER_VALIDATOR

pvillard31: Then you can do something like:

````java
.addValidator(StandardValidators.createDataSizeBoundsValidator(1, Integer.MAX_VALUE))
````

Using the data size approach makes it more user friendly as the user can just provide something like '1 MB', etc.

File: nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java [hunk 3]
```diff
@@ -303,7 +311,8 @@ protected Connection createConnection(ProcessContext context, ExecutorService ex
         final ConnectionFactory cf = new ConnectionFactory();
         cf.setUsername(context.getProperty(USER).evaluateAttributeExpressions().getValue());
         cf.setPassword(context.getProperty(PASSWORD).getValue());
-
+        cf.setMaxInboundMessageBodySize(context.getProperty(MAX_INBOUND_MESSAGE_BODY_SIZE).evaluateAttributeExpressions().asDataSize(DataUnit.B).intValue());
+
```
pvillard31, line 315: Please make sure the build is valid with `mvn clean install -Pcontrib-check`

````
Warning:  src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java:[315] (regexp) RegexpSinglelineJava: Line has trailing whitespace.
````

```suggestion
        cf.setMaxInboundMessageBodySize(context.getProperty(MAX_INBOUND_MESSAGE_BODY_SIZE).evaluateAttributeExpressions().asDataSize(DataUnit.B).intValue());

```

pvillard31 (changes requested):

pvillard31 (changes requested):
In addition to the checkstyle violation mentioned in my previous comment, another comment: right now the new property is added to the common properties in the Abstract class, meaning that the property will also show in the PublishAMQP processor which does not really make sense. I think that it should not be added in PROPERTY_DESCRIPTORS of the Abstract class and instead added in PROPERTY_DESCRIPTORS of the Consume class. Thoughts?

hr-ireh:
It is true that the settings are related to the consumer, but like the port and IP, they must be set in the ConnectionFactory, which is created in the AbstractAMQPProcessor.
How to set MaxInboundMessageBodySize in ConsumeAMQP!

pvillard31:
The checkstyle validation needs to be fixed.

Regarding the property, what I'm saying is that you can remove it from the PROPERTY_DESCRIPTORS list of the Abstract class and be added into the PROPERTY_DESCRIPTORS list of the Consume processor. Everything else remains unchanged. This only affects where the property will be visible in the UI.

Liammarwood:
This seemed to have broken the PublishAMQP processor now as its expecting the Max property but there is no Validators for it

PR #9778, opened 2025-03-06
NIFI-14324: Code changes regarding extract text processor

samar121:
The processor now accurately extracts and stores the matched text only once, resolving the previous issue of multiple extractions i.e. more than one. This fix ensures that duplicate entries no longer occur when extracting text using the specified regex pattern.


pvillard31:
While I agree with the changes and I totally understand the value here, I believe this change could be considered as a breaking change in case someone is referencing the attribute downstream in the flow (with a RouteOnAttribute for example) using the attribute suffixed with the index (that no longer exists with this change) instead of the attribute without the index suffix.

I'm not against this PR but, if approved, this will need to be clearly mentioned in the release notes.

exceptionfactory:
Thanks for the contribution @samar121.

I agree with @pvillard31 that this would constitute a breaking change as it stands. This is highlighted by the fact that multiple unit tests are failing.

I believe this change could be introduced through a new property to control the behavior, although that needs some additional consideration.

Given that there are multiple unit tests failing, and this is a breaking change as it stands, I am closing this pull request for now. However, please feel free to submit a new pull request after updating the unit tests and considering a way forward in terms of introducing the changes in a non-breaking way.

We can also discuss options on the associated Jira issue if you would like to consider options before moving directly to code changes.

PR #9782, opened 2025-03-06
NIFI:14326: Reverting 'REMOTE_POLL_BATCH_SIZE property for ListSFTP'

gresockj:
# Summary

This reverts #8390, allowing ListSFTP to return more than just the configured batch size.

[NIFI-14326](https://issues.apache.org/jira/browse/NIFI-14326)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/ListFileTransfer.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/ListFileTransfer.java
@@ -31,6 +31,7 @@ import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


EndzeitBegins:
Thank you for looking into this @gresockj.

Reverting the change #8390 certainly.
However, is removing support for the property `REMOTE_POLL_BATCH_SIZE` from `ListSFTP` not essentially a breaking change, now that it present in previous versions?

I'm aware of the discussions around a new release of NiFi on the mailing list soon. 
Wouldn't an easy, non-breaking fix (in terms of processor API) be to set the `maxResults` in [SFTPTransfer.java#L284](https://github.com/apache/nifi/blob/fb17c10b2fc2ebf78f96f7e9b105f2143c3b5235/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/util/SFTPTransfer.java#L284) to a fixed value of `Integer.MAX_VALUE`.
We could then look into a proper fix once the release is done.

exceptionfactory (changes requested):
Concurring with @EndzeitBegins, removing the property should also include an implementation of the `migrateProperties()` method, so that the framework will remove the property from the flow configuration when upgrading.

gresockj:
I see your point.  This sounds like two different approaches: either keep the property and essentially ignore it and use `Integer.MAX_VALUE`, or remove the property and include this in `migrateProperties()`.  I'm thinking the latter is the approach we should take for now.

Note that `SFTPTransfer#getListing` is used by more than just ListSFTP, so making it a fixed value of `Integer.MAX_VALUE` would change more than just this processor's behavior.

pvillard31 (approved):

PR #9791, opened 2025-03-10
[NIFI-14349] - Bump version of NX to the latest (20.5.0)

rfellows:
[NIFI-14349](https://issues.apache.org/jira/browse/NIFI-14349)

This also resolves an issue that dependabot found with the version of axios that was included: https://github.com/apache/nifi/security/dependabot/412


mcgilman (approved):
Thanks @rfellows!

PR #9793, opened 2025-03-10
NIFI-14343 Replaced IdentifyMimeType property descriptors Config File and Config Body with one property descriptor using identifiesExternalResource.

dan-s1:
# Summary

[NIFI-14343](https://issues.apache.org/jira/browse/NIFI-14343)
This PR removes the two property descriptors Config File and Config Body and replaces it with a single property descriptor which uses identifiesExternalResource which thereby allows a single property descriptor to support a configuration body, configuration file path or a configuration specified by a URL.  In addition validation was added to ensure if a custom configuration is replacing the defaults or being merged with the defaults that it is indeed a valid configuration. This PR  no longer supports expression language for this new property descriptor since if if a file path is specified in an attribute, it will be treated as a configuration body and not a file path.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
@@ -111,7 +111,7 @@ public class IdentifyMimeType extends AbstractProcessor {
             .build();
 
     public static final PropertyDescriptor CUSTOM_MIME_CONFIGURATION = new PropertyDescriptor.Builder()
-            .name("Custom Configuration")
+            .name("Custom MIME Configuration")
             .description("A URL or file path to a custom Tika Mime type configuration or the actual content of a custom Tika Mime type configuration.")
             .required(true)
             .identifiesExternalResource(ResourceCardinality.SINGLE, ResourceType.FILE, ResourceType.URL, ResourceType.TEXT)
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
@@ -16,6 +16,7 @@
  */
 package org.apache.nifi.processors.standard;
 
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.nifi.annotation.behavior.InputRequirement;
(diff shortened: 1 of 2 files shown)


mosermw:
Can you update the AllowableValues for REPLACE and MERGE to include the words "custom MIME configurations" to make it more clear?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/IdentifyMimeType.java
```diff
@@ -173,31 +174,23 @@ public void setup(final ProcessContext context) throws IOException {
             this.detector = config.getDetector();
             this.mimeTypes = config.getMimeRepository();
         } else {
-            setCustomMimeTypes(configStrategy, context);
-        }
-
-        this.encodingDetector = config.getEncodingDetector();
-    }
-
-    private void setCustomMimeTypes(String configStrategy, ProcessContext context) {
```
mosermw, line 182: Since you have two places where the MimeTypesFactory.create() logic is duplicated (in customValidate() and setup() ), it seems like a perfect opportunity to keep that logic in a method.  Suggest something like:

```
    private MimeTypes setCustomMimeTypes(String configStrategy, PropertyContext context) throws MimeTypeException, IOException {
        try (final InputStream customInputStream = context.getProperty(CUSTOM_MIME_CONFIGURATION).asResource().read()) {
            if (configStrategy.equals(REPLACE.getValue())) {
                return MimeTypesFactory.create(customInputStream);
            } else {
                try (final InputStream nifiInputStream = getClass().getClassLoader().getResourceAsStream(CUSTOM_MIME_TYPES_FILENAME);
                     final InputStream tikaInputStream = MimeTypes.class.getClassLoader().getResourceAsStream(DEFAULT_MIME_TYPES_PATH)) {
                    return MimeTypesFactory.create(customInputStream, nifiInputStream, tikaInputStream);
                }
            }
        }
    }
```

dan-s1: @mosermw I learnt something new as I did not realize the try with resources block ensures the resource is closed even when there is a return statement inside the try block. See more [here](https://www.google.com/search?q=if+return+inside+java+try+with+resources+will+the+resource+be+closed&oq=if+return+inside+java+try+with+resources+will+the+resource+be+closed+&gs_lcrp=EgZjaHJvbWUyBggAEEUYOdIBCjQzMjUzajBqMTWoAgiwAgE&sourceid=chrome&ie=UTF-8)

mosermw:
Thanks for the changes @dan-s1, this looks good now.  I tested a bunch of scenarios and the migration worked in each.
+1 will merge.

PR #9812, opened 2025-03-19
NIFI-14358 ConsumeKinesisStream added pausing on backpressure

dariuszseweryn:
# Summary

[NIFI-14358](https://issues.apache.org/jira/browse/NIFI-14358)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

- [x] (no new) New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [x] (no new) New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [x] (no changes) Documentation formatting appears as expected in rendered files


exceptionfactory (changes requested):
Thanks for proposing these changes @dariuszseweryn. Honoring back pressure is an important improvement, but I think the interface naming and implementation needs to be reworked.

The `Pause` naming is very generic, and as this relates more specifically the consuming records, a more specific naming strategy would be more intuitive.

On the read side in the RecordProcessor classes, treating this as more of a conditional rather than `consumePause()` would be much clearer. There are some analogies to a [Semaphore](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Semaphore.html), although I don't think that actual class should be used here because of unnecessary complexity. The [ManagedBlocker](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.ManagedBlocker.html) seems a bit closer to what is intended here. What do you think about naming the basic interface something like `RecordProcessorBlocker`? That highlights the usage in `RecordProcessor` and also connotes blocking, which is the purpose of the abstraction. Perhaps `RecordProcessorLatch` could be better, inspired by the `CountDownLatch` class, but not necessarily tied to that implementation. The basic method could be `await()` following the `CountDownLatch` pattern.

The write side in the ConsumeKinesisStream Processor class would need a different set of methods, which could be declared on a concrete class if only referenced in that Processor. On that side, I don't think it is necessary pass in the `ProcessContext` in an `onTrigger()` method. Instead, the Processor itself could check the available relationship status, and then call a method. As far as naming, perhaps something using `block`  could be useful, such as `block()` and `unblock()`.

With that background, this highlights some of the shortcomings of the current Processor design, as described in other Jira issues, but it seems like a reasonable way forward with the right implementation.

dariuszseweryn:
Hey @exceptionfactory 
Thanks for the suggestions! Much appreciated as "naming things is hard™". I'll incorporate your suggestions.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/pause/TestSwitchableRecordProcessorBlocker.java
```diff
@@ -0,0 +1,111 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.function.Supplier;
+
+public class TestSwitchableRecordProcessorBlocker {
+
+    @Test
+    public void testResumeAfterPause() {
```
exceptionfactory, line 27: It looks like these test methods should be renamed to align with the new method names.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/pause/TestSwitchableRecordProcessorBlocker.java
```diff
@@ -0,0 +1,111 @@
...
+        final TestSwitchableRecordProcessorBlockerInspector blockerInspector = new TestSwitchableRecordProcessorBlockerInspector();
+
+        final SwitchableRecordProcessorBlocker recordProcessorBlocker = SwitchableRecordProcessorBlocker.create(blockerInspector);
+
+        recordProcessorBlocker.block();
+        final Thread thread = new Thread(createPauseRunnable(recordProcessorBlocker));
+        thread.start();
+
+        blockerInspector.awaitAwaited();
+        Assertions.assertTrue(thread.isAlive());
```
exceptionfactory, line 37: The general convention is to use `import static` for assert methods.
```suggestion
        assertTrue(thread.isAlive());
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/SwitchableRecordProcessorBlockerInspector.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.aws.kinesis.stream.pause;
+
+interface SwitchableRecordProcessorBlockerInspector {
```
exceptionfactory, line 19: Is this interface only necessary for testing? It seems like it could be removed. If not, then also needs some naming attention.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/SwitchableRecordProcessorBlocker.java
```diff
@@ -0,0 +1,67 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.aws.kinesis.stream.pause;
+
+import java.util.concurrent.CountDownLatch;
+
+public class SwitchableRecordProcessorBlocker implements RecordProcessorBlocker {
```
exceptionfactory, line 21: As this is the regular implementation, recommend using `Standard` for the name:
```suggestion
public class StandardRecordProcessorBlocker implements RecordProcessorBlocker {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/SwitchableRecordProcessorBlocker.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private final SwitchableRecordProcessorBlockerInspector switchableRecordProcessorBlockerInspector;
+    private CountDownLatch isPaused = new CountDownLatch(0);
```
exceptionfactory, line 24:
```suggestion
    private CountDownLatch blocker = new CountDownLatch(0);
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/SwitchableRecordProcessorBlocker.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public static SwitchableRecordProcessorBlocker createNonBlocking() {
```
exceptionfactory, line 26: This appears to be used only in tests. If so, recommend extracting any non-runtime code from this class.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/SwitchableRecordProcessorBlocker.java
dariuszseweryn: I would argue that standard does not provide much information on what behavior the class has

dariuszseweryn: Anyway, changed

dariuszseweryn:
Yet again, great catches. I probably shouldn't code on evenings!

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/SwitchableRecordProcessorBlocker.java
exceptionfactory: I agree that `Standard` doesn't say anything about what it does, but I think it is sufficient in this case as `Switchable` doesn't make it clear what can be switched. Thanks for making the adjustment.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/record/TestAbstractKinesisRecordProcessor.java
```diff
@@ -72,7 +73,7 @@ public void setUp() {
 
         // default test fixture will try operations twice with very little wait in between
         fixture = new AbstractKinesisRecordProcessor(processSessionFactory, runner.getLogger(), "kinesis-test",
-                "endpoint-prefix", null, 10_000L, 1L, 2, DATE_TIME_FORMATTER) {
+                "endpoint-prefix", null, 10_000L, 1L, 2, DATE_TIME_FORMATTER, NoopRecordProcessorBlocker.create()) {
```
exceptionfactory, line 76: Can the `StandardRecordProcessorBlocker` be used here? It seems like the Noop instance is not required for standard behavior in this case.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/pause/TestStandardRecordProcessorBlocker.java
```diff
@@ -0,0 +1,114 @@
...
+
+        public void onPauseAwaited() {
+            onPauseAwaited = true;
+        }
+
+        public void onPauseFinished() {
+            onPauseFinished = true;
+        }
+
+        public void awaitAwaited() {
```
exceptionfactory, line 97:
```suggestion
        public void awaitStarted() {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/pause/TestStandardRecordProcessorBlocker.java
```diff
@@ -0,0 +1,114 @@
...
+        private boolean onPauseAwaited = false;
+        private boolean onPauseFinished = false;
+
+        public void onPauseAwaited() {
+            onPauseAwaited = true;
+        }
+
+        public void onPauseFinished() {
+            onPauseFinished = true;
+        }
```
exceptionfactory, line 95: It looks like these variables and methods should be updated to reflect the `Blocker` naming.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/StandardRecordProcessorBlocker.java
```diff
@@ -0,0 +1,44 @@
...
+    private CountDownLatch blocker = new CountDownLatch(0);
+
+    public static StandardRecordProcessorBlocker create() {
+        return new StandardRecordProcessorBlocker();
+    }
+
+    protected StandardRecordProcessorBlocker() { }
+
+    public void await() throws InterruptedException {
+        blocker.await();
```
exceptionfactory, line 32: What happens if `unblock()` and `block()` get called while the RecordProcessor is waiting on this call, since the `CountDownLatch` instance would be changed in the `block()` method?

The `CyclicBarrier` can be reset, but doesn't have all the semantics that seem to be required.

exceptionfactory (changes requested):
Thanks for the updates @dariuszseweryn, I raised one more key question regarding the `CountDownLatch` instance, and some test-related items, but otherwise this looks close to completion.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/StandardRecordProcessorBlocker.java
dariuszseweryn: `block()` does not create a new `CountDownLatch` if previous `CountDownLatch#getCount` above 0. Since `block` and `unblock` are synchronized there should not be a case that a RecordProcessor would wait indefinitely

sfc-gh-dhandermann: The `await()` method is not synchronized, so the Processor could call `unblock()` again. The issue doesn't seem to be with waiting indefinitely, but what happens to a RecordProcessor referencing a `CountDownLatch` instance that could be changed.

exceptionfactory: The `await()` method is not synchronized, so the issue isn't waiting indefinitely, but whether there are any issues with the `CountDownLatch` instance changing.

PR #9813, opened 2025-03-20
NIFI-14381: Add flow analysis rule to identify unset SSL Context Service controller services

mattyb149:
# Summary

[NIFI-14381](https://issues.apache.org/jira/browse/NIFI-14381) This PR introduces a Flow Analysis Rule that will examine specified (or all if the Component Type is left blank) components and if they have an SSL Context Service property that is unset will issue a violation.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireSecureConnection.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+
+@Tags({"component", "processor", "controller service", "type"})
+@CapabilityDescription("Produces rule violations for each component (i.e. processors or controller services) having a property "
+        + "identifying an SSLContextService that is not set.")
+@UseCase(
+        description = "Ensure that no ports can be opened for insecure (plaintext, e.g.) communications.",
```
exceptionfactory, line 45: Related to the name, this is not completely accurate for all cases.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireSecureConnection.java
```diff
@@ -0,0 +1,121 @@
...
+        configuration = """
+                To avoid the violation, ensure that the "SSL Context Service" property is set for the specified component(s).
+                """
+)
+public class RequireSecureConnection extends AbstractFlowAnalysisRule {
```
exceptionfactory, line 50: The name is somewhat misleading because certain components, such as `InvokeHTTP`, have a secure connection without configuring an SSL Context Service. Although more verbose, something like `RequireCustomSSLContext` would be more precise.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireSecureConnection.java
```diff
@@ -0,0 +1,121 @@
...
+    public static final PropertyDescriptor COMPONENT_TYPE = new PropertyDescriptor.Builder()
+            .name("component-type")
+            .displayName("Component Type(s)")
```
exceptionfactory, line 53: The `name()` should be used without the `displayName`, and the parentheses are not necessary.
```suggestion
            .name("Component Types")
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireSecureConnection.java
```diff
@@ -0,0 +1,121 @@
...
+            .description("A comma-separated list of component types. Components of the given type that have a property identifying an SSL Context Service will produce a rule violation "
+                    + "if the service is not set. If no components are specified (i.e. this property is blank), the rule will apply to all components.")
```
exceptionfactory, line 55: It would be helpful to indicate a few examples, such as `Processor`, and `ControllerService`, although this could be error-prone and may be better as a limited set of allowable values.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/pom.xml
```diff
@@ -23,7 +23,13 @@
     <artifactId>nifi-standard-rules</artifactId>
     <packaging>jar</packaging>
 
-    <dependencies />
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-ssl-context-service-api</artifactId>
+            <scope>compile</scope>
```
exceptionfactory, line 30: Should the scope be set to `provided`?

exceptionfactory (changes requested):
Thanks for proposing this new rule @mattyb149.

I noted a few questions and concerns related to the rule name, as `RequireSecureConnection` implies that `SSL Context Service` is required, which is not always the case.

The fact that some components do not require an `SSL Context Service` raises the question as to the validity of this rule for general purposes. As mentioned in the detailed comments, `InvokeHTTP` defaults to the system trust store, as do some other components. The fact that the nature of the default behavior can be different raises a question about the usefulness of this rule. For example, it would not work in a flow that requires a mixture of standard trusted certificates and custom trusted certificates for internal services. That could be addressed by always requiring a custom SSL Context Service configuration.

At minimum, changing the name and adjusting the description is important to convey the actual behavior. Thoughts?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireSecureConnection.java
mattyb149: I'm open to suggestions for the name, the idea is that you can't add a ListenHttp to the canvas and open an insecure port.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireSecureConnection.java
mattyb149: Works for me, will change the name

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireSecureConnection.java
mattyb149: Copy-paste :) will change

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireSecureConnection.java
exceptionfactory: Thanks for the reply. If that's the case, it sounds like it would be cleaner to limit this rule to the ListenHTTP Processor. That would avoid a lot of the potential ambiguity.

mattyb149: Maybe for all TCP-based Listen processors? ListenHTTP, ListenTCP, ListenOTLP, etc.?

exceptionfactory: The Component Types property is a bit ambiguous as it stands, because "Component Type" could imply Processor versus Controller Service. `Component Classes` might be clearer, requiring the full class name. However, if the goal is to focus on components that can receive connections, it seems like the list could be embedded in the Rule itself, not requiring the property.

mattyb149: I had to go a different route so I used a list of affected processors, but I also couldn't get at the actual PropertyDescriptor for the components, just the VersionedPropertyDescriptor, which does not identify the Controller Service reference, just that the property refers to a CS. This solution is brittle (based on property names) but I don't see a better way unless we can guarantee all components are InstantiatedVersionedComponents.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireCustomSSLContext.java
```diff
@@ -0,0 +1,84 @@
...
+                To avoid the violation, ensure that the "SSL Context Service" property is set for the specified component(s).
+                """
+)
+public class RequireCustomSSLContext extends AbstractFlowAnalysisRule {
+
+    private final List<String> componentTypes = List.of(
+            "ListenFTP",
+            "ListenHTTP",
+            "ListenTCP",
+            "ListenOLTP",
```
exceptionfactory, line 49: `ListenOTLP` always requires an `SSL Context Service`, so it can be removed.

pvillard31 (changes requested):
The JSON flow definitions in the tests contain vendor specific versions, can we have the same flows where vendor specific versions are replaced by Apache NiFi versions?

Also, just a suggestion: do we want to expose an optional property where an admin can provide additional processors to check? I'm thinking about custom extensions that are quite frequent. I remember numerous NiFi users with custom variations of ListenHTTP / HandleHTTPRequest processors.

exceptionfactory:
Regarding custom class names, I think the problem is that the Property Name has to be `SSL Context Service`, and that isn't guaranteed with other third-party components.  From that angle, I recommend using the full class name, not just the simple name, to make it clear that this applies only to Apache NiFi components.

pvillard31:
Yep, ok, my bad, I thought we had the information and we could do something like:

````java
if (component instanceof VersionedConfigurableComponent versionedConfigurableComponent) {
    final Map<String, VersionedPropertyDescriptor> props = versionedConfigurableComponent.getPropertyDescriptors();
    for (Entry<String, VersionedPropertyDescriptor> entry : props.entrySet()) {
        if (entry.getValue().getIdentifiesControllerService()) {
            // entry.getValue().getControllerServiceDefinition();
        }
    }
}
````

and have the information about the controller service interface that this property expects but this it not something available. That could be an interesting improvement I guess.

**Comment by mattyb149** [comment]
> Regarding custom class names, I think the problem is that the Property Name has to be `SSL Context Service`, and that isn't guaranteed with other third-party components. From that angle, I recommend using the full class name, not just the simple name, to make it clear that this applies only to Apache NiFi components.

I can't get the class name for it (see Pierre's comments above), we'd need an API change to add a method to retrieve the identified controller service unless there's some other way to get it (but I couldn't find one).

File: nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireCustomSSLContext.java [hunk 7]
```diff
@@ -0,0 +1,84 @@
...
+                To avoid the violation, ensure that the "SSL Context Service" property is set for the specified component(s).
+                """
+)
+public class RequireCustomSSLContext extends AbstractFlowAnalysisRule {
+
+    private final List<String> componentTypes = List.of(
+            "ListenFTP",
+            "ListenHTTP",
+            "ListenTCP",
+            "ListenTCPRecord",
```
exceptionfactory, line 49: `ListenTCPRecord` was removed in NiFi 2.0, so this reference should be removed.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/resources/docs/org.apache.nifi.flowanalysis.rules.RequireCustomSSLContext/additionalDetails.md
```diff
@@ -0,0 +1,28 @@
...
+# RequireCustomSSLContext
+
+## Usage Information
+
+This rule should be applied to prevent flow designers from opening ports for insecure connections. The list of components to which this rule applies is as follows:
+
+- **ListenFTP**
+- **ListenHTTP**
+- **ListenTCP**
+- **ListenTCPRecord**
```
exceptionfactory, line 25:
```suggestion
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireCustomSSLContext.java
```diff
@@ -0,0 +1,83 @@
...
+@Tags({"component", "processor", "controller service", "type", "ssl", "tls", "listen"})
+@CapabilityDescription("Produces rule violations for each component (i.e. processors or controller services) having a property "
+        + "identifying an SSLContextService that is not set.")
+@UseCase(
+        description = "Ensure that an SSL Context has been configured for the specified components. This helps avoid ports being opened for insecure (plaintext, e.g.) communications.",
+        configuration = """
+                To avoid the violation, ensure that the "SSL Context Service" property is set for the specified component(s).
+                """
+)
+public class RequireCustomSSLContext extends AbstractFlowAnalysisRule {
```
exceptionfactory, line 43: With the implement scope narrowed to components that listen for incoming connections, as opposed to any component, it would be helpful to adjust the rule name now. What do you think about `RequireServerSSLContextService`?

exceptionfactory (changes requested):
Thanks for the updates @mattyb149. This looks close to completion. Now that the scope is clear, I have one more recommendation regarding the rule name, then I think this is ready to proceed.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireCustomSSLContext.java
mattyb149: Works for me, will change

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireServerSSLContextService.java
```diff
@@ -0,0 +1,83 @@
...
+        + "identifying an SSLContextService that is not set.")
+@UseCase(
+        description = "Ensure that an SSL Context has been configured for the specified components. This helps avoid ports being opened for insecure (plaintext, e.g.) communications.",
+        configuration = """
+                To avoid the violation, ensure that the "SSL Context Service" property is set for the specified component(s).
+                """
+)
+public class RequireServerSSLContextService extends AbstractFlowAnalysisRule {
+
+    private final List<String> componentTypes = List.of(
```
exceptionfactory, line 45:
```suggestion
    private static final List<String> componentTypes = List.of(
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireServerSSLContextService.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public Collection<ComponentAnalysisResult> analyzeComponent(VersionedComponent component, FlowAnalysisRuleContext context) {
+        Collection<ComponentAnalysisResult> results = new HashSet<>();
+
+        if (component instanceof VersionedConfigurableExtension versionedConfigurableExtension) {
+
+            String encounteredComponentType = versionedConfigurableExtension.getType();
+            String encounteredSimpleComponentType = encounteredComponentType.substring(encounteredComponentType.lastIndexOf(".") + 1);
+
+            if (componentTypes.isEmpty() || componentTypes.contains(encounteredComponentType) || componentTypes.contains(encounteredSimpleComponentType)) {
```
exceptionfactory, line 63: This logic can be simplified since `componentTypes` is static:
```suggestion
            if (componentTypes.contains(encounteredComponentType)) {
```

exceptionfactory (changes requested):
Thanks for renaming, on final review, I noticed that the logic to check the component type includes some unnecessary checking for the simple type, and for an empty collection, neither of which are necessary.

exceptionfactory (approved):
Thanks again @mattyb149, the latest version looks good!

Any remaining comments @pvillard31?

pvillard31 (approved):

PR #9814, opened 2025-03-21
NIFI-14382: Create processor to list box metadata templates for file

ncover21:
# Summary

[NIFI-14382](https://issues.apache.org/jira/browse/NIFI-14382)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ListBoxFileMetadataTemplates.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ListBoxFileMetadataTemplates.java
@@ -38,16 +38,6 @@ import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
-import org.apache.nifi.schema.access.SchemaNotFoundException;
-import org.apache.nifi.serialization.RecordSetWriter;
-import org.apache.nifi.serialization.RecordSetWriterFactory;
-import org.apache.nifi.serialization.SimpleRecordSchema;
-import org.apache.nifi.serialization.WriteResult;
-import org.apache.nifi.serialization.record.MapRecord;
-import org.apache.nifi.serialization.record.Record;
-import org.apache.nifi.serialization.record.RecordField;
-import org.apache.nifi.serialization.record.RecordFieldType;
-import org.apache.nifi.serialization.record.RecordSchema;
 
 import java.io.IOException;
 import java.io.OutputStream;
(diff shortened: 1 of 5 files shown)


nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ListBoxFileMetadataTemplates.java
```diff
@@ -0,0 +1,231 @@
...
+                recordAttributes.put("box.metadata.templates.names", String.join(",", templateNames));
+                recordAttributes.put("box.metadata.templates.count", String.valueOf(templatesList.size()));
+                flowFile = session.putAllAttributes(flowFile, recordAttributes);
+
+                session.getProvenanceReporter().receive(flowFile, BoxFileUtils.BOX_URL + fileId);
+                session.transfer(flowFile, REL_SUCCESS);
+            } catch (final IOException e) {
+                getLogger().error("Failed writing metadata templates from file [{}]", fileId, e);
+                flowFile = session.putAttribute(flowFile, ERROR_MESSAGE, e.getMessage());
+                flowFile = session.penalize(flowFile);
```
pvillard31, line 199: I don't think I would penalize the flowfile here, if the user wants to retry the flowfile with a penalty duration, the user should use the retry mechanism at framework level instead of using a self loop relationship.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ListBoxFileMetadataTemplates.java
```diff
@@ -0,0 +1,231 @@
...
+                session.transfer(flowFile, REL_NOT_FOUND);
+            } else {
+                getLogger().error("Couldn't fetch metadata templates from file with id [{}]", fileId, e);
+                flowFile = session.penalize(flowFile);
+                session.transfer(flowFile, REL_FAILURE);
+            }
+        } catch (final Exception e) {
+            getLogger().error("Failed to process metadata templates for file [{}]", fileId, e);
+            flowFile = session.putAttribute(flowFile, ERROR_MESSAGE, e.getMessage());
+            flowFile = session.penalize(flowFile);
```
pvillard31, line 217: I don't think I would penalize the flowfile here, if the user wants to retry the flowfile with a penalty duration, the user should use the retry mechanism at framework level instead of using a self loop relationship.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ListBoxFileMetadataTemplates.java
```diff
@@ -0,0 +1,229 @@
...
+
+        } catch (final BoxAPIResponseException e) {
+            flowFile = session.putAttribute(flowFile, ERROR_CODE, valueOf(e.getResponseCode()));
+            flowFile = session.putAttribute(flowFile, ERROR_MESSAGE, e.getMessage());
+            if (e.getResponseCode() == 404) {
+                getLogger().warn("Box file with ID {} was not found.", fileId);
+                session.transfer(flowFile, REL_NOT_FOUND);
+            } else {
+                getLogger().error("Couldn't fetch metadata templates from file with id [{}]", fileId, e);
+                flowFile = session.penalize(flowFile);
```
pvillard31, line 210: Sorry for missing it earlier but I'd not penalize the flowfile here either. Unless there is a relationship that we really expect to be a self loop relationship on the component, I don't think it makes sense to penalize the flowfile. In case of a failure like this, the relationship would likely be connected to another processor for dealing with the failure and there is no reason to wait for the penalty duration before dealing with that error. And if the user wants to retry that failure, then that should be done by selecting the "retry" option on the failure relationship directly.

ncover21: Good with me, that makes sense. Removed

PR #9816, opened 2025-03-21
NIFI-14380 - OAuth 2.0 Controller Service to support JWT Bearer Flow

pvillard31:
# Summary

[NIFI-14380](https://issues.apache.org/jira/browse/NIFI-14380) - OAuth 2.0 Controller Service to support JWT Bearer Flow

This feature has been tested against
- Microsoft Sharepoint & MSGraph APIs
- Box
- Salesforce
- Google Identity for Google Cloud

Configuration settings for each scenario can be found in the additional details documentation of the controller service.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
@@ -19,7 +19,6 @@ package org.apache.nifi.oauth2;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
-import com.nimbusds.common.contenttype.ContentType;
 import com.nimbusds.jose.JOSEException;
 import com.nimbusds.jose.JOSEObjectType;
 import com.nimbusds.jose.JWSAlgorithm;
@@ -51,13 +50,13 @@ import org.apache.nifi.expression.AttributeExpression;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.key.service.api.PrivateKeyService;
 import org.apache.nifi.logging.ComponentLog;
+import org.apache.nifi.oauth2.key.Ed25519Signer;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.apache.nifi.ssl.SSLContextProvider;
 import org.apache.nifi.web.client.api.HttpRequestHeadersSpec;
 import org.apache.nifi.web.client.api.HttpResponseEntity;
 import org.apache.nifi.web.client.api.WebClientService;
 import org.apache.nifi.web.client.provider.api.WebClientServiceProvider;
-import org.apache.nifi.web.security.jwt.key.Ed25519Signer;
 
 import javax.net.ssl.X509ExtendedKeyManager;
 
@@ -103,10 +102,6 @@ import java.util.stream.Collectors;
 })
 public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerService implements OAuth2AccessTokenProvider, VerifiableControllerService {
 
-    private static final ObjectMapper ACCESS_DETAILS_MAPPER = new ObjectMapper()
-            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
-            .setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE);
-
     public static final PropertyDescriptor PRIVATE_KEY_SERVICE = new PropertyDescriptor.Builder()
             .name("Private Key Service")
             .description("The private key service to use for signing JWTs.")
@@ -117,9 +112,17 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
     public static final PropertyDescriptor SIGNING_ALGORITHM = new PropertyDescriptor.Builder()
             .name("Signing Algorithm")
             .description("The algorithm to use for signing the JWT.")
-            .allowableValues(JWSAlgorithm.RS256.getName(), JWSAlgorithm.RS384.getName(), JWSAlgorithm.RS512.getName(),
-                    JWSAlgorithm.PS256.getName(), JWSAlgorithm.PS384.getName(), JWSAlgorithm.PS512.getName(),
-                    JWSAlgorithm.ES256.getName(), JWSAlgorithm.ES384.getName(), JWSAlgorithm.ES512.getName(), JWSAlgorithm.Ed25519.getName())
+            .allowableValues(
+                    JWSAlgorithm.RS256.getName(),
+                    JWSAlgorithm.RS384.getName(),
+                    JWSAlgorithm.RS512.getName(),
+                    JWSAlgorithm.PS256.getName(),
+                    JWSAlgorithm.PS384.getName(),
+                    JWSAlgorithm.PS512.getName(),
+                    JWSAlgorithm.ES256.getName(),
+                    JWSAlgorithm.ES384.getName(),
+                    JWSAlgorithm.ES512.getName(),
+                    JWSAlgorithm.Ed25519.getName())
             .defaultValue(JWSAlgorithm.PS256.getName())
             .required(true)
             .addValidator(Validator.VALID)
@@ -175,27 +178,8 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
             .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
             .build();
 
-    public static final PropertyDescriptor NBF = new PropertyDescriptor.Builder()
-            .name("Set Not Before Time")
-            .description("The nbf (not before) claim identifies the time before which the JWT MUST NOT be accepted for processing."
-                    + " If set to true, the current time will be used.")
-            .required(true)
-            .allowableValues("true", "false")
-            .defaultValue("false")
-            .addValidator(Validator.VALID)
-            .build();
-
-    public static final PropertyDescriptor HEADER_TYPE = new PropertyDescriptor.Builder()
-            .name("Set JWT Header Type")
-            .description("If true, the JWT header type (typ) will be set to JWT.")
-            .required(true)
-            .allowableValues("true", "false")
-            .defaultValue("false")
-            .addValidator(Validator.VALID)
-            .build();
-
     public static final PropertyDescriptor HEADER_X5T = new PropertyDescriptor.Builder()
-            .name("Set JWT Header x5t")
+            .name("Set JWT Header X.509 Cert Thumbprint")
             .description("If true, will set the JWT header x5t field with the base64url-encoded SHA-256 thumbprint of the X.509 certificate's DER encoding."
                     + " If set to true, an instance of SSLContextProvider must be configured with a certificate using RSA algorithm.")
             .required(true)
@@ -272,9 +256,7 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
             SUBJECT,
             AUDIENCE,
             SCOPE,
-            NBF,
             JTI,
-            HEADER_TYPE,
             HEADER_X5T,
             SSL_CONTEXT_PROVIDER,
             KID,
@@ -283,6 +265,10 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
             REFRESH_WINDOW,
             JWT_VALIDITY
     );
+
+    private static final ObjectMapper ACCESS_DETAILS_MAPPER = new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE);
 
     private volatile AccessToken accessDetails;
     private volatile JWSSigner signer;
@@ -301,8 +287,6 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
     private volatile String kid;
     private volatile String grantType;
     private volatile String assertion;
-    private volatile boolean nbf;
-    private volatile boolean headerType;
     private volatile boolean headerX5T;
 
     private volatile Map<String, String> customClaims;
@@ -311,6 +295,9 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
     static final String CLAIM_PREFIX = "CLAIM.";
     static final String FORM_PREFIX = "FORM.";
 
+    private static final String APPLICATION_JSON = "application/json";
+    private static final String APPLICATION_URLENCODED = "application/x-www-form-urlencoded";
+
     @Override
     public List<PropertyDescriptor> getSupportedPropertyDescriptors() {
         return PROPERTY_DESCRIPTORS;
@@ -318,7 +305,7 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
 
     @OnEnabled
     public void onEnabled(ConfigurationContext context) {
-        getProperties(context);
+        initProperties(context);
         initJWTSigner();
     }
 
@@ -372,9 +359,8 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
         }
 
         if (privateKey instanceof RSAPrivateKey) {
-            if (algorithmName.startsWith("RS") || algorithmName.startsWith("PS")) {
-                return validationResults;
-            } else {
+            if (!algorithmName.startsWith("RS")
+                    && !algorithmName.startsWith("PS")) {
                 validationResults.add(new ValidationResult.Builder()
                         .subject(SIGNING_ALGORITHM.getDisplayName())
                         .valid(false)
@@ -382,18 +368,16 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
                         .build());
             }
         } else if (privateKey instanceof ECPrivateKey) {
-            if (algorithm.equals(JWSAlgorithm.ES256) || algorithm.equals(JWSAlgorithm.ES384) || algorithm.equals(JWSAlgorithm.ES512)) {
-                return validationResults;
-            } else {
+            if (!algorithm.equals(JWSAlgorithm.ES256)
+                    && !algorithm.equals(JWSAlgorithm.ES384)
+                    && !algorithm.equals(JWSAlgorithm.ES512)) {
                 validationResults.add(new ValidationResult.Builder()
                         .subject(SIGNING_ALGORITHM.getDisplayName())
                         .valid(false)
                         .explanation("The private key is of EC type, the signing algorithm must be either ES256, ES384 or ES512.")
                         .build());
             }
-        } else if (algorithm.equals(JWSAlgorithm.Ed25519)) {
-            return validationResults;
-        } else {
+        } else if (!algorithm.equals(JWSAlgorithm.Ed25519)) {
             validationResults.add(new ValidationResult.Builder()
                     .subject(SIGNING_ALGORITHM.getDisplayName())
                     .valid(false)
@@ -406,13 +390,13 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
 
     @Override
     public List<ConfigVerificationResult> verify(ConfigurationContext context, ComponentLog verificationLogger, Map<String, String> variables) {
-        getProperties(context);
+        initProperties(context);
         initJWTSigner();
         ConfigVerificationResult.Builder builder = new ConfigVerificationResult.Builder().verificationStepName("Acquire token");
         try {
             getAccessDetails();
             builder.outcome(ConfigVerificationResult.Outcome.SUCCESSFUL);
-        } catch (Exception ex) {
+        } catch (final Exception ex) {
             String explanation = ex.getMessage();
             if (ex.getCause() != null) {
                 explanation += " (" + ex.getCause().getMessage() + ")";
@@ -427,8 +411,8 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
         if (this.accessDetails == null || isRefreshRequired()) {
             try {
                 acquireAccessDetails();
-            } catch (Exception e) {
-                throw new AccessTokenRetrievalException("Failed to retrieve access token details", e);
+            } catch (final Exception e) {
+                throw new AccessTokenRetrievalException("Failed to acquire Access Token", e);
             }
         }
         return accessDetails;
@@ -454,10 +438,12 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
         getLogger().debug("New Access Token request started [{}]", tokenEndpoint);
 
         final Instant now = Instant.now();
+        final Date nowDate = Date.from(now);
 
         Builder claimsSetBuilder = new JWTClaimsSet.Builder()
                 .expirationTime(Date.from(now.plus(jwtValidity)))
-                .issueTime(Date.from(now));
+                .issueTime(nowDate)
+                .notBeforeTime(nowDate);
 
         if (issuer != null) {
             claimsSetBuilder.issuer(issuer);
@@ -475,10 +461,6 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
             claimsSetBuilder.claim("scope", scope);
         }
 
-        if (nbf) {
-            claimsSetBuilder.notBeforeTime(Date.from(now));
-        }
-
         if (jti != null) {
             claimsSetBuilder.jwtID(jti);
         }
@@ -486,10 +468,7 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
         customClaims.forEach(claimsSetBuilder::claim);
 
         JWSHeader.Builder headerBuilder = new JWSHeader.Builder(JWSAlgorithm.parse(algorithmName));
-
-        if (headerType) {
-            headerBuilder.type(JOSEObjectType.JWT);
-        }
+        headerBuilder.type(JOSEObjectType.JWT);
 
         if (kid != null) {
             headerBuilder.keyID(kid);
@@ -499,15 +478,15 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
             try {
                 final String alias = keyManager.chooseClientAlias(new String[] {"RSA"}, null, null);
                 if (alias == null) {
-                    getLogger().warn("No key alias found, setting x5t header to null.");
+                    throw new AccessTokenRetrievalException("Cannot set x5t header because no key alias found");
                 } else {
                     final PrivateKey privateKey = keyManager.getPrivateKey(alias);
                     if (privateKey == null) {
-                        getLogger().warn("No private key found, setting x5t header to null.");
+                        throw new AccessTokenRetrievalException(String.format("Cannot set x5t header because no private key found for alias %s", alias));
                     } else {
                         final X509Certificate[] certificates = keyManager.getCertificateChain(alias);
                         if (certificates == null || certificates.length == 0) {
-                            getLogger().warn("No certificates found");
+                            throw new AccessTokenRetrievalException(String.format("Cannot set x5t header because no certificate chain found for alias %s", alias));
                         } else {
                             final MessageDigest messageDigest = MessageDigest.getInstance("SHA-256");
                             final byte[] bytes = messageDigest.digest(certificates[0].getEncoded());
@@ -516,8 +495,10 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
                         }
                     }
                 }
-            } catch (Exception e) {
-                getLogger().warn("Failed to set x5t header, setting to null.", e);
+            } catch (final AccessTokenRetrievalException e) {
+                throw e;
+            } catch (final Exception e) {
+                throw new AccessTokenRetrievalException("Failed to set x5t header", e);
             }
         }
 
@@ -533,8 +514,8 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
         HttpRequestHeadersSpec request = webClientService
                 .post()
                 .uri(new URI(tokenEndpoint))
-                .header("Accept", ContentType.APPLICATION_JSON.getType())
-                .header("Content-Type", ContentType.APPLICATION_URLENCODED.getType())
+                .header("Accept", APPLICATION_JSON)
+                .header("Content-Type", APPLICATION_URLENCODED)
                 .body(formParams.entrySet()
                         .stream()
                         .map(param -> param.getKey() + "=" + param.getValue())
@@ -542,7 +523,7 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
 
         try (final HttpResponseEntity response = request.retrieve()) {
             if (response.statusCode() != 200) {
-                final String message = "Failed to retrieve access details. Status code: %s, response body: %s".formatted(
+                final String message = "Failed to retrieve Access Token: HTTP %s with Response [%s]".formatted(
                         response.statusCode(),
                         readBodyAsText(response).orElse("[no body]"));
 
@@ -573,7 +554,7 @@ public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerServic
         }
     }
 
-    private void getProperties(ConfigurationContext context) {
+    private void initProperties(ConfigurationContext context) {
         privateKey = context.getProperty(PRIVATE_KEY_SERVICE).asControllerService(PrivateKeyService.class).getPrivateKey();
         tokenEndpoint = context.getProperty(TOKEN_ENDPOINT).getValue();
         webClientService = context.getProperty(WEB_CLIENT_SERVICE).asControllerService(WebClientServiceProvider.class).getWebClientService();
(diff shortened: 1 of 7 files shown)


exceptionfactory:
Thanks for putting this together @pvillard31, I plan to review soon.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/test/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProviderTest.java
```diff
@@ -0,0 +1,300 @@
...
+
+    @Mock
+    private ComponentLog mockLogger;
+
+    private JWTBearerOAuth2AccessTokenProviderForTests provider;
+
+    @BeforeEach
+    void setUp() {
+        // Initialize mocks
+        MockitoAnnotations.openMocks(this);
```
exceptionfactory, line 81: This should be removed and replaced with the `ExtendWith(MockitoExtension.class)` annotation on the class.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/test/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProviderTest.java
```diff
@@ -0,0 +1,300 @@
...
+        when(mockWebClientServiceProvider.getWebClientService()).thenReturn(mockWebClientService);
+    }
+
+    @Test
+    void testClaimsAreSetProperly() throws Exception {
+        // Setup context
+        setContext();
+
+        setPrivateKeyMock(RSAPrivateKey.class);
+        when(mockContext.getProperty(JWTBearerOAuth2AccessTokenProvider.SIGNING_ALGORITHM).getValue()).thenReturn(JWSAlgorithm.RS256.getName());
```
exceptionfactory, line 102: Instead of mocking the context calls, it seems like it would be cleaner to use the TestRunner.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/pom.xml
```diff
@@ -27,4 +27,17 @@
         <module>nifi-oauth2-provider-service</module>
         <module>nifi-oauth2-provider-nar</module>
     </modules>
+    <repositories>
+        <!-- Shibboleth Repository required for OpenSAML -->
```
exceptionfactory, line 31: The Controller Service does not use OpenSAML, so is this required? It seems like it could be removed with the removal of `nifi-web-security`.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+import org.apache.nifi.expression.ExpressionLanguageScope;
+import org.apache.nifi.key.service.api.PrivateKeyService;
+import org.apache.nifi.logging.ComponentLog;
+import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.ssl.SSLContextProvider;
+import org.apache.nifi.web.client.api.HttpRequestHeadersSpec;
+import org.apache.nifi.web.client.api.HttpResponseEntity;
+import org.apache.nifi.web.client.api.WebClientService;
+import org.apache.nifi.web.client.provider.api.WebClientServiceProvider;
+import org.apache.nifi.web.security.jwt.key.Ed25519Signer;
```
exceptionfactory, line 60: Instead of importing this class from the `nifi-web-security` module, a separate copy should be placed in this module to avoid the dependency.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+                name = "FORM.Request parameter name",
+                value = "Request parameter value",
+                expressionLanguageScope = ExpressionLanguageScope.ENVIRONMENT,
+                description = "Custom parameters that should be added to the body of the request against the token endpoint.")
+})
+public class JWTBearerOAuth2AccessTokenProvider extends AbstractControllerService implements OAuth2AccessTokenProvider, VerifiableControllerService {
+
+    private static final ObjectMapper ACCESS_DETAILS_MAPPER = new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE);
```
exceptionfactory, line 108: Recommend moving this after the property descriptors.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+            .identifiesControllerService(PrivateKeyService.class)
+            .required(true)
+            .build();
+
+    public static final PropertyDescriptor SIGNING_ALGORITHM = new PropertyDescriptor.Builder()
+            .name("Signing Algorithm")
+            .description("The algorithm to use for signing the JWT.")
+            .allowableValues(JWSAlgorithm.RS256.getName(), JWSAlgorithm.RS384.getName(), JWSAlgorithm.RS512.getName(),
+                    JWSAlgorithm.PS256.getName(), JWSAlgorithm.PS384.getName(), JWSAlgorithm.PS512.getName(),
+                    JWSAlgorithm.ES256.getName(), JWSAlgorithm.ES384.getName(), JWSAlgorithm.ES512.getName(), JWSAlgorithm.Ed25519.getName())
```
exceptionfactory, line 122: Recommend formatting these values as one per line for readability.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+            .required(false)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
+            .build();
+
+    public static final PropertyDescriptor GRANT_TYPE = new PropertyDescriptor.Builder()
+            .name("Grant Type")
+            .description("Value to set for the grant_type parameter in the request to the token endpoint.")
+            .required(true)
+            .defaultValue("urn:ietf:params:oauth:grant-type:jwt-bearer")
```
exceptionfactory, line 164: Given that this implementation is `JWTBearer`, should the `Grant Type` be configurable?

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+            .name("Assertion Parameter Name")
+            .description("Name of the parameter to use for the JWT assertion in the request to the token endpoint.")
+            .required(true)
+            .defaultValue("assertion")
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
+            .build();
+
+    public static final PropertyDescriptor NBF = new PropertyDescriptor.Builder()
+            .name("Set Not Before Time")
```
exceptionfactory, line 179: Is there any reason not to set the `nbf` claim? It seems like it should always be included.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+            .description("The nbf (not before) claim identifies the time before which the JWT MUST NOT be accepted for processing."
+                    + " If set to true, the current time will be used.")
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .addValidator(Validator.VALID)
+            .build();
+
+    public static final PropertyDescriptor HEADER_TYPE = new PropertyDescriptor.Builder()
+            .name("Set JWT Header Type")
```
exceptionfactory, line 189: Is there any reason not to set this to `JWT` for all tokens as described in [RFC 7519](https://www.rfc-editor.org/rfc/rfc7519#section-5.1)?

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+            .description("If true, the JWT header type (typ) will be set to JWT.")
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .addValidator(Validator.VALID)
+            .build();
+
+    public static final PropertyDescriptor HEADER_X5T = new PropertyDescriptor.Builder()
+            .name("Set JWT Header x5t")
```
exceptionfactory, line 198: Perhaps using the more descriptive name would be helpful?
```suggestion
            .name("Set JWT Header X.509 Certificate Thumbprint")
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+            validationResults.add(new ValidationResult.Builder()
+                    .subject(HEADER_X5T.getDisplayName())
+                    .valid(false)
+                    .explanation("The private key must be an RSA key to set the x5t header.")
+                    .build());
+        }
+
+        if (privateKey instanceof RSAPrivateKey) {
+            if (algorithmName.startsWith("RS") || algorithmName.startsWith("PS")) {
+                return validationResults;
```
exceptionfactory, line 376: Having multiple nested `return` statements makes this method difficult to follow. Recommend refactoring to a single return and adjust the conditionals accordingly.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+
+    @Override
+    public List<ConfigVerificationResult> verify(ConfigurationContext context, ComponentLog verificationLogger, Map<String, String> variables) {
+        getProperties(context);
+        initJWTSigner();
+        ConfigVerificationResult.Builder builder = new ConfigVerificationResult.Builder().verificationStepName("Acquire token");
+        try {
+            getAccessDetails();
+            builder.outcome(ConfigVerificationResult.Outcome.SUCCESSFUL);
+        } catch (Exception ex) {
```
exceptionfactory, line 415: Minor convention recommendation:
```suggestion
        } catch (final Exception e) {
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+        return Arrays.asList(builder.build());
+    }
+
+    @Override
+    public AccessToken getAccessDetails() {
+        if (this.accessDetails == null || isRefreshRequired()) {
+            try {
+                acquireAccessDetails();
+            } catch (Exception e) {
+                throw new AccessTokenRetrievalException("Failed to retrieve access token details", e);
```
exceptionfactory, line 431:
```suggestion
                throw new AccessTokenRetrievalException("Failed to acquire Access Token", e);
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+
+        if (kid != null) {
+            headerBuilder.keyID(kid);
+        }
+
+        if (headerX5T) {
+            try {
+                final String alias = keyManager.chooseClientAlias(new String[] {"RSA"}, null, null);
+                if (alias == null) {
+                    getLogger().warn("No key alias found, setting x5t header to null.");
```
exceptionfactory, line 502: It is best to avoid ending log messages with a period character. However, shouldn't this throw an exception instead of logging a warning?
```suggestion
                    getLogger().warn("No key alias found, setting x5t header to null");
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+                } else {
+                    final PrivateKey privateKey = keyManager.getPrivateKey(alias);
+                    if (privateKey == null) {
+                        getLogger().warn("No private key found, setting x5t header to null.");
```
exceptionfactory, line 506: As above, shouldn't this condition throw an exception?

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+                        } else {
+                            final MessageDigest messageDigest = MessageDigest.getInstance("SHA-256");
+                            final byte[] bytes = messageDigest.digest(certificates[0].getEncoded());
+                            String url = Base64.getEncoder().encodeToString(bytes);
+                            headerBuilder.x509CertSHA256Thumbprint(new Base64URL(url));
+                        }
+                    }
+                }
+            } catch (Exception e) {
+                getLogger().warn("Failed to set x5t header, setting to null.", e);
```
exceptionfactory, line 520: See above notes on exception and message formatting.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+                .header("Accept", ContentType.APPLICATION_JSON.getType())
+                .header("Content-Type", ContentType.APPLICATION_URLENCODED.getType())
+                .body(formParams.entrySet()
+                        .stream()
+                        .map(param -> param.getKey() + "=" + param.getValue())
+                        .collect(Collectors.joining("&")));
+
+        try (final HttpResponseEntity response = request.retrieve()) {
+            if (response.statusCode() != 200) {
+                final String message = "Failed to retrieve access details. Status code: %s, response body: %s".formatted(
```
exceptionfactory, line 545: Recommend adjusting the formatting
```suggestion
                final String message = "Failed to retrieve Access Token: HTTP %s with Response [%s]".formatted(
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,677 @@
...
+        try (final InputStream is = response.body()) {
+            final String body = new String(is.readAllBytes(), StandardCharsets.UTF_8);
+            return Optional.of(body);
+        } catch (final IOException e) {
+            getLogger().warn("Failed to read response body", e);
+            return Optional.empty();
+        }
+    }
+
+    private void getProperties(ConfigurationContext context) {
```
exceptionfactory, line 576: This seems like it should be named `setProperties()` since it doesn't return anything.

exceptionfactory (changes requested):
Thanks for putting this together @pvillard31. It covers a lot of use cases, which adds complexity, but reflects the state of various services. I noted a number of recommendations, with the most important one being the removal of the `nifi-web-security` dependency.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
pvillard31: Based on the documentation, there are implementations not expecting it and where it fails if it is set... I could make the tests again to tell you again which test implementation exactly does not like it when set.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
pvillard31: Same answer: some implementations are very strict in terms of expected fields in the header. I will test again to double check to be sure.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
pvillard31: So I felt like not setting the field, logging a warning to have a bulletin on the controller service, and still trying to acquire the token was the best approach but happy to consider an exception instead if we feel like it's better to not even try.

exceptionfactory: Given that the property configuration expects the Certificate Thumbprint to be set, I think this should throw an exception as opposed to trying something that seems very unlikely to work.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
pvillard31: My bad, Salesforce seems to be OK when the header type is set, so I'll remove this property

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
pvillard31: My bad as well :) tested again all 4 services and they are all fine with NBF being set. Will remove.

pvillard31:
Thanks for the initial review @exceptionfactory - I pushed a commit to address your comments.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/test/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProviderTest.java
pvillard31: Pushed a commit where I'm using the test runner instead. Not ideal as I can't call the customValidate directly and specifically check what is making the controller service invalid but that could be a follow up improvement in the test framework.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+            .identifiesControllerService(SSLContextProvider.class)
+            .build();
+
+    public static final PropertyDescriptor JTI = new PropertyDescriptor.Builder()
+            .name("JWT ID")
+            .description("The \"jti\" (JWT ID) claim provides a unique identifier for the JWT. The identifier value must be assigned in a "
+                    + "manner that ensures that there's a negligible probability that the same value will be accidentally assigned to a "
+                    + "different data object; if the application uses multiple issuers, collisions MUST be prevented among values produced "
+                    + "by different issuers as well. The \"jti\" value is a case-sensitive string. If set, it is recommended to set this "
+                    + "value to ${UUID()}.")
```
exceptionfactory, line 206: This could be reformatted using a multi line string instead of using concatenation.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+            .required(true)
+            .addValidator(StandardValidators.URL_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor WEB_CLIENT_SERVICE = new PropertyDescriptor.Builder()
+            .name("Web Client Service")
+            .description("The Web Client Service to use for calling the token endpoint.")
+            .identifiesControllerService(WebClientServiceProvider.class)
+            .required(true)
+            .build();
```
exceptionfactory, line 232: It seems like the URL and Web Client Service properties should be moved near the top along with other required properties.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+            .description("The service will attempt to refresh tokens expiring within the refresh window, subtracting the configured duration from the token expiration.")
+            .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
+            .defaultValue("5 minutes")
+            .required(true)
+            .build();
+
+    static final PropertyDescriptor JWT_VALIDITY = new PropertyDescriptor.Builder()
+            .name("JWT Expiration Time")
+            .description("Expiration time used to set the corresponding claim of the JWT. In case the returned access token does not include an expiration time, this will"
+                    + " be used with the refresh window to re-acquire a new access token.")
```
exceptionfactory, line 245: Multi line string could be used.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+    }
+
+    private void acquireAccessDetails() throws URISyntaxException, JOSEException {
+        getLogger().debug("New Access Token request started [{}]", tokenEndpoint);
+
+        final Instant now = Instant.now();
+        final Date nowDate = Date.from(now);
+
+        Builder claimsSetBuilder = new JWTClaimsSet.Builder()
+                .expirationTime(Date.from(now.plus(jwtValidity)))
```
exceptionfactory, line 446: Recommend declaration `expirationTime` as a separate variable for readability and potential debugging.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+                        final X509Certificate[] certificates = keyManager.getCertificateChain(alias);
+                        if (certificates == null || certificates.length == 0) {
+                            throw new AccessTokenRetrievalException(String.format("Cannot set x5t header because no certificate chain found for alias %s", alias));
+                        } else {
+                            final MessageDigest messageDigest = MessageDigest.getInstance("SHA-256");
+                            final byte[] bytes = messageDigest.digest(certificates[0].getEncoded());
+                            String url = Base64.getEncoder().encodeToString(bytes);
+                            headerBuilder.x509CertSHA256Thumbprint(new Base64URL(url));
+                        }
+                    }
```
exceptionfactory, line 498: Recommend moving the digest calculation to a private method.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+        signedJWT.sign(signer);
+        return signedJWT.serialize();
+    }
+
+    private Optional<String> readBodyAsText(final HttpResponseEntity response) {
+        try (final InputStream is = response.body()) {
+            final String body = new String(is.readAllBytes(), StandardCharsets.UTF_8);
+            return Optional.of(body);
+        } catch (final IOException e) {
+            getLogger().warn("Failed to read response body", e);
```
exceptionfactory, line 554: It would be helpful to include the HTTP Status Code, and the URL if available for troubleshooting this warning.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+                }
+            }
+        }
+    }
+
+    private void initJWTSigner() {
+        JWSAlgorithm algorithm = JWSAlgorithm.parse(algorithmName);
+
+        if (privateKey instanceof RSAPrivateKey) {
+            if (algorithmName.startsWith("RS") || algorithmName.startsWith("PS")) {
```
exceptionfactory, line 631: Is the algorithmName check required here? It should be enforced through validation, and the `RSASSASigner` should also check the key type.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+                            URLEncoder.encode(context.getProperty(descriptor).evaluateAttributeExpressions().getValue(), StandardCharsets.UTF_8));
+                }
+            }
+        }
+    }
+
+    private void initJWTSigner() {
+        JWSAlgorithm algorithm = JWSAlgorithm.parse(algorithmName);
+
+        if (privateKey instanceof RSAPrivateKey) {
```
exceptionfactory, line 630: The cast variable name can be used, instead of casting below.
```suggestion
        if (privateKey instanceof RSAPrivateKey rsaPrivateKey) {
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
```diff
@@ -0,0 +1,664 @@
...
+            if (algorithmName.startsWith("RS") || algorithmName.startsWith("PS")) {
+                signer = new RSASSASigner((RSAPrivateKey) privateKey);
+            }
+        } else if (privateKey instanceof ECPrivateKey) {
```
exceptionfactory, line 634:
```suggestion
        } else if (privateKey instanceof ECPrivateKey ecPrivateKey) {
```

exceptionfactory (changes requested):
Thanks for moving this forward @pvillard31, it looks closer to completion, I noted a few minor adjustments and recommendations.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/JWTBearerOAuth2AccessTokenProvider.java
pvillard31: This method is called from from error handling in `requestTokenEndpoint` where we log the HTTP Code and response. I added the URL as well. Not sure we should also add this information in this specific warn log, no?

exceptionfactory: Good point, in this case, I recommend removing this warning. It would probably be better to throw the exception, catch it in the calling method, and throw it as the cause of the `AccessTokenRetrievalException`.

pvillard31: Fair enough, pushed a commit to have a better logic there

exceptionfactory: Thanks, the updated approach looks good.

exceptionfactory (approved):
Thanks for working through the feedback @pvillard31, the latest version looks good. +1 merging

PR #9822, opened 2025-03-25
NIFI-14389 - Provide the option to force refresh Access Token in OAuth2AccessTokenProvider

pvillard31:
# Summary

[NIFI-14389](https://issues.apache.org/jira/browse/NIFI-14389) - Provide the option to force refresh Access Token in OAuth2AccessTokenProvider

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
@@ -29,9 +29,10 @@ public interface OAuth2AccessTokenProvider extends ControllerService {
     AccessToken getAccessDetails();
 
     /**
-     * Refreshes the access token even if it has not expired yet
+     * Request a new Access Token based on configured properties regardless of
+     * current expiration status. The default implementation does not perform any
+     * action.
      */
     default void refreshAccessDetails() {
-        // no default implementation
     }
 }
--- a/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
@@ -29,12 +29,9 @@ public interface OAuth2AccessTokenProvider extends ControllerService {
     AccessToken getAccessDetails();
 
     /**
-     * @param forceAccessTokenRefresh true if the access token should be refreshed
-     *                                even if it is not expired
-     * @return A valid access token (refreshed automatically if needed) and
-     *         additional metadata (provided by the OAuth2 access server)
+     * Refreshes the access token even if it has not expired yet
      */
-    default AccessToken getAccessDetails(final boolean forceAccessTokenRefresh) {
-        return getAccessDetails();
+    default void refreshAccessDetails() {
+        // no default implementation
     }
 }
--- a/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
@@ -27,4 +27,14 @@ public interface OAuth2AccessTokenProvider extends ControllerService {
      * @return A valid access token (refreshed automatically if needed) and additional metadata (provided by the OAuth2 access server)
      */
     AccessToken getAccessDetails();
+
+    /**
+     * @param forceAccessTokenRefresh true if the access token should be refreshed
+     *                                even if it is not expired
+     * @return A valid access token (refreshed automatically if needed) and
+     *         additional metadata (provided by the OAuth2 access server)
+     */
+    default AccessToken getAccessDetails(final boolean forceAccessTokenRefresh) {
+        return getAccessDetails();
+    }
 }
--- a/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/TokenRefreshStrategy.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/TokenRefreshStrategy.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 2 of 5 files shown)


exceptionfactory (changes requested):
Thanks for proposing this improvement @pvillard31. The general concept makes sense, but what do you think about implementing this as a strategy property, such as `Token Refresh Strategy`? That could provide the opportunity for clearer definition of options, such as `On Token Expiration` for current behavior and `On Unauthorized Status` for the proposed behavior. It would also make it easier to implement additional options in the future if needed.

pvillard31:
Yep, good point @exceptionfactory, it will be cleaner

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/TokenRefreshStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package org.apache.nifi.oauth2;
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum TokenRefreshStrategy implements DescribedValue {
+
+    ON_TOKEN_EXPIRATION("The token will be refreshed based on its expiration time and the configured refresh window"),
+
+    ON_NON_AUTHORIZED_REQUEST("A new token will be requested in case of a non-authorized request (HTTP 401) even if the current token has not expired");
```
exceptionfactory, line 25: I recommend `ON_UNAUTHORIZED_STATUS` or `ON_UNAUTHORIZED_RESPONSE` to align with the standard name for [HTTP 401](https://http.dev/401) status.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
```diff
@@ -27,4 +27,14 @@ public interface OAuth2AccessTokenProvider extends ControllerService {
...
      */
     AccessToken getAccessDetails();
+
+    /**
+     * @param forceAccessTokenRefresh true if the access token should be refreshed
+     *                                even if it is not expired
+     * @return A valid access token (refreshed automatically if needed) and
+     *         additional metadata (provided by the OAuth2 access server)
+     */
+    default AccessToken getAccessDetails(final boolean forceAccessTokenRefresh) {
```
exceptionfactory, line 37: Instead of adding this new method and option, what do you think about promoting `refreshAccessDetails()` as an interface method? This would require calling both `refreshAccessDetails()` and then `getAccessDetails()`, but it would clarify that these are separate actions, the first of which could fail, allowing the caller to determine the desired behavior. The default implementation could be nothing, making it easier for third-party implementations to implement when desired, aligning with the optional usage of the new Token Refresh Strategy.

exceptionfactory (changes requested):
Thanks for adjusting the property definition approach @pvillard31. Reviewing the Access Token Provider method, I raised a question on the method definition.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
pvillard31: Yeah, you're right, I think it makes it easier from a caller perspective to understand the expected usage of those methods.

pvillard31: Thinking a bit more about it... `refreshAccessDetails()`, as of now, assumes that the refresh token is not null. Some services do not send a refresh token with the access token. So I'll have to adjust the logic just a bit.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
```diff
@@ -27,4 +27,11 @@ public interface OAuth2AccessTokenProvider extends ControllerService {
      * @return A valid access token (refreshed automatically if needed) and additional metadata (provided by the OAuth2 access server)
      */
     AccessToken getAccessDetails();
+
+    /**
+     * Refreshes the access token even if it has not expired yet
```
exceptionfactory, line 32:
```suggestion
     * Request a new Access Token based on configured properties regardless of current expiration status. The default implementation does not perform any action.
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/OAuth2AccessTokenProvider.java
```diff
@@ -27,4 +27,11 @@ public interface OAuth2AccessTokenProvider extends ControllerService {
...
+     */
+    default void refreshAccessDetails() {
+        // no default implementation
```
exceptionfactory, line 35: Recommend removing this comment and incorporating in the method comments.
```suggestion
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/StandardOauth2AccessTokenProvider.java
```diff
@@ -339,6 +340,31 @@ public AccessToken getAccessDetails() {
         return accessDetails;
     }
 
+    @Override
+    public void refreshAccessDetails() {
+
```
exceptionfactory, line 345: Unnecessary newline can be removed:
```suggestion
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/StandardOauth2AccessTokenProvider.java
```diff
@@ -339,6 +340,31 @@ public AccessToken getAccessDetails() {
...
+        addFormData(refreshTokenBuilder);
+
+        AccessToken newAccessDetails = requestToken(refreshTokenBuilder);
+
+        if (newAccessDetails.getRefreshToken() == null) {
+            newAccessDetails.setRefreshToken(this.accessDetails.getRefreshToken());
+        }
+
+        this.accessDetails = newAccessDetails;
```
exceptionfactory, line 365: Minor, but I recommend placing the logic in an `else` condition instead of the short-circuit return.

exceptionfactory (changes requested):
Thanks for the updates @pvillard31, the `refreshAccessDetails()` method and implementation look good. Just a couple minor documentation and structural adjustments, and then this looks ready to go.

exceptionfactory (approved):

PR #9843, opened 2025-04-03
NIFI-14413 - Add User Agent to NiFi User Model

arturchyzy:
# Summary

[NIFI-14413](https://issues.apache.org/jira/browse/NIFI-14413)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for adding the User Agent property @arturchyzy.

The general approach looks good, but I noticed a gap related to token authentication. I noticed the `StandardJwtAuthenticationConverter` also creates a NiFi User, but is not included in these updates. That converter does not have access to the request information, which highlights a potential design issue.

Spring Security has an `AuthenticationDetailsSource`, with implementations that return additional information about the request. The `WebAuthenticationDetails` is a useful implementation, which includes the Remote Address. Extending that approach seems like the right abstraction for adding the User Agent. That appears to require some other adjustments to the general implementation strategy, but I think that is what needs to happen.

sfc-gh-achyzy:
@exceptionfactory I've pushed the changes using AuthenticationDetailsSource. I have to admit it was a bit challenging to track all the authentication types but I think I was able to support all of them.

exceptionfactory (approved):
Thanks for making the adjustments to use the Authentication Details @arturchyzy, the latest version looks good! +1 merging

PR #9847, opened 2025-04-04
NIFI-1842 Added Apache Camel bundle

luigiba:
# Summary

[NIFI-1842](https://issues.apache.org/jira/browse/NIFI-1842)
[NIFI-924](https://issues.apache.org/jira/browse/NIFI-924)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-camel-bundle/nifi-camel-nar/pom.xml
+++ b/nifi-extension-bundles/nifi-camel-bundle/nifi-camel-nar/pom.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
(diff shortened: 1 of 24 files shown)


exceptionfactory (changes requested):
Thanks for proposing this integration with Camel @luigiba, however, I'm going to vote this down based the general impedance mismatch between Camel and NiFi.

In past years, NiFi has had other Processors that supported generalized framework integration, including Apache Flume and Spring Framework. Although this type of generalized component seems appealing at first, it presents a large number of integration and troubleshooting challenges. Generalized framework integration often requires a large number of dependencies, only some of which are actually required at runtime. Due to the abstract nature of such integrations, using them requires a thorough understanding of both NiFi and the integrated framework. This leads to troubleshooting difficulties when attempting to track down behavior or performance issues, making it unclear whether the problem is with NiFi or with the integrated framework.

This is not a negative reflection on Apache Camel or other generalized framework. It reflects a number of lessons learned from historical attempts to integrate disparate frameworks in NiFi 1, which were abandoned and removed in NiFi 2.

This is also not ruling out the potential value of such integrations for certain use cases. For this reason, I recommend providing the components in a separate GitHub repository as a way to gather feedback and improvements.

Thank you again for presenting the pull request, but I am voting -1 and closing it for now.

luigiba:
Thanks for your feedback. I understand your point of view. However, I’d like to clarify that the Camel integration I’m proposing is tailored specifically to a small subset of Camel components on which we can discuss its utility. In this pull request Camel is not integrated as a generic framework but instead as a set of further declarative APIs (YAML/XML) that can be used to transform and enrich the flow files content/attributes. This makes the camel integration much more light-weight and NIFI still remains the main in / out framework.

joewitt:
Given our history of integrating similar projects and how it worked out in practice from both a maintenance and usability point of view I share @exceptionfactory view that this should not proceed at this time.

However, I do note that the idea itself in terms of usefulness to users which do have Camel knowledge and want to benefit from the combined strength - makes sense.

What I recommend is rather that introducing this to the Apache NiFI community itself you simply create an extension nar in your own github repository and start a simply project for this.  You can let folks know about it and they can bring it into their NiFi deployments.  This is done for many other components.  Some things may in a sense graduate to the apache nifi repo or some things are just great/useful on their own and people can use them.  Or some ideas are found in practice to not really work out as intended and it is no harm no foul.

Thanks

PR #9848, opened 2025-04-04
NIFI-14435 - FlowComparator should offer deep recursive comparison

pvillard31:
# Summary

[NIFI-14435](https://issues.apache.org/jira/browse/NIFI-14435) - FlowComparator should offer deep recursive comparison

> Consider a versioned process group PG.
> Add a new child process group inside of PG.
> Do many modifications in the child PG (add processors, etc).
> If showing local changes for PG, it'll only say that a process group childPG has been added but nothing about what has been done in childPG.
> This JIRA is to add a new Flow Comparator Versioned Strategy in order to offer the option for a full exhaustive listing of the differences to also include the list of what is being added inside a newly added process group.
> There is no intent with this work to change what is returned to the UI when showing local changes as it should be further evaluated in terms of side effects.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/FlowComparatorVersionedStrategy.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/FlowComparatorVersionedStrategy.java
@@ -26,12 +26,6 @@ public enum FlowComparatorVersionedStrategy {
      */
     DEEP,
     /**
-     * The whole versioned process group is considered and every component change
-     * should appear as a difference. This also includes the list of changes within
-     * a newly added process group. Version information is included as well.
-     */
-    DEEP_WITH_RECURSIVE_PG,
-    /**
      * The comparator should disregard individual changes but only look for changes
      * in the version information.
      */
--- a/nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/FlowComparatorVersionedStrategy.java
+++ b/nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/FlowComparatorVersionedStrategy.java
@@ -26,7 +26,14 @@ public enum FlowComparatorVersionedStrategy {
      */
     DEEP,
     /**
-     * The comparator should disregard individual changes but only look for changes in the version information.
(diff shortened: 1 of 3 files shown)


exceptionfactory (commented):
Thanks for proposing these changes @pvillard31.

On initial review, I'm wondering if the Process Group addition should be included automatically in the `DEEP` strategy. On the one hand, it would surface a large number of additional changes, but it seems to align with the intended meaning of `DEEP`.

On the other hand, `DEEP_WITH_RECURSIVE_PG` doesn't seem to precisely reflect the desired behavior, as this applies only to an added PG, correct? What about removing a child PG?

Perhaps a different strategy name could help, but introducing the new strategy just for this particular aspect of the comparison seems worth some additional consideration on the potential best way forward.

pvillard31:
Thanks @exceptionfactory for the feedback.

Great point on deletion. My thoughts were really in the context of showing changes between two versions (n and n+1) and in my mind, whoever is looking at the changes is most likely already familiar with version n, so I thought that showing only that the process group (and its content) was deleted is enough because I assumed that the person already knows what was inside the deleted process group. When adding a process group, there is no way to already know what's inside. That's the way I approached it. I can definitely also consider full details for the removal of a process group.

I do agree with you that the existing strategy `DEEP` had a good meaning but it sounds like this is more about the level of metadata being returned rather than "deep" in the sense of recursive listing.

I did think about just updating the DEEP strategy and I'm OK taking this approach to be honest, I was just concerned about potential side effects on existing APIs that rely on this (show local changes for versioned process groups being one).

With this additional context, do you have any preference?

exceptionfactory (commented):
Thanks for the additional background @pvillard31.

Considering the options, I'm inclined to think that enumerating the full list of changes as part of the current `DEEP` strategy is the best way to go.

Although it will expand the set of changes shown, I think it aligns best with the current description of the `DEEP` strategy, enabling users to see the full list of changes, as opposed to a summary Process Group that may contain a large set of changes.

If that becomes a problem for some reason, I could see revisiting the available strategies to narrow it down in places, but it seems better to include the full set of details as the standard behavior.

pvillard31:
Thanks @exceptionfactory - I just pushed a commit that removes the new strategy and instead updates the existing `DEEP` strategy. I also added full details for both deletion and addition of a child process group.

exceptionfactory (approved):
Thanks for making the adjustments @pvillard31, the updated approach looks good. +1 merging

PR #9849, opened 2025-04-04
NIFI-14191 Added the previous process group to the URL

homer-rich:
Added the previous process group to the end of the URL so that it may be centered with the click of the zoom button when navigating up a process group via the LeaveProcessGroup function.

# Summary

[NIFI-14191](https://issues.apache.org/jira/browse/NIFI-14191)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-14191) issue created

### Pull Request Tracking

- [X] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-14191`
- [X] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-14191`

### Pull Request Formatting

# Verification

Tested using the hot reloading available via `npm nx serve` in the frontend folder.

### Build

- [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/canvas.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/canvas.component.ts
@@ -172,7 +172,7 @@ export class Canvas implements OnInit, OnDestroy {
                 if (skipTransform) {
                     this.store.dispatch(setSkipTransform({ skipTransform: false }));
                 } else {
-                    this.store.dispatch(centerSelectedComponents({ request: { allowTransition } }));
+                    this.store.dispatch(restoreViewport());
                 }
             });
 
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/state/flow/flow.effects.ts
@@ -2808,16 +2808,25 @@ export class FlowEffects {
         { dispatch: false }
     );
 
-    leaveProcessGroup$ = createEffect(
-        () =>
-            this.actions$.pipe(
-                ofType(FlowActions.leaveProcessGroup),
-                concatLatestFrom(() => this.store.select(selectParentProcessGroupId)),
-                filter(([, parentProcessGroupId]) => parentProcessGroupId != null),
-                tap(([, parentProcessGroupId]) => {
-                    this.router.navigate(['/process-groups', parentProcessGroupId]);
-                })
-            ),
+    leaveProcessGroup$ = createEffect(() =>
+        this.actions$.pipe(
+            ofType(FlowActions.leaveProcessGroup),
+            concatLatestFrom(() => [
+                this.store.select(selectParentProcessGroupId).pipe(isDefinedAndNotNull()),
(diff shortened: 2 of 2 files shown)


mcgilman:
Thanks for the PR @homer-rich! I'll have a look.

mcgilman:
I haven't had a chance to fully review, but I wanted to clarify the desired behavior. The current behavior of leave process group will navigate up one level and restore the users previous viewport. Here we are trying to retain that functionality but also select the child Process Group was just exited.

I have some concern at first glance (but I need to verify) that we're losing the centering behavior which would trigger when a user navigates to a selected component without having established a viewport to restore (like through a deep link).

homer-rich:
That is the correct interpretation of what I am trying to accomplish with the desired behavior.  I was wondering if there was a way to respond differently if the link was deep vice using the Leave Group function.  I'm new to JS/TS development, apologies in advance.

mosermw:
This feels like a really useful feature to me.  I tested it and I like how it works.

To test @mcgilman's concern, I grabbed a deep link to a processor that was two groups down.  Then I cleared my browser cache and cookies and restarted the browser.  When I navigated to the deep link, the processor was selected and the viewport was centered in the PG.  Leave Group twice also centered those viewports.  I moved the viewport on the way back down to the original processor, and on the way back out the shifted viewports were restored properly.

Even though I don't understand the intricacies of the Typescript code in this PR, it appears simple enough that it can be merged with thorough testing.  So if this PR expires, I'm inclined to re-open and merge.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/canvas.component.ts
```diff
@@ -172,7 +172,7 @@ export class Canvas implements OnInit, OnDestroy {
                 if (skipTransform) {
                     this.store.dispatch(setSkipTransform({ skipTransform: false }));
                 } else {
-                    this.store.dispatch(centerSelectedComponents({ request: { allowTransition } }));
+                    this.store.dispatch(restoreViewport());
```
mcgilman, line 175: This change is problematic. This was the concern I mentioned initially though the example I gave was wrong. It's removes centering behavior and replaces it with restoring the most recent viewport. An example of a sequence where this becomes an issue is when using the Search capabilities to find another component on the canvas in the current Process Group that is currently out of view. Attempting to navigate to the component through the search results won't work since the user is already centered on the current viewport. 

This change shouldn't be needed though but I think I understand why it was included. This sequence will fire twice... once for when the route changes and once for when the store is reloaded (because we're changing Process Groups). These both fire independently/separately. As a result, the Process Group is centering when that wasn't the intent. I would need to consider this scenario more carefully. Other instances of `navigateWithoutTransform` happen within the current Process Group so the current Process Group doesn't fire and doesn't showcase this issue.

mcgilman (changes requested):
Thanks for the PR @homer-rich. Apologies for losing track of this one. Please see the feedback below. The behavior that you were trying to work around with the call to `restoreViewport` will need further consideration to preventing the centering behavior.

exceptionfactory:
Thanks for the initial work on this @homer-rich, and thanks for reviewing and highlighting issues @mosermw and @mcgilman. Given the lack of activity over multiple months, and the current implementation issue, I am going to close this pull request for now. Feel free to reopen when you are ready to address the feedback.

PR #9850, opened 2025-04-04
NIFI-14392 Migrate nifi-1 local state

tpalfy:
# Summary

[NIFI-14392](https://issues.apache.org/jira/browse/NIFI-14392)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-write-ahead-log/src/main/java/org/wali/MinimalLockingWriteAheadLog.java
+++ b/nifi-commons/nifi-write-ahead-log/src/main/java/org/wali/MinimalLockingWriteAheadLog.java
@@ -1,0 +1,1185 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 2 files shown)


joewitt:
I dont think we're looking to re-introduce deprecated and removed code to provide a migration tool but back into the nifi codebase.

We need to look at migration tooling as its own repo or bespoke tool or some other solution.

exceptionfactory (changes requested):
Concurring with @joewitt I am voting -1 on this proposed addition of the deprecated implementation.

Historical implementations could be provided outside of the project, but that does not present a migration strategy within the project itself.

As a general approach, a current implementation could attempt to evaluate the current state and migrate to the new format. This would take some work to implement properly, but it would provide a migration strategy that avoids introducing deprecated components.

exceptionfactory (commented):
For clarification, the migration in the current implementation seems on a good path, but it should not require reintroducing the entire deprecated class.

exceptionfactory:
@tpalfy Given the feedback, it seems like the best option is to start a discussion on the mailing list about introducing this change in the NiFi 1 support branch for a potential 1.28.2 version. It would require some concurrence and general agreement on putting in the work to conduct another release, but that would align with other migration patterns that exist only in that version.

Unless any other strategies come to mind, it looks like this should be closed.

PR #9851, opened 2025-04-06
NIFI-14436 TestRunner ensures properties with unsatisfied dependencies are not used

EndzeitBegins:
# Summary

[NIFI-14436](https://issues.apache.org/jira/browse/NIFI-14436)

Adjusts the `TestRunner` to ensure properties whose dependencies are not satisfied are not used. See [NIFI-14400](https://issues.apache.org/jira/browse/NIFI-14400) for more background on the motivation behind this change.

~~The added behaviour is opt-out using `runner.setProhibitUseOfPropertiesWithUnsatisfiedDependencies(false)`. There are several existing components which do not conform to the new check. These are to be addressed in separate PRs under NIFI-14437 or a sub-task thereof.~~
The behaviour cannot be disabled. All components that did not pass the added check were refactored accordingly. 

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/ConsumeAMQP.java
+++ b/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/ConsumeAMQP.java
@@ -22,7 +22,6 @@ import com.rabbitmq.client.AMQP.BasicProperties;
 import com.rabbitmq.client.Connection;
 import com.rabbitmq.client.Envelope;
 import com.rabbitmq.client.GetResponse;
-import java.util.EnumSet;
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
 import org.apache.nifi.annotation.behavior.WritesAttribute;
@@ -122,7 +121,7 @@ public class ConsumeAMQP extends AbstractAMQPProcessor<AMQPConsumer> {
         .name("header.format")
         .displayName("Header Output Format")
         .description("Defines how to output headers from the received message")
-        .allowableValues(OutputHeaderFormat.getAllowedValues())
+        .allowableValues(OutputHeaderFormat.class)
         .defaultValue(OutputHeaderFormat.COMMA_SEPARATED_STRING)
         .required(true)
         .build();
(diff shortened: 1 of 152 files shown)


pvillard31:
Hey @EndzeitBegins - thanks for putting this together!

While I understand the multi steps approach, I feel like we should not even allow the option to opt-out. I looked at some of the cases where the problem is exhibited and in those cases it would not trigger a bug (ie. we do try to get the value of the property even if the dependency is not satisfied but we later check the dependency or if the returned value is null), it's more about having a better code structure.

I'm happy to collaborate with you to fix the places where we have the failure if you'd like to divide and conquer. Let me know.

exceptionfactory (commented):
I agree with @pvillard31 that the TestRunner should not have the option to opt out of the expected behavior.

Although I agree that there is value in general with making incremental changes, I think this is a case where it is better to make all of the changes together.

EndzeitBegins:
Alright. Then I'll continuously append commits addressing the different components to this PR instead. 

That way, changes should still be relatively easy to track while all work is kept under a single PR and may be squashed before merge if desired.
Sounds good to both of you @pvillard31 & @exceptionfactory?

exceptionfactory:
> Alright. Then I'll continuously append commits addressing the different components to this PR instead. In the meantime I'll mark the PR as draft.
> 
> That way, changes should still be relatively easy to track while all work is kept under a single PR and may be squashed before merge if desired. Sounds good to both of you @pvillard31 & @exceptionfactory?

That sounds like a good plan, thanks @EndzeitBegins!

EndzeitBegins:
I refactored all components where the added check failed and removed the opt-out functionality.

Most changes were straightforward, primarily involving conditional property checks or adjustments to the execution path. In rare instances, tests revealed that the processor relied on a specific property (or its default value) even when the dependency constraint was unmet. In these cases, I removed the constraint entirely.

I have rebased these changes onto the latest `main` branch to resolve merge conflicts. While system tests encountered issues yesterday due to changes merged into `main`, a subsequent rebase appears to have resolved these problems.

I'd appreciate it if you could take a look, @pvillard31 & @exceptionfactory.

exceptionfactory (approved):
Thanks for the extensive work on making the changes to align expected behavior @EndzeitBegins. The changes looks good.

Any additional feedback @pvillard31?

pvillard31 (approved):
Thanks for this work @EndzeitBegins - I'm a +1

PR #9858, opened 2025-04-08
NIFI-14450: Setting Max retry to static value to FetchBoxFileRepresentation

ncover21:
NIFI-14450:

Edge case has been occurring once in a while were Box API gets stuck and goes in a loop. The box service which generates the representation gets stuck and the default method sets the default max retries to Integer.MAX_VALUE
```
public void getRepresentationContent(String representationHint, String assetPath, OutputStream output) {
        this.getRepresentationContent(representationHint, assetPath, output, Integer.MAX_VALUE);
    }
```

- setting a static value for the retry count 

# Summary

[NIFI-14450](https://issues.apache.org/jira/browse/NIFI-14450)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/FetchBoxFileRepresentation.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/FetchBoxFileRepresentation.java
@@ -120,6 +120,13 @@ public class FetchBoxFileRepresentation extends AbstractProcessor implements Ver
             REL_REPRESENTATION_NOT_FOUND
     );
 
+    /*
+     * The maximum number of retries for polling the status of the generated representation.
+     * Each retry waits for 100 milliseconds before the next attempt, set in the Box SDK.
+     * Total wait time is 5 seconds.
+     */
+    private static final int MAX_RETRIES = 50;
+
     private volatile BoxAPIConnection boxAPIConnection;
 
     @Override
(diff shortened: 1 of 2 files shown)


ncover21:
@pvillard31 some advice might be helpful here. I've set the max retry count to 50, given that the sdk hardcodes its sleep between requests to 100 milliseconds. https://github.com/box/box-java-sdk/blob/6ee705e863e6bd93864c4b036f8aac9c13edce8e/src/main/java/com/box/sdk/BoxFile.java#L607 

Making it 5 seconds total. Do you think this is an ok number?

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/FetchBoxFileRepresentation.java
```diff
@@ -120,6 +120,13 @@ public class FetchBoxFileRepresentation extends AbstractProcessor implements Ver
             REL_REPRESENTATION_NOT_FOUND
     );
 
+    /*
+     * The maximum number of retries for polling the status of the generated representation.
+     * Each retry waits for 100 milliseconds before the next attempt, set in the Box SDK.
+     * Total wait time is 5 seconds.
+     */
+    public static final int MAX_RETRIES = 50;
```
awelless, line 128: Is there any particular reason this is public?

awelless (approved):

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/FetchBoxFileRepresentation.java
sfc-gh-ncover: Oops, nope. Fixed thanks

ncover21: Oops nope, fixed thanks

pvillard31:
Yeah, I believe this is a fair value. If it still fails after the max retry number, are we properly handling the failure? is it throwing an exception?

ncover21:
> Yeah, I believe this is a fair value. If it still fails after the max retry number, are we properly handling the failure? is it throwing an exception?

Yes, boxApiException is thrown and routed based on status code like normal.

pvillard31:
Cool, assuming all checks are green, I'll +1 and merge

PR #9863, opened 2025-04-09
NIFI-14435 - FlowComparator should offer deep recursive comparison

pvillard31:
# Summary

[NIFI-14435](https://issues.apache.org/jira/browse/NIFI-14435) - FlowComparator should offer deep recursive comparison

Follow up of reverted PR #9848.

Instead of instantiating an empty versioned process group which causes a class cast exception, we just call the compare methods for the components as well as for the configuration of the PG (only in the case of a newly added PG) to check if the defaults have changed.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -25,6 +25,7 @@ on:
       - 'nifi-framework-api/**'
       - 'nifi-framework-bundle/**'
       - 'nifi-extension-bundles/nifi-py4j-extension-bundle/**'
+      - 'nifi-registry/nifi-registry-core/nifi-registry-flow-diff/**'
       - 'nifi-stateless/**'
       - 'nifi-mock/**'
   pull_request:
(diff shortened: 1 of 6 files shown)


exceptionfactory (commented):
Thanks for making the adjustments @pvillard31. The latest version looks good in general, but I pushed a change to trigger the system tests workflow for changes to the `nifi-registry-flow-diff` module to ensure current and future updates work at the system level.

exceptionfactory (changes requested):
It looks like the following system test is failing based on unexpected differences.

```
JoinClusterWithDifferentFlowIT.testStartupWithDifferentFlow » Timeout testStartupWithDifferentFlow() timed out after 5 minutes
```

pvillard31:
Thanks @exceptionfactory - I have pushed a commit to address the system test failure

PR #9868, opened 2025-04-10
NIFI-14391 Support to initiate administrative access for group of users

EndzeitBegins:
# Summary

[NIFI-14391](https://issues.apache.org/jira/browse/NIFI-14391)

Adds support for defining `Initial Group Identity 123` and `Initial Admin Group` similar to the existing `Initial User Identity 123` and `Initial Admin Identity`.
This way a group of initial admins can be defined when using OIDC instead of relying on a specific user or a service account.

When using the docker image, `INITIAL_ADMIN_GROUP` can be used besides or instead of `INITIAL_ADMIN_IDENTITY`.

Tested functionality using NiFi docker image and local instance of Keycloak. I couldn't get nested claim structures to work, e.g. defining `NIFI_SECURITY_USER_OIDC_CLAIM_GROUPS=nested.namespace.groups`. However, that should be out-of-scope of this issue.  

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


EndzeitBegins:
As we've [discussed this on Slack](https://apachenifi.slack.com/archives/C0L9VCD47/p1742923692051819) a few weeks ago, when you get a chance, I'd appreciate a review by you @exceptionfactory.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
```diff
@@ -168,10 +171,21 @@ public void onConfigured(AuthorizerConfigurationContext configurationContext) th
...
             // extract any node identities
             initialUserIdentities = new HashSet<>();
+            initialGroupIdentities = new HashSet<>();
             for (Map.Entry<String, String> entry : configurationContext.getProperties().entrySet()) {
-                Matcher matcher = INITIAL_USER_IDENTITY_PATTERN.matcher(entry.getKey());
-                if (matcher.matches() && !StringUtils.isBlank(entry.getValue())) {
-                    initialUserIdentities.add(IdentityMappingUtil.mapIdentity(entry.getValue(), identityMappings));
+                Matcher userMatcher = INITIAL_USER_IDENTITY_PATTERN.matcher(entry.getKey());
+                if (userMatcher.matches() && !StringUtils.isBlank(entry.getValue())) {
+                    if (StringUtils.isNotBlank(entry.getValue())) {
```
mosermw: Since the first if() uses && !StringUtils.isBlank(), the second if( StringUtils.isNotBlank() ) is unnecessary and can be removed.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
```diff
@@ -168,10 +171,21 @@ public void onConfigured(AuthorizerConfigurationContext configurationContext) th
...
+                        initialUserIdentities.add(IdentityMappingUtil.mapIdentity(entry.getValue(), identityMappings));
+                    }
+                    continue;
+                }
+
+                Matcher groupMatcher = INITIAL_GROUP_IDENTITY_PATTERN.matcher(entry.getKey());
+                if (groupMatcher.matches() && !StringUtils.isBlank(entry.getValue())) {
+                    if (StringUtils.isNotBlank(entry.getValue())) {
```
mosermw: Same ... second if (StringUtils.isNotBlank() ) can be removed.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/test/java/org/apache/nifi/authorization/FileUserGroupProviderTest.java
```diff
@@ -196,6 +229,27 @@ public void testOnConfiguredWhenTenantsExistAndInitialUsersProvided() throws Exc
...
+        when(configurationContext.getProperty(eq(FileUserGroupProvider.PROP_INITIAL_GROUP_IDENTITY_PREFIX + "1")))
+                .thenReturn(new StandardPropertyValue(adminGroupIdentity, null, ParameterLookup.EMPTY));
+        when(configurationContext.getProperty(eq(FileUserGroupProvider.PROP_INITIAL_GROUP_IDENTITY_PREFIX + "2")))
+                .thenReturn(new StandardPropertyValue(otherGroupIdentity, null, ParameterLookup.EMPTY));
+
+        writeFile(primaryTenants, SIMPLE_TENANTS_BY_USER);
+        userGroupProvider.onConfigured(configurationContext);
+
+        final Set<User> users = userGroupProvider.getUsers();
+        assertEquals(2, users.size());
```
mosermw: This is checking the number of Users but instead it should be checking the number of Groups.  In this case, since the tenants is not empty, verify that 0 groups are created.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/test/java/org/apache/nifi/authorization/FileAccessPolicyProviderTest.java
```diff
@@ -385,6 +414,121 @@ public void testOnConfiguredWhenInitialAdminProvidedAndFlowIsNull() throws Excep
...
+
+        assertFalse(foundRootGroupPolicy);
+    }
+
+    @Test
+    public void testOnConfiguredWhenInitialAdminGroupProvidedAndFlowIsNull() throws Exception {
+        // setup NiFi properties to return a file that does not exist
+        properties = mock(NiFiProperties.class);
+        when(properties.getRestoreDirectory()).thenReturn(restoreAuthorizations.getParentFile());
+        when(properties.getFlowConfigurationFile()).thenReturn(new File("src/test/resources/does-not-exist.json.gz"));
```
mosermw: Looks like testOnConfiguredWhenInitialAdminGroupProvidedAndNoFlowExists() and testOnConfiguredWhenInitialAdminGroupProvidedAndFlowIsNull() are exactly the same.  I believe this line should be

when(properties.getFlowConfigurationFile()).thenReturn(null);

Would you also fix the testOnConfiguredWhenInitialAdminProvidedAndFlowIsNull() method where this same change is needed?  I believe this was a mistake when removing support for flow.xml.gz

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
```diff
@@ -737,6 +758,35 @@ private void createUser(final Tenants tenants, final String userIdentity) {
...
+     *
+     * @param tenants the Tenants reference
+     * @param groupName the name of the group to look for
+     */
+    private void createGroup(final Tenants tenants, final String groupName) {
+        if (StringUtils.isBlank(groupName)) {
+            return;
+        }
+
+        // TODO stream api?
```
mosermw: Since createGroup() and createUser() are so similar, I think you can remove this TODO and keep what you have here.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
```diff
@@ -737,6 +758,35 @@ private void createUser(final Tenants tenants, final String userIdentity) {
...
+        org.apache.nifi.authorization.file.tenants.generated.Group foundGroup = null;
+        for (org.apache.nifi.authorization.file.tenants.generated.Group group : tenants.getGroups().getGroup()) {
+            if (group.getName().equals(groupName)) {
+                foundGroup = group;
+                break;
+            }
+        }
+
+        if (foundGroup == null) {
+            final Group newGroup = new Group.Builder().identifierGenerateFromSeed(groupName).name(groupName).build();
```
mosermw: Looking at this line and comparing it to the same line in the createUser() method above, I think we should do a small refactor on the createUser() method.

Would you modify createUser() to use User.Builder().identifierGenerateFromSeed()?  I believe this would allow us to remove the IdentifierUtil.java file entirely because it would not be used anymore.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
```diff
@@ -670,6 +684,7 @@ private synchronized void load() throws JAXBException, IllegalStateException {
 
         if (emptyTenants) {
             populateInitialUsers(tenants);
+            populateInitialGroups(tenants);
```
mosermw, line 684: This is more of an idea than a comment.  How would you feel about adding all of the initial users as a member of each of the initial groups?  As I was testing this, I found myself wanting to add myself (the Initial Admin) to my Initial Group, because managing policies at the group level has always proven to be easier.

mosermw (commented):
I tested this and it works well @EndzeitBegins.  Nice job!  I just had a few comments and questions to get your thoughts.

EndzeitBegins:
Thank you for the review & feedback @mosermw. As I've been on vacation I haven't been able to take a detailed look at it yet but will try to revisit the PR in the next days.

mosermw:
@EndzeitBegins Take your time, because I'm going on vacation now and wouldn't be able to merge immediately anyway.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
EndzeitBegins: I adjust the code for users and was able to remove `IdentifierUtil` entirely indeed, good catch!

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
EndzeitBegins: I revisited the code and refactored both implementations a little.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/test/java/org/apache/nifi/authorization/FileUserGroupProviderTest.java
EndzeitBegins: Good catch, you're absolutely right. I adjusted the test.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
EndzeitBegins: I see where you're going with this, however I'm not sure if that's a good default.
Say I want initial users `Alice`, `Bob`, and `Cleo` as well as initial groups `Users` and `Admins`. 
I do not think it makes a lot of sense to add the users to both groups in this case. 

In regards to the `Initial Admin Identity` (User)  and `Initial Admin Group`  however, one could argue that it would be a good idea to add the initial admin user to the initial admin group, in case both are defined. What do you think @mosermw?

EndzeitBegins:
Thank you for the review @mosermw. 

Since the PR has been open for some time, I've decided to rebase onto the latest `main`. I addressed all your comments and  added a comment in regards to adding initial users to initial groups but did not change behaviour yet.

I would appreciate another review once you have some time to spare. I hope you have / had a relaxing vacation.

nifi-framework-bundle/nifi-framework/nifi-file-authorizer/src/main/java/org/apache/nifi/authorization/FileUserGroupProvider.java
mosermw: Thanks for considering this @EndzeitBegins. I remember being unsure if there was an obviously correct approach to this.  Looking back now, I agree that adding all initial users to all initial groups is going too far.

Adding the `Initial Admin Identity` user to the `Initial Admin Group` does feel like an obvious thing to do, though.  If you can make that happen, I think it makes a lot of sense.  If you make this change, consider mentioning it in the documentation.

EndzeitBegins:
I've pushed another commit that adds the initial admin user to the initial admin group instead of granting policies to the user directly, if both are specified and the user group providers supports updating groups.

I'd appreciate you taking another look @mosermw.
Maybe we can get this to land on `main` before @pvillard31 kicks off the upcoming release.

mosermw:
Code look good to me, and the tests I ran worked as expected.  +1

PR #9869, opened 2025-04-10
NIFI-14458 - Added run command to nifi.cmd

philci52:
# Summary

[NIFI-14458](https://issues.apache.org/jira/browse/NIFI-14458)
Added run command to nifi.cmd to allow NiFi to run in the foreground and be used with Apache ProcRun

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-docs/src/main/asciidoc/administration-guide.adoc
+++ b/nifi-docs/src/main/asciidoc/administration-guide.adoc
@@ -58,7 +58,9 @@ NOTE: Under sustained and extremely high throughput the CodeCache settings may n
 *** `start`: starts NiFi in the background
 *** `stop`: stops NiFi that is running in the background
 *** `status`: provides the current status of NiFi
+*** `run`: runs NiFi in the foreground and waits for a Ctrl-C to initiate shutdown of NiFi.
 
+NOTE: When using Ctrl-C to initiate shutdown of NiFi on Windows, it is recommended to answer `N` to the `Terminate batch job (Y/N)?` prompt to allow the nifi.cmd to exit cleanly and restore the current working directory.  Alternatively, a `.\nifi.cmd stop` can be run from another console.
 
 When NiFi first starts up, the following files and directories are created:
 
--- a/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/bin/nifi.cmd
+++ b/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/bin/nifi.cmd
@@ -55,6 +55,9 @@ if %RUN_COMMAND% == "set-single-user-credentials" (
 ) else if %RUN_COMMAND% == "start" (
   rem Start bootstrap process in new minimized window
   call start /MIN "Apache NiFi" "%JAVA_EXE%" %JAVA_MEMORY% %JAVA_PARAMS% org.apache.nifi.bootstrap.BootstrapProcess %RUN_COMMAND%
+) else if %RUN_COMMAND% == "run" (
+  rem Start bootstrap process in the foreground
+  call "%JAVA_EXE%" %JAVA_MEMORY% %JAVA_PARAMS% org.apache.nifi.bootstrap.BootstrapProcess start
 ) else (
   call "%JAVA_EXE%" %JAVA_MEMORY% %JAVA_PARAMS% org.apache.nifi.bootstrap.BootstrapProcess %RUN_COMMAND%
 )


exceptionfactory (changes requested):
Thanks for proposing this change @philci52.

However, calling `BootstrapProcess` with the `start` command is different than the behavior of the `nifi.sh` script. From that perspective, the behavior should be the same across both platforms.

If you aren't in a position to implement those details, I may be able to take a look at that approach.

philci52:
I guess I don't understand the difference.  Is it in the Java code?  Will I still be able to run "nifi stop" to stop the application if run that way?

exceptionfactory:
Here's the section of `nifi.sh` that handles the `run` command:

https://github.com/apache/nifi/blob/main/nifi-framework-bundle/nifi-framework/nifi-resources/src/main/resources/bin/nifi.sh#L272

As shown, it calls the BootstrapProcess class with `get-run-command`, then executes the returned command.

Yes, `nifi.cmd stop` will either way.

philci52:
Looks like there will be more issues with paths that have spaces (probably an issue on linux too).  https://github.com/apache/nifi/blob/main/nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/command/GetRunCommandBootstrapCommand.java#L78

exceptionfactory:
Thanks for the thoughtful evaluation of options @philci52.

I think there are two different elements to consider.

The first aspect is bootstrap.conf. Property values using `java.arg` contain user-controller values, and it seems acceptable to depend the user to introduce quotes when necessary, such as the following:

```
java.arg.100=-Dsystem.property.name=escaped\\ string
```

The second aspect is the standard arguments to the command, which include the following from `StandardProcessBuilderProvider`:

1. Class Path argument containing file paths
2. Log Directory system property containing log directory path
3. Application Properties system property containing file path
4. Management Server Address
5. Additional arguments from bootstrap.conf
6. Application class name to run

Arguments 1, 2, and 3 contain paths that may need to be quoted, so for those cases, quoting the value of the argument seems like the way to go. Arguments 4 and 6 do not need to be quoted. The additional arguments in 5 can be addressed as described above.

With that background, it seems like making adjustments to `StandardProcessBuilderProvider` could provide a workable solution that should apply to all platforms.

philci52:
@exceptionfactory Thank you for the quick feedback.  There is one other path that may need to be quoted and that is the (7) path to java itself.

With regards to using StandardProcessBuilderProvider, that class will have to know if its creating arguments for the shell/console or for using ProcessBuilder.  When using ProcessBuilder, passing `"argument"`, the child process will actually get `"argument"` as opposed to `argument` when passed from the shell/console.  Is that acceptable?

philci52:
Here is a comment from some of our internal code regarding the use of process builder and windows/linux differences.  This application uses an XML file for configuration, but I figured it would be relevant to this conversation

```
Java Application Operating system Differences:
MS Windows:
When calling a java process on MS Windows from OURAPP, java will automatically
expand glob characters and Java uses double quotes to suppress glob expansion.  Glob characters
are []{}*?.  This can lead to some unexpected behavior when calling java processes from OURAPP.
To suppress this expansion when calling a java application from within OURAPP, you must surround the
argument with double quotes.  For example <arg>&quot;test_*&quot;</arg> or
<arg><![CDATA["test_*"]]></arg>.  The application will then get test_* as the parameter.
If you need to place quotes in argument itself, you must surround the argument with double quotes
and the use two double quotes to represent one double quote.  For example, if you want the string 
P"J"D, then you would use <arg><![CDATA["P""J""D"]]></arg>
Linux/Unix:
When calling a java process from within from OURAPP on Linux, the java application will get
the exact value that you pass it, except for -cp argument, which java will expand.
If you want glob expansion, you must wrap the java application with a call to the shell.

```

philci52:
@exceptionfactory  Any updates on how you want to handle the issues here.

exceptionfactory:
Thanks for checking @philci52, I evaluated some options last week using the existing run command, but didn't find a satisfactory solution.

At the moment, evaluating the impact of selective quoting in the `StandardProcessBuilderProvider`, and updating the Windows `nifi.cmd` to use `get-run-command` seems like the best way forward.

Are you in a position to implement and evaluate the approach for Windows, along with the impact on Linux platforms? If not, I may be able to take a closer look.

philci52:
I can work on this, but it will take me a few weeks to get back to it.

exceptionfactory:
Thanks again for the discussion on this issue @philci52. Given the direction and time, I am closing this for now. I might be able to revisit it, and glad to follow up on the Jira issue if you have further questions about implementation.

philci52:
@exceptionfactory Sorry, I have just not had the time to get back this.  I'll open a new pull request when/if I do.

For now, others can use the scripts I've uploaded to the ticket and in this PR to get it working with ProcRun although it may not be the desired solution.

PR #9872, opened 2025-04-11
NIFI-14469 - Support custom fields in AccessToken

pvillard31:
# Summary

[NIFI-14469](https://issues.apache.org/jira/browse/NIFI-14469) - Support custom fields in AccessToken

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/pom.xml
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/pom.xml
@@ -22,4 +22,10 @@
     </parent>
     <artifactId>nifi-oauth2-provider-api</artifactId>
     <packaging>jar</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+    </dependencies>
 </project>
--- a/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/AccessToken.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/AccessToken.java
@@ -17,7 +17,11 @@
 
 package org.apache.nifi.oauth2;
 
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+
 import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
 
 public class AccessToken {
     private String accessToken;
@@ -27,6 +31,9 @@ public class AccessToken {
     private String scopes;
 
     private final Instant fetchTime;
+
+    // Map to store custom fields
+    private final Map<String, String> customFields = new HashMap<>();
 
     public AccessToken() {
         this.fetchTime = Instant.now();
@@ -92,5 +99,14 @@ public class AccessToken {
 
     Instant now() {
         return Instant.now();
+    }
+
+    @JsonAnySetter
+    public void setCustomField(String key, String value) {
+        customFields.put(key, value);
+    }
+
+    public Map<String, String> getCustomFields() {
+        return customFields;
     }
 }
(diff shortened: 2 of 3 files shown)


nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/pom.xml
```diff
@@ -22,4 +22,10 @@
     </parent>
     <artifactId>nifi-oauth2-provider-api</artifactId>
     <packaging>jar</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
```
exceptionfactory, line 28: Introducing this dependency at the API level seems problematic. At minimum, this should be limited to the jackson-annotation library. However, I recommend an alternative approach that handles the custom fields through an ObjectMapper configuration in the implementation class.

exceptionfactory (changes requested):
Thanks @pvillard31, the concept sounds good, but it looks like the implementation approach should be adjusted. I can provide more details on options if needed, but it should be possible to configure the Jackson implementation to handle the custom fields, without necessarily annotating the AccessToken class itself.

exceptionfactory (commented):
Thanks for the updates @pvillard31.

Have you evaluated using the [JsonView](https://www.baeldung.com/jackson-json-view-annotation) annotation on a subclass of `AccessToken` within the `StandardOauth2AccessTokenProvider`? Having an internal subclass of `AccessToken`, with an interface method annotated with `JsonAnySetter` seems like an approach that would work without having to implement a full Deserializer.

pvillard31:
Thanks @exceptionfactory - I was not aware of the JsonView annotation. I just pushed a commit to leverage it. Let me know if this is what you had in mind.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/AccessToken.java
```diff
@@ -41,6 +45,16 @@ public AccessToken(String accessToken, String refreshToken, String tokenType, lo
         this.scope = scope;
     }
 
+    public AccessToken(String accessToken, String refreshToken, String tokenType, long expiresIn, String scope, Map<String, String> customFields) {
```
exceptionfactory, line 48: Is this constructor required?

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-api/src/main/java/org/apache/nifi/oauth2/AccessTokenViews.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nifi.oauth2;
+
+public class AccessTokenViews {
```
exceptionfactory, line 20: This should be placed in the implementation module, instead of the API module.

exceptionfactory (changes requested):
Thanks for the updates @pvillard31, this looks close to what I had in mind. I noted a couple adjustments to keep the mapping view internal to the implementation module, otherwise, this looks close to completion.

pvillard31:
Addressed your comments - this concept of JsonView is new to me, appreciate the guidance!

exceptionfactory (commented):
Thanks for the updates @pvillard31, after taking another look at available options, the mix-in strategy appears to be the most concise solution, enabling a simple interface with matching method signature to apply the `JsonAnySetter` annotation.

I pushed that change and also renamed the Map to `additionalParameters`, which follows the convention of Spring Security's `OAuth2AccessTokenReponse` class. I also changed the Map value type from `String` to `Object`, since it could be something else, such as a number.

If these changes look good, I circle back to complete the review.

pvillard31:
Changes LGTM, thanks @exceptionfactory

exceptionfactory (approved):
Thanks for the concurrence @pvillard31! +1 merging

PR #9874, opened 2025-04-11
NIFI-14426: Add support for HSSF format in ExcelReader processor

zhtk:
# Summary

[NIFI-14426](https://issues.apache.org/jira/browse/NIFI-14426): Add support for HSSF format in ExcelReader processor

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
@@ -53,13 +53,13 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
-@Tags({"excel", "spreadsheet", "xlsx", "parse", "record", "row", "reader", "values", "cell"})
+@Tags({"excel", "spreadsheet", "xls", "xlsx", "parse", "record", "row", "reader", "values", "cell"})
 @CapabilityDescription("Parses a Microsoft Excel document returning each row in each sheet as a separate record. "
         + "This reader allows for inferring a schema from all the required sheets "
         + "or providing an explicit schema for interpreting the values."
         + "See Controller Service's Usage for further documentation. "
-        + "This reader is currently only capable of processing .xlsx "
-        + "(XSSF 2007 OOXML file format) Excel documents and not older .xls (HSSF '97(-2007) file format) documents.")
+        + "This reader is capable of processing both password and non password protected .xlsx (XSSF 2007 OOXML file format) "
+        + "and older .xls (HSSF '97(-2007) file format) Excel documents.")
 public class ExcelReader extends SchemaRegistryService implements RecordReaderFactory {
 
     public static final PropertyDescriptor REQUIRED_SHEETS = new PropertyDescriptor
@@ -83,6 +83,15 @@ public class ExcelReader extends SchemaRegistryService implements RecordReaderFa
             .defaultValue("1")
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .addValidator(StandardValidators.POSITIVE_INTEGER_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor INPUT_FILE_TYPE = new PropertyDescriptor
+            .Builder().name("Input File Type")
+            .displayName("Input File Type")
+            .description("Specifies type of Excel input file.")
+            .required(true)
+            .allowableValues(InputFileType.class)
+            .defaultValue(InputFileType.XLSX)
             .build();
 
     public static final PropertyDescriptor PROTECTION_TYPE = new PropertyDescriptor
@@ -128,6 +137,7 @@ public class ExcelReader extends SchemaRegistryService implements RecordReaderFa
         final List<String> requiredSheets = getRequiredSheets(variables);
         final int firstRow = getStartingRow(variables);
         final String password = configurationContext.getProperty(PASSWORD).getValue();
+        final InputFileType inputFileType = configurationContext.getProperty(INPUT_FILE_TYPE).asAllowableValue(InputFileType.class);
         final ExcelRecordReaderConfiguration configuration = new ExcelRecordReaderConfiguration.Builder()
                 .withDateFormat(dateFormat)
                 .withRequiredSheets(requiredSheets)
@@ -136,6 +146,7 @@ public class ExcelReader extends SchemaRegistryService implements RecordReaderFa
                 .withTimeFormat(timeFormat)
                 .withTimestampFormat(timestampFormat)
                 .withPassword(password)
+                .withInputFileType(inputFileType)
                 .build();
 
         return new ExcelRecordReader(configuration, in, logger);
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/InputFileType.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/InputFileType.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.excel;
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum InputFileType implements DescribedValue {
+    XLS("XLS", "XLS 1997-2007 file format"),
+    XLSX("XLSX", "XLSX 2007-present OOXML file format");
+
+    InputFileType(String displayName, String description) {
+        this.displayName = displayName;
+        this.description = description;
+    }
+
+    private final String displayName;
+    private final String description;
+
+    @Override
+    public String getValue() {
+        return name();
+    }
+
+    @Override
+    public String getDisplayName() {
+        return displayName;
+    }
+
+    @Override
+    public String getDescription() {
+        return description;
+    }
+}
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/InputFileType.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/InputFileType.java
@@ -19,15 +19,13 @@ package org.apache.nifi.excel;
 import org.apache.nifi.components.DescribedValue;
 
 public enum InputFileType implements DescribedValue {
-    XLS("XLS", "XLS 1997-2007 file format"),
-    XLSX("XLSX", "XLSX 2007-present OOXML file format");
+    XLS("XLS 1997-2007 file format"),
+    XLSX("XLSX 2007-present OOXML file format");
 
-    InputFileType(String displayName, String description) {
-        this.displayName = displayName;
+    InputFileType(String description) {
         this.description = description;
     }
 
-    private final String displayName;
     private final String description;
 
     @Override
@@ -37,7 +35,7 @@ public enum InputFileType implements DescribedValue {
 
     @Override
     public String getDisplayName() {
-        return displayName;
+        return name();
     }
 
     @Override
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
@@ -19,6 +19,8 @@ package org.apache.nifi.excel;
 import com.github.pjfanning.xlsx.StreamingReader;
 import org.apache.nifi.logging.ComponentLog;
 import org.apache.nifi.processor.exception.ProcessException;
+import org.apache.poi.hssf.record.crypto.Biff8EncryptionKey;
+import org.apache.poi.hssf.usermodel.HSSFWorkbook;
 import org.apache.poi.ss.usermodel.Row;
 import org.apache.poi.ss.usermodel.Sheet;
 import org.apache.poi.ss.usermodel.Workbook;
@@ -42,13 +44,26 @@ class RowIterator implements Iterator<Row>, Closeable {
     private Row currentRow;
 
     RowIterator(final InputStream in, final ExcelRecordReaderConfiguration configuration, final ComponentLog logger) {
-        this.workbook = StreamingReader.builder()
-                .rowCacheSize(100)
-                .bufferSize(4096)
-                .password(configuration.getPassword())
-                .setAvoidTempFiles(configuration.isAvoidTempFiles())
-                .setReadSharedFormulas(true) // NOTE: If not set to true, then data with shared formulas fail.
-                .open(in);
+        if (configuration.getInputFileType() == InputFileType.XLSX) {
+            this.workbook = StreamingReader.builder()
+                    .rowCacheSize(100)
+                    .bufferSize(4096)
+                    .password(configuration.getPassword())
+                    .setAvoidTempFiles(configuration.isAvoidTempFiles())
+                    .setReadSharedFormulas(true) // NOTE: If not set to true, then data with shared formulas fail.
+                    .open(in);
+        } else {
+            // Providing the password to the HSSFWorkbook is done by setting a thread variable managed by
+            // Biff8EncryptionKey. After the workbook is created, the thread variable can be cleared.
+            Biff8EncryptionKey.setCurrentUserPassword(configuration.getPassword());
+            try {
+                this.workbook = new HSSFWorkbook(in);
+            } catch (final IOException e) {
+                throw new ProcessException("Failed to open XLS file", e);
+            } finally {
+                Biff8EncryptionKey.setCurrentUserPassword(null);
+            }
+        }
 
         final List<String> requiredSheets = configuration.getRequiredSheets();
         if (requiredSheets == null || requiredSheets.isEmpty()) {
--- a/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
+++ b/nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
@@ -16,7 +16,6 @@
  */
 package org.apache.nifi.excel;
 
-import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.nifi.logging.ComponentLog;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.serialization.MalformedRecordException;
@@ -25,7 +24,8 @@ import org.apache.nifi.serialization.record.Record;
 import org.apache.nifi.serialization.record.RecordField;
 import org.apache.nifi.serialization.record.RecordFieldType;
 import org.apache.nifi.serialization.record.RecordSchema;
-import com.github.pjfanning.xlsx.exceptions.ReadException;
+import org.apache.poi.hssf.record.crypto.Biff8EncryptionKey;
+import org.apache.poi.hssf.usermodel.HSSFWorkbook;
 import org.apache.poi.openxml4j.opc.OPCPackage;
 import org.apache.poi.poifs.crypt.EncryptionInfo;
 import org.apache.poi.poifs.crypt.EncryptionMode;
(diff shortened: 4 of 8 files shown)


zhtk:
Yes, rebased @exceptionfactory

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -42,13 +51,26 @@ class RowIterator implements Iterator<Row>, Closeable {
...
+                    .rowCacheSize(100)
+                    .bufferSize(4096)
+                    .password(configuration.getPassword())
+                    .setAvoidTempFiles(configuration.isAvoidTempFiles())
+                    .setReadSharedFormulas(true) // NOTE: If not set to true, then data with shared formulas fail.
+                    .open(in);
+        } else {
+            // Providing the password to the HSSFWorkbook is done by setting a thread variable managed by
+            // Biff8EncryptionKey. After the workbook is created, the thread variable can be cleared.
+            Biff8EncryptionKey.setCurrentUserPassword(configuration.getPassword());
```
exceptionfactory, line 65: This is problematic for multi-threaded use of ExcelReader, which is very possible. Under the circumstances, it seems best to avoid attempting to support encrypted XLS files right now. This will likely throw exceptions when `HSSFWorkbook` attempts to read, but that is acceptable under the circumstances, as it is a limitation of the library.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -129,4 +151,37 @@ private boolean hasExhaustedRows() {
         }
         return exhausted;
     }
+
+    private boolean isXSSFExcelFile(InputStream in, final String password) {
```
exceptionfactory, line 155: Minor naming recommend:
```suggestion
    private boolean isXmlFormat(final InputStream in, final String password) {
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -129,4 +151,37 @@ private boolean hasExhaustedRows() {
...
+        }
+        if (fileType != FileMagic.OLE2) {
+            // Not an Excel file
+            return false;
+        }
+        if (password == null) {
+            // If there is no password on the OLE2 file, then it's an XLS file. Otherwise, the type must be further
+            // investigated.
+            return false;
+        }
```
exceptionfactory, line 175: It would be helpful to refactor this method to use a single return.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -129,4 +151,37 @@ private boolean hasExhaustedRows() {
...
+
+        DirectoryNode root;
+        try {
+            root = new POIFSFileSystem(new NonCloseableInputStream(in)).getRoot();
+            in.reset();
```
exceptionfactory, line 180: It looks like the `reset()` call should be moved to a `finally` block in case of exceptions.

exceptionfactory (changes requested):
Thanks for working on this @zhtk. The general approach looks straightforward. Although there are potential memory implications for the older XLS format, this seems to be the best implementation strategy for now.

I noted one particular issue related to password handling for XLS files, and the lack of thread safety. As mentioned, I think the best option for now is to not support encrypted XLS, and let the read operation fail.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -129,4 +151,37 @@ private boolean hasExhaustedRows() {
...
+        }
+        if (password == null) {
+            // If there is no password on the OLE2 file, then it's an XLS file. Otherwise, the type must be further
+            // investigated.
+            return false;
+        }
+
+        DirectoryNode root;
+        try {
+            root = new POIFSFileSystem(new NonCloseableInputStream(in)).getRoot();
```
dan-s1, line 179: In addition I think the creation of `POIFSFileSystem` and the assignment to `root` should be done on separate lines. Also the return statement should be all done in the try block as if the instantiation of `POIFSFileSystem` throws an IOException the return statement is never reached.

```suggestion
        DirectoryNode root;
        try {
           final POIFSFileSystem poifsFileSystem = new POIFSFileSystem(new NonCloseableInputStream(in)).getRoot();
           final DirectoryNode root = poifsFileSystem.getRoot() ;
           return root.hasEntryCaseInsensitive(DEFAULT_POIFS_ENTRY) || root.hasEntryCaseInsensitive(OOXML_PACKAGE);
           }
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: or perhaps
```suggestion
    private boolean isXSSF(InputStream in, final String password) {
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: @exceptionfactory Just want to clarify since `reset` can also throw an `IOException` the `finally` would have to also have a try catch block.

exceptionfactory: Yes, that is correct.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
exceptionfactory: Although making the Input File Type a configurable property is an option, there are positives and negatives to consider requiring the configuration. On the one hand, IdentifyMimeType could be used with a RouteOnAttribute Processor to determine the appropriate routing. On the other hand, that increases the potential flow design complexity when attempting to support both XLS and XLSX.

I agree that introducing a new configuration property would simplify this logic and remove the need for File Type detection.

Given the mismatch in functionality, with and without support for password-protection, there does seem to be some value in requiring one type or the other.

It is also worth noting that XLS has not been supported previously, and that it has different memory usage characteristics as previously discussed.

With that in mind, I would be open to adding a new property named `Input File Type` with `XLSX` and `XLS` values, where `XLSX` would be the default type. The Property could support Expression Language, which is resolvable from the Map of variables passed to `createRecordReader()`.

dan-s1: @zhtk Are you clear on the changes needed?

exceptionfactory: @zhtk The thread safety issue is that calling a static method applies to all instances of the Reader. That means if there are multiple instances of the Excel Reader configured, with different passwords, different threads could step on each other. I recommend taking a closer look at the POI library to see if there are any other options. If this is the only approach supported in the library itself, then it would require more significant custom code to implement a solution that works in the context of NiFi.

zhtk: @exceptionfactory, okay, the method is static, and the variable to which the method writes is static, but it's instance of `ThreadLocal<String>`. According to my understanding, any reads/writes to this variable will be only visible by thread accessing the variable.

exceptionfactory: Thanks for pointing to the source implementation of `Biff8EncryptionKey`, I see the use of `ThreadLocal` and the associated comments. That makes a significant difference, and alleviates the concerns of setting a shared variable.

With that clarification, I withdraw my earlier concerns. I would simply add a comment indicating that Biff8EncryptionKey stores the value in a ThreadLocal variable, and proceed.

zhtk:
After reading whole discussion I decided to remove type detection and to add `Input File Type` property. There are 2 reasons for that:
1. In case of passworded Excel file, to determine file type the whole Flow File stream must be read. At the moment the stream is resettable only up to 1024 * 1024 mark (i.e. in `createRecordReader` there is a call `in.mark(1024 * 1024);`). Reseting stream on files larger than 1 MB could be problematic. Working around this problem would require contributions to both POI and StreamingReader libraries.
2. Performance characteristics of reading XLS files is different than XLSX.

dan-s1:
@zhtk From message above it seems there are conflicts you must resolve

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
```diff
@@ -53,13 +53,13 @@
...
-@Tags({"excel", "spreadsheet", "xlsx", "parse", "record", "row", "reader", "values", "cell"})
+@Tags({"excel", "spreadsheet", "xls", "xlsx", "parse", "record", "row", "reader", "values", "cell"})
 @CapabilityDescription("Parses a Microsoft Excel document returning each row in each sheet as a separate record. "
         + "This reader allows for inferring a schema from all the required sheets "
         + "or providing an explicit schema for interpreting the values."
         + "See Controller Service's Usage for further documentation. "
-        + "This reader is currently only capable of processing .xlsx "
-        + "(XSSF 2007 OOXML file format) Excel documents and not older .xls (HSSF '97(-2007) file format) documents.")
+        + "This reader is capable of processing .xlsx (XSSF 2007 OOXML file format) "
+        + "Excel documents and older .xls (HSSF '97(-2007) file format) documents.")
```
dan-s1, line 62:
```suggestion
        + "This reader is capable of processing both password and non password protected .xlsx (XSSF 2007 OOXML file format) "
        +  " and older .xls (HSSF '97(-2007) file format) Excel documents.")
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/InputFileType.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.excel;
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum InputFileType implements DescribedValue {
+    XLS("XLS", "An XLS file"),
+    XLSX("XLSX", "An XLSX file");
```
dan-s1, line 23:
```suggestion
    XLS("XLS", "HSSF '97(-2007) file format"),
    XLSX("XLSX", ""XSSF 2007 OOXML file format"");
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
```diff
@@ -42,13 +44,26 @@ class RowIterator implements Iterator<Row>, Closeable {
...
 
     RowIterator(final InputStream in, final ExcelRecordReaderConfiguration configuration, final ComponentLog logger) {
-        this.workbook = StreamingReader.builder()
-                .rowCacheSize(100)
-                .bufferSize(4096)
-                .password(configuration.getPassword())
-                .setAvoidTempFiles(configuration.isAvoidTempFiles())
-                .setReadSharedFormulas(true) // NOTE: If not set to true, then data with shared formulas fail.
-                .open(in);
+        if (configuration.getInputFileType() == InputFileType.XLSX) {
```
dan-s1, line 47: This comment is regarding the whole if else statements between lines 47-66. Instead of an if else, an assignment with a switch statement would be clearer to indicate what code is needed for each Excel type. I was thinking along the line of this:

```
this.workbook = switch(configuration.getInputFileType()) {
            case XLSX -> StreamingReader.builder()
                    .rowCacheSize(100)
                    .bufferSize(4096)
                    .password(configuration.getPassword())
                    .setAvoidTempFiles(configuration.isAvoidTempFiles())
                    .setReadSharedFormulas(true) // NOTE: If not set to true, then data with shared formulas fail.
                    .open(in);
            case XLS -> {
                // Providing the password to the HSSFWorkbook is done by setting a thread variable managed by
                // Biff8EncryptionKey. After the workbook is created, the thread variable can be cleared.
                Biff8EncryptionKey.setCurrentUserPassword(configuration.getPassword());
                try {
                    yield new HSSFWorkbook(in);
                } catch (final IOException e) {
                    throw new ProcessException("Failed to open XLS file", e);
                } finally {
                    Biff8EncryptionKey.setCurrentUserPassword(null);
                }
            }
        };
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
```diff
@@ -122,20 +112,32 @@ public void testNonExcelFile() {
...
+        assertInstanceOf(ProcessException.class, cause);
     }
 
     @Test
-    public void testOlderExcelFormatFile() {
-        ExcelRecordReaderConfiguration configuration = new ExcelRecordReaderConfiguration.Builder().build();
-        MalformedRecordException mre = assertThrows(MalformedRecordException.class, () -> new ExcelRecordReader(configuration, getInputStream("olderFormat.xls"), logger));
-        assertTrue(ExceptionUtils.getStackTrace(mre).contains("data appears to be in the OLE2 Format"));
+    public void testOlderExcelFormatFile() throws MalformedRecordException {
+        final List<RecordField> fields = Arrays.asList(
```
dan-s1, line 120:
```suggestion
        final List<RecordField> fields = List.of(
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
```diff
@@ -318,17 +328,47 @@ void testPasswordProtected() throws Exception {
...
+                .withAvoidTempFiles(true)
+                .withInputFileType(InputFileType.XLS)
+                .build();
+
+        InputStream inputStream = new ByteArrayInputStream(PASSWORD_PROTECTED_OLDER_EXCEL.toByteArray());
+        ExcelRecordReader recordReader = new ExcelRecordReader(configuration, inputStream, logger);
+        List<Record> records = getRecords(recordReader, false, false);
+
+        assertEquals(DATA.length, records.size());
+    }
```
dan-s1, line 346: Instead of a new test, combine this test and the previous test line 314-329 into a JUnit 5 org.junit.jupiter.params.ParameterizedTest with an org.junit.jupiter.params.provider.EnumSource. You can use the enum value in the configuration and to choose which password protected workbook to use.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
```diff
@@ -318,17 +328,47 @@ void testPasswordProtected() throws Exception {
...
+    void testPasswordProtectedOlderExcelWithoutPassword() {
+        RecordSchema schema = getPasswordProtectedSchema();
+        ExcelRecordReaderConfiguration configuration = new ExcelRecordReaderConfiguration.Builder()
+                .withSchema(schema)
+                .withInputFileType(InputFileType.XLS)
+                .build();
+
+        InputStream inputStream = new ByteArrayInputStream(PASSWORD_PROTECTED_OLDER_EXCEL.toByteArray());
+        assertThrows(Exception.class, () -> new ExcelRecordReader(configuration, inputStream, logger));
+    }
```
dan-s1, line 370: Same as previous comment make this a Paramaterized test by combining this test with the one on lines 348-358

dan-s1 (commented):
@zhtk great work so far. I had a few more suggestions.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/InputFileType.java
exceptionfactory: The `HSSF` acronym should not be used in user-facing labels and documentation. It is an acronym internal to the Apache POI project, and a pejorative one.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
dan-s1: @exceptionfactory Although I believe this is clearer code it would necessitate making changes in TestExcelHeaderSchemaStrategy to account for setting the InputFileType in the context otherwise the unit tests there will fail. Is this still worth it?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/InputFileType.java
exceptionfactory: Yes, including the years in the description is helpful.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
exceptionfactory: I'm not quite following the question, is what still worth it? Are you referring to "if-else" versus "switch", adding the new Input File Type property?

dan-s1: I meant the switch. I feel its clearer seeing which code gets called for which type.

exceptionfactory: I prefer if-else for a smaller numbers of options, and switch when it comes to larger numbers. I would leave it as-is, but would be fine either way.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelReader.java
zhtk: Changed

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/InputFileType.java
zhtk: Good point, changed

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/RowIterator.java
zhtk: Keeping as is, but switch also makes sense here

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
zhtk: Changed

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/excel/TestExcelRecordReader.java
zhtk: Changed

dan-s1:
@zhtk Can you please fix the conflicts you have locally and then push?

zhtk:
TBH I don't see conflicts, github interface says: "No conflicts with base branch. Changes can be cleanly merged.", and local merge also was performed cleanly. To make sure everything is ok, I have rebased and squashed all commits on the branch, hope now it's okay

dan-s1:
> TBH I don't see conflicts, github interface says: "No conflicts with base branch. Changes can be cleanly merged.", and local merge also was performed cleanly. To make sure everything is ok, I have rebased and squashed all commits on the branch, hope now it's okay

Things look aky now with the force push. Thanks!

dan-s1 (commented):
I noticed one more thing (hopefully the last :) )

dan-s1:
@exceptionfactory I approve these changes. Is there anything else you see which needs changing?

exceptionfactory (approved):
Thanks for bringing this to completion @zhtk, and thanks for the reviews along the way @dan-s1, the latest version looks good.

I was going to note that the `displayName()` element on the property is not necessary, but I see that follows the other properties for now, so removing the `displayName()` across the board can be handled separately.

+1 merging

zhtk:
I also thank you for reviews and for patience in getting these changes merged 🙇

PR #9888, opened 2025-04-22
NIFI-14071: Labels not rendering lines starting with spaces

samar121:
🛠️ Fix: Label rendering issue in boundedMultilineEllipsis
This MR addresses a bug where SVG labels using boundedMultilineEllipsis were not displaying correctly when rendered via D3, especially during interactions like dragging or viewport changes.

✅ Root Cause:
The issue stemmed from inaccurate line height and line count calculations, which affected label wrapping and positioning. In some cases, the initial tspan was not receiving the proper y attribute, causing the entire label to be misaligned or not visible.

🔧 Fix Details:
Reordered the logic inside the while (word) loop to:

1. Assign y position immediately on first word append, ensuring the tspan gets positioned correctly even for single-word lines.
2. Calculate lineHeight and lineCount using getBoundingClientRect() only once (on the first word of the first line), improving performance and consistency.
3. Preserve leading spaces using xml:space="preserve" to match label formatting.
4. Ensure all wrapped lines are properly positioned using dy="1.2em" and line limit logic.

📷 Screenshots:
Before (Broken Labels):
Screenshot from 2025-04-21 22-23-15

After (Fixed Labels):
Screenshot from 2025-04-22 10-02-05

🔬 Validation:
Tested rendering behavior on:
1. Long and short label strings
2. Multi-line wrapping scenarios
3. Resizing, dragging, zooming, and panning

All scenarios now render correctly as expected.

📌 Note:
This fix improves layout consistency and ensures robust label handling across different canvas states and screen sizes.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
@@ -1484,7 +1484,7 @@ export class CanvasUtils {
 
             // Split words normally, letting internal whitespace collapse
             const words = trimmedLine.split(/\s+/).reverse();
-            if (words.length > 0) {
+            if (leadingWhitespace.length > 0 && words.length > 0) {
                 words[words.length - 1] = leadingWhitespace + words[words.length - 1]; // Prepend leading space to the first word
             }
 
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
@@ -1478,16 +1478,20 @@ export class CanvasUtils {
         let lineHeight = height;
 
         for (const fullLine of lines) {
-            // Preserve leading spaces per word
-            const words: string[] = fullLine.match(/(\s*\S+)/g)?.reverse() || [];
+            // Extract and preserve only the leading whitespace at the start of the line
+            const trimmedLine = fullLine.trimStart();
+            const leadingWhitespace = fullLine.slice(0, fullLine.length - trimmedLine.length);
+
+            // Split words normally, letting internal whitespace collapse
+            const words = trimmedLine.split(/\s+/).reverse();
+            if (words.length > 0) {
+                words[words.length - 1] = leadingWhitespace + words[words.length - 1]; // Prepend leading space to the first word
+            }
 
             let newLine = true;
             let line: string[] = [];
 
-            let tspan = selection.append('tspan')
-                .attr('x', x)
-                .attr('width', width)
-                .attr('xml:space', 'preserve'); // preserve leading spaces
+            let tspan = selection.append('tspan').attr('x', x).attr('width', width).attr('xml:space', 'preserve');
 
             // go through each word
             let word = words.pop();
@@ -1495,9 +1499,8 @@ export class CanvasUtils {
             while (word) {
                 // add the current word
                 line.push(word);
-
                 // update the label text
-                tspan.text(line.join(''));
+                tspan.text(line.join(' '));
 
                 if (!lineCountCalculated) {
                     const bbox = tspan.node().getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


samar121:
Thanks for the feedback! I’ve now added comments back into the boundedMultilineEllipsis method to explain the reasoning behind the changes and key parts of the logic.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
```diff
@@ -1478,11 +1478,16 @@ export class CanvasUtils {
         let lineHeight = height;
 
         for (const fullLine of lines) {
-            const words: string[] = fullLine.split(/\s+/).reverse();
+            // Preserve leading spaces per word
+            const words: string[] = fullLine.match(/(\s*\S+)/g)?.reverse() || [];
```
rfellows, line 1482: This might be too aggressive. Now, all leading whitespace is considered part of the subsequent word treating space, tab, and line breaks as regular characters and not as an opportunity to wrap text at those points. Prior to this change, whitespace was treated similar to how it renders in html, multiple consecutive spaces would render as a single one and would allow wrapping there. Now, if you have many spaces together between words the leading spaces before the word are now bound together and will not allow wrapping in the middle of the spaces, treating a string like "          hello" as 15 characters rather than 5.

Consider the string `Hello                                           World`
Before change:
Screenshot 2025-04-25 at 09 06 20

After change:
Screenshot 2025-04-25 at 09 06 51

I think the best solution would be to only preserve whitespace at the beginning of a line(start of string or following line break) and let whitespace collapse elsewhere like it did before.

Another consequence of the current solution is that since the leading space is preserved, it can make lines look indented when they do wrap to the next line:
Screenshot 2025-04-25 at 09 02 54

rfellows (changes requested):
Thanks for adding the comments back in but I think there is an opportunity to make this a bit better still...

samar121:
Thanks a lot for the thoughtful feedback, @rfellows!
You were absolutely right — the earlier change was too aggressive in preserving all leading whitespace, which unintentionally affected how the lines wrapped, especially when there were multiple spaces between words.
I've updated the logic to preserve leading whitespace only at the beginning of each line (i.e., at the start of a string or immediately after a line break) while allowing normal whitespace collapsing between words,

✅ After the change — for a string like      Hello                                           World
image

✅ Leading space preservation and unintended indentation issues are also fixed:
image

Please let me know if you spot anything else!

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/service/canvas-utils.service.ts
```diff
@@ -1478,19 +1478,27 @@ export class CanvasUtils {
...
-            const words: string[] = fullLine.split(/\s+/).reverse();
+            // Extract and preserve only the leading whitespace at the start of the line
+            const trimmedLine = fullLine.trimStart();
+            const leadingWhitespace = fullLine.slice(0, fullLine.length - trimmedLine.length);
+
+            // Split words normally, letting internal whitespace collapse
+            const words = trimmedLine.split(/\s+/).reverse();
+            if (words.length > 0) {
+                words[words.length - 1] = leadingWhitespace + words[words.length - 1]; // Prepend leading space to the first word
+            }
```
rfellows, line 1489: There is a minor performance hit here. It can make resizing of the label feel jittery. It can be fixed by adding an additional condition here to only update the array in this manor if there is actually whitespace to preserve:

```suggestion
            if (leadingWhitespace.length > 0 && words.length > 0) {
                words[words.length - 1] = leadingWhitespace + words[words.length - 1]; // Prepend leading space to the first word
            }
```

rfellows (changes requested):
One last thing...

samar121:
Thanks for the catch @rfellows 
I've updated the logic to only prepend leading whitespace if it actually exists, as suggested.
This should eliminate unnecessary string operations and make the label resizing feel smoother.

samar121:
Hi team,

Thank you for merging the changes and mentioning me!
My GitHub username and Apache JIRA ID are different, so the issue was not automatically assigned to me.

Could you please assign NIFI-14071 to me for proper tracking?
My Apache JIRA ID is:  samarpratap

Thanks again!
cc: @rfellows

rfellows:
@samar121 Thanks again for the pull request! I have updated the JIRA to be assigned to you.

PR #9889, opened 2025-04-22
NIFI-14489 Allow for custom named Process Groups when importing flows from Registry

buczi:
# Summary

[NIFI-14489](https://issues.apache.org/jira/browse/NIFI-14489)

New input field added in the `Import From Registry` form. It is optional and user can add there a new name of the processor. It satisfies a case when user creates multiple flows from the same source and wants a convenient way to rename them before creation. Currently user would have to at first import flow and then change the name through after going to `Settings` page.

### Now
Old-Version

### After Changes
NIFI-14489

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessGroupResource.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessGroupResource.java
@@ -1047,6 +1047,11 @@ public class ProcessGroupResource extends FlowUpdateResource<ProcessGroupImportE
             // If there are any Parameter Providers referenced by Parameter Contexts, resolve these to point to the appropriate Parameter Provider, if we are able to.
             unresolvedParameterProviders.addAll(serviceFacade.resolveParameterProviders(flowSnapshot, NiFiUserUtils.getNiFiUser()));
 
+            // If Process Group name was provided in the request, prioritize it over the name from version control
+            if (requestProcessGroupEntity.getComponent().getName() != null) {
+                flowSnapshot.getFlowContents().setName(requestProcessGroupEntity.getComponent().getName());
+            }
+
             // Step 6: Update contents of the ProcessGroupDTO passed in to include the components that need to be added.
             requestProcessGroupEntity.setVersionedFlowSnapshot(flowSnapshot);
         }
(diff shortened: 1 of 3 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
```diff
@@ -397,6 +399,7 @@ export class ImportFromRegistry extends CloseOnEscapeDialog implements OnInit {
                 }),
                 disconnectedNodeAcknowledged: this.clusterConnectionService.isDisconnectionAcknowledged(),
                 component: {
+                    name: this.importFromRegistryForm.get('customFlowName')?.value,
```
sardell, line 402: If a user types in a value in the Custom Versioned Flow Name field then removes the value, an empty string is sent to the backend which returns a 400 (see screenshot below). There are a couple of ways you could handle this, but this is the most straightforward IMO.
```suggestion
                    name: this.importFromRegistryForm.get('customFlowName')?.value || null,
```
Screenshot 2025-05-05 at 11 52 41 AM

sardell (changes requested):
Thanks for the PR, @buczi! I left one comment but otherwise looks good to me.

sardell (approved):
+1 (non-binding). Thanks, @buczi!

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessGroupResource.java
```diff
@@ -1047,6 +1047,11 @@ public Response createProcessGroup(
             // If there are any Parameter Providers referenced by Parameter Contexts, resolve these to point to the appropriate Parameter Provider, if we are able to.
             unresolvedParameterProviders.addAll(serviceFacade.resolveParameterProviders(flowSnapshot, NiFiUserUtils.getNiFiUser()));
 
+            // If Process Group name was provided in the request, prioritize it over the name from version control
+            if (StringUtils.isNotBlank(requestProcessGroupEntity.getComponent().getName())) {
+                flowSnapshot.getFlowContents().setName(requestProcessGroupEntity.getComponent().getName());
+            }
+
```
mosermw, line 1054: I feel like this section of code doesn't belong here, but instead it belongs below (near line 1115) before the call to serviceFacade.updateProcessGroupContents().  You'll even see a comment there "we do want to use the name of the Process Group that is in the Flow Contents".  You might consider updating that comment, too.

exceptionfactory (changes requested):
Thanks for suggesting the change @buczi, and thanks for the feedback @mosermw and @sardell.

On review, however, adding this option does not seem like a good idea right now because it introduces an additional option for something that is can already be changed later. It is important to be careful when considering UI changes, as too many options can create unnecessary clutter.

It would be better to consider the desired use more completely on the associated Jira issue, instead of making this change directly.

For these reasons, I'm closing the PR for now.

PR #9891, opened 2025-04-22
NIFI-14491 Support other types in CreateBoxFileMetadataInstance

awelless:
# Summary

[NIFI-14491](https://issues.apache.org/jira/browse/NIFI-14491)
Adding support for numeric, date and array types in `CreateBoxFileMetadataInstance`.
Adding support for date type in `UpdateBoxFileMetadataInstance`.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CreateBoxFileMetadataInstance.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CreateBoxFileMetadataInstance.java
@@ -37,14 +37,20 @@ import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.processors.box.utils.BoxDate;
 import org.apache.nifi.serialization.RecordReader;
 import org.apache.nifi.serialization.RecordReaderFactory;
 import org.apache.nifi.serialization.record.Record;
+import org.apache.nifi.serialization.record.RecordField;
+import org.apache.nifi.serialization.record.RecordFieldType;
 
 import java.io.InputStream;
+import java.time.LocalDate;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import static java.lang.String.valueOf;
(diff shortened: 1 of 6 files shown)


nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CreateBoxFileMetadataInstance.java
```diff
@@ -223,20 +229,58 @@ private void processRecord(Record record, Metadata metadata, List<String> errors
...
+        if (record.getValue(field) == null) {
+            return;
+        }
+
+        final RecordFieldType fieldType = field.getDataType().getFieldType();
+        final String fieldName = field.getFieldName();
+        final String path = "/" + fieldName;
+
+        if (isNumber(fieldType)) {
+            metadata.add(path, record.getAsDouble(fieldName));
```
sfc-gh-rpaulin, line 254: Since isNumber is checking up to BigInteger and BigDecimal (arbitrary precision) it's possible that a number larger than double precision will be retrieved.  Given box only accepts up to double precision this seems appropriate.  Just wanted to make sure we're calling out the preference that we are opting for being more accepting of numbers being rounded or set to positive or negative infinity over throwing an error.  Agree with the approach but wanted to call out the design decision.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/test/java/org/apache/nifi/processors/box/CreateBoxFileMetadataInstanceTest.java
```diff
@@ -82,7 +84,12 @@ void testSuccessfulMetadataCreation() {
                   "documentType": "Q1 plans",
                   "competitiveDocument": "no",
                   "status": "active",
-                  "author": "Jones"
+                  "author": "Jones",
+                  "int": 1,
+                  "double": 1.234,
```
sfc-gh-rpaulin, line 89: Might be good to add a large number to demonstrate we are preferring rounding over erroring when numbers requiring more than 64 bits.

awelless: Good point. Added a value presenting that

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CreateBoxFileMetadataInstance.java
```diff
@@ -223,20 +229,58 @@ private void processRecord(Record record, Metadata metadata, List<String> errors
...
+    }
+
+    private boolean isNumber(final RecordFieldType fieldType) {
+        final boolean isInteger = RecordFieldType.BIGINT.equals(fieldType) || RecordFieldType.BIGINT.isWiderThan(fieldType);
+        final boolean isFloat = RecordFieldType.DECIMAL.equals(fieldType) || RecordFieldType.DECIMAL.isWiderThan(fieldType);
+        return isInteger || isFloat;
+    }
+
+    private boolean isDate(final RecordFieldType fieldType) {
+        return RecordFieldType.DATE.equals(fieldType);
```
sfc-gh-ncover, line 277:
```suggestion
        return RecordFieldType.DATE.equals(fieldType) || RecordFieldType.TIMESTAMP.equals(fieldType);
```
There is also a timestamp record type, would this be picked up by date or should it be also added in?

awelless: Box supports dates only. I.e. the time part is `00:00:00.000`. I find that treating timestamps as a date is quite risky, as we'll have to drop the time.

pvillard31 (approved):

PR #9897, opened 2025-04-23
NIFI-14470 - PublishKafka; compatibility with PublishKafka26 keys

greyp9:
The JIRA notes a difference in the handling of the `kafka.key` setting in the new `PublishKafka` processor, when
compared to the deprecated `PublishKafka_2_6` processor.

* as expression language is noted as supported in the component property, the code should honor that
* the expression language resolved value should be used as is, rather than as a lookup into the FlowFile attribute set

There were a couple of other component properties that employ the expression language evaluation, so those
PropertyDescriptors have been updated to reflect the code usage.

As NiFi 2.0 was released in October 2024, there might be significant uptake of the processor (with its incompatible behavior).  Weighing this against the compatibility expectations of the community, this seems like the best path forward, but it is also a reasonable point of discussion.

# Summary

[NIFI-14470](https://issues.apache.org/jira/browse/NIFI-14470)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


exceptionfactory (commented):
Thank for proposing a solution to this issue @greyp9.

On the one hand, the current implementation in arguably broken, since the `Kafka Key` property indicates support for Expression Language.

On the other hand, supporting EL for the property value introduces a level of indirection that could also be confusing.

Since this change aligns with the earlier implementation, it seems like the best way forward, but I will leave this open for a bit longer to see if there is any additional feedback.

exceptionfactory (approved):
Thanks again for addressing this issue @greyp9, this looks good as it stands. +1 merging

PR #9918, opened 2025-05-07
NIFI-14541 Add Scoped Authorization for Flow Registry Clients

exceptionfactory:
# Summary

[NIFI-14541](https://issues.apache.org/jira/browse/NIFI-14541) Adds scoped authorization for listing, adding, modifying, and removing Flow Registry Clients.

The changes are compatible with existing authorization policies based on the read and write access to the Controller resource. Introducing a new Registry Client Resource Type aligns with other component authorization, and enables authorization checking for operations on Flow Registry Clients nested under the Controller resource.

The `StandardFlowRegistryClientNode` already returns the Controller resource as the parent authorizable object, supporting nested authorization checking.

Enabling the standard Managed Authorizer and configuring a Flow Registry Client exercises new authorization checking against existing policies at the Controller level.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-authorization/src/main/java/org/apache/nifi/authorization/resource/ResourceFactory.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-authorization/src/main/java/org/apache/nifi/authorization/resource/ResourceFactory.java
@@ -553,6 +553,7 @@ public final class ResourceFactory {
                     case InputPort -> "Input Port";
                     case OutputPort -> "Output Port";
                     case Processor -> "Processor";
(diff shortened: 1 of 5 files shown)


pvillard31 (approved):
Thanks for the changes @exceptionfactory - it looks straightforward and thanks for fixing `uuid` versus `id`, I believe this should be fixed in some other places in this `ControllerResource` class but it is definitely not critical. I also noticed some inconsistencies with the `SecurityRequirement` annotations and the permissions on potential referenced controller services but I don't think we need to go into that level of details of registry clients. +1, merging

PR #9923, opened 2025-05-08
NIFI-14545 Remove casting for PublishKafka to BufferedInputStream

sammu97:
This PR aims to avoids restriction on mark() supplied length, currently set to around 1MB, when publishing payloads to Kafka. This fix seems to have been lost during the Kafka Refactor, as part of [NIFI-14375](https://issues.apache.org/jira/browse/NIFI-14375).

# Summary

[NIFI-14545](https://issues.apache.org/jira/browse/NIFI-14545)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/PublishKafkaIT.java
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/PublishKafkaIT.java
@@ -20,11 +20,7 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.common.header.Header;
-import org.apache.nifi.json.JsonRecordSetWriter;
-import org.apache.nifi.json.JsonTreeReader;
 import org.apache.nifi.reporting.InitializationException;
-import org.apache.nifi.schema.access.SchemaAccessUtils;
-import org.apache.nifi.util.MockFlowFile;
 import org.apache.nifi.util.TestRunner;
 import org.apache.nifi.util.TestRunners;
 import org.junit.jupiter.api.MethodOrderer;
@@ -32,16 +28,12 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestMethodOrder;
 
 import java.nio.charset.StandardCharsets;
-import java.time.Duration;
-import java.util.ArrayList;
-import java.util.Properties;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import static org.apache.kafka.clients.consumer.ConsumerConfig.MAX_POLL_RECORDS_CONFIG;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 @TestMethodOrder(MethodOrderer.MethodName.class)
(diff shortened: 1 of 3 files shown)


sammu97:
I was finding difficulties actually performing a test for this, specifically mocking a KafkaProducerService. Feel free to suggest any methods of unit testing for this.

pvillard31:
You can fix the checkstyle issue by removing the unused import. And you can apply this patch to add an integration test that reproduces the issue and confirms that you fixed it. Thanks a lot for your contribution!

[0001-Adding-integration-test.patch](https://github.com/user-attachments/files/20102296/0001-Adding-integration-test.patch)

sammu97:
Thanks @pvillard31 !

pvillard31 (changes requested):
Instead of

````java

        public void process(final InputStream in) {
            try (final InputStream is = in) {
                final Iterator<KafkaRecord> records = kafkaConverter.convert(attributes, is, inputLength);
                producerService.send(records, publishContext);
            } catch (final Exception e) {
                publishContext.setException(e); // on data pre-process failure, indicate this to controller service
                producerService.send(Collections.emptyIterator(), publishContext);
            }
        }
````

we can have:

````
        public void process(final InputStream in) {
            try {
                final Iterator<KafkaRecord> records = kafkaConverter.convert(attributes, in, inputLength);
                producerService.send(records, publishContext);
            } catch (final Exception e) {
                publishContext.setException(e); // on data pre-process failure, indicate this to controller service
                producerService.send(Collections.emptyIterator(), publishContext);
            }
        }
````

**Review by greyp9 — COMMENTED** [review]
Hi @sammu97, this is a nice catch!  Thanks for noting the problem, and working through the process to contribute a fix.

I would encourage you to perform some manner of verification on the records sent by `PublishKafka`.  To help leverage the design of the existing integration tests, it would also be better to break out the the new test into a separate test class.

In order to speed up the suite of integration tests associated with the kafka processors, they all derive from `AbstractKafkaBaseIT`, which starts a single kafka server via `testcontainers`, which is used for all derived tests.  In order to prevent the tests from interfering with each other, the Kafka topic used for a test is derived from the test name.

```
runner.setProperty(PublishKafka.TOPIC_NAME, getClass().getName());
```

The current `PublishKafka` tests use NiFi's `TestRunner` to publish the payload under test, and then verify directly via kafka client library APIs.  There are several examples of this in the package you're already working in.

In the test publish method, this assertion would verify the expected record count associated with the test FlowFile:

```
        MockFlowFile flowFile = runner.getFlowFilesForRelationship(PublishKafka.REL_SUCCESS).getFirst();
        assertEquals(Integer.toString(EXPECTED_RECORD_COUNT), flowFile.getAttribute("msg.count"));
```

For the suggested consume method, you'll want to make a copy of an existing consume, with a couple of modifications:

- Rather than accept the default Kafka consume count of 500, you'll want to set that higher.

```
        final Properties kafkaConsumerProperties = getKafkaConsumerProperties();
        kafkaConsumerProperties.setProperty(MAX_POLL_RECORDS_CONFIG, Integer.toString(EXPECTED_RECORD_COUNT * 2));
```

- Even with this change, the Kafka consumer client chunks the records it consumes into smaller batches.  So rather than calling `consumer.poll()` once, you'll want to do multiple times.  Something like:

```
            final List<ConsumerRecord<String, String>> records = new ArrayList<>();
            for (int i = 0; (i < 5); ++i) {
                final ConsumerRecords<String, String> recordsPoll = consumer.poll(Duration.ofSeconds(1));
                recordsPoll.forEach(records::add);;
            }
            assertEquals(EXPECTED_RECORD_COUNT, records.size());
```

This works on my machine, but may need adjustment to work everywhere it needs to work.


[1] https://github.com/apache/nifi/blob/main/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/PublishKafkaIT.java#L51

**Comment by greyp9** [comment]
Since you found the trigger for the behavior to be a payload of >1MB, you might also adjust the test to a smaller size, which is still above the threshold:

```
    private static final int EXPECTED_RECORD_COUNT = 20000; // Expect 20,000 records
    private static final int MAX_MESSAGE_SIZE = 2 * 1024 * 1024; // 2MB
```

sammu97:
@pvillard31 @greyp9 Thanks for the recommendations and overall help on this! I have applied all changes requested, including a new consume test to test on the large batch.

greyp9:
> @pvillard31 @greyp9 Thanks for the recommendations and overall help on this! I have applied all changes requested, including a new consume test to test on the large batch.

If you would, please create a new file called `PublishKafkaLargePayloadIT.java`, and move the test changes there.

sammu97:
@greyp9 Apologies, missed that in your previous comment! Will do

greyp9 (approved):
Thanks for working through the process; this looks great!

pvillard31 (approved):

pvillard31:
Merged with https://github.com/apache/nifi/commit/4ea4e6397f7d27e8b90613d2cee5543c6ef9e5e2

PR #9927, opened 2025-05-09
NIFI-14555 - Option to empty parameter values in git registry clients

pvillard31:
# Summary

[NIFI-14555](https://issues.apache.org/jira/browse/NIFI-14555) - Option to empty parameter values in git registry clients

The goal of this PR is to add an optional property for all Git based registry clients. The property "Empty Parameters" will default to false to retain current behavior. If set to true, all values of all parameters will be removed before the flow is being versioned and stored in the git repo.

Parameters are usually environment dependent and it might not be desirable to include in the versioned flow the parameters values that have been set in a lower/dev environment while designing the flow.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
@@ -17,7 +17,7 @@
 
 package org.apache.nifi.registry.flow.git;
 
-import org.apache.nifi.components.AllowableValue;
+import org.apache.nifi.components.DescribedValue;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.ValidationContext;
 import org.apache.nifi.components.ValidationResult;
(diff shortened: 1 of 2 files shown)


pvillard31:
Thanks @exceptionfactory - good point, adjusted the code

nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
```diff
@@ -93,12 +94,14 @@ public abstract class AbstractGitFlowRegistryClient extends AbstractFlowRegistry
...
             .build();
 
-    public static final PropertyDescriptor EMPTY_PARAMETERS = new PropertyDescriptor.Builder()
-            .name("Remove Parameter Values")
-            .description("If true, the values of all parameters will be removed from the flow before storing it.")
-            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
-            .allowableValues("true", "false")
-            .defaultValue("false")
+    static final AllowableValue RETAIN = new AllowableValue("retain", "Retain", "Will not modify the parameter values");
+    static final AllowableValue REMOVE = new AllowableValue("remove", "Remove", "Will remove the parameter values");
```
exceptionfactory, line 98: Instead of using `AllowableValue`, it would be helpful to define this as an `enum`, there the enum name can be the value itself, as opposed to the lowercase `retain` and `remove`. This also makes the property retrieval cleaner with `asAllowableValue()`.

exceptionfactory (changes requested):
Thanks for adjusting the approach @pvillard31. I noted one more recommendation to use an `enum` implementing `DescribedValue` for the definition, then this should be ready to go.

nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
pvillard31: Thanks @exceptionfactory - pushed a commit to use an enum instead

exceptionfactory (approved):

PR #9941, opened 2025-05-13
[NIFI-14563] - Display inherited parameter contexts in the correct order

rfellows:
# Summary
[NIFI-14563](https://issues.apache.org/jira/browse/NIFI-14563)

Inherited parameter contexts don't display in the correct order

The issue was that the initial order of the selected/inherited parameter contexts was based off of the order of the list of ALL parameter contexts and not of the inherited contexts defined by the one being edited.


scottyaslan (commented):
Reviewing...

scottyaslan (approved):
+1 Verified that the correct order is displayed for the selected parameter context inheritance order. Thanks for the update and the unit tests!

PR #9967, opened 2025-05-21
NIFI-14442 Show bundle version changes in versioned flow differences

mosermw:
# Summary

[NIFI-14442](https://issues.apache.org/jira/browse/NIFI-14442)

This PR collects bundle version differences between a local Versioned Process Group and its FlowRegistry version.
These differences will only appear after another local change has been made.  Bundle version differences still do not
count when determining a Versioned Process Group's state.

This does not handle the case when a Versioned Process Group is imported from a FlowRegistry and then local changes
are immediately made.  In this scenario, there will be no bundle version differences.  This is because that Process Group's
versioned snapshot is modified from what is in the FlowRegistry, to update bundle versions and references to controller services
and parameter providers.  However, if you restart NiFi, the PG's versioned snapshot is pulled from the FlowRegisry on
startup, and then bundle version differences will be shown.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
@@ -73,6 +73,10 @@ public class FlowDifferenceFilters {
             || isLogFileSuffixChange(difference);
     }
 
+    public static boolean isBundleChange(final FlowDifference difference) {
+        return difference.getDifferenceType() == DifferenceType.BUNDLE_CHANGED;
+    }
+
     private static boolean isSensitivePropertyDueToGhosting(final FlowDifference difference, final FlowManager flowManager) {
         final DifferenceType differenceType = difference.getDifferenceType();
         if (differenceType != DifferenceType.PROPERTY_SENSITIVITY_CHANGED && differenceType != DifferenceType.PROPERTY_ADDED) {
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
@@ -5531,12 +5531,7 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
                 + " but cannot find a Flow Registry with that identifier");
         }
 
-        VersionedProcessGroup registryGroup = null;
-        final VersionControlInformation vci = processGroup.getVersionControlInformation();
-        if (vci != null) {
-            registryGroup = vci.getFlowSnapshot();
-        }
-
+        VersionedProcessGroup registryGroup = versionControlInfo.getFlowSnapshot();
         if (registryGroup == null) {
             try {
                 final FlowVersionLocation flowVersionLocation = new FlowVersionLocation(versionControlInfo.getBranch(), versionControlInfo.getBucketIdentifier(),
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
@@ -286,6 +286,8 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public final class DtoFactory {
@@ -2770,10 +2772,18 @@ public final class DtoFactory {
 
    public Set<ComponentDifferenceDTO> createComponentDifferenceDtosForLocalModifications(final FlowComparison comparison, final VersionedProcessGroup localGroup, final FlowManager flowManager) {
        final Map<ComponentDifferenceDTO, List<DifferenceDTO>> differencesByComponent = new HashMap<>();
+       final Map<ComponentDifferenceDTO, Set<DifferenceDTO>> bundleDifferencesByComponent = new HashMap<>();
 
        final Map<String, VersionedProcessGroup> versionedGroups = flattenProcessGroups(comparison.getFlowA().getContents());
 
        for (final FlowDifference difference : comparison.getDifferences()) {
+           // capture bundle differences and dedupe those differences
+           if (FlowDifferenceFilters.isBundleChange(difference)) {
+               final ComponentDifferenceDTO componentDiff = createBundleDifference(difference);
+               final Set<DifferenceDTO> differences = bundleDifferencesByComponent.computeIfAbsent(componentDiff, key -> new HashSet<>());
+               differences.add(createDifferenceDto(difference));
+           }
+
            // Ignore any environment-specific change
            if (FlowDifferenceFilters.isEnvironmentalChange(difference, localGroup, flowManager)) {
                continue;
@@ -2787,12 +2797,19 @@ public final class DtoFactory {
 
            final ComponentDifferenceDTO componentDiff = createComponentDifference(difference);
            final List<DifferenceDTO> differences = differencesByComponent.computeIfAbsent(componentDiff, key -> new ArrayList<>());
+           differences.add(createDifferenceDto(difference));
+       }
 
-           final DifferenceDTO dto = new DifferenceDTO();
-           dto.setDifferenceType(difference.getDifferenceType().getDescription());
-           dto.setDifference(difference.getDescription());
-
-           differences.add(dto);
+       if (!differencesByComponent.isEmpty()) {
+           // differences were found, so now let's add back in any BUNDLE_CHANGED differences
+           // since they were initially filtered out as an environment-specific change
+           bundleDifferencesByComponent.forEach((key, value) -> {
+               List<DifferenceDTO> values = value.stream().toList();
+               differencesByComponent.merge(key, values, (v1, v2) -> {
+                   v1.addAll(v2);
+                   return v1;
+               });
+           });
        }
 
        for (final Map.Entry<ComponentDifferenceDTO, List<DifferenceDTO>> entry : differencesByComponent.entrySet()) {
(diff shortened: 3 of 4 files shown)


NissimShiman:
Reviewing...

joewitt:
@bbende Can you take a look from a review perspective?

NissimShiman:
I think letting users know that a component bundle update is occurring is a nice addition.  Especially the way it is being done conditionally.
I've noticed that the list of local changes has a "Go To" option next to the "Component Bundle Changed" which doesn't lead anywhere.  Would it be a big deal to remove that option for this?

Also noticed that if a versioned PG is in flow, NiFi is restarted, PG is updated (and new version pushed to registry) and then is rolled back and then a new change is made (but not committed yet) , the component bundle update does not show up as a local change, but if NiFi is again restarted it will.  Is this expected?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -5531,12 +5531,7 @@ public FlowComparisonEntity getLocalModifications(final String processGroupId) {
                 + " but cannot find a Flow Registry with that identifier");
         }
```
NissimShiman, line 5583: Nice catch, based on the above code

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
```diff
@@ -2802,6 +2819,24 @@ public Set<ComponentDifferenceDTO> createComponentDifferenceDtosForLocalModifica
...
+    * Create a custom difference with a simpler difference description, only for matching descriptions.
+    */
+   DifferenceDTO createDifferenceDto(final FlowDifference difference) {
+       final DifferenceDTO dto = new DifferenceDTO();
+       dto.setDifferenceType(difference.getDifferenceType().getDescription());
+
+       String description = difference.getDescription();
+       final Matcher matcher = DIFFERENCE_DESCRIPTION_PATTERN.matcher(description);
+       if (matcher.matches()) {
+           description = matcher.group(1) + " version " + matcher.group(2) + " to " + matcher.group(3);
```
NissimShiman: I think this is a good idea as well.  The default is unwieldy.

mosermw:
Thanks for having a look at this PR @NissimShiman

> I've noticed that the list of local changes has a "Go To" option next to the "Component Bundle Changed" which doesn't lead anywhere. Would it be a big deal to remove that option for this?

I was able to get something to work, then I noticed there is an [open PR for NIFI-14637](https://github.com/apache/nifi/pull/10028) that does almost exactly what I was going to do.  So I think we can wait for that PR to merge, then I can modify this PR to rebase and take advantage of that code.

> Also noticed that if a versioned PG is in flow, NiFi is restarted, PG is updated (and new version pushed to registry) and then is rolled back and then a new change is made (but not committed yet) , the component bundle update does not show up as a local change, but if NiFi is again restarted it will. Is this expected?

Yes it's expected.  This is because the ProcessGroup stores a copy of its versioned snapshot, which typically has already been modified from what's in the FlowRegistry.

mosermw:
> > I've noticed that the list of local changes has a "Go To" option next to the "Component Bundle Changed" which doesn't lead anywhere. Would it be a big deal to remove that option for this?
> 
> I was able to get something to work, then I noticed there is an [open PR for NIFI-14637](https://github.com/apache/nifi/pull/10028) that does almost exactly what I was going to do. So I think we can wait for that PR to merge, then I can modify this PR to rebase and take advantage of that code.

The PR 10028 got merged, so I rebased this PR and added the condition to prevent showing the "Go To" action for bundle version differences.

NissimShiman:
I have tested the "Go To" option will not appear for these changes (and looked over the code as well).
Thank you @mosermw for resolving that.

This now performs as described in the jira (with the known edge case in the comment still present).

This lets the user know in many cases that a bundle version change is about to be persisted to registry, but maybe the edge case could be worked out as well.

More work would be needed so a process group's versioned snapshot should not be modified from what is in the FlowRegistry to make this edge case work as well, but that is a lot of work.

@bbende Is this something that would make sense to be done as part of this PR? or this something that should be left alone?
(and, if it should be left alone, is the PR good as is?  If we are OK leaving the edge case out, I am +1 on this.)

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
```diff
@@ -2802,6 +2819,24 @@ public Set<ComponentDifferenceDTO> createComponentDifferenceDtosForLocalModifica
        return differencesByComponent.keySet();
    }
 
+   private static final Pattern DIFFERENCE_DESCRIPTION_PATTERN = Pattern.compile(".*artifact='(.*)'.*version='(.*)'].*version='(.*)']'$");
```
exceptionfactory: Introducing this pattern to reformat the description seems potentially brittle. It would be better to either change it elsewhere, or build the description using values available from the `FlowDifference`.

exceptionfactory (changes requested):
Thanks for proposing this improvement to highlighting bundle version differences @mosermw. The concept seems useful in some scenarios, but I'm not sure about the implementation, particularly reformatting the structure in the `DtoFactory`. It seems like this should be implemented elsewhere as part of the comparison process.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
mosermw: I appreciate your eyes on this PR @exceptionfactory.  This was the least intrusive approach I could think of at the time, but I agree it is potentially brittle.  I will see what I can do to make improvements.

exceptionfactory: Thanks for looking into alternative approaches, sounds good.

mosermw: I moved this code to the ConciseEvolvingDifferenceDescriptor class where some other difference descriptions are generated, and I do think this made the code cleaner.

Should I rebase this PR onto main, to try to resolve the macos build/test issues in the CI checks?  Or can those checks be ignored for now?

exceptionfactory: Thanks, rebasing would be a good next step.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -73,6 +73,10 @@ public static boolean isEnvironmentalChange(final FlowDifference difference, fin
             || isLogFileSuffixChange(difference);
     }
 
+    public static boolean isBundleChange(final FlowDifference difference) {
```
exceptionfactory, line 109: This only seems to be used once, is it necessary?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
```diff
@@ -2821,6 +2846,18 @@ private void flattenProcessGroups(final VersionedProcessGroup group, final Map<S
...
    }
 
+   private ComponentDifferenceDTO createBundleDifference(final FlowDifference difference) {
+       VersionedComponent component = difference.getComponentB();
+
+       final ComponentDifferenceDTO dto = new ComponentDifferenceDTO();
+       dto.setComponentType(component.getComponentType().toString());
+       // bundle difference could apply to many components, but use the Process Group identifier as
+       // the Component Identifier here so that many similar component differences can be easily deduped
+       dto.setComponentId(component.getGroupIdentifier());
```
exceptionfactory, line 2885: This seems questionable, would it hide multiple component version differences in a given Process Group?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
```diff
@@ -2792,12 +2800,19 @@ public Set<ComponentDifferenceDTO> createComponentDifferenceDtosForLocalModifica
...
-           dto.setDifferenceType(difference.getDifferenceType().getDescription());
-           dto.setDifference(difference.getDescription());
-
-           differences.add(dto);
+           differences.add(createDifferenceDto(difference));
+       }
+
+       if (!differencesByComponent.isEmpty()) {
+           // differences were found, so now let's add back in any BUNDLE_CHANGED differences
+           // since they were initially filtered out as an environment-specific change
```
exceptionfactory, line 2837: Does this mean that bundle changes are no longer considered environment-specific changes?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
mosermw: Ah, I meant to call the new isBundleChange method from the isEnvironmentalChange method above.  This would follow the pattern of the other methods called by isEnvironmentalChange.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
mosermw: Bundle changes are still considered to be an environmental change, **_unless_** another local change has been made.  I didn't want bundle differences to be the only reason a Versioned Process Group would appear as locally changed, because a NiFi upgrade could potentially cause all Versioned PGs to appear locally changed.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/dto/DtoFactory.java
mosermw: This quite nicely dedupes only similar component differences.  As an example, if your Versioned Process Group contains 20 processors from nifi-standard-nar and 1 processor from nifi-jolt-nar, then once you make any non-environmental change, you would see 1 difference in the nifi-jolt-nar version and only 1 difference in the nifi-standard-nar version.  I didn't think showing 21 differences with 20 being the same would be desirable.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
mosermw: I added a commit for this.  Also rebased onto main because of conflicts and because this PR didn't build anymore due to the npm package.lock changes that were needed from a month ago.

NissimShiman:
Thank you @exceptionfactory for the review.
I did a quick recompile and test and it performs as described.
I am +1

NissimShiman (approved):

exceptionfactory:
Thanks for the confirmation @NissimShiman, and thanks again @mosermw! +1 merging

PR #9993, opened 2025-06-05
NIFI-14630 Corrected usage of KafkaClientComponent interface

turcsanyip:
The PR removes `KafkaPublishComponent` / `KafkaClientComponent` interface implementation from `PublishKafka`. The single property that was used from `KafkaPublishComponent` has been moved to the processor (as other publisher specific properties).
`Kafka3ConnectionService` now implements `KafkaClientComponent` in order for the properties be defined in a single location. The properties in the interface have been updated from the controller service (they were used from the service effectively, the interface just provided property names for the shared code).

# Summary

[NIFI-14630](https://issues.apache.org/jira/browse/NIFI-14630)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


mark-bathori (approved):
Thanks @turcsanyip for these fixes and corrections. The changes looks good.
LGTM +1

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-shared/src/main/java/org/apache/nifi/kafka/shared/component/KafkaClientComponent.java
```diff
@@ -32,12 +32,11 @@ public interface KafkaClientComponent {
...
     PropertyDescriptor BOOTSTRAP_SERVERS = new PropertyDescriptor.Builder()
             .name("bootstrap.servers")
-            .displayName("Kafka Brokers")
-            .description("Comma-separated list of Kafka Brokers in the format host:port")
+            .displayName("Bootstrap Servers")
+            .description("Comma-separated list of Kafka Bootstrap Servers in the format host:port. Corresponds to Kafka bootstrap.servers property")
             .required(true)
-            .addValidator(StandardValidators.HOSTNAME_PORT_LIST_VALIDATOR)
-            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
-            .defaultValue("localhost:9092")
```
pvillard31: Any reason for not keeping this validator and the default value?

turcsanyip: Thanks for flagging this!

In general, I did not want to change the behavior of the properties in order to eliminate compatibility issues at all. So I took the property definitions from `Kafka3ConnectionService` because those were used effectively.

On the other hand, it makes sense to add `HOSTNAME_PORT_LIST_VALIDATOR` because the existing `Bootstrap Servers` property values must be following this format already. Otherwise the property is unusable. So this seems a minimal risk change and therefore I'm leaning toward adding the validator.

Regarding `localhost:9092` default value: The "localhost" values are typically used during development only and they must be changed in real life use cases almost all the time. Personally, I don't prefer "developer defaults" exposed in prod apps (even if it is handy for devs). However, I see that we use "localhost" defaults in some places so if you think it is kind of a convention in NiFi, I can add it.

So I'm +1 for the validator and rather -1 for the default value. Please share what you think based on the background info provided above and I will change the code accordingly.

pvillard31: Yep, fair enough, sounds good to me, thanks @turcsanyip

exceptionfactory: I agree with removing the `localhost:9092` default value, and instead having it null is best, thanks @turcsanyip and @pvillard31.

pvillard31 (approved):
Thanks @turcsanyip - I had another look at the changes and I'm a +1
@exceptionfactory - feel free to merge if you're a +1 as well or I can do it

PR #9995, opened 2025-06-06
NIFI-14629 Added timeout blocking to Kinesis StandardRecordProcessorB…

dariuszseweryn:
…locker

# Summary

[NIFI-14629](https://issues.apache.org/jira/browse/NIFI-14629)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

- [x] (there are no new dependencies) New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [x] (there are no new dependencies) New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [x] (there is no new documentation) Documentation formatting appears as expected in rendered files


nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/StandardRecordProcessorBlocker.java
```diff
@@ -38,7 +76,17 @@ public synchronized void block() {
...
-    public synchronized void unblock() {
+    public void unblock() {
+        unblockPrivate(timestampProvider.get());
+    }
+
+    public void unblockAndDisableTimeout() {
+        unblockPrivate(Long.MAX_VALUE);
+    }
+
+    private synchronized void unblockPrivate(final long newTimeoutAnchor) {
```
exceptionfactory: Use of `Private` in the method name should be avoided. `Internal` could work, but better to be more descriptive, such as `countDown()` based on what the method is doing.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/StandardRecordProcessorBlocker.java
```diff
@@ -17,19 +17,57 @@
...
+ * Unblocking is done by calling {@link #unblock()} or {@link #unblockAndDisableTimeout()}. The latter is used to disable the timeout mechanism to allow shutdown of background threads that could
+ * otherwise block indefinitely due to no calls to {@link #unblock()}.
+ */
 public class StandardRecordProcessorBlocker implements RecordProcessorBlocker {
 
     private CountDownLatch blocker = new CountDownLatch(0);
 
+    static final long BLOCK_AFTER_TIMEOUT_MILLIS = TimeUnit.SECONDS.toMillis(2);
+    private final Supplier<Long> timestampProvider;
+    private long timeoutAnchor = 0L;
```
exceptionfactory: The naming of this property is unclear, it should be renamed to something that more clearly describes its purpose.

exceptionfactory (changes requested):
Thanks for evaluating potential improvements to Kinesis processing @dariuszseweryn. On initial review, I noticed several issues.

The introduction of the new `unblockAndDisableTimeout()` method is only on the Standard implementation, which goes outside the bounds of the interface definition.

More to the substance of the implementation, however, the `Blocker` class now includes two CountDownLatch instances, and synchronization at different points, leading to a difficult to follow approach. I think the approach needs to be revisited and simplified, otherwise it is introducing more complexity to something that is already getting too complex to be maintainable. Some of this relates to fundamental design issues with the Processor itself, and attempting to work around them is going to be a losing battle in the end.

dariuszseweryn:
> The introduction of the new unblockAndDisableTimeout() method is only on the Standard implementation, which goes outside the bounds of the interface definition.

The interface defines only `await()` method — `block()`, `unblock()` are only on the Standard implementation — this was the accepted solution in the prior PR. `unblockAndDisableTimeout()` only builds on that approach. What you propose?

I have extracted `TimeoutActivationLatch` class and changed naming. Hopefully it is now clearer. PTAL

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/pause/StandardRecordProcessorBlocker.java
dariuszseweryn: Renamed to `timeoutBase`

dariuszseweryn:
> From an implementation point of view, it would be helpful to make use of java.time.Duration in several places, versus the separate TimeUnit and long milliseconds.

Changed to used Duration. I'll use it more often going forward.
> As far as the Supplier approach for the timestamp, is that necessary? It adds a layer of indirection that makes the implementation more difficult to follow, although I can see where it might be useful only for testing.

Yes, it helps for testing, that is the purpose. Public interface is oblivious of the Supplier though.
> Regarding testing, is the Noop implementation added to avoid timing issues with the tests?

Correct. I have changed to use the standard implementation and call `unblockInfinitely()` on it.
> One more high-level question, is there a reason for maintaining both the CountDownLatch and the Timeout Latch, as opposed to just the Timeout Latch? This still seems like some more fundamental that needs to be reconsidered.

I wanted to have a separation of concerns, but since I have merged TimeoutActivationLatch into StandardRecordProcessorBlocker I am now using a single latch and change behavior by changing `timeoutBase` parameter.

exceptionfactory (approved):
Thanks for making the adjustments @dariuszseweryn, the latest version is much more streamlined, and looks like a reasonable approach given the current implementation constraints with the Kinesis library.

PR #9975, opened 2025-05-28
NIFI-14596 Added logic to the ExcelHeaderSchemaStrategy to rename duplicate column names thereby ensuring no data loss or skewing of data.

dan-s1:
# Summary

[NIFI-14596](https://issues.apache.org/jira/browse/NIFI-14596)

# Tracking

Please complete the following tracking steps prior to pull request creation.

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI) issue created

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-00000`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-00000`

### Pull Request Formatting

- [x] Pull Request based on current revision of the `main` branch
- [x] Pull Request refers to a feature branch with one commit containing changes

# Verification

Please indicate the verification steps performed prior to pull request creation.

### Build

- [x] Build completed using `mvn clean install -P contrib-check`
  - [x] JDK 21

### Licensing

- [ ] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [ ] New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [ ] Documentation formatting appears as expected in rendered files

PR #9998, opened 2025-06-06
NIFI-14626 - Fix MongoDB connection string creation

pvillard31:
# Summary

[NIFI-14626](https://issues.apache.org/jira/browse/NIFI-14626) - Fix MongoDB connection string creation

When calling the constructor of ConnectionString, some authentication method absolutely require the username to be coming from the connection string regardless of the information provided via the Credentials that are set into the Client builder. This PR injects the username and password into the URI when creating the connection string to solve the problem.

I've confirmed proper behavior before/after this PR against a MongoDB Atlas instance in GCP. This way, the user does not have to provide the username/password in the URI property of the controller service that would expose sensitive information.

We may (I didn't investigate) be able to eliminate this approach if upgrading the client to a more recent major release but that would require additional work as they are breaking changes.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java
+++ b/nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java
@@ -24,6 +24,7 @@ import com.mongodb.WriteConcern;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoClients;
 import com.mongodb.client.MongoDatabase;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.Tags;
 import org.apache.nifi.annotation.lifecycle.OnDisabled;
@@ -96,9 +97,18 @@ public class MongoDBControllerService extends AbstractControllerService implemen
         }
 
         try {
-            final String uri = context.getProperty(URI).evaluateAttributeExpressions().getValue();
+            String uri = context.getProperty(URI).evaluateAttributeExpressions().getValue();
             final String user = context.getProperty(DB_USER).evaluateAttributeExpressions().getValue();
             final String passw = context.getProperty(DB_PASSWORD).evaluateAttributeExpressions().getValue();
+
+            // we need to create a new URI string that contains the user and password
+            // if not the creation of the connection string will fail for some
+            // authentication methods when calling the constructor of ConnectionString
+            if (user != null && passw != null && !uri.contains("@")) {
+                final String beforeInstance = StringUtils.substringBefore(uri, "://");
+                final String afterInstance = StringUtils.substringAfter(uri, "://");
+                uri = beforeInstance + "://" + user + ":" + passw + "@" + afterInstance;
+            }
 
             final MongoClientSettings.Builder builder = MongoClientSettings.builder();
             final ConnectionString cs = new ConnectionString(uri);


nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java
```diff
@@ -96,10 +97,19 @@ protected MongoClient createClient(ConfigurationContext context, MongoClient exi
...
             final String user = context.getProperty(DB_USER).evaluateAttributeExpressions().getValue();
             final String passw = context.getProperty(DB_PASSWORD).evaluateAttributeExpressions().getValue();
 
+            // we need to create a new URI string that contains the user and password
+            // if not the creation of the connection string will fail for some
+            // authentication methods when calling the constructor of ConnectionString
+            if (user != null && passw != null && !uri.contains("@")) {
+                final String beforeInstance = StringUtils.substringBefore(uri, "://");
+                final String afterInstance = StringUtils.substringAfter(uri, "://");
+                uri = beforeInstance + "://" + user + ":" + passw + "@" + afterInstance;
```
dan-s1, line 110: Per changes I needed to make for #9975, the following would be cleaner
```suggestion
                uri = "%s://%s:%s@%s".formatted(beforeInstance, user, passw, afterInstance); 
```

nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java
```diff
@@ -96,10 +97,19 @@ protected MongoClient createClient(ConfigurationContext context, MongoClient exi
...
+            }
```
dan-s1: It would be clearer to use URI terms when parsing out the specific parts of the URI. Also
per changes I needed to make for #9975 instead of using string concatenation use the `formatted` method.
```suggestion
            if (user != null && passw != null && !uri.contains("@")) {
                final String scheme = StringUtils.substringBefore(uri, "://");
                final String hostPortPath = StringUtils.substringAfter(uri, "://");
                uri = "%s://%s:%s@%s".formatted(scheme, user, passw, hostPortPath); 
            }
```

exceptionfactory (changes requested):
Thanks for working on this issue @pvillard31.

On closer review, I'm not sure that adding the username and password to the URI is the best solution.

The MongoCredential has a property for authentication mechanism, which is currently unset. Instead for appending the username and password, what about parsing the connection string for the authMechanism and using that if found? Would that address the issue?

pvillard31:
> Thanks for working on this issue @pvillard31.
> 
> On closer review, I'm not sure that adding the username and password to the URI is the best solution.
> 
> The MongoCredential has a property for authentication mechanism, which is currently unset. Instead for appending the username and password, what about parsing the connection string for the authMechanism and using that if found? Would that address the issue?

Hey @exceptionfactory - AFAICT this would not help. The exception is currently raised when calling

````java
final ConnectionString cs = new ConnectionString(uri);
````

While it is possible to create the `MongoCredential` with a specific authentication method, we still need to create the `ConnectionString` and give it to the client builder. Or are you suggesting something else?

**Comment by exceptionfactory** [comment]
> Hey @exceptionfactory - AFAICT this would not help. The exception is currently raised when calling
> 
> ```java
> final ConnectionString cs = new ConnectionString(uri);
> ```
> 
> While it is possible to create the `MongoCredential` with a specific authentication method, we still need to create the `ConnectionString` and give it to the client builder. Or are you suggesting something else?

Thanks for the reply. I am suggesting to parse and _remove_ the `authMechanism` parameter from the input URI string. Walking through the `ConnectionString` class, it appears the presence of that parameter triggers looking for a username and password, so seems like remove it from the input URI string would resolve the issue, without adding the username and password to the URI.

File: nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java [hunk 3]
```diff
@@ -96,25 +101,52 @@ protected MongoClient createClient(ConfigurationContext context, MongoClient exi
...
+                // we extract all specified parameters in the URI
+                final Map<String, String> result = new LinkedHashMap<>();
+                for (String param : uri.split("&")) {
+                    String[] pair = param.split("=", 2);
+                    String key = URLDecoder.decode(pair[0], StandardCharsets.UTF_8).toLowerCase();
+                    String value = pair.length > 1 ? URLDecoder.decode(pair[1], StandardCharsets.UTF_8) : "";
+                    result.put(key, value);
+                }
+
+                uri = uri.replaceAll("(?i)authmechanism=[^&]*&?", "");
```
exceptionfactory: Instead of parsing the entire query string into a Map, it seems like it would be cleaner to define the regular expression pattern that can be used to both remove the value from the URI and find to the value of `authMechanism`. This could also be used in the conditional itself, instead of the `containsIgnoreCase()` method.

nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java
```diff
@@ -96,25 +101,52 @@ protected MongoClient createClient(ConfigurationContext context, MongoClient exi
...
+                    case PLAIN -> builder.credential(MongoCredential.createPlainCredential(user, database, passw.toCharArray()));
+                    default -> throw new IllegalArgumentException("Unsupported authentication mechanism with username and password: " + mechanism);
+                }
+
+                builder.applyConnectionString(cs);
+            } else {
+                final ConnectionString cs = new ConnectionString(uri);
+
+                if (user != null && passw != null) {
+                    final String database = cs.getDatabase() == null ? "admin" : cs.getDatabase();
```
exceptionfactory: Given the special nature of this `admin` database case, it would be best to put it in a single place, outside of the conditional, as opposed to using the same line in both conditional cases.

nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java
```diff
@@ -96,25 +101,52 @@ protected MongoClient createClient(ConfigurationContext context, MongoClient exi
...
+                    final MongoCredential credential = MongoCredential.createCredential(user, database, passw.toCharArray());
+                    builder.credential(credential);
+                }
+
+                builder.applyConnectionString(cs);
```
exceptionfactory: This call can be moved outside the conditional, using the `final` value of the `ConnectionString` object from either branch.

exceptionfactory (changes requested):
Thanks for working on the revised approach @pvillard31, this aligns more closely with what I had in mind. I recommended some adjustments to the implementation, and then I think this should be ready to go.

pvillard31:
Thanks @exceptionfactory ! I pushed another commit to follow your recommendations

nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java
```diff
@@ -96,27 +99,53 @@ protected MongoClient createClient(ConfigurationContext context, MongoClient exi
...
             final String user = context.getProperty(DB_USER).evaluateAttributeExpressions().getValue();
             final String passw = context.getProperty(DB_PASSWORD).evaluateAttributeExpressions().getValue();
 
             final MongoClientSettings.Builder builder = MongoClientSettings.builder();
-            final ConnectionString cs = new ConnectionString(uri);
 
-            if (user != null && passw != null) {
-                final String database = cs.getDatabase() == null ? "admin" : cs.getDatabase();
+            // Regex to find authMechanism value (case-insensitive)
+            final Pattern authMechanismPattern = Pattern.compile("(?i)(?:[?&])authmechanism=([^&]*)");
```
exceptionfactory: This pattern can be defined in a static variable.

exceptionfactory (commented):
Thanks for the update @pvillard31, one more minor adjustment and then this looks ready to go.

nifi-extension-bundles/nifi-mongodb-bundle/nifi-mongodb-services/src/main/java/org/apache/nifi/mongodb/MongoDBControllerService.java
pvillard31: Done

exceptionfactory (approved):
Thanks for working through the feedback and adjusting the implementation @pvillard31, the latest version looks good! +1 merging

PR #10000, opened 2025-06-09
NIFI-14579 Aligned the Use Starting Row strategy to use a similar approach as the Infer Schema strategy and added a new strategy for using the starting row as the schema field names and reading the whole file for determining the schema.

dan-s1:
# Summary

[NIFI-14579](https://issues.apache.org/jira/browse/NIFI-14579)
This PR aims to align the `Infer Schema` and `Use Starting Row` strategies to be both classes which implement the `org.apache.nifi.schema.inference.SchemaInferenceEngine` interface.  It also provides two implementations for `Use Starting Row`,` Standard` and `All.` As before `Use Starting Row` always uses column names in the defined first row (by user) but now the user has an option to either use as before the first 10 rows to infer the schema or all the rows.
The new class created `ExcelHeaderSchemaInference`, implements the `SchemaInferenceEngine<Row>` interface and incorporates all the functionality the `ExcelHeaderSchemaStrategy` class had including the latest changes made in PR [#9975](https://github.com/apache/nifi/pull/9975).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


dan-s1:
@exceptionfactory 
When you have a chance can you please take a look at this PR? I tried to follow your directives on the ticket to have `Use Starting Row` now implement `org.apache.nifi.schema.inference.SchemaInferenceEngine` interface aligning it with `Infer Schema`.  Now there are two options to `Use Starting Row` to infer from the first 10 rows (Standard) or the whole file.
Thanks!

dan-s1:
> Thanks for working on this improvement @dan-s1.
> 
> The general implementation approach is along the lines of the discussion in Jira, but the property values do not quite align.
> 
> The existing `Use Strategy Row` strategy value needs to be retained to avoid breaking changes.
> 
> Instead of introducing two new options with `Standard` and `All`, I was thinking that there would be a new dependent property with selectable values of `Standard` and `All Rows`. The property would be selectable based on existing schema inference strategy options, instead of introducing new ones.

I thought I avoided breaking changes by adding `migrateProperties` in ExcelReader (line 160)

exceptionfactory:
> I thought I avoided breaking changes by adding `migrateProperties` in ExcelReader (line 160)

Sorry for the confusion, yes, you are correct, the `migrateProperties` approach would avoid breaking changes.

With that being said, I think it would be helpful to retain the existing strategy options, but introduce a separate dependent property for clarity of behavior.

dan-s1:
@exceptionfactory I added the property as you suggested. I must admit that actually simplified the code. Thank you for suggesting that.

dan-s1:
@exceptionfactory Please at your earliest convenience review the changes. Thank you!

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StartingRowStrategy.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.apache.nifi.components.AllowableValue;
+import org.apache.nifi.components.DescribedValue;
+
+public enum StartingRowStrategy implements DescribedValue {
+    STANDARD("Use %s rows after the starting row.".formatted(StartingRowStrategy.NUM_ROWS_TO_DETERMINE_TYPES)),
+    ALL("Use all the rows after the starting row.");
+
+    static final int NUM_ROWS_TO_DETERMINE_TYPES = 10; // NOTE: This number is arbitrary.
+    static final AllowableValue USE_STARTING_ROW = new AllowableValue("Use Starting Row", "Use Starting Row",
```
exceptionfactory: This `AllowableValue` does not fit well inside of this `enum`. Creating a subclass of `AllowableValue` as a separate class is one option to share this value.

exceptionfactory (changes requested):
Thanks for the updates @dan-s1, the adjusted approach is on a good track.

I think the name `Starting Row Strategy` for the property name and enumeration is a bit too generic to be helpful. The purpose of the property is to determine the number of rows to read. Something along the lines of `Row Reading Strategy` or `Row Evaluation Strategy` seems better.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/StartingRowStrategy.java
dan-s1: Can I place it as a variable in class `ExcelStartingRowSchemaInference` ? It was in that class before I had renamed the class.

dan-s1:
> Thanks for the updates @dan-s1, the adjusted approach is on a good track.
> 
> I think the name `Starting Row Strategy` for the property name and enumeration is a bit too generic to be helpful. The purpose of the property is to determine the number of rows to read. Something along the lines of `Row Reading Strategy` or `Row Evaluation Strategy` seems better.

Just to clarify are you wanting me rename the property or the enum? The property name has the same name as it did before.

dan-s1:
@exceptionfactory I believe I addressed the concerns you raised. Can you please take a look at the changes I made when you get a chance? Thanks!

exceptionfactory (approved):
Thanks for the work on this @dan-s1, the latest version looks good. Having the alternative `All Rows` strategy should be helpful for additional use cases. +1 merging

PR #10003, opened 2025-06-10
[NIFI-14636] - Consider form pending status as not valid

rfellows:
# Summary

[NIFI-14636](https://issues.apache.org/jira/browse/NIFI-14636)

* added checks for form `pending` status when determining if for submit buttons are disabled or not.


mcgilman:
Will review...

mcgilman (changes requested):
Thanks for the PR @rfellows! I think we may also need this additional condition added to the form in `fetch-parameter-provider-parameters.component.html`.

mcgilman (approved):

PR #10005, opened 2025-06-10
[NIFI-14648] - Skeleton/Spinner in Import from Registry dialog

rfellows:
# Summary

[NIFI-14648](https://issues.apache.org/jira/browse/NIFI-14648)

* Added a skeleton loader to the import from registry dialog while versions are loading.
* Added an error state if loading versions fails


mcgilman:
Reviewing...

PR #10006, opened 2025-06-10
NIFI-9770 Added nifi-cql-bundle to bring support for Cassandra 4.X an…

MikeThomsen:
…d ScyllaDB.

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-assembly/pom.xml
+++ b/nifi-assembly/pom.xml
@@ -521,6 +521,24 @@ language governing permissions and limitations under the License. -->
         </dependency>
         <dependency>
             <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-cql-nar</artifactId>
+            <version>2.5.0-SNAPSHOT</version>
+            <type>nar</type>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-cql-services-api-nar</artifactId>
+            <version>2.5.0-SNAPSHOT</version>
+            <type>nar</type>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-cassandra-session-provider-service-nar</artifactId>
+            <version>2.5.0-SNAPSHOT</version>
+            <type>nar</type>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-registry-nar</artifactId>
             <version>2.5.0-SNAPSHOT</version>
             <type>nar</type>
--- a/nifi-extension-bundles/nifi-cql-bundle/nifi-cql-nar/pom.xml
+++ b/nifi-extension-bundles/nifi-cql-bundle/nifi-cql-nar/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements. See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License. You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.apache.nifi</groupId>
+        <artifactId>nifi-cql-bundle</artifactId>
+        <version>2.5.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>nifi-cql-nar</artifactId>
+    <packaging>nar</packaging>
+
+    <dependencyManagement>
+        <dependencies>
+            <!-- Provided in nifi-cassandra-services-api-nar -->
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <scope>provided</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-cql-services-api-nar</artifactId>
+            <version>2.5.0-SNAPSHOT</version>
+            <type>nar</type>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-cql-processors</artifactId>
+        </dependency>
+    </dependencies>
+</project>
--- a/nifi-extension-bundles/nifi-cql-bundle/nifi-cql-nar/src/main/resources/META-INF/NOTICE
+++ b/nifi-extension-bundles/nifi-cql-bundle/nifi-cql-nar/src/main/resources/META-INF/NOTICE
@@ -1,0 +1,328 @@
+nifi-cassandra-nar
+Copyright 2016-2020 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+******************
+Apache Software License v2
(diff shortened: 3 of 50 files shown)


MikeThomsen:
@exceptionfactory @mattyb149 ScyllaDB aims for 100% compatibility at the interface level on Java drivers, and the basic testcontainers-powered testing I did seems to support it being a drop-in replacement for the Java DataStax driver.

The two "codec" Java files that are in the Cassandra provider package were generated with ChatGPT.

exceptionfactory (changes requested):
Thanks for putting this together @MikeThomsen.

This will take some time to review, but on a quick scan, I observes a number of issues related to Property Descriptor naming, logging, and use of legacy classes where more modern options are available.

The convention issues can be addressed, but one more fundamental concern is the use of the [com.datadata.oss](https://central.sonatype.com/artifact/com.datastax.oss/java-driver-core) driver, which is over a year old. As mentioned on the [DataStax Java Driver page](https://docs.datastax.com/en/developer/java-driver/index.html), the driver is now maintained under the Apache Cassandra project.

nifi-extension-bundles/nifi-cql-bundle/nifi-cql-nar/pom.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
joewitt, line 1: given 'nifi-cql-bundle' is the top level concept it really should say the word 'cassandra' fully written out.  

so 'nifi-cassandra-bundle' at the top level at least.

nifi-extension-bundles/nifi-cql-bundle/nifi-cql-nar/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,328 @@
...
+    This product includes software developed by Coda Hale and Yammer, Inc.
+
+    This product includes code derived from the JSR-166 project (ThreadLocalRandom, Striped64,
+    LongAdder), which was released with the following comments:
+
+        Written by Doug Lea with assistance from members of JCP JSR-166
+        Expert Group and released to the public domain, as explained at
+        http://creativecommons.org/publicdomain/zero/1.0/
+
+  (ASLv2) The Netty Project
```
joewitt, line 105: please look for other exmaples of how we reference netty.  all this text should not be needed

nifi-assembly/pom.xml
```diff
@@ -519,6 +519,24 @@ language governing permissions and limitations under the License. -->
             <version>2.5.0-SNAPSHOT</version>
             <type>nar</type>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-cql-nar</artifactId>
```
joewitt, line 524: I am not in favor of including cassandra by default in our convenenience binary.  It makes great sense to build a great supportable cassandra interface again but it needs to be really well written and should not be bundled by default until we find out more about usage.

mattyb149: For all external system integrations I'm in favor of a (perhaps set of) interface(s) that separate the external system's API/SDK from the NiFi components in order to BYO library so we can try to be future-proof for upcoming versions and such. This is often a lot more difficult and time-consuming than when we used to just accept all contributions that connected to these systems.

Having said that, I don't want to set a negative tone for PR contributions for external system integrations or else everyone will end up with their own private repos and NiFi's bread and butter of universal data distribution will suffer.

nifi-extension-bundles/nifi-cql-bundle/nifi-cql-nar/pom.xml
MikeThomsen: The bundle includes a driver NAR for ScyllaDB support, so I think calling it `nifi-cql-bundle` makes more sense here. That said, it's been a while AFAICT since Cassandra and ScyllaDB support were discussed, so I'm open to splitting this into two separate bundles because we can't predict where Cassandra and ScyllaDB will be in the future.

MikeThomsen:
@exceptionfactory We should be good to go now on the dependency update.

If you, @joewitt  and @mattyb149 could weight in on keeping this as a single bundle vs two (one Cassandra, one ScyllaDB) I'd like to know your thoughts. I see more advantages in having a single, unified CQL bundle, but the disadvantage is that API divergence between the two databases could necessitate breaking compatibility later.

joewitt:
I don't know what CQL is but i'm guessing Cassandra Query Language.

Im not familiar with Cassandra vs ScyllaDB

But I am very familiar with complex dependencies, maintenance, and vulnerability considerations.  This one just makes me more attentive than others because it tends to tick those boxes of concern.

To be clear...I do recognize Cassandra is very popular and an excellent tool.  And the nifi community is better off when we give them a great integration.  I get it.  I just dont have a good enough sense of this contrib in doing that.

I also noted the AI generated looking code and thought that was a potential flag.  I'm not sure we can use that as is but maybe...

exceptionfactory (changes requested):
Thanks for the updating the driver references @MikeThomsen. I plan to take a closer at the dependency tree, as it appears to include a number of libraries.

As far as the code itself, there are several areas that need to be addressed to align with current conventions including the following:

1. Remove use of `displayName()` from Property Descriptors
2. Use human-readable Title Case names for the `name()` in Property Descriptors
3. Define supported Property Descriptors statically using `List.of()`
4. Defined Relationships statically using `Set.of()`
5. Rewrite log messages to avoid use of `due to {}` with an `Exception`, since the `Exception` is not an argument for the log message itself
6. Use an `enum` that implements `DescribedValue` for defining allowable values and read the `enum` from the Context
7. Standardize naming using either `CQL` or `Cql` in some class names. It looks like the driver uses `CqlSession`, although NiFi components use `SQL`, so using `CQL` seems better in places
8. Align Java package names more closely with module names, such as `processors.cql` and `services.scylladb` in applicable modules

If you can work through these items across the board, that would address a number of things and enable focus on the implementation itself.

MikeThomsen:
@joewitt 

> I don't know what CQL is but i'm guessing Cassandra Query Language.

> Im not familiar with Cassandra vs ScyllaDB

TL;DR, yes CQL = Cassandra Query Language. ScyllaDB is an increasingly popular, C++ rewrite of Apache Cassandra that aims for (or seems to aim for) 100% driver API compatibility so it is a drop-in replacement as far as Cassandra users are concerned.

> I also noted the AI generated looking code and thought that was a potential flag. I'm not sure we can use that as is but maybe...

I understand the concern, but when I reviewed the output from ChatGPT the TypeCodec implementations looked like what any reasonable person would write by hand to implement those mappings so I'm not sure there is enough novelty here to raise red flags. That said, I can try to do some googling on Cassandra to see if OpenAI was taking liberties with someone else's code.

MikeThomsen:
@exceptionfactory 

> If you can work through these items across the board, that would address a number of things and enable focus on the implementation itself.

Yeah. Understood. Thanks for this. FWIW, the implementation is directly inspired by how the graph bundle generally handles the interaction between its controller services and processors (since Cypher and Gremin are so radically different as protocols)

nifi-extension-bundles/nifi-cql-bundle/nifi-cql-services-api/src/main/java/org/apache/nifi/service/cql/api/CQLExecutionService.java
```diff
@@ -0,0 +1,156 @@
...
+
+    PropertyDescriptor PROP_SSL_CONTEXT_SERVICE = new PropertyDescriptor.Builder()
+            .name("SSL Context Service")
+            .description("The SSL Context Service used to provide client certificate information for TLS/SSL "
+                    + "connections.")
+            .required(false)
+            .identifiesControllerService(SSLContextService.class)
+            .build();
+
+    PropertyDescriptor CLIENT_AUTH = new PropertyDescriptor.Builder()
```
exceptionfactory, line 71: This property should not be used for client operations, it is only relevant to services that implement listening sockets.

nifi-extension-bundles/nifi-cql-bundle/nifi-cql-services-api/src/main/java/org/apache/nifi/service/cql/api/CQLExecutionService.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.service.cql.api.exception.QueryFailureException;
+import org.apache.nifi.ssl.SSLContextService;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public interface CQLExecutionService extends ControllerService {
+    PropertyDescriptor CONTACT_POINTS = new PropertyDescriptor.Builder()
```
exceptionfactory, line 34: These Property Descriptors should not be included on the interface. The Controller Service interface defines contract behavior, but the Property Descriptors are specific to a given implementation, and should never be included on a Controller Service interface definition.

PR #10011, opened 2025-06-11
NIFI-14620 Correct toDate EL Function handling of time zone

exceptionfactory:
# Summary

[NIFI-14620](https://issues.apache.org/jira/browse/NIFI-14620) Corrects the behavior of the `toDate()` Expression Language function when parsing values using the optional second argument containing a time zone identifier.

Prior to changes in [NIFI-8161](https://issues.apache.org/jira/browse/NIFI-8161), the `toDate` function used the [SimpleDateFormat.setTimeZone()](https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html#setTimeZone-java.util.TimeZone-) method to set the Time Zone for string parsing.

With the migration to [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html) in NiFi 2.0.0-M1, the [parseBest()](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#parseBest-java.lang.CharSequence-java.time.temporal.TemporalQuery...-) method covered various types of results, but effectively ignored the Time Zone applied to the `DateTimeFormatter` using the `withTimeZone()` method.

Changes in this pull request update the `FormatUtils.parseInstant()` method, used only for Expression Language functions, to use the [DateTimeFormatter.parse()](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#parse-java.lang.CharSequence-) method. The utility method then queries the parsed `TemporalAccessor` for available fields, first constructing a `LocalDateTime` and then accounting for both the `ZoneId` passed as an argument and the `ZoneOffset` when present in the input string.

The updated approach passes existing tests for `toDate()` with and without time zone, and also passes a new unit test the compares a `java.util.Date` result with an expected instance constructed to match.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-expression-language/src/test/java/org/apache/nifi/attribute/expression/language/TestQuery.java
+++ b/nifi-commons/nifi-expression-language/src/test/java/org/apache/nifi/attribute/expression/language/TestQuery.java
@@ -1012,6 +1012,29 @@ public class TestQuery {
     }
 
     @Test
+    public void testToDateAdjustedWithTimeZone() {
+        final String dateFormat = "yyyy-MM-dd";
+        final String created = "2025-06-01";
+        final String timeZoneId = "UTC";
+        final TimeZone timeZone = TimeZone.getTimeZone(timeZoneId);
+
+        // Build Calendar for java.util.Date to match expected result of toDate() function
+        final Calendar calendar = Calendar.getInstance();
+        calendar.clear();
+        calendar.set(Calendar.YEAR, 2025);
+        calendar.set(Calendar.MONTH, Calendar.JUNE);
+        calendar.set(Calendar.DAY_OF_MONTH, 1);
+        calendar.setTimeZone(timeZone);
+
+        // Expected Date with System Default Time Zone and hours adjusted based on Time Zone specified
+        final Date expected = calendar.getTime();
+
+        final String expression = "${created:toDate('%s', '%s')}".formatted(dateFormat, timeZoneId);
+        final Map<String, String> attributes = Map.of("created", created);
+        verifyEquals(expression, attributes, expected);
+    }
+
+    @Test
     public void testInstantConversion() {
         final Map<String, String> attributes = new HashMap<>();
         attributes.put("instant", "1403620278642");
@@ -2591,6 +2614,8 @@ public class TestQuery {
             }
         } else if (expectedResult instanceof Boolean) {
             assertEquals(ResultType.BOOLEAN, result.getResultType());
+        } else if (expectedResult instanceof Date) {
+            assertEquals(ResultType.DATE, result.getResultType());
         } else {
             assertEquals(ResultType.STRING, result.getResultType());
         }
--- a/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java
+++ b/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java
@@ -21,15 +21,15 @@ import org.apache.nifi.time.DurationFormat;
 
 import java.text.NumberFormat;
 import java.time.Instant;
-import java.time.LocalDate;
 import java.time.LocalDateTime;
-import java.time.LocalTime;
-import java.time.Year;
-import java.time.YearMonth;
+import java.time.OffsetDateTime;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
+import java.time.temporal.ChronoField;
 import java.time.temporal.TemporalAccessor;
+import java.time.temporal.TemporalQueries;
 import java.util.Locale;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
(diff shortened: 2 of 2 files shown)


pvillard31 (approved):
Changes are looking good and the new unit test is a useful addition. Thanks for the changes @exceptionfactory

dan-s1:
@exceptionfactory 
Thank you for the contribution. I am still noting two things which seem to be off.
I ran your fix with parameter date "20250601 and the various time zones the user had mentioned in the ticket and I get the following. Please note the time zone of my computer is UTC. 

| Time Zone | toDate result| Expected Result|
|-----------------|--------------------|------------------------|
| Europe/Berlin            | Sat May 31 22:00:00 UTC 2025|Sun Jun 01 02:00:00 CEST 2025|
| UTC            | Sun Jun 01 00:00:00 UTC 2025|Sun Jun 01 00:00:00 UTC 2025|
| Europe/Kiev | Sat May 31 21:00:00 UTC 2025|Sun Jun 01 03:00:00 EEST 2025|
| America/Detroit | Sun Jun 01 04:00:00 UTC 2025|Sat May 31 20:00:00 EDT 2025|

One thing to note is all the examples have UTC as the timezone. One of the user's issues was to have the correct timezone in the timestamp and that still is not happening. Furthermore all the later time zones than UTC such as Europe/Berlin and Europe/Kiev should have the same date Sun Jun 01 and later times 02:00:00 and  03:00:00. Also timezone America/Detroit  should be an earlier date Sat May 31 and time 20:00:00.

 I just wanted to add, I produced the output of the above table with the following code in `TestQuery`

```
@ParameterizedTest
@ValueSource(strings = {"Europe/Berlin", "UTC", "Europe/Kiev", "America/Detroit"})
    void testToDate(String timezone) {
        final Query query = Query.compile("${dateTime:toDate('yyyy-MM-dd', 'timezone')}".replaceFirst("timezone", timezone));
        final Map<String, String> attributes = new HashMap<>();
        attributes.put("dateTime", "2025-06-01");

        final QueryResult<?> result = query.evaluate(new StandardEvaluationContext(attributes));
        System.out.printf("%s %s\n", timezone, result.getValue());
}
```

exceptionfactory:
Thanks for following up on this issue @dan-s1.

There are several things to consider when evaluating the `java.util.Date` behavior.

The `toDate()` function returns a `java.util.Date` instance, and the `Date.toString()` representation is what is shown in the table. Internally, the class represents information using a `long` field named `fastTime` and and internal calendar with associated zone information. The zone information is what appears when calling `Date.toString()`.

The Time Zone argument to the `toDate()` function influences the behavior of the date parsing, previously `SimpleDateFormat` and now `DateTimeFormatter`. In both cases, however, the returned `java.util.Date` will display the local system time zone when using the `toString()` method.

As you described in the Jira issue, formatting `java.util.Date` can be accomplished using the `format()` EL function. The `toInstant()` function is properly more understandable in terms of actual behavior, because the resolved value is always computed according to UTC, whereas `toDate()` is more relative to local system time.

Glad to consider the issue further in Jira, but some of this is running into the issues inherent in `java.util.Date`.

dan-s1:
@exceptionfactory 
Thanks for responding. It seems from the user in 1.x the correct timezone was displayed when calling toDate with a timezone but now its only the local one.
Should we be changing the documentation per the user's question?

exceptionfactory:
Yes, it is probably worth adding some clarification to the `toDate()` function, indicating that it influences parsing behavior only.

dan-s1:
I assume in a different ticket, right?

exceptionfactory:
> I assume in a different ticket, right?

Yes, a separate ticket for tracking the documentation changes, linked to this one, would be helpful.

dan-s1:
@exceptionfactory 
I do not mean to belabor this but I did not see the answer to my question earlier why the date and times are skewed for the other time zones. I understand why the UTC is present but shouldn't each of the time zones have the correct date and time per their timezone? Per you explanation I would still expect to see the following

| Time Zone | toDate result| Expected Result|
|-----------------|--------------------|------------------------|
| Europe/Berlin            | Sat May 31 22:00:00 UTC 2025|Sun Jun 01 02:00:00 UTC 2025|
| UTC            | Sun Jun 01 00:00:00 UTC 2025|Sun Jun 01 00:00:00 UTC 2025|
| Europe/Kiev | Sat May 31 21:00:00 UTC 2025|Sun Jun 01 03:00:00 UTC 2025|
| America/Detroit | Sun Jun 01 04:00:00 UTC 2025|Sat May 31 20:00:00 UTC 2025|

exceptionfactory:
> @exceptionfactory I do not mean to belabor this but I did not see the answer to my question earlier why the date and times are skewed for the other time zones. I understand why the UTC is present but shouldn't each of the time zones have the correct date and time per their timezone? Per you explanation I would still expect to see the following
> Time Zone 	toDate result 	Expected Result
> Europe/Berlin 	Sat May 31 22:00:00 UTC 2025 	Sun Jun 01 02:00:00 UTC 2025
> UTC 	Sun Jun 01 00:00:00 UTC 2025 	Sun Jun 01 00:00:00 UTC 2025
> Europe/Kiev 	Sat May 31 21:00:00 UTC 2025 	Sun Jun 01 03:00:00 UTC 2025
> America/Detroit 	Sun Jun 01 04:00:00 UTC 2025 	Sat May 31 20:00:00 UTC 2025


The apparent skew is the result of interpreting the input time zone offset relative to the system local time zone. This gets back to the behavior of `java.util.Date`.

During parsing, the `parseToInstant` method computes a `LocalDateTime`, then adjusts it based on the input Time Zone argument. After that adjustment, `Date.from()` takes the epoch time in milliseconds from the parsed `java.time.Instant`. Due to the way that `java.util.Date` always computes `toString()` relative to the system default Time Zone, the string representation is also adjusted.

This gets back to the fundamental problems with `java.util.Date` that prompted to move to more precise classes in the `java.time` package.

PR #10023, opened 2025-06-16
NIFI-14638: Allowing Stop Version Control even when there are no Regi…

mcgilman:
…stry Clients.


rfellows:
Will review...

rfellows (approved):
Thanks for the updates, @mcgilman 👍

PR #10030, opened 2025-06-20
NIFI-14671 NIFI-14675 record path long handling improvements

awelless:
# Summary

[NIFI-14671](https://issues.apache.org/jira/browse/NIFI-14671)
[NIFI-14675](https://issues.apache.org/jira/browse/NIFI-14675)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


nifi-commons/nifi-record-path/src/main/java/org/apache/nifi/record/path/filter/NumericBinaryOperatorFilter.java
```diff
@@ -65,10 +65,10 @@ protected boolean test(final FieldValue fieldValue, final Object rhsValue) {
         }
 
         final String fieldName = fieldValue.getField() == null ? "<Anonymous Inner Field>" : fieldValue.getField().getFieldName();
-        final Number lhsNumber = lhsLongCompatible ? DataTypeUtils.toLong(value, fieldName) : DataTypeUtils.toDouble(value, fieldName);
```
awelless, line 68: The result of this expression was always upcasted to`Double`, even when the first branch was chosen.
To remediate this we should separate conversion to longs and doubles.

exceptionfactory:
Thanks for the latest update, I agree that the existing behavior of returning an `int` should be reserved when the number is within the range of an `Integer`.

awelless:
@exceptionfactory.
After reviewing the tests I decided to preserve the original behavior in `RecordPathCompiler`. Since these value can land into a record and impact a record schema, some processors downstream can, theoretically, expect the `INT` type specifically.

So, a literal number is being saved as `long` only when it doesn't fit into `int`.

nifi-commons/nifi-record-path/src/main/java/org/apache/nifi/record/path/filter/NumericComparisonFilter.java
```diff
@@ -0,0 +1,44 @@
...
+        final int comparisonResult = isLongCompatible(lhsNumber) && isLongCompatible(rhsNumber)
+                ? Long.compare(lhsNumber.longValue(), rhsNumber.longValue())
+                : Double.compare(lhsNumber.doubleValue(), rhsNumber.doubleValue());
+
+        return verifyComparisonResult(comparisonResult);
+    }
+
+    protected abstract boolean verifyComparisonResult(final int comparisonResult);
+
+    private boolean isLongCompatible(final Number value) {
```
exceptionfactory: This check is duplicative of existing checks in the parent class, which already determines whether the numbers are compatible with the `Long` type.

nifi-commons/nifi-record-path/src/main/java/org/apache/nifi/record/path/filter/NumericBinaryOperatorFilter.java
```diff
@@ -65,10 +65,10 @@ protected boolean test(final FieldValue fieldValue, final Object rhsValue) {
...
-        final Number rhsNumber = rhsLongCompatible ? DataTypeUtils.toLong(rhsValue, fieldName) : DataTypeUtils.toDouble(rhsValue, fieldName);
-        return compare(lhsNumber, rhsNumber);
+        return lhsLongCompatible && rhsLongCompatible
+                ? test(DataTypeUtils.toLong(value, fieldName), DataTypeUtils.toLong(rhsValue, fieldName))
+                : test(DataTypeUtils.toDouble(value, fieldName), DataTypeUtils.toDouble(rhsValue, fieldName));
     }
 
-    protected abstract boolean compare(final Number lhsNumber, final Number rhsNumber);
+    protected abstract boolean test(final Number lhsNumber, final Number rhsNumber);
```
exceptionfactory: Instead of passing the untyped `Number`, separate concrete methods should be defined, one for `long` types and one for `double` types. Those separate implementations can then call the abstract `verifyComparisonResult`.

exceptionfactory (changes requested):
Thanks for the updates to maintain compatibility @awelless.

On further review, the new `NumberComparisonFilter` introduces a duplicative type check for the `Number` arguments. As noted in comments, the approach should be refactored to call type-specific methods avoid the `instanceof` checks.

nifi-commons/nifi-record-path/src/main/java/org/apache/nifi/record/path/filter/NumericBinaryOperatorFilter.java
awelless: The concern was that not all subclasses of `NumericBinaryOperatorFilter` would care about the specific types. But since there are only comparison operators for now, we should be good with introducing 2 separate methods for longs and doubles.
Changed.

PR #10031, opened 2025-06-20
NIFI-10752: Implement GetCouchbase and PutCouchbase processors with SDK3

mark-bathori:
# Summary

This PR introduces an architectural pattern similar to the Kafka bundle. Its main idea is to decouple top-level Couchbase processors and services from specific SDK dependencies. This is achieved by introducing a Controller Service layer that directly implements the service API methods based on a selected SDK. The advantage of this approach is that if a new Couchbase SDK is released, we avoid duplicating existing top-level processors and services. Instead, a new service bundle can be created to implement the new SDK, and components can simply be reconfigured to use the updated service.

[NIFI-10752](https://issues.apache.org/jira/browse/NIFI-10752)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/pom.xml
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/pom.xml
@@ -54,7 +54,7 @@
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>couchbase</artifactId>
-            <version>1.21.3</version>
+            <version>${testcontainers.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
@@ -27,10 +27,10 @@ import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.apache.nifi.services.couchbase.CouchbaseClient;
+import org.apache.nifi.services.couchbase.exception.ExceptionCategory;
 import org.apache.nifi.services.couchbase.utils.CouchbaseContext;
 import org.apache.nifi.services.couchbase.utils.DocumentType;
 import org.apache.nifi.services.couchbase.CouchbaseConnectionService;
-import org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler;
 import org.apache.nifi.services.couchbase.exception.CouchbaseException;
 import org.apache.nifi.stream.io.StreamUtils;
 
@@ -177,8 +177,8 @@ public abstract class AbstractCouchbaseProcessor extends AbstractProcessor {
                                             ComponentLog logger, FlowFile flowFile, CouchbaseException e, String errorMessage) {
         final Throwable throwable = (e.getCause() != null) ? e.getCause() : e;
         logger.error(errorMessage, throwable);
-        final CouchbaseErrorHandler.ErrorHandlingStrategy strategy = couchbaseClient.getErrorHandler().getStrategy(throwable);
-        switch (strategy) {
+        final ExceptionCategory exceptionCategory = couchbaseClient.getExceptionCategory(throwable);
+        switch (exceptionCategory) {
             case ROLLBACK -> session.rollback();
             case FAILURE -> {
                 context.yield();
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseClient.java
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseClient.java
@@ -16,8 +16,8 @@
  */
 package org.apache.nifi.services.couchbase;
 
-import org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler;
 import org.apache.nifi.services.couchbase.exception.CouchbaseException;
+import org.apache.nifi.services.couchbase.exception.ExceptionCategory;
 import org.apache.nifi.services.couchbase.utils.CouchbaseGetResult;
 import org.apache.nifi.services.couchbase.utils.CouchbaseUpsertResult;
 
@@ -27,5 +27,5 @@ public interface CouchbaseClient {
 
     CouchbaseUpsertResult upsertDocument(String documentId, byte[] content) throws CouchbaseException;
 
-    CouchbaseErrorHandler getErrorHandler();
+    ExceptionCategory getExceptionCategory(Throwable throwable);
 }
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/utils/DocumentType.java
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/utils/DocumentType.java
@@ -21,8 +21,8 @@ package org.apache.nifi.services.couchbase.utils;
  */
 public enum DocumentType {
 
-    JSON( "application/json"),
-    BINARY( "application/octet-stream");
+    JSON("application/json"),
+    BINARY("application/octet-stream");
 
     private final String mimeType;
 
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/pom.xml
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/pom.xml
@@ -27,7 +27,7 @@
     <packaging>jar</packaging>
 
     <properties>
-        <couchbase.version>3.9.0</couchbase.version>
+        <couchbase.version>3.9.1</couchbase.version>
     </properties>
 
     <dependencies>
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseClient.java
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseClient.java
@@ -51,8 +51,8 @@ import com.couchbase.client.java.kv.MutationResult;
 import com.couchbase.client.java.kv.PersistTo;
 import com.couchbase.client.java.kv.ReplicateTo;
 import com.couchbase.client.java.kv.UpsertOptions;
-import org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler;
 import org.apache.nifi.services.couchbase.exception.CouchbaseException;
+import org.apache.nifi.services.couchbase.exception.ExceptionCategory;
 import org.apache.nifi.services.couchbase.utils.CouchbaseGetResult;
 import org.apache.nifi.services.couchbase.utils.CouchbaseUpsertResult;
 import org.apache.nifi.services.couchbase.utils.DocumentType;
@@ -63,9 +63,9 @@ import java.util.Map;
 import java.util.function.Predicate;
 
 import static java.util.Map.entry;
-import static org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler.ErrorHandlingStrategy.FAILURE;
-import static org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler.ErrorHandlingStrategy.RETRY;
-import static org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler.ErrorHandlingStrategy.ROLLBACK;
+import static org.apache.nifi.services.couchbase.exception.ExceptionCategory.FAILURE;
+import static org.apache.nifi.services.couchbase.exception.ExceptionCategory.RETRY;
+import static org.apache.nifi.services.couchbase.exception.ExceptionCategory.ROLLBACK;
 
 class StandardCouchbaseClient implements CouchbaseClient {
 
@@ -73,6 +73,36 @@ class StandardCouchbaseClient implements CouchbaseClient {
     private final DocumentType documentType;
     private final PersistTo persistTo;
     private final ReplicateTo replicateTo;
+
+    private final Map<Class<? extends Exception>, ExceptionCategory> exceptionMapping = Map.ofEntries(
+            entry(AuthenticationFailureException.class, ROLLBACK),
+            entry(BucketNotFoundException.class, ROLLBACK),
+            entry(ScopeNotFoundException.class, ROLLBACK),
+            entry(CollectionNotFoundException.class, ROLLBACK),
+            entry(DatasetNotFoundException.class, ROLLBACK),
+            entry(ServiceNotAvailableException.class, ROLLBACK),
+            entry(FeatureNotAvailableException.class, ROLLBACK),
+            entry(InvalidArgumentException.class, ROLLBACK),
+            entry(ConfigException.class, ROLLBACK),
+
+            entry(RequestCanceledException.class, RETRY),
+            entry(TemporaryFailureException.class, RETRY),
+            entry(DurableWriteInProgressException.class, RETRY),
+            entry(DurableWriteReCommitInProgressException.class, RETRY),
+            entry(ServerOutOfMemoryException.class, RETRY),
+            entry(DocumentLockedException.class, RETRY),
+            entry(DocumentMutationLostException.class, RETRY),
+            entry(DocumentUnretrievableException.class, RETRY),
+            entry(DocumentNotLockedException.class, RETRY),
+
+            entry(DocumentNotFoundException.class, FAILURE),
+            entry(DocumentExistsException.class, FAILURE),
+            entry(CasMismatchException.class, FAILURE),
+            entry(DecodingFailureException.class, FAILURE),
+            entry(PathNotFoundException.class, FAILURE),
+            entry(PathExistsException.class, FAILURE),
+            entry(ValueTooLargeException.class, FAILURE)
+    );
 
     StandardCouchbaseClient(Collection collection, DocumentType documentType, PersistTo persistTo, ReplicateTo replicateTo) {
         this.collection = collection;
@@ -88,7 +118,7 @@ class StandardCouchbaseClient implements CouchbaseClient {
 
             return new CouchbaseGetResult(result.contentAsBytes(), result.cas());
         } catch (Exception e) {
-            throw new CouchbaseException("An error occurred while getting the document from Couchbase", e);
+            throw new CouchbaseException("Failed to get document [%s] from Couchbase".formatted(documentId), e);
         }
     }
 
@@ -107,43 +137,13 @@ class StandardCouchbaseClient implements CouchbaseClient {
 
             return new CouchbaseUpsertResult(result.cas());
         } catch (Exception e) {
-            throw new CouchbaseException("An error occurred while upserting the document in Couchbase", e);
+            throw new CouchbaseException("Failed to upsert document [%s] in Couchbase".formatted(documentId), e);
         }
     }
 
     @Override
-    public CouchbaseErrorHandler getErrorHandler() {
-        final Map<Class<? extends Exception>, CouchbaseErrorHandler.ErrorHandlingStrategy> exceptionMapping = Map.ofEntries(
-                entry(AuthenticationFailureException.class, ROLLBACK),
-                entry(BucketNotFoundException.class, ROLLBACK),
-                entry(ScopeNotFoundException.class, ROLLBACK),
-                entry(CollectionNotFoundException.class, ROLLBACK),
-                entry(DatasetNotFoundException.class, ROLLBACK),
-                entry(ServiceNotAvailableException.class, ROLLBACK),
-                entry(FeatureNotAvailableException.class, ROLLBACK),
-                entry(InvalidArgumentException.class, ROLLBACK),
-                entry(ConfigException.class, ROLLBACK),
-
-                entry(RequestCanceledException.class, RETRY),
-                entry(TemporaryFailureException.class, RETRY),
-                entry(DurableWriteInProgressException.class, RETRY),
-                entry(DurableWriteReCommitInProgressException.class, RETRY),
-                entry(ServerOutOfMemoryException.class, RETRY),
-                entry(DocumentLockedException.class, RETRY),
-                entry(DocumentMutationLostException.class, RETRY),
-                entry(DocumentUnretrievableException.class, RETRY),
-                entry(DocumentNotLockedException.class, RETRY),
-
-                entry(DocumentNotFoundException.class, FAILURE),
-                entry(DocumentExistsException.class, FAILURE),
-                entry(CasMismatchException.class, FAILURE),
-                entry(DecodingFailureException.class, FAILURE),
-                entry(PathNotFoundException.class, FAILURE),
-                entry(PathExistsException.class, FAILURE),
-                entry(ValueTooLargeException.class, FAILURE)
-        );
-
-        return new CouchbaseErrorHandler(exceptionMapping);
+    public ExceptionCategory getExceptionCategory(Throwable throwable) {
+        return exceptionMapping.getOrDefault(throwable.getClass(), FAILURE);
     }
 
     private Transcoder getTranscoder(DocumentType documentType) {
--- a/nifi-assembly/pom.xml
+++ b/nifi-assembly/pom.xml
@@ -885,24 +885,6 @@ language governing permissions and limitations under the License. -->
             <version>2.7.0-SNAPSHOT</version>
             <type>nar</type>
         </dependency>
-        <dependency>
-            <groupId>org.apache.nifi</groupId>
-            <artifactId>nifi-couchbase-services-api-nar</artifactId>
-            <version>2.7.0-SNAPSHOT</version>
-            <type>nar</type>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.nifi</groupId>
-            <artifactId>nifi-couchbase-standard-services-nar</artifactId>
-            <version>2.7.0-SNAPSHOT</version>
-            <type>nar</type>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.nifi</groupId>
-            <artifactId>nifi-couchbase-nar</artifactId>
-            <version>2.7.0-SNAPSHOT</version>
-            <type>nar</type>
-        </dependency>
         <!-- AspectJ library needed by the Java Agent used for native library loading (see bootstrap.conf) -->
         <dependency>
             <groupId>org.aspectj</groupId>
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/test/java/org/apache/nifi/processors/couchbase/GetCouchbaseTest.java
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/test/java/org/apache/nifi/processors/couchbase/GetCouchbaseTest.java
@@ -21,8 +21,8 @@ import org.apache.nifi.provenance.ProvenanceEventType;
 import org.apache.nifi.reporting.InitializationException;
 import org.apache.nifi.services.couchbase.CouchbaseClient;
 import org.apache.nifi.services.couchbase.CouchbaseConnectionService;
-import org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler;
 import org.apache.nifi.services.couchbase.exception.CouchbaseException;
+import org.apache.nifi.services.couchbase.exception.ExceptionCategory;
 import org.apache.nifi.services.couchbase.utils.CouchbaseGetResult;
 import org.apache.nifi.util.MockFlowFile;
 import org.apache.nifi.util.StringUtils;
@@ -33,7 +33,6 @@ import org.junit.jupiter.api.Test;
 
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -82,7 +81,6 @@ public class GetCouchbaseTest {
         final CouchbaseGetResult result = new CouchbaseGetResult(content.getBytes(), TEST_CAS);
 
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(Collections.emptyMap()));
         when(client.getDocument(anyString())).thenReturn(result);
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
@@ -123,7 +121,6 @@ public class GetCouchbaseTest {
         final CouchbaseGetResult result = new CouchbaseGetResult(content.getBytes(), TEST_CAS);
 
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(Collections.emptyMap()));
         when(client.getDocument(anyString())).thenReturn(result);
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
@@ -166,7 +163,6 @@ public class GetCouchbaseTest {
         final CouchbaseGetResult result = new CouchbaseGetResult(content.getBytes(), TEST_CAS);
 
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(Collections.emptyMap()));
         when(client.getDocument(anyString())).thenReturn(result);
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
@@ -205,7 +201,7 @@ public class GetCouchbaseTest {
     @Test
     public void testWithFailure() throws CouchbaseException, InitializationException {
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(Collections.emptyMap()));
+        when(client.getExceptionCategory(any())).thenReturn(ExceptionCategory.FAILURE);
         when(client.getDocument(anyString())).thenThrow(new CouchbaseException("", new TestCouchbaseException("Test exception")));
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/test/java/org/apache/nifi/processors/couchbase/PutCouchbaseTest.java
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/test/java/org/apache/nifi/processors/couchbase/PutCouchbaseTest.java
@@ -21,8 +21,8 @@ import org.apache.nifi.provenance.ProvenanceEventType;
 import org.apache.nifi.reporting.InitializationException;
 import org.apache.nifi.services.couchbase.CouchbaseClient;
 import org.apache.nifi.services.couchbase.CouchbaseConnectionService;
-import org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler;
 import org.apache.nifi.services.couchbase.exception.CouchbaseException;
+import org.apache.nifi.services.couchbase.exception.ExceptionCategory;
 import org.apache.nifi.services.couchbase.utils.CouchbaseUpsertResult;
 import org.apache.nifi.util.MockFlowFile;
 import org.apache.nifi.util.StringUtils;
@@ -32,7 +32,6 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -75,7 +74,6 @@ public class PutCouchbaseTest {
     @Test
     public void testWithDocumentIdAsFlowFileAttribute() throws CouchbaseException, InitializationException {
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(Collections.emptyMap()));
         when(client.upsertDocument(anyString(), any())).thenReturn(new CouchbaseUpsertResult(TEST_CAS));
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
@@ -118,11 +116,8 @@ public class PutCouchbaseTest {
 
     @Test
     public void testWithFailure() throws CouchbaseException, InitializationException {
-        final Map<Class<? extends Exception>, CouchbaseErrorHandler.ErrorHandlingStrategy> exceptionMapping =
-                Collections.singletonMap(TestCouchbaseException.class, CouchbaseErrorHandler.ErrorHandlingStrategy.FAILURE);
-
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(exceptionMapping));
+        when(client.getExceptionCategory(any())).thenReturn(ExceptionCategory.FAILURE);
         when(client.upsertDocument(anyString(), any())).thenThrow(new CouchbaseException("", new TestCouchbaseException("Test exception")));
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
@@ -144,11 +139,8 @@ public class PutCouchbaseTest {
 
     @Test
     public void testWithRetry() throws CouchbaseException, InitializationException {
-        final Map<Class<? extends Exception>, CouchbaseErrorHandler.ErrorHandlingStrategy> exceptionMapping =
-                Collections.singletonMap(TestCouchbaseException.class, CouchbaseErrorHandler.ErrorHandlingStrategy.RETRY);
-
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(exceptionMapping));
+        when(client.getExceptionCategory(any())).thenReturn(ExceptionCategory.RETRY);
         when(client.upsertDocument(anyString(), any())).thenThrow(new CouchbaseException("", new TestCouchbaseException("Test exception")));
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
@@ -170,11 +162,8 @@ public class PutCouchbaseTest {
 
     @Test
     public void testWithRollback() throws CouchbaseException, InitializationException {
-        final Map<Class<? extends Exception>, CouchbaseErrorHandler.ErrorHandlingStrategy> exceptionMapping =
-                Collections.singletonMap(TestCouchbaseException.class, CouchbaseErrorHandler.ErrorHandlingStrategy.ROLLBACK);
-
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(exceptionMapping));
+        when(client.getExceptionCategory(any())).thenReturn(ExceptionCategory.ROLLBACK);
         when(client.upsertDocument(anyString(), any())).thenThrow(new CouchbaseException("", new TestCouchbaseException("Test exception")));
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
@@ -197,7 +186,7 @@ public class PutCouchbaseTest {
     @Test
     public void testWithUnknownException() throws CouchbaseException, InitializationException {
         final CouchbaseClient client = mock(CouchbaseClient.class);
-        when(client.getErrorHandler()).thenReturn(new CouchbaseErrorHandler(Collections.emptyMap()));
+        when(client.getExceptionCategory(any())).thenReturn(ExceptionCategory.FAILURE);
         when(client.upsertDocument(anyString(), any())).thenThrow(new CouchbaseException("", new TestCouchbaseException("Test exception")));
 
         final CouchbaseConnectionService service = mock(CouchbaseConnectionService.class);
--- a/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/exception/ExceptionCategory.java
+++ b/nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/exception/ExceptionCategory.java
@@ -16,23 +16,8 @@
  */
 package org.apache.nifi.services.couchbase.exception;
 
-import java.util.Map;
-
-import static org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler.ErrorHandlingStrategy.FAILURE;
-
-public class CouchbaseErrorHandler {
-
-    private final Map<Class<? extends Exception>, ErrorHandlingStrategy> exceptionMapping;
-
-    public CouchbaseErrorHandler(Map<Class<? extends Exception>, ErrorHandlingStrategy> exceptionMapping) {
-        this.exceptionMapping = exceptionMapping;
-    }
-
-    public ErrorHandlingStrategy getStrategy(Throwable throwable) {
-        return exceptionMapping.getOrDefault(throwable.getClass(), FAILURE);
-    }
-
-    public enum ErrorHandlingStrategy {
-        ROLLBACK, FAILURE, RETRY
-    }
+public enum ExceptionCategory {
+    ROLLBACK,
+    FAILURE,
+    RETRY
 }


nifi-extension-bundles/nifi-couchbase-bundle/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+        <artifactId>nifi-standard-shared-bom</artifactId>
+        <version>2.5.0-SNAPSHOT</version>
+        <relativePath>../nifi-standard-shared-bundle/nifi-standard-shared-bom</relativePath>
+    </parent>
+
+    <artifactId>nifi-couchbase-bundle</artifactId>
+    <packaging>pom</packaging>
+
+    <properties>
+        <couchbase.version>3.7.8</couchbase.version>
```
pvillard31: Recommend using latest version 3.8.2

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseConnectionService.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.controller.ControllerService;
+import org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler;
+import org.apache.nifi.services.couchbase.exception.CouchbaseException;
+import org.apache.nifi.services.couchbase.utils.CouchbaseGetResult;
+import org.apache.nifi.services.couchbase.utils.CouchbasePutResult;
+
+public interface CouchbaseConnectionService extends ControllerService {
+
+    PropertyDescriptor COUCHBASE_CLUSTER_SERVICE = new PropertyDescriptor.Builder()
```
exceptionfactory: Including Property Descriptors in a service interface definition is generally poor practice, because it is not part of the Controller Service contract. I recommend removing this Descriptor from the interface.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/exception/CouchbaseErrorHandler.java
```diff
@@ -0,0 +1,38 @@
...
+    public CouchbaseErrorHandler(Map<Class<? extends Exception>, ErrorHandlingStrategy> exceptionMapping) {
+        this.exceptionMapping = exceptionMapping;
+    }
+
+    public ErrorHandlingStrategy getStrategy(Exception exception) {
+        return exceptionMapping.getOrDefault(exception.getClass(), Failure);
+    }
+
+    public enum ErrorHandlingStrategy {
+        Rollback, Failure, Retry
```
exceptionfactory: In general, `enum`  values should be uppercased. Is there a reason for this naming convention?

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/exception/CouchbaseErrorHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.services.couchbase.exception;
+
+import java.util.Map;
+
+import static org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler.ErrorHandlingStrategy.Failure;
+
+public class CouchbaseErrorHandler {
```
exceptionfactory: Is there a reason for including this class in the API module? It seems like it should be located in one of the implementation modules, or in a `shared` module if needed.

exceptionfactory (commented):
Thanks for putting together new support for Couchbase @mark-bathori!

I noted a couple initial recommendations, and I would like to take a closer look soon.

One other general note, did you consider creating an integration test using [Testcontainers](https://testcontainers.com/modules/couchbase/)? That could be very useful for verifying behavior against a running instance of Couchbase.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/exception/CouchbaseErrorHandler.java
mark-bathori: The idea here is to create a common class for error handling. SDK-specific implementations will then determine which exceptions are considered fatal, recoverable, or other categories by adding the proper mapping to this class. Based on that, it felt like the proper place for this class was alongside other classes used in the API interfaces.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/exception/CouchbaseErrorHandler.java
mark-bathori: They should be uppercase, I'll fix them.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseConnectionService.java
mark-bathori: Yes it doesn't seem right to put a  Property Descriptors into this interface, the implementations will need to use a cluster service and somehow I wanted to avoid the duplication, will rework this.

turcsanyip:
@mark-bathori Thanks for putting this together!

I checked the components and how to configure them and I would have a suggestion at high level:

I feel a bit complex that two controller services and a processor need to be set up for a flow. Also, the bucket/scope/collection properties are configured in `Couchbase3ConnectionService` so a separate controller service is needed for each target collection. Bucket, scope and collection are lightweight objects so I do not think it is necessary to initialize them once and store them in the CS. The processor can pass the names to the service. Even more, they could come from FF attributes at runtime.

So my idea would be:
- `CouchbaseConnectionService` CS interface (`Couchbase3ConnectionService`implementation) with method:
CouchbaseClient getClient(String bucket, String, scope, String collection, ?)
- CouchbaseClient regular (non-CS) interface (`Couchbase3Client` implementation) with methods `getDocument()`, `putDocument()`, etc.

I believe this fits better to the "Kafka pattern". In my opinion, `CouchbaseClientProviderService` name would be more appropriate for the CS (like in case of `SmbClientProviderService`) but `CouchbaseConnectionService` is also fine (like the Kafka service).

What do you think?

@exceptionfactory @pvillard31 Could you please share your thoughts as well?

mark-bathori:
Thanks @turcsanyip for the review.

The main idea behind the current implementation was to completely separate any Couchbase SDK-specific logic and attributes from the processor level. This is why the bucket, scope, and collection attributes are defined on the controller service. I'm unsure if we can assume these properties will exist in future SDK versions, considering the scope attribute didn't exist in SDK 2. For example, if a new abstraction level were introduced in the next SDK, it would require modifications to the processors, which is something I aimed to avoid.

I agree that configuring the controller service with fixed values for bucket, scope, and collection isn't the best approach. I agree that it would be better if these could get values from FlowFile attributes, so users don't need to configure multiple Controller Services for different bucket, scope, and collection values.

Regarding the double controller service layer, it seemed like a good idea to separate cluster initialization-specific properties from database action-related ones. However, I'm not against merging them into a single service.

exceptionfactory:
Thanks for the discussion @mark-bathori and @turcsanyip.

I agree with @turcsanyip that having the two layers of Controller Services introduces too much complexity from a flow designer perspective. Implementing something along the lines of the Kafka Connection Service approach could work, avoiding the use of the generic `T` return type for a Couchbase Cluster.

mark-bathori:
Thanks @exceptionfactory and @turcsanyip for your review and opinions, I'll rework and update the PR.

chedim:
@mark-bathori , thank you for updating the plugin! One small thing — mentioning SDK version in the bundle name (couchbase-3-...) makes it feel like it is couchbase version, not the SDK version, and couchbase version 3 is very ancient (--we're on major 7 now-- it's already 8)

exceptionfactory:
> @mark-bathori , thank you for updating the plugin! One small thing — mentioning SDK version in the bundle name (couchbase-3-...) makes it feel like it is couchbase version, not the SDK version, and couchbase version 3 is very ancient (we're on major 7 now).

That's a good point @chedim. Clearly the Client SDK version follows a different strategy than the Couchbase server version.

From a flow designer perspective, naming the component according to the SDK version is probably more confusing.

Naming the component with the Couchbase server version could also be unnecessarily limiting.

Decoupling the API and implementation is a good approach, so when it comes to naming, perhaps just calling the implementation `StandardCouchbase...` is the best option. If we come to the point where different implementations require a different version, those could be named differently.

What do you think @mark-bathori and @turcsanyip?

turcsanyip:
@exceptionfactory I agree that `couchbase-3-...` is not ideal due to the mismatch between Couchbase server and client versions and also because we may be able to make an in-place upgrade to version 4 later on (as in case of Kafka).

Not sure how to name the next generation after "standard" but `StandardCouchbase...` looks fine for now.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-3-services-nar/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,39 @@
+nifi-couchbase-services-nar
+Copyright 2014-2023 The Apache Software Foundation
```
turcsanyip: Typo / copy-paste error:
```suggestion
Copyright 2014-2025 The Apache Software Foundation
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-3-services/src/main/java/org/apache/nifi/services/couchbase/Couchbase3Client.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler.ErrorHandlingStrategy.ROLLBACK;
+
+public class Couchbase3Client implements CouchbaseClient {
+
+    private final Collection collection;
+    private final DocumentType documentType;
+    private final PersistTo persistTo;
+    private final ReplicateTo replicateTo;
+
+    public Couchbase3Client(Collection collection, DocumentType documentType, PersistTo persistTo, ReplicateTo replicateTo) {
```
turcsanyip: As the client implementation is not intended to be used directly by the processors, the class and its constructor should be package-private.
```suggestion
class Couchbase3Client implements CouchbaseClient {

    private final Collection collection;
    private final DocumentType documentType;
    private final PersistTo persistTo;
    private final ReplicateTo replicateTo;

   Couchbase3Client(Collection collection, DocumentType documentType, PersistTo persistTo, ReplicateTo replicateTo) {
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-3-services/src/main/java/org/apache/nifi/services/couchbase/Couchbase3Client.java
```diff
@@ -0,0 +1,161 @@
...
+            final GetResult result = collection.get(documentId, GetOptions.getOptions().transcoder(getTranscoder(documentType)));
+
+            return new CouchbaseGetResult(result.contentAsBytes(), result.cas());
+        } catch (Exception e) {
+            throw new CouchbaseException(e);
+        }
+    }
+
+    @Override
+    public long upsertDocument(String documentId, byte[] content) throws CouchbaseException {
```
turcsanyip: Though it is a single value (at the moment), I would consider `CouchbaseUpsertResult`.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-3-services/src/main/java/org/apache/nifi/services/couchbase/Couchbase3ConnectionService.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.List;
+
+@CapabilityDescription("Provides a Couchbase SDK 3 based implementation.")
+@Tags({"nosql", "couchbase", "database", "connection"})
+public class Couchbase3ConnectionService extends AbstractControllerService implements CouchbaseConnectionService {
+
+    public static final PropertyDescriptor CONNECTION_STRING = new PropertyDescriptor.Builder()
+            .name("Connection String")
+            .description("The hostnames or ip addresses of the bootstraping nodes and optional parameters."
+                    + " Syntax) couchbase://node1,node2,nodeN?param1=value1&param2=value2&paramN=valueN")
```
turcsanyip: Typo:
```suggestion
                    + " Syntax: couchbase://node1,node2,nodeN?param1=value1&param2=value2&paramN=valueN")
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+            .description("All FlowFile that fail due to server/cluster availability go to this relationship.")
+            .build();
+
+    private static final List<PropertyDescriptor> PROPERTIES = List.of(
+            DOCUMENT_ID,
+            BUCKET_NAME,
+            SCOPE_NAME,
+            COLLECTION_NAME,
+            DOCUMENT_TYPE,
+            COUCHBASE_CONNECTION_SERVICE
```
turcsanyip: I suggest the following property order:
```suggestion
            COUCHBASE_CONNECTION_SERVICE,
            BUCKET_NAME,
            SCOPE_NAME,
            COLLECTION_NAME,
            DOCUMENT_TYPE,
            DOCUMENT_ID
```
Connection service properties are usually located at the top. Then following bucket-scope-collection hierarchy. And Document descriptors at the end.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/GetCouchbase.java
```diff
@@ -0,0 +1,105 @@
...
+            throw new ProcessException("Document ID is missing. Please provide a valid Document ID through processor property or FlowFile content.");
+        }
+
+        final String bucketName = context.getProperty(BUCKET_NAME).evaluateAttributeExpressions(flowFile).getValue();
+        final String scopeName = context.getProperty(SCOPE_NAME).evaluateAttributeExpressions(flowFile).getValue();
+        final String collectionName = context.getProperty(COLLECTION_NAME).evaluateAttributeExpressions(flowFile).getValue();
+        final DocumentType documentType = DocumentType.valueOf(context.getProperty(DOCUMENT_TYPE).getValue());
+
+        final CouchbaseContext couchbaseContext = new CouchbaseContext(bucketName, scopeName, collectionName, documentType);
+        final CouchbaseClient couchbaseClient = connectionService.getClient(couchbaseContext);
```
turcsanyip: This section is the same as in the Put processor and may be extracted to the abstract class.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/utils/CouchbaseAttributes.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.couchbase.utils;
+
+public class CouchbaseAttributes {
+
+    public static final String BUCKET_ATTRIBUTE = "couchbase.bucket";
```
turcsanyip, line 25: I suggest extracting the descriptions into constants too (as they are the same for Put and Get).
Similar to [GoogleDriveAttibutes](https://github.com/apache/nifi/blob/main/nifi-extension-bundles/nifi-gcp-bundle/nifi-gcp-processors/src/main/java/org/apache/nifi/processors/gcp/drive/GoogleDriveAttributes.java).

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    protected byte[] readFlowFileContent(ProcessSession session, FlowFile flowFile) {
+        final byte[] content = new byte[(int) flowFile.getSize()];
+        session.read(flowFile, in -> StreamUtils.fillBuffer(in, content, true));
+
+        return content;
+    }
+
+    protected String createTransitUrl(CouchbaseContext couchbaseContext, String documentId) {
```
turcsanyip:
```suggestion
    protected String createTransitUri(CouchbaseContext couchbaseContext, String documentId) {
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+        return String.join(".", couchbaseContext.bucket(), couchbaseContext.scope(), couchbaseContext.collection(), documentId);
```
turcsanyip: The URI (or URI-like identifier) of the external system (eg. database host, kafka brokers) is usually added in the Transit Uri, so I think it should be prefixed with the Couchbase connection URL (without the parameters).

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/GetCouchbase.java
mark-bathori: This section wasn't extracted because the bucket, scope, and collection values are also used as attributes on the output FlowFile, so it didn't seem worth to extract only the `CouchbaseContext` and `CouchbaseClient` creation parts. What do you think about this?

mark-bathori:
Thanks, @chedim, @exceptionfactory, and @turcsanyip for the reviews.

I agree that "standard" naming is better than using the SDK version in the module name, as it avoids confusion, so I've renamed the modules accordingly.

I also agree that using the `CAS` attribute would be beneficial. However, as I've observed, `upsert` and `replace` are not equivalent in terms of functionality. They're not simply interchangeable because `replace` doesn't provide insertion capability. It would likely be best to add a separate processor specifically to support the `replace` function with `CAS`.

chedim:
@mark-bathori, thank you for your work on this! please let me know if you need any help with the PR, I can take some tasks, if needed.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/GetCouchbase.java
turcsanyip: The method can return the already existing `CouchbaseContext` object that could be used to retrieve the bucket, collection, etc.

turcsanyip:
> I also agree that using the CAS attribute would be beneficial. However, as I've observed, upsert and replace are not equivalent in terms of functionality. They're not simply interchangeable because replace doesn't provide insertion capability. It would likely be best to add a separate processor specifically to support the replace function with CAS.

`insert()` + `replace()` could be used to achieve the same upsert functionality. It can be the scope of a follow-up jira because it is beyond the basic functionality. As far as I can see it now, I would not add a separate processor because naming could be problematic and there are also other options we may want to support later (e.g. insert only, update only) so creating new processors would be overkill.

exceptionfactory:
On a quick check, the `CertificateAuthenticator.fromKey()` method may be an option, after retrieving the PrivateKey and Certificates from the KeyManager available from `SSLContextProvider`

mark-bathori:
Thanks @exceptionfactory for the comment.

I based the `KeyManagerFactory` creation on a similar approach used in the [MQTT](https://github.com/apache/nifi/blob/main/nifi-extension-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/adapters/HiveMqV5ClientAdapter.java#L201) bundle. I have since checked the `KeyManager` approach you suggested, but unfortunately, it requires providing an alias as parameter to get the `PrivateKey`, which is not suitable for our use case.

The `TrustStore` is currently being created in the same way. Is this approach acceptable for the `TrustStore`, or should it be handled differently?

I think it would be useful to extend the `SSLContextProvider` to create both `KeyManagerFactory` and `TrustManagerFactory` or for now, we could remove keystore-based authentication from the scope of this pull request and revisit it later.

Let me know what you think.

exceptionfactory:
> Thanks @exceptionfactory for the comment.
> 
> I based the `KeyManagerFactory` creation on a similar approach used in the [MQTT](https://github.com/apache/nifi/blob/main/nifi-extension-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/adapters/HiveMqV5ClientAdapter.java#L201) bundle. I have since checked the `KeyManager` approach you suggested, but unfortunately, it requires providing an alias as parameter to get the `PrivateKey`, which is not suitable for our use case.
> 

MQTT is one of the poor examples to which I was referring, so from this perspective, it does not provide a good model to follow.

> The `TrustStore` is currently being created in the same way. Is this approach acceptable for the `TrustStore`, or should it be handled differently?

Yes, the trust store should also be handled differently.

> 
> I think it would be useful to extend the `SSLContextProvider` to create both `KeyManagerFactory` and `TrustManagerFactory` or for now, we could remove keystore-based authentication from the scope of this pull request and revisit it later.

That might be worth considering, but it would need to be a more common usage pattern.

> 
> Let me know what you think.

On further review, the Couchbase Client [Authenticator interface](https://github.com/couchbase/couchbase-jvm-clients/blob/002cdca7341ee529ba2dce15e3f5dbd2ecb9562b/core-io/src/main/java/com/couchbase/client/core/env/Authenticator.java#L67) has an `applyTlsProperties()` method that takes a Netty [SslContextBuilder](https://netty.io/4.1/api/io/netty/handler/ssl/SslContextBuilder.html). The `SslContextBuilder` itself supports being configured directly with a `KeyManager`, instead of other options. Creating a custom implementation of that interface should allow the `SSLContextProvider` Controller Service interface to be used. With this approach, the Couchbase Service can support either the PEM-based implementation, or the JKS/PKCS12 implementation.

exceptionfactory:
I did note the that `Authenticator` interface recommends using available implementations, since the method stability is marked as `Internal`. As a compromise along those lines, extending `CertificateAuthenticator` and just overriding the `applyTlsProperties()` method could also work, as it inherits the rest of the behavior. It is also considering a pull request to the Couchbase driver itself to support the `KeyManager` directly, but as a targeted custom implementation, it seems like the best way forward.

mark-bathori:
Thanks for the clarification on the MQTT solution and for the recommendations. I'll review other approaches for both Keystore and Truststore usage and will update the PR.

mark-bathori:
I checked the `Authenticator` class for custom implementations. However, since its methods are marked as 'Internal' and described as 'unstable,' I think it's better to reach out to Couchbase to request support for `KeyManager` first. Because of this, I've removed the keystore support from this pull request.

@chedim, what do you think about submitting this request to Couchbase?"

chedim:
Hi @mark-bathori!
If I read the code and conversation correctly, I would not recommend extending Authenticator interface. 
It looks that code in this PR configures a TLS connection to a CB cluster by providing our client with list of acceptable issuer or CA certificates, which is [basic channel-level TLS](https://docs.couchbase.com/operator/current/concept-tls.html#basic-tls-configuration) that prevents MiM attacks. The CertificateAuthenticator is used for [mutual tls connection and client authorization via certificate](https://docs.couchbase.com/operator/current/concept-tls.html#tls-client-authentication).

Going back to @exceptionfactory's original concern, it looks like  `org.apache.nifi.ssl.SSLContextProvider#createTrustManager` and `javax.net.ssl.X509TrustManager#getAcceptedIssuers` can be used to obtain a list of CA certificates to configure TLS. 

Upon further inspection, if mTLS is desired then I would recommend using `com.couchbase.client.core.env.CertificateAuthenticator#fromKeyManagerFactory` together with `com.couchbase.client.core.deps.io.netty.handler.ssl.util.KeyManagerFactoryWrapper` (or a custom wrapper) to bridge the `SSLContextProvider` with Couchbase mTLS configuration:

```
SSLContextProvider scp = <...>;
CertificateAuthenticator.fromKeyManagerFactory(() -> new KeyManagerFactoryWrapper(scp.createKeyManager()));
```

mark-bathori:
Thanks, @chedim for the response.

I modified the trust store initialization in my previous commit as you described, and this further confirms that the solution should now be good.

Regarding the mTLS part, the KeyManagerFactoryWrapper seems to be a good solution for our current use case and is much simpler than creating a custom implementation. Thanks for the tip. Apart from that, I would add this within a new PR.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/GetCouchbase.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        final long startNanos = System.nanoTime();
+        final String documentId = context.getProperty(DOCUMENT_ID).isSet()
+                ? context.getProperty(DOCUMENT_ID).evaluateAttributeExpressions(flowFile).getValue()
+                : new String(readFlowFileContent(session, flowFile), StandardCharsets.UTF_8);
+
+        if (isEmpty(documentId)) {
+            throw new ProcessException("Document ID is missing. Please provide a valid Document ID through processor property or FlowFile content.");
+        }
```
turcsanyip: @mark-bathori Thanks for the review changes so far!

I found only one more thing: When the Document ID is missing (not configured either in the property or in the FlowFile content), the FlowFile gets rolled back via throwing ProcessException. I think it should go the Failure instead because the FlowFile is invalid and the processor cannot process it in the next round either.

exceptionfactory:
Thanks for reviewing and tagging me @turcsanyip. I will take a closer look at the authentication strategy, and take one more pass through the implementation structure, then follow up soon.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-ssl-context-service-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.21.3</version>
```
exceptionfactory: This version should be removed as it is managed in the root pom.xml through the testcontainers-bom.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+import static org.apache.nifi.processors.couchbase.utils.CouchbaseAttributes.DEFAULT_BUCKET;
+import static org.apache.nifi.processors.couchbase.utils.CouchbaseAttributes.DEFAULT_COLLECTION;
+import static org.apache.nifi.processors.couchbase.utils.CouchbaseAttributes.DEFAULT_SCOPE;
+import static org.apache.nifi.processors.couchbase.utils.CouchbaseAttributes.DOCUMENT_ID_ATTRIBUTE;
+import static org.apache.nifi.processors.couchbase.utils.CouchbaseAttributes.SCOPE_ATTRIBUTE;
+
+public abstract class AbstractCouchbaseProcessor extends AbstractProcessor {
+
+    public static final PropertyDescriptor DOCUMENT_ID = new PropertyDescriptor.Builder()
+            .name("Document Id")
```
exceptionfactory: Minor naming adjustment:
```suggestion
            .name("Document ID")
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+            .description("Couchbase document id, or an expression to construct the Couchbase document id.")
```
exceptionfactory:
```suggestion
            .description("Couchbase document identifier, or an expression to construct the Couchbase document identifier")
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+    public static final PropertyDescriptor COUCHBASE_CONNECTION_SERVICE = new PropertyDescriptor.Builder()
+            .name("Couchbase Connection Service")
+            .description("A Couchbase Connection Service which manages connections to a Couchbase cluster.")
+            .required(true)
+            .identifiesControllerService(CouchbaseConnectionService.class)
+            .build();
+
+    public static final PropertyDescriptor BUCKET_NAME = new PropertyDescriptor.Builder()
+            .name("Bucket Name")
+            .description("The name of bucket.")
```
exceptionfactory: This description just reiterates the property name. It would be helpful to describe what "Bucket" means in a sentence for clarity.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+            .required(true)
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .defaultValue(DEFAULT_BUCKET)
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .build();
+
+    public static final PropertyDescriptor SCOPE_NAME = new PropertyDescriptor.Builder()
+            .name("Scope Name")
+            .description("The name of scope.")
```
exceptionfactory: Similar to Bucket Name, more detail is needed.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+            .description("The name of collection.")
+            .required(true)
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .defaultValue(DEFAULT_COLLECTION)
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .build();
+
+    public static final PropertyDescriptor DOCUMENT_TYPE = new PropertyDescriptor.Builder()
+            .name("Document Type")
+            .description("The type of content.")
```
exceptionfactory: More detail needed in the description.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+    private static final List<PropertyDescriptor> PROPERTIES = List.of(
+            COUCHBASE_CONNECTION_SERVICE,
+            BUCKET_NAME,
+            SCOPE_NAME,
+            COLLECTION_NAME,
+            DOCUMENT_TYPE,
+            DOCUMENT_ID
+    );
+
+    public static final Set<Relationship> RELATIONSHIPS = Set.of(REL_SUCCESS, REL_FAILURE, REL_RETRY);
```
exceptionfactory: This should be private.
```suggestion
    private static final Set<Relationship> RELATIONSHIPS = Set.of(REL_SUCCESS, REL_FAILURE, REL_RETRY);
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+    protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+        return PROPERTIES;
+    }
+
+    @Override
+    public Set<Relationship> getRelationships() {
+        return RELATIONSHIPS;
+    }
+
+    protected CouchbaseConnectionService connectionService;
```
exceptionfactory: This should be declared before any methods.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+
+    @OnScheduled
+    public void onScheduled(final ProcessContext context) {
+        connectionService = context.getProperty(COUCHBASE_CONNECTION_SERVICE).asControllerService(CouchbaseConnectionService.class);
+    }
+
+    protected byte[] readFlowFileContent(ProcessSession session, FlowFile flowFile) {
```
exceptionfactory, line 146: This could easily lead to heap memory exhaustion with larger files, is this required?

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/GetCouchbase.java
```diff
@@ -0,0 +1,98 @@
...
+    public void onTrigger(ProcessContext context, ProcessSession session) throws ProcessException {
+        FlowFile flowFile = session.get();
+        if (flowFile == null) {
+            return;
+        }
+
+        final long startNanos = System.nanoTime();
+        final String documentId = context.getProperty(DOCUMENT_ID).isSet()
+                ? context.getProperty(DOCUMENT_ID).evaluateAttributeExpressions(flowFile).getValue()
+                : new String(readFlowFileContent(session, flowFile), StandardCharsets.UTF_8);
```
exceptionfactory: Reading the Document ID from FlowFile content does not sound like a feature that should be supported. What is the intended use case? Either way, reading using a FlowFile attribute should be the only supported option.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/PutCouchbase.java
```diff
@@ -0,0 +1,94 @@
...
+        FlowFile flowFile = session.get();
+        if (flowFile == null) {
+            return;
+        }
+
+        final long startNanos = System.nanoTime();
+        final CouchbaseConnectionService connectionService = context.getProperty(COUCHBASE_CONNECTION_SERVICE).asControllerService(CouchbaseConnectionService.class);
+        final String documentId = context.getProperty(DOCUMENT_ID).isSet()
+                ? context.getProperty(DOCUMENT_ID).evaluateAttributeExpressions().getValue()
+                : flowFile.getAttribute(CoreAttributes.UUID.key());
```
exceptionfactory: This should not fall back to the FlowFile `UUID`, instead the property should be required to provide clear behavior.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/test/java/org/apache/nifi/processors/couchbase/TestGetCouchbase.java
```diff
@@ -0,0 +1,223 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class TestGetCouchbase {
```
exceptionfactory: For consistency with Integration Tests, I recommend renaming these unit test classes with `Test` as the suffix:
```suggestion
public class GetCouchbaseTest {
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseClient.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.nifi.services.couchbase.utils.CouchbaseGetResult;
+import org.apache.nifi.services.couchbase.utils.CouchbaseUpsertResult;
+
+public interface CouchbaseClient {
+
+    CouchbaseGetResult getDocument(String documentId) throws CouchbaseException;
+
+    CouchbaseUpsertResult upsertDocument(String documentId, byte[] content) throws CouchbaseException;
+
+    CouchbaseErrorHandler getErrorHandler();
```
exceptionfactory: This does not seem like a method that belongs on the Client interface. Instead, having a shared class for Processors could accomplishing something similar if a common strategy is desired.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseClient.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.nifi.services.couchbase.exception.CouchbaseErrorHandler;
+import org.apache.nifi.services.couchbase.exception.CouchbaseException;
+import org.apache.nifi.services.couchbase.utils.CouchbaseGetResult;
+import org.apache.nifi.services.couchbase.utils.CouchbaseUpsertResult;
+
+public interface CouchbaseClient {
+
+    CouchbaseGetResult getDocument(String documentId) throws CouchbaseException;
+
+    CouchbaseUpsertResult upsertDocument(String documentId, byte[] content) throws CouchbaseException;
```
exceptionfactory, line 28: Can the `content` be completely unstructured, or are there any other formatting requirements?

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/exception/CouchbaseErrorHandler.java
exceptionfactory: Thanks for the reply. On consideration, it seems like this still belongs at the Processor level.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/utils/CouchbaseGetResult.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.services.couchbase.utils;
+
+public record CouchbaseGetResult(byte[] resultContent, long cas) {
```
exceptionfactory, line 24: What is `cas` in this scenario? Adding JavaDoc comments and perhaps changing the name would be helpful.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/utils/DocumentType.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package org.apache.nifi.services.couchbase.utils;
+
+/**
+ * Couchbase document types.
+ */
+public enum DocumentType {
+
+    Json( "application/json"),
+    Binary( "application/octet-stream");
```
exceptionfactory: By convention, enums should be all caps:
```suggestion
    JSON( "application/json"),
    BINARY( "application/octet-stream");
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services-nar/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,39 @@
...
+      The following NOTICE information applies:
+        Couchbase JVM Core IO
+        Copyright 2014 Couchbase, Inc.
+
+    (ASLv2) Non-Blocking Reactive Foundation for the JVM
+      The following NOTICE information applies:
+        Non-Blocking Reactive Foundation for the JVM
+        Copyright 2005 - 2024 Broadcom
+
+    (ASLv2) JSpecify annotations
```
exceptionfactory: Is this annotation needed at runtime? If not, it should be excluded

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseClient.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    public CouchbaseGetResult getDocument(String documentId) throws CouchbaseException {
+        try {
+            final GetResult result = collection.get(documentId, GetOptions.getOptions().transcoder(getTranscoder(documentType)));
+
+            return new CouchbaseGetResult(result.contentAsBytes(), result.cas());
+        } catch (Exception e) {
+            throw new CouchbaseException(e);
```
exceptionfactory: A message should always be included when wrapping an exception, recommend changing the constructor to require a message and a cause.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseClient.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    @Override
+    public CouchbaseUpsertResult upsertDocument(String documentId, byte[] content) throws CouchbaseException {
+        try {
+            if (!getInputValidator(documentType).test(content)) {
+                throw new CouchbaseException("The provided input is invalid.");
```
exceptionfactory: Trailing `.` should be avoided in messages.
```suggestion
                throw new CouchbaseException("The provided input is invalid");
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseConnectionService.java
```diff
@@ -0,0 +1,154 @@
...
+            .build();
+
+    public static PropertyDescriptor SSL_CONTEXT_SERVICE = new PropertyDescriptor.Builder()
+            .name("SSL Context Service")
+            .description("Service supporting SSL communication configuration. The service is using one-way SSL, so only the trust store properties will be used.")
+            .identifiesControllerService(SSLContextService.class)
+            .build();
+
+    public static final PropertyDescriptor PERSIST_TO = new PropertyDescriptor.Builder()
+            .name("Persist To")
```
exceptionfactory: The naming of this property seems slightly odd. Something like `Persistence Strategy` or `Persistent Storage Destination` seems better. What do you think?

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseConnectionService.java
```diff
@@ -0,0 +1,154 @@
...
+            .description("Durability constraint about disk persistence.")
+            .required(true)
+            .allowableValues(PersistTo.values())
+            .defaultValue(PersistTo.NONE.toString())
+            .build();
+
+    public static final PropertyDescriptor REPLICATE_TO = new PropertyDescriptor.Builder()
+            .name("Replicate To")
```
exceptionfactory: Similar to above, `Replication Strategy` or `Durability Constraint` seems more precise.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseConnectionService.java
```diff
@@ -0,0 +1,154 @@
...
+    private volatile Cluster cluster;
+
+    @Override
+    protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+        return PROPERTIES;
+    }
+
+    private String connectionString;
+    private PersistTo persistTo;
+    private ReplicateTo replicateTo;
```
exceptionfactory: These declarations should be located above all methods.

exceptionfactory (changes requested):
Thanks for the work on this @mark-bathori. I noted a handful of minor items, as well as several property naming and description elements that would be very helpful to improve before incorporating this into the project.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/GetCouchbase.java
mark-bathori: The old implementation supported this, so I kept this functionality to maintain similar behavior. However, I believe it can be removed.

exceptionfactory: Thanks, I recommend removing this feature in this new Processor.

exceptionfactory:
@mark-bathori On review, it seems like making these Processors more specifically named, such as `GetCouchbaseDocument` and `PutCouchbaseDocument` would be helpful. This would indicate that each Processor only supports a single document, and leaves room for the potential of multi-document record-oriented Processors. What do you think?

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
mark-bathori: Currently, the processor supports two document types: `JSON` and `BINARY`. The Couchbase `upsert` method accepts content as an `Object`, but it requires a `Transcoder` to define what type of content it can process. For the document types mentioned, we are using [RawBinaryTranscoder](https://github.com/couchbase/couchbase-jvm-clients/blob/master/java-client/src/main/java/com/couchbase/client/java/codec/RawBinaryTranscoder.java) and [RawJsonTranscoder](https://github.com/couchbase/couchbase-jvm-clients/blob/master/java-client/src/main/java/com/couchbase/client/java/codec/RawJsonTranscoder.java). If you check the encode method in both implementations, `byte[]` is the common type they can handle.

I also checked the Couchbase documentation, which states that the maximum size for a single document is 20MB. Of course, this does not prevent a user from attempting to send larger files to the processor. Should I add a size check for the FlowFile?

mark-bathori:
@exceptionfactory I support adding the extensibility for multi-document, record-oriented implementations, but I'm not sure what the best approach is yet.

The name `Document` might not be the most suitable, as the processors can also handle binary data, which is more of a "Key-Value Store"-like function rather than a "Document Database" function. At the moment, however, I don't have a better name.

I'm not sure what the "standard approach" is in NiFi for handling different types of inputs in a processor. For example, some processors (like PublishMQTT) can process both FlowFile content and a Record Reader, while other solutions implement these two functions in separate processors (e.g., the old ConsumeKafka and ConsumeKafkaRecord).

Which one do you think is the correct way?

exceptionfactory:
> @exceptionfactory I support adding the extensibility for multi-document, record-oriented implementations, but I'm not sure what the best approach is yet.
> 
> The name `Document` might not be the most suitable, as the processors can also handle binary data, which is more of a "Key-Value Store"-like function rather than a "Document Database" function. At the moment, however, I don't have a better name.
> 
> I'm not sure what the "standard approach" is in NiFi for handling different types of inputs in a processor. For example, some processors (like PublishMQTT) can process both FlowFile content and a Record Reader, while other solutions implement these two functions in separate processors (e.g., the old ConsumeKafka and ConsumeKafkaRecord).
> 
> Which one do you think is the correct way?

Thanks for the reply. On further consideration, the current naming can work. As you noted, one strategy is to introduce new Processors with the `Record` suffix, and another approach is to introduce record-oriented handling as a strategy property.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseClient.java
mark-bathori: Currently there are two type of content supported: `JSON` and `BINARY`. The selected document type determines which transcoder is used during the `upsert` action but Couchbase itself doesn't validate the content. I've added custom data validation in NiFi side in the `getInputValidator` method which you can find in the `StandardCouchbaseClient` class.

exceptionfactory: Thanks for the reference, I noticed the validator and observed that Binary allows literally anything. Is that a desirable use case?

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseClient.java
mark-bathori: While the class could be moved to the processor level, maintaining the current functionality would require the ErrorHandlingStrategy enum to remain at the API level to ensure consistent error grouping based on Couchbase SDK exceptions.
What is your opinion? Do you have any ideas on how this could be moved to the processor bundle while retaining the ability to categorize SDK exceptions without adding three new methods to the API?

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/CouchbaseClient.java
mark-bathori: Based on the Couchbase documentation binary type can be anything therefor there is no structural restriction on them: https://docs.couchbase.com/go-sdk/current/concept-docs/nonjson.html

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/src/main/java/org/apache/nifi/processors/couchbase/AbstractCouchbaseProcessor.java
exceptionfactory: Thanks for the background. It might be worth considering an client-side enforcement, but for now, leaving it to the Couchbase server seems reasonable.

exceptionfactory (commented):
Thanks for the updates @mark-bathori, this looks close to completion, I plan to follow up soon.

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-processors/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>couchbase</artifactId>
+            <version>1.21.3</version>
```
exceptionfactory: This version should either be removed, or reference `testcontainers.version`

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-services-api/src/main/java/org/apache/nifi/services/couchbase/utils/DocumentType.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package org.apache.nifi.services.couchbase.utils;
+
+/**
+ * Couchbase document types.
+ */
+public enum DocumentType {
+
+    JSON( "application/json"),
+    BINARY( "application/octet-stream");
```
exceptionfactory:
```suggestion
    JSON("application/json"),
    BINARY("application/octet-stream");
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+        <groupId>org.apache.nifi</groupId>
+        <artifactId>nifi-couchbase-bundle</artifactId>
+        <version>2.7.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>nifi-couchbase-standard-services</artifactId>
+    <packaging>jar</packaging>
+
+    <properties>
+        <couchbase.version>3.9.0</couchbase.version>
```
exceptionfactory: This can now be updated to 3.9.1

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseClient.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    public CouchbaseGetResult getDocument(String documentId) throws CouchbaseException {
+        try {
+            final GetResult result = collection.get(documentId, GetOptions.getOptions().transcoder(getTranscoder(documentType)));
+
+            return new CouchbaseGetResult(result.contentAsBytes(), result.cas());
+        } catch (Exception e) {
+            throw new CouchbaseException("An error occurred while getting the document from Couchbase", e);
```
exceptionfactory: Recommend including the `documentId` in the message
```suggestion
            throw new CouchbaseException("Failed to get document [%s] from Couchbase".formatted(documentId), e);
```

nifi-extension-bundles/nifi-couchbase-bundle/nifi-couchbase-standard-services/src/main/java/org/apache/nifi/services/couchbase/StandardCouchbaseClient.java
```diff
@@ -0,0 +1,162 @@
...
+
+            final MutationResult result = collection.upsert(documentId, content,
+                    UpsertOptions.upsertOptions()
+                            .durability(persistTo, replicateTo)
+                            .transcoder(getTranscoder(documentType))
+                            .clientContext(new HashMap<>()));
+
+            return new CouchbaseUpsertResult(result.cas());
+        } catch (Exception e) {
+            throw new CouchbaseException("An error occurred while upserting the document in Couchbase", e);
```
exceptionfactory:
```suggestion
            throw new CouchbaseException("Failed to upsert document [%s] in Couchbase".formatted(documentId), e);
```

exceptionfactory (changes requested):
Reviewing the latest updates, I noted a few more minor issues, and I think resolving the following items should allow this to be ready:

1. Error Handling Strategy: Instead of having `CouchbaseErrorHandler` as the Service API response, what do you think about changing this to `getExceptionCategory`? The `ErrorHandlingStrategy` enum could be renamed to `ExceptionCategory`, and then the Processor could implement the appropriate logic. The difference is that the Service API and implementation define the category of the exception, but the Processor is responsible for deciding how to handle it. The net result is similar, but it keeps the responsibility more clearly defined.

2. For this initial pull request, I recommend removing the Couchbase NARs from the `nifi-assembly`. Although the Service API and Processor NARs have no other dependencies, which is great, the Standard Service NAR includes the Couchbase Core IO, which is about 30 MB. The reason it is so large is that it includes shaded versions of multiple libraries, including gRPC, Jackson, and Netty. Specifically, it includes platform-dependent native libraries from Netty. From a dependency management perspective, this makes it impossible to override the shaded version of Netty, creating potential vulnerability issues when Netty publishes vulnerabilities. Couchbase is not unique in this way, but it is less common. Given the size and shading, I think it is best to leave it out of the standard binary distribution, at least for the initial version.

chedim:
@exceptionfactory , can I start working on the other parts of the other CB PR with the rest of the components that were there?

mark-bathori:
@chedim I've got them almost ready, opening PR soon.

PR #10039, opened 2025-06-21
NIFI-14682 - Help user with JDBC Driver configuration in DBCP controller service

pvillard31:
# Summary

[NIFI-14682](https://issues.apache.org/jira/browse/NIFI-14682) - Help user with JDBC Driver configuration in DBCP controller service

The goal is to improve the `verify` method in the `Abstract` class as well as the error handling in the `getDriver` method of the controller service implementation.

We cannot use a `customValidate` approach as we may be in situation where one of those situations happen:

- specified resources are not loaded in the classpath
- some specified resources are using the URL approach and we may not have the resource locally available at the time
- we may have some weirdly packaged JARs with very custom Drivers

Given the above, we don't want to take the risk of making the component invalid and unusable.

Instead we improve things in two places:

- the `verify()` method but we cannot expect all of the specified resources to be loaded in the classpath at the time the method is called so the best approach is to have a static analysis of whatever is available to us and provide a non exhaustive list of what drivers can be used.
- the `getDriver()` method - in this case, resources are loaded in the classpath so we can also verify that the classes that we suspect are potential drivers could indeed be loaded as drivers.

The changes have been tested with many drivers, specified in different ways (asset, directory, file, URL) with many different vendors (H2, Derby, MS SQL, Postgres, ClickHouse, Snowflake).

While this does not cover all possible scenarios of how the component has been configured, it does provide a more helpful error message in most cases when the specified driver class is not correct.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/AbstractDBCPConnectionPool.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/AbstractDBCPConnectionPool.java
@@ -16,14 +16,6 @@
  */
 package org.apache.nifi.dbcp;
 
-import java.sql.Connection;
-import java.sql.Driver;
-import java.sql.SQLException;
-import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
 import org.apache.commons.dbcp2.BasicDataSource;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.annotation.lifecycle.OnDisabled;
@@ -31,6 +23,9 @@ import org.apache.nifi.annotation.lifecycle.OnEnabled;
 import org.apache.nifi.components.ConfigVerificationResult;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.PropertyValue;
+import org.apache.nifi.components.resource.ResourceReference;
+import org.apache.nifi.components.resource.ResourceReferences;
+import org.apache.nifi.components.resource.ResourceType;
 import org.apache.nifi.controller.AbstractControllerService;
 import org.apache.nifi.controller.ConfigurationContext;
 import org.apache.nifi.controller.VerifiableControllerService;
@@ -43,8 +38,25 @@ import org.apache.nifi.security.krb.KerberosAction;
 import org.apache.nifi.security.krb.KerberosLoginException;
 import org.apache.nifi.security.krb.KerberosUser;
 
+import java.io.File;
+import java.sql.Connection;
+import java.sql.Driver;
+import java.sql.SQLException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Map;
+import java.util.Scanner;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.stream.Collectors;
+
 import static org.apache.nifi.components.ConfigVerificationResult.Outcome.FAILED;
 import static org.apache.nifi.components.ConfigVerificationResult.Outcome.SUCCESSFUL;
+import static org.apache.nifi.dbcp.utils.DBCPProperties.DB_DRIVERNAME;
 import static org.apache.nifi.dbcp.utils.DBCPProperties.DB_DRIVER_LOCATION;
 import static org.apache.nifi.dbcp.utils.DBCPProperties.KERBEROS_USER_SERVICE;
 
@@ -58,6 +70,7 @@ public abstract class AbstractDBCPConnectionPool extends AbstractControllerServi
         List<ConfigVerificationResult> results = new ArrayList<>();
 
         KerberosUser kerberosUser = null;
+
         try {
             kerberosUser = getKerberosUser(context);
             if (kerberosUser != null) {
@@ -103,14 +116,26 @@ public abstract class AbstractDBCPConnectionPool extends AbstractControllerServi
         } catch (final Exception e) {
             String message = "Failed to configure Data Source.";
             if (e.getCause() instanceof ClassNotFoundException) {
-                message += String.format("  Ensure changes to the '%s' property are applied before verifying",
+                message += String.format(" Ensure changes to the '%s' property are applied before verifying.",
                         DB_DRIVER_LOCATION.getDisplayName());
             }
             verificationLogger.error(message, e);
+
+            final String driverName = context.getProperty(DB_DRIVERNAME).evaluateAttributeExpressions().getValue();
+            final ResourceReferences driverResources = context.getProperty(DB_DRIVER_LOCATION).evaluateAttributeExpressions().asResources();
+
+            if (StringUtils.isNotBlank(driverName) && driverResources.getCount() != 0) {
+                List<String> availableDrivers = discoverDriverClassesStatic(driverResources);
+                if (!availableDrivers.isEmpty() && !availableDrivers.contains(driverName)) {
+                    message += String.format(" Driver class '%s' not found in provided resources. Available driver classes found: %s.",
+                            driverName, String.join(", ", availableDrivers));
+                }
+            }
+
             results.add(new ConfigVerificationResult.Builder()
                     .verificationStepName("Configure Data Source")
                     .outcome(FAILED)
-                    .explanation(message + ": " + e.getMessage())
+                    .explanation(message + " Exception: " + e.getMessage())
                     .build());
         } finally {
             try {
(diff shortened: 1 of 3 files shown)


exceptionfactory (commented):
Thanks for working on this improvement @pvillard31.

I plan to take a closer look, but on a quick review, it looks like a number of the patterns and listed class names should be defined as static variables instead of local method variables.

pvillard31:
I have pushed some changes to remove the checks based on naming and fairly common database types to only rely on `META-INF/services` at this point - for the verify where class loading may not have happened yet. I have improved the error messages in case the verify is used before or after configuration is applied and driver is loaded. I tested with a few large drivers including the phoenix driver that is about 150MB, all of the operations took below or around one second to complete.

Screenshot 2025-07-30 at 18 13 06
Screenshot 2025-07-30 at 18 13 30
Screenshot 2025-07-30 at 18 13 59
Screenshot 2025-07-30 at 18 14 50

nifi-extension-bundles/nifi-extension-utils/nifi-dbcp-utils/src/main/java/org/apache/nifi/dbcp/utils/DriverUtils.java
```diff
@@ -0,0 +1,215 @@
...
+     */
+    private static boolean isPotentialDriverClass(final String className) {
+        final String lowerClassName = className.toLowerCase();
+
+        // Common patterns for JDBC driver class names
+        boolean isCandidate = lowerClassName.contains("driver")
+                || lowerClassName.contains("jdbc")
+                || className.matches(".*Driver$")
+                || className.matches(".*JdbcDriver$")
+                || className.matches(".*SQLServerDriver$");
```
exceptionfactory: For clarity of maintenance, I recommend creating a single regular expression pattern and declaring it as a static variable.

nifi-extension-bundles/nifi-extension-utils/nifi-dbcp-utils/src/main/java/org/apache/nifi/dbcp/utils/DriverUtils.java
```diff
@@ -0,0 +1,215 @@
...
+public class DriverUtils {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DriverUtils.class);
+
+    /**
+     * Discovers JDBC driver classes using static JAR scanning only (no class
+     * loading) This is used during validation when resources aren't loaded into
+     * classpath yet
+     */
+    public static List<String> discoverDriverClassesStatic(final ResourceReferences driverResources) {
```
exceptionfactory: The word `Static` can be removed from this and other methods.
```suggestion
    public static List<String> findDriverClassNames(final ResourceReferences driverResources) {
```

nifi-extension-bundles/nifi-extension-utils/nifi-dbcp-utils/src/main/java/org/apache/nifi/dbcp/utils/DriverUtils.java
```diff
@@ -0,0 +1,215 @@
...
+        LOGGER.debug("Static driver discovery completed. Found {} potential driver classes", driverClasses.size());
+        return new ArrayList<>(driverClasses);
+    }
+
+    /**
+     * Scans a JAR file for potential JDBC driver class names using static analysis
+     * only This uses improved heuristics including checking META-INF/services and
+     * common patterns
+     */
+    private static Set<String> scanJarForDriverClassesStatic(final File jarFile) {
```
exceptionfactory:
```suggestion
    private static Set<String> scanJarForDriverClasses(final File jarFile) {
```

nifi-extension-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/AbstractDBCPConnectionPool.java
```diff
@@ -101,16 +105,30 @@ public List<ConfigVerificationResult> verify(final ConfigurationContext context,
...
+            final String driverName = context.getProperty(DB_DRIVERNAME).evaluateAttributeExpressions().getValue();
+            final ResourceReferences driverResources = context.getProperty(DB_DRIVER_LOCATION).evaluateAttributeExpressions().asResources();
+
+            if (StringUtils.isNotBlank(driverName) && driverResources.getCount() != 0) {
+                List<String> availableDrivers = DriverUtils.discoverDriverClassesStatic(driverResources);
+                if (!availableDrivers.isEmpty() && !availableDrivers.contains(driverName)) {
+                    messageBuilder.append(String.format(" Driver class '%s' not found in provided resources. Available driver classes found: %s.",
+                            driverName, String.join(", ", availableDrivers)));
+                } else if (e.getCause() instanceof ClassNotFoundException && availableDrivers.contains(driverName)) {
+                    messageBuilder.append(" Driver found but ensure you apply the controller service configuration before verifying again.");
```
exceptionfactory: Recommend avoiding `you` in messages, perhaps something like the following
```suggestion
                    messageBuilder.append(" Driver Class found but not loaded: Apply configuration before verifying");
```

nifi-extension-bundles/nifi-extension-utils/nifi-dbcp-base/src/main/java/org/apache/nifi/dbcp/AbstractDBCPConnectionPool.java
```diff
@@ -101,16 +105,30 @@ public List<ConfigVerificationResult> verify(final ConfigurationContext context,
...
+            StringBuilder messageBuilder = new StringBuilder("Failed to configure Data Source.");
+            verificationLogger.error(messageBuilder.toString(), e);
+
+            final String driverName = context.getProperty(DB_DRIVERNAME).evaluateAttributeExpressions().getValue();
+            final ResourceReferences driverResources = context.getProperty(DB_DRIVER_LOCATION).evaluateAttributeExpressions().asResources();
+
+            if (StringUtils.isNotBlank(driverName) && driverResources.getCount() != 0) {
+                List<String> availableDrivers = DriverUtils.discoverDriverClassesStatic(driverResources);
+                if (!availableDrivers.isEmpty() && !availableDrivers.contains(driverName)) {
+                    messageBuilder.append(String.format(" Driver class '%s' not found in provided resources. Available driver classes found: %s.",
```
exceptionfactory: Recommend avoiding the trailing `.` character:
```suggestion
                    messageBuilder.append(String.format(" Driver class [%s] not found in provided resources. Available driver classes found: %s",
```

pvillard31:
Done, thanks @exceptionfactory

nifi-extension-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-dbcp-service/src/test/java/org/apache/nifi/dbcp/DBCPServiceTest.java
```diff
@@ -205,6 +211,41 @@ public void testGetDataSourceProperties() throws SQLException {
         service.getDataSource().close();
     }
 
+    @Test
+    public void testInvalidDirverH2() throws URISyntaxException {
```
exceptionfactory:
```suggestion
    public void testInvalidDriverH2() throws URISyntaxException {
```

nifi-extension-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-dbcp-service/src/test/java/org/apache/nifi/dbcp/DBCPServiceTest.java
```diff
@@ -205,6 +211,41 @@ public void testGetDataSourceProperties() throws SQLException {
...
+        final URL driverURL = org.h2.Driver.class
+                .getProtectionDomain()
+                .getCodeSource()
+                .getLocation();
```
exceptionfactory: For testing purposes, using the Derby JDBC Driver looks like it would avoid this reference to H2, which otherwise appears to be unused.

nifi-extension-bundles/nifi-standard-services/nifi-dbcp-service-bundle/nifi-dbcp-service/src/main/java/org/apache/nifi/dbcp/DBCPConnectionPool.java
```diff
@@ -203,7 +207,33 @@ protected Driver getDriver(final String driverName, final String url) {
...
+            } catch (Exception ignored) {
+                // Context might not be available, continue without it
+            }
+
+            final List<String> availableDrivers = (driverResources != null && driverResources.getCount() != 0) ? DriverUtils.discoverDriverClasses(driverResources) : List.of();
+
+            StringBuilder errorMessage = new StringBuilder(String.format("JDBC driver class '%s' not found.", driverName));
+
+            if (!availableDrivers.isEmpty()) {
+                errorMessage.append(String.format(" Available driver classes found in resources: %s.", String.join(", ", availableDrivers)));
```
exceptionfactory: Minor note, but using `" Available driver classes...".formatted()` instead of `String.format()` would reduce some of the nesting, making this a bit easier to read.

pvillard31:
Thanks @exceptionfactory - have made the recommended changes and switched to Derby for the unit test

exceptionfactory (approved):

PR #10053, opened 2025-06-27
NIFI-14696 Fixed KinesisRecordProcessorRecord schema changes handling

dariuszseweryn:
# Summary

[NIFI-14696](https://issues.apache.org/jira/browse/NIFI-14696)

`KinesisRecordProcessorRecord` did not handle situations where schema of subsequent records in a batch changed (either due to having embedded schema or schema inference) which lead to various scenarios:
- records columns being silently dropped
- records couldn't be written (i.e. `IllegalTypeConversionException` could be thrown)

Additionally the last batch record caused an exception that was "handled" by retries in `AbstractKinesisRecordProcessor` the FlowFile was not properly completed — i.e. it was not having attributes set.

Now these are fixed:
- in case of schema change either previous FlowFile is being properly completed and a new FlowFile is started, or records with different schemas get grouped in separate FlowFiles

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

- [x] (no new) New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [x] (no new) New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [x] (no new) Documentation formatting appears as expected in rendered files


nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -99,27 +100,44 @@ void processRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kin
...
                 final WriteResult writeResult = writer.write(outputRecord);
-                recordCount += writeResult.getRecordCount();
-
-                // complete the FlowFile if there are no more incoming Kinesis Records and no more records in this RecordSet
-                if (lastRecord && !recordSet.isAnotherRecord()) {
-                    completeFlowFile(flowFiles, session, recordCount, writeResult, kinesisRecord, stopWatch);
-                }
-                firstOutputRecord = false;
+                // definitely no longer empty
+                emptyFlowFileToRemove = null;
```
awelless: Nit: perhaps, it would be simpler to keep `flowFile` but introduce `boolean flowFileEmpty` to check if a file should be removed. Or even we can use `firstSuccessfulWriteInfo` to check if anything has been written.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -99,27 +100,44 @@ void processRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kin
...
+            } catch (IOException e) {
+                getLogger().error("Failed to complete a FlowFile, dropped records from stream: {}, shardId: {}, sequence number range: [{}, {}], due to {}",
+                        getStreamName(),
+                        getKinesisShardId(),
+                        firstSuccessfulWriteInfo.kinesisRecord.sequenceNumber(),
+                        lastSuccessfulWriteInfo.kinesisRecord.sequenceNumber(),
+                        e.getLocalizedMessage(),
+                        e);
+            }
+            firstSuccessfulWriteInfo = null;
```
awelless: Previously we did `outputRawRecordOnException` when `completeFlowFile` threw any error. After refactoring this out, should we preserve the error handling logic there?

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/record/TestKinesisRecordProcessorRecord.java
```diff
@@ -343,6 +347,23 @@ public void testProcessUnparsableRecordWithRawOutputWithCheckpoint() throws Shut
         session.assertNotRolledBack();
     }
 
+    @Test
+    public void testProcessUnparsableRecordWithRawOutputWithCheckpoint_lastRecordInvalid() throws ShutdownException, InvalidStateException {
```
awelless: Nit: `@ParameterizedTest` with `@MethodSource` instead of `testProcessUnparsableRecordWithRawOutputWithCheckpoint` and `testProcessUnparsableRecordWithRawOutputWithCheckpoint_lastRecordInvalid` should make the tests clearer.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: I have reverted to using a boolean (based on `firstSuccessfulWriteInfo`) to mark if the FlowFile should be removed or not

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: I initially thought that it is not worth it but I re-added `outputRawRecordOnException` to have the same behavior to the extent possible

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -82,15 +85,18 @@ void startProcessingRecords() {
...
     void processRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kinesisRecord, final boolean lastRecord,
                        final ProcessSession session, final StopWatch stopWatch) {
-        boolean firstOutputRecord = true;
-        int recordCount = 0;
+        if (flowFiles.size() > 1) {
+            // historically this code has assumed that the FlowFile it operates on is the first one in the list.
+            // changing this behavior would exceed the scope of the bugfix change this comment was added in.
+            // leaving this comment to inform future maintainers that assumption is rather weak and should be refactored
+            getLogger().warn("More than one FlowFile is being processed at once, this is not expected.", flowFiles);
+        }
```
exceptionfactory: Instead of introducing this log and message, please create a Jira issue for tracking.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -75,51 +82,93 @@ public KinesisRecordProcessorRecord(final ProcessSessionFactory sessionFactory,
...
+        if (flowFileState != null && !flowFiles.contains(flowFileState.flowFile)) {
+            getLogger().warn("Currently processed FlowFile is no longer available, this is not expected.", flowFiles);
+            closeSafe(flowFileState, "FlowFile State");
+            flowFileState = null;
         }
+        if (kinesisRecord == failedKinesisRecordCausingDataLoss)  {
+            // AbstractKinesisRecordProcessor does retry processing of failed records. however in case of CompleteFlowFileMultipleKinesisRecordException it is impossible to determine the state of
+            // the affected FlowFile and replay all records that were in it. To prevent data loss, the exception needs to be rethrown until it is given up by the abstract processor. still there may be
+            // other cases where we end up in undetermined state.
+            throw failedKinesisRecordCausingDataLossException;
```
awelless: Would it be possible to adjust the retrying logic instead? By telling apart retriable and persistent errors we can avoid doing unnecessary retries in `AbstractKinesisRecordProcessor`.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -197,7 +278,75 @@ private void outputRawRecordOnException(final boolean firstOutputRecord, final F
...
+                closeable.close();
+            } catch (final IOException e) {
+                getLogger().warn("Failed to close {} due to {}", closeableName, e.getLocalizedMessage(), e);
+            }
+        }
+    }
+
+    private record SuccessfulWriteInfo(KinesisClientRecord kinesisRecord, WriteResult writeResult) { }
+
+    private class FlowFileState implements Closeable {
```
awelless: Nit: will `SchematizedFlowFile` or `FlowFileWithSchema` better represent the idea behind this class?

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -47,16 +51,19 @@
...
+import java.util.Optional;
 
 public class KinesisRecordProcessorRecord extends AbstractKinesisRecordProcessor {
     final RecordReaderFactory readerFactory;
     final RecordSetWriterFactory writerFactory;
     final Map<String, String> schemaRetrievalVariables;
 
-    private RecordSetWriter writer;
-    private OutputStream outputStream;
+    private FlowFileState flowFileState;
```
awelless: Nit: I reckon `(current|open)FlowFile` name will better convey the meaning of this field.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -75,51 +82,93 @@ public KinesisRecordProcessorRecord(final ProcessSessionFactory sessionFactory,
...
+                return;
+            }
+            if (!flowFiles.contains(flowFileState.flowFile)) {
+                getLogger().warn("Currently processed FlowFile is no longer available at processing end, this is not expected.", flowFiles);
+                closeSafe(flowFileState, "FlowFile State");
+                return;
+            }
+            completeFlowFile(flowFiles, session, stopWatch);
+        } catch (CompleteFlowFileSingleKinesisRecordException e) {
+            final boolean removeFirstFlowFileIfAvailable = true;
```
awelless: Nit: that's "current" file now, not "first"

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -128,62 +177,94 @@ void processRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kin
...
+            closeSafe(flowFileState, "FlowFile State");
+            final String message = "Failed to complete a FlowFile containing records from Stream Name: %s, Shard Id: %s, Sequence/Subsequence No range: [%s/%d, %s/%d)".formatted(
+                    getStreamName(),
+                    getKinesisShardId(),
+                    flowFileState.firstSuccessfulWriteInfo.kinesisRecord.sequenceNumber(),
+                    flowFileState.firstSuccessfulWriteInfo.kinesisRecord.subSequenceNumber(),
+                    flowFileState.lastSuccessfulWriteInfo.kinesisRecord.sequenceNumber(),
+                    flowFileState.lastSuccessfulWriteInfo.kinesisRecord.subSequenceNumber()
+            );
+            final boolean isSingleKinesisRecordInFlowFile = flowFileState.firstSuccessfulWriteInfo.kinesisRecord.equals(flowFileState.lastSuccessfulWriteInfo.kinesisRecord);
```
awelless: Nit: let's incapsulate the logic of determining how many records there are in the current flow file into `FlowFileState` class.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -128,62 +177,94 @@ void processRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kin
...
+            if (isSingleKinesisRecordInFlowFile) {
+                // in case of a single Kinesis Record in FlowFile, so we can route it to failure relationship
+                throw new CompleteFlowFileSingleKinesisRecordException(message, e, flowFileState.firstSuccessfulWriteInfo.kinesisRecord);
+            }
+            // in case of multiple Kinesis Records the whole batch needs to be failed, otherwise data can be lost
+            throw new CompleteFlowFileMultipleKinesisRecordException(message, e);
```
awelless: For me different exception types seem counterintuitive.
Should we better do this check in the error handling piece of code?

awelless (commented):
From functional point of view the changes look good.
Left a few comments regarding error handling code organization.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: Yup, added a `KinesisBatchUnrecoverableException` at `AbstractKinesisRecordProcessor`

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: It is a little bit more than that — I wanted to give the notion of all the state associated with the `FlowFile`, not only schema

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: `currentFlowFileState` it is then

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/AbstractKinesisRecordProcessor.java
```diff
@@ -181,8 +186,11 @@ private boolean attemptProcessRecord(final List<FlowFile> flowFiles, final Kines
                                          final ProcessSession session, final StopWatch stopWatch) {
         boolean processedSuccessfully = false;
         try {
-            processRecord(flowFiles, kinesisRecord, lastRecord, session, stopWatch);
```
dariuszseweryn, line 184: TODO: remove `lastRecord` from the `attemptProcessRecord` signature

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
```diff
@@ -686,7 +686,7 @@ private static DataType inferRecordDataType(final Map<String, ?> map) {
      * @param strict check for a strict match, i.e. all fields in the record should have a corresponding entry in the schema
      * @return True if the object is compatible with the schema
      */
-    private static boolean isRecordTypeCompatible(RecordSchema schema, Object value, boolean strict) {
```
dariuszseweryn: If there is a better suited function available — feel free to suggest

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/AbstractKinesisRecordProcessor.java
```diff
@@ -177,12 +182,15 @@ private int processRecordsWithRetries(final List<KinesisClientRecord> records, f
...
-    private boolean attemptProcessRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kinesisRecord, final boolean lastRecord,
+    private boolean attemptProcessRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kinesisRecord,
                                          final ProcessSession session, final StopWatch stopWatch) {
         boolean processedSuccessfully = false;
         try {
-            processRecord(flowFiles, kinesisRecord, lastRecord, session, stopWatch);
+            processRecord(flowFiles, kinesisRecord, session, stopWatch);
             processedSuccessfully = true;
+        } catch (final KinesisBatchUnrecoverableException e) {
+            // don't attempt retry, rethrow
```
exceptionfactory: Recommend removing this comment since it simply describes the behavior, and the exception name itself indicates the reason.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/record/TestKinesisRecordProcessorRecord.java
```diff
@@ -284,22 +290,63 @@ public void testProcessPoisonPillRecordButNoRawOutputWithCheckpoint() throws Shu
...
+    private static Stream<Arguments> unparsableRecordsLists() {
+        final KinesisClientRecord unparsableRecordMock = mock(KinesisClientRecord.class);
+        return Stream.of(
+                Arguments.argumentSet("Unparsable At The Beginning",
+                        Arrays.asList(
+                                unparsableRecordMock,
+                                KinesisClientRecord.builder().approximateArrivalTimestamp(null)
+                                        .partitionKey("partition-1")
+                                        .sequenceNumber("1")
+                                        .data(ByteBuffer.wrap("{\"record\":\"1\"}".getBytes(StandardCharsets.UTF_8)))
```
exceptionfactory: It looks the data content for each of these examples could be declared statically and reused.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -34,10 +34,14 @@
 import org.apache.nifi.serialization.record.PushBackRecordSet;
 import org.apache.nifi.serialization.record.Record;
 import org.apache.nifi.serialization.record.RecordSchema;
+import org.apache.nifi.serialization.record.util.DataTypeUtils;
+import org.apache.nifi.serialization.record.util.IllegalTypeConversionException;
 import org.apache.nifi.util.StopWatch;
+import org.jetbrains.annotations.NotNull;
```
exceptionfactory: The `NotNull` annotation from JetBrains should generally be avoided given the lack of the explicit dependency.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -75,51 +79,85 @@ public KinesisRecordProcessorRecord(final ProcessSessionFactory sessionFactory,
...
+
+    @Override
+    void finishProcessingRecords(final ProcessSession session, final List<FlowFile> flowFiles, final StopWatch stopWatch) {
+        super.finishProcessingRecords(session, flowFiles, stopWatch);
+        try {
+            if (currentFlowFileState == null) {
+                return;
+            }
+            if (!flowFiles.contains(currentFlowFileState.flowFile)) {
+                getLogger().warn("Currently processed FlowFile is no longer available at processing end, this is not expected.", flowFiles);
```
exceptionfactory: The `flowFiles` argument is not included as a parameter in this warning log message. Recommend logging the specified FlowFile and just the number of others.

However, can this occur? Instead of logging a warning, should an `IllegalStateException` be thrown instead?

```suggestion
                getLogger().warn("{} is not available in provided FlowFiles [{}]", currentFlowFileState.flowFile, flowFiles.size());
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -75,51 +79,85 @@ public KinesisRecordProcessorRecord(final ProcessSessionFactory sessionFactory,
...
+    void processRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kinesisRecord,
                        final ProcessSession session, final StopWatch stopWatch) {
-        boolean firstOutputRecord = true;
-        int recordCount = 0;
-        final ByteBuffer dataBuffer = kinesisRecord.data();
-        byte[] data = dataBuffer != null ? new byte[dataBuffer.remaining()] : new byte[0];
-        if (dataBuffer != null) {
-            dataBuffer.get(data);
+        if (currentFlowFileState != null && !flowFiles.contains(currentFlowFileState.flowFile)) {
+            getLogger().warn("Currently processed FlowFile is no longer available, this is not expected.", flowFiles);
```
exceptionfactory: See note above on similar behavior.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -75,51 +79,85 @@ public KinesisRecordProcessorRecord(final ProcessSessionFactory sessionFactory,
...
+                    flowFiles.add(currentFlowFileState.flowFile);
                 }
 
-                final WriteResult writeResult = writer.write(outputRecord);
-                recordCount += writeResult.getRecordCount();
-
-                // complete the FlowFile if there are no more incoming Kinesis Records and no more records in this RecordSet
-                if (lastRecord && !recordSet.isAnotherRecord()) {
-                    completeFlowFile(flowFiles, session, recordCount, writeResult, kinesisRecord, stopWatch);
+                if (!DataTypeUtils.isRecordTypeCompatible(currentFlowFileState.recordSchema, outputRecord, false)) {
```
exceptionfactory: Although the explicit check is an option, is there a reason for this approach as opposed to catching exceptions when the type is not compatible?

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -75,51 +79,85 @@ public KinesisRecordProcessorRecord(final ProcessSessionFactory sessionFactory,
...
+                    // subsequent records may have different schema. if so, try complete current FlowFile and start a new one with wider schema to continue
```
exceptionfactory: The comment raises a question about the implementation approach. Schema changes should either work with an inferred schema, or should throw an exception because of a schema mismatch.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -128,62 +166,88 @@ void processRecord(final List<FlowFile> flowFiles, final KinesisClientRecord kin
...
 
-    private void createWriter(final FlowFile flowFile, final ProcessSession session, final Record outputRecord)
-            throws IOException, SchemaNotFoundException {
-
-        final RecordSchema readerSchema = outputRecord.getSchema();
-        final RecordSchema writeSchema = writerFactory.getSchema(schemaRetrievalVariables, readerSchema);
-        outputStream = session.write(flowFile);
-        writer = writerFactory.createWriter(getLogger(), writeSchema, outputStream, flowFile);
-        writer.beginRecordSet();
+    private static byte @NotNull [] getData(final KinesisClientRecord kinesisRecord) {
```
exceptionfactory:
```suggestion
    private static byte [] getData(final KinesisClientRecord kinesisRecord) {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -197,7 +261,64 @@ private void outputRawRecordOnException(final boolean firstOutputRecord, final F
...
-        return getDefaultAttributes(sequenceNumber, partitionKey, approximateArrivalTimestamp);
+        return getDefaultAttributes(sequenceNumber, subSequenceNumber, partitionKey, approximateArrivalTimestamp);
+    }
+
+    void closeSafe(final Closeable closeable, final String closeableName) {
+        if (closeable != null) {
+            try {
+                closeable.close();
+            } catch (final IOException e) {
+                getLogger().warn("Failed to close {} due to {}", closeableName, e.getLocalizedMessage(), e);
```
exceptionfactory: It is not necessary to repeat the exception message since it is included in the stack trace of the cause.
```suggestion
                getLogger().warn("Failed to close {}", closeableName, e);
```

dariuszseweryn:
Given the schema is widen on each next FlowFile produced by a batch – where split may happen multiple times in a single aggregated kinesis record – a subsequence number allows for better understanding on what exactly is in a FlowFile.

The writer schema is determined by the first record in a batch. A batch/shard is not necessarily composed of messages sharing the same schema but we can test this scenario as an edge case which will rather use other strategy than schema inference. Even if they share the schema, inference may assume a number is a 32 bit integer if it fits in range on the first record — all subsequent messages may contain numbers that won't fit as the type is 64 bit integer in reality.

Schema widening should happen only when schema is inferred — unfortunately we cannot identify this case programmatically without enhancing RecordSchema or RecordReader interfaces which would raise even more questions. Or maybe this behaviour should be guarded by a new property?

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: That is a bit of defensive programming in case someone would change the superclass without paying enough attention to subclasses. It should not happen

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: TL;DR: Apparently some writers are leaving the FlowFile in an inconsistent state after throwing exceptions (see example below). Since NiFi is a mature project, I assume that it is a known info and is handled by other processors.

Example:
first record: `{ "name": 10 }` / inferred schema `{ "name": INT }`
second record: `{ "name": 100000000000000 }` / inferred schema `{ "name": LONG }`

What WriteJsonResult leaves in FlowFile after throwing:
```
{ "name": 10 }
{ "name" }
```

Since the processor does not know how to read/fix the written file — we should fail the processing in this case. Therefore we would need to do one of:
- check/fix all the writers so they would not leave FlowFiles in inconsistent state — this can impact their performance
- add other abstraction for rewinding FlowFiles for fixing content
- check the schema before writing as proposed in this PR — there may be a different way to do it

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: Should all the record readers default to biggest number representations when inferring schema?

Or you suggest that RecordSchema/RecordReader would expose information if it was inferred and then the processor should do two passes over the record batch? First for creating a schema for all records and second for actual serialization?

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: Changing to `IllegalStateException`

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/kinesis/stream/record/TestKinesisRecordProcessorRecord.java
dariuszseweryn: Extracted declaration in function

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: I have reverted the schema widening and just kept the `REL_PARSE_FAILURE` routing. I will create a ticket about schema inference for numeric types. ([NIFI-14734](https://issues.apache.org/jira/browse/NIFI-14734))

Now I think that there is yet another way forward: To close the previous FlowFile and create the new one with the new schema — just remove the widening part. Edit: this still may work surprisingly if subsequent messages have only partially matching schemas.

dariuszseweryn:
Hm... Apparently `TestStandardRecordProcessorBlocker$TimeoutBlock#testBlockAndUnblock` is flaky. I will create a ticket to bump the timeout period to 5 seconds — in the meantime — could you re-run the tests?

Edit: Created ticket [NIFI-14747](https://issues.apache.org/jira/browse/NIFI-14747)

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -75,51 +77,79 @@ public KinesisRecordProcessorRecord(final ProcessSessionFactory sessionFactory,
...
+                return;
+            }
+            if (!flowFiles.contains(currentFlowFileState.flowFile)) {
+                // this is unexpected, flowFiles have been altered not in this class after the start of processing
+                throw new IllegalStateException("%s is not available in provided FlowFiles [%d]".formatted(currentFlowFileState.flowFile, flowFiles.size()));
+            }
+            completeFlowFile(flowFiles, session, stopWatch);
+        } catch (final FlowFileCompletionException e) {
+            if (!currentFlowFileState.containsDataFromExactlyOneKinesisRecord()) {
+                throw new KinesisBatchUnrecoverableException(e.getMessage(), e);
```
exceptionfactory: This should include a specific message, such as "FlowFile does not contain one data from one Kinesis Record"

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
```diff
@@ -75,51 +77,79 @@ public KinesisRecordProcessorRecord(final ProcessSessionFactory sessionFactory,
...
+            }
+            final boolean removeCurrentStateFlowFileIfAvailable = true;
```
exceptionfactory: If this is always `true`, it would be better to set it as a static final variable instead of a method local variable.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
exceptionfactory: Thanks for outlining the issues and describing the potential options.

Each Record Writer is responsible for its own FlowFile writing behavior. For this reason, attempting to resolve the issue here does not appear to be the best location. The JSON Record Set Writer may behave one way, but other Record Writers may be more lenient in certain cases.

In this case, it does not seem like the Processor should introduce schema compatibility checking, as that is the responsibility of the Record Writer.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: I'll refactor the code so it won't need this parameter

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: If so, I have created a ticket for it. ([NIFI-14753](https://issues.apache.org/jira/browse/NIFI-14753))

exceptionfactory: Thanks for creating the new ticket for consideration. I mentioned the concerns with the initial pull request approach, but there may be other options to consider.

At a general level, throwing an exception can and should end a FlowFile transaction, so the issue about having an inconsistent FlowFile content is not quite clear.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
exceptionfactory: Thanks for highlighting the schema inference issue, that sounds like a good thing to address separately.

Although attempting to create a new FlowFile when the schema changes is an interesting option, I agree with you that partially matching schemas could present other problems, so I don't recommend going in that direction.

At a base level, having this Processor attempt to compensate for schema mismatches does not seem like the right approach, and failures should be the expected behavior.

dariuszseweryn: I've already removed this code. No changes to the emitted FlowFile schema. Only a single FlowFile is emitted.

exceptionfactory (changes requested):
Reviewing the latest changes, I'm not clear on what elements remain necessary. Taking a step back, if the fundamental issue is the `IllegalTypeConversionException`, then it would be helpful to reduce the change set to that for clarity. This also relates to the comments in [NIFI-14753](https://issues.apache.org/jira/browse/NIFI-14753) regarding various options at a general level.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
exceptionfactory: Thanks for confirming.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/KinesisRecordProcessorRecord.java
dariuszseweryn: Actually there is another legitimate case that needs schema compatibility validation — schema encoded in the message payload, (see `ConfluentEncodedSchemaReferenceReader` and `GlueEncodedSchemaReferenceReader`). Each message read can have a different schema — logical solution would be to send out multiple FlowFiles with corresponding schemas in that case.

dariuszseweryn:
Taking a step back then. 

# Problems with `KinesisRecordProcessorRecord`

There are several functional problems with `KinesisRecordProcessorRecord`:
1. The class does not complete a FlowFile if there is any exception thrown when the last record in a batch is processed. This makes NiFi log a warning but it was observed that such FlowFile was retrieved by another processor downstream without any attributes.
1. The class does not handle well cases where schema of subsequent records differs especially when it is handled reactively because the exceptions thrown by the writers may leave the FlowFile in an inconsistent state — requiring a full rollback of transaction effectively halting the flow until human intervention. This is a problem in two situations:
    - When schema is inferred from the record itself — I can understand that it is considered "best effort" support, yet it is a problem from correctness standpoint
    - When schema is referenced in the record header via usage of `ConfluentEncodedSchemaReferenceReader` or `GlueEncodedSchemaReferenceReader` — which, I think, should be a valid, fully supported use-case. There are of-course several approaches to that which I will continue in a separate paragraph.
1. The class incorrectly calculates `record.count` attribute when multiple `intermediateRecord` are available in `kinesisClientRecord.data()` content for the last processed kinesis record in batch. (If there are 2 intermediate records in the last kinesis record, `record.count` is reported as a 1 bigger than it should, if 3 intermediate records = +3, if 3 = +6, 4 = +10 and so on).

There is a non-functional problem with `KinesisRecordProcessorRecord` — it contains and assumes a lot of state (FlowFile, RecordWriter, OutputStream, Session, contents of passed List<FlowFile>). Consolidation of state management is recommended for lowering cognitive burden for future readers. In this PR I have consolidated the state into `FlowFileState` class which is instantiated only in `#processRecord` function and encapsulates other state-altering interactions.

# Approaches to records that have incompatible schema

Currently the processor assumes the first kinesis record determines the schema for the whole FlowFile. If subsequent record schema does not match, depending on the mismatch case, it will either be not fully written (additional fields in subsequent records will get omitted) or an exception will get thrown (if field's content cannot be coerced, too large numeric value or a different data type) potentially leaving the FlowFile in inconsistent state (verified for `JsonRecordSetWriter`).

Given no guarantee that subsequent records will have matching case (due to wrong inference or schema encoding in header) the current behavior is not well suited. What can be done?

1. Add to contract of all RecordWriter implementations that in case of exception thrown the FlowFile is left in consistent state. This would need changes in potentially all implemented writers and as a way was rejected in [NIFI-14753](https://issues.apache.org/jira/browse/NIFI-14753). This would allow reactive approach to schema mismatches and routing offending records to Parse Failure (or rather introduce a new relationship) and continuing processing — otherwise the processing is halted without human intervention because the FlowFile needs to be dropped as it is in inconsistent state (I am unsure if readers downstream would be able to cope with a FlowFile in such state).
1. Add schema validation in `KinesisRecordProcessorRecord` for each subsequent record which would allow completing an in-flight FlowFile and opening a new one. It was [rejected](https://github.com/apache/nifi/pull/10053#discussion_r2206166722) though.
    - If there is a fear of performance impact we could add a parameter to the `ConsumeKinesisStream` processor whether it should expect multiple schemas in a single batch of Kinesis records and do the compatibility check. It doesn't look like a good way forward from the UX perspective as the reading logic would need to spill outside of the `RecordReader` definition. The processor does not have means to introspect the reader settings though, so it cannot auto-configure.
1. Make the processor do two passes over all `KinesisClientRecord` in batch. First to determine the encompassing schema, second to write data. At the first glance this approach seems surprising to me if I was a user of a stream that contains data in many schemas that the resulting FlowFile has schema that is an aggregation.

Personally I think that from the correctness point of view the processor should be able to close a currently processed FlowFile when the subsequent record's schema does not match and create a new FlowFile. This could have performance implications for users that have records with alternating schemas on each one — it should be technically correct implementation though. For users that have a single, static schema setup the only potential downside would be the performance impact of schema validation for each record — it may be negligible though anyway.

exceptionfactory:
Thanks for the substantive review of options and current discussion @dariuszseweryn, that is helpful framing.

Reviewing the options, and considering similar issues in the `ConsumeKafka` Processor, the use case related to an embedded schema is a good candidate for consideration. Solving that scenario could also benefit the schema inference scenario.

Here is an additional approach to consider:

The `ConsumeKafka` Processor does different output FlowFiles when records have different associated Schemas. However, instead of comparing whether a Record Schema is compatible with Schema from the first Record, it groups output based on Record Schema equality. It still sends invalid records to a `parse.failure` Relationship, but it provides a logical grouping based on common Record Schema references. This fits the scenario of embedded schema references, and can also work for scenarios where schema inference produces different Record Schema results for different Records. This approach would also avoid some of the performance concerns related to evaluate a `KinesisRecord` multiple times, and avoids the need for changes to Record Writers.

If that sounds viable, it is probably worth creating a new pull request, since it presents a more substantive direction than the initial focus here on improving error handling. However, it seems like such an approach gets closer to resolving the core issues.

dariuszseweryn:
> The ConsumeKafka Processor does different output FlowFiles when records have different associated Schemas. However, instead of comparing whether a Record Schema is compatible with Schema from the first Record, it groups output based on Record Schema equality. It still sends invalid records to a parse.failure Relationship, but it provides a logical grouping based on common Record Schema references. This fits the scenario of embedded schema references, and can also work for scenarios where schema inference produces different Record Schema results for different Records. This approach would also avoid some of the performance concerns related to evaluate a KinesisRecord multiple times, and avoids the need for changes to Record Writers.

The downside of such approach is that offset tracking gets impossible without per-row sequence/sub-sequence numbers. Was this considered for in `ConsumeKafka`?

exceptionfactory:
> > The ConsumeKafka Processor does different output FlowFiles when records have different associated Schemas. However, instead of comparing whether a Record Schema is compatible with Schema from the first Record, it groups output based on Record Schema equality. It still sends invalid records to a parse.failure Relationship, but it provides a logical grouping based on common Record Schema references. This fits the scenario of embedded schema references, and can also work for scenarios where schema inference produces different Record Schema results for different Records. This approach would also avoid some of the performance concerns related to evaluate a KinesisRecord multiple times, and avoids the need for changes to Record Writers.
> 
> The downside of such approach is that offset tracking gets impossible without per-row sequence/sub-sequence numbers. Was this considered for in `ConsumeKafka`?

Each Kafka `ConsumerRecord` has its own unique offset, so that works well in that scenario. Using the subsequence number for Kinesis sounds like it should provide the same level of tracking. I had initially raised a concern about adding the subsequence number in the context of addressing error handling, but considering the larger picture, including it to accomplish precise `KinesisRecord` tracking makes sense. With the subsequence number accounted for, do you see any other issues with Record Schema-based grouping?

dariuszseweryn:
There are two aspects that make me think:
- flows that assume sequential FlowFile contents
- humans auditing produced FlowFile contents completeness

# Flows assuming sequential FlowFile contents

All produced FlowFiles contain `aws.kinesis.sequence.number` of the last record. Given the flow uses FIFO or `aws.kinesis.sequence.number` as a prioritizer, subsequent FlowFiles for the same shard contained non-overlapping ranges of subsequent records under normal circumstances. 

Using grouping by default could break this assumption. To be backwards-compatible in this manner, the default strategy should be to close one FlowFile and start another when the schema changes and have grouping as an option.

# Humans auditing produced FlowFile contents completeness

I am considering auditability of processing completeness, as this processor does not work well with the Stateless Engine, i.e. it does not support Exactly Once semantics. Not all users will use the wrapping mechanism — having a way to determine/audit from the processed FlowFiles if all records were processed successfully could be useful.

Up till now, FlowFiles contained sequential records minus those that could not be parsed and went with Parsing Failure relationship. Incremental check of sequential FlowFiles and counting records between sequence numbers was enough to verify completeness.

With grouping it is harder to reason if all records were processed just by looking at subsequent FlowFiles attributes, as one would need to match the schema to find sequential FlowFiles and there is no guarantee how much back one would need to look for the previous FlowFile in sequence for a given schema. 

Easy ways for verification with grouping:
1. in wrapper mode — this is less of a problem, we have sequence/subsequence information on the processed records themselves
2. in non-wrapping mode — apart from having a sequence/subsequence number of the last record on each FlowFile, every FlowFile created from a single batch should have some identification of the batch it came from — sequence/subsequence number of the first record in the batch and count of FlowFiles produced by the batch. This would allow for easy grouping of FlowFiles produced by a batch and counting messages processed.

# TL;DR
There should be a new property on the processor, e.g. `Record Schema Change Handling`:
- `Create New FlowFile` (default) — close old FlowFile, create a new one
- `Group By FlowFile` — grouping similar to Kafka
Additionally, when transferring files to the Success Relationship, each FlowFile should have batch start sequence/subsequence number and count of FlowFiles emitted to the Success Relationship by processing the batch.

> If that sounds viable, it is probably worth creating a new pull request, since it presents a more substantive direction than the initial focus here on improving error handling. However, it seems like such an approach gets closer to resolving the core issues.

I will rephrase the original JIRA naming/description to match the core problem better (schema mismatches due to embedded schema) with additional error handling improvements since we cleared-out that that errors on writing FlowFiles should generally break a transaction.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/property/FlowFileHandlingOnSchemaChangeStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.aws.kinesis.property;
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum FlowFileHandlingOnSchemaChangeStrategy implements DescribedValue {
```
exceptionfactory: Recommend shortening this to something like `SchemaDifferenceHandlingStrategy`.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/property/FlowFileHandlingOnSchemaChangeStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+    ROLL_FLOW_FILES("Roll FlowFile", "Create a new FlowFile for each record with a different schema. The previous FlowFile will be completed with the records that have the previous schema. Emitted" +
+            " FlowFiles will contain continuous record sequences."),
+    GROUP_FLOW_FILES("Group FlowFiles", "Group records with the same schema into a single FlowFile. If a record with a different schema is encountered, a new FlowFile will be created for the new" +
```
exceptionfactory: This name does not quite fit the behavior, since it is grouping Records, instead of FlowFiles. The description is clear, so perhaps something like `Group Records By Schema`.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/statehandlerstrategy/StateHandlerStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.nifi.schema.access.SchemaNotFoundException;
+import org.apache.nifi.serialization.record.Record;
+import org.apache.nifi.serialization.record.RecordSchema;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+public abstract class StateHandlerStrategy<State, ActionContext, FinalizeException extends Throwable> {
```
exceptionfactory: The level of abstraction with the multiple generic types is a bit difficult to follow on initial read. The nested interfaces also seems to make this more complicated than necessary. Instead of having this abstract class and minimal implementations, it seems more straightforward to have a single class with a conditional for the minor difference in handling.

exceptionfactory (changes requested):
Thanks for working on the re-scoped direction @dariuszseweryn.

The general concept looks good, but the implementation seems more complicated than necessary with the multiple classes and nested interfaces. I recommend making adjustments to reduce the level of abstraction to move this forward.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/statehandlerstrategy/StateHandlerStrategy.java
dariuszseweryn: I have collapsed implementations into `StateHandlerStrategy` class and dropped "abstract" modifier.

Nested interfaces are needed due to checked exceptions that can be thrown from the actions on the state (initializing and completing) — I wanted to have a clean responsibility of `StateHandlerStrategy` and avoid leaking too many state initialization/completion details into it.

exceptionfactory: Thanks for consolidating the implementations, that is helpful.

The generic types do not seem necessary, since there is only one implementation of this class, so it would be helpful to simplify that abstraction as well.

dariuszseweryn: This change makes the encapsulation harder to maintain but removing types as requested.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/StateHandlerStrategy.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private FlowFileState getOrCreateNewState(final Record record, final BatchProcessingContext flowFileContext) throws IOException, SchemaNotFoundException {
+        final FlowFileState previousState = activeStateMap.get(record.getSchema());
+        if (previousState != null) {
+            return previousState;
+        }
+        final FlowFileState newState = stateInitializerAction.init(record, flowFileContext);
+        activeStateMap.put(record.getSchema(), newState);
+        return newState;
+    }
```
exceptionfactory: The two different methods appear to be the same, except for the `stateFinalizerAction` call, I recommend collapsing them and just checking for the `Roll FlowFiles` strategy, unless there is some other key difference I am missing.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/property/SchemaDifferenceHandlingStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.aws.kinesis.property;
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum SchemaDifferenceHandlingStrategy implements DescribedValue {
+    ROLL_FLOW_FILES("Roll FlowFile", "Create a new FlowFile for each record with a different schema. The previous FlowFile will be completed with the records that have the previous schema. Emitted" +
+            " FlowFiles will contain continuous record sequences."),
+    GROUP_FLOW_FILES("Group Records By Schema", "Group records with the same schema into a single FlowFile. If a record with a different schema is encountered, a new FlowFile will be created for" +
```
exceptionfactory: The enum name should align more closely to the label, such as `GROUP_RECORDS` or `GROUP_RECORDS_BY_SCHEMA`.
```suggestion
    GROUP_RECORDS("Group Records By Schema", "Group records with the same schema into a single FlowFile. If a record with a different schema is encountered, a new FlowFile will be created for" +
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/property/SchemaDifferenceHandlingStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.aws.kinesis.property;
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum SchemaDifferenceHandlingStrategy implements DescribedValue {
+    ROLL_FLOW_FILES("Roll FlowFile", "Create a new FlowFile for each record with a different schema. The previous FlowFile will be completed with the records that have the previous schema. Emitted" +
```
exceptionfactory: Perhaps `CREATE_FLOW_FILE` is clearer based on the description, and the enum name should be singular to match the label.
```suggestion
    CREATE_FLOW_FILE("Roll FlowFile", "Create a new FlowFile for each record with a different schema. The previous FlowFile will be completed with the records that have the previous schema. Emitted" +
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/ConsumeKinesisStream.java
```diff
@@ -319,6 +320,15 @@ public class ConsumeKinesisStream extends AbstractAwsAsyncProcessor<KinesisAsync
             .dependsOn(RECORD_WRITER)
             .build();
 
+    public static final PropertyDescriptor FLOW_FILE_HANDLING_ON_SCHEMA_CHANGE_STRATEGY = new PropertyDescriptor.Builder()
+            .name("FlowFile Handling On Schema Change Strategy")
```
exceptionfactory: The name can be adjusted along the lines of the enum.
```suggestion
            .name("FlowFile Handling On Schema Difference")
```

exceptionfactory (changes requested):
Thanks for making the adjustments @dariuszseweryn, this looks closer to completion, I noted a few more areas of refinement, which appeared to be leftovers from the previous abstraction.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/record/StateHandlerStrategy.java
dariuszseweryn: Collapsed

PR #10056, opened 2025-06-27
NIFI-14700 - Option to ignore parameter value changes in git registry clients

pvillard31:
# Summary

[NIFI-14700](https://issues.apache.org/jira/browse/NIFI-14700) - Option to ignore parameter value changes in git registry clients

In [NIFI-14555](https://issues.apache.org/jira/browse/NIFI-14555) we added the option to commit a new versioned flow with empty values for all parameters. This is a good strategy in case you want to make sure that you're not inadvertently committing a new version that would contain some parameter values that are specific to a lower environment.

However this strategy is not great when you have parameters where you want to specify a default value. In such a case, and when using the strategy implemented in [NIFI-14555](https://issues.apache.org/jira/browse/NIFI-14555), you'd lose the default values for your parameters when you commit a new version.

This JIRA is to add a new strategy allowing to commit a version where any already existing parameter is ignored. It means that we change the version we are committing by copying the existing values of the existing parameters into the new version. This is particularly useful as changing the value of a parameter is not considered as a local change so it can be confusing for a user to commit a new version where it is also including parameter value changes.

This is not a perfect solution because this strategy does not allow the user to change a default value that would have been set previously. But in that case the user can switch between the strategies at the registry client level to execute the right sequence of steps to get the expected end result.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-git-flow-registry/src/main/java/org/apache/nifi/registry/flow/git/AbstractGitFlowRegistryClient.java
@@ -26,6 +26,8 @@ import org.apache.nifi.flow.Position;
 import org.apache.nifi.flow.VersionedComponent;
 import org.apache.nifi.flow.VersionedConnection;
 import org.apache.nifi.flow.VersionedFlowCoordinates;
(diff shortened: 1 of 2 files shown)


exceptionfactory (approved):
Thanks for this improvement @pvillard31, the new `IGNORE_CHANGES` strategy looks like a useful and straightforward option for handling Parameter Context Values. +1 merging

PR #10058, opened 2025-06-30
NIFI-14702 Improve SplitExcel processor memory usage

zhtk:
# Summary

[NIFI-14702](https://issues.apache.org/jira/browse/NIFI-14702) Improve SplitExcel processor memory usage

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-14702) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


dan-s1:
@zhtk I think overall this is a great idea. I had noticed SplitExcel was sluggish but until you, I had not heard any complaints about it. What you have implemented is very similar to what POI has in `HSSFRow` with the method `copyRowFrom`. I had learnt about this when implementing [NIFI-14538](https://issues.apache.org/jira/browse/NIFI-14538). I have linked that ticket to your ticket. Hence I believe we should submit a ticket to POI for them to implement a similar method for XSSFSheet. I think you should give the justification you gave in the ticket that the current `copyRows` is a memory hog. If POI does implement this then I would say for you to rework this PR with that change.  I am waiting for their next release anyway for the changes I made for [NIFI-14538](https://issues.apache.org/jira/browse/NIFI-14538). Only in the event POI does not want to implement this method, I would then say go ahead with your code.
Also I just wanted to confirm that this PR only helps the memory issue for XSSF but not HSSF as all the changes you made are for XSSF and could not be applied to HSSF.
One more thing  I see which may have to be addressed at least in documentation is the notes given on the `SXSSFWorkbook` [javadoc](https://poi.apache.org/apidocs/dev/org/apache/poi/xssf/streaming/SXSSFWorkbook.html)

> Please note that there are still things that still may consume a large amount of memory based on which features you are using, e.g. merged regions, comments, ... are still only stored in memory and thus may require a lot of memory if used extensively. SXSSFWorkbook defaults to using inline strings instead of a shared strings table. This is very efficient, since no document content needs to be kept in memory, but is also known to produce documents that are incompatible with some clients. With shared strings enabled all unique strings in the document has to be kept in memory. Depending on your document content this could use a lot more resources than with shared strings disabled. Carefully review your memory budget and compatibility needs before deciding whether to enable shared strings or not.

zhtk:
@dan-s1, at first glance submitting a patch to POI seems to be a good idea. After some thought I have considerations to this approach. What we would like to achieve in POI is not just support for copying of rows, but memory-efficient copying of rows using StreamingReader. The problem is that StreamingSheet object returned by StreamingReader doesn't implement all methods from POI Sheet interface. I.e. in many cases it throws UnsupportedOperationException. As a result, such util method submitted to POI repository would be very sensitive to all refactorings and code simplifications, as the code using only streaming approach would be more complicated than code using random memory access. So we may easily end up with situation where after some refactor / change in POI we can't upgrade POI dependency in NiFi, because someone unknowingly broke copy method. One way to mitigate this issue could be to add test in POI, but this would introduce circular dependency: Apache POI -> pjfanning.xlsx.StreamingReader -> Apache POI tests. Not a great pattern.

Regarding enabling of shared strings in SXSSFWorkbook, I'm not sure how to approach this. On one hand enabling it may blow memory usage again, on the other hand breaking compatibility for some clients sounds bad. If only these clients were listed in Javadoc.

By the way, can you briefly explain why the code doesn't work for HSSF? I know for sure that it won't help with memory usage, because format of HSSF enforces to keep the whole file content in memory.

dan-s1:
@zhtk I am not understanding your hesitation. Isn't the copying code (i.e copying rows) from a StreamingSheet? We are not trying to call copy on a streaming object we are trying to copy from them which I thought would not touch the unsupported methods. This is what your changes have essentially done. I would assume in a streaming class this method would also not be supported.

> By the way, can you briefly explain why the code doesn't work for HSSF? I know for sure that it won't help with memory usage, because format of HSSF enforces to keep the whole file content in memory.

Currently SplitExcel does not support HSSF as you can see in the original code the streaming library is used which I understood supports only XSSF type files. In fact you had made the changes in `RowIterator` to only use the streaming API if the specified type was XLSX (which I believe is synonymous with XSSF). The changes you have made to use `SXSSFWorkbook` also is XSSF centric as that class only supports XSSF type files.

zhtk:
@dan-s1, perhaps I didn't understand your proposal. What is the scope of changes you propose to move to Apache POI? Note that in code I provided method signature changes and some parts of implementation were removed (e.g. ability to shift row numbers between source and destination sheets, some copy settings). As I understood, you ask to add  `copyRows(Sheet sourceSheet)` method to `SXSSFSheet` class, correct? Then the method implementation can ensure that it works correctly with SXSSFSheet destination, but it's hard to ensure that after some refactoring/changes to SXSSFSheet the method will work with parameter `sourceSheet` set to `StreamingSheet` class. The class coming from a different library, inheriting POI interfaces and implementing only subset of methods from `Sheet` interface. Does it make sense?

dan-s1:
@zhtk Yes my proposal is to have at least the logic from your `copyRowToSheet` although maybe just perhaps without the sheet and similar to what POI has for HSSF `copyRowFrom` for the `SXSSFSheet` class (if that is possible) and if it makes sense perhaps also for `XSSFSheet`.  BTW the author of the StreamingSheet (PJ Fanning) I believe is one of the main developers for POI so he intimately aware of both code bases.

pjfanning:
I don't want to get too involved here but we received a long email to the POI dev list - to which I would say that all you can do is trial and error. There are too many applications that read Excel format files to know how they all behave.
One suggestion here is that you make SXSSFWorkbook support optional. Users who need to use less memory (SXSSF) might end up with xlsx files that might be less compatible with their favourite spreadsheet app.
You could make whether SXSSFWorkbook uses shared strings another option. SXSSFWorkbook with shared strings will use more memory - but you could use https://github.com/pjfanning/poi-shared-strings to reduce the extra memory.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/excel/ExcelRowCopier.java
```diff
@@ -0,0 +1,71 @@
...
+        .cellFormula(false)
+        .cellStyle(CellCopyPolicy.DEFAULT_COPY_CELL_STYLE_POLICY)
+        .cellValue(CellCopyPolicy.DEFAULT_COPY_CELL_VALUE_POLICY)
+        .condenseRows(CellCopyPolicy.DEFAULT_CONDENSE_ROWS_POLICY)
+        .copyHyperlink(CellCopyPolicy.DEFAULT_COPY_HYPERLINK_POLICY)
+        .mergeHyperlink(CellCopyPolicy.DEFAULT_MERGE_HYPERLINK_POLICY)
+        .mergedRegions(false)
+        .rowHeight(CellCopyPolicy.DEFAULT_COPY_ROW_HEIGHT_POLICY)
+        .build();
+    private static final CellCopyContext CELL_COPY_CONTEXT = new CellCopyContext();
```
pjfanning: * you can't share the context as a static
* it is meant to be one context instance per copy job (eg if you are copying one sheet to another sheet, share the context across all the related copy cell calls) - you can't reuse that context if you are start on a new workbook

pjfanning: this is written in the javadoc - https://javadoc.io/static/org.apache.poi/poi/5.4.1/org/apache/poi/ss/usermodel/CellCopyContext.html

dan-s1:
@exceptionfactory My main concern is where it is implemented.

exceptionfactory:
> @exceptionfactory My main concern is where it is implemented.

Thanks for the clarification. On initial review, this seems much more narrow in scope that the similar discussion we had on copying rows for `HSSFRow`, so that's why I'm more inclined to proceed with the approach after reviewing the behavior and implementation details.

dan-s1:
Do we make `SXSSFWorkbook` support optional as pjfanning suggested? 
I assume either way we would need to have some additional details regarding this as I previously mentioned since there are some warning using `SXSSFWorkbook` as seen in the javadocs.

exceptionfactory:
> Do we make `SXSSFWorkbook` support optional as pjfanning suggested? I assume either way we would need to have some additional details regarding this as I previously mentioned since there are some warning using `SXSSFWorkbook` as seen in the javadocs.

If the capabilities are the same, then it seems like we can proceed with swapping it out, unless I am missing something.

zhtk:
@pjfanning, thanks too for detailed explanation. Fixed your comments.

@exceptionfactory, moved all logic from ExcelRowCopier back to processor.

@dan-s1, accordingly to the response received in mailing list (https://lists.apache.org/thread/o8fj0685pf115mnbzhk8rssscrtp8syr), the method was made to be similar to https://github.com/pjfanning/excel-streaming-reader/blob/main/src/test/java/com/github/pjfanning/xlsx/CopyToSXSSFUtil.java

I have adjusted SXSSFWorkbook to use shared strings to ensure compatibility with all clients. Tests with file described in the Jira ticket didn't show significant changes in memory usage. If we need to, we can optimise it later and take a look at https://github.com/pjfanning/poi-shared-strings. From my perspective, we can merge PR, it should give the same functionality as code before changes.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -127,7 +130,7 @@ public class SplitExcel extends AbstractProcessor {
             .condenseRows(CellCopyPolicy.DEFAULT_CONDENSE_ROWS_POLICY)
             .copyHyperlink(CellCopyPolicy.DEFAULT_COPY_HYPERLINK_POLICY)
             .mergeHyperlink(CellCopyPolicy.DEFAULT_MERGE_HYPERLINK_POLICY)
-            .mergedRegions(CellCopyPolicy.DEFAULT_COPY_MERGED_REGIONS_POLICY)
+            .mergedRegions(false)
```
exceptionfactory: What is the reason for changing this setting? Is it set to `false` because of the explicit merge region handling in the copyRows method?

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -167,21 +170,14 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
...
-                        List<Row> originalRows = new ArrayList<>();
-                        for (Row originalRow : originalSheet) {
-                            originalRows.add(originalRow);
-                        }
-
-                        if (!originalRows.isEmpty()) {
-                            newSheet.copyRows(originalRows, originalSheet.getFirstRowNum(), CELL_COPY_POLICY);
-                        }
+                    try (Workbook newWorkbook = new SXSSFWorkbook(null, SXSSFWorkbook.DEFAULT_WINDOW_SIZE, false, true)) {
+                        Sheet newSheet = newWorkbook.createSheet(originalSheetName);
```
exceptionfactory:
```suggestion
                        final Sheet newSheet = newWorkbook.createSheet(originalSheetName);
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -230,6 +226,29 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
         session.transfer(flowFileSplits, REL_SPLIT);
     }
 
+    public static int copyRows(final Sheet originalSheet, final Sheet destinationSheet) {
```
exceptionfactory: This should not be public, and doesn't necessarily need to be static.
```suggestion
    private int copyRows(final Sheet originalSheet, final Sheet destinationSheet) {
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -230,6 +226,29 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
...
+
+            for (final Cell sourceCell : sourceRow) {
+                final Cell destCell = destinationRow.createCell(sourceCell.getColumnIndex());
+                CellUtil.copyCell(sourceCell, destCell, CELL_COPY_POLICY, cellCopyContext);
+            }
+
+            ++rowCount;
+        }
+
+        for (CellRangeAddress sourceRegion : originalSheet.getMergedRegions()) {
```
exceptionfactory:
```suggestion
        for (final CellRangeAddress sourceRegion : originalSheet.getMergedRegions()) {
```

exceptionfactory (changes requested):
Thanks for the updates @zhtk, this looks much more focused now. I noted a few minor recommendations.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
zhtk: Yes, added a comment with explanation

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
zhtk: Done

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
zhtk: Good catch, fixed

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
zhtk: Done

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -127,7 +130,7 @@ public class SplitExcel extends AbstractProcessor {
             .condenseRows(CellCopyPolicy.DEFAULT_CONDENSE_ROWS_POLICY)
             .copyHyperlink(CellCopyPolicy.DEFAULT_COPY_HYPERLINK_POLICY)
             .mergeHyperlink(CellCopyPolicy.DEFAULT_MERGE_HYPERLINK_POLICY)
-            .mergedRegions(CellCopyPolicy.DEFAULT_COPY_MERGED_REGIONS_POLICY)
+            .mergedRegions(false) // NOTE: set to false because of the explicit merge region handling in the copyRows method.
```
pjfanning, line 136: so the support for mergedRegions in POI's CellUtils doesn't work generally or doesn't work specifically for SXSSF?

zhtk: It works, I just disabled it based on [this comment](https://github.com/apache/poi/blob/67320b6453d0d81fec13d9cf3cf4f1d1ba3ac73a/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFSheet.java#L3003):

    final CellCopyPolicy options = new CellCopyPolicy(policy);
    // avoid O(N^2) performance scanning through all regions for each row
    // merged regions will be copied after all the rows have been copied
    options.setCopyMergedRegions(false);

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -230,6 +226,29 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
...
+
+            ++rowCount;
+        }
+
+        for (final CellRangeAddress sourceRegion : originalSheet.getMergedRegions()) {
+            destinationSheet.addMergedRegion(sourceRegion.copy());
+        }
+
+        return rowCount;
+    }
```
dan-s1: @zhtk I understand you patterned this method after the code suggested in [CopyToSXSSFUtil](https://github.com/pjfanning/excel-streaming-reader/blob/main/src/test/java/com/github/pjfanning/xlsx/CopyToSXSSFUtil.java). What I do not understand is why there are the hyperlinks dealt with separately there but not here?

sfc-gh-ktrysinski: Hi @dan-s1, sorry for the delay - I had to take over this PR from @zhtk and only today I finally found some time.

I've tested the processor and yes - the code without explicit hyperlink copy did not work properly. I've changed it so now it works like in [CopyToSXSSFUtil](https://github.com/pjfanning/excel-streaming-reader/blob/main/src/test/java/com/github/pjfanning/xlsx/CopyToSXSSFUtil.java), and I've also added a unit test to make sure hyperlinks are copied correctly.

Kamilkime: Hi @dan-s1, sorry for the delay - I had to take over this PR from @zhtk and only today I finally found some time.

I've tested the processor and yes - the code without explicit hyperlink copy did not work properly. I've changed it so now it works like in [CopyToSXSSFUtil](https://github.com/pjfanning/excel-streaming-reader/blob/main/src/test/java/com/github/pjfanning/xlsx/CopyToSXSSFUtil.java), and I've also added a unit test to make sure hyperlinks are copied correctly.

sfc-gh-ktrysinski:
I bumped the `excel-streaming-reader` version to 5.1.1 (thanks @pjfanning for the fast release!). It fixes an issue with hyperlinks without location (without the `#` sign in the url). I also added a newer .xlsx file to test that.

Kamilkime:
I bumped the `excel-streaming-reader` version to 5.1.1 (thanks @pjfanning for the fast release!). It fixes an issue with hyperlinks without location (without the `#` sign in the url). I also added a newer .xlsx file to test that.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/processors/excel/TestSplitExcel.java
```diff
@@ -239,6 +241,38 @@ void testCopyDateTime() throws Exception {
...
+        runner.assertTransferCount(SplitExcel.REL_ORIGINAL, 1);
+        runner.assertTransferCount(SplitExcel.REL_FAILURE, 0);
+
+        final MockFlowFile flowFile = runner.getFlowFilesForRelationship(SplitExcel.REL_SPLIT).getFirst();
+        try (XSSFWorkbook workbook = new XSSFWorkbook(flowFile.getContentStream())) {
+            final Sheet sheet = workbook.getSheetAt(0);
+            assertEquals("Sheet1", sheet.getSheetName());
+
+            final List<XSSFHyperlink> hyperlinks = (List<XSSFHyperlink>) sheet.getHyperlinkList();
+            Assertions.assertIterableEquals(
```
exceptionfactory: The `import static` approach should be used instead of `Assertions`:
```suggestion
            assertIterableEquals(
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -230,6 +229,32 @@ public void onTrigger(ProcessContext context, ProcessSession session) throws Pro
...
+        for (final Row sourceRow : originalSheet) {
+            final Row destinationRow = destinationSheet.createRow(sourceRow.getRowNum());
+            destinationRow.setHeight(sourceRow.getHeight());
+
+            for (final Cell sourceCell : sourceRow) {
+                final Cell destCell = destinationRow.createCell(sourceCell.getColumnIndex());
+                CellUtil.copyCell(sourceCell, destCell, CELL_COPY_POLICY, cellCopyContext);
+            }
+
+            ++rowCount;
```
exceptionfactory: Is there a reason for the prefixed `++` as opposed to the suffix approach?
```suggestion
            rowCount++;
```

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
```diff
@@ -125,9 +131,9 @@ public class SplitExcel extends AbstractProcessor {
             .cellStyle(CellCopyPolicy.DEFAULT_COPY_CELL_STYLE_POLICY)
             .cellValue(CellCopyPolicy.DEFAULT_COPY_CELL_VALUE_POLICY)
             .condenseRows(CellCopyPolicy.DEFAULT_CONDENSE_ROWS_POLICY)
-            .copyHyperlink(CellCopyPolicy.DEFAULT_COPY_HYPERLINK_POLICY)
+            .copyHyperlink(false) // NOTE: the hyperlinks appear at end of sheet, so we need to iterate them separately at the end.
```
exceptionfactory, line 134: What does the comment mean that hyperlinks appear "at the end of the sheet"?

exceptionfactory (commented):
Thanks for the updates to this PR @Kamilkime. I plan to take another look at running behavior, but noted a couple more minor things. Otherwise this looks close to completion.

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
sfc-gh-ktrysinski: I don't think so, changed that one to suffix

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/processors/excel/TestSplitExcel.java
Kamilkime: Done

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
Kamilkime: I don't think so, changed that one to suffix

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
Kamilkime: I copied this comment from here: https://github.com/pjfanning/excel-streaming-reader/blob/main/src/test/java/com/github/pjfanning/xlsx/CopyToSXSSFUtil.java#L19

I think in the final sheet XML the hyperlinks are at the end of the file, after the standard content

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/test/java/org/apache/nifi/processors/excel/TestSplitExcel.java
```diff
@@ -26,9 +26,11 @@
 import org.apache.poi.ss.usermodel.DateUtil;
 import org.apache.poi.ss.usermodel.Row;
 import org.apache.poi.ss.usermodel.Sheet;
+import org.apache.poi.xssf.usermodel.XSSFHyperlink;
 import org.apache.poi.xssf.usermodel.XSSFSheet;
 import org.apache.poi.xssf.usermodel.XSSFWorkbook;
 import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
```
exceptionfactory:
```
src/test/java/org/apache/nifi/processors/excel/TestSplitExcel.java:[33,8] (imports) UnusedImports: Unused import - org.junit.jupiter.api.Assertions.
```

Kamilkime: Done

nifi-extension-bundles/nifi-poi-bundle/nifi-poi-services/src/main/java/org/apache/nifi/processors/excel/SplitExcel.java
exceptionfactory: Thanks for the reference.

exceptionfactory (approved):
Thanks for bringing this to completion @Kamilkime, and thanks for the collaboration @zhtk, @dan-s1, and @pjfanning! The latest version looks good. +1 merging

PR #10074, opened 2025-07-08
NIFI-14710 - Stateless should trigger failure callback when queues are not empty after process group is stopped

pvillard31:
# Summary

[NIFI-14710](https://issues.apache.org/jira/browse/NIFI-14710) - Stateless should trigger failure callback when queues are not empty after process group is stopped

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


sfc-gh-mpayne:
Hey @pvillard31 thanks for jumping in here. I think this solution is close to what we need but not exactly. The issue is that checking `isFlowFileQueued()` will return `true` if any data is queued up at all, including queued up in the root group's Output Ports. When a stateless group is run, it will queue up the data in output ports until it's been acknowledged, and I don't think we want to count these connections here. So instead of `isFlowFileQueued()` I think we want to use `rootGroup.isDataQueuedForProcessing()`

pvillard31:
Thanks @markap14 - very good point! Pushed a commit to use `isDataQueuedForProcessing()` instead

markap14 (approved):

PR #10079, opened 2025-07-10
NIFI-14713 short circuit getting rule violations

esecules:
# Summary

[NIFI-14713](https://issues.apache.org/jira/browse/NIFI-14713)
short circuit getting rule violations when there are no rule violations or when the user is getting violations for the root process group. All violations apply to the root process group by default. 

We need to short circuit the execution because otherwise it will recurse and visit all process groups in a flow. Added a patch from @tpalfy which greatly improves the performance when one of the other short cuts can't be taken.

Added a similar short circuit on the StandardFlowAnalyzer so that it's basically a no-op if there are no configured rules.

Also changed one of the unit tests for getting rule violations recursively so that it is distinct from getting all violations from the root of a tree of process groups. Added two more unit tests, one for no violations and one for checking the root process group.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

- n/a New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- n/a New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- n/a Documentation formatting appears as expected in rendered files

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
@@ -6846,8 +6846,9 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
     }
 
     public Stream<RuleViolation> getRuleViolationStream(String processGroupId) {
-        if (ruleViolationsManager.isEmpty())
+        if (ruleViolationsManager.isEmpty()) {
             return Stream.empty();
+        }
 
         ProcessGroup processGroup = processGroupDAO.getProcessGroup(processGroupId);
 


exceptionfactory (commented):
Thanks for putting this together @esecules. I plan to take a closer look soon, but on initial review, I noticed some style issues with the use of the wildcard import and missing braces in the if-else conditionals. If you can review those style issues, I can take a closer look at the implementation soon.

esecules:
> Thanks for putting this together @esecules. I plan to take a closer look soon, but on initial review, I noticed some style issues with the use of the wildcard import and missing braces in the if-else conditionals. If you can review those style issues, I can take a closer look at the implementation soon.

I see them and I'll make those changes now.

esecules:
FYI @exceptionfactory this is a subtask of a broader ticket for improving the performance of this endpoint. Adding these short circuits would be good enough for my use case and hopefully most others. I made a couple other suggestions on the ticket too. Where it falls short is on flows that have a ton of process groups but the tree of them is either very deep or doesn't widen out until after the root node. Either way, removing all the validation rules will provide an escape hatch so having the canvas open, especially at the root doesn't hog so many CPU cycles.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -6846,18 +6846,25 @@ public FlowAnalysisResultEntity getFlowAnalysisResult(String processGroupId) {
     }
 
     public Stream<RuleViolation> getRuleViolationStream(String processGroupId) {
```
esecules, line 6848: If I may check my understanding, @exceptionfactory, this method would require a read lock on the flow config for the duration of its execution because it's a get.* method in the StandardNifiServiceFacade?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -6846,18 +6846,25 @@ public FlowAnalysisResultEntity getFlowAnalysisResult(String processGroupId) {
...
-
-        Stream<RuleViolation> ruleViolationStreamForGroupAndAllChildren = Stream.concat(
-            ruleViolations.stream(),
-            processGroup.getProcessGroups().stream()
-                .map(ProcessGroup::getIdentifier)
-                .flatMap(this::getRuleViolationStream)
-        );
-
-        return ruleViolationStreamForGroupAndAllChildren;
+        if (ruleViolationsManager.isEmpty())
```
tpalfy: Do we need the comments? The code looks clear enough. (Maintaining comments is not only a hassle but doomed to degradation. Even local variables (with descriptive names) are better in that regard because - being part of the code - they are less likely to degrade.)

I personally prefer a single return statement but not a dealbreaker for me.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/StandardNiFiServiceFacadeTest.java
```diff
@@ -1076,13 +1076,90 @@ public void testGetRuleViolationsForGroupIsRecursive() throws Exception {
                 grandChildRuleViolation1, grandChildRuleViolation2, grandChildRuleViolation3
         ));
 
+        when(processGroupDAO.getProcessGroup("root")).thenReturn(processGroup);
+        reset(ruleViolationsManager);
+        when(ruleViolationsManager.isEmpty()).thenReturn(false);
```
tpalfy: These are not needed I think. 

```suggestion
```

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/StandardNiFiServiceFacadeTest.java
```diff
@@ -1076,13 +1076,90 @@ public void testGetRuleViolationsForGroupIsRecursive() throws Exception {
...
+                Arrays.asList(ruleViolation1, ruleViolation2)
+        );
+
+        Collection<RuleViolation> expected = new HashSet<>(Arrays.asList(
+                childRuleViolation1, childRuleViolation2,
+                grandChildRuleViolation1, grandChildRuleViolation2, grandChildRuleViolation3
+        ));
+
+        when(processGroupDAO.getProcessGroup(groupId)).thenReturn(processGroup);
+        when(ruleViolationsManager.isEmpty()).thenReturn(false);
```
tpalfy:
```suggestion
```
These are not needed. The first is taken care of by the `mockProcessGroup` call.
The second is default behaviour for mocks. Just increases boilerplate.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/StandardNiFiServiceFacadeTest.java
```diff
@@ -1076,13 +1076,90 @@ public void testGetRuleViolationsForGroupIsRecursive() throws Exception {
                 grandChildRuleViolation1, grandChildRuleViolation2, grandChildRuleViolation3
         ));
 
+        when(processGroupDAO.getProcessGroup("root")).thenReturn(processGroup);
```
tpalfy:
```suggestion
        when(processGroupDAO.getProcessGroup(FlowManager.ROOT_GROUP_ID_ALIAS)).thenReturn(processGroup);
```

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/StandardNiFiServiceFacadeTest.java
```diff
@@ -1076,13 +1076,90 @@ public void testGetRuleViolationsForGroupIsRecursive() throws Exception {
...
+        reset(ruleViolationsManager);
+        when(ruleViolationsManager.isEmpty()).thenReturn(false);
+        when(ruleViolationsManager.getAllRuleViolations()).thenReturn(expected);
+
+        // WHEN
+        Collection<RuleViolation> actual = serviceFacade.getRuleViolationStream(processGroup.getIdentifier()).collect(Collectors.toSet());
```
tpalfy, line 1082: Shouldn't we have a case for when we call `getRuleViolationStream` with `FlowManager.ROOT_GROUP_ID_ALIAS` instead of an ID?

```java
        // WHEN
        Collection<RuleViolation> actual = serviceFacade.getRuleViolationStream(FlowManager.ROOT_GROUP_ID_ALIAS).collect(Collectors.toSet());
```

tpalfy (commented):
In general I see the need for performance improvement. My issue is we are already going for one of the optional solutions without having decided which one we want in the end.
Because this change is not needed if we go with option 2. or option 3.

Do we want this as a quick-fix?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/StandardNiFiServiceFacadeTest.java
esecules: I will add one.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/StandardNiFiServiceFacadeTest.java
esecules: Thanks for the suggestions and explanation.

esecules:
> In general I see the need for performance improvement. My issue is we are already going for one of the optional solutions without having decided which one we want in the end. Because this change is not needed if we go with option 2. or option 3.
> 
> Do we want this as a quick-fix?

@tpalfy I was hoping for a quick fix that could make it to the next release that allows an "escape hatch" for the users like me ("there are dozens of us!") who have a data flow with a large number of process groups.

Personally it's not too urgent for me since we have a workaround with our API Gateway. It could be a problem for others.

I also wanted time to understand the write access patterns more before I make larger changes so I don't inadvertently make adding / deleting violations slower at the expense of reading and overall make it worse.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/StandardNiFiServiceFacadeTest.java
esecules: I'd need both for full branch coverage of the `else if()` block in `getRuleViolationStream`

esecules:
Working through contrib-check errors on my side. Stay tuned. 

Edit: errors fixed!

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -6846,18 +6846,23 @@ public FlowAnalysisResultEntity getFlowAnalysisResult(String processGroupId) {
     }
 
     public Stream<RuleViolation> getRuleViolationStream(String processGroupId) {
-        ProcessGroup processGroup = processGroupDAO.getProcessGroup(processGroupId);
```
esecules, line 6849: I have a question about the original implementation. Would it have allowed a user to view violations in a child process group of the current process group regardless of the level of permission the user had on that process group?

I see no evidence that it is making this check. Should it be?

If the answer is yes, I'll log another ticket instead of lumping the fix into this PR.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
esecules: Did some reading and experiments and I understand now that this does need a read lock.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
esecules: FYI the endpoint requires /flow access only and doesn't check anything else. Let's not change it in this PR but it's worth looking into since validation errors could expose something from a process group or processor config that wouldn't otherwise be visible.

esecules:
I see some unit test failures after integrating the patch. I will investigate. --- Fixed! Sorry, I am still not familiar with what runs as part of contrib-check. Otherwise I would have caught this before.

esecules:
The failing system test passed on the build on my fork of NiFi https://github.com/esecules/nifi/actions/runs/16232390927.

esecules:
@exceptionfactory @tpalfy please review when you get a chance. Thanks!

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/flowanalysis/StandardFlowAnalyzer.java
```diff
@@ -190,6 +190,7 @@ private void analyzeProcessGroup(
             Collection<RuleViolation> groupViolations,
             Map<VersionedComponent, Collection<RuleViolation>> componentToRuleViolations
     ) {
+        if (flowAnalysisRules.isEmpty()) return;
```
tpalfy:
```suggestion
        if (flowAnalysisRules.isEmpty()) {
            return;
        }
```

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -6846,18 +6846,33 @@ public FlowAnalysisResultEntity getFlowAnalysisResult(String processGroupId) {
     }
 
     public Stream<RuleViolation> getRuleViolationStream(String processGroupId) {
+        if (ruleViolationsManager.isEmpty())
+            return Stream.empty();
```
tpalfy:
```suggestion
        if (ruleViolationsManager.isEmpty()) {
            return Stream.empty();
        }
```

tpalfy (commented):
Looks good in general, just found 2 places with missing curly braces. In fact I thought they are supposed to cause checkstyle violation, even for one-liners.

In any case, if those are fixed (or declared to be acceptable) I can merge this.

tpalfy:
LGTM
Thank you for your work @esecules and for the review @exceptionfactory !
Merged into main.

PR #10081, opened 2025-07-10
NIFI-14737: Add Flow Analysis Rule to Require Load-Balanced Connections from a Source Processor

mattyb149:
# Summary

[NIFI-14737](https://issues.apache.org/jira/browse/NIFI-14737) This PR proposes a rule to enforce that all downstream connections from a source processor (no incoming connections) must be Load-Balanced in some way.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation


pvillard31 (changes requested):
I'm confused by this rule. Shouldn't it be only for source components configured to run on Primary node only? If not then, the data is likely already balanced and adding load balancing connection would impact performances. Consider a ConsumeKafka processor as an example.

And even in the case of Primary Node only, there are use cases where you don't want data to be load balanced across the nodes because you want to maintain ordering of the flow files.

Unless I'm missing something, I don't think this rule makes sense.

mattyb149:
Yes this is for Primary Node only source processors, I'll add that to the code. The use case is for non-expert flow designers to not design a flow for a cluster that actually only operates on the primary node.

mosermw:
I want to make sure this PR doesn't get into a place where it overlaps the NIFI-14204 PR #9685 which was rejected.

exceptionfactory (changes requested):
I concur with @pvillard31, this does not seem like a good candidate for inclusion in the project as a rule with general applicability. Can you provide any additional details on how this fits as a general use case @mattyb149?

mattyb149:
@pvillard31 mentioned one specific example (but I presume there are more), which doesn't suggest it's not a general use case. I will push a commit for Primary Node-only processors but wouldn't this be a generally applicable (and optional) rule which speaks to the idea of the rules engine being used for Best Practices (if you opt in)? It's doesn't bring in any extra significant  footprint.

exceptionfactory:
The example of `ConsumeKafka` is a good one, where running on the Primary Node is not necessarily desirable, and actually poor practice, because each node is responsible for its own partitions. This is the case with other types of event handling Processors as well. That's the reason why having this type of rule doesn't seem like helpful, because it would recommend something contrary to best practices.

mattyb149:
True, if each node is responsible for its own partitions. In other examples like fetching from a source database you wouldn't want each node to do the same work. But I would think this rule would be helpful and would only be enabled by an admin that wants to enforce it.

exceptionfactory:
> True, if each node is responsible for its own partitions. In other examples like fetching from a source database you wouldn't want each node to do the same work. But I would think this rule would be helpful and would only be enabled by an admin that wants to enforce it.

That's correct, and this rule would be a problem because it could not be applied in a selective way, if a flow had both of these scenarios together.

For this reason, I don't think it is a good candidate for inclusion.

nifi-registry/nifi-registry-core/nifi-registry-docker/dockerhub/sh/update_flow_provider.sh
pgyori, line 26: Although this change is absolutely necessary to make the script idempotent, I think it should be in its own commit, as I do not see how this is related to new flow analysis rule.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+                sourceProcessors.remove(connection.getDestination().getId());
+            }
+        });
+
+        processGroup.getConnections().forEach(connection -> {
+            if (connection.getSource().getId() != null) {
+                VersionedProcessor sourceProcessor = sourceProcessors.get(connection.getSource().getId());
+                if (sourceProcessor != null) {
+                    // Check if the connection is a Load-Balanced Connection
+                    if (Objects.equals(connection.getLoadBalanceStrategy(), "DO_NOT_LOAD_BALANCE")) {
```
pgyori: To make sure a potential null load balance strategy does not go unnoticed, I recommend:
`if (connection.getLoadBalanceStrategy() == null || "DO_NOT_LOAD_BALANCE".equals(connection.getLoadBalanceStrategy()))`

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.nifi.scheduling.ExecutionNode;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+
+@Tags({"connection", "source", "load", "balance"})
+@CapabilityDescription("Produces rule violation when a source processor does not have a Load-Balanced Connection (LBC) downstream"
```
pgyori: Can you please mention in the CapabilityDescription that it only applies to processors running on the Primary Node?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.nifi.flowanalysis.GroupAnalysisResult;
+import org.apache.nifi.scheduling.ExecutionNode;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+
+@Tags({"connection", "source", "load", "balance"})
```
pgyori: I think it is worth adding something like "primary" or "primarynodeonly" to the tags.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+            if (connection.getSource().getId() != null) {
+                VersionedProcessor sourceProcessor = sourceProcessors.get(connection.getSource().getId());
+                if (sourceProcessor != null) {
+                    // Check if the connection is a Load-Balanced Connection
+                    if (Objects.equals(connection.getLoadBalanceStrategy(), "DO_NOT_LOAD_BALANCE")) {
+                        // If not, we need to report this as a violation
+                        analysisResults.add(
+                                GroupAnalysisResult.forComponent(sourceProcessor,
+                                                connection.getIdentifier() + "_" + "LoadBalancedConnectionRequired",
+                                                "Source processors must configure their downstream connections to use a Load Balancing Strategy")
```
pgyori: Could you please update this message to something like:
"Source processors running on the Primary Node must configure ..."

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+        processGroup.getConnections().forEach(connection -> {
+            if (connection.getSource().getId() != null) {
+                VersionedProcessor sourceProcessor = sourceProcessors.get(connection.getSource().getId());
+                if (sourceProcessor != null) {
+                    // Check if the connection is a Load-Balanced Connection
+                    if (Objects.equals(connection.getLoadBalanceStrategy(), "DO_NOT_LOAD_BALANCE")) {
+                        // If not, we need to report this as a violation
+                        analysisResults.add(
+                                GroupAnalysisResult.forComponent(sourceProcessor,
+                                                connection.getIdentifier() + "_" + "LoadBalancedConnectionRequired",
```
pgyori: Nitpicking: the "_" + "LoadBalancedConnectionRequired" could simply be "_LoadBalancedConnectionRequired"

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+
+
+    @Override
+    public Collection<GroupAnalysisResult> analyzeProcessGroup(VersionedProcessGroup processGroup, FlowAnalysisRuleContext context) {
+
+        Collection<GroupAnalysisResult> analysisResults = new HashSet<>();
+
+        Map<String, VersionedProcessor> sourceProcessors = new HashMap<>();
+        for (VersionedProcessor processor : processGroup.getProcessors()) {
+            if (ExecutionNode.PRIMARY.toString().equals(processor.getExecutionNode())) {
```
pgyori: Instead of .toString() I recommend using
`ExecutionNode.PRIMARY.name()`

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+                sourceProcessors.put(processor.getIdentifier(), processor);
+            }
+        }
+        // Remove all processors that are not source processors or not running
+        processGroup.getConnections().forEach(connection -> {
+            if (connection.getDestination().getType() == ConnectableComponentType.PROCESSOR) {
```
pgyori: To be safe, I recommend a null check:
`if (connection.getDestination() != null && connection.getDestination().getType() == ConnectableComponentType.PROCESSOR)`

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+                sourceProcessors.remove(connection.getDestination().getId());
+            }
+        });
+
+        processGroup.getConnections().forEach(connection -> {
+            if (connection.getSource().getId() != null) {
```
pgyori: To be safe, I recommend a null check:
`if (connection.getSource() != null && connection.getSource().getId() != null)`

pgyori (changes requested):
Thank you @mattyb149  for the PR. Please check my review comments and let me know what you think.

nifi-registry/nifi-registry-core/nifi-registry-docker/dockerhub/sh/update_flow_provider.sh
exceptionfactory, line 26: This should not be included, it appears to be mistakenly included from a rebase

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
exceptionfactory: This needs to be updated to reflect the current behavior, Primary Node only

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Objects;
+
+@Tags({"connection", "source", "load", "balance"})
+@CapabilityDescription("Produces rule violation when a source processor does not have a Load-Balanced Connection (LBC) downstream"
+        + " in all outgoing connections")
+public class RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor extends AbstractFlowAnalysisRule {
+
+
+    @Override
+    public Collection<GroupAnalysisResult> analyzeProcessGroup(VersionedProcessGroup processGroup, FlowAnalysisRuleContext context) {
```
exceptionfactory:
```suggestion
    public Collection<GroupAnalysisResult> analyzeProcessGroup(final VersionedProcessGroup processGroup, final FlowAnalysisRuleContext context) {
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+
+        Collection<GroupAnalysisResult> analysisResults = new HashSet<>();
```
exceptionfactory:
```suggestion
        final Collection<GroupAnalysisResult> analysisResults = new HashSet<>();
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-rules/src/main/java/org/apache/nifi/flowanalysis/rules/RequireLoadBalancedConnectionAfterPrimaryNodeSourceProcessor.java
exceptionfactory: I recommend using a format string instead of concatenation

exceptionfactory (changes requested):
Concurring with feedback from @pgyori, I also noted a few other adjustments

mattyb149:
The build failures seem to be from an intermittently failing test:
`TestSocketLoadBalancedFlowFileQueue.testGetQueueSnapshotIncludesRebalancingPartition:788 expected: <0> but was: <4>`

pgyori (approved):
I think it looks good now. Thank you @mattyb149 !

PR #10097, opened 2025-07-16
NIFI-14243 add Load Balance Status to the Connections tab of the Summ…

markobean:
…ary page

# Summary

Add a column to the table on the Connections tab of the Summary page. The load balancing status is now indicated. Possible values are Not Configured, Active and Inactive. These states are defined in the nifi-api. The column is sortable and can be used as a filter.

This feature makes it possible to find all connections throughout a flow which have load balancing configured.

[NIFI-14243](https://issues.apache.org/jira/browse/NIFI-14243)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.html
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.html
@@ -196,7 +196,7 @@
                             <!-- Load Balance Status Column -->
                             <ng-container matColumnDef="loadBalanceStatus">
                                 <th mat-header-cell *matHeaderCellDef mat-sort-header>
-                                    <div class="overflow-ellipsis overflow-hidden whitespace-nowrap">Load Balance</div>
+                                    <div class="overflow-ellipsis overflow-hidden whitespace-nowrap">Load Balancing</div>
                                 </th>
                                 <td mat-cell *matCellDef="let item" [title]="formatLoadBalanceStatus(item)">
                                     {{ formatLoadBalanceStatus(item) }}
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.html
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.html
@@ -196,7 +196,7 @@
                             <!-- Load Balance Status Column -->
                             <ng-container matColumnDef="loadBalanceStatus">
                                 <th mat-header-cell *matHeaderCellDef mat-sort-header>
-                                    <div class="overflow-ellipsis overflow-hidden whitespace-nowrap">Load Balance Status</div>
+                                    <div class="overflow-ellipsis overflow-hidden whitespace-nowrap">Load Balance</div>
                                 </th>
                                 <td mat-cell *matCellDef="let item" [title]="formatLoadBalanceStatus(item)">
                                     {{ formatLoadBalanceStatus(item) }}
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.html
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.html
@@ -193,6 +193,16 @@
                                 </td>
                             </ng-container>
 
+                            <!-- Load Balance Status Column -->
+                            <ng-container matColumnDef="loadBalanceStatus">
+                                <th mat-header-cell *matHeaderCellDef mat-sort-header>
+                                    <div class="overflow-ellipsis overflow-hidden whitespace-nowrap">Load Balance Status</div>
+                                </th>
+                                <td mat-cell *matCellDef="let item" [title]="formatLoadBalanceStatus(item)">
+                                    {{ formatLoadBalanceStatus(item) }}
+                                </td>
+                            </ng-container>
+
                             <ng-container matColumnDef="actions">
                                 <th mat-header-cell *matHeaderCellDef></th>
                                 <td mat-cell *matCellDef="let item">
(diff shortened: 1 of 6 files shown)


markobean:
Now that 2.5.0 has been released, I rebased this PR to the current `main` updating to the latest version.

Also, I added a small change to re-word the column heading from "Load Balance Status" so simply "Load Balance". This takes up less horizontal real estate allowing additional column(s) to be added in the future without getting overly cramped.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.ts
```diff
@@ -124,12 +126,27 @@ export class ConnectionStatusTable extends ComponentStatusTable<ConnectionStatus
...
+            case 'LOAD_BALANCE_NOT_CONFIGURED':
+                return 'Not Configured';
+            case 'LOAD_BALANCE_ACTIVE':
+                return 'Active';
+            case 'LOAD_BALANCE_INACTIVE':
+                return 'Inactive';
+            default:
+                return 'unknown';
+
+            }
```
exceptionfactory: To make this a bit more dynamic, I recommend simply removing `LOAD_BALANCE_` from the `loadBalanceStatus` string. That way `unknown` is not possible.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.html
```diff
@@ -193,6 +193,16 @@
                                 </td>
                             </ng-container>
 
+                            <!-- Load Balance Status Column -->
+                            <ng-container matColumnDef="loadBalanceStatus">
+                                <th mat-header-cell *matHeaderCellDef mat-sort-header>
+                                    <div class="overflow-ellipsis overflow-hidden whitespace-nowrap">Load Balance</div>
```
exceptionfactory: Recommend labeling this `Load Balancing`:
```suggestion
                                    <div class="overflow-ellipsis overflow-hidden whitespace-nowrap">Load Balancing</div>
```

exceptionfactory (changes requested):
Thanks for the updates @markobean. Reviewing the frontend changes again, this looks straightforward. I recommended two minor adjustments, and otherwise I think it looks good.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.ts
nifiguy: I'm not certain what you mean by "more dynamic". I do not like simply striking `LOAD_BALANCE_` from the string leaving `NOT_CONFIGURED`, `ACTIVE`, and `INACTIVE`. This does not render nicely in the UI, i.e. the ALL CAPS presentation is not appropriate in the UI as it feels over emphasized. Similarly, to make it all lower case using `.toLowerCase()` isn't as elegant as a natural reading presentation having the first character capitalized; it also leaves behind an odd underscore in the case of "not_configured". 

Also, in reality, the `unknown` case should never occur by the nature of the `loadBalanceStatus` variable and its  underlying allowable values, but is necessary for the switch statement completeness. Since it should never occur, I deliberately left it as all lowercase such that it would stand out as inconsistent and suggest a "problem" if ever it presents itself in the future.

Bottom line, I'm thinking of appropriate presentation in the UI more than elegant and minimal lines of code.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.html
markobean: Done.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/summary/ui/connection-status-listing/connection-status-table/connection-status-table.component.ts
markobean: Oops - I was logged in to GitHub as a different user. The @nifiguy comments came from @markobean.

exceptionfactory: Thanks for clarifying. The application uses uppercased values in other places, but I agree that the title case rendering is more natural. On balance then, I recommend returning the `loadBalanceStatus` value itself, instead of `unknown`.

markobean: Good suggestion. I've updated to return `connection.connectionStatusSnapshot.loadBalanceStatus` in the default case.

exceptionfactory:
Thanks @markobean! Can you rebase on the current `main` branch? There is an issue with the GitHub Action for the `ci-workflow` that is now resolved.

markobean:
@exceptionfactory Sure thing. Rebase to latest `main` complete.

markobean:
> Reviewing the latest updates, this looks close to completion.
> 
> I noticed one minor issue that could lead to confusion related to filtering. `Active` and `Inactive` work, but filtering on `Not Configured` does not work because it doesn't match the enumerated value, but `NOT_CONFIGURED` does work. This is not the most intuitive behavior, given the displayed value is `Not Configured`. One option is to render `NOT_CONFIGURED` for the column value. Another option could be to replace the space character in the `filterTerm` with `_` to allow matching.

Seems reasonable. I modified the filtering for the loadBalanceStatus column so that it replaces spaces with underscores. Now, it will match the enum value. For good measure, I also trimmed the search term to eliminate leading and trailing whitespace that may inadvertently add erroneous underscores.

exceptionfactory (approved):
Thanks for adjusting the filter term handling @markobean, the latest version looks good! +1 merging

PR #10102, opened 2025-07-17
NIFI-14736 Add EOF attribute to ExecuteSQL for the last FlowFile output for a ResultSet

takraj:
# Summary

[NIFI-14736](https://issues.apache.org/jira/browse/NIFI-14736)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AbstractExecuteSQL.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AbstractExecuteSQL.java
@@ -247,7 +247,7 @@ public abstract class AbstractExecuteSQL extends AbstractSessionFactoryProcessor
     }
 
     protected void onTrigger(final ProcessContext context, final ProcessSession session, final ProcessSessionFactory sessionFactory) throws ProcessException {
-        final ExecuteSQLConfiguration config; // TODO remove
+        final ExecuteSQLConfiguration config;
 
         // Ensure that fileToProcess can only be accessed through resultSetFragments, because it might get removed
         // while outputting in batches, because the session needs to be committed.


exceptionfactory (changes requested):
Thanks for working on this improvement @takraj.

This pull request includes a large number of unrelated restructuring changes. Although those might be worth considering, they should be reverted. There is little value in sharing Property Descriptor references, and it often creates complexity as opposed to removing it.

Rescoping the pull request to the substance of the new attribute would put this in a much better position for potential review.

takraj:
@exceptionfactory Thank you for your comment. I've removed most of the unrelated changes, I hope this helps. Could you please check again?

takraj:
There is also a different approach to do this, by creating an extra 'sentinel' Flow File, in addition to the actual fragments. This would be then have no content, or empty record set, but would have the attribute. So downstream processors would know that the query is complete, when they process this empty file. What do you think?

exceptionfactory:
> There is also a different approach to do this, by creating an extra 'sentinel' Flow File, in addition to the actual fragments. This would be then have no content, or empty record set, but would have the attribute. So downstream processors would know that the query is complete, when they process this empty file. What do you think?

Thanks for the updates, the scoped-down set of changes are helpful.

I will have to take a closer look at the details, but on initial consideration, creating an extra empty FlowFile does not sound like a good approach, as it could disrupt existing flow expectations.

takraj:
> ... on initial consideration, creating an extra empty FlowFile does not sound like a good approach, as it could disrupt existing flow expectations.

@exceptionfactory Could you open up a little, in what sense would it distrupt the existing flow expectations?

We would also put it behind a switch, so `ExecuteSQL` and `ExecuteSQLRecord` would then both have a new property, like 'Add EOF Flow File' or 'Add Sentinel Flow File', with true/false settings, by default set to false. So unless it is intentionally turned on, these processors would behave the original way.

I have some concerns about the current approach, because it is a bit complicated with this Flow File delaying + migrating between sessions solution. Adding an extra Flow File is simpler, and simpler is usually better from maintenance perspective. What do you think?

exceptionfactory:
> > ... on initial consideration, creating an extra empty FlowFile does not sound like a good approach, as it could disrupt existing flow expectations.
> 
> @exceptionfactory Could you open up a little, in what sense would it distrupt the existing flow expectations?
> 
> We would also put it behind a switch, so `ExecuteSQL` and `ExecuteSQLRecord` would then both have a new property, like 'Add EOF Flow File' or 'Add Sentinel Flow File', with true/false settings, by default set to false. So unless it is intentionally turned on, these processors would behave the original way.
> 
> I have some concerns about the current approach, because it is a bit complicated with this Flow File delaying + migrating between sessions solution. Adding an extra Flow File is simpler, and simpler is usually better from maintenance perspective. What do you think?

@takraj Writing an empty FlowFile could be disruptive for existing flows that expect to receive FlowFiles containing at least one result. Some flow designs may be built to handle empty FlowFiles, others may not.

Adding a new property could be one way to implement this feature in a compatible way, so that could be considered. However, introducing new properties adds some level of complexity for the flow designer, which is more important to consider than code complexity.

With that being said, transferring FlowFiles between ProcessSessions raises some implementation questions. I need to take a closer look at the implementation details, but at the moment it is not clear why it is necessary to create more than one session. Is it not possible to set the attribute on the last of Result Set FlowFiles before transferring?

takraj:
@exceptionfactory 

> ... at the moment it is not clear why it is necessary to create more than one session. Is it not possible to set the attribute on the last of Result Set FlowFiles before transferring?

It is possible, but there is 'Output Batch Size' property. When this is set to a value, the created Flow Files need to be transferred, and the session is also need to be committed before we finish reading the records. So at the point, when we know there is no more, we might have already committed the session, so we cannot modify the attributes of the last Flow File.

My solution here attempts to address this by always holding back the last added Flow File, and excluding it from the mid-session commits. But this is only possible, if you temporarily remove it from the session to be committed. This guarantees that the very last Flow File will be only committed, when reading records have completed. So we can add the EOF property, because we already know there will be no more.

That's the basic concept of this PR. I'm open to alternative solutions, if you see anything better.

exceptionfactory:
> @exceptionfactory
> 
> > ... at the moment it is not clear why it is necessary to create more than one session. Is it not possible to set the attribute on the last of Result Set FlowFiles before transferring?
> 
> It is possible, but there is 'Output Batch Size' property. When this is set to a value, the created Flow Files need to be transferred, and the session is also need to be committed before we finish reading the records. So at the point, when we know there is no more, we might have already committed the session, so we cannot modify the attributes of the last Flow File.
> 
> My solution here attempts to address this by always holding back the last added Flow File, and excluding it from the mid-session commits. But this is only possible, if you temporarily remove it from the session to be committed. This guarantees that the very last Flow File will be only committed, when reading records have completed. So we can add the EOF property, because we already know there will be no more.
> 
> That's the basic concept of this PR. I'm open to alternative solutions, if you see anything better.

Thanks for clarifying the issues @takraj, that makes sense in terms of the `Output Batch Size` property.

Under the circumstances, I think the best option is to go with the FlowFile attribute approach (although I would adjust the name and not include the word `executesql`). However, if the `Output Batch Size` property is configured, then I would note in the description that this disables writing the FlowFile attribute.

takraj:
@exceptionfactory 

> although I would adjust the name and not include the word `executesql`

Most of other attributes, except the fragment id/index/count include this word. Is there any reason to remove it?

> However, if the Output Batch Size property is configured, then I would note in the description that this disables writing the FlowFile attribute.

Uhm. It doesn't, and it shouldn't. Currently this attribute is always added to the last Flow File, regardless to the `Output Batch Size` and `Max Rows Per Flow File` settings. I can scope it down to only add, if *both* of these have a value, if you think it is reasonable.

Currently the description of the attribute itself says:

> This attribute is added to the very last created FlowFile, to indicate end of results. This is primarily serving the downstream processors, when 'Output Batch Size' is set, in which case no other information is available about the total number of fragments created from a single result set.

takraj:
For reference, here is the alternative approach, that creates a sentinel Flow File with the attribute, when requested:
* https://github.com/apache/nifi/pull/10121

exceptionfactory:
> Uhm. It doesn't, and it shouldn't. Currently this attribute is always added to the last Flow File, regardless to the `Output Batch Size` and `Max Rows Per Flow File` settings. I can scope it down to only add, if _both_ of these have a value, if you think it is reasonable.
> 
> Currently the description of the attribute itself says:
> 
> > This attribute is added to the very last created FlowFile, to indicate end of results. This is primarily serving the downstream processors, when 'Output Batch Size' is set, in which case no other information is available about the total number of fragments created from a single result set.

Thanks for pointing that out, I was reading an earlier comment as thinking that this should be changed.

Regarding the `executesql` prefix, I see what you mean about some other attributes already using it, although there is a mixture of values.

Taking a second look, however, now I'm question whether this change is necessary at all. The Processor already writes `fragment.count` and `fragment.index` in this scenario, so simply comparing those values indicates the presence of the last FlowFile in a batch.

takraj:
@exceptionfactory 

> The Processor already writes fragment.count and fragment.index in this scenario, so simply comparing those values indicates the presence of the last FlowFile in a batch.

This is true, except for the case, when `Output Batch Size` is set. In that case, `fragment.count` is not added, because we don't know in advance how many Flow Files will be created.

Here is what the documentation says:

> If 'Max Rows Per Flow File' is set then this is the total number of FlowFiles produced by a single ResultSet. This can be used in conjunction with the fragment.identifier attribute in order to know how many FlowFiles belonged to the same incoming ResultSet. If Output Batch Size is set, then this attribute will not be populated.

exceptionfactory:
> This is true, except for the case, when `Output Batch Size` is set. In that case, `fragment.count` is not added, because we don't know in advance how many Flow Files will be created.

So for clarity, you are proposing to add this new FlowFile attribute to cover the case when `Output Batch Size` is set? Perhaps taking a step back and evaluating the core goal of the change would be helpful. It sounds like the new attribute is not necessary at all, the real issue is that the Result Set batching does not provide a way to know when the end of the batch is reached. Is that accurate? If that sounds correct, then the question is why this needs to be supported, as these two things are mutually exclusive for a reason.

takraj:
@exceptionfactory 

> So for clarity, you are proposing to add this new FlowFile attribute to cover the case when Output Batch Size is set?

Yes, this is the case.

> ... the real issue is that the Result Set batching does not provide a way to know when the end of the batch is reached. Is that accurate?

Almost. We don't know in advance how many fragments we will have, and result set batching outputs fragments while we are still fetching data from the DB. We only know the number of total fragments, when we have reached the very end of the result set.

exceptionfactory:
> @exceptionfactory
> 
> > So for clarity, you are proposing to add this new FlowFile attribute to cover the case when Output Batch Size is set?
> 
> Yes, this is the case.
> 
> > ... the real issue is that the Result Set batching does not provide a way to know when the end of the batch is reached. Is that accurate?
> 
> Almost. We don't know in advance how many fragments we will have, and result set batching outputs fragments while we are still fetching data from the DB. We only know the number of total fragments, when we have reached the very end of the result set.

Thanks for clarifying, that is very helpful.

Taking a step back then, the issue is not really adding the new FlowFile attribute, as it can be easily inferred.

As far as the optimal solution, did you consider calling `ResultSet.hasNext()` before calling `ProcessSession.commitAsync()`? That would indicate whether there are more rows to retrieve, or whether the last FlowFile created is actually the last FlowFile of the batch.

takraj:
@exceptionfactory 

> ... did you consider calling `ResultSet.hasNext()` before calling `ProcessSession.commitAsync()`?

I'm afraid this is not that simple, because `ResultSet` does not provide a way to determine if the cursor is standing at the last row. Your only option is to call `next()`, which also moves the cursor.

https://docs.oracle.com/en/java/javase/21/docs/api/java.sql/java/sql/ResultSet.html

exceptionfactory:
> @exceptionfactory
> 
> > ... did you consider calling `ResultSet.hasNext()` before calling `ProcessSession.commitAsync()`?
> 
> I'm afraid this is not that simple, because `ResultSet` does not provide a way to determine if the cursor is standing at the last row. Your only option is to call `next()`, which also moves the cursor.
> 
> https://docs.oracle.com/en/java/javase/21/docs/api/java.sql/java/sql/ResultSet.html

Good point, I was thinking of an `Iterator` initially without reviewing the documentation, but what about another method like `isAfterLast()`? It can depend on the JDBC implementation, but could be an option to consider.

takraj:
@exceptionfactory 

> ... what about another method like isAfterLast()? It can depend on the JDBC implementation, but could be an option to consider.

This sounds very risky, since it depends on the driver implementation. And there are also other concerns:

**Concern 1:** The result set is driven by the `SqlWriter`, like so: `sqlWriter.writeResultSet(resultSet, out, getLogger(), null)`. This means that cursor operations are delegated to another class, including end of resultset check. This is not the biggest problem, but it might complicate things further.

**Concern 2:** I'm not familiar with this entirely, but according to the current implementation, the result set might be fragmented in some way. If we reach the 'end' of the result set, there is a logic to get further results, which can mean we might or might not create further Flow Files. And moreover, at this point we might have already sent out the last Flow File. So this is something, that is pretty difficult to tackle with.

Source: https://github.com/apache/nifi/blob/d9b4984b1270f7659660e0ae58a52afe6a44f7e3/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/AbstractExecuteSQL.java#L444

```java
// are there anymore result sets?
try {
  hasResults = st.getMoreResults(Statement.CLOSE_CURRENT_RESULT);
  hasUpdateCount = st.getUpdateCount() != -1;
} catch (SQLException ex) {
  hasResults = false;
  hasUpdateCount = false;
}
```

Meanwhile, I have realized that, instead of the new EOF attribute, we could reuse the already existing `fragment.count`. The documentation says this is missing only in case of a set `Output Batch Size`. But with my change here, we could add this attribute to the FlowFiles in the very last batch. Practically meaning that at least the last FlowFile will contain information about the total number of fragments.

Now, since these `fragment.*` attributes are primarily for `Merge*` processors, if those can work with only one FlowFile having this information, we have already achieved our goal: The FlowFiles from `ExecuteSQL*` can be merged into a single FlowFile, even if `Output Batch Size` is set.

And this scenario is already supported by `MergeContent`, since its documentation says:

`fragment.count`
----------------------
> Applicable only if the <Merge Strategy> property is set to Defragment. This attribute indicates how many FlowFiles should be expected in the given bundle. **At least one FlowFile must have this attribute** in the bundle. If multiple FlowFiles contain the "fragment.count" attribute in a given bundle, all must have the same value.

Unfortunately `MergeRecord` doesn't, but it also could be an easy task to add.

This approach would be a compatible way to achieve what this ticket is about.

exceptionfactory (changes requested):
Thanks for the thorough reply @takraj.

On the latter part of your reply, I agree that using the existing `fragment.count` attribute is the right approach, obviating the need for a new FlowFile attribute as originally proposed in this pull request and Jira issue. I could also see updating `MergeRecord` separately if needed.

I remain stuck on the issue with the `Output Batch Size` property, and I don't think the current proposed approach in the pull request is the right way forward. I'm less concerned about the potential uncertainty around `ResultSet.isAfterLast()` unless this is not a common feature, which would require some evaluation.

The `Output Batch Size` is intended to release FlowFiles as quickly as possible, which is orthogonal keeping track of a batch of FlowFiles.

Given the changes in direction, regarding even the attribute itself, I think it would be better to revisit this in the Jira issue and evaluate options, then return to a new pull request once the proposal is clear.

exceptionfactory:
@takraj Following up on this, are you still evaluating an alternative approach? As mentioned, it may be best to continue discussion on the Jira issue.

takraj:
@exceptionfactory 

> Given the changes in direction, regarding even the attribute itself, I think it would be better to revisit this in the Jira issue and evaluate options, then return to a new pull request once the proposal is clear.

OK. Let's close this PR for now, and continue the design in the related Jira, or maybe in a new PR.

> are you still evaluating an alternative approach?

Not currently, but maybe you are right, and the design could be simplified by just using `isAfterLast()`, if support for this is common enough.

exceptionfactory:
Thanks @takraj, sounds good.

PR #10106, opened 2025-07-18
NIFI-14765 - Flow Registry Client policies break user endpoint

pvillard31:
# Summary

[NIFI-14765](https://issues.apache.org/jira/browse/NIFI-14765) - Flow Registry Client policies break user endpoint

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


exceptionfactory (changes requested):
Thanks for working on this issue @pvillard31.

Taking a closer look, I believe the existing `/controller` prefix needs to remain so that these two nested Resource Types could be managed through the parent policy on the Controller. That was the intent of the implementation.

I have reproduced the error, and I think the issue can be resolved with better parsing of the path hierarchy. I will look into the details further and may follow up with an alternative approach.

exceptionfactory:
@pvillard31 After further evaluation, I put together a new pull request that corrects the Resource Type resolution, preserving the current nested approach and correcting the lookup behavior. #10110

PR #10107, opened 2025-07-18
NIFI-14766 Implemented PMD's ExhaustiveSwitchHasDefault, ImplicitSwitchFallThrough and TooFewBranchesForSwitch rules

dan-s1:
# Summary

[NIFI-14766](https://issues.apache.org/jira/browse/NIFI-14766)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/c2/c2-protocol/c2-protocol-api/src/main/java/org/apache/nifi/c2/protocol/api/C2OperationState.java
+++ b/c2/c2-protocol/c2-protocol-api/src/main/java/org/apache/nifi/c2/protocol/api/C2OperationState.java
@@ -144,19 +144,13 @@ public class C2OperationState implements Serializable {
          * @return predefined ordinal
          */
         public static int toOrdinal(OperationState state) {
-            switch (state) {
-                case FULLY_APPLIED:
-                    return 0;
-                case PARTIALLY_APPLIED:
-                    return 1;
-                case OPERATION_NOT_UNDERSTOOD:
-                    return 2;
-                case NO_OPERATION:
-                    return 3;
-                case NOT_APPLIED:
-                default:
-                    return 4;
-            }
+            return switch (state) {
+                case FULLY_APPLIED -> 0;
+                case PARTIALLY_APPLIED -> 1;
+                case OPERATION_NOT_UNDERSTOOD -> 2;
+                case NO_OPERATION -> 3;
+                default -> 4;
+            };
         }
     }
 
--- a/minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/status/reporters/StatusLogger.java
+++ b/minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/status/reporters/StatusLogger.java
@@ -22,6 +22,8 @@ import static org.apache.nifi.minifi.commons.api.MiNiFiProperties.NIFI_MINIFI_ST
 import static org.apache.nifi.minifi.commons.api.MiNiFiProperties.NIFI_MINIFI_STATUS_REPORTER_LOG_QUERY;
 
 import java.io.IOException;
+import java.util.Objects;
+
 import org.apache.nifi.logging.LogLevel;
 import org.apache.nifi.minifi.bootstrap.QueryableStatusAggregator;
 import org.apache.nifi.minifi.bootstrap.status.PeriodicStatusReporter;
@@ -116,12 +118,10 @@ public class StatusLogger extends PeriodicStatusReporter {
                 }
 
             } catch (Exception e) {
-                switch (logLevel) {
-                    case ERROR:
-                        logger.error("Unexpected exception when attempting to report the status", e);
-                        break;
-                    default:
-                        logger.warn("Unexpected exception when attempting to report the status", e);
+                if (Objects.requireNonNull(logLevel) == LogLevel.ERROR) {
+                    logger.error("Unexpected exception when attempting to report the status", e);
+                } else {
+                    logger.warn("Unexpected exception when attempting to report the status", e);
                 }
             }
         }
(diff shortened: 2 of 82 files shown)


dan-s1:
@exceptionfactory Are there parts of the code which are not built when we build locally? Its weird as I built locally and did not have all these errors which I have seen the last few times.

exceptionfactory:
> @exceptionfactory Are there parts of the code which are not built when we build locally? Its weird as I built locally and did not have all these errors which I have seen the last few times.

What command are you running for a local build? There are optional build profiles, and some integration tests, which are not enabled by default, but which are enabled in the GitHub build.

dan-s1:
`./mvnw clean install -Pcontrib-check`

mosermw:
Do we need the TooFewBranchesForSwitch rule?  It seems like the solution creates less readable code.

exceptionfactory:
> Do we need the TooFewBranchesForSwitch rule? It seems like the solution creates less readable code.

I agree with @mosermw. I think this rule is a bit too opinionated for general purposes.

dan-s1:
Would it make sense to set the property `minimumNumberCaseForASwitch` from the default value of 3 to 2 since most of the changes I have made are as a result of the switch accommodating enums which only have two values.

exceptionfactory:
> Would it make sense to set the property `minimumNumberCaseForASwitch` from the default value of 3 to 2 since most of the changes I have made are as a result of the switch accommodating enums which only have two values.

If 3 aligns with current implementation, with just a few changes, that could be reasonable. If not, maybe 4.

dan-s1:
Right now the default is 3 hence resulting in the opinionated rule.  :).

exceptionfactory:
> Right now the default is 3 hence resulting in the opinionated rule. :).

Thanks, I misread the default setting. Yes, changing to 2 sounds like it could work if it reduced the scope of changes. What do you think @mosermw?

mosermw:
I agree `minimumNumberCaseForASwitch` of 2 seems reasonable.  A switch with 1 case shouldn't be using switch at all :)

exceptionfactory (changes requested):
Thanks for adjusting the rule scope and changes @dan-s1. The updates look good. If you can rebase and incorporate the recent updates to resolve the current branch conflict, this should be ready to go.

dan-s1:
@exceptionfactory 
I would like to address the original questions I had sent you by email  regarding the following classes which were flagged by PMD's ImplicitSwitchFallThrough rule.  I wanted to double check whether the logic in these classes are correct or perhaps they need to have break statements. The ones crossed out I believe are not issues. `RetryFlowFile` just set a different log level but uses the same log statement hence the fallthrough.  The `CastEvaluator` family of classes all  have a switch statement in the previous case statement which covers all cases and each case has a return statement (in fact I am not sure why PMD is flagging these). Its just the other ones I am just not sure whether the logic is correct or not

- org.apache.nifi.controller.serialization.AffectedComponentSet after line 550 there is no break statement.
- org.apache.nifi.processors.standard.ControlRate after line  280 there is no break statement.
- org.apache.nifi.processors.standard.TailFile after line 1034 there is no break statement.
- org.apache.nifi.authorization.StandardAuthorizableLookup after line 592 there is no break statement.
- ~~org.apache.nifi.processors.standard.RetryFlowFile after line 248 there is no break statement.~~
- ~~org.apache.nifi.attribute.expression.language.evaluation.cast.WholeNumberCastEvaluator after line 72 there is  no break statement.~~
- ~~org.apache.nifi.attribute.expression.language.evaluation.cast.DecimalCastEvaluator after line 70 there is no break statement.~~
- ~~org.apache.nifi.attribute.expression.language.evaluation.cast.NumberCastEvaluator after line 77 there is no break statement.~~

Please advise. Thanks!

dan-s1:
> I agree `minimumNumberCaseForASwitch` of 2 seems reasonable. A switch with 1 case shouldn't be using switch at all :)

@mosermw You would be surprised how many switch statements there were in the code base that only had one case statement.

mosermw:
> I wanted to double check whether the logic in these classes are correct or perhaps they need to have break statements. 
> 
> * org.apache.nifi.controller.serialization.AffectedComponentSet after line 550 there is no break statement.
> * org.apache.nifi.processors.standard.ControlRate after line  280 there is no break statement.
> * org.apache.nifi.processors.standard.TailFile after line 1034 there is no break statement.
> * org.apache.nifi.authorization.StandardAuthorizableLookup after line 592 there is no break statement.
> 
> Please advise. Thanks!

I had a look at these:

- I think AffectedComponentSet is missing a break statement there. I'm not sure the case ever happens, or if it does happen then falling through has no negative effects, but I think a break is intended.
- For ControlRate it appears the fallthrough is intentional, because the "Maximum Rate" property must be filled in when using either of DATA_RATE, FLOWFILE_RATE or ATTRIBUTE_RATE.  Having a comment there is good to make that clear.
- After reading the javadoc for TailFile, it appears the fallthrough is intentional.  Having a comment there is good to make that clear.
- For StandardAuthorizableLookup, I don't think it's possible to reach the end of the case, because a return or thrown exception is the only path out.  Having a break statement there probably won't hurt if it makes PMD happy.

dan-s1:
@mosermw Thank you for commenting on this! It is much appreciated.

exceptionfactory (approved):
Thanks @dan-s1, and thanks for the feedback and confirmation @mosermw, the latest version looks good. +1 merging

dan-s1:
@exceptionfactory I was actually working to address @mosermw comments. What should I do with those changes?

exceptionfactory:
> @exceptionfactory I was actually working to address @mosermw comments. What should I do with those changes?

Thanks @dan-s1, sorry for the misunderstanding. If you can open a new PR with those changes, that should be fine.

dan-s1:
@exceptionfactory All good! Thank you!

PR #10115, opened 2025-07-21
NIFI-14770: Allowing the user to reset all counters in a single action.

mcgilman:


mcgilman:
> I reviewed this from a UI perspective...
> 
> The concern that I have is that the "Reset All" button and dialog might be a bit ambiguous when the list of Counters has been filtered. The users might assume that only the Counters that match the filter criteria will be reset. Maybe the description in the confirmation dialog should indicate the total number of Counters getting reset?
> 
> I suppose another, more involved option would be to give the user the option to reset only the "filtered" Counters in addition to all. Not sure how beneficial that might be however. Might be enough to just make it more clear in the dialog that you are resetting `X` number of Counters.

Thanks for reviewing @rfellows! I've pushed another commit that incorporates the total count into the reset prompt. Happy to tweak the language if you have further suggestions.

exceptionfactory (commented):
Thanks for working on this @mcgilman! On a quick review of the REST API, I was wondering about the potential use of the `DELETE` method instead of `PUT`, but given the behavior, `PUT` looks good. I suppose `DELETE` would work as a way to actually delete existing counters, but that is a more destructive, so this looks like the right way to go.

rfellows (approved):
Thanks for the changes @mcgilman. This is a nice productivity boosting feature.

PR #10123, opened 2025-07-22
NIFI-14778: Refactoring location of ProcessGroupReferences

mcgilman:
- Move ProcessGroupReferences into top level common components directory so that it is readily consumable throughout.
- Changes here are mostly around keeping the imports updated.


scottyaslan:
Reviewing...

scottyaslan (approved):

PR #10128, opened 2025-07-23
NIFI-14666: Add hyperlink to component ID in Action Details view

hsahu-ksolves123:
**Summary**
This PR implements [NIFI-14666](https://issues.apache.org/jira/browse/NIFI-14666), adding a clickable hyperlink to the **component ID** shown in the **Action Details** view of the Flow Configuration History page.

**Changes**

- Updated `action-details.component.html` to wrap the component ID in an anchor tag.
- Uses `routerLink` to enable navigation directly to the component in the UI.

 **Verification**

- Verified that the hyperlink renders correctly and navigates to the intended component.
- No regressions observed in the flow configuration history view.
Screenshot from 2025-07-24 01-41-49

Let me know if any improvements or test coverage updates are needed.


hsahu-ksolves123:
@exceptionfactory Thank you for the feedback. You're right - hardcoding 'root' won't work for components in nested Process Groups.
I'm investigating the action data structure to find the correct Process Group ID and implement dynamic routing.

hsahu-ksolves123:
I've made the requested changes according to the suggestions. Please review the latest commits and let me know if any further updates are needed. Thanks!

exceptionfactory (changes requested):
The hyperlink appears to be unchanged, it still references the root group. In addition, all `console.log()` statements should be removed, and all changes to `package-lock.json` should be reverted.

exceptionfactory (commented):
@hsahu-ksolves123 Given that the latest changes are simply adding log statements, and the fundamental approach is not correct, I am going to close this pull request for now. If you have an approach that works for the correct Process Group ID, feel free to open a new pull request scoped to the appropriate changes.

PR #10134, opened 2025-07-24
NIFI-14787 Implemented code improvements to avoid PMD's ImplicitSwitchFallThrough rule.

dan-s1:
# Summary

[NIFI-14787](https://issues.apache.org/jira/browse/NIFI-14787)
This PR involved adding the suggestions made in PR #10107 which are

- org.apache.nifi.controller.serialization.AffectedComponentSet after line 550 added a break statement.
- org.apache.nifi.processors.standard.ControlRate after line 280 added a comment why there is no break statement.
org.apache.nifi.processors.standard.TailFile after line 1034 added a comment why there is no break statement.
org.apache.nifi.authorization.StandardAuthorizableLookup refactored to remove the nested switch statement into its own method in order to make the outer switch easier to read and thereby avoiding any violation of the  PMD ImplicitSwitchFallThrough rule.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
@@ -569,8 +569,7 @@ public class StandardAuthorizableLookup implements AuthorizableLookup {
             case DataTransfer:
             case ProvenanceData:
             case Operation:
-                final Authorizable authorizable = handleResourceTypeContainingOtherResourceType(resource, resourceType);
-                return authorizable;
+                return handleResourceTypeContainingOtherResourceType(resource, resourceType);
             case Controller:
                 // Handle Nested Resource Types such as Flow Analysis Rules and Flow Registry Clients
                 final ResourceType nestedResourceType;
@@ -623,7 +622,7 @@ public class StandardAuthorizableLookup implements AuthorizableLookup {
             case DataTransfer -> new DataTransferAuthorizable(getAccessPolicy(baseResourceType, resource));
             case ProvenanceData -> new ProvenanceDataAuthorizable(getAccessPolicy(baseResourceType, resource));
             case Operation -> new OperationAuthorizable(getAccessPolicy(baseResourceType, resource));
-            default -> null;
+            default -> throw new IllegalArgumentException("Cannot use resource type %s in this method".formatted(resourceType));
         };
     }
 
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
@@ -569,28 +569,8 @@ public class StandardAuthorizableLookup implements AuthorizableLookup {
             case DataTransfer:
             case ProvenanceData:
             case Operation:
-
-                // get the resource type
-                final String baseResource = StringUtils.substringAfter(resource, resourceType.getValue());
-                final ResourceType baseResourceType = ResourceType.fromRawValue(baseResource);
-
-                if (baseResourceType == null) {
-                    throw new ResourceNotFoundException("Unrecognized base resource: " + resource);
-                }
-
-                switch (resourceType) {
-                    case Policy:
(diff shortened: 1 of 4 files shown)


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
```diff
@@ -569,28 +569,8 @@ public Authorizable getAuthorizableFromResource(final String resource) {
...
-                    case DataTransfer:
-                        return new DataTransferAuthorizable(getAccessPolicy(baseResourceType, resource));
-                    case ProvenanceData:
-                        return new ProvenanceDataAuthorizable(getAccessPolicy(baseResourceType, resource));
-                    case Operation:
-                        return new OperationAuthorizable(getAccessPolicy(baseResourceType, resource));
-                }
-            // fallthrough
+                final Authorizable authorizable = handleResourceTypeContainingOtherResourceType(resource, resourceType);
+                return authorizable;
```
exceptionfactory: It looks like this could be collapsed into a single line with a return.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
```diff
@@ -628,6 +608,25 @@ public Authorizable getAuthorizableFromResource(final String resource) {
...
+            throw new ResourceNotFoundException("Unrecognized base resource: " + resource);
+        }
+
+        return switch (resourceType) {
+            case Policy -> new AccessPolicyAuthorizable(getAccessPolicy(baseResourceType, resource));
+            case Data -> new DataAuthorizable(getAccessPolicy(baseResourceType, resource));
+            case DataTransfer -> new DataTransferAuthorizable(getAccessPolicy(baseResourceType, resource));
+            case ProvenanceData -> new ProvenanceDataAuthorizable(getAccessPolicy(baseResourceType, resource));
+            case Operation -> new OperationAuthorizable(getAccessPolicy(baseResourceType, resource));
+            default -> null;
```
exceptionfactory: Did the previous fall through case return `null` as well? It seems like this should throw an exception.

exceptionfactory (commented):
Thanks for coming back to wrap this up @dan-s1. I noted one question on the Lookup class change.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
dan-s1: Sure I can do that although from past reviews I thought you seem to prefer having a variable declared.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
dan-s1: It did not. I initially did not have a `default` and the compiler complained that something has to be returned so I put a return null after the switch statement. Then Intellij suggested to refactor to an enhanced switch statement whereby it inserted the default with a return of null. Honestly this method should never be used except for in the five cases where a resource type can have another resource type so the default should never be reached. Its only there to make the compiler happy.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
exceptionfactory: There are different preferences, I actually prefer the one-line return, others prefer it declared. In general I would ignore it, but I highlighted it here since there is nothing else in the code block.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/authorization/StandardAuthorizableLookup.java
exceptionfactory: Thanks, I was thinking it was a rare scenario. It seems like throwing an `IllegalArgumentException` would help clarify that this should not happen, what do you think?

PR #10136, opened 2025-07-25
NIFI-14789 In PutSQL send incoming invalid FlowFiles to failure.

tpalfy:
# Summary

[NIFI-14789](https://issues.apache.org/jira/browse/NIFI-14789)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
@@ -32,6 +32,7 @@ import org.apache.nifi.components.ValidationResult;
 import org.apache.nifi.dbcp.DBCPService;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
+import org.apache.nifi.flowfile.attributes.CoreAttributes;
 import org.apache.nifi.flowfile.attributes.FragmentAttributes;
 import org.apache.nifi.processor.AbstractSessionFactoryProcessor;
 import org.apache.nifi.processor.FlowFileFilter;
@@ -64,6 +65,7 @@ import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
@@ -78,6 +80,9 @@ import static java.lang.String.valueOf;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
 import static java.util.stream.Collectors.toList;
+import static org.apache.nifi.processor.FlowFileFilter.FlowFileFilterResult.ACCEPT_AND_CONTINUE;
+import static org.apache.nifi.processor.FlowFileFilter.FlowFileFilterResult.ACCEPT_AND_TERMINATE;
+import static org.apache.nifi.processor.FlowFileFilter.FlowFileFilterResult.REJECT_AND_CONTINUE;
 import static org.apache.nifi.processor.util.pattern.ExceptionHandler.createOnError;
 
 @SupportsBatching
@@ -700,17 +705,32 @@ public class PutSQL extends AbstractSessionFactoryProcessor {
         boolean fragmentedTransaction = false;
 
         final int batchSize = context.getProperty(BATCH_SIZE).asInteger();
-        final FlowFileFilter dbcpServiceFlowFileFilter = context.getProperty(CONNECTION_POOL).asControllerService(DBCPService.class).getFlowFileFilter(batchSize);
+        final DBCPService dbcpService = context.getProperty(CONNECTION_POOL).asControllerService(DBCPService.class);
+        final FlowFileFilter batchingDbcpServiceFlowFileFilter = dbcpService.getFlowFileFilter(batchSize);
+        final FlowFileFilter nonbatchingDbcpServiceFlowFileFilter = dbcpService.getFlowFileFilter();
+
         List<FlowFile> flowFiles;
         if (useTransactions) {
-            final TransactionalFlowFileFilter filter = new TransactionalFlowFileFilter(dbcpServiceFlowFileFilter);
+            final TransactionalFlowFileFilter filter = new TransactionalFlowFileFilter(batchingDbcpServiceFlowFileFilter);
             flowFiles = session.get(filter);
             fragmentedTransaction = filter.isFragmentedTransaction();
         } else {
-            if (dbcpServiceFlowFileFilter == null) {
+            if (batchingDbcpServiceFlowFileFilter == null) {
                 flowFiles = session.get(batchSize);
             } else {
-                flowFiles = session.get(dbcpServiceFlowFileFilter);
+                PutSQLFlowFileFilter flowFileFilter = new PutSQLFlowFileFilter(
+                        batchingDbcpServiceFlowFileFilter,
+                        nonbatchingDbcpServiceFlowFileFilter
+                );
+
+                List<FlowFile> filteredFlowFiles = session.get(flowFileFilter);
+
+                if (flowFileFilter.isCollectingInvalidFlowFiles()) {
+                    result.routeTo(filteredFlowFiles, REL_FAILURE);
+                    flowFiles = Collections.emptyList();
+                } else {
+                    flowFiles = filteredFlowFiles;
+                }
             }
         }
 
@@ -934,7 +954,7 @@ public class PutSQL extends AbstractSessionFactoryProcessor {
                 if (fragmentId == null || "1".equals(fragCount)) {
                     return filterNonFragmentedTransaction(flowFile);
                 } else {
-                    return FlowFileFilterResult.REJECT_AND_CONTINUE;
+                    return REJECT_AND_CONTINUE;
                 }
             }
 
@@ -945,7 +965,7 @@ public class PutSQL extends AbstractSessionFactoryProcessor {
                     return filterNonFragmentedTransaction(flowFile);
                 } else {
                     // we've already selected 1 FlowFile, and this one doesn't match.
-                    return FlowFileFilterResult.REJECT_AND_CONTINUE;
+                    return REJECT_AND_CONTINUE;
                 }
             }
 
@@ -967,14 +987,14 @@ public class PutSQL extends AbstractSessionFactoryProcessor {
 
                 if (numSelected >= numFragments - 1) {
                     // We have all of the fragments we need for this transaction.
-                    return FlowFileFilterResult.ACCEPT_AND_TERMINATE;
+                    return ACCEPT_AND_TERMINATE;
                 } else {
                     // We still need more fragments for this transaction, so accept this one and continue.
                     numSelected++;
                     return FlowFileFilterResult.ACCEPT_AND_CONTINUE;
                 }
             } else {
-                return FlowFileFilterResult.REJECT_AND_CONTINUE;
+                return REJECT_AND_CONTINUE;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


tpalfy:
Thank you for the feedback @exceptionfactory.

As for the `getFlowFileFilter()` without batch size - I'm not sure what the question is about.
- The `getFLowFileFilter(batchSize)` will delegate to the `getFlowFileFilter()`.
- As far as the explicit call in my implementation - it's only when collecting the invalid FlowFiles. The `Batch Size` property is supposed to be controlling the database side after all. (As per the property's description: "The preferred number of FlowFiles to put to the database in a single transaction").

But more importantly, do I understand correctly, that we shouldn't change the `PutSQL`, but change `DBCPConnectionPoolLookup.getFlowFileFilter()`? (That's what concretely initiated this issue.)
How should `DBCPConnectionPoolLookup.getFlowFileFilter()` work in that case? If the FlowFile simply returns a REJECT_AND_CONTINUE, it will still end up staying in the incoming queue indefinitely.

I want to add that in the case of `ExecuteSQL` for example, in this scenario the invalid FlowFile does get sent to failure.

exceptionfactory:
Thanks for the reply @tpalfy.

Regarding the `getFlowFileFilter()` method calls, I was referring to this:

```
final FlowFileFilter batchingDbcpServiceFlowFileFilter = dbcpService.getFlowFileFilter(batchSize);
final FlowFileFilter nonbatchingDbcpServiceFlowFileFilter = dbcpService.getFlowFileFilter();
```

I believe only the first one is needed, as is currently used.

Regarding where to make changes, yes, keeping `PutSQL` unchanged, but changing `DBConnectionPoolLookup` seems like the right place to make improvements.

I see the current behavior is to throw a `ProcessException` if the `database.name` attribute is missing. That appears to be wrong, as mentioned, given that a `FlowFileFilter` should not throw exceptions. As you noted, returning `REJECT_AND_CONTINUE` keeps the FlowFile in the queue indefinitely. However, that seems like the correct behavior, because it indicates a flow design problem. To handle this, a `RouteOnAttribute` to check for the presence of `database.name` seems appropriate. This would align with existing behavior in cases where the value of `database.name` does not match with one of the configured values in the `DBCPConnectionLookup` service.

I would be supportive of making those changes in the `DBCPConnectionPoolLookup.getFlowFileFilter()`.

exceptionfactory:
Regarding `ExecuteSQL`, it does not appear to use the FlowFileFilter, unless I am missing a reference, so that would be a different point of consideration.

tpalfy:
@exceptionfactory I can agree with most what you say on a technical level. But there are conceptual levels to consider in my opinion.

In general, a robust, professional system is better off when prepared for various user errors as well. It's not great when the requirement for "database.name" means that if it is missing, the whole flow can get stuck and potentially crash a production system. While user errors can lead to that either way, NiFi is a perfect candidate for a component to prevent that.

The `ExecuteSQL` doesn't use the FlowFileFilter filter indeed, but that's just a technical detail. From a user perspective, they can use it with a `DBCPConnectionPoolLookup` and rely on it sending faulty FlowFiles to the failure relationship.

If anything, `PutSQL` should definitely do this even more so - when it's `Rollback On Failure` is set to `false`, as explicitly promises to do so by the documention.
Having to add another processor _instead_ is hardly a solution post-mortem and in general a bad user experience.

I do agree that the `DBCPConnectionPoolLookup.getFlowFileFilter()` should have been designed in a way - together with PutSQL - that it doesn't throw an exception. But the fact of the matter is, this change has been approved in this way - as part of the core mechanic of the implementation. Considering a major refactor is hardly a proper cost-benefit analysis result.

I updated the PR so that the adjustment only takes effect when `Rollback On Failure` is set to `false` though. This hopefully both limits any backward compatibility liability and makes the functionality more in tune with common sense user expectation and with the documentation.

exceptionfactory:
Thanks for the reply @tpalfy.

I agree with you that error handling should be more robust, and should attempt to mitigate potential user error. It is important to note, however, that the `DBCPConnectionPoolLookup` makes `database.name` an explicit part of the usage contract. This is unfortunate, but it is an existing implementation problem.

Part of the problem with the current situation is that the `getFlowFileFilter()` method never should have been added to the `DBCPService` interface, as it leads to the confusion of responsibility currently in place for `PutSQL`. I think we should separately look at deprecating this method for removal.

To help move things in the right direction, one option is to implement checking for the `database.name` attribute in `PutSQL` itself. In the `pollFlowFiles()` method, if the `DBCPService` returns a non-null instance of the `FlowFileFilter`, then there could be two calls to `session.get()`. The first call would use the provided filter. The second call would find FlowFiles missing the `database.name` attribute, and when found, those FlowFiles could be routed to failure.

This approach clarifies the tight coupling that already exists with the implementation in `DBCPConnectionPoolLookup`, and also moves in the direction of subsequent deprecation for removal.

I'm open to other alternatives that also help move things in this direction, but I think the exception handling approach is less clear, and doesn't move things in the best direction.

tpalfy:
@exceptionfactory Your suggestion is exactly (well, almost) what I'm trying to do! I only touched the `PutSQL` and no other class. I added a PutSQL-specific FlowFileFilter implementation there. It does this separation in one filter cycle, but the result is basically same.

Maybe the solution being complex confuscates what is really going on, but I deemed it necessary due to the fact that the `database.name` requirement only exists for `PutSQL` when the `DBCPService` is of the type `DBCPConnectionPoolLookup`. I didn't want to add a check for this specific combination.

exceptionfactory:
@tpalfy Yes, in terms of the FlowFileFilter handling, I agree that the current proposal is similar to what I had suggested. The main difference is how to determine the failure condition. Looking for the `database.name` attribute, as opposed to catching the `ProcessException` helps move in the direction away from the `DBCPConnectionPoolLookup` implementation.

tpalfy:
@exceptionfactory I think I could come up with an implementation according to your suggested idea.

exceptionfactory:
> @exceptionfactory I think I could come up with an implementation according to your suggested idea.

Thanks @tpalfy, the revised direction looks like a good path, I will take a closer look soon.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
```diff
@@ -701,46 +703,64 @@ private FlowFilePoll pollFlowFiles(final ProcessContext context, final ProcessSe
...
             } else {
-                flowFiles = session.get(dbcpServiceFlowFileFilter);
+                selectedFlowFiles = session.get(dbcpServiceFlowFileFilter);
             }
         }
 
-        if (flowFiles.isEmpty()) {
+        boolean selectedFlowFilesShouldHaveDataBaseNameButDont = dbcpServiceFlowFileFilter != null
+                && !selectedFlowFiles.isEmpty()
+                && selectedFlowFiles.stream().findAny().get().getAttribute("database.name") == null;
```
exceptionfactory: This complex conditional and name is a bit hard to follow on initial read, although the concept makes sense. I think having a first-level condition based on the presence of `dbcpServiceFlowFileFilter`, then checking for FlowFiles missing `database.name` would be a bit easier to follow.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
```diff
@@ -701,46 +703,64 @@ private FlowFilePoll pollFlowFiles(final ProcessContext context, final ProcessSe
...
+
+        if (selectedFlowFilesShouldHaveDataBaseNameButDont) {
+            selectedFlowFiles.forEach(flowFile -> getLogger().warn(
+                    "FlowFile {} is invalid because it's missing a 'database.name' attribute. Routing to '{}'.",
+                    flowFile.getAttribute(CoreAttributes.UUID.key()),
+                    REL_FAILURE.getName()
+            ));
```
exceptionfactory: Recommend adjusting the message since the FlowFile UUID and other information is included in the standard `toString()` method. I also recommend removing the trailing `.` character.
```suggestion
            selectedFlowFiles.forEach(flowFile -> getLogger().warn(
                    "{} missing attribute named [database.name] Routing to [{}]",
                    flowFile,
                    REL_FAILURE.getName()
            ));
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
```diff
@@ -805,50 +825,50 @@ boolean isFragmentedTransactionReady(final List<FlowFile> flowFiles, final Long
                 return true;
             } else if (fragmentCount == null) {
                 throw illegal.apply("Cannot process %s because there are %d FlowFiles with the same fragment.identifier "
-                        + "attribute but not all FlowFiles have a fragment.count attribute", new Object[] {flowFile, flowFiles.size()});
+                        + "attribute but not all FlowFiles have a fragment.count attribute", new Object[]{flowFile, flowFiles.size()});
```
exceptionfactory: This looks like a good opportunity to remove the unnecessary `new Object[]{}` wrapper, which would be a helpful cleanup. If you prefer to avoid that change, I would revert the spacing changes.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
tpalfy: I'm not sure what you mean by this to be honest. Can you provide a code suggestion?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
tpalfy: I guess this makes sense. Feels like it can clutter the bulletin, but probably that's how it's done in other places as well.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
tpalfy: I think I accidentally did a full-file formatting. I'll revert the unrelated ones.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutSQL.java
exceptionfactory: Sure, here is one possible approach, there are a probably a few other ways to implement it. I also noticed that the current `stream().findAny()` approach does not appear to be correct, since it is only retrieving one item at random instead of checking the entire collection.

```
final List<FlowFile> validFlowFiles;
if (dbcpServiceFlowFileFilter == null) {
    validFlowFiles = selectedFlowFiles;
} else {
    final boolean invalidFlowFilesFound = selectedFlowFiles.stream()
        .filter(selectedFlowFile -> 
            selectedFlowFile.getAttribute("database.name") == null
        )
        .findAny()
        .isPresent();
    if (invalidFlowFilesFound) {
        // Log warning for each FlowFile...
        validFlowFiles = List.of();
        result.routeTo(selectedFlowFiles, REL_FAILURE);
    } else {
        validFlowFiles = selectedFlowFiles;
    }
}
```

tpalfy: I think the objective disadvantages of such an approach is just too high. This has more complexity with a nested if and also has 3 clauses, giving the impression that we deal with 3 different scenarios.
Also doesn't do anything to explain the connection between the `database.name` requirement and the type of the `FlowFileFilter`.

How about I go the classic route and inline the variable and add a comment instead?

As for the `stream().findAny()`, the returned collection is expected to be homogenous (whether it contains valid flowfiles - in which case all `database.name` have the same value - or invalid flowfiles - in which case all `database.name` are missing).

Not the same, but similar to how this is relied upon in `PutSQL` when creating a connection:
```java
    private final PartialFunctions.InitConnection<FunctionContext, Connection> initConnection = (c, s, fc, ffs) -> {
        final Connection connection = c.getProperty(CONNECTION_POOL).asControllerService(DBCPService.class)
                .getConnection(ffs == null || ffs.isEmpty() ? emptyMap() : ffs.getFirst().getAttributes());
```
note
`ffs.getFirst().getAttributes()`

exceptionfactory: Thanks for the reply and clarifying the `findAny()` behavior in light of the list contents. In that case, can it be changed to a simple `getFirst()`? With a streamlined invocation and a comment, the more concise approach sounds good.

exceptionfactory:
Thanks for the updates @tpalfy. It looks like some recent PMD changes are causing the code-compliance scan to fail on an unrelated module. If you can rebase the PR, that should resolve the issue.

exceptionfactory:
@tpalfy One procedural note, the commits on the branch have the "Unverified" tag, stating that "No user is associated with the committer email." This sounds like a Git setting where the email address for the commit author needs to be adjusted. Can you review this setting and push an update?

tpalfy:
@exceptionfactory Thanks for pointing out the email issue. I thought I took care of that already but I thought wrong. I sorted it out.

exceptionfactory:
> @exceptionfactory Thanks for pointing out the email issue. I thought I took care of that already but I thought wrong. I sorted it out.

Thanks @tpalfy, I see Verified now!

exceptionfactory:
It looks like the last update has a minor PMD finding that needs to be corrected:

```
PMD Failure: org.apache.nifi.processors.standard.TestPutSQL:1638 Rule:LooseCoupling Priority:3 Avoid using implementation types like 'HashMap'; use the interface instead.
```

PR #10145, opened 2025-07-28
NIFI-14599 - Use parameter context revision when updating multiple PCs via parameter groups from parameter provider

pvillard31:
# Summary

[NIFI-14599](https://issues.apache.org/jira/browse/NIFI-14599) - Use parameter context revision when updating multiple PCs via parameter groups from parameter provider

Note that the system test being added here does not, unfortunately, expose the problem because of the way the client ID is set via the system test NiFi Client Util. But since I created this test as part of investigating the issue and since this is still a good test to have, I kept it as part of this PR.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


pvillard31:
The system test I added is currently failing because it also requires the code from #10137 to fix proper deletion of parameters

exceptionfactory (approved):
Thanks for tracking down this issue @pvillard31, the additional test is also helpful. +1 merging

PR #10148, opened 2025-07-29
NIFI-14800 - Shorten y-axis integer labels when 100,000 or more using standard abbreviations

rfellows:
# Summary

[NIFI-14800](https://issues.apache.org/jira/browse/NIFI-14800)

### Changes Made:

**1. Enhanced `formatInteger` method in `NiFiCommon` service:**
- **Purpose:** Solve y-axis label clipping issues in charts when displaying large numbers
- **Implementation:** 
  - Optional argument `useCompactNotation` added to support new behavior
  - Numbers ≥ 100,000 now use compact notation via `Intl.NumberFormat` with `notation: 'compact'`
  - Shows up to 2 decimal places for precision (`maximumFractionDigits: 2`)
  - Numbers < 100,000 continue using existing comma formatting
  - Examples: `100000` → `"100K"`, `1250000` → `"1.25M"`, `2750000000` → `"2.75B"`

**2. Comprehensive test coverage:**
- **Added:** new test cases covering:
  - Small numbers with comma formatting (0 to 99,999)
  - Large numbers with compact notation (≥ 100,000)
  - Negative number handling
  - Edge cases around the 100,000 threshold
  - Very large numbers (billions, trillions)

### Impact:
This change directly addresses chart label clipping issues by making large integer labels significantly shorter while maintaining readability. The status history chart component will automatically benefit from this improvement since it uses the `formatInteger` method through its formatters.

#### Before (labels aren't quite long enough to force clipping, but big enough to convey the idea)
Screenshot 2025-07-29 at 13 17 23

#### After abbreviating the y axis labels
Screenshot 2025-07-30 at 11 18 25

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/status-history/status-history-chart/status-history-chart.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/status-history/status-history-chart/status-history-chart.component.ts
@@ -111,6 +111,28 @@ export class StatusHistoryChart implements OnDestroy {
             return this.nifiCommon.formatFloat(d / 1000000);
         }
     };
+
+    // Formatters for y-axis use with compact notation for large numbers
+    private yAxisFormatters: any = {
+        DURATION: (d: number) => {
+            return this.nifiCommon.formatDuration(d);
+        },
+        COUNT: (d: number) => {
+            // need to handle floating point number since this formatter
+            // will also be used for average values
+            if (d % 1 === 0) {
+                return this.nifiCommon.formatInteger(d, true); // Enable compact notation for y-axis
+            } else {
+                return this.nifiCommon.formatFloat(d);
+            }
+        },
+        DATA_SIZE: (d: number) => {
+            return this.nifiCommon.formatDataSize(d);
+        },
+        FRACTION: (d: number) => {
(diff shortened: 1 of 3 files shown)


pvillard31:
The abbreviation seems to also be used for Min/Max/Mean on the left side as well, not just the y-axis. Is that expected? We may want the exact number display there?

rfellows:
> The abbreviation seems to also be used for Min/Max/Mean on the left side as well, not just the y-axis. Is that expected? We may want the exact number display there?

Maybe not explicitly intended. They use the same formatting function that is modified in this PR. I am not certain that the exact number is required here, but can see the argument for it. They are only shortened/abbreviated if the number is 6 digits or more (so 100,000+). If you think the use case for the exact number is common, I can see about leaving it the full number.

sfc-gh-jwitt:
Abbreviation on the graph on the right - perfect.
Abbreviation on the numbers on the left - please avoid if possible.

I am constantly copying/pasting the real specific numbers from the left.

rfellows:
> Abbreviation on the graph on the right - perfect. Abbreviation on the numbers on the left - please avoid if possible.
> 
> I am constantly copying/pasting the real specific numbers from the left.

Thanks for the feedback @sfc-gh-jwitt and @pvillard31, I'll take it into account and update this PR.

scottyaslan:
Reviewing...

PR #10149, opened 2025-07-30
NIFI-14801 Adding registryClientUrl and sslContextServiceId to create-reg-client and update-reg-client toolkit commands

grishick:
# Summary

[NIFI-14801](https://issues.apache.org/jira/browse/NIFI-14801)
Adding registryClientUrl and sslContextServiceId to create-reg-client and update-reg-client toolkit commands
# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

- [x] JDK 21

### Licensing

- [x] No new dependencies

### Documentation

### DIFF
--- a/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/CommandOption.java
+++ b/nifi-toolkit/nifi-toolkit-cli/src/main/java/org/apache/nifi/toolkit/cli/impl/command/CommandOption.java
@@ -82,7 +82,8 @@ public enum CommandOption {
     REGISTRY_CLIENT_URL("rcu", "registryClientUrl", "The url of the registry client", true),
     REGISTRY_CLIENT_DESC("rcd", "registryClientDesc", "The description of the registry client", true),
     REGISTRY_CLIENT_TYPE("rct", "registryClientType", "The type of the registry client", true),
-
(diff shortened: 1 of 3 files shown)


pvillard31 (changes requested):
I'm not in favor of this change as this is really assuming the creation of a NiFi Registry Client. But we have many additional implementations where `url` and/or `ssl-context-service` are not existing properties in the implementations.

So I'd like to understand the statement:
> update-reg-client commands in nifi-toolkit do not allow creating a functional registry client without additional API operations via cURL

I am using the combination of
````
./bin/cli.sh nifi create-reg-client
./bin/cli.sh nifi set-reg-client-property
````

and I have not seen such limitation.

grishick:
@pvillard31 I get your point. I made this change on 1.2x because there wasn't set-reg-client-property command available there and I don't see it in the new docs either: https://nifi.apache.org/nifi-docs/toolkit-guide.html. Now that you've mentioned it I found it in the code though. I think my main problem with create-reg-client is still that it always leaves me with an incomplete registry client, and in order to get a working registry client, I have to make an additional operation.

pvillard31:
I'll close the PR as I don't think this is the right approach given this is currently implementation specific. If you do want/need to have everything in the same command I recommend looking at adding support for the `input` argument where one can give a file with a JSON representation of the component to instantiate. This is similar to existing commands such as `nifi create-service` for example.

PR #10159, opened 2025-08-01
NIFI-14808 Introduce MockParameterLookup class for parameters passing

kubaugust:
# Summary

[NIFI-14808](https://issues.apache.org/jira/browse/NIFI-14808)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-mock/src/main/java/org/apache/nifi/util/MockValidationContext.java
+++ b/nifi-mock/src/main/java/org/apache/nifi/util/MockValidationContext.java
@@ -40,11 +40,12 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-public class MockValidationContext extends MockControllerServiceLookup implements ValidationContext, ControllerServiceLookup {
+public class MockValidationContext extends MockControllerServiceLookup implements ValidationContext {
 
     private final MockProcessContext context;
     private final Map<String, Boolean> expressionLanguageSupported;
     private final StateManager stateManager;
+    private final ParameterLookup parameterLookup;
     private volatile boolean validateExpressions = true;
 
     public MockValidationContext(final MockProcessContext processContext) {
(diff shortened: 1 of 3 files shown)


nifi-mock/src/main/java/org/apache/nifi/util/MockValidationContext.java
```diff
@@ -40,20 +40,30 @@
...
     public MockValidationContext(final MockProcessContext processContext) {
-        this(processContext, null);
+        this(processContext, (StateManager) null);
     }
 
     public MockValidationContext(final MockProcessContext processContext, final StateManager stateManager) {
+        this(processContext, stateManager, ParameterLookup.EMPTY);
+    }
+
+    public MockValidationContext(final MockProcessContext processContext, final Map<String, String> parameters) {
```
sfc-gh-dhandermann: Although the `Map<String, String>` of parameters is convenient, it is not necessarily intuitive for all use cases, so I recommend removing this constructor. Removing it also eliminates the need to disambiguate existing usage of `null` for the `StateManager`.

sfc-gh-dhandermann (changes requested):
Thanks for proposing this improvement @kubaugust. Most of the changes look straightforward. The only concern I noted is the ambiguity of the two-argument constructor. Although the `MockValidationContext` has minimal direct use in the project, it could also be used outside, and requiring the disambiguation is not optimal for compatibility. Eliminating that and simply supporting the three-argument constructor with the `ParameterLookup` instance should resolve both issues. That should remove the need to change other classes as well, so those changes can be reverted.

nifi-mock/src/main/java/org/apache/nifi/util/MockValidationContext.java
kubaugust: Thanks David for noticing! Indeed, this might be a bit unintuitive... Removed this constructor

PR #10166, opened 2025-08-02
NIFI-14816 - Add Azure Git DevOps Flow Registry Client

pvillard31:
# Summary

NIFI-14816 - Add Azure Git DevOps Flow Registry Client

We currently have support for Github, Gitlab and Bitbucket. This Jira is for adding a new implementation for Azure Git DevOps. The initial implementation will support only one authentication method with service principals and the client credentials OAuth2 flow.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
+++ b/nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
@@ -17,6 +17,7 @@
 
 package org.apache.nifi.azure.devops;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.nifi.oauth2.OAuth2AccessTokenProvider;
@@ -34,7 +35,10 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -57,12 +61,81 @@ public class AzureDevOpsRepositoryClient implements GitRepositoryClient {
     private static final String CONTENT_TYPE_HEADER = "Content-Type";
     private static final String BEARER = "Bearer ";
 
+    // Azure DevOps Git security namespace identifier used for permissions API
+    private static final String GIT_SECURITY_NAMESPACE = "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87";
+    // Security token format used for repository-scoped permission checks
+    private static final String SECURITY_TOKEN_FORMAT = "repoV2/%s/%s";
+
+    // Common string values
+    private static final String EMPTY_STRING = "";
+    private static final String FORWARD_SLASH = "/";
+
+    // Path segments
+    private static final String SEGMENT_APIS = "_apis";
+    private static final String SEGMENT_PERMISSIONS = "permissions";
+    private static final String SEGMENT_GIT = "git";
+    private static final String SEGMENT_REPOSITORIES = "repositories";
+    private static final String SEGMENT_ITEMS = "items";
+    private static final String SEGMENT_REFS = "refs";
+    private static final String SEGMENT_COMMITS = "commits";
+    private static final String SEGMENT_PUSHES = "pushes";
+
+    // Permission bit for Generic Contribute
+    private static final String GENERIC_CONTRIBUTE_PERMISSION_BIT = "4";
+
+    // Query parameter names
+    private static final String PARAM_TOKENS = "tokens";
+    private static final String PARAM_FILTER = "filter";
+    private static final String PARAM_INCLUDE_CONTENT = "includeContent";
+    private static final String PARAM_FORMAT = "format";
+    private static final String PARAM_VERSION = "version";
+    private static final String PARAM_VERSION_TYPE = "versionType";
+    private static final String PARAM_PATH = "path";
+    private static final String PARAM_SCOPE_PATH = "scopePath";
+    private static final String PARAM_TOP = "$top";
+    private static final String PARAM_SEARCH_ITEM_PATH = "searchCriteria.itemPath";
+    private static final String PARAM_SEARCH_ITEM_VERSION = "searchCriteria.itemVersion.version";
+    private static final String PARAM_SEARCH_ITEM_VERSION_TYPE = "searchCriteria.itemVersion.versionType";
+
+    // Common parameter values
+    private static final String INCLUDE_TRUE = "true";
+    private static final String FORMAT_OCTET_STREAM = "octetStream";
+    private static final String VERSION_TYPE_COMMIT = "commit";
+    private static final String REFS_HEADS_PREFIX = "refs/heads/";
+    private static final String FILTER_HEADS_PREFIX = "heads/";
+    private static final String OBJECT_TYPE_TREE = "tree";
+    private static final String OBJECT_TYPE_BLOB = "blob";
+    private static final String JSON_FIELD_VALUE = "value";
+    private static final String JSON_FIELD_PROJECT = "project";
+    private static final String JSON_FIELD_ID = "id";
+    private static final String JSON_FIELD_NAME = "name";
+    private static final String JSON_FIELD_GIT_OBJECT_TYPE = "gitObjectType";
+    private static final String JSON_FIELD_PATH = "path";
+    private static final String JSON_FIELD_AUTHOR = "author";
+    private static final String JSON_FIELD_COMMENT = "comment";
+    private static final String JSON_FIELD_DATE = "date";
+    private static final String JSON_FIELD_COMMIT_ID = "commitId";
+    private static final String JSON_FIELD_OBJECT_ID = "objectId";
+
+    // Change request constants
+    private static final String CHANGE_TYPE_ADD = "add";
+    private static final String CHANGE_TYPE_EDIT = "edit";
+    private static final String CHANGE_TYPE_DELETE = "delete";
+    private static final String CONTENT_TYPE_BASE64 = "base64encoded";
+
+    // Common query parameter names and values
+    private static final String VERSION_DESCRIPTOR_VERSION = "versionDescriptor.version";
+    private static final String VERSION_DESCRIPTOR_VERSION_TYPE = "versionDescriptor.versionType";
+    private static final String VERSION_TYPE_BRANCH = "branch";
+    private static final String RECURSION_LEVEL = "recursionLevel";
+    private static final String RECURSION_LEVEL_ONE_LEVEL = "OneLevel";
+
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
     private final String organization;
     private final String project;
     private final String repoName;
-    private final String apiUrl;
+    private final URL apiUrl;
     private final String repoPath;
     private final String clientId;
     private final OAuth2AccessTokenProvider tokenProvider;
@@ -92,8 +165,8 @@ public class AzureDevOpsRepositoryClient implements GitRepositoryClient {
         // attempt to retrieve repository information to validate access
         final URI uri = getUriBuilder().build();
         final JsonNode response = executeGet(uri);
-        final String projectId = response.get("project").get("id").asText();
-        final String repoId = response.get("id").asText();
+        final String projectId = response.get(JSON_FIELD_PROJECT).get(JSON_FIELD_ID).asText();
+        final String repoId = response.get(JSON_FIELD_ID).asText();
 
         canRead = true;
         canWrite = hasWriteAccess(projectId, repoId);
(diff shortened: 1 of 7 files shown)


nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureGitDevOpsFlowRegistryClient.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.nifi.registry.flow.FlowRegistryException;
+import org.apache.nifi.registry.flow.git.AbstractGitFlowRegistryClient;
+import org.apache.nifi.registry.flow.git.client.GitRepositoryClient;
+import org.apache.nifi.web.client.provider.api.WebClientServiceProvider;
+
+import java.util.List;
+
+@Tags({ "git", "azure", "devops", "registry", "flow" })
+@CapabilityDescription("Flow Registry Client that uses the Azure DevOps Git REST API to version control flows in a repository.")
+public class AzureGitDevOpsFlowRegistryClient extends AbstractGitFlowRegistryClient {
```
exceptionfactory: I recommend either removing `Git` from the class name, or moving it around as `AzureDevOpsGitFlowRegistryClient`
```suggestion
public class AzureDevOpsFlowRegistryClient extends AbstractGitFlowRegistryClient {
```

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureGitDevOpsFlowRegistryClient.java
```diff
@@ -0,0 +1,131 @@
...
+
+    static final PropertyDescriptor AZURE_DEVOPS_API_URL = new PropertyDescriptor.Builder()
+            .name("Azure DevOps API URL")
+            .description("The base URL of the Azure DevOps instance")
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .defaultValue("dev.azure.com")
```
exceptionfactory: Since the property has `URL`, the value should be `https://dev.azure.com` and this should use a URL validator. Alternatively, this property could be renamed, but it looks like the scheme should be carried through to the implementation that builds the URL.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+    /**
+     * Checks if the user has write access to the repository.
+     *
+     * @param projectId The ID of the project containing the repository.
+     * @param repoId    The ID of the repository.
+     * @return true if the user has write access, false otherwise.
+     * @throws FlowRegistryException if an error occurs while checking access.
+     */
+    private boolean hasWriteAccess(final String projectId, final String repoId) throws FlowRegistryException {
+        final String securityNamespace = "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"; // Azure DevOps Git security namespace
```
exceptionfactory: Recommend declaring this value as a private static variable, with the comment indicating the security namespace reference.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+        final String securityToken = String.format("repoV2/%s/%s", projectId, repoId);
```
exceptionfactory: Recommend declaring the format string as a private static variable named `SECURITY_TOKEN_FORMAT` or similar.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+
+        final URI uri = this.webClient.getHttpUriBuilder()
+                .scheme("https")
```
exceptionfactory: The `scheme` could be pulled from the `apiUrl` if the variable type was `URL`.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+                .host(apiUrl)
```
exceptionfactory: As noted on the property descriptor, if the value is a URL, then this could be `apiUrl.getHost()`

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+                .addPathSegment("permissions")
+                .addPathSegment(securityNamespace)
+                .addPathSegment("4") // 4 is the permission bit for "Generic Contribute" in Azure DevOps Git
+                .addQueryParameter("tokens", securityToken)
+                .addQueryParameter(API, API_VERSION)
+                .build();
+
+        final JsonNode response = executeGet(uri);
+
+        return response.has("value")
```
exceptionfactory: I recommend calling `response.get("value")` and then performing the checks on the `JsonNode`, instead of calling `get("value")` multiple times.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+        }
+        return result;
+    }
+
+    @Override
+    public Set<String> getTopLevelDirectoryNames(final String branch) throws FlowRegistryException {
+        final URI uri = listingUrl("")
+                .addQueryParameter("versionDescriptor.version", branch)
+                .addQueryParameter("versionDescriptor.versionType", "branch")
+                .addQueryParameter("recursionLevel", "OneLevel")
```
exceptionfactory: With these query parameters used in multiple places, recommend declaring static named values.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+            final Optional<String> existingSha = getContentSha(request.getPath(), branch);
+            changeType = existingSha.isPresent() ? "edit" : "add";
+        } else {
+            changeType = "edit";
+        }
+
+        final String encoded = Base64.getEncoder().encodeToString(request.getContent().getBytes(StandardCharsets.UTF_8));
+        final String json = String.format("{\"refUpdates\":[{\"name\":\"refs/heads/%s\",\"oldObjectId\":\"%s\"}],"
+                + "\"commits\":[{\"comment\":\"%s\",\"changes\":[{\"changeType\":\"%s\",\"item\":{\"path\":\"%s\"},"
+                + "\"newContent\":{\"content\":\"%s\",\"contentType\":\"base64encoded\"}}]}]}",
```
exceptionfactory: A multiline string could be used for formatting. It may also be worth defining the objects as `records` inside this class to avoid the string formatting.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+                .addQueryParameter(API, API_VERSION)
+                .build();
+        final JsonNode refResponse = executeGet(refUri);
+        final String oldObjectId = refResponse.get("value").get(0).get("objectId").asText();
+
+        final URI pushUri = getUriBuilder().addPathSegment("pushes")
+                .addQueryParameter(API, API_VERSION)
+                .build();
+
+        final String json = String.format("{\"refUpdates\":[{\"name\":\"refs/heads/%s\",\"oldObjectId\":\"%s\"}],"
```
exceptionfactory: See above note on JSON formatting, multiple uses may also be a reason to create `records` for serialization.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+
+    /**
+     * Create URI builder for accessing the repository.
+     *
+     * @return HTTP URL Builder configured for repository access
+     */
+    private HttpUriBuilder getUriBuilder() {
+        return this.webClient.getHttpUriBuilder()
+                .scheme("https")
+                .host(apiUrl)
```
exceptionfactory: Noting this reference on changing `apiUrl` to a type of `URL` instead of string.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,618 @@
...
+        final String prefix = (repoPath == null || repoPath.isBlank())
+                ? ""
+                : (repoPath.startsWith("/") ? repoPath : "/" + repoPath);
+
+        final String normalizedPath = (path == null || path.isBlank())
+                ? ""
+                : (path.startsWith("/") ? path : "/" + path);
+
+        if (prefix.isEmpty()) {
+            return normalizedPath.isEmpty() ? "/" : normalizedPath;
```
exceptionfactory: Recommend using `EMPTY_STRING` and `FORWARD_SLASH` static values. It also looks like a single method could be defined and reused for constructing both the `prefix` and `normalizedPath` using the same content and starts with checks.

pvillard31:
Thanks for the thorough review @exceptionfactory - I pushed a commit to address your comments and confirmed that everything is working as expected with my repo in Azure.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,739 @@
...
+        // attempt to retrieve repository information to validate access
+        final URI uri = getUriBuilder().build();
+        final JsonNode response = executeGet(uri);
+        final String projectId = response.get(JSON_FIELD_PROJECT).get(JSON_FIELD_ID).asText();
+        final String repoId = response.get(JSON_FIELD_ID).asText();
+
+        canRead = true;
+        canWrite = hasWriteAccess(projectId, repoId);
+
+        LOGGER.info("Created {} for clientId = [{}], repository [{}]", getClass().getSimpleName(), clientId, repoName);
```
exceptionfactory: The `clientId` seems like it could be confused with the Client ID for authentication. Also if it is only used for logging, is it needed?

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
```diff
@@ -0,0 +1,739 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Git Repository Client implementation for Azure DevOps using the REST API.
+ */
+public class AzureDevOpsRepositoryClient implements GitRepositoryClient {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AzureDevOpsRepositoryClient.class);
```
exceptionfactory: Passing the `ComponentLog` would enable associating logs with the Flow Registry instance, instead of just passing the component identifier.

exceptionfactory (commented):
Thanks for the updates @pvillard31. This looks close to completion, just one additional question related to logging and the component identifier.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
pvillard31: I followed the approach already taken in the GitLab and BitBucket implementations. I can remove this client ID if you think this is adding confusion at the moment. Or make it more clear by explicitly saying Flow Registry Client ID or something similar.

nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
pvillard31: Yep good point and I guess a follow up improvement could be to do the same on the other flow registry client implementations

exceptionfactory (approved):
Thanks for the latest updates @pvillard31, the latest version looks good! +1 merging

PR #10200, opened 2025-08-13
NIFI-14857 In ConsumeMQTT and PublishMQTT, remove 'final' keyword from 'getSupportedPropertyDescriptors()' method.

tpalfy:
# Summary

[NIFI-14857](https://issues.apache.org/jira/browse/NIFI-14857)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/ConsumeMQTT.java
+++ b/nifi-extension-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/ConsumeMQTT.java
@@ -304,7 +304,7 @@ public class ConsumeMQTT extends AbstractMQTTProcessor {
     }
 
     @Override
-    public final List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+    public List<PropertyDescriptor> getSupportedPropertyDescriptors() {
         return PROPERTY_DESCRIPTORS;
     }
 
--- a/nifi-extension-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/PublishMQTT.java
+++ b/nifi-extension-bundles/nifi-mqtt-bundle/nifi-mqtt-processors/src/main/java/org/apache/nifi/processors/mqtt/PublishMQTT.java
@@ -169,7 +169,7 @@ public class PublishMQTT extends AbstractMQTTProcessor {
     }
 
     @Override
-    public final List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+    public List<PropertyDescriptor> getSupportedPropertyDescriptors() {
         return PROPERTY_DESCRIPTORS;
     }
 


joewitt:
I can comment here or on the JIRA as preferred.  The goal here is not clear.

Is the intent that by removing 'final' on these method signatures these components become open for extension?  By whom?

tpalfy:
@joewitt Yeah, the goal is to make it possible in general to create specialized processors based on the existing one by inheriting from them.

exceptionfactory (commented):
Although the change itself is simple, and most Processors do not use the `final` keyword for this method, Processors are not designed for extension. The MQTT Processors in particular contain a number of private methods that do not lend themselves to extension, so making this specific change, for the purpose of extension, does not sound like a supported path forward.

joewitt:
I definitely share what I think is your leaning here @exceptionfactory 

However, I think all this does is remove two generally pointless final declarations.

It doesn't mean or imply we're saying this processor/set of processors are design for extension.  They're really not.

So with that very narrow view of what this I am +1 on this. You ok with that?

exceptionfactory:
> I definitely share what I think is your leaning here @exceptionfactory
> 
> However, I think all this does is remove two generally pointless final declarations.
> 
> It doesn't mean or imply we're saying this processor/set of processors are design for extension. They're really not.
> 
> So with that very narrow view of what this I am +1 on this. You ok with that?

Thanks @joewitt, yes, with this background, I'm fine with the change in this particular PR.

tpalfy:
@joewitt @exceptionfactory I can change the title and description of the Jira (the commit message is technical already). Although I think overall it would only be to NiFi's benefit if we considered components extensible.

exceptionfactory:
> @joewitt @exceptionfactory I can change the title and description of the Jira (the commit message is technical already). Although I think overall it would only be to NiFi's benefit if we considered components extensible.

Thanks, updating the Jira issue would be helpful for reference purposes.

Designing Processors for extension brings a much greater burden of complexity and maintainability, which is part of the reason that the classes themselves are not considered open for extension.

tpalfy:
Updated the Jira.

joewitt:
@tpalfy

> Although I think overall it would only be to NiFi's benefit if we considered components extensible.

I infer you are not referring to the already existing incredibly extensive extensibility powers within NiFI but instead you are talking about 'we need to ensure extensions we build are themselves extensible'.

This becomes purely about how each component is written and designed.  You can demonstrate and model what you'd like to see done to support this with components.  In fact every author of a component should consider whether they want that thing extended by others but also the cost to the community to maintain such additional APIs.

Notably simply removing 'final' from a method or two does not do this.  The extension developer has to consider whether they're going to make specific points of designed extensibility for a given component such that it might be open for extension but closed for modification.  Or a totally different extensibility model where the extender could in fact break the thing.  Whatever it is you as the author of the component get to choose that.

It would also be interesting to see or consider how such APIs would get published.  We maintain the purity of the 'nifi-api' so you'd also then need to consider 'nifi-exampleprocessor-api-X.Y.Z' and so on.  The moment you declare a component itself extensible you are declaring whatever its API is as public.  Now you also cannot make changes without risking breaking code written against them.

I'll keep an eye out for a proposal or component where you share and show what you have in mind.

tpalfy:
@joewitt I just want to create custom processors based on the existing ones. Add some properties, remove/re-default others.

pvillard31 (approved):

PR #10220, opened 2025-08-19
NIFI-14880: Setting noopener and noreferrer when invoking window.open().

mcgilman:


mcgilman:
Thanks for the review @exceptionfactory! I missed that in the docs initially. While it appears that `noreferrer` predates `noopener` modern browsers at this point should support the implication and make the `noopener` redundant. Let me remove it here.

rfellows (approved):
Thanks @mcgilman 👍

exceptionfactory (approved):

PR #10238, opened 2025-08-25
NIFI-14900 - Adding support for dropStateKeySupported Stateful flag

pvillard31:
# Summary

NIFI-14900 - Adding support for dropStateKeySupported Stateful flag

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/state/StandardStateManager.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/state/StandardStateManager.java
@@ -24,8 +24,8 @@ import org.apache.nifi.components.state.StateProvider;
 import org.apache.nifi.logging.ComponentLog;
 import org.apache.nifi.logging.LogRepository;
 import org.apache.nifi.logging.LogRepositoryFactory;
-import org.apache.nifi.processor.SimpleProcessLogger;
 import org.apache.nifi.logging.StandardLoggingContext;
+import org.apache.nifi.processor.SimpleProcessLogger;
 
 import java.io.IOException;
 import java.util.Map;
@@ -34,11 +34,18 @@ public class StandardStateManager implements StateManager {
     private final StateProvider localProvider;
     private final StateProvider clusterProvider;
     private final String componentId;
+    private final boolean dropStateKeySupported;
 
     public StandardStateManager(final StateProvider localProvider, final StateProvider clusterProvider, final String componentId) {
+        this(localProvider, clusterProvider, componentId, false);
+    }
+
+    public StandardStateManager(final StateProvider localProvider, final StateProvider clusterProvider,
+            final String componentId, final boolean dropStateKeySupported) {
         this.localProvider = localProvider;
         this.clusterProvider = clusterProvider;
         this.componentId = componentId;
+        this.dropStateKeySupported = dropStateKeySupported;
     }
 
     private StateProvider getProvider(final Scope scope) {
@@ -84,6 +91,11 @@ public class StandardStateManager implements StateManager {
     public void clear(final Scope scope) throws IOException {
         getLogger(componentId).debug("Clearing {} State", scope);
         getProvider(scope).clear(componentId);
+    }
+
+    @Override
+    public boolean isStateKeyDropSupported() {
+        return dropStateKeySupported;
     }
 
     @Override
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/ExtensionBuilder.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/ExtensionBuilder.java
@@ -19,6 +19,7 @@ package org.apache.nifi.controller;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.annotation.behavior.RequiresInstanceClassLoading;
+import org.apache.nifi.annotation.behavior.Stateful;
 import org.apache.nifi.annotation.behavior.SupportsBatching;
 import org.apache.nifi.annotation.configuration.DefaultSettings;
 import org.apache.nifi.bundle.Bundle;
@@ -85,6 +86,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.net.ssl.SSLContext;
+
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Proxy;
 import java.net.URL;
@@ -507,6 +509,10 @@ public class ExtensionBuilder {
        applyDefaultSettings(procNode);
        applyDefaultRunDuration(procNode);
 
+       final Stateful stateful = processor.getComponent().getClass().getAnnotation(Stateful.class);
+       final boolean dropStateKeySupported = stateful != null && stateful.dropStateKeySupported();
+       stateManagerProvider.getStateManager(identifier, dropStateKeySupported);
+
        return procNode;
    }
 
@@ -525,6 +531,10 @@ public class ExtensionBuilder {
                    componentType, type, reloadComponent, extensionManager, validationTrigger, true);
            taskNode.setName(componentType);
        }
+
+       final Stateful stateful = reportingTask.getComponent().getClass().getAnnotation(Stateful.class);
+       final boolean dropStateKeySupported = stateful != null && stateful.dropStateKeySupported();
+       stateManagerProvider.getStateManager(identifier, dropStateKeySupported);
 
        return taskNode;
    }
@@ -550,6 +560,10 @@ public class ExtensionBuilder {
                    extensionManager, validationTrigger, true);
            parameterProviderNode.setName(componentType);
        }
+
+       final Stateful stateful = parameterProvider.getComponent().getClass().getAnnotation(Stateful.class);
+       final boolean dropStateKeySupported = stateful != null && stateful.dropStateKeySupported();
+       stateManagerProvider.getStateManager(identifier, dropStateKeySupported);
 
        return parameterProviderNode;
    }
@@ -668,7 +682,10 @@ public class ExtensionBuilder {
            final ComponentLog serviceLogger = new SimpleProcessLogger(identifier, serviceImpl, new StandardLoggingContext(null));
            final TerminationAwareLogger terminationAwareLogger = new TerminationAwareLogger(serviceLogger);
 
-           final StateManager stateManager = stateManagerProvider.getStateManager(identifier);
+           final Stateful stateful = serviceImpl.getClass().getAnnotation(Stateful.class);
+           final boolean dropStateKeySupported = stateful != null && stateful.dropStateKeySupported();
+           final StateManager stateManager = stateManagerProvider.getStateManager(identifier, dropStateKeySupported);
+
            final ControllerServiceInitializationContext initContext = new StandardControllerServiceInitializationContext(identifier, terminationAwareLogger,
                    serviceProvider, stateManager, kerberosConfig, nodeTypeProvider);
            serviceImpl.initialize(initContext);
@@ -845,6 +862,10 @@ public class ExtensionBuilder {
                    componentType, type, reloadComponent, extensionManager, validationTrigger, true);
            ruleNode.setName(componentType);
        }
+
+       final Stateful stateful = flowAnalysisRule.getComponent().getClass().getAnnotation(Stateful.class);
+       final boolean dropStateKeySupported = stateful != null && stateful.dropStateKeySupported();
+       stateManagerProvider.getStateManager(identifier, dropStateKeySupported);
 
        return ruleNode;
    }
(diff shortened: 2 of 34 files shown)


pvillard31:
Yeah, I noticed and I'm trying to figure out what is the problem because everything is working as expected locally and I can't really understand how it can report a state with a value of 2 at this point of the test...

pvillard31:
Understood what was happening... When running locally, things are done really fast so, even though there is a 2 nodes cluster, the state is updated at the same time by the processor on each node leading to state value of 1. In Github testing, the processor instances are likely not executing at the exact same time so each instance on each node increments the state once, leading to a value of 2. Given what we want to test here, I just used the PrimaryNodeOnly annotation to solve the problem and we still have the appropriate testing of the new feature.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/ExtensionBuilder.java
```diff
@@ -507,6 +509,10 @@ private ProcessorNode createProcessorNode(final LoggableComponent<Processor> pro
        applyDefaultSettings(procNode);
        applyDefaultRunDuration(procNode);
 
+       final Stateful stateful = processor.getComponent().getClass().getAnnotation(Stateful.class);
+       final boolean dropStateKeySupported = stateful != null && stateful.dropStateKeySupported();
```
dan-s1: These lines are duplicated five time in this class. I think it would be worthwhile to create a reusable method to return a boolean which you could assign to the `dropStateKeySupported` variable. e.g.

```
private boolean isDropStateKeySupported(final ConfigurableComponent configurableComponent) {
       final Stateful stateful = configurableComponent.getClass().getAnnotation(Stateful.class);
       return stateful != null && stateful.dropStateKeySupported();
   }
```

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessorResource.java
```diff
@@ -489,7 +489,7 @@ public Response getState(
      * @throws InterruptedException if interrupted
      */
     @POST
-    @Consumes(MediaType.WILDCARD)
+    @Consumes(MediaType.APPLICATION_JSON)
```
exceptionfactory: Following the pattern of other Resource methods, it looks like this should support both wildcard and JSON, correct?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/ControllerServiceDAO.java
```diff
@@ -150,6 +151,7 @@ public interface ControllerServiceDAO {
      * Clears the state of the specified controller service.
      *
      * @param controllerServiceId controller service id
+     * @param componentStateDTO   state of the controller service
      */
-    void clearState(String controllerServiceId);
+    void clearState(final String controllerServiceId, final ComponentStateDTO componentStateDTO);
```
exceptionfactory: Minor, but interface method arguments do not need `final`

nifi-system-tests/nifi-system-test-extensions-bundle/nifi-system-test-extensions/src/main/java/org/apache/nifi/processors/tests/system/MultiKeyState.java
```diff
@@ -0,0 +1,77 @@
...
+@Stateful(scopes = Scope.CLUSTER, description = "Stores three counters in state", dropStateKeySupported = true)
+public class MultiKeyState extends AbstractProcessor {
+    public static final Relationship REL_SUCCESS = new Relationship.Builder()
+            .name("success")
+            .autoTerminateDefault(true)
+            .build();
+
+    @Override
+    public Set<Relationship> getRelationships() {
+        return new HashSet<>(Collections.singletonList(REL_SUCCESS));
```
exceptionfactory:
```suggestion
        return Set.of(REL_SUCCESS);
```

nifi-system-tests/nifi-system-test-extensions-bundle/nifi-system-test-extensions/src/main/java/org/apache/nifi/processors/tests/system/MultiKeyStateNotDroppable.java
```diff
@@ -0,0 +1,77 @@
...
+@Stateful(scopes = Scope.CLUSTER, description = "Stores three counters in state", dropStateKeySupported = false)
+public class MultiKeyStateNotDroppable extends AbstractProcessor {
+    public static final Relationship REL_SUCCESS = new Relationship.Builder()
+            .name("success")
+            .autoTerminateDefault(true)
+            .build();
+
+    @Override
+    public Set<Relationship> getRelationships() {
+        return new HashSet<>(Collections.singletonList(REL_SUCCESS));
```
exceptionfactory:
```suggestion
        return Set.of(REL_SUCCESS);
```

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardComponentStateDAO.java
```diff
@@ -42,38 +46,98 @@ private StateMap getState(final String componentId, final Scope scope) {
...
+                manager.clear(Scope.CLUSTER);
+                manager.clear(Scope.LOCAL);
+            }
         } catch (final IOException ioe) {
-            throw new IllegalStateException(String.format("Unable to clear the state for the specified component %s: %s", componentId, ioe), ioe);
+            throw new IllegalStateException("Unable to clear the state for the specified component %s: %s".formatted(componentId, ioe), ioe);
         }
     }
 
+    public boolean hasExactlyOneKeyRemoved(Map<String, String> currentState, Map<String, String> newState) {
```
exceptionfactory: Should this method be `public`?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardComponentStateDAO.java
```diff
@@ -42,38 +46,98 @@ private StateMap getState(final String componentId, final Scope scope) {
...
             }
 
-            // clear both state's at the same time
-            manager.clear(Scope.CLUSTER);
-            manager.clear(Scope.LOCAL);
+            if (componentStateDTO == null) {
+                // clear both state's at the same time
+                manager.clear(Scope.CLUSTER);
+                manager.clear(Scope.LOCAL);
+            } else if (manager.getState(Scope.LOCAL) != null && !manager.getState(Scope.LOCAL).toMap().isEmpty() && !stateManagerProvider.isClusterProviderEnabled()) {
```
exceptionfactory: The conditional in each of these clauses is a bit hard to follow. It seems like nesting branches for `componentStateDTO` not null would support a single check for the existing of local state, and then subsequent checks would be easier to follow.

exceptionfactory (changes requested):
Great work on implementing this new feature support @pvillard31. On closer review, the majority of the changes look ready to go. The core details in `StandardComponentStateDAO` are a bit hard to follow, so if that could be reworked a bit, then this should be ready.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessorResource.java
pvillard31: Yes, absolutely, good catch, thanks

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardComponentStateDAO.java
pvillard31: It should not, good catch

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardComponentStateDAO.java
pvillard31: Pushed a commit with an alternative that is easier to read. It has intermediary returns though. Happy to go back to a full if/else branches approach that would be a bit simplified as you suggested.

pvillard31:
Pushed a commit to add a null safe guard in the state map providers for ZK and k8s. The problem was introduced with the recent decoupling of ZK and was causing the added system tests of this PR to fail although it was OK before the rebase.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/ExtensionBuilder.java
```diff
@@ -507,6 +509,9 @@ private ProcessorNode createProcessorNode(final LoggableComponent<Processor> pro
        applyDefaultSettings(procNode);
        applyDefaultRunDuration(procNode);
 
+       // Ensure StateManager is initialized with component's Stateful capabilities
+       getStateManagerForComponent(identifier, processor.getComponent().getClass());
```
exceptionfactory: Calling this method raised some questions about the approach to creating the StateManager. This surfaces the fact that the Standard implementation of the StateManagerProvider caches instances, but the ExtensionBuilder should not need to be aware of this implementation detail. For this reason, these calls should be removed, and another approach should be considered for returning the appropriate instance of a StateManager.

pvillard31:
Thanks for the review @exceptionfactory - I pushed a change where I no longer assume state capability in ExtensionBuilder. Instead I introduced a pluggable resolver into StandardStateManagerProvider to avoid any ordering/caching concerns and evaluate the capability lazily. All of this gets wired into FlowControllerConfiguration. Let me know what you think.

pvillard31:
Thanks for the feedback @exceptionfactory - I pushed a commit that, I believe, is going in the direction you're suggesting. Let me know what yo think.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/scheduling/RepositoryContextFactory.java
```diff
@@ -54,7 +54,13 @@ public RepositoryContextFactory(final ContentRepository contentRepository, final
     }
 
     public RepositoryContext newProcessContext(final Connectable connectable, final AtomicLong connectionIndex) {
-        final StateManager stateManager = stateManagerProvider.getStateManager(connectable.getIdentifier());
+        final StateManager stateManager;
+        if (connectable instanceof org.apache.nifi.controller.ProcessorNode processorNode) {
```
exceptionfactory: Is there a reason this is limited to Processors? Other components can also use state, and the annotation evaluation should not need to know about the type of class.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/state/StandardStateManager.java
```diff
@@ -24,21 +24,34 @@
...
     private final StateProvider localProvider;
     private final StateProvider clusterProvider;
     private final String componentId;
+    private final Supplier<Boolean> dropStateKeySupportedSupplier;
 
     public StandardStateManager(final StateProvider localProvider, final StateProvider clusterProvider, final String componentId) {
+        this(localProvider, clusterProvider, componentId, () -> false);
+    }
+
+    public StandardStateManager(final StateProvider localProvider, final StateProvider clusterProvider,
```
exceptionfactory, line 44: Is this constructor needed in addition to the `Supplier` based constructor?

exceptionfactory (commented):
Thanks for the latest update @pvillard31, the revised approach to tracking support looks to be going in a good direction.

I noted several places where there is a check to determine whether a particular Node is a ProcessorNode, do you think that is needed? For ConnectableTasks, perhaps that is reasonable, but on the other hand, the Stateful annotation evaluation should handle any class, so it seems simpler to avoid the ProcessorNode checks unless it is needed for some other reason.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/scheduling/RepositoryContextFactory.java
pvillard31: No, you're right, this change is a leftover when I tried a few things. Reverting this change.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/state/StandardStateManager.java
pvillard31: No longer needed, also a leftover. Removing.

pvillard31:
Thanks David, I cleaned up the code. Sorry about that, changing the directions a couple of times across so many classes didn't help. I also double checked how we were doing the seeding for other component types and added calls in ExtensionBuilder. While this is not needed for reporting tasks / flow analysis rules, it is needed for controller services because this is the only place we construct its initialization context and wire a StateManager. For consistency I added it on all components to be on the safe side even if it's redundant.

PR #10241, opened 2025-08-25
NIFI-14874 Unable to move destination of a Connection to a Process Group when source component is running

alepekhin:
# Summary

[NIFI-14874](https://issues.apache.org/jira/browse/NIFI-14874)

# Tracking

### Issue Tracking

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-14874`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-14874`

### Pull Request Formatting

# Verification

To reproduce, create any two processors, Processor A and Processor B, with a Connection from Processor A (source) to Processor B (destination.) Also, create a Process Group which contains an Input Port within it. Start Processor A.

Attempt to move the destination of the Connection from Processor B to the Process Group's Input Port.
It should allow to edit connection and select input port 

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/connectable/StandardConnection.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/connectable/StandardConnection.java
@@ -271,12 +271,6 @@ public final class StandardConnection implements Connection {
         }
 
         try {
-            getSource().verifyCanUpdate();
-        } catch (final IllegalStateException ise) {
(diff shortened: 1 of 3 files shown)


joewitt:
I am pretty sure the solution here wont simply be to not check whether the change is valid on the server side.  

There are some important reasons we would not want to change the target in the midst of processing and thus why we enforce this.

alepekhin:
@joewitt We have discussed it in the issue. We do change target when we move connection to another target Process. Why we can not do the same when we move connection to another Input Port?

PR #10077, opened 2025-07-09
[WIP] NIFI-14727 Create ConsumeKinesisStreamV2

awelless:
# Summary

[NIFI-14727](https://issues.apache.org/jira/browse/NIFI-14727)

**The pull request is still work in progress!**
Support for records is waiting for #10053 to get approved.

Feedback on the present parts is much appreciated.

## Changes
Extended `MockProcessSession` which allows testing NiFi session rollback scenarios.

Created a new `ConsumeKinesis3Stream` processor which doesn't rely on `ProcessSessionFactory`.
Incoming records are buffered in `RecordBuffer`, where they're later consumed from by the procesor.

Created a new NAR `nifi-aws-kinesis-3-nar` which contains the new processor.

### ConsumeKinesis3Stream
A new version of the processor. Spins up KCL scheduler, read records from `RecordBuffer` and writes them into flow files.

### RecordBuffer
A facade which is responsible for buffering all incoming records, leverages backpressure when needed, and ensures exclusive read access on the records to processors tasks.
Each incoming record is routed to its according `ShardBuffer`, where it stays until consumed and committed by the processor.

### ShardBuffer
A buffer that stores records for a single shard. The records are stored in 2 collections to ensure replayability on session rollbacks. Note, that `ConsumeKinesis3Stream` doesn't interact with shard buffers directly. `RecordBuffer` behaves as a router here.

# Tracking

Please complete the following tracking steps prior to pull request creation.

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI) issue created

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-00000`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-00000`

### Pull Request Formatting

- [x] Pull Request based on current revision of the `main` branch
- [x] Pull Request refers to a feature branch with one commit containing changes

# Verification

Please indicate the verification steps performed prior to pull request creation.

### Build

- [ ] Build completed using `mvn clean install -P contrib-check`
  - [x] JDK 21

### Licensing

- [ ] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html) - **Checking out.**
- [ ] New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [ ] Documentation formatting appears as expected in rendered files

PR #10245, opened 2025-08-27
NIFI-14907 Selectively fail commits in MockProcessSession

awelless:
# Summary

[NIFI-14907](https://issues.apache.org/jira/browse/NIFI-14907)
The first commit introduces a new property which sets the behavior for session commits.
The second commit introduces a builder to `MockProcessSession`.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-mock/src/main/java/org/apache/nifi/util/MockProcessSession.java
+++ b/nifi-mock/src/main/java/org/apache/nifi/util/MockProcessSession.java
@@ -310,7 +310,7 @@ public class MockProcessSession implements ProcessSession {
         }
 
         if (failCommit) {
-            throw new TestFailedCommitException();
+            throw new FlowFileHandlingException("Cannot commit session because the session was requested to fail by a test");
         }
 
         closeStreams(openInputStreams, enforceStreamsClosed);
@@ -1480,12 +1480,6 @@ public class MockProcessSession implements ProcessSession {
         final String curUuid = curFlowFile.getAttribute(CoreAttributes.UUID.key());
         final String providedUuid = curFlowFile.getAttribute(CoreAttributes.UUID.key());
         return curUuid.equals(providedUuid);
-    }
-
-    public static final class TestFailedCommitException extends RuntimeException {
-        private TestFailedCommitException() {
-            super("Failing the commit, as requested by the test");
-        }
     }
 
     public static final class Builder {


exceptionfactory (changes requested):
Thanks for separating out these proposed changes @awelless.

The `nifi-mock` library is used widely for external components, so public constructors and methods cannot be changed. The builder pattern and constructor removal changes need to be reverted, then the core changes can be considered.

awelless:
@exceptionfactory 
For some reason I thought that `nifi-mock` is an internal module and not a part of external API.
Sure, constructors should not be changed. I reverted that.

I still believe that builder introduces readability, compared a constructor with 7 parameters. What's your opinion on that?
Shall we promote the builder usage by, e.g. deprecating the constructors with a large number of arguments?

nifi-mock/src/main/java/org/apache/nifi/util/MockProcessSession.java
```diff
@@ -1456,4 +1481,73 @@ boolean isFlowFileKnown(final FlowFile flowFile) {
...
+    public static final class Builder {
+
+        private final SharedSessionState sharedState;
+        private final Processor processor;
+        private StateManager stateManager;
+
+        private boolean enforceStreamsClosed = true;
+        private boolean allowRecursiveReads = false;
+        private boolean allowSynchronousCommits = false;
+        private boolean shouldFailCommit = false;
```
exceptionfactory: With the default value being false, setting these values to `false` is unnecessary.

nifi-framework-bundle/nifi-framework/nifi-site-to-site/src/test/java/org/apache/nifi/remote/protocol/http/TestHttpFlowFileServerProtocol.java
```diff
@@ -525,7 +524,9 @@ private void setupMockProcessSession() {
         when(rootGroupPort.getIdentifier()).thenReturn("root-group-port-id");
 
         sessionState = new SharedSessionState(rootGroupPort, new AtomicLong(0));
-        processSession = new MockProcessSession(sessionState, rootGroupPort, true, new MockStateManager(rootGroupPort), true);
+        processSession = MockProcessSession.builder(sessionState, rootGroupPort)
+                .allowSynchronousCommits()
+                .build();
```
exceptionfactory: Recommend reverting this change to keep the scope of this PR focused on `nifi-mock`.

nifi-mock/src/main/java/org/apache/nifi/util/MockProcessSession.java
```diff
@@ -285,14 +296,23 @@ public void commit() {
...
+        }
     }
 
     private void commitInternal() {
         if (!beingProcessed.isEmpty()) {
             throw new FlowFileHandlingException("Cannot commit session because the following FlowFiles have not been removed or transferred: " + beingProcessed);
         }
 
+        if (shouldFailCommit) {
+            throw new TestFailedCommitException();
```
exceptionfactory: Recommend throwing a `FlowFileHandlingException` instead of creating a new Exception class.

nifi-mock/src/main/java/org/apache/nifi/util/MockProcessSession.java
```diff
@@ -88,6 +88,7 @@ public class MockProcessSession implements ProcessSession {
     private final StateManager stateManager;
     private final boolean allowSynchronousCommits;
     private final boolean allowRecursiveReads;
+    private final boolean shouldFailCommit;
```
exceptionfactory: Recommend naming this `failCommit` and updating references, since `should` implies optional behavior, versus expected behavior.
```suggestion
    private final boolean failCommit;
```

exceptionfactory (changes requested):
Thanks for re-scoping the changes @awelless. This looks on the right track.

The addition of the `Builder` is useful, although I would not deprecate the constructors. I noted a few other minor adjustments.

nifi-mock/src/main/java/org/apache/nifi/util/MockProcessSession.java
awelless: I wanted to make the default values explicit, provided that `enforseStreamsClosed` is set to `true` by default.
Removed default `false` values.

nifi-mock/src/main/java/org/apache/nifi/util/MockProcessSession.java
awelless: A separate exception class allows us to distinguish between an exception caused by `failCommit` and an exception caused by illegal flow file handling.

E.g. [here](https://github.com/apache/nifi/pull/10077/files#diff-60594ac5c47e49ef5a5ce87340edb29188754d494046dbeced840263f87e2d8eR333) we're waiting specifically for a "scripted" rollback. A thrown `FlowFileHandlingException` during a test would mean that source code doesn't handle flow files correctly and should fail the test.

Throwing `FlowFileHandlingException` and catching it in tests might conceal other failure cases.

exceptionfactory: That is true that using `FlowFileHandlingException` could be confused with other exceptions. On the other hand, introducing a new Exception class expands the API surface of `nifi-mock`.

The `StandardProcessSession` can throw several different kinds of exceptions, including `FlowFileAccessException`, and of course incorrect behavior in a Processor can lead to other exceptions. An `UncheckedIOException` could be another option. I think it is better than any test code expect a generic `RuntimeException`, as opposed to a specific exception class, as other test methods can evaluate correct behavior.

awelless: I see. Then `FlowFileHandlingException` seems like the best option. I removed the new exception class.

PR #10249, opened 2025-08-28
NIFI-14910 NIFI-14911 NiFi - Upgrade Jetty to 12.1.1 and move to Jakarta EE11

pvillard31:
# Summary

NIFI-14910 NIFI-14911 NIFI-14913 NiFi - Upgrade Jetty to 12.1.1 and move to Jakarta EE11

**Important notes**:
- Given that Spring Boot 3.x relies on EE10, it is not possible to upgrade NiFi Registry to use EE10. This will be a separate effort that is tracked by NIFI-14912 & NIFI-14914. In the meantime, the pom dependencies have been adjusted to move NiFi to Jetty 12.1.1 and Jakarta EE11 while NiFi Registry remains on Jakarta EE10.
- This PR is currently as a draft to allow some time for reviewers. But also because it is worth mentioning that:
  - the Jersey dependency is currently set to `4.0.0-M4`. No clarity as to when the final `4.0.0` release is expected.
  - the swagger codegen plugin in Toolkit API is still using Jersey 3 as the configured library (https://github.com/apache/nifi/blob/main/nifi-toolkit/nifi-toolkit-api/pom.xml#L89) given that Jersey 4 is not an option yet - I do not believe however that this is an issue
- I have not moved the JMS bundle from `javax.jms-api` to `jakarta.jms-api` as this would be a breaking change for users with existing JMS components where a specific JMS client is provided as part of the configuration. This is something we can evaluate later.
- Jakarta EE11 brings support for Virtual Threads with Java 21. This could be evaluated and leveraged in follow up efforts.
- Went through NOTICE files / docs to clean up references and make sure we are aligned with latest. There is a lot of changes related to NOTICE files... sorry to the reviewers, I tried to do my best to make sure we are back to a good state

**Testing**:
- Beyond confirming that the full build with system tests is green, some additional manual testing has been done to confirm proper behavior of the UI with custom UIs, content viewer, documentation rendering, etc.
- Confirmed that NiFi Registry is working with Jetty 12.1/EE10

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/JettyServer.java
```diff
@@ -679,13 +673,13 @@ private void addDocsServlets(WebAppContext webAppContext) {
...
 
-            final ServletHolder docs = new ServletHolder("docs", ResourceServlet.class);
+            final ServletHolder docs = new ServletHolder("docs", DefaultServlet.class);
             final Path htmlBaseResource = docsDir.toPath().resolve("html");
             docs.setInitParameter("baseResource", htmlBaseResource.toString());
             docs.setInitParameter("dirAllowed", "false");
             webAppContext.addServlet(docs, "/html/*");
 
-            final ServletHolder restApi = new ServletHolder("rest-api", ResourceServlet.class);
+            final ServletHolder restApi = new ServletHolder("rest-api", DefaultServlet.class);
```
exceptionfactory: Is there a particular reason for changing from the ResourceServlet to the DefaultServlet? This previously resulted in warnings related to the DefaultServlet and path configuration.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/pom.xml
```diff
@@ -85,6 +85,25 @@
             <artifactId>nifi-web-servlet-shared</artifactId>
             <version>2.6.0-SNAPSHOT</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-deploy</artifactId>
```
exceptionfactory, line 90: The `jetty-deploy` library brings in several other dependencies. Is there a reason this is necessary in Jetty 12.1 and not 12.0?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/StandardServerProvider.java
```diff
@@ -63,7 +63,7 @@ class StandardServerProvider implements ServerProvider {
     public Server getServer(final NiFiProperties properties) {
         Objects.requireNonNull(properties, "Properties required");
 
-        final QueuedThreadPool threadPool = new QueuedThreadPool(properties.getWebThreads());
+        final VirtualThreadPool threadPool = new VirtualThreadPool(properties.getWebThreads());
```
exceptionfactory: I recommend leaving this as the `QueuedThreadPool` for now, because the web threads property is less meaningful with Virtual Threads. It may be worth considering deprecating it. Either way, it seems like something better to consider in a separate issue.

nifi-registry/pom.xml
```diff
@@ -35,6 +35,10 @@
         <module>nifi-registry-docker-maven</module>
     </modules>
     <properties>
+        <nifi.registry.jetty.version>12.0.25</nifi.registry.jetty.version>
```
exceptionfactory: Is there a reason to introduce this new property as opposed to overriding the value from the parent Maven configuration?

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/FlowSynchronizationIT.java
```diff
@@ -606,9 +606,14 @@ public void testAddControllerServiceReferencingExistingService() throws NiFiClie
 
         // Delete the CountFlowFiles processor, and countB and countC services, disable A.
         getClientUtil().stopProcessor(countFlowFiles);
+        // Ensure the processor has fully stopped before disabling services to avoid race conditions
+        getClientUtil().waitForStoppedProcessor(countFlowFiles.getId());
```
exceptionfactory, line 610: Should these changes be handled separately?

exceptionfactory (changes requested):
Thanks for the thorough work on these changes @pvillard31. The major of the adjustments appear to be straightforward.

Aside from the thread pool handling, the only code-related adjusted that raised questions is in `JettyServer`, with the associated changes to add the `jetty-deploy` dependency.

Regarding Spring Boot and NiFi Registry, with the way the application is structured, it seems like it should be possible to use 12.1 and override some references. Is the problem there related to Jakarta EE 11 as opposed to Jetty 12.1 itself? It is fine to address that separately, but it would be helpful to have more details.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/JettyServer.java
pvillard31: Can't remember why, reverting.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/pom.xml
pvillard31: Yes, this is one of the changes with Jetty 12.1: `DeploymentManager` no longer exists and I'm replacing it by `StandardDeployer` which comes from `jetty-deploy` in `JettyServer`

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/clustering/FlowSynchronizationIT.java
pvillard31: The test was systematically failing after the upgrade to Jetty 12.1. It may have been because of the virtual threads because it seems to be working fine now that I reverted the change. Will revert here as well.

nifi-registry/pom.xml
pvillard31: This is what I tried initially but it was kinda breaking things in terms of BOM import and mixing the EE10 vs EE11 layers. There may be a way to make things work but I found this approach cleaner

pvillard31:
For NiFi Registry, reviewing further, the problem is specifically related to EE11. I have made the changes to also uses Jetty 12.1.1 in NiFi Registry but keep the ee10 layer.

nifi-registry/pom.xml
exceptionfactory: Sounds good, thanks for the background.

pvillard31:
OK... so while using Jetty 12.1.1 works for NiFi Registry (I was able to properly build it, run it, and use it with NiFi), the integration tests are failing because of the way Spring Boot is used there (I remember now :)). And we can also see the failure on the system test that I was mentioning with the changes on `FlowSynchronizationIT`.

exceptionfactory:
> OK... so while using Jetty 12.1.1 works for NiFi Registry (I was able to properly build it, run it, and use it with NiFi), the integration tests are failing because of the way Spring Boot is used there (I remember now :)). And we can also see the failure on the system test that I was mentioning with the changes on `FlowSynchronizationIT`.

Thanks @pvillard31.

As far as the NiFI Registry integration tests, I am open to removing those because they do not follow the actual pattern of a running NiFi Registry. The NiFi Registry Docker tests also provide some basic checks for a running NiFi Registry. What do you think?

On the system test failure, it is a bit strange to see the change, but if the previous adjustments corrected it and were otherwise reasonable, that sounds good.

pvillard31:
Yeah so for NiFi Registry, I noticed that we were bringing Jetty through Spring Boot for running the tests. I removed this import to instead used the embedded Tomcat to still execute those tests successfully. I'm not against completely removing those.

exceptionfactory:
> Yeah so for NiFi Registry, I noticed that we were bringing Jetty through Spring Boot for running the tests. I removed this import to instead used the embedded Tomcat to still execute those tests successfully. I'm not against completely removing those.

That looks like a reasonable and minimal change to keep those tests passing. If that works and doesn't impact runtime behavior for NiFi Registry, that sounds good.

exceptionfactory (approved):
Thanks for the extensive work here @pvillard31.

On final validation, I noticed an exception thrown from `LogoutCompleteRedirectFilter` related to the `Fragment` portion of the redirect URI when passed to `sendRedirect()`. Changing the approach to set the `Location` header and send an HTTP 302 as in `QueryStringToFragmentFilter` worked as expected, so I will include that change in the merge commit.

PR #10253, opened 2025-08-28
NIFI-14918: Removing dependencies from LICENSE files and in allowedCo…

mcgilman:
…mmonJsDependencies in project.json for dependencies that are no longer used.


scottyaslan (approved):
Thanks for catching these! LGTM!

PR #10268, opened 2025-09-03
NIFI-14639: Explicitly popping off the top back navigation when the u…

mcgilman:
…ser clicks the back link.

The underlying issue can be reproduced by following the Go To Service option from a Processor to a Service. Then in the Service following the Go To Service option on another Service reference. So the flow would have a Processor dependent on a Service which is dependent on another Service. Once the user has followed this navigation, current `main` branch wouldn't allow the user to navigate all the way back to the Processor. This PR changes how we previously visited routes are popped off the stack to ensure that they user can break free from a given route boundary.


scottyaslan:
Reviewing...

scottyaslan (approved):
Confirmed the user is now able to return to the originating processor.

PR #10269, opened 2025-09-03
NIFI-14614 - Fix QueryRecord column/value misalignment with Calcite by enforcing projection pushdown

pvillard31:
# Summary

NIFI-14614 - Fix QueryRecord column/value misalignment with Calcite by enforcing projection pushdown

I have tested many (realllllyyyyy many) things/approaches but I think this is now properly fixed ! 🚀 

**Summary**
- Fixes incorrect value mapping in QueryRecord when SELECT reorders columns and for UNION ALL queries under the new Calcite integration
- Restores behavior parity with NiFi 1.x by pushing projections into the table scan so row values are produced in the exact SELECT order

**Root Cause**

With the NiFi 2.x calcite-utils integration, Calcite did not always push projections into the Scan. The JDBC layer then returned rows in base table order while ResultSet metadata reflected the SELECT projection, leading to label/value misalignment (e.g., reorders and UNION ALL appearing as duplicated/mislabeled rows)
  
**Changes**

- Projection Pushdown (calcite-utils):
    - Made NiFiProjectTableScanRule public and broadened it to match core Project
    - Relaxed operand to .anyInputs() and verified the input is NiFiTableScan in onMatch
    - Transform Project + NiFiTableScan to a projected NiFiTableScan carrying selected field indexes in the correct order
    - Ensured rule registration in NiFiTableScan.register() and added a safe constructor registration to avoid cases where register() is not invoked
  
- ResultSetRecordSet simplification (record module):
    - Since projection pushdown ensures SELECT order matches row index order, simplified value retrieval
    - Index-first (aligned to schema order), with label-based fallback only if index getter returns null (to keep mocks/drivers compatible).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-calcite-utils/src/main/java/org/apache/nifi/sql/internal/NiFiProjectTableScanRule.java
+++ b/nifi-commons/nifi-calcite-utils/src/main/java/org/apache/nifi/sql/internal/NiFiProjectTableScanRule.java
@@ -70,7 +70,6 @@ public class NiFiProjectTableScanRule extends RelRule<NiFiProjectTableScanRule.C
     }
 
     public interface Config extends RelRule.Config {
-        // This impl comes directly from the Calcite documentation.
         Config DEFAULT = new StandardConfig()
             .withOperandSupplier(b0 -> b0.operand(Project.class).anyInputs());
 


nifi-commons/nifi-calcite-utils/src/main/java/org/apache/nifi/sql/internal/NiFiTableScan.java
```diff
@@ -51,6 +51,15 @@ protected NiFiTableScan(final RelOptCluster cluster, final RelOptTable table, fi
...
         fieldExpression = Expressions.constant(fields);
+        // Ensure projection pushdown rule is registered with the planner, as some Calcite versions
+        // may not invoke the RelNode#register() method.
+        try {
+            cluster.getPlanner().addRule(new NiFiProjectTableScanRule(NiFiProjectTableScanRule.Config.DEFAULT));
+        } catch (Exception e) {
+            // Rule may already be registered; intentionally ignored to avoid duplicate registration errors.
+            // Touch the exception to avoid PMD EmptyCatchBlock warnings
+            e.getClass();
+        }
```
dan-s1: Naming the exception variable `ignored` informs PMD to leave this alone. See PMD [EmptyCatchBlock](https://pmd.github.io/pmd/pmd_rules_java_errorprone.html#emptycatchblock) documentation for more details
```suggestion
        } catch (Exception ignored) {
            // Rule may already be registered; intentionally ignored to avoid duplicate registration errors.
        }
```

pvillard31: oh nice, thanks, I didn't know that!

nifi-commons/nifi-calcite-utils/src/main/java/org/apache/nifi/sql/internal/NiFiProjectTableScanRule.java
```diff
@@ -19,24 +19,29 @@
...
     public void onMatch(final RelOptRuleCall call) {
-        final LogicalProject project = call.rel(0);
-        final NiFiTableScan scan = call.rel(1);
+        final Project project = call.rel(0);
+
+        // Attempt to locate NiFiTableScan as immediate input
+        if (!(project.getInput() instanceof NiFiTableScan)) {
+            return;
+        }
+        final NiFiTableScan scan = (NiFiTableScan) project.getInput();
```
dan-s1: Per Intellij, the declaration of  the `scan` variable can be replaced with a pattern variable
```suggestion
        // Attempt to locate NiFiTableScan as immediate input
        if (!(project.getInput() instanceof NiFiTableScan scan)) {
            return;
        }
```

pvillard31: Yep, good point, I need to use this pattern matching more

nifi-commons/nifi-calcite-utils/src/main/java/org/apache/nifi/sql/internal/NiFiProjectTableScanRule.java
```diff
@@ -68,8 +73,7 @@ private static int[] getProjectionFields(final List<RexNode> expressions) {
     public interface Config extends RelRule.Config {
         // This impl comes directly from the Calcite documentation.
```
dan-s1: With the recent changes, is this comment still relevant i.e. is this code still directly from the Calcite documentation?

pvillard31: I guess this was coming from https://calcite.apache.org/javadocAggregate/org/apache/calcite/plan/RelRule.html
So yeah we can probably remove that line.

```suggestion
```

PR #10275, opened 2025-09-04
NIFI-13838 - Allow Change Version on a ghosted component when bundle becomes available

pvillard31:
# Summary

NIFI-13838 - Allow Change Version on a ghosted component when bundle becomes available

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/api/dto/DtoFactoryTest.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/api/dto/DtoFactoryTest.java
@@ -16,11 +16,17 @@
  */
 package org.apache.nifi.web.api.dto;
 
+import org.apache.nifi.bundle.Bundle;
+import org.apache.nifi.bundle.BundleCoordinate;
+import org.apache.nifi.bundle.BundleDetails;
 import org.apache.nifi.components.AllowableValue;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.validation.ValidationStatus;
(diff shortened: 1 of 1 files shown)


pvillard31:
Thanks for the review @exceptionfactory.

My use case is the following: I have version X that comes with NiFi OOTB, then I add a version Y of that NAR using the nar-manager API., In my flow, I have an instance of the processor version Y. If I force delete the NAR (version Y) using the nar-manager API, then the component becomes ghosted and I have no option to switch it back to version X. This change is letting the UI know that it should show the change version option and let me switch back to version X.

This boolean `multipleVersionsAvailable` is only used by the UI as a hint to determine whether or not "Change Version" should be shown in the contextual menu of the component.

One could argue that if you have a ghost component version Y, and a bundle with the same coordinates except for a version X, you do have multiple versions in NiFi at this point in time. With this change, on the ghosted component, I can right click and change version to X, and once done, if I right click on the component, I won't see Change Version anymore.

I have not evaluated another approach but I guess we could imagine a completely new field that has a better meaning and make changes in the UI to use that new field but I really don't think it makes much sense as this current field would not be used anymore... So it would be kinda renaming with deprecation I guess...

I have not looked into adding a system test for this scenario but I can definitely look into adding one if you think this is valuable. Let me know your thoughts.

bbende:
One alternative could be to add a new field like `compatibleBundleCount` which sort of encompasses the existing `multipleVersionsAvailable` since checking the count can tell you if there is more than 1, but it can also be checked to see if when `isExtensionMissing` is `compatibleBundleCount` > 0, but then as you said it also requires coordinated UI change.

pvillard31:
Will update the test class 👍🏻

PR #10286, opened 2025-09-09
NIFI-14944 - Fix flow upgrade when CS reference goes from external to scoped controller service

pvillard31:
# Summary

NIFI-14944 - Fix flow upgrade when CS reference goes from external to scoped controller service

Current situation:

- Considering a versioned flow PG1 containing GenerateFlowFile -> ConvertRecord -> UpdateAttribute.
- Where, in v1, ConvertRecord references a JSON Reader and JSON Writer defined at root level (external controller services in the versioned flow definition).
- Where, in v2, ConvertRecord references newly created JSON Reader and JSON Writer defined at PG1 level.

Then, when upgrading from v1 to v2, ConvertRecord would still reference the external controller services, which is not expected.

This PR fixes the logic to identify the controller service that should be referenced. A unit test has been added to reproduce the above scenario with versioned flow definitions. I also added another unit test to show that if the processor ConvertRecord is in a nested PG2 process group, then we would still update the referenced controller services to the ones defined in PG1. We basically do a lookup starting in the current process group and then go up to the parent, and so on.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -1616,35 +1616,38 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
 
                 final String value;
                 if (descriptor != null && referencesService && (proposedProperties.get(propertyName) != null)) {
-                    // Need to determine if the component's property descriptor for this service is already set to an id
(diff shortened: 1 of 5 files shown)


exceptionfactory (commented):
Thanks for working on this @pvillard31. Given the age of the pull requests, it looks like it needs to be rebased, and the unit test could use some alignment with standard conventions.

pvillard31:
Thanks for the ping @exceptionfactory - I rebased the PR and simplified the approach / testing coverage

exceptionfactory (approved):
Thanks for revisiting and revising the tests @pvillard31, looks good!

PR #10287, opened 2025-09-09
NIFI-14950: Allowing individual state entries to be removed when the …

mcgilman:
…underlying component supports it.

This PR is the UI counterpart to https://github.com/apache/nifi/pull/10238. It should not be merged until the back end changes land first.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/component-state/component-state.component.html
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/component-state/component-state.component.html
@@ -117,7 +117,7 @@
                                             color="primary"
                                             (click)="clearComponentStateEntry(item)"
                                             [title]="'Clear state entry: ' + item.key">
-                                            <i class="fa fa-trash" aria-hidden="true"></i>
+                                            <i class="fa fa-trash"></i>
                                         </button>
                                     </td>
                                 </ng-container>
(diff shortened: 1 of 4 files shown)


pvillard31:
Thanks for the changes @mcgilman ! I cannot really comment on the code changes but I have built this PR in combination with #10238 and it works as expected based on the testing I did.

As a side-note, the MacOS build failure is because of the known issue that has been fixed in #10288 in case you want to rebase on `main` and have full green builds.

rfellows:
reviewing...

nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/component-state/component-state.component.html
```diff
@@ -89,6 +106,23 @@ <h2 mat-dialog-title>Component State</h2>
...
+                                <ng-container matColumnDef="actions">
+                                    <th mat-header-cell *matHeaderCellDef></th>
+                                    <td mat-cell *matCellDef="let item">
+                                        <button
+                                            [disabled]="isClearing"
+                                            mat-icon-button
+                                            color="primary"
+                                            (click)="clearComponentStateEntry(item)"
+                                            [title]="'Clear state entry: ' + item.key">
+                                            <i class="fa fa-trash" aria-hidden="true"></i>
```
rfellows: from MDN:
> `aria-hidden="true"` should not be used on elements that can receive focus

These trash can icons can be tabbed to, so we should not be setting this attribute.

PR #10291, opened 2025-09-10
NIFI-14953 Validate inlined protobuf schemas

lkuchars:
# Summary

[NIFI-14953](https://issues.apache.org/jira/browse/NIFI-14953)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/ProtobufSchemaCompiler.java
+++ b/nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/ProtobufSchemaCompiler.java
@@ -130,7 +130,8 @@ final class ProtobufSchemaCompiler {
                 logger.debug("Successfully compiled schema for identifier: {}", schemaDefinition.getIdentifier());
                 return compiledSchema;
             } catch (final IllegalStateException e) {
-                throw new SchemaCompilationException(e); // Illegal state exception is thrown by the wire library for schema issues
+                // Illegal state exception is thrown by the wire library for schema issues
+                throw new SchemaCompilationException("Could not compile schema: %s".formatted(schemaDefinition.toString()), e);
             } catch (final Exception e) {
                 throw new RuntimeException("Failed to compile Protobuf schema for identifier: " + schemaDefinition.getIdentifier(), e);
             }
--- a/nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
+++ b/nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
@@ -28,8 +28,10 @@ import org.apache.nifi.components.ValidationResult;
 import org.apache.nifi.context.PropertyContext;
 import org.apache.nifi.controller.AbstractControllerService;
 import org.apache.nifi.controller.ConfigurationContext;
+import org.apache.nifi.controller.ControllerServiceInitializationContext;
 import org.apache.nifi.logging.ComponentLog;
 import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.reporting.InitializationException;
 import org.apache.nifi.schema.access.SchemaNotFoundException;
 import org.apache.nifi.schemaregistry.services.MessageName;
 import org.apache.nifi.schemaregistry.services.MessageNameResolver;
@@ -52,7 +54,6 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HexFormat;
 import java.util.List;
 import java.util.Map;
@@ -139,7 +140,6 @@ public class StandardProtobufReader extends SchemaRegistryService implements Rec
         schemaText = context.getProperty(SCHEMA_TEXT);
         schemaBranchName = context.getProperty(SCHEMA_BRANCH_NAME);
         schemaVersion = context.getProperty(SCHEMA_VERSION);
-        schemaCompiler = new ProtobufSchemaCompiler(getIdentifier(), getLogger());
     }
 
     @Override
@@ -161,6 +161,12 @@ public class StandardProtobufReader extends SchemaRegistryService implements Rec
     }
 
     @Override
+    protected void init(final ControllerServiceInitializationContext config) throws InitializationException {
+        super.init(config);
+        schemaCompiler = new ProtobufSchemaCompiler(getIdentifier(), getLogger());
+    }
+
+    @Override
     protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
         final List<PropertyDescriptor> properties = new ArrayList<>(super.getSupportedPropertyDescriptors());
         properties.add(MESSAGE_NAME_RESOLUTION_STRATEGY);
@@ -176,7 +182,7 @@ public class StandardProtobufReader extends SchemaRegistryService implements Rec
 
     @Override
     protected Collection<ValidationResult> customValidate(final ValidationContext validationContext) {
-        final List<ValidationResult> problems = new ArrayList<>(super.customValidate(validationContext));
+        final List<ValidationResult> results = new ArrayList<>(super.customValidate(validationContext));
         final String schemaAccessStrategyValue = validationContext.getProperty(SCHEMA_ACCESS_STRATEGY).getValue();
 
         // Only validate when using Schema Text property strategy
@@ -184,23 +190,22 @@ public class StandardProtobufReader extends SchemaRegistryService implements Rec
             final PropertyValue schemaTextProperty = validationContext.getProperty(SCHEMA_TEXT);
             final String schemaTextValue = schemaTextProperty.getValue();
 
-            if (validationContext.isExpressionLanguageSupported(SCHEMA_TEXT.getName())
-                && validationContext.isExpressionLanguagePresent(schemaTextValue)) {
-                return Collections.emptyList();
+            if (validationContext.isExpressionLanguagePresent(schemaTextValue)) {
+                return results;
             }
 
             if (schemaTextValue == null || schemaTextValue.isBlank()) {
-                problems.add(new ValidationResult.Builder()
+                results.add(new ValidationResult.Builder()
                     .subject(SCHEMA_TEXT.getDisplayName())
                     .input(schemaTextValue)
                     .valid(false)
-                    .explanation("Schema Text cannot be empty when using \"Use 'Schema Text' Property\" strategy")
+                    .explanation("Schema Text value is missing")
                     .build());
-                return problems;
+                return results;
             }
 
             try {
-                // Try to compile the schema to validate it's valid protobuf format
+                // Try to compile the schema to validate protobuf format
                 final String hash = sha256Hex(schemaTextValue);
                 final SchemaIdentifier schemaIdentifier = SchemaIdentifier.builder()
                     .name(hash + PROTO_EXTENSION)
@@ -215,13 +220,13 @@ public class StandardProtobufReader extends SchemaRegistryService implements Rec
 
                     if (validationContext.isExpressionLanguageSupported(MESSAGE_NAME.getName())
                         && validationContext.isExpressionLanguagePresent(messageNameValue)) {
-                        return problems;
+                        return results;
                     }
 
                     if (messageNameValue != null && !messageNameValue.isBlank()) {
                         // Check if the message name exists in the compiled schema
                         if (compiledSchema.getType(messageNameValue) == null) {
-                            problems.add(new ValidationResult.Builder()
+                            results.add(new ValidationResult.Builder()
                                 .subject(MESSAGE_NAME.getDisplayName())
                                 .input(messageNameValue)
                                 .valid(false)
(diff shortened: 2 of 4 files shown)


nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/ProtobufSchemaCompiler.java
```diff
@@ -129,7 +129,8 @@ private Schema compileSchemaDefinition(final SchemaDefinition schemaDefinition)
                 final Schema compiledSchema = createAndLoadSchema(tempDir);
                 logger.debug("Successfully compiled schema for identifier: {}", schemaDefinition.getIdentifier());
                 return compiledSchema;
-
+            } catch (final IllegalStateException e) {
+                throw new SchemaCompilationException(e); // Illegal state exception is thrown by the wire library for schema issues
```
exceptionfactory: A message should be included along with the `IllegalStateException` cause

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/test/java/org/apache/nifi/services/protobuf/TestStandardProtobufReaderPropertyValidation.java
```diff
@@ -84,6 +84,62 @@ void testValidWhenSchemaTextPropertyRemoved() { // default value of ${proto.sche
...
+                }
+                """;
+
+            runner.setProperty(standardProtobufReader, SCHEMA_TEXT, validProtobufSchema);
+            runner.setProperty(standardProtobufReader, MESSAGE_NAME_RESOLUTION_STRATEGY, MESSAGE_NAME_PROPERTY);
+            runner.setProperty(standardProtobufReader, MESSAGE_NAME, "test.NonExistentMessage");
+
+            final ValidationResult invalidResult = verifyExactlyOneValidationError();
+
+            assertTrue(invalidResult.getExplanation().contains("Message name 'test.NonExistentMessage' cannot be found"));
```
exceptionfactory: Recommend reducing the search string:
```suggestion
            assertTrue(invalidResult.getExplanation().contains("test.NonExistentMessage"));
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
```diff
@@ -170,6 +174,84 @@ protected PropertyDescriptor buildSchemaTextProperty() {
...
+                    .valid(false)
+                    .explanation("Invalid protobuf schema format: " + e.getMessage())
+                    .build());
+            }
+        }
+
+        return problems;
+    }
+
+    // Compile the schema to validate it's correct. The method is used only for validation purposes.
```
exceptionfactory: This comment mostly reiterates the method name, and the usage in validation is clear, so recommend removing this comment.

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
```diff
@@ -170,6 +174,84 @@ protected PropertyDescriptor buildSchemaTextProperty() {
         return PROTOBUF_SCHEMA_TEXT;
     }
 
+    @Override
+    protected Collection<ValidationResult> customValidate(final ValidationContext validationContext) {
+        final List<ValidationResult> problems = new ArrayList<>(super.customValidate(validationContext));
```
exceptionfactory: Results can also indicate a valid status, so the variable name should be changed.
```suggestion
        final List<ValidationResult> results = new ArrayList<>(super.customValidate(validationContext));
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
```diff
@@ -170,6 +174,84 @@ protected PropertyDescriptor buildSchemaTextProperty() {
...
+        final String schemaAccessStrategyValue = validationContext.getProperty(SCHEMA_ACCESS_STRATEGY).getValue();
+
+        // Only validate when using Schema Text property strategy
+        if (SCHEMA_TEXT_PROPERTY.getValue().equals(schemaAccessStrategyValue)) {
+            final PropertyValue schemaTextProperty = validationContext.getProperty(SCHEMA_TEXT);
+            final String schemaTextValue = schemaTextProperty.getValue();
+
+            if (validationContext.isExpressionLanguageSupported(SCHEMA_TEXT.getName())
```
exceptionfactory: Is the `isExpressionLanguageSupported` check needed?

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
```diff
@@ -170,6 +174,84 @@ protected PropertyDescriptor buildSchemaTextProperty() {
...
+                && validationContext.isExpressionLanguagePresent(schemaTextValue)) {
+                return Collections.emptyList();
```
exceptionfactory: Creating the results array and then having multiple return statements is not ideal. It would be better to refactor this method to have a single return at the end.

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
```diff
@@ -170,6 +174,84 @@ protected PropertyDescriptor buildSchemaTextProperty() {
...
+            }
+
+            if (schemaTextValue == null || schemaTextValue.isBlank()) {
+                problems.add(new ValidationResult.Builder()
+                    .subject(SCHEMA_TEXT.getDisplayName())
+                    .input(schemaTextValue)
+                    .valid(false)
+                    .explanation("Schema Text cannot be empty when using \"Use 'Schema Text' Property\" strategy")
```
exceptionfactory: Recommend avoiding the reference to the other property name:
```suggestion
                    .explanation("Schema Text value is missing)
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
```diff
@@ -170,6 +174,84 @@ protected PropertyDescriptor buildSchemaTextProperty() {
...
+                    .build());
+                return problems;
+            }
+
+            try {
+                // Try to compile the schema to validate it's valid protobuf format
```
exceptionfactory:
```suggestion
                // Try to compile the schema to validate protobuf format
```

exceptionfactory (changes requested):
Thanks for this additional validation @lkuchars.

As this requires writing out temporary files, it seems like it could be a bit expensive as written, given that `customValidate` can be triggered many times. One option is to implement cached checking of the schema so that it only runs the compilation when the Schema Text property changes. Another option is to implement this as a manual verify method, requiring user interaction. It seems better to consider the verify method approach, to avoid complexity around caching, and because this validation only applies to the Schema Text option, but I'm open to considering either approach.

lkuchars:
> Thanks for this additional validation @lkuchars.
> 
> As this requires writing out temporary files, it seems like it could be a bit expensive as written, given that `customValidate` can be triggered many times. One option is to implement cached checking of the schema so that it only runs the compilation when the Schema Text property changes. Another option is to implement this as a manual verify method, requiring user interaction. It seems better to consider the verify method approach, to avoid complexity around caching, and because this validation only applies to the Schema Text option, but I'm open to considering either approach.

@exceptionfactory thank you for taking your time to look at this. It would be nice to get the validation of the schema without manual intervention. I think the original ProtobufReader worked that way, and it used cache.  I think your suggestion with a cache is spot on. 
I'll just remove the throwaway compiler
`// Create a throwaway schema compiler for validation to avoid polluting the main compiler cache.
        final ProtobufSchemaCompiler validationCompiler = new ProtobufSchemaCompiler(getIdentifier() + "_validation", getLogger());`
I think my concerns about not polluting the global compiler cache are unjustified.
I can use the schema compiler initialized in onEnabled. That way, subsequent validation calls will use the cache implementation from the ProtobufSchemaCompiler.
WDYT?

exceptionfactory:
> @exceptionfactory thank you for taking your time to look at this. It would be nice to get the validation of the schema without manual intervention. I think the original ProtobufReader worked that way, and it used cache. I think your suggestion with a cache is spot on. I'll just remove the throwaway compiler `// Create a throwaway schema compiler for validation to avoid polluting the main compiler cache. final ProtobufSchemaCompiler validationCompiler = new ProtobufSchemaCompiler(getIdentifier() + "_validation", getLogger());` I think my concerns about not polluting the global compiler cache are unjustified. I can use the schema compiler initialized in onEnabled. That way, subsequent validation calls will use the cache implementation from the ProtobufSchemaCompiler. WDYT?

Thanks for the reply @lkuchars. Using the same `schemaCompiler` instance could work, but it would have to be instantiated somewhere else, because `OnEnabled` methods are invoked manually, after validation.

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
lkuchars: If we refactor to have a single return, we would need to use nested if-else statements or boolean flags, which would actually make the code less readable and more complex. The early returns here serve a legitimate purpose - they represent different validation scenarios where further validation is either unnecessary or impossible.

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/ProtobufSchemaCompiler.java
lkuchars: fixed

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/test/java/org/apache/nifi/services/protobuf/TestStandardProtobufReaderPropertyValidation.java
lkuchars: fixed

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
lkuchars: fixed

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
lkuchars: fixed

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
lkuchars: removed

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
lkuchars: fixed

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufReader.java
lkuchars: fixed

lkuchars:
> > @exceptionfactory thank you for taking your time to look at this. It would be nice to get the validation of the schema without manual intervention. I think the original ProtobufReader worked that way, and it used cache. I think your suggestion with a cache is spot on. I'll just remove the throwaway compiler `// Create a throwaway schema compiler for validation to avoid polluting the main compiler cache. final ProtobufSchemaCompiler validationCompiler = new ProtobufSchemaCompiler(getIdentifier() + "_validation", getLogger());` I think my concerns about not polluting the global compiler cache are unjustified. I can use the schema compiler initialized in onEnabled. That way, subsequent validation calls will use the cache implementation from the ProtobufSchemaCompiler. WDYT?
> 
> Thanks for the reply @lkuchars. Using the same `schemaCompiler` instance could work, but it would have to be instantiated somewhere else, because `OnEnabled` methods are invoked manually, after validation.

I moved the initialization of the schemaCompiler to the init() override.

PR #10292, opened 2025-09-10
NIFI-14892 Aligned those processors and controller services whose property names do not match the human-friendly display names in extended bundles nifi-airtable-bundle, nifi-amqp-bundle, nifi-asana-bundle, nifi-asn1-bundle and nifi-aws-bundle.

dan-s1:
# Summary

[NIFI-14892](https://issues.apache.org/jira/browse/NIFI-14892)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation


dan-s1:
@exceptionfactory I understand for the those unit tests which test the properties defined in the actual processor. But what about processors which inherit properties from a parent and have their own properties defined? I would think we would want to to test `super.migrateProperties` was called to ensure all properties are renamed. In fact I had pointed out to you on the Jira ticket a case in point where `super.migrateProperties` was not called but should have been. Also when working the Azure code I noticed there are properties defined in utility classes and not the actual processor/controller service and used across various processors/controller services. Wouldn't it be a good idea to to test those migrations?

exceptionfactory:
Thanks for the reply @dan-s1, that's a good point about Processors with migrate methods in super classes, or those that use utilities for migration. I am supportive of unit tests for those scenarios. What do you think about using that as the decision point, and removing some of the unit tests where the migration is a simple one-to-one in the same class?

dan-s1:
@exceptionfactory I just want to clarify what I should do for those classes which inherit properties from a parent but do not have properties of there own to rename hence they do not define override `migrateProperties`.  In that case only the parent `migrateProperties` is called.  Its not exactly a simple one-to-one but it possibly could be considered one. Do we want unit tests for those scenarios (e.g. `CopyS3Object`) to ensure those properties were renamed?


Also a separate question I have is I noticed that in `AbstractAwsMachineLearningJobStatusProcessor` it has the line

`config.renameProperty("aws-region", REGION.getName());`

The `REGION` property is defined in `AbstractAwsProcessor` and is used in many AWS related classes yet I do not see a renaming in any other class except this one. I added a rename to `PutSQS` although I am not sure if I should have .  I am thinking the rename should probably be in done in the `migrateProperties` method of `AbstractAwsProcessor` as it is the ultimate parent of many of the AWS related processors. Please advise.

dan-s1:
@exceptionfactory If you are okay with it, I am going to pull up the 
`config.renameProperty("aws-region", REGION.getName());`
to `AbstractAwsProcessor` where I believe it rightfully should be.

exceptionfactory:
> @exceptionfactory If you are okay with it, I am going to pull up the `config.renameProperty("aws-region", REGION.getName());` to `AbstractAwsProcessor` where I believe it rightfully should be.

Sounds good, thanks.

dan-s1:
Please see the tables in the Summary above detailing which classes are overriding `migrateProperties` and should be calling  `super.migrateProperties`.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -210,6 +210,11 @@ protected AMQPPublisher createAMQPWorker(final ProcessContext context, final Con
         return new AMQPPublisher(connection, getLogger());
     }
 
+    @Override
+    public void migrateProperties(final PropertyConfiguration config) {
```
exceptionfactory, line 214: It looks like is missing a `super.migrateProperties()` call.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/ConsumeAMQP.java
```diff
@@ -181,6 +175,17 @@ public class ConsumeAMQP extends AbstractAMQPProcessor<AMQPConsumer> {
 
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
+    @Override
+    public void migrateProperties(final PropertyConfiguration config) {
```
exceptionfactory, line 179: Missing `super.migrateProperties()`

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/PutKinesisStream.java
```diff
@@ -95,8 +94,7 @@ public class PutKinesisStream extends AbstractAwsSyncProcessor<KinesisClient, Ki
             .build();
 
     public static final PropertyDescriptor MAX_MESSAGE_BUFFER_SIZE_MB = new PropertyDescriptor.Builder()
-            .name("max-message-buffer-size")
-            .displayName("Max message buffer size (MB)")
+            .name("Max message buffer size (MB)")
```
exceptionfactory: The name should be changed to standard conventions:
```suggestion
            .name("Max Message Buffer Size")
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/PutKinesisStream.java
```diff
@@ -95,8 +94,7 @@ public class PutKinesisStream extends AbstractAwsSyncProcessor<KinesisClient, Ki
...
             .description("Max message buffer size in Mega-bytes")
```
exceptionfactory: This should also be adjusted.
```suggestion
            .description("Max message buffer size defined with standard data size units")
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/PutKinesisStream.java
```diff
@@ -105,8 +103,7 @@ public class PutKinesisStream extends AbstractAwsSyncProcessor<KinesisClient, Ki
             .build();
 
     static final PropertyDescriptor KINESIS_STREAM_NAME = new PropertyDescriptor.Builder()
-            .name("kinesis-stream-name")
-            .displayName("Amazon Kinesis Stream Name")
+            .name("Amazon Kinesis Stream Name")
```
exceptionfactory: This can be simplified:
```suggestion
            .name("Stream Name")
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/stream/PutKinesisStream.java
```diff
@@ -76,17 +77,15 @@ public class PutKinesisStream extends AbstractAwsSyncProcessor<KinesisClient, Ki
     public static final String AWS_KINESIS_SEQUENCE_NUMBER = "aws.kinesis.sequence.number";
 
     public static final PropertyDescriptor KINESIS_PARTITION_KEY = new PropertyDescriptor.Builder()
-        .displayName("Amazon Kinesis Stream Partition Key")
-        .name("amazon-kinesis-stream-partition-key")
+        .name("Amazon Kinesis Stream Partition Key")
```
exceptionfactory: "Amazon Kinesis" is redundant since it is in the context of the Processor.
```suggestion
        .name("Stream Partition Key")
```

exceptionfactory (approved):
Thanks @dan-s1, the latest version looks good. +1 merging

dan-s1:
@exceptionfactory  Thanks for approving and merging!

PR #10294, opened 2025-09-11
NIFI-14956 - AWS Credentials Provider - Add support for STS AssumeRoleWithWebIdentity

pvillard31:
# Summary

NIFI-14956 - AWS Credentials Provider - Add support for STS AssumeRoleWithWebIdentity

**Given the EOL of the SDK v1, I only implemented a version for the SDK v2**

Steps to setup an AWS only test environment for testing this new feature.
(note the commands are using `--profile perso`, change base on your needs)
### Setup variables

````
export REGION=eu-central-1
export POOL_NAME=nifi-oidc-demo
export USERNAME=test
export PASSWORD='NewStrongPassword#123'
export DOMAIN_PREFIX=nifi-demo-pv
export CONF_CLIENT_NAME=NiFi
export CALLBACK_URL=http://localhost:53682/callback
export LOGOUT_URL=http://localhost:53682/logout
````

### Create Cognito User Pool

````
aws cognito-idp create-user-pool --region $REGION --pool-name "$POOL_NAME" --policies 'PasswordPolicy={MinimumLength=12,RequireUppercase=true,RequireLowercase=true,RequireNumbers=true,RequireSymbols=true}' --auto-verified-attributes email --profile perso


USER_POOL_ID=$(aws cognito-idp list-user-pools --region "$REGION" --max-results 60 --profile perso | jq -r --arg name "$POOL_NAME" '.UserPools[] | select(.Name==$name) | .Id')
````

### Create Hosted UI Domain

````
aws cognito-idp create-user-pool-domain --region $REGION --user-pool-id $USER_POOL_ID --domain $DOMAIN_PREFIX --profile perso
````

### Create Confidential App Client (with Auth Code Flow)

````
aws cognito-idp create-user-pool-client \
    --region $REGION \
    --user-pool-id $USER_POOL_ID \
    --client-name "$CONF_CLIENT_NAME" \
    --generate-secret \
    --allowed-o-auth-flows-user-pool-client \
    --allowed-o-auth-flows code \
    --supported-identity-providers COGNITO \
    --allowed-o-auth-scopes openid email profile \
    --callback-urls "$CALLBACK_URL" \
    --logout-urls "$LOGOUT_URL" \
    --explicit-auth-flows ALLOW_USER_PASSWORD_AUTH ALLOW_USER_SRP_AUTH ALLOW_REFRESH_TOKEN_AUTH \
    --profile perso
````

From the output, capture the below:

````
export CONF_CLIENT_ID=...
export CONF_CLIENT_SECRET=...
````

And:

````
export TOKEN_ENDPOINT=https://$DOMAIN_PREFIX.auth.$REGION.amazoncognito.com/oauth2/token
export ISSUER=https://cognito-idp.$REGION.amazonaws.com/$USER_POOL_ID
````

### Create and confirm a user

````
aws cognito-idp admin-create-user --region $REGION --user-pool-id $USER_POOL_ID --username $USERNAME --temporary-password "$PASSWORD" --profile perso

aws cognito-idp admin-set-user-password --region $REGION --user-pool-id $USER_POOL_ID --username $USERNAME --password "$PASSWORD" --permanent --profile perso
````

### Get a Refresh Token for the user

````
SECRET_HASH=$(printf '%s' "$USERNAME""$CONF_CLIENT_ID" | openssl dgst -sha256 -hmac "$CONF_CLIENT_SECRET" -binary | base64)

REFRESH_TOKEN=$(aws cognito-idp initiate-auth \
        --region $REGION \
        --client-id $CONF_CLIENT_ID \
        --auth-flow USER_PASSWORD_AUTH \
        --auth-parameters USERNAME=$USERNAME,PASSWORD="$PASSWORD",SECRET_HASH="$SECRET_HASH" \
        --profile perso \
        | jq -r .AuthenticationResult.RefreshToken)
````

### Go to AWS IAM and Create Role

- Create Role
- Select Web Identity
- Select your Cognito pool
- Select the audience with the client ID of your app
- Next
- Select a policy, for example `AmazonSQSFullAccess`
- Next
- Give a name
- Create Role

Capture the Role ARN for the role you just created. Example:

````
export ROLE_ARN=arn:aws:iam::802551555016:role/nifi-test
````

### Configure NiFi

  - `StandardOauth2AccessTokenProvider`
      - Authorization Server URL: `$TOKEN_ENDPOINT`
      - Client Authentication Strategy: Basic Authentication
      - Grant Type: Refresh Token
      - Refresh Token: `$REFRESH_TOKEN`
      - Client ID: `$CONF_CLIENT_ID`
      - Client Secret: `$CONF_CLIENT_SECRET`
      - Scope: (leave empty)

  - AWS Credentials Provider (Web Identity)
      - OAuth2 Access Token Provider: select provider above
      - Assume Role ARN: `$ROLE_ARN`
      - Assume Role Session Name: `nifi-oidc-test`
      - Assume Role STS Region: `$REGION`

  - Use an AWS SDK v2 processor (e.g., `PutSQS`) with this credentials provider and test

### Important notes

  - STS requires `aud`; using `refresh_token` (or code) yields an `id_token` with `aud` that matches the app client ID.
  - Only v2 implementation so cannot be used with S3 processors at the moment as they are still using SDK v1.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/AssumeRoleCredentialsStrategy.java
+++ b/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/AssumeRoleCredentialsStrategy.java
@@ -37,8 +37,8 @@ import software.amazon.awssdk.services.sts.model.AssumeRoleRequest;
 import javax.net.ssl.SSLContext;
 import java.net.Proxy;
 import java.net.URI;
-import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 
 import static org.apache.nifi.processors.aws.credentials.provider.service.AWSCredentialsProviderControllerService.ASSUME_ROLE_ARN;
 import static org.apache.nifi.processors.aws.credentials.provider.service.AWSCredentialsProviderControllerService.ASSUME_ROLE_EXTERNAL_ID;
@@ -89,22 +89,11 @@ public class AssumeRoleCredentialsStrategy extends AbstractCredentialsStrategy {
     @Override
     public Collection<ValidationResult> validate(final ValidationContext validationContext,
                                                  final CredentialsStrategy primaryStrategy) {
-        final Collection<ValidationResult> validationFailureResults  = new ArrayList<>();
-
-        final boolean assumeRoleArnIsSet = validationContext.getProperty(ASSUME_ROLE_ARN).isSet();
-
-        if (assumeRoleArnIsSet) {
-            final Integer maxSessionTime = validationContext.getProperty(MAX_SESSION_TIME).asInteger();
-
-            // Session time only b/w 900 to 3600 sec (see software.amazon.awssdk.services.sts.model.AssumeRoleRequest#durationSeconds)
-            if (maxSessionTime < 900 || maxSessionTime > 3600) {
-                validationFailureResults.add(new ValidationResult.Builder().valid(false).input(maxSessionTime + "")
-                        .explanation(MAX_SESSION_TIME.getDisplayName() +
-                                " must be between 900 and 3600 seconds").build());
-            }
-        }
-
-        return validationFailureResults;
+        // Assume Role participates as a derived strategy or reused property group.
+        // Do not produce cross-strategy validation failures here; required/missing
+        // fields are enforced by PropertyDescriptor requirements and selected
+        // strategies, and derived selection is handled separately.
+        return Collections.emptyList();
     }
 
     @Override
--- a/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/WebIdentityCredentialsStrategy.java
+++ b/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/WebIdentityCredentialsStrategy.java
@@ -46,6 +46,7 @@ import java.net.URI;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
 
@@ -65,7 +66,7 @@ import static org.apache.nifi.processors.aws.credentials.provider.service.AWSCre
 public class WebIdentityCredentialsStrategy extends AbstractCredentialsStrategy implements CredentialsStrategy {
 
     public WebIdentityCredentialsStrategy() {
-        super("Web Identity (OIDC)", new PropertyDescriptor[]{
+        super("Web Identity", new PropertyDescriptor[]{
                 OAUTH2_ACCESS_TOKEN_PROVIDER,
                 ASSUME_ROLE_ARN,
                 ASSUME_ROLE_NAME
@@ -76,7 +77,7 @@ public class WebIdentityCredentialsStrategy extends AbstractCredentialsStrategy 
     public Collection<ValidationResult> validate(final ValidationContext validationContext, final CredentialsStrategy primaryStrategy) {
         // Avoid cross-strategy validation conflicts: Web Identity reuses Assume Role properties.
         // Controller-level validation enforces required combinations when OAuth2 is configured.
-        return null;
+        return Collections.emptyList();
     }
 
     @Override
@@ -165,12 +166,12 @@ public class WebIdentityCredentialsStrategy extends AbstractCredentialsStrategy 
                     return cached;
                 }
 
-                final String webIdentityJwt = getWebIdentityToken();
+                final String webIdentityToken = getWebIdentityToken();
 
                 final AssumeRoleWithWebIdentityRequest.Builder reqBuilder = AssumeRoleWithWebIdentityRequest.builder()
                         .roleArn(roleArn)
                         .roleSessionName(roleSessionName)
-                        .webIdentityToken(webIdentityJwt);
+                        .webIdentityToken(webIdentityToken);
 
                 if (sessionSeconds != null) {
                     reqBuilder.durationSeconds(sessionSeconds);
@@ -196,19 +197,22 @@ public class WebIdentityCredentialsStrategy extends AbstractCredentialsStrategy 
             // Prefer id_token when available
             final Map<String, Object> additional = accessToken.getAdditionalParameters();
             if (additional != null) {
-                final Object idTokenObj = additional.get("id_token");
-                if (idTokenObj instanceof String && !((String) idTokenObj).isEmpty()) {
-                    return (String) idTokenObj;
+                final String idToken = (String) additional.get("id_token");
+                if (idToken != null) {
+                    if (StringUtils.isBlank(idToken)) {
+                        throw new IllegalStateException("OAuth2AccessTokenProvider returned an empty id_token");
+                    } else {
+                        return idToken;
+                    }
                 }
             }
 
-            final String token = accessToken.getAccessToken();
-
-            if (StringUtils.isBlank(token)) {
+            final String accessTokenValue = accessToken.getAccessToken();
+            if (StringUtils.isBlank(accessTokenValue)) {
                 throw new IllegalStateException("No usable token found in AccessToken (id_token or access_token)");
+            } else {
+                return accessTokenValue;
             }
-
-            return token;
         }
     }
 }
--- a/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/service/AWSCredentialsProviderControllerService.java
+++ b/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/service/AWSCredentialsProviderControllerService.java
@@ -223,6 +223,7 @@ public class AWSCredentialsProviderControllerService extends AbstractControllerS
         .description("Controller Service providing OAuth2/OIDC tokens to exchange for AWS temporary credentials using STS AssumeRoleWithWebIdentity.")
         .identifiesControllerService(OAuth2AccessTokenProvider.class)
         .required(false)
+        .dependsOn(ASSUME_ROLE_ARN)
         .build();
 
 
@@ -342,6 +343,17 @@ public class AWSCredentialsProviderControllerService extends AbstractControllerS
                         .subject(ASSUME_ROLE_ARN.getDisplayName())
                         .valid(false)
                         .explanation("Web Identity (OIDC) requires both '" + ASSUME_ROLE_ARN.getDisplayName() + "' and '" + ASSUME_ROLE_NAME.getDisplayName() + "' to be set")
+                        .build());
+            }
+        }
+
+        if (validationContext.getProperty(ASSUME_ROLE_ARN).isSet()) {
+            final Integer maxSessionTime = validationContext.getProperty(MAX_SESSION_TIME).asInteger();
+            if (maxSessionTime != null && (maxSessionTime < 900 || maxSessionTime > 3600)) {
+                validationFailureResults.add(new ValidationResult.Builder()
+                        .subject(MAX_SESSION_TIME.getDisplayName())
+                        .valid(false)
+                        .explanation(MAX_SESSION_TIME.getDisplayName() + " must be between 900 and 3600 seconds")
                         .build());
             }
         }

### CODE REVIEW

**Comment by pvillard31** [comment]
Thanks for the feedback @exceptionfactory. As I noted in the PR description, what you're pointing out is a pure limitation of using Cognito but I wanted to provide an example that is easy to reproduce for a reviewer willing to try the change in AWS. I would not expect users to actually use Cognito in this context.

- STS WebIdentity is not inherently end‑user only. AWS uses it for machine identities too (e.g., EKS IRSA). What STS requires is a valid OIDC JWT with an issuer it trusts and an aud that matches a registered client ID.
- Cognito client‑credentials returns an access token without aud, and Cognito does not let you override standard claims for access tokens. It also does not issue ID tokens for client‑credentials. So Cognito + client‑credentials cannot satisfy STS trust policy checks wrt aud.

I also tested this change with NiFi running on GCP:
- NiFi runs on GCP
- I created a dummy OAuth Token Provider to get a Google OIDC ID token using the metadata server of my GCE instance
- Created a Google OIDC provider in AWS and specify an audience such as sts.amazonaws.com
- Then create a role using WebIdentity referencing that identity provider and also specifying the audience (as well as the service account attached to my GCE instance)

As you can see this is way more involved for a reviewer and I did have to build a custom NAR for the token provider implementation in GCP to get the token.

For reference:
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-idp_oidc.html
https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html
https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html

The last link gives an example where STS AssumeRoleWithWebIdentity is used with a service account in the context of EKS with the below role:

````json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "arn:aws:iam::$account_id:oidc-provider/$oidc_provider"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "$oidc_provider:aud": "sts.amazonaws.com",
          "$oidc_provider:sub": "system:serviceaccount:$namespace:$service_account"
        }
      }
    }
  ]
}
````

Let me know what you think.

File: nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/AssumeRoleCredentialsStrategy.java [hunk 1]
```diff
@@ -74,18 +75,32 @@ public class AssumeRoleCredentialsStrategy extends AbstractCredentialsStrategy {
...
     }
 
+    @Override
+    public Collection<ValidationResult> validate(final ValidationContext validationContext,
+                                                 final CredentialsStrategy primaryStrategy) {
+        // Assume Role participates as a derived strategy or reused property group.
+        // Do not produce cross-strategy validation failures here; required/missing
+        // fields are enforced by PropertyDescriptor requirements and selected
+        // strategies, and derived selection is handled separately.
+        return null;
```
exceptionfactory: This should return an empty collection instead of `null`.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/WebIdentityCredentialsStrategy.java
```diff
@@ -0,0 +1,219 @@
...
+import static org.apache.nifi.processors.aws.credentials.provider.service.AWSCredentialsProviderControllerService.OAUTH2_ACCESS_TOKEN_PROVIDER;
+
+/**
+ * Supports AWS SDK v2 credentials using STS AssumeRoleWithWebIdentity and an OAuth2/OIDC token
+ * provided by an {@link OAuth2AccessTokenProvider}. This is a primary strategy for SDK v2 only.
+ */
+public class WebIdentityCredentialsStrategy extends AbstractCredentialsStrategy implements CredentialsStrategy {
+
+    public WebIdentityCredentialsStrategy() {
+        super("Web Identity (OIDC)", new PropertyDescriptor[]{
```
exceptionfactory: I recommend removing `(OIDC)` and just mentioning it in the description where applicable.
```suggestion
        super("Web Identity", new PropertyDescriptor[]{
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/WebIdentityCredentialsStrategy.java
```diff
@@ -0,0 +1,219 @@
...
+                ASSUME_ROLE_ARN,
+                ASSUME_ROLE_NAME
+        });
+    }
+
+    @Override
+    public Collection<ValidationResult> validate(final ValidationContext validationContext, final CredentialsStrategy primaryStrategy) {
+        // Avoid cross-strategy validation conflicts: Web Identity reuses Assume Role properties.
+        // Controller-level validation enforces required combinations when OAuth2 is configured.
+        return null;
```
exceptionfactory: I believe this should return an empty Collection instead of `null`.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/WebIdentityCredentialsStrategy.java
```diff
@@ -0,0 +1,219 @@
...
+            if (current != null && currentExpiration != null && now.isBefore(currentExpiration.minus(SKEW))) {
+                return current;
+            }
+
+            synchronized (this) {
+                if (cached != null && expiration != null && Instant.now().isBefore(expiration.minus(SKEW))) {
+                    return cached;
+                }
+
+                final String webIdentityJwt = getWebIdentityToken();
```
exceptionfactory: Minor, but I recommend retaining `Token` in the variable name:
```suggestion
                final String webIdentityToken = getWebIdentityToken();
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/WebIdentityCredentialsStrategy.java
```diff
@@ -0,0 +1,219 @@
...
+                }
+            }
+
+            final String token = accessToken.getAccessToken();
+
+            if (StringUtils.isBlank(token)) {
+                throw new IllegalStateException("No usable token found in AccessToken (id_token or access_token)");
+            }
+
+            return token;
```
exceptionfactory: I recommend refactoring this method to a single return.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/factory/strategies/WebIdentityCredentialsStrategy.java
```diff
@@ -0,0 +1,219 @@
...
+            final AccessToken accessToken = oauth2AccessTokenProvider.getAccessDetails();
+            if (accessToken == null) {
+                throw new IllegalStateException("OAuth2AccessTokenProvider returned null AccessToken");
+            }
+
+            // Prefer id_token when available
+            final Map<String, Object> additional = accessToken.getAdditionalParameters();
+            if (additional != null) {
+                final Object idTokenObj = additional.get("id_token");
+                if (idTokenObj instanceof String && !((String) idTokenObj).isEmpty()) {
```
exceptionfactory: For simplicity, and clarity of behavior, I recommend simply checking of `idToken` is not null, and the throwing an `IllegalStateException` if the string itself is empty.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/service/AWSCredentialsProviderControllerService.java
```diff
@@ -260,6 +263,12 @@ public class AWSCredentialsProviderControllerService extends AbstractControllerS
         .dynamicallyModifiesClasspath(true)
         .build();
 
+    public static final PropertyDescriptor OAUTH2_ACCESS_TOKEN_PROVIDER = new PropertyDescriptor.Builder()
+        .name("OAuth2 Access Token Provider")
+        .description("Controller Service providing OAuth2/OIDC tokens to exchange for AWS temporary credentials using STS AssumeRoleWithWebIdentity.")
+        .identifiesControllerService(OAuth2AccessTokenProvider.class)
+        .required(false)
```
exceptionfactory, line 225: What options are available to make this dependent on other properties? Perhaps the `Assume Role ARN`? It may also be helpful to introduce some type of strategy property where `WebIdentity` is one of several allowed values, on which this property would depend. That should help make it clear when this service is actually needed.

exceptionfactory (changes requested):
Thanks for the helpful and substantive background on WebIdentity @pvillard31. The use case for platforms other than web application is clear now, and I see why the Cognito example was useful for testing purposes, if not applicable for general use cases.

I provided some feedback on the implementation details. Most of it seems clear, although I think some adjustment of property descriptors would help make it clear when this particular type of credentials provider is being configured.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/credentials/provider/service/AWSCredentialsProviderControllerService.java
pvillard31: Yeah adding `.dependsOn(ASSUME_ROLE_ARN)`. I did think about adding a strategy property, but it was not obvious how to make it work and be 100% sure that it would not break existing flows during an upgrade. We have quite a lot of possible combinations here.

pvillard31:
Thanks @exceptionfactory, rebased against main, hopefully I didn't mess it up :)

exceptionfactory (approved):
Thanks @pvillard31, the latest version looks good! +1 merging

dbuijs:
I suspect this pull request may have changed the behaviour of the AWSCredentialsProviderControllerService. In AWS with IRSA configured as per the AWS documentation, I can run Nifi 2.6.0 and simply by configuring it to use default credentials it just works and assumes the AWS Role. 

In Nifi 2.7.0 and later (2.7.2, 2.8.0) the AWSCredentialsProviderControllerService will validate with these settings, but when I try to use it for ListS3Bucket, it fails validation with:

```
nifi-2-node.nifi-headless.nifi.svc.cluster.local:8443 - Failed to list contents of bucket 'swipilot-onyxia': Unable to load credentials from any of the providers in the chain AwsCredentialsProviderChain(credentialsProviders=[SystemPropertyCredentialsProvider(), EnvironmentVariableCredentialsProvider(), WebIdentityTokenFileCredentialsProvider(), ProfileCredentialsProvider(profileName=default, profileFile=ProfileFile(sections=[])), ContainerCredentialsProvider(), InstanceProfileCredentialsProvider()]) : [SystemPropertyCredentialsProvider(): Unable to load credentials from system settings. Access key must be specified either via environment variable (AWS_ACCESS_KEY_ID) or system property (aws.accessKeyId)., EnvironmentVariableCredentialsProvider(): Unable to load credentials from system settings. Access key must be specified either via environment variable (AWS_ACCESS_KEY_ID) or system property (aws.accessKeyId)., WebIdentityTokenFileCredentialsProvider(): Connection pool shut down, ProfileCredentialsProvider(profileName=default, profileFile=ProfileFile(sections=[])): Profile file contained no credentials for profile 'default': ProfileFile(sections=[]), ContainerCredentialsProvider(): Cannot fetch credentials from container - neither AWS_CONTAINER_CREDENTIALS_FULL_URI or AWS_CONTAINER_CREDENTIALS_RELATIVE_URI environment variables are set., InstanceProfileCredentialsProvider(): Failed to load credentials from IMDS.]  
```

Inside the AWS EKS pod, I have confirmed that this works: 
```
export TOKEN=$(cat $AWS_WEB_IDENTITY_TOKEN_FILE)
aws sts assume-role-with-web-identity \
    --role-arn $AWS_ROLE_ARN \
    --role-session-name "nifi-debug" \
    --web-identity-token $TOKEN
```
It is true that neither AWS_CONTAINER_CREDENTIALS_FULL_URI nor AWS_CONTAINER_CREDENTIALS_RELATIVE_URI are set, but this did not seem to be necessary in Nifi 2.6.0 and earlier.

pvillard31:
Thanks for reporting this @dbuijs. This particular PR should not be the cause - the WebIdentity credentials strategy only activates when the OAuth2 Access Token Provider property is explicitly configured, which doesn't apply to your default credentials / IRSA setup.

The more likely cause is NIFI-13192 (#10460), which migrated S3 processors from AWS SDK v1 to v2 in the same release (NiFi 2.7.0). In SDK v1, the default credential chain handled IRSA via a different internal implementation. In SDK v2, the DefaultCredentialsProvider uses WebIdentityTokenFileCredentialsProvider which creates its own internal STS client with an HTTP connection pool - and that pool appears to get shut down during the processor lifecycle.

I believe this is the same problem as in https://issues.apache.org/jira/browse/NIFI-15535. I discussed with the reporter of this JIRA and the problem was solved by moving to using the EKS pod roles instead. If we want to solve the problem for IRSA, my investigation at the time made me think that we would need a custom implementation to by-pass the issue in the AWS SDK and not have this pool shut down issue.

pvillard31:
If moving away from IRSA is not an option, I can look into having an improvement to read directly from the mounted `AWS_WEB_IDENTITY_TOKEN_FILE` and have our own STS client to avoid connection pool lifecycle issues with the default AWS SDK credentials chain.

PR #10304, opened 2025-09-15
NIFI-14965: Adding the property value tooltip that helps see embedded…

mcgilman:
… parameter values.

### DIFF
--- a/nifi-frontend/src/main/frontend/libs/shared/src/components/codemirror/autocomplete/parameter-tip/parameter-tip.component.html
+++ b/nifi-frontend/src/main/frontend/libs/shared/src/components/codemirror/autocomplete/parameter-tip/parameter-tip.component.html
@@ -25,7 +25,7 @@
                 } @else if (parameter.value === '') {
                     <div class="unset neutral-color">Empty string set</div>
                 } @else {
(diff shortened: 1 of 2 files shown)


nifi-frontend/src/main/frontend/libs/shared/src/components/codemirror/autocomplete/parameter-tip/parameter-tip.component.html
```diff
@@ -19,6 +19,15 @@
     @if (data?.parameter; as parameter) {
         <div class="flex flex-col gap-y-3">
             <div class="parameter-name text-lg font-bold">{{ parameter.name }}</div>
+            @if (!parameter.sensitive) {
+                @if (parameter.value === null) {
+                    <div class="unset neutral-color">No value set</div>
+                } @else if (parameter.value === '') {
+                    <div class="unset neutral-color">Empty string set</div>
+                } @else {
+                    <div class="line-clamp-[20] leading-4 whitespace-pre font-mono text-sm">{{ parameter.value }}</div>
```
rfellows: should add `truncate` class here to be consistent.

PR #10316, opened 2025-09-17
NIFI-14984 - prevent ENTER from adding a new line in codemirror editors if the component is not dirty

rfellows:
[NIFI-14984](https://issues.apache.org/jira/browse/NIFI-14984)


mcgilman:
Reviewing...

PR #10317, opened 2025-09-17
NIFI-14985 - Removed property should be considered as an environmental change

pvillard31:
# Summary

NIFI-14985 - Removed property should be considered as an environmental change

Let's consider the scenario:

- I have a processor version X which is configured with properties A, B, C
- I have a versioned flow V that contains this processor version X
- Then I create a version Y of the processor where I now only have properties A and B and I did implement the removeProperty() call

If I import the flow when only the processor version Y is available, the versioned process group will show a local change with the property being removed. We should not show this as a local change since it's an expected modification of the flow with the call of removeProperty() (we should however check that this is not a dynamic property).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


exceptionfactory (approved):
Thanks for this improvement @pvillard31, the implementation and tests look good. +1 merging

PR #10320, opened 2025-09-17
NIFI-14988: Fixing issue with default values for dynamic properties.

mcgilman:
- In the combo editor, the form was not marked as dirty which caused the Apply button to not be enabled.
- In the nf editor, the default value was not considered.


rfellows:
reviewing...

PR #10326, opened 2025-09-19
NIFI-14985 - Fix logic to check for dynamic properties for properties being removed showing as environmental changes

pvillard31:
# Summary

NIFI-14985 - Fix logic to check for dynamic properties for properties being removed showing as environmental changes

This is a follow-up for the change in #10317 where the implemented logic was not correct.

The implemented logic is not correct when it comes to checking if we may be in the case of a dynamic property being removed. Checking the property descriptor is not the right solution because it will always be showing as dynamic: this is the default behavior of NiFi when a property is removed, NiFi will keep the property and convert it as a dynamic one by default. We should only rely on the annotations on the component to check if we're indeed facing a change where the property being removed was a static property and its removal should be ignored and not showing as a local change.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
@@ -112,12 +112,7 @@ public class FlowDifferenceFilters {
 
     }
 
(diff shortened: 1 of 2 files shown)


exceptionfactory (approved):
Thanks for correcting this issue @pvillard31, it makes sense to avoid relying on the `getPropertyDescriptor()` response as described. +1 merging

PR #10327, opened 2025-09-19
NIFI-14995 - Improve Flow Differences Filter to account for renameProperty and createControllerService

pvillard31:
# Summary

NIFI-14995 - Improve Flow Differences Filter to account for renameProperty and createControllerService

As we use more and more the `migrateProperties` capabilities, we see that we have some corner cases where the automated changes are showing as local changes which is confusing because a user is upgrading without making any changes to their versioned flows and the UI is showing that some local modifications have been made. Such scenarios should be better handled to not show local changes.

One scenario has been handled in [NIFI-14985](https://issues.apache.org/jira/browse/NIFI-14985) to better deal with removed properties.

This change is to handle two additional scenarios (this is combined in a single PR because of the approach taken that is introducing a concept of environmental context)

### Rename property with parameters

Let's consider a processor version N with

````java
    public static final PropertyDescriptor ACCESS_KEY_ID = new PropertyDescriptor.Builder()
        .name("Access Key")
        .displayName("Access Key ID")
        .required(false)
        .sensitive(true)
        .build();
````

And then version N+1 with

````java
    public static final PropertyDescriptor ACCESS_KEY_ID = new PropertyDescriptor.Builder()
        .name("Access Key ID")
        .required(false)
        .sensitive(true)
        .build();
...
    public void migrateProperties(PropertyConfiguration config) {
        config.renameProperty("Access Key", ACCESS_KEY_ID.getName());
    }
````

If the versioned flow is containing:

````json
       "groupIdentifier" : "flow-contents-group",
       "identifier" : "8de3b31b-0a70-3e01-a01c-4ed2fe3a4cbe",
       "name" : "AWS Credentials Provider",
       "properties" : {
         "Access Key" : "#{AWS Access Key ID}",
         "default-credentials" : "false",
         "Session Time" : "3600",
         "assume-role-sts-signer-override" : "Default Signature",
         "assume-role-sts-region" : "us-west-2",
         "Secret Key" : "#{AWS Secret Access Key}",
         "anonymous-credentials" : "false"
       }, 
````

Then we would show two local changes:
- Property Parameterization Removed - Property 'Access Key ID' is no longer a parameter reference
- Property Parameterized - Property 'Access Key ID' was parameterized

We should not be showing this as local changes.

### Create Controller Service

When using something like:

````java
     @Override
     public void migrateProperties(final PropertyConfiguration config) {
         if (!config.isPropertySet(ABC.getName())) {
             final String serviceId = config.createControllerService(
                     "org.apache.nifi.foo",
                     Map.of());
             config.setProperty(ABC, serviceId);
         }
         super.migrateProperties(config);
     } 
````

Then we would be showing two local changes:
- Property Added: the property ABC has been added
- Component Created: a controller service has been created

We should check that if the property being added is not a dynamic property, and is referencing the UUID of the created controller service, then both changes should not be showing as local changes.

### Approach

Introducing a richer `EnvironmentalChangeContext` that is reused across all comparisons. This is how we can link a couple of differences as related to the same change and decide to filter out the differences or not.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
@@ -76,7 +76,7 @@ public class FlowDifferenceFilters {
 
     public static boolean isEnvironmentalChange(final FlowDifference difference, final VersionedProcessGroup localGroup, final FlowManager flowManager,
                                                 final EnvironmentalChangeContext context) {
-        final EnvironmentalChangeContext evaluatedContext = context == null ? EnvironmentalChangeContext.empty() : context;
+        final EnvironmentalChangeContext evaluatedContext = Objects.requireNonNull(context, "EnvironmentalChangeContext required");
         return difference.getDifferenceType() == DifferenceType.BUNDLE_CHANGED
             || isSensitivePropertyDueToGhosting(difference, flowManager)
             || isRpgUrlChange(difference)
@@ -134,12 +134,11 @@ public class FlowDifferenceFilters {
     }
 
     private static ComponentNode getComponent(final FlowManager flowManager, final VersionedComponent component) {
-        if (!(component instanceof InstantiatedVersionedComponent)) {
+        if (component instanceof InstantiatedVersionedComponent instantiatedComponent) {
+            return getComponent(flowManager, component.getComponentType(), instantiatedComponent.getInstanceIdentifier());
+        } else {
             return null;
         }
-
-        final InstantiatedVersionedComponent instantiatedComponent = (InstantiatedVersionedComponent) component;
-        return getComponent(flowManager, component.getComponentType(), instantiatedComponent.getInstanceIdentifier());
     }
 
     private static boolean supportsDynamicProperties(final ConfigurableComponent component) {
@@ -597,21 +596,19 @@ public class FlowDifferenceFilters {
                         final PropertyDescriptor propertyDescriptor = componentNode.getPropertyDescriptor(fieldNameOptional.get());
                         if (propertyDescriptor != null && !propertyDescriptor.isDynamic() && propertyDescriptor.getControllerServiceDefinition() != null) {
                             final Object valueB = difference.getValueB();
-                            if (valueB instanceof String) {
-                                serviceIdsReferencedByNewProperties.add((String) valueB);
+                            if (valueB instanceof String serviceId) {
+                                serviceIdsReferencedByNewProperties.add(serviceId);
                             }
                         }
                     }
                 }
-            }
-
-            if (difference.getDifferenceType() == DifferenceType.PROPERTY_PARAMETERIZED
+            } else if (difference.getDifferenceType() == DifferenceType.PROPERTY_PARAMETERIZED
                     || difference.getDifferenceType() == DifferenceType.PROPERTY_PARAMETERIZATION_REMOVED) {
 
                 final Optional<String> propertyNameOptional = difference.getFieldName();
                 final Optional<String> componentIdOptional = getComponentInstanceIdentifier(difference);
 
-                if (!propertyNameOptional.isPresent() || !componentIdOptional.isPresent()) {
+                if (propertyNameOptional.isEmpty() || componentIdOptional.isEmpty()) {
                     continue;
                 }
 
@@ -630,14 +627,16 @@ public class FlowDifferenceFilters {
             }
         }
 
-        Set<String> serviceIdsWithMatchingAdditions = Collections.emptySet();
+        final Set<String> serviceIdsWithMatchingAdditions;
         if (!serviceIdsReferencedByNewProperties.isEmpty()) {
             serviceIdsWithMatchingAdditions = differences.stream()
                     .filter(diff -> diff.getDifferenceType() == DifferenceType.COMPONENT_ADDED)
                     .map(FlowDifferenceFilters::extractControllerServiceIdentifier)
                     .filter(Objects::nonNull)
                     .filter(serviceIdsReferencedByNewProperties::contains)
-                    .collect(Collectors.toCollection(HashSet::new));
+                    .collect(Collectors.toSet());
+        } else {
+            serviceIdsWithMatchingAdditions = Collections.emptySet();
         }
 
         final Set<FlowDifference> parameterizedPropertyRenameDifferences = new HashSet<>();
@@ -677,15 +676,19 @@ public class FlowDifferenceFilters {
             }
         }
 
+        final EnvironmentalChangeContext environmentalChangeContext;
         if (serviceIdsWithMatchingAdditions.isEmpty() && parameterizedPropertyRenameDifferences.isEmpty()) {
-            return EnvironmentalChangeContext.empty();
+            environmentalChangeContext = EnvironmentalChangeContext.empty();
+        } else {
+            environmentalChangeContext = new EnvironmentalChangeContext(serviceIdsWithMatchingAdditions, parameterizedPropertyRenameDifferences);
         }
 
-        return new EnvironmentalChangeContext(serviceIdsWithMatchingAdditions, parameterizedPropertyRenameDifferences);
+        return environmentalChangeContext;
     }
 
     public static boolean isControllerServiceCreatedForNewProperty(final FlowDifference difference, final EnvironmentalChangeContext context) {
-        return isControllerServiceCreatedForNewPropertyInternal(difference, context == null ? EnvironmentalChangeContext.empty() : context);
+        final EnvironmentalChangeContext evaluatedContext = Objects.requireNonNull(context, "EnvironmentalChangeContext required");
+        return isControllerServiceCreatedForNewPropertyInternal(difference, evaluatedContext);
     }
 
     private static boolean isControllerServiceCreatedForNewPropertyInternal(final FlowDifference difference, final EnvironmentalChangeContext context) {
@@ -695,9 +698,7 @@ public class FlowDifferenceFilters {
 
         if (difference.getDifferenceType() == DifferenceType.PROPERTY_ADDED) {
             final Object valueB = difference.getValueB();
-            if (valueB instanceof String) {
-                return context.serviceIdsCreatedForNewProperties().contains(valueB);
-            }
+            return valueB != null && context.serviceIdsCreatedForNewProperties().contains(valueB);
         }
 
         if (difference.getDifferenceType() == DifferenceType.COMPONENT_ADDED) {
@@ -734,7 +735,8 @@ public class FlowDifferenceFilters {
     }
 
     public static boolean isPropertyParameterizationRename(final FlowDifference difference, final EnvironmentalChangeContext context) {
-        return (context == null || context.parameterizedPropertyRenames().isEmpty()) ? false : context.parameterizedPropertyRenames().contains(difference);
+        final EnvironmentalChangeContext evaluatedContext = Objects.requireNonNull(context, "EnvironmentalChangeContext required");
+        return evaluatedContext.parameterizedPropertyRenames().isEmpty() ? false : evaluatedContext.parameterizedPropertyRenames().contains(difference);
     }
 
     private static Optional<String> getComponentInstanceIdentifier(final FlowDifference difference) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java [hunk 1]
```diff
@@ -58,6 +71,12 @@ public class FlowDifferenceFilters {
      * @return <code>true</code> if the change is an environment-specific change, <code>false</code> otherwise
      */
     public static boolean isEnvironmentalChange(final FlowDifference difference, final VersionedProcessGroup localGroup, final FlowManager flowManager) {
+        return isEnvironmentalChange(difference, localGroup, flowManager, EnvironmentalChangeContext.empty());
+    }
+
+    public static boolean isEnvironmentalChange(final FlowDifference difference, final VersionedProcessGroup localGroup, final FlowManager flowManager,
+                                                final EnvironmentalChangeContext context) {
+        final EnvironmentalChangeContext evaluatedContext = context == null ? EnvironmentalChangeContext.empty() : context;
```
exceptionfactory: I recommend requiring the `EnvironmentalChangeContext` for this method, instead of substituting an `empty()` context.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -112,6 +133,15 @@ private static ComponentNode getComponent(final FlowManager flowManager, final C
 
     }
 
+    private static ComponentNode getComponent(final FlowManager flowManager, final VersionedComponent component) {
+        if (!(component instanceof InstantiatedVersionedComponent)) {
+            return null;
+        }
+
+        final InstantiatedVersionedComponent instantiatedComponent = (InstantiatedVersionedComponent) component;
+        return getComponent(flowManager, component.getComponentType(), instantiatedComponent.getInstanceIdentifier());
```
exceptionfactory: Minor implementation note, it looks like this could be adjusted to use `instanceof InstantiatedVersionedComponent instantiated) { ...` and otherwise return `null`, instead of the short-circuit return and casting.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+                final ComponentNode componentNode = Optional.ofNullable(getComponent(flowManager, difference.getComponentB()))
+                        .orElseGet(() -> getComponent(flowManager, difference.getComponentA()));
+                if (componentNode != null) {
+                    final Optional<String> fieldNameOptional = difference.getFieldName();
+                    if (fieldNameOptional.isPresent()) {
+                        final PropertyDescriptor propertyDescriptor = componentNode.getPropertyDescriptor(fieldNameOptional.get());
+                        if (propertyDescriptor != null && !propertyDescriptor.isDynamic() && propertyDescriptor.getControllerServiceDefinition() != null) {
+                            final Object valueB = difference.getValueB();
+                            if (valueB instanceof String) {
+                                serviceIdsReferencedByNewProperties.add((String) valueB);
```
exceptionfactory: Minor adjustment to remove the need for casting:
```suggestion
                            if (valueB instanceof String serviceId) {
                                serviceIdsReferencedByNewProperties.add(serviceId);
```

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (difference.getDifferenceType() == DifferenceType.PROPERTY_PARAMETERIZED
```
exceptionfactory: It looks like this should be an `else if` instead of a standalone `if` statement.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+                    || difference.getDifferenceType() == DifferenceType.PROPERTY_PARAMETERIZATION_REMOVED) {
+
+                final Optional<String> propertyNameOptional = difference.getFieldName();
+                final Optional<String> componentIdOptional = getComponentInstanceIdentifier(difference);
+
+                if (!propertyNameOptional.isPresent() || !componentIdOptional.isPresent()) {
```
exceptionfactory: The negated `isPresent()` can be changed to `isEmpty()`

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+                    parameterizedAddsByComponent.computeIfAbsent(componentId, key -> new ArrayList<>()).add(diffInfo);
+                } else {
+                    parameterizationRemovalsByComponent.computeIfAbsent(componentId, key -> new ArrayList<>()).add(diffInfo);
+                }
+            }
+        }
+
+        Set<String> serviceIdsWithMatchingAdditions = Collections.emptySet();
+        if (!serviceIdsReferencedByNewProperties.isEmpty()) {
+            serviceIdsWithMatchingAdditions = differences.stream()
```
exceptionfactory: I recommend declaring the Set `final` and setting it in an `else` block.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+                    .filter(diff -> diff.getDifferenceType() == DifferenceType.COMPONENT_ADDED)
+                    .map(FlowDifferenceFilters::extractControllerServiceIdentifier)
+                    .filter(Objects::nonNull)
+                    .filter(serviceIdsReferencedByNewProperties::contains)
+                    .collect(Collectors.toCollection(HashSet::new));
```
exceptionfactory: Can `toSet()` be used instead of a new HashSet?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+                    parameterizedPropertyRenameDifferences.add(matchingAddition.difference());
+                }
+            }
+        }
+
+        if (serviceIdsWithMatchingAdditions.isEmpty() && parameterizedPropertyRenameDifferences.isEmpty()) {
+            return EnvironmentalChangeContext.empty();
+        }
+
+        return new EnvironmentalChangeContext(serviceIdsWithMatchingAdditions, parameterizedPropertyRenameDifferences);
```
exceptionfactory: Minor, but since these are close to the end of the method, I recommend declaring a `final EnvironmentChangeContext` and setting the value as opposed to having multiple returns.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+    }
+
+    public static boolean isControllerServiceCreatedForNewProperty(final FlowDifference difference, final EnvironmentalChangeContext context) {
```
exceptionfactory, line 689: Can `context` be null? Similar to other public methods, I recommend using `Objects.requireNonNull()`

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+    private static boolean isControllerServiceCreatedForNewPropertyInternal(final FlowDifference difference, final EnvironmentalChangeContext context) {
+        if (context.serviceIdsCreatedForNewProperties().isEmpty()) {
+            return false;
+        }
+
+        if (difference.getDifferenceType() == DifferenceType.PROPERTY_ADDED) {
+            final Object valueB = difference.getValueB();
+            if (valueB instanceof String) {
+                return context.serviceIdsCreatedForNewProperties().contains(valueB);
+            }
```
exceptionfactory: The `instanceof` check seems unnecessary since it would be handled by `contains`, unless `valueB` can be `null`.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+
+        if (component instanceof VersionedControllerService) {
+            return component.getIdentifier();
+        }
+
+        return null;
+    }
+
+    public static boolean isPropertyParameterizationRename(final FlowDifference difference, final EnvironmentalChangeContext context) {
+        return (context == null || context.parameterizedPropertyRenames().isEmpty()) ? false : context.parameterizedPropertyRenames().contains(difference);
```
exceptionfactory: See note on `null` possibility for `context`.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+        return Optional.ofNullable(component.getIdentifier());
+    }
+
+    private static Optional<String> getParameterReferenceValue(final FlowDifference difference, final boolean fromComponentA) {
+        final VersionedComponent primaryComponent = fromComponentA ? difference.getComponentA() : difference.getComponentB();
+        final VersionedComponent secondaryComponent = fromComponentA ? difference.getComponentB() : difference.getComponentA();
+
+        final Map<String, String> primaryProperties = getProperties(primaryComponent);
+        if (primaryProperties.isEmpty()) {
+            return Optional.empty();
```
exceptionfactory: This method is a bit longer, making the multiple return statements a bit more complex. What do you think about refactoring to a single return?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -547,4 +577,302 @@ private static boolean isParameterContextChange(final FlowDifference flowDiffere
...
+            if (!secondaryProperties.containsKey(propertyName)) {
+                return Optional.of(propertyValue);
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private static Map<String, String> getProperties(final VersionedComponent component) {
+        if (component == null) {
```
exceptionfactory: Recommend refactoring this method to a single return an use of `if ... else if ...`

exceptionfactory (commented):
Thanks for working on these improvements @pvillard31, this is very helpful when it comes to distinguishing actual changes from changes related to property migration.

I noted a handful of small implementation recommendations, but the basic approach looks good.

pvillard31:
Thanks for the review @exceptionfactory - I pushed a commit to address your comments

PR #10332, opened 2025-09-22
docs: fix small typos & formatting in README

Username46786:
“SAML 2” → “SAML 2.0” (matches admin guide terminology) 
Corrected grep example: nifi-app*.log and quoted pattern
Added missing comma in license sentence
Docs-only changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ for thousands of companies worldwide across every industry.
   - Support for Processors in native Python
   - REST API for orchestration and monitoring
 - Secure Configuration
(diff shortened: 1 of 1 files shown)


exceptionfactory (changes requested):
Thanks for suggesting these changes @Username46786, however, they are not necessary, and in the case of the `grep` command, the change is not desirable. All pull requests also need an associated Apache NiFi Jira issue for tracking. In light of these details, I am closing this pull request.

PR #10334, opened 2025-09-22
NIFI-15001 - Fix PutElasticsearchRecord with recursive reads

pvillard31:
# Summary

NIFI-15001 - Fix `PutElasticsearchRecord` with recursive reads

When `PutElasticsearchRecord` processed FlowFiles with batch sizes smaller than the record count, it re-read the FlowFile to build error batches. With recursive reads disabled (the default), the second `session.read()` triggered `IllegalStateException: FlowFile … already in use`, so entire FlowFiles were routed to failure.

The change keeps processing single-pass: every record is cloned exactly once when read from the RecordReader. The mutated copy is used for the Elasticsearch `_bulk` request, while an untouched clone is retained for provenance and error routing. This eliminates the recursive read and keeps RecordPath mutations from leaking into error output.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecord.java
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecord.java
@@ -61,6 +61,7 @@ import org.apache.nifi.serialization.RecordReaderFactory;
 import org.apache.nifi.serialization.RecordSetWriter;
 import org.apache.nifi.serialization.RecordSetWriterFactory;
 import org.apache.nifi.serialization.record.DataType;
+import org.apache.nifi.serialization.record.MapRecord;
 import org.apache.nifi.serialization.record.PushBackRecordSet;
 import org.apache.nifi.serialization.record.Record;
 import org.apache.nifi.serialization.record.RecordField;
@@ -79,6 +80,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecord.java
```diff
@@ -773,6 +770,47 @@ private Object coerceStringToLong(final String fieldName, final String stringVal
...
+
+    private Map<String, Object> cloneValues(final Map<String, Object> values) {
+        final Map<String, Object> cloned = new LinkedHashMap<>(values.size());
+        for (final Map.Entry<String, Object> entry : values.entrySet()) {
+            cloned.put(entry.getKey(), cloneValue(entry.getValue()));
+        }
+        return cloned;
+    }
+
+    private Object cloneValue(final Object value) {
```
dan-s1, line 786: Instead of creating all these clone methods, it would be cleaner to use something like Apache Commons `SerializationUtils#clone`. The only thing you would have to make sure though is all of the data to be cloned implements the Serializable interface.

pvillard31: In our case the records are `MapRecord` instances, and `MapRecord` (and other `Record` implementations) do not implement `Serializable`. Besides, Java serialization would add noticeable overhead on a hot path that can process large batches, whereas the current approach only walk the map/collection structures we actually store.

dan-s1: Thanks for clarifying that. Could this cloning be reused? In other words would it make sense to have a clone/deepClone method in the  `Record` interface and implemented in the implementations?

pvillard31: I guess it could, but I'm not sure this is really a pattern we would want to encourage. The scenario here is quite specific. Others may have a different opinion though.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchRecord.java
```diff
@@ -420,22 +423,28 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-            final List<Record> originals = new ArrayList<>();
+            final List<Record> processedRecords = new ArrayList<>();
+            final List<Record> originalRecords = new ArrayList<>();
 
             Record record;
             while ((record = recordSet.next()) != null) {
-                addOperation(operationList, record, indexOperationParameters, indices, types);
-                originals.add(record);
+                final Record originalRecord = cloneRecord(record);
+                final Record processedRecord = cloneRecord(record);
```
exceptionfactory, line 434: Is it necessary to make two clones of the original `record`?

pvillard31: Yes it is. `recordSet.next()` reuses the same Record instance each time, so we take one deep copy before adding it to `processedRecords`. It lets us mutate fields (for example through the various `fieldValue.updateValue(...)` calls) without those changes leaking into later iterations. We take a second deep copy for `originalRecords` so that, when `indexDocuments()` builds FlowFiles for failures, it can write out an untouched version of the input rather than the mutated version or re-read from the original FlowFile.

exceptionfactory: Thanks for describing the background, that makes sense.

PR #10343, opened 2025-09-24
NIFI-14949 - Display bulletin stack trace on bulletin board if it exi…

rfellows:
…sts. Add the stack trace to the copied content if available.

# Summary

[NIFI-14949](https://issues.apache.org/jira/browse/NIFI-14949)

* If a bulletin has a stack trace, it will be offered to view through an expandable link on the bulletin board.

* Changed the bulletin copy mechanism slightly: the copy icon was moved to the end of the bulletin description in both the bulletin board and the bulletin tool-tip component. In both cases, if a stack trace is present it will be copied to the clipboard as well, appended after the message.

Screenshot 2025-09-24 at 12 10 15
Screenshot 2025-09-24 at 12 10 31


scottyaslan:
Reviewing...

PR #10344, opened 2025-09-24
NIFI-14999: Adding support to stop a starting processor.

mcgilman:
## Summary
This PR adds support for exposing the **physical state** of processors in the NiFi API, enabling the UI to allow stopping processors that are in the `STARTING` physical state.

## Problem Statement
Previously, when a processor was starting up (executing `@OnScheduled` methods), its logical state was `STOPPED` but the physical state was `STARTING`. The UI had no visibility into this physical state. Users couldn't stop processors that were stuck in the `STARTING` state, requiring process group-level stops as a workaround

## Solution Overview
Added `physicalState` field to processor DTOs and entities, with full support for:
- API exposure with proper schema documentation
- Cluster response merging with transition state prioritization  
- Backend validation allowing stops of `STARTING` processors
- Frontend updates to allow the user to stop a STARTING processor

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardProcessorDAO.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardProcessorDAO.java
@@ -413,7 +413,15 @@ public class StandardProcessorDAO extends ComponentDAO implements ProcessorDAO {
                         case DISABLED:
                             processor.verifyCanDisable();
                             break;
+                        case STARTING:
+                        case STOPPING:
+                            // These are internal transition states and should not be set directly via API
+                            throw new IllegalArgumentException("Cannot set processor state to " + purposedScheduledState);
                     }
+                } else if (purposedScheduledState == ScheduledState.STOPPED && processor.getPhysicalScheduledState() == ScheduledState.STARTING) {
+                    // Handle special case: verify stopping a processor that's physically starting
+                    processor.getProcessGroup().verifyCanScheduleComponentsIndividually();
+                    processor.verifyCanStop();
                 }
             } catch (IllegalArgumentException iae) {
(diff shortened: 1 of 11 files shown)


rfellows (approved):
UI changes look good 👍

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardProcessorDAO.java
```diff
@@ -552,6 +564,15 @@ public ProcessorNode updateProcessor(ProcessorDTO processorDTO) {
                 } catch (Exception e) {
                     throw new NiFiCoreException("Unable to update processor run state: " + e, e);
                 }
+            } else if  (purposedScheduledState == ScheduledState.STOPPED && processor.getPhysicalScheduledState() == ScheduledState.STARTING) {
+                // Handle special case: allow stopping a processor that's physically starting
+                try {
+                    parentGroup.stopProcessor(processor);
+                } catch (NullPointerException npe) {
+                    throw new NiFiCoreException("Unable to stop starting processor.", npe);
```
sfc-gh-mpayne: We should not be catching NullPointers

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardProcessorDAO.java
```diff
@@ -552,6 +564,15 @@ public ProcessorNode updateProcessor(ProcessorDTO processorDTO) {
...
+                } catch (Exception e) {
+                    throw new NiFiCoreException("Unable to stop starting processor: " + e, e);
```
sfc-gh-mpayne: The Exception in this case should probably include the `toString()` of the Processor in question.

sfc-gh-mpayne (approved):

PR #10345, opened 2025-09-24
NIFI-15014 allow advanced UpdateAttribute Rules to be searched even i…

markobean:
…f missing Comments field such as when a processor was upgraded from prior to Comments being available

# Summary

In certain cases, the Search feature of Rules in the advanced UpdateAttribute does not work. The cause was determined to be a missing Comments field; not an empty field with no value, but <comments> were missing from the definition in the flow.json.gz. Likely, this occurs only when the processor has been upgraded from a version prior to when Rule Comments were available. 

The fix ensures the Comments field is present before attempting to search in its value. Arguably, the check could be performed on other fields being searched. But in reality all other fields have always been present and should never be missing. Based on review/feedback, I would not be opposed to addressing additional fields if it was deemed appropriate.

[NIFI-15014](https://issues.apache.org/jira/browse/NIFI-15014)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

A thorough verification would mean editing a flow definition or flow.json.gz to remove the <comments> field entirely from an UpdateAttribute processor. The search should still function even after doing so. 

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/update-attribute/src/app/pages/update-attribute/ui/rule-listing/rule-listing.component.ts
@@ -163,7 +163,7 @@ export class RuleListing {
         return (
             rule.id.toLowerCase().includes(filterText) ||
             rule.name.toLowerCase().includes(filterText) ||
-            rule.comments.toLowerCase().includes(filterText) ||
+            (rule.comments ?? '').toLowerCase().includes(filterText) ||
             rule.conditions.some((condition) => this.conditionMatches(condition, filterText)) ||
             rule.actions.some((action) => this.actionMatches(action, filterText))
         );


markobean:
@pvillard31 This issue is not directly related to the UI; I'm not sure the label is appropriate. While the search functionality is accessed through the UI, the only update is in the backend code serving the search results.

pvillard31:
Yeah fair point, given this is typescript, I assumed that the UI folks are more familiar with this code

rfellows:
@markobean I'm confused by this statement
>...the only update is in the backend code serving the search results

This change is most definitely a UI change and NOT a backend change. It just "massages" the list of rules it already has loaded from the backend.

Is it your intent to change this on the backend? If so, we should close this PR and address this issue elsewhere (maybe in `org.apache.nifi.update.attributes.dto.DtoFactory`). If not, this does fix the issue described and I'd be fine to merge it. Just want to make sure this is the solution you intended?

markobean:
@rfellows I misspoke when I said backend. What I really meant was that it doesn't change any user-facing UI elements, e.g. adding a new button or rearranging layout, etc. My intent was to change the behavior in the UI as per the PR. No backend changes are intended or needed.

PR #10346, opened 2025-09-24
NIFI-14320 buckets listing and management, (manage bucket policies is…

scottyaslan:
# Summary

Buckets listing, filtering, create, edit, delete, Also fixed favicon.

[NIFI-14320](https://issues.apache.org/jira/browse/NIFI-14320)
Screenshot 2025-09-24 at 4 24 58 PM
Screenshot 2025-09-24 at 4 24 40 PM
Screenshot 2025-09-24 at 4 25 10 PM
Screenshot 2025-09-24 at 4 25 33 PM
Screenshot 2025-09-24 at 4 25 42 PM
Screenshot 2025-09-24 at 4 25 55 PM
Screenshot 2025-09-24 at 4 26 14 PM

# Tracking

Please complete the following tracking steps prior to pull request creation.

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI) issue created

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-14320`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-14320`

### Pull Request Formatting

- [x] Pull Request based on current revision of the `main` branch
- [x] Pull Request refers to a feature branch with one commit containing changes

# Verification

Please indicate the verification steps performed prior to pull request creation.

### Build

- [x] Build completed using `./mvnw clean install -P contrib-check`
  - [x] JDK 21
  - [x] JDK 25

### Licensing

- [x] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [x] New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [x] Documentation formatting appears as expected in rendered files

PR #10347, opened 2025-09-24
NIFI-15004 nifi registry about dialog

scottyaslan:
# Summary

[NIFI-15004](https://issues.apache.org/jira/browse/NIFI-15004)

Screenshot 2025-09-24 at 5 37 01 PM

# Tracking

### Issue Tracking

### Pull Request Tracking

- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-15004`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-15004`

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [x] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/service/about.service.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/service/about.service.ts
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { inject, Injectable } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+import { Observable } from 'rxjs';
+import { LoadAboutResponse } from '../state/about';
+
+@Injectable({ providedIn: 'root' })
+export class AboutService {
+    private httpClient = inject(HttpClient);
+
+    private static readonly API: string = '../nifi-registry-api';
+
+    getAbout(): Observable<LoadAboutResponse> {
+        // const mockError: HttpErrorResponse = new HttpErrorResponse({
+        //     status: 404,
+        //     statusText: 'Bad Gateway',
+        //     url: `${AboutService.API}/buckets`,
+        //     error: {
+        //         message: 'Mock error: unable to GET about information.',
+        //         timestamp: new Date().toISOString()
+        //     }
+        // });
+        // return throwError(() => mockError);
+
+        return this.httpClient.get<LoadAboutResponse>(`${AboutService.API}/about`);
+    }
+}
--- a/nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/ui/common/about-dialog/about-dialog.component.html
+++ b/nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/ui/common/about-dialog/about-dialog.component.html
@@ -18,7 +18,6 @@
 <div class="about-dialog">
     <h2 mat-dialog-title>About NiFi Registry</h2>
     <mat-dialog-content>
-        <context-error-banner [context]="ErrorContextKey.ABOUT"></context-error-banner>
         <div class="flex flex-col justify-between gap-y-5">
             @if (about$ | async; as about) {
                 <div class="flex justify-center">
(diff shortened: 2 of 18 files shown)


sardell:
Reviewing now...

nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/service/about.service.ts
```diff
@@ -0,0 +1,43 @@
...
+import { LoadAboutResponse } from '../state/about';
+
+@Injectable({ providedIn: 'root' })
+export class AboutService {
+    private httpClient = inject(HttpClient);
+
+    private static readonly API: string = '../nifi-registry-api';
+
+    getAbout(): Observable<LoadAboutResponse> {
+        // const mockError: HttpErrorResponse = new HttpErrorResponse({
```
sardell, line 30: I mentioned this in your buckets PR, but I'm not sure of the value of keeping this commented code in each service.  I think this code snippet could be useful at some point when adding new features or debugging, but instead of leaving commented code inside the codebase itself I feel it would be better to leave it in some sort of README in the service directory or in some local development documentation.

sardell (commented):
I left one comment around some commented code in the service, but otherwise this looks great!

nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/service/about.service.ts
sardell, line 30: Same comment I left on your other PRs regarding leaving this snippet in the codebase as a comment.[1][2]

[1] https://github.com/apache/nifi/pull/10346/files#r2379890984
[2] https://github.com/apache/nifi/pull/10347#discussion_r2380262296

scottyaslan: I think we do want to keep these here. They are here to make it easy to test all the error handling scenarios. We don't want to leave the error scenario turned on for the dev server as we want the application to behave normally but we want to be able to quickly test error scenarios.

scottyaslan: I think we do want to keep these here. They are here to make it easy to test all the error handling scenarios. We don't want to leave the error scenario turned on for the dev server as we want the application to behave normally but we want to be able to quickly test error scenarios.

sardell: I see the usefulness of a snippet like this, I just think it should live outside of the actual code in a README to be copy/pasted when it's temporarily needed for testing. However, this is not a blocker issue for me and I'm fine if you feel strongly about leaving these snippets in the service.

scottyaslan: Thanks.

nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/ui/common/about-dialog/about-dialog.component.ts
```diff
@@ -0,0 +1,59 @@
...
+        NgOptimizedImage,
+        ContextErrorBanner
+    ]
+})
+export class AboutDialogComponent extends CloseOnEscapeDialog {
+    private store = inject<Store<AboutState>>(Store);
+
+    constructor() {
+        super();
+        this.store.dispatch(loadAbout());
```
rfellows: Its inconsistent with how the nifi about dialog works. load about is not loaded as part of the dialog open, rather it is loaded as part of a higher-level component (navigation, content viewer, ...)

also, the about in those cases doesn't show an error banner in the dialog. Pretty sure it is a snackbar or global error.

nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/ui/common/about-dialog/about-dialog.component.html
```diff
@@ -0,0 +1,47 @@
...
+    <mat-dialog-content>
+        <context-error-banner [context]="ErrorContextKey.ABOUT"></context-error-banner>
+        <div class="flex flex-col justify-between gap-y-5">
+            @if (about$ | async; as about) {
+                <div class="flex justify-center">
+                    <img
+                        ngSrc="assets/icons/registry-logo-web-app.svg"
+                        priority
+                        width="300"
+                        height="128"
```
rfellows: Weird dev tools console warning about this size: 
```
common.mjs:1775 NG02952: The NgOptimizedImage directive (activated on an <img> element with the `ngSrc="assets/icons/registry-logo-web-app.svg"`) has detected that the aspect ratio of the image does not match the aspect ratio indicated by the width and height attributes. 
Intrinsic image size: 117w x 150h (aspect-ratio: 0.78). 
Supplied width and height attributes: 300w x 128h (aspect-ratio: 2.34). 
To fix this, update the width and height attributes.
```

scottyaslan: fixed

nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/ui/common/about-dialog/about-dialog.component.ts
scottyaslan: In the current nifi registry application the about dialog calls GET on the about url each time the dialog is opened:
https://github.com/apache/nifi/blob/main/nifi-registry/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js#L1113

I tried to follow that approach since that is how it worked previously. I can appreciate that we are changing some things so I am open to discussing when and where this call should be happening. I was simply trying to replace what exists.

To that point, I realize we have changed some things. Also to that point, the legacy nifi registry UI did not have any error handling for the GET call to the about url so no snackbar or banner error was ever surfaced to the user. If we want to follow what nifi has done I can.

rfellows (approved):

PR #10355, opened 2025-09-28
NIFI-15025 : LookupFailureException with large HTTP responses due to BufferedInputStream mark/reset limitation

ravinarayansingh:
…st mark/reset support, add related tests.

# Summary

[NIFI-15025](https://issues.apache.org/jira/browse/NIFI-15025)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


ravinarayansingh:
> Thanks for proposing this improvement @ravinarayansingh.
> 
> On initial review, I'm concerned about introducing a custom `InputStream` implementation for this particular use case.
> 
> The `RestLookupService` is not designed for handling very large responses in general, although accommodating something larger seems reasonable. Instead of adding a custom `InputStream` with associated temporary file management, what do you think about simply using a larger buffer size for the `BufferedInputStream`? I agree that introducing a property is probably unnecessary, but increasing the buffer size to something like 1 MB or 8 MB seems like it would provide a reasonable size to cover more use cases, without introducing more complexity.

Hi @exceptionfactory 
Agreed, an 8 MB buffer size sounds like the right balance. I’ll go ahead and make the change accordingly

exceptionfactory:
> Hi @exceptionfactory Agreed, an 8 MB buffer size sounds like the right balance. I’ll go ahead and make the change accordingly

Thanks @ravinarayansingh, that sounds good.

exceptionfactory (approved):
Thanks for adjusting the approach and adding the test @ravinarayansingh, this looks good! +1 merging

PR #10356, opened 2025-09-29
NIFI-15016 - Ensure nested versioned groups detect scheduled-state changes

pvillard31:
# Summary

NIFI-15016 - Ensure nested versioned groups detect scheduled-state changes

The versioned flow stores the same flow coordinates on every embedded process group in a versioned flow. When we do flow design changes or configuration edits, we clear those coordinates and the old comparator logic would see this as a condition to check the changes in a recursive manner (deep comparison). When we do operational changes like starting/stopping, enabling/disabling, we do not clear the coordinates and this is why we were missing enabling/disabling a component as a local change in nested process groups. Clearing coordinates when scheduling changes occur would falsely mark the group as referencing a different flow version, contradicting the runtime-versus-definition. Consequently, I have updated the recursion condition to inspect any non-empty child group. It preserves the coordinate contract while surfacing schedule state changes exactly like other edits.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/test/java/org/apache/nifi/registry/flow/diff/TestStandardFlowComparator.java
```diff
@@ -285,6 +286,60 @@ public void testDeepStrategyWithChildPGs() {
...
+        registryProcessor.setProperties(Collections.emptyMap());
+        registryProcessor.setPropertyDescriptors(Collections.emptyMap());
+        registryNested.getProcessors().add(registryProcessor);
+
+        final VersionedProcessor localProcessor = new VersionedProcessor();
+        localProcessor.setIdentifier("processorZ");
+        localProcessor.setScheduledState(ScheduledState.DISABLED);
+        localProcessor.setProperties(Collections.emptyMap());
+        localProcessor.setPropertyDescriptors(Collections.emptyMap());
+        localNested.getProcessors().add(localProcessor);
```
exceptionfactory, line 322: It would be helpful to declare static values, or at least method-local variables for the various component identifiers.

exceptionfactory (commented):
Thanks for this improvement @pvillard31. The change makes sense, and the new test is helpful. I noted a minor suggestion to declare and reuse string identifiers for the test method, but otherwise this looks good.

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/test/java/org/apache/nifi/registry/flow/diff/TestStandardFlowComparator.java
pvillard31: Thanks @exceptionfactory , done

PR #10366, opened 2025-10-02
NIFI-15027 - adjust AvroWriter handling of invalid payloads; ConsumeKafka impact

greyp9:
# Summary

[NIFI-15027](https://issues.apache.org/jira/browse/NIFI-15027)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [x] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/avro/WriteAvroResultWithSchema.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/avro/WriteAvroResultWithSchema.java
@@ -61,7 +61,7 @@ public class WriteAvroResultWithSchema extends AbstractRecordSetWriter {
         try {
             dataFileWriter.append(rec);
         } catch (final DataFileWriter.AppendWriteException e) {
-            throw new IOException(e);
+            throw new IOException("AppendWriteException while writing a datum to the Avro record buffer", e);
         }
         return Collections.emptyMap();
     }


nifi-extension-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/avro/WriteAvroResultWithSchema.java
```diff
@@ -58,7 +58,11 @@ public void flush() throws IOException {
     @Override
     public Map<String, String> writeRecord(final Record record) throws IOException {
         final GenericRecord rec = AvroTypeUtil.createAvroRecord(record, schema);
-        dataFileWriter.append(rec);
+        try {
+            dataFileWriter.append(rec);
+        } catch (final DataFileWriter.AppendWriteException e) {
+            throw new IOException(e);
```
jrsteinebrey: Thanks for working on this ticket. 
This changed line breaks other writeRecord() callers who explicitly catch DataFileWriter.AppendWriteException like this example
https://github.com/jrsteinebrey/nifi/blob/b0f29ef94e95be8160ec2cd5fbdfbef373451f90/nifi-extension-bundles/nifi-extension-utils/nifi-database-utils/src/main/java/org/apache/nifi/util/db/JdbcCommon.java#L466
They would need to be changed to catch IOException instead of AppendWriteException. 
Instead of this change here in WriteAvroResultWithSchema.java,
I suggest that you consider changing the Kafka code here 
https://github.com/apache/nifi/blob/1457950040d0fe86ade53770def6c5a95b6f0252/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/convert/AbstractRecordStreamKafkaMessageConverter.java#L112-L120\
to catch (Exception) instead of specific exception classes. Then the ticket is resolved and any future created  exception classes also route to failure.

greyp9: That's reasonable; thanks.

I'm not familiar with the reason for the "catch all" in `AbstractRecordStreamKafkaMessageConverter`.

To me, the problem seems to be that the Avro writer implementation throws a particular exception (class) that is not visible in the classpath of the Kafka implementation.  So we can't act based on that particular exception.

Another variation would be for AvroWriter to throw `MalformedRecordException` instead of `IOException`, as that better conveys the particular problem (bad data).

There are potential side effects to either of these potential paths forward; hopefully others in the community will chime in.

pvillard31: I think I'd also go with the change only in the Kafka class where we would catch all exceptions to route to parse failure. Thoughts @exceptionfactory @markap14 ?

exceptionfactory: Reviewing the call structure, I favor the proposed approach that catches the `AppendWriteException` and throws something more specific. Wrapping it and throwing an `IOException` seems appropriate based on the description of `AppendWriteException`, although I would add a message to the `IOException`.

For broader context, the `JdbcCommon` handling of `dataWriter.append()` is not directly related, and in that case, catching `AppendWriteException` only serves to allow for more specific exception messaging.

The contract of `RecordReaderFactory.createRecordReader()` defines the three checked exceptions, which the `KafkaMessageConverter` handles as parse failures. Any other exceptions propagate to `ConsumeKafka.onTrigger()`, where the transaction is rolled back. For this reason, catching a general `Exception` as a parse failure could mask other issues that indicate a programming bug, versus a problem with the record or schema.

greyp9: Thanks @jrsteinebrey @pvillard31 @exceptionfactory for your input!

I propose making this update to the changeset:

> Another variation would be for AvroWriter to throw MalformedRecordException instead of IOException, as that better conveys the particular problem (bad data).

Does that work for everyone?

exceptionfactory: Although `MalformedRecordException` is the most precise, it does not align with the `writeRecord` method signature, since `MalformedRecordException` extends the base `Exception` class.

jrsteinebrey: Non-binding: I am good with IOException being thrown here like @exceptionfactory recommended.

greyp9: Sure, that makes sense.  I will leave things as is.

exceptionfactory: Thanks, with that determined, the only other change I recommend is including a message for the IOException, such as `Failed to write Avro Record`.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/ConsumeKafkaRecordWithNullIT.java
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Demonstrate handling of incoming Avro payloads with null fields where the associated Avro schema specifies that
+ * the field is required.  Invalid payloads should be routed to PARSE_FAILURE, as stated in the documentation.
+ */
+class ConsumeKafkaRecordWithNullIT extends AbstractConsumeKafkaIT {
+    private static final int FIRST_PARTITION = 0;
+
+    private static final String RESOURCE_AVRO_SCHEMA_NULLABLE = "src/test/resources/org/apache/nifi/kafka/reader/schemaNullable.avsc.json";
+    private static final String RESOURCE_AVRO_SCHEMA_REQUIRED = "src/test/resources/org/apache/nifi/kafka/reader/schemaRequired.avsc.json";
```
exceptionfactory: It might be useful to make these JSON files multiline strings within the test class, but I will defer to the current implementation if you prefer to leave it as is.

greyp9: Thanks; my opinion is that embedded multiline resources (especially JSON) can be harder to read when the needed escapes are present.  So, I'd like to retain the current implementation of those.

exceptionfactory: Yes, I agree that escaped JSON is harder to read. Multiline strings do not need escaping, which is the reason for the suggestion, but I'm fine with leaving the current approach for now.

greyp9: Maybe I misunderstood?  Are you saying that lines 64, 65 would be better as four lines?

exceptionfactory: I meant that the JSON could be defined as follows with a multiline string:
```
private static final String SCHEMA_JSON = """
{
  "name": "test",
  "type": "record",
  "fields": [
    {
      "name": "text",
      "type": "string"
    },
    {
      "name": "ordinal",
      "type": "long"
    }
  ]
}
""";

greyp9: TIL!  I'll make that change.

exceptionfactory:
Thanks for the updates to the tests @greyp9, if you can just add a message to the `IOException`, this looks ready to go.

greyp9:
> Thanks for the updates to the tests @greyp9, if you can just add a message to the `IOException`, this looks ready to go.

I borrowed from the intent of the javadoc for that Avro class.

PR #10368, opened 2025-10-02
NIFI-15038: Fix Snowflake connection unwrap class after driver upgrade

mark-bathori:
# Summary

[NIFI-15038](https://issues.apache.org/jira/browse/NIFI-15038)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [x] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services-api/src/main/java/org/apache/nifi/processors/snowflake/SnowflakeConnectionWrapper.java
+++ b/nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services-api/src/main/java/org/apache/nifi/processors/snowflake/SnowflakeConnectionWrapper.java
@@ -18,6 +18,7 @@
 package org.apache.nifi.processors.snowflake;
 
 import net.snowflake.client.jdbc.SnowflakeConnection;
+import net.snowflake.ingest.internal.net.snowflake.client.jdbc.SnowflakeConnectionV1;
 
 import java.sql.Connection;
 import java.sql.SQLException;
@@ -35,7 +36,7 @@ public class SnowflakeConnectionWrapper implements AutoCloseable {
         connection.close();
     }
 
-    public SnowflakeConnection unwrap() throws SQLException {
-        return connection.unwrap(SnowflakeConnection.class);
+    public SnowflakeConnectionV1 unwrap() throws SQLException {
+        return connection.unwrap(SnowflakeConnectionV1.class);
     }
 }


exceptionfactory:
On a cursory review, it looks like this could be related to the Ingest SDK including the shaded JDBC Driver. A clean solution would involve switching to the unshaded version. I'm fine with this change as it stands if it works for now, but I may look at refactoring the dependencies, which should allow switching back to the standard `SnowflakeConnection` interface.

mark-bathori:
Thanks @exceptionfactory and @pvillard31 for the comments. It seems like keeping the interface and changing to the net.snowflake.ingest.internal.net.snowflake.client.jdbc packege also solves the issue so referencing directly to the V1 class is not needed.
I'll check if I can make an IT test for this.

mark-bathori:
@exceptionfactory @pvillard31 After investigating this a bit more I think we have two options for solving this properly:
1. Use an unshaded version from the `snowflake-ingest-sdk`.
2. Remove the `snowflake-jdbc` dependency and rely on the version included in the `snowflake-ingest-sdk`. The main drawback here is that we lose the ability to upgrade the `snowflake-jdbc` dependency version independently.

Which one do you think would be better?

exceptionfactory:
> @exceptionfactory @pvillard31 After investigating this a bit more I think we have two options for solving this properly:
> 
>     1. Use an unshaded version from the `snowflake-ingest-sdk`.
> 
>     2. Remove the `snowflake-jdbc` dependency and rely on the version included in the `snowflake-ingest-sdk`. The main drawback here is that we lose the ability to upgrade the `snowflake-jdbc` dependency version independently.
> 
> 
> Which one do you think would be better?

Going with the unshaded version of `snowflake-ingest-sdk` is definitely better from a dependency management perspective, but it does require more careful evaluation. I've spent some time working on this for a draft pull request in support of Snowpipe Streaming, so this is something I could evaluate separately if that would be helpful.

exceptionfactory:
@mark-bathori I have submitted an alternative pull request #10425 that changes the dependencies to use the Snowflake JDBC Thin and Snowflake Ingest SDK unshaded drivers. As described in the pull request, this change supports better control over transitive dependencies, and also addresses the Connection unwrap issue described in the Jira issue. The Snowflake Ingest SDK uses the Snowflake JDBC Driver to upload files, so the coupling remains, but that already existed with the standard and shaded versions of these libraries.
 
 I'm closing this pull request in favor of the new one, and glad to continue the discussion there as needed.

PR #10372, opened 2025-10-02
NIFI-15039 Replace usage of Enum.values() when used often

mosermw:
# Summary

[NIFI-15039](https://issues.apache.org/jira/browse/NIFI-15039)

In Enums that are used often, created a static List of Enum values, added a getValues() method to return that list, and used that getValues() method where appropriate.   Did not modify usage of Enum.values() cases when it clearly isn't called often, such as in PropertyDescriptor allowedValues().

Modified IndexOperationRequest forValue() method, in nifi-elasticsearch-bundle, to loop through values rather than using streams, which performs much better for a list of 5 values.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-security-cert/src/main/java/org/apache/nifi/security/cert/GeneralNameType.java
+++ b/nifi-commons/nifi-security-cert/src/main/java/org/apache/nifi/security/cert/GeneralNameType.java
@@ -16,6 +16,8 @@
  */
 package org.apache.nifi.security.cert;
 
+import java.util.List;
+
 /**
  * Subject Alternative Name General Name Types defined according to RFC 3280 Section 4.2.1.7
  */
@@ -37,6 +39,12 @@ public enum GeneralNameType {
     IP_ADDRESS(7, "iPAddress", String.class),
 
     REGISTERED_ID(8, "registeredID", byte[].class);
+
+    private static final List<GeneralNameType> VALUES = List.of(GeneralNameType.values());
+
+    public static List<GeneralNameType> getValues() {
+        return VALUES;
+    }
 
     private final int nameType;
 
--- a/nifi-commons/nifi-security-cert/src/main/java/org/apache/nifi/security/cert/GeneralNameType.java
+++ b/nifi-commons/nifi-security-cert/src/main/java/org/apache/nifi/security/cert/GeneralNameType.java
@@ -16,8 +16,6 @@
  */
 package org.apache.nifi.security.cert;
 
-import java.util.List;
-
 /**
  * Subject Alternative Name General Name Types defined according to RFC 3280 Section 4.2.1.7
  */
@@ -39,12 +37,6 @@ public enum GeneralNameType {
     IP_ADDRESS(7, "iPAddress", String.class),
 
     REGISTERED_ID(8, "registeredID", byte[].class);
-
-    private static final List<GeneralNameType> VALUES = List.of(GeneralNameType.values());
-
-    public static List<GeneralNameType> getValues() {
-        return VALUES;
-    }
 
     private final int nameType;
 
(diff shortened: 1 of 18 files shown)


joewitt:
Is there more to help explain or quantify the benefit of these changes?

mosermw:
> Is there more to help explain or quantify the benefit of these changes?

Hi Joe!  I found this blog to be interesting and useful.  https://www.gamlor.info/wordpress/2017/08/javas-enum-values-hidden-allocations/

We found this scenario in our own code, then I found InvokeHTTP experiences this in every onTrigger().  When looking into that I found more cases in the NiFi codebase.

Hard to quantify in garbage collection scenarios, other than "every little bit helps".

joewitt:
Cool yeah.  Any chance you have a profiled example of before after such as what you found with ListenHTTP?

i defer to others of course who are more closely helping with code quality/efficiency than I.  But generally speaking it would be good for cases like this where we're adding code but not function - that we can quantify the benefit a bit.

Not trying to cause busy work.  You're probably right... Just would be nice to see.

mosermw:
Sorry, I didn't do any profiling of InvokeHTTP or other NiFi code (there is a change in the PR that looks like would help PutElasticsearchRecord).  I did profile my own code, mainly by using Java FlightRecorder to capture a million iterations of a unit test.  Those results would always show fewer garbage collections occurred after these changes, indicating less garbage generated.  Exactly how much this would affect GC in NiFi overall depends on lots of things.  Taking this PR on its own, GC improvements might not even be noticeable.

In NiFi overall, we found the most noticeable GC savings comes from using the `-XX:+UseStringDeduplication` flag in G1, which perhaps is making FlowFile attribute storage more efficient.

exceptionfactory (commented):
Thanks for proposing this change @mosermw.

I'm familiar with the potential optimization, but overall, I'm not in favor of introducing a `getValues()` method as a general pattern for enums. I think it would be easy to miss using `getValues()` versus `values()`, and thus introducing it across the board does not seem generally beneficially.

On the other hand, if there are particular hot spots, I think it is better to declare a local `List` instead of adding the convenience method on the `enum` itself. What do you think about that approach?

joewitt:
I feel similarly 

However very interested in Michael's other finding related to -XX:+UseStringDeduplication

mosermw:
Removed most of the changes, but left the InvokeHTTP change using a local static List, because it's called for every flowfile.  Also left the IndexOperationRequest change, which is used by PutElasticsearchJson for every flowfile and by PutElasticsearchRecord for every Record.

mosermw:
For -XX:+UseStringDeduplication, I guess it's anecdotal story time.  We had a heavily loaded cluster that experienced a sluggish UI and sometimes the UI would completely stop responding.  Also, the nodes would often drop from the cluster.  We traced this to excessive garbage collection and stop-the-world time.  It used G1GC and was doing full gc every 10 seconds with 5 seconds of stop-the-world.

Once we applied the UseStringDeduplication flag, we saw a 50% reduction in full gc events with stop-the-world times in milliseconds.  This made the UI happy and cluster nodes stopped dropping out.

Of course, every NiFi flow is different, so it's not predictable how much this can help in general cases.  We had some NiFi clusters that never do a full gc.  Our prevailing theory was this cluster dealt with a hundreds of flowfile attributes on every file, which of course are all Strings.  It also had custom components that also dealt with a lot of Strings while processing each file.

exceptionfactory (approved):
Thanks for narrowing the scope of changes @mosermw, I agree with the remaining adjustments given the usage on every FlowFile processed. +1 merging

Thanks for sharing the background experience on UseStringDeduplication, it makes sense that it could be helpful in flows that handle large numbers of FlowFiles with many attributes per FlowFile.

kevdoran:
@mosermw Interesting, thanks for sharing! I captured the potential benefits of adding `-XX:+UseStringDeduplication` to NiFi's default JVM GC args in [NIFI-15091](https://issues.apache.org/jira/browse/NIFI-15091) and will try to do some experimentation and profiling soon.

PR #10377, opened 2025-10-03
NIFI-15045 - Prevent empty tooltip in help docs for property definition allowable values

rfellows:
# Summary

[NIFI-15045](https://issues.apache.org/jira/browse/NIFI-15045)

If there is no description for an allowable value, the UI would still show an info icon with an empty tooltip. This PR just removes the icon so they can't see the empty tooltip if there is no description.

One example can be found in the docs for `org.apache.nifi.controller.MonitorMemory`, in particular the Memory Pool allowable values:

This is what it looked like before:
image


mcgilman:
Reviewing...

mcgilman (approved):
Thanks for the PR @rfellows!

PR #10383, opened 2025-10-03
NIFI-14471 clear branch value when switching between registry clients…

markobean:
… in the import from registry form

# Summary

NiFi can be configured with multiple Registry clients. When multiple clients have at least one that supports branching, e.g. GitHubFlowRegistryClient, and another which does not, e.g. NifiRegistryFlowRegistryClient, the client not supporting branching was being handed branch information. The information was a remnant from the branch-supporting client. Worse, if that branch name was not "main", an error was thrown.

On the "import from registry" form in the UI, the code was updated so that the branch details are wiped each time the registry selection is changed thereby preventing branches from being included in the API calls for non-supporting clients.

[NIFI-14471](https://issues.apache.org/jira/browse/NIFI-14471)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

See Jira issue for steps to reproduce - except where the error is reported in the issue, the import of a dataflow should succeed.

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
@@ -199,7 +199,6 @@ export class ImportFromRegistry extends CloseOnEscapeDialog implements OnInit {
         if (this.supportsBranching) {
             this.loadBranches(registryId);
         } else {
-            this.clearBuckets();
             this.loadBuckets(registryId);
         }
     }
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
@@ -195,8 +195,8 @@ export class ImportFromRegistry extends CloseOnEscapeDialog implements OnInit {
 
     registryChanged(registryId: string): void {
         this.supportsBranching = this.clientBranchingSupportMap.get(registryId) || false;
+        this.clearBranches();
         if (this.supportsBranching) {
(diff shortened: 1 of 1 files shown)


rfellows:
Will review...

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
```diff
@@ -195,8 +195,8 @@ export class ImportFromRegistry extends CloseOnEscapeDialog implements OnInit {
 
     registryChanged(registryId: string): void {
         this.supportsBranching = this.clientBranchingSupportMap.get(registryId) || false;
+        this.clearBranches();
         if (this.supportsBranching) {
-            this.clearBranches();
             this.loadBranches(registryId);
         } else {
             this.clearBuckets();
```
rfellows: Moving the `clearBranches` call outside of the conditional affords us the opportunity to forgo calling `clearBuckets` here since `clearBranches` will call `clearBuckets`.

```suggestion
```

rfellows (changes requested):
Your solution works well, there is just one minor optimization we can now take as a result.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/import-from-registry/import-from-registry.component.ts
markobean: Good catch. I will update.

rfellows (approved):
Nice work @markobean, thanks for the fix! 👍

markobean:
FWIW - after the latest change, I repeated my test in a running NiFi. Everything continued to work well.

rfellows:
> FWIW - after the latest change, I repeated my test in a running NiFi. Everything continued to work well.

Thanks, I had tested out that change locally before suggesting it as well.

I'll merge this once CI is green. Thanks again!

markobean:
Thanks Rob

PR #10384, opened 2025-10-03
NIFI-15043: Update Parameter Context Rest API endpoint fails on NiFi cluster

jrsteinebrey:
# Summary

[NIFI-15043](https://issues.apache.org/jira/browse/NIFI-15043)

# Tracking

### Issue Tracking

- [X] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-15043) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

I created a NiFi flow on a NiFi cluster which calls the nifi-api rest endpoint affected by this bug.
Before the code change, the nifi-api call failed and I stepped through the debugger to determine the cause.
After the code change, the nifi-api endpoint works in both cluster and standalone modes.

### Build

  - [X] JDK 21
  - [X] JDK 25

### Licensing

No dependency changes.

### Documentation

No documentation changes needed.

### DIFF
--- a/nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/NiFiClientUtil.java
+++ b/nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/NiFiClientUtil.java
@@ -741,7 +741,6 @@ public class NiFiClientUtil {
 
         final ParameterContextEntity entityUpdate = createParameterContextEntity(existingEntity.getComponent().getName(), existingEntity.getComponent().getDescription(),
             parameterEntities, inheritedParameterContextIds, null);
-        entityUpdate.setId(existingEntity.getId());
         entityUpdate.setRevision(existingEntity.getRevision());
         entityUpdate.getComponent().setId(existingEntity.getComponent().getId());
 


jrsteinebrey:
The problem occurs on a clustered NiFi node because it generates a uri to replicate the request to the other Nifi nodes in the cluster. The uri gets looks like this:[ *https://redacted:8443/nifi-api/parameter-contexts/null*|https://redacted:8443/nifi-api/parameter-contexts/null] and gets created on this line:
https://github.com/jrsteinebrey/nifi/blob/423381d4c53e5765d9a92bebf72804f36294bfa7/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/util/ParameterUpdateManager.java#L217
                updateUri = new URI(exampleUri.getScheme(), exampleUri.getUserInfo(), exampleUri.getHost(),
                        exampleUri.getPort(), "/nifi-api/parameter-contexts/" + updatedContext.getId(), null, exampleUri.getFragment());

The uri has "null" in it because ParameterContextEntity updatedContext has a null value for its id instance variable because it never gets set. ParameterContextEntity which extends ComponentEntity has an instance variable called parameterContextDTO which stores the actual id. Other ComponentEntity subclasses store the id directly in their id  instance variable.

A robust solution is to make ParameterContextEntity behave safely by setting the ParameterContextEntity id to equal the parameterContextDTO's id which it holds.

pvillard31:
Could a system test be added to reproduce the issue? In ClusteredParameterContextIT for example? Or we don't have a way to reproduce it easily?

jrsteinebrey:
Thanks, @pvillard31. I will look into ClusteredParameterContextIT.

jrsteinebrey:
@pvillard31 I changed a piece of test code in the second commit. Now my first code change is tested and needed for a successful result by over a dozen integration tests now.

PR #10387, opened 2025-10-04
NIFI-13402 - Multiple flowfiles as output for Python Processors

pvillard31:
# Summary

NIFI-13402 - Multiple flowfiles as output for Python Processors

### Context

The current Python bridge allows each invocation of a Python Processor to return exactly one `FlowFileSourceResult` or `FlowFileTransformResult`. This works for simple use-cases but prevents Python processors from emitting multiple FlowFiles in a single trigger, which can be common for some use cases (pagination, splitting, etc).

The taken approach in this PR is to have the minimal changes needed across the Java bridge and the Python helper library to support multiple results while preserving backward compatibility.

### Goals

- Allow Python source processors to emit zero, one, or many FlowFiles per invocation without additional Java code
- Allow Python transform processors to emit zero, one, or many FlowFiles derived from the incoming FlowFile in a single `onTrigger` call
- Preserve existing processor behavior and API surface for implementations that continue to return a single result object
- Provide predictable handling of the original FlowFile when multiple results are returned

### Non-Goals

- Supporting Java-side streaming of arbitrarily large result sequences (iterables will be materialised into memory per call)
- Changing RecordTransform behavior

### Python Return Contract

Processors may return either:

- a single `FlowFile*Result` instance (current behaviour), or
- any iterable (list, tuple, generator) containing `FlowFile*Result` instances

Returning `None` continues to mean "no output".

Python helper classes (`FlowFileTransform` and `FlowFileSource`) will normalise generators into lists before returning to Py4J to ensure predictable conversion.

### FlowFileSourceProxy Changes

- Detect when the bridge returns a `Collection` or `Iterator` and iterate over each element, creating and transferring a FlowFile per result
- Maintain the current fast-path for single results to avoid overhead
- Continue to honour the implicit `success` relationship while allowing per-result relationship names
- `FlowFileSourceResult.free()` is invoked for every element to avoid leaking proxies

### FlowFileTransformProxy Changes

- Fan-out the incoming FlowFile when multiple results are returned by the Python side
- Default behaviour when more than one result is provided:
  - clone the original FlowFile once per result
  - apply the same attribute/content updates that are currently performed
  - transfer each clone to its requested relationship
- Single-result behaviour remains unchanged: mutate the original FlowFile in-place and transfer to the requested relationship, cloning only if the `original` relationship is expected (matching existing logic)
- The original FlowFile will be routed to `REL_ORIGINAL` when that relationship is not auto-terminated and at least one result was processed. If the relationship is auto-terminated we remove the clone as currently implemented
- Any result that requests the `failure` relationship causes the original FlowFile to be routed to failure once, mirroring today's semantics. Additional `failure` results on the same trigger is ignored with a warning to avoid double-transfer attempts
- Invalid result types (for example, returning plain strings) will trigger a `ProcessException` with a detailed message

### Backward Compatibility

- Existing processors that return a single `FlowFile*Result` continue to work without modification
- Processors that currently return a single result wrapped in a list will behave the same after this change
- The default cloning behaviour for multi-result transforms mirrors NiFi’s Java processors like `SplitText`

### Error Handling

- Null or empty iterables result in no FlowFiles being transferred
- Mixed results where some are `None` will be skipped with a warning and without failing the entire batch
- Any exception raised while handling a specific result (for example, write failure) will roll back the session and rethrow a `ProcessException`

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-docs/src/main/asciidoc/python-developer-guide.adoc
+++ b/nifi-docs/src/main/asciidoc/python-developer-guide.adoc
@@ -223,6 +223,12 @@ The constructor has a single required positional argument, the `relationship` to
 the FlowFile are to be updated, the FlowFile's new contents should be returned via the `contents` argument. Any FlowFile attributes that
 are to be added or modified may additionally provided using the `attributes` argument.
 
+The method may also return an iterable (for example, a list or generator) of `FlowFileTransformResult` instances. In this case the
+framework clones the original FlowFile once per result, applies the contents and attributes described by that result, and transfers each
+clone to the requested relationship. After the iterable has been processed the original FlowFile is routed to the `original`
+relationship (unless that relationship is auto-terminated). Returning the `failure` relationship is only supported when a single result
+is produced.
+
 
 
 [[record-transform]]
@@ -328,6 +334,10 @@ empty `FlowFileSourceResult`. When `create()` returns with `None`, the processor
 When there is nothing to return, it might be useful to yield the processor's resources and not schedule the processor
 to run for the period of time defined by the processor's Yield Duration. This can be achieved by calling
 `context.yield_resources()` from the processor's `create` method right before returning `None`.
+
+A source processor may also return an iterable of `FlowFileSourceResult` objects. Each entry in the iterable causes the framework to
+create a new FlowFile, apply the specified contents and attributes, and transfer it to the requested relationship. Returning `None` or
+an empty iterable results in no FlowFiles being created for that invocation.
 
 
 [[property-descriptors]]
--- a/nifi-extension-bundles/nifi-py4j-extension-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/python/processor/FlowFileSource.java
+++ b/nifi-extension-bundles/nifi-py4j-extension-bundle/nifi-py4j-bridge/src/main/java/org/apache/nifi/python/processor/FlowFileSource.java
@@ -19,6 +19,6 @@ package org.apache.nifi.python.processor;
 
 public interface FlowFileSource extends PythonProcessor {
 
-    FlowFileSourceResult createFlowFile();
+    Object createFlowFile();
 
 }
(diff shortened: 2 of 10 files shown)


exceptionfactory (changes requested):
Thanks for proposing this improvement to the Python Processor API @pvillard31.

I appreciate the goal of maintaining compatibility with existing implementations, and I agree that any improvement along these lines should not require changing existing Python Processors.

On initial review, however, I'm not in favor of changing the interface contract for `PythonProcessor` interfaces to return a generic `Object`. This seems too ambiguous, and moves in an uncertain direction when it comes to supported and expected output types.

The Python Processor API is purposefully opinionated and limited, so rather than attempting to fit support for multiple FlowFiles into existing interfaces, I think it would be better to introduce a new type of interface that explicitly expects to return a collection of FlowFile results. This would avoid making the existing interfaces too generic, and also make it clear when a Python Processor is expecting to return a collection of results.

I'm open to further consideration, but that's my initial recommendation. The Jira issue mentions a limited use case, and it is questionable whether splitting one FlowFile into multiple FlowFiles is best done in Python when existing Java Processors can handle many uses. I'm sure there are reasonable use cases for returning multiple FlowFiles, but I think this needs further consideration.

pvillard31:
Thanks @exceptionfactory - I understand the concern. It felt like the best approach to not overcomplicate the Python API with completely new types of interfaces.

exceptionfactory (changes requested):

On further consideration, I'm still concerned about introducing support for multiple FlowFiles using this strategy. I think it is better to keep the clear interface definitions, and not support returning multiple FlowFiles.

It may be better to consider something like this in the context of supporting the ability to read multiple FlowFiles. That is a more significant change, but it provides the opportunity for a new Python Processor interface, versus changing the existing interfaces.

pvillard31:
Fair enough. Closing the PR, thanks @exceptionfactory

henrikjohansen:
I was sooo happy to see that someone finally took a stab at implementing support for [NIFI-13402](https://issues.apache.org/jira/browse/NIFI-13402) since I have several Processor implementations that are collecting dust until this feature arrives ...

The inability to output multiple flowfiles continues to be a **major** limitation for Python based processors making it impossible to build certain types of processors or use Python processors for certain workloads.

Thank you @pvillard31 for the effort even though the PR was rejected.

mattyb149:
Couldn't we just add an overloaded default method to FlowFileTransform and the other Transforms to return a List<FlowFileTranformResult>? Not sure how default methods work in Py4J but it seems like it would be a welcome addition, not a change in existing functionality.

PR #10388, opened 2025-10-04
NIFI-13673 - Add example, testing and documentation for dependent properties in Python processors

pvillard31:
# Summary

NIFI-13673 - Add example, testing and documentation for dependent properties in Python processors

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/properties.py
+++ b/nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/properties.py
@@ -16,7 +16,7 @@
 from enum import Enum
 import re
 
-from py4j.protocol import Py4JJavaError
+from py4j.protocol import Py4JError
 
 from nifiapi.componentstate import StateManager
 from nifiapi.__jvm__ import JvmHolder
@@ -254,6 +254,8 @@ class PropertyDescriptor:
         return builder.build()
 
 
+
+
     def __get_allowable_values(self, gateway):
         if self.allowableValues is None:
             return None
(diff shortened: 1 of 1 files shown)


pvillard31:
Yeah @exceptionfactory , I did think the same and all of this is because of the changes coming as part of the work around NIFI-14400.

I pushed a commit where I drop the catch-and-skip pattern and I now build a deterministic view of descriptors, values, and dependency relationships using recursive checks in ProcessContext/ValidationContext. This way we only read properties whose prerequisites are satisfied; optional Java hooks like getSupportedPropertyDescriptors() are invoked defensively to avoid Py4J reflection errors. This keeps Java-side failures visible while still filtering out dependent descriptors appropriately.

This is probably not perfect for some scenarios but we are no longer doing a catch all approach.

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/properties.py
```diff
@@ -293,6 +298,106 @@ def __add_resource_definition(self, gateway, resource_definition, builder):
...
 
 
+def _collect_descriptor_metadata(java_context):
+    descriptor_list = []
+    descriptor_lookup = {}
+    value_lookup = {}
+
+    try:
+        java_properties = java_context.getProperties()
+    except AttributeError:
```
exceptionfactory, line 308: What can trigger an `AttributeError`? Should this be logged or otherwise thrown?

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/properties.py
```diff
@@ -293,6 +298,106 @@ def __add_resource_definition(self, gateway, resource_definition, builder):
...
+    try:
+        supported_descriptor_supplier = getattr(java_context, 'getSupportedPropertyDescriptors')
+    except AttributeError:
+        supported_descriptor_supplier = None
+
+    supported_descriptors = None
+    if callable(supported_descriptor_supplier):
+        try:
+            supported_descriptors = supported_descriptor_supplier()
+        except Py4JError:
```
exceptionfactory, line 331: What can trigger the `Py4JError`?

exceptionfactory (commented):
Thanks for the updates @pvillard31, the revised approach looks much clearer.

I noted two questions related to particular errors, which seem like they could at least use a comment, if not other potential changes.

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/properties.py
pvillard31: OK so I don't think I have a perfect understanding of all the nuances here but... based on my understanding we might have this in some cases. When the Python side wraps the incoming `java_context`, we assume it is a full NiFi `ProcessContext/ValidationContext`, but during a few code paths NiFi gives us lighter-weight stubs that only expose the descriptor discovery APIs, and trying to call `getProperties()` raises an `AttributeError`. The try/except lets us ignore those cases: if the stub doesn’t carry any configured values yet, we just treat the property map as missing and continue with whatever metadata we can obtain from `getSupportedPropertyDescriptors()`. Maybe I misunderstood and this is not absolutely necessary.

nifi-framework-bundle/nifi-framework-extensions/nifi-py4j-framework-bundle/nifi-python-extension-api/src/main/python/src/nifiapi/properties.py
pvillard31: So here we’re calling back into whatever Java proxy Py4J has wrapped as `java_context`. When Py4J can’t actually make that call (like cannot call `getSupportedPropertyDescriptors` or the proxy has gone stale because the component was reloaded and the Java side disposed the old object ID), we would get the Py4JError. So it's a way to gracefully handle this case.

exceptionfactory (approved):
Thanks for responses @pvillard31, on further review, the latest version looks good! +1 merging

PR #10391, opened 2025-10-05
NIFI-15057 - Add Record Path like helper methods for Record Transform Python processors

pvillard31:
# Summary

NIFI-15057 - Add Record Path like helper methods for Record Transform Python processors

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-docs/src/main/asciidoc/python-developer-guide.adoc
+++ b/nifi-docs/src/main/asciidoc/python-developer-guide.adoc
@@ -283,6 +283,30 @@ If the partition has more than one field in the dictionary, all fields in the di
 the Records to be written to the same output FlowFile.
 
 
+==== Working with Record Paths
+
+Manipulating just a small portion of an incoming record is a very common use case for `RecordTransform` processors. To make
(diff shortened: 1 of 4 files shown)


exceptionfactory (commented):
Thanks for proposing this addition @pvillard31.

It looks like a useful feature, and I agree that attempting to integrate existing NiFi RecordPath support would be challenging on several levels.

I was wondering whether this is something that should be part of the NiFi Python Processor API, so I looked around for similar options and came across the [glom](https://pypi.org/project/glom/) library. The syntax uses `.` instead of `/` for nested elements, but it otherwise has some similarities.

If that could serve a similar purpose, it may be better to reference it as an example for now, and avoid introducing something directly into the NiFi API.

What do you think?

pvillard31:
Thanks for the review @exceptionfactory - I followed your suggestion: I removed the nifiapi changes and updated my processor example and testing to rely on the suggested glom library.

exceptionfactory (approved):
Thanks for implementing the revised approach based on `glom`, looks good! +1 merging

PR #10393, opened 2025-10-05
NIFI-15058 - Make git based Flow Registry Clients verifiable

pvillard31:
# Summary

NIFI-15058 - Make git based Flow Registry Clients verifiable

Screenshot 2025-10-05 at 16 09 54

Screenshot 2025-10-05 at 16 10 14

Screenshot 2025-10-05 at 16 10 25

Screenshot 2025-10-05 at 16 11 34

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/settings/ui/registry-clients/edit-registry-client/edit-registry-client.component.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/settings/ui/registry-clients/edit-registry-client/edit-registry-client.component.ts
@@ -22,7 +22,7 @@ import { MatInputModule } from '@angular/material/input';
 import { MatCheckboxModule } from '@angular/material/checkbox';
 import { MatButtonModule } from '@angular/material/button';
 import { AsyncPipe } from '@angular/common';
-import { Observable } from 'rxjs';
+import { Observable, of } from 'rxjs';
 import {
     InlineServiceCreationRequest,
     InlineServiceCreationResponse,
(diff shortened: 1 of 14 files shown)


rfellows:
Reviewing UI changes...

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/settings/ui/registry-clients/edit-registry-client/edit-registry-client.component.ts
```diff
@@ -142,9 +151,28 @@ export class EditRegistryClient extends TabbedDialog {
         });
     }
 
+    private getModifiedProperties(): ModifiedProperties {
+        const propertyControl: AbstractControl | null = this.editRegistryClientForm.get('properties');
+        if (propertyControl) {
```
rfellows: There is a bug in this implementation. If you open a saved github registry client and attempt to Verify it without making any changes, it will fail every time. Once you make a change to one of the properties and try to verify again, it will pass (assuming the configuration is actually good)


The problem is that if the form is not `dirty`, you are passing essentially the whole configuration back. Since the sensitive value isn't known to the UI at this point, 5 asterisks (`*****`) is passed down as the value of the sensitive property and it is failing.

The API only expects changed properties to be included in the payload. You get away with it when initially configuring the client since the user is actually providing the sensitive value at that point.


```suggestion
        if (propertyControl && propertyControl.dirty) {
```

rfellows (changes requested):
I do have a question/concern in addition to a suggested change (see inline comment).

This implementation and offering of Verification for registry clients is cool, but it is a bit misleading for non-github clients. Verification always passes. This threw me for a loop when I tried to verify my local registry client configuration that was not running but the verification passed. This is typically how we handle verification for things that don't support it, but in this case it felt like it should have worked. There is no enable option for these things that would then raise a bulletin if it failed, you only find out when you go to try to use the client. Not saying we have to implement verification support for all of the clients out of the box, but it is something to be aware of and could cause confusion to users.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/settings/ui/registry-clients/edit-registry-client/edit-registry-client.component.ts
pvillard31: thanks @rfellows, good catch, I committed your suggested change

pvillard31:
Thanks for the review @rfellows - you're right that if the verify method is not implemented and the configuration is correct (based on the validators on the properties), it'd say that this is successful even though it may not true. I agree that this could be confusing but I do think this is aligned with what we have right now: a processor without a verify method would show successful at verification if configuration is valid but could very well throw errors when the processor is started. I think this is OK. Also the only implementation that is not git based in Apache NiFi is the NiFi Registry implementation for which we could also implement the verify method in a follow up improvement.

PR #10396, opened 2025-10-06
NIFI-1501 - Allow users to only show controller services for the current process group (scope) in the listing

rfellows:
# Summary

[NIFI-15051](https://issues.apache.org/jira/browse/NIFI-15051)

## Add Filter to Show Current Scope Controller Services Only

This PR introduces a filtering feature to the Controller Services listing page that allows users to view only the controller services defined in the current process group, filtering out inherited services from parent process groups.

### Changes

**User-Facing Features:**
- Added a "Show current scope only" checkbox filter with help tooltip on the Controller Services page
- Filter only appears when viewing non-root process groups (hidden at root level)
- Tooltip displays the current process group name (or ID if not readable)
- Filter state defaults to unchecked, showing all services (current + inherited)

**Technical Implementation:**
- Converted `serviceState$` observable to signal-based state management using `selectSignal()`
- Added `showCurrentScopeOnly` signal to track filter state
- Implemented `filteredControllerServices` computed property for reactive filtering
- Added `shouldShowFilter` computed property to conditionally display filter based on hierarchy
- Updated module imports to include `FormsModule`, `MatCheckboxModule`, and `NifiTooltipDirective`

**Testing:**
- Added comprehensive test suite with 16 tests covering:
  - Component initialization
  - Filter functionality (enabled/disabled states, edge cases)
  - Computed property behavior
  - Template integration and user interactions
  - Root process group handling
- All tests passing ✅

### Why This Change?

Users working in nested process groups often inherit many controller services from parent groups, making it difficult to identify which services are defined at the current scope. This filter provides a quick way to focus on locally-defined services while still allowing visibility into inherited services when needed.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/service/guard/base-guard.utils.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi-registry/src/app/service/guard/base-guard.utils.ts
@@ -16,7 +16,15 @@
  */
 
 import { inject } from '@angular/core';
-import { ActivatedRouteSnapshot, CanActivateFn, CanMatchFn, Router, RouterStateSnapshot, UrlSegment, UrlTree } from '@angular/router';
+import {
+    ActivatedRouteSnapshot,
+    CanActivateFn,
+    CanMatchFn,
+    Router,
+    RouterStateSnapshot,
+    UrlSegment,
+    UrlTree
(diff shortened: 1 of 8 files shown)


joewitt:
@rfellows you'll want to edit the commit message to NIFI-15051

scottyaslan:
Reviewing...

PR #10400, opened 2025-10-07
NIFI-15062 Add PutIcebergRecord Processor and Services

exceptionfactory:
# Summary

[NIFI-15062](https://issues.apache.org/jira/browse/NIFI-15062) Adds a `PutIcebergRecord` Processor and several Controller Services that provide initial integration for storing records in Apache Iceberg tables.

The Apache Iceberg ecosystem supports a wide variety of catalogs, storage providers, and file formats. The purpose of this pull request is to provide several Controller Service abstractions that enable extensible integration, with a specific implementation of each Controller Service. With the number of potential integration options, Apache NiFi should not necessarily implement support for every possible solution, but should provide extension points that enable focused types of integration. The `iceberg-api` library is the foundation for this approach.

The `nifi-iceberg-bundle` includes multiple modules that have the following dependency hierarchy:

- `nifi-iceberg-shared-nar`
  - `nifi-iceberg-services-api-nar`
    - `nifi-iceberg-processors-nar`
    - `nifi-iceberg-rest-catalog-nar`
    - `nifi-iceberg-aws-nar`
    - `nifi-iceberg-parquet-writer-nar`

The `nifi-iceberg-shared-nar` contains the `iceberg-api` and `iceberg-core` libraries along with transitive dependencies.

The `nifi-iceberg-services-api-nar` depends on `iceberg-api` and incorporates the Apache NiFi Controller Service interfaces that align with `iceberg-api` interfaces.

The `nifi-iceberg-processors-nar` contains the `PutIcebergRecord` Processor, which references properties for the following Controller Services:

- `IcebergCatalog`
- `IcebergWriter`
- `IcebergFileIOProvider`

These three interfaces define the primary extension points for external integration.

The `nifi-iceberg-rest-catalog-nar` contains the `RESTIcebergCatalog` implementation of the `IcebergCatalog` Controller Service. This implementation configures the `RESTSessionCatalog` from the `iceberg-core` library and supports Catalog Authentication using OAuth2 with Client Credentials or Bearer Tokens. Building on the `iceberg-core` library provided in `nifi-iceberg-shared-nar`, the `nifi-iceberg-rest-catalog-nar` does not have any additional dependencies. The `RESTIcebergCatalog` defines a `FileIO Provider` property that supports configurable Controller Services for Iceberg `FileIO` implementations.

The `nifi-iceberg-aws-nar` contains the `S3IcebergFileIOProvider` which configures and returns the Iceberg `S3FileIO` class. Support for S3 requires a number of AWS SDK 2 libraries, which is one of the primary reasons for separate packaging of `FileIOProvider` implementations. The S3 implementation supports configurable authentication using Basic or Session Credentials, as well as Vended Credentials, where the REST Catalog is expected to provide the required credentials.

The `nifi-iceberg-parquet-writer-nar` contains the `ParquetIcebergWriter` Controller Service, supporting Apache Parquet serialization. Apache Parquet has a number of transitive dependencies, including a dependency on the `hadoop-common` library. The NAR packaging excludes many unnecessary transitive dependencies and has an explicit list of dependencies required at runtime for Parquet serialization.

This implementation structure and Controller Service design strategy should serve as the basis for additional storage provider implementations. With Apache Parquet being the predominant format for Apache Iceberg, direct support for other file formats may not be necessary. The variety of Iceberg REST Catalog implementations may require additional configuration options in the future, but the core `IcebergCatalog` Controller Service abstraction provides a decoupled strategy for future implementation.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/PutIcebergRecord.java
+++ b/nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/PutIcebergRecord.java
@@ -1,0 +1,264 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.iceberg;
+
+import org.apache.iceberg.types.Types;
+import org.apache.nifi.processors.iceberg.record.DelegatedRecord;
+import org.apache.nifi.processors.iceberg.record.StandardStructTypeProvider;
+import org.apache.nifi.processors.iceberg.record.StructTypeProvider;
+import org.apache.nifi.serialization.record.RecordSchema;
+import org.apache.nifi.services.iceberg.IcebergCatalog;
+import org.apache.nifi.services.iceberg.IcebergRowWriter;
+import org.apache.nifi.services.iceberg.IcebergWriter;
+import org.apache.iceberg.AppendFiles;
+import org.apache.iceberg.DataFile;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.catalog.Catalog;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.nifi.annotation.behavior.InputRequirement;
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.annotation.lifecycle.OnScheduled;
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.expression.ExpressionLanguageScope;
+import org.apache.nifi.flowfile.FlowFile;
+import org.apache.nifi.processor.AbstractProcessor;
+import org.apache.nifi.processor.ProcessContext;
+import org.apache.nifi.processor.ProcessSession;
+import org.apache.nifi.processor.Relationship;
+import org.apache.nifi.processor.exception.ProcessException;
+import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.serialization.MalformedRecordException;
+import org.apache.nifi.serialization.RecordReader;
+import org.apache.nifi.serialization.RecordReaderFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.time.Clock;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+@Tags({"iceberg", "analytics", "polaris", "s3"})
+@CapabilityDescription("Store records in Iceberg Table using configurable Catalog for managing namespaces and tables.")
+@InputRequirement(InputRequirement.Requirement.INPUT_REQUIRED)
+public class PutIcebergRecord extends AbstractProcessor {
+
+    static final PropertyDescriptor ICEBERG_CATALOG = new PropertyDescriptor.Builder()
+            .name("Iceberg Catalog")
+            .description("Provider Service for Iceberg Catalog")
+            .required(true)
+            .identifiesControllerService(IcebergCatalog.class)
+            .build();
+
+    static final PropertyDescriptor ICEBERG_WRITER = new PropertyDescriptor.Builder()
+            .name("Iceberg Writer")
+            .description("Provider Service for Iceberg Row Writers responsible for producing formatted Iceberg Data Files")
+            .required(true)
+            .identifiesControllerService(IcebergWriter.class)
+            .build();
+
+    static final PropertyDescriptor RECORD_READER = new PropertyDescriptor.Builder()
+            .name("Record Reader")
+            .description("Record Reader for incoming FlowFiles")
+            .required(true)
+            .identifiesControllerService(RecordReaderFactory.class)
+            .build();
+
+    static final PropertyDescriptor NAMESPACE = new PropertyDescriptor.Builder()
+            .name("Namespace")
+            .description("Iceberg Namespace containing Tables")
+            .required(true)
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .build();
+
+    static final PropertyDescriptor TABLE_NAME = new PropertyDescriptor.Builder()
+            .name("Table Name")
+            .description("Iceberg Table Name")
+            .required(true)
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
+            .build();
+
+    static final Relationship SUCCESS = new Relationship.Builder()
+            .name("success")
+            .description("FlowFiles transferred to Iceberg")
+            .build();
+
+    static final Relationship FAILURE = new Relationship.Builder()
+            .name("failure")
+            .description("FlowFiles not transferred to Iceberg")
+            .build();
+
+    static final String RECORDS_PROCESSED_COUNTER = "Records Processed";
+
+    static final String DATA_FILES_PROCESSED_COUNTER = "Data Files Processed";
+
+    private static final List<PropertyDescriptor> properties = List.of(
+            ICEBERG_CATALOG,
+            ICEBERG_WRITER,
+            RECORD_READER,
+            NAMESPACE,
+            TABLE_NAME
+    );
+
+    private static final Set<Relationship> relationships = Set.of(SUCCESS, FAILURE);
+
+    private static final long MAXIMUM_BYTES = 536870912;
+
+    private static final StructTypeProvider structTypeProvider = new StandardStructTypeProvider();
+
+    private final Clock clock = Clock.systemDefaultZone();
+
+    private volatile Catalog catalog;
+
+    private volatile IcebergWriter icebergWriter;
+
+    @Override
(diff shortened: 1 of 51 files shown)


nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/record/StandardStructTypeProvider.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public Types.StructType getStructType(final RecordSchema recordSchema) {
+        Objects.requireNonNull(recordSchema, "Record Schema required");
+
+        final List<Types.NestedField> nestedFields = new ArrayList<>();
+
+        final ListIterator<RecordField> recordFields = recordSchema.getFields().listIterator();
+        while (recordFields.hasNext()) {
+            final RecordField recordField = recordFields.next();
+            final int id = recordFields.nextIndex();
```
pvillard31: Isn't it going to be an issue for nested records? When `getStructType()` calls itself for nested records we will have duplicate IDs in the same schema tree, no? Iceberg requires every ID in a schema tree to be globally unique.

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/record/StandardStructTypeProvider.java
```diff
@@ -0,0 +1,109 @@
...
+        };
+    }
+
+    private Type getDecimalType(final DecimalDataType dataType) {
+        return Types.DecimalType.of(dataType.getPrecision(), dataType.getScale());
+    }
+
+    private Type getListType(final ArrayDataType dataType) {
+        final DataType elementType = dataType.getElementType();
+        return Types.ListType.ofOptional(0, getType(elementType));
```
pvillard31: Wondering if that's going to be an issue here... Iceberg is going to use that `int 0` as the element’s field ID, no? This is problematic as the ID must be strictly positive I believe and it must not collide with others.

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/record/StandardStructTypeProvider.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private Type getMapType(final MapDataType dataType) {
+        return Types.MapType.ofOptional(0, 0, Types.StringType.get(), getType(dataType.getValueType()));
```
pvillard31: Same comment here.

pvillard31 (commented):
I'll try to run tests as soon as possible but I left some comments after going through the code.

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/record/StandardStructTypeProvider.java
exceptionfactory: That's a very good point, most of the initial testing did not use nested types. I will review the implementation options and implement a better strategy for nested field identification.

exceptionfactory: I refactored `PutIcebergRecord` to use the Table Schema and Struct instead of depending on this conversion from Record Schema to Iceberg Struct Type. Although this may be useful for other scenarios such as initial table creation, it is not needed in the current implementation, so I have removed this class.

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/record/StandardStructTypeProvider.java
exceptionfactory: Resolved with refactored approach, eliminating this class.

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-rest-catalog/src/main/java/org/apache/nifi/services/iceberg/catalog/RESTIcebergCatalog.java
```diff
@@ -0,0 +1,331 @@
...
+
+    @OnEnabled
+    public void onEnabled(final ConfigurationContext context) {
+        sessionCatalog = getInitializedCatalog(context);
+    }
+
+    @OnDisabled
+    public void onDisabled() {
+        try {
+            sessionCatalog.close();
```
pvillard31: Null check? I believe that if onEnabled fails and sessionCatalog is not set, we would still expect the user to disable the controller service and we would get a NPE, no?

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-rest-catalog/src/main/java/org/apache/nifi/services/iceberg/catalog/RESTIcebergCatalog.java
```diff
@@ -0,0 +1,331 @@
...
+            properties.put(CatalogProperties.WAREHOUSE_LOCATION, warehouseLocation);
+        }
+
+        final AuthenticationStrategy authenticationStrategy = context.getProperty(AUTHENTICATION_STRATEGY).asAllowableValue(AuthenticationStrategy.class);
+        final Function<Map<String, String>, RESTClient> restClientBuilder;
+
+        if (AuthenticationStrategy.BEARER == authenticationStrategy) {
+            final String bearerToken = context.getProperty(BEARER_TOKEN).getValue();
+            properties.put(OAuth2Properties.TOKEN, bearerToken);
+            properties.put(AuthProperties.AUTH_TYPE, AuthProperties.AUTH_TYPE_OAUTH2);
```
pvillard31, line 254: Should be `AuthProperties.AUTH_TYPE_BEARER`, no?

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/PutIcebergRecord.java
```diff
@@ -0,0 +1,259 @@
...
+                    writeRecords(recordReader, rowWriter, struct, recordsProcessed);
+                    session.adjustCounter(RECORDS_PROCESSED_COUNTER, recordsProcessed.get(), false);
+                } catch (final Exception e) {
+                    getLogger().error("Write Rows to Table [{}] failed {}", tableIdentifier, flowFile, e);
+                    abortWriter(rowWriter, tableIdentifier);
+                    relationship.set(FAILURE);
+                }
+            } catch (final Exception e) {
+                getLogger().error("Processing Records for Table [{}] failed {}", tableIdentifier, flowFile, e);
+                relationship.set(FAILURE);
```
pvillard31, line 186: Shouldn't we abort the writer here too?

```suggestion
                getLogger().error("Processing Records for Table [{}] failed {}", tableIdentifier, flowFile, e);
                abortWriter(rowWriter, tableIdentifier);
                relationship.set(FAILURE);
```

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/PutIcebergRecord.java
```diff
@@ -0,0 +1,259 @@
...
+        }
+
+        try {
+            final DataFile[] dataFiles = rowWriter.dataFiles();
+            appendDataFiles(table, dataFiles);
+            session.adjustCounter(DATA_FILES_PROCESSED_COUNTER, dataFiles.length, false);
+        } catch (final Exception e) {
+            getLogger().error("Appending Data Files to Table [{}] failed", tableIdentifier, e);
+            relationship.set(FAILURE);
+        }
```
pvillard31: Shouldn't we do the append only in case of success?

```suggestion
        if (SUCCESS.equals(relationship.get())) {
            try {
                final DataFile[] dataFiles = rowWriter.dataFiles();
                appendDataFiles(table, dataFiles);
                session.adjustCounter(DATA_FILES_PROCESSED_COUNTER, dataFiles.length, false);
            } catch (final Exception e) {
                getLogger().error("Appending Data Files to Table [{}] failed", tableIdentifier, e);
                relationship.set(FAILURE);
                abortWriter(rowWriter, tableIdentifier);
            }
        } else {
            abortWriter(rowWriter, tableIdentifier);
        }
```

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-rest-catalog/src/main/java/org/apache/nifi/services/iceberg/catalog/RESTIcebergCatalog.java
exceptionfactory: There is no `AUTH_TYPE_BEARER` and the use of `AUTH_TYPE_OAUTH2` aligns with the derived value set in Iceberg [AuthManagers](https://github.com/apache/iceberg/blob/cd097eb7e94686854f8b0cb42c77d3d1362a9ab8/core/src/main/java/org/apache/iceberg/rest/auth/AuthManagers.java#L56) when the `TOKEN` property is provided.

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-processors/src/main/java/org/apache/nifi/processors/iceberg/PutIcebergRecord.java
exceptionfactory: Good catch, yes!

exceptionfactory:
Thanks for testing and feedback @pvillard31, I pushed an update to address the items mentioned.

nifi-extension-bundles/nifi-iceberg-bundle/nifi-iceberg-rest-catalog/src/main/java/org/apache/nifi/services/iceberg/catalog/RESTIcebergCatalog.java
pvillard31: My bad I wrongly assumed that we had the same enum for our AuthStrategy and the AuthProperties. Thanks!

PR #10401, opened 2025-10-07
NIFI-15070 Fixed OAuth2AccessTokenProvider to properly handle permane…

turcsanyip:
…nt access tokens that have no expiration time

# Summary

[NIFI-15070](https://issues.apache.org/jira/browse/NIFI-15070)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/StandardOauth2AccessTokenProvider.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/StandardOauth2AccessTokenProvider.java
@@ -66,7 +66,6 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 @SupportsSensitiveDynamicProperties
@@ -468,7 +467,7 @@ public class StandardOauth2AccessTokenProvider extends AbstractControllerService
     }
 
     private boolean isRefreshRequired() {
-        final long expirationTime = Optional.ofNullable(accessDetails.getExpiresIn()).orElse(defaultExpirationTimeSeconds);
+        final long expirationTime = accessDetails.getExpiresIn() != null ? accessDetails.getExpiresIn() : defaultExpirationTimeSeconds;
 
         final Instant expirationRefreshTime = accessDetails.getFetchTime()
                 .plusSeconds(expirationTime)


turcsanyip:
@exceptionfactory Thanks for highlighting this issue. Actually, the changed behavior was not applied in `JWTBearerOAuth2AccessTokenProvider` which should be adjusted. However, I would not introduce alternative methods because the real problem is that the current `getExpiredIn()` and `isExpired()` methods do not handle a valid scenario (permanent tokens). In that case we would keep `getExpiredIn()` and `isExpired()` fully compatible but containing the bug.

Instead, I would suggest modifying the type to `Long` and adjusting all usages in the codebase. As this is a component interface and is not part of `nifi-api`, I don't think we must follow strict compatibility.

exceptionfactory:
Thanks for the reply @turcsanyip. Although this is not part of the `nifi-api`, the signature change could have theoretical impact of components built to interact with the Controller Service interface. With that being said, changing from `long` to `Long` should have minimal impact, so I'm willing to go forward with the change and updating all references in the project repository

turcsanyip:
Thanks @exceptionfactory. Updated the PR with the change in `JWTBearerOAuth2AccessTokenProvider` (tested with Salesforce and also with another provider that sets the `expires_in` field).

turcsanyip:
@exceptionfactory Based on further testing, it looks Salesforce expires the session even if the access token has no expiration time. So maybe it would be a better approach to add a user configurable `Expiration Time` property to `StandardOauth2AccessTokenProvider` (as in case of `JWTBearerOAuth2AccessTokenProvider`) and use this value as default when no expiration time is provided.
Please give me some time to test it thoroughly and put the review on hold till then.
Thanks

exceptionfactory:
> @exceptionfactory Based on further testing, it looks Salesforce expires the session even if the access token has no expiration time. So maybe it would be a better approach to add a user configurable `Expiration Time` property to `StandardOauth2AccessTokenProvider` (as in case of `JWTBearerOAuth2AccessTokenProvider`) and use this value as default when no expiration time is provided. Please give me some time to test it thoroughly and put the review on hold till then. Thanks

Thanks for the update, I will hold off on further feedback until you have run through some additional tests. Perhaps introducing a strategy property, something like `Expiration Handling` with `Access Token` and `Configured` as possible values, where `Configured` requires an explicit `Expiration Period` property could work for this generic service. I'm open to considering options. It may also require something more specific for Salesforce, so that is another thing to keep in mind in terms of scoping potential changes.

exceptionfactory:
@turcsanyip Are you still in the process of evaluating other changes? If it is something longer-term, perhaps it may be best to close this PR for now and revisit when ready?

turcsanyip:
@exceptionfactory Thanks for the heads-up. The improved solution has recently been tested. Will update the PR soon (later this week).

exceptionfactory:
> @exceptionfactory Thanks for the heads-up. The improved solution has recently been tested. Will update the PR soon (later this week).

Sounds good @turcsanyip, thanks for the update!

turcsanyip:
@exceptionfactory The PR has been updated. I added `Default Expiration Time` property to `StandardOauth2AccessTokenProvider` (similar to `JWT Expiration Time` in `JWTBearerOAuth2AccessTokenProvider`). The service uses this value as default when no expiration time is provided in the access token by the OAuth server.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/StandardOauth2AccessTokenProvider.java
```diff
@@ -456,11 +467,19 @@ private void getProperties(ConfigurationContext context) {
...
 
-        return Instant.now().isAfter(expirationRefreshTime);
+            return Instant.now().isAfter(expirationRefreshTime);
+        } else {
+            final Instant expirationRefreshTime = accessDetails.getFetchTime()
+                    .plusSeconds(defaultExpirationTimeSeconds)
+                    .minusSeconds(refreshWindowSeconds);
+
+            return Instant.now().isAfter(expirationRefreshTime);
+        }
```
exceptionfactory: It looks like this could be refactored to a single `return` and also the value of expiration time could be the element to set conditionally so that it would read something like
```
final long expirationRefreshTime;
if (accessDetails.getExpiresIn() == null) {
    expirationRefreshTime = defaultExpirationTimeSeconds;
} else {
    expirationRefreshTime = accessDetails.getExpiresIn;
}
```

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/StandardOauth2AccessTokenProvider.java
```diff
@@ -456,8 +468,10 @@ private void getProperties(ConfigurationContext context) {
     }
 
     private boolean isRefreshRequired() {
+        final long expirationTime = Optional.ofNullable(accessDetails.getExpiresIn()).orElse(defaultExpirationTimeSeconds);
```
exceptionfactory: The `Optional.ofNullable()` approach creates an unnecessary `Optional` wrapper. I recommend replacing this with a standard conditional

exceptionfactory (changes requested):
Thanks for the adjustment @turcsanyip. I recommend avoiding the `Optional.ofNullable()` and functional style in this case, then this looks ready to go.

nifi-extension-bundles/nifi-standard-services/nifi-oauth2-provider-bundle/nifi-oauth2-provider-service/src/main/java/org/apache/nifi/oauth2/StandardOauth2AccessTokenProvider.java
turcsanyip: > The `Optional.ofNullable()` approach creates an unnecessary `Optional` wrapper. I recommend replacing this with a standard conditional

I'm not a big fan of it, but I tend to use it. I don't insist on it at all.
Just to avoid one more round: what do you think about a ternary operator? (no wrapper but still compact)
I'm also fine with the regular `if-else` if that is preferable.

exceptionfactory: Thanks, sure, a ternary is also fine with me, as it also avoids the unnecessary wrapper creation.

PR #9109, opened 2024-07-24
NIFI-13125: Add view for comparing different versions of registered flows

sardell:
# Summary

[NIFI-13125](https://issues.apache.org/jira/browse/NIFI-13125)

### Testing
1. Start NiFi Registry and create a bucket. Import a flow to that bucket. 
2. Open the controller settings modal in the NiFi UI and add a registry client. Point the url to the NiFi Registry url. 
3. Add a Process Group and import the flow you just created in registry. 
4. Make some changes to the flow (can be as simple as renaming a processor). Right click on the process group and select Version > Commit local changes. Save your changes.
5. Repeat step 4 so you have another set of committed changes (so 3 versions total, though you could do more if you’d like).
6. Right click on the process group and select Version > Change version
7. You’ll see a new icon to the left of the Version column in the versions table (see 1st image below). Click on an icon that isn’t the current version. You should see a modal launch listing the differences between the two versions (see 2nd image below). 
8. Try adjusting the version selection dropdowns. If you compare the same version, you should see a message informing you that there are no differences (see 3rd image below).
9. The filter should apply whatever input it has across the 3 columns (component name, change type, and difference).

Screenshot 2024-07-24 at 4 49 06 PM

Screenshot 2024-07-24 at 4 49 19 PM

Screenshot 2024-07-24 at 4 49 24 PM

# Tracking

Please complete the following tracking steps prior to pull request creation.

### Issue Tracking

- [ ] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI) issue created

### Pull Request Tracking

- [ ] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-00000`
- [ ] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-00000`

### Pull Request Formatting

- [ ] Pull Request based on current revision of the `main` branch
- [ ] Pull Request refers to a feature branch with one commit containing changes

# Verification

Please indicate the verification steps performed prior to pull request creation.

### Build

- [ ] Build completed using `mvn clean install -P contrib-check`
  - [ ] JDK 21

### Licensing

- [ ] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [ ] New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [ ] Documentation formatting appears as expected in rendered files

PR #10473, opened 2025-10-28
NIFI-13125 - Flow Version Diff View

pvillard31:
# Summary

NIFI-13125 - Flow Version Diff View

I used #9109 as a starting point and coded something similar for NiFi 2.x. Definitely not a UI expert so my approach may not be the best one. Happy to collaborate with someone more experienced on the UI side.

This feature is particularly useful when a user sees that a versioned process group can be updated to a new version and the user would like to see what are the differences with the current version before upgrading the process group.

Access to the view is made available when listing versions for a versioned process group:

Screenshot 2025-10-28 at 23 10 04

Clicking on the diff icon for the desired version will show the diff between the current version and the selected version

Screenshot 2025-10-28 at 23 11 14

Versions can be selected via the dropdown lists:

Screenshot 2025-10-28 at 23 18 38

Filtering is also possible to filter the content of the table, and the columns can also be sorted alphabetically.

The current feature does not allow comparison across branches but that could be considered as a follow-up improvement if the community sees value into it (I'm not sure I picture a use case where it would be very useful).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/change-version-dialog/change-version-dialog.ts
```diff
@@ -47,16 +49,18 @@ export class ChangeVersionDialog extends CloseOnEscapeDialog {
     private nifiCommon = inject(NiFiCommon);
     private store = inject<Store<CanvasState>>(Store);
 
-    displayedColumns: string[] = ['current', 'version', 'created', 'comments'];
+    displayedColumns: string[] = ['actions', 'current', 'version', 'created', 'comments'];
```
scottyaslan: The actions column should be displayed as the last column in the table listing.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/change-version-dialog/change-version-dialog.html
```diff
@@ -78,6 +78,24 @@ <h2 mat-dialog-title>
                         (matSortChange)="sortData($event)"
                         [matSortActive]="sort.active"
                         [matSortDirection]="sort.direction">
+                        <!-- Actions Column -->
```
scottyaslan: The actions column should be displayed as the last column in the table listing.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/change-version-dialog/change-version-dialog.scss
```diff
@@ -22,6 +22,10 @@
 
     .listing-table {
         table {
+            .mat-column-actions {
+                width: 54px;
+            }
+
```
scottyaslan: actions column is already set for all listings so we don't need this here

```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.html
```diff
@@ -0,0 +1,123 @@
...
+                        [formControl]="filterControl"
+                        placeholder="Filter differences"
+                        data-qa="flow-diff-filter" />
+                </mat-form-field>
+            </div>
+
+            <div class="flow-diff-table flex-1">
+                <div class="flow-diff-loading" *ngIf="isLoading">
+                    <mat-progress-spinner diameter="36" mode="indeterminate"></mat-progress-spinner>
+                </div>
```
scottyaslan: The listings in nifi do not show mat-progress-spinner or any other loading UX.
```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.html
```diff
@@ -0,0 +1,123 @@
...
+
+                <div *ngIf="!isLoading">
+                    <div *ngIf="errorMessage" class="flow-diff-error" data-qa="flow-diff-error">
+                        {{ errorMessage }}
+                    </div>
```
scottyaslan: The listings in nifi do not show errors like this. If there is an error when requesting the diff it should be displayed in a context error banner at the top of the dialog like we do in other cases.


```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.html
```diff
@@ -0,0 +1,123 @@
...
+
+                    <ng-container *ngIf="!errorMessage">
+                        <div *ngIf="noDifferences" class="flow-diff-empty" data-qa="flow-diff-empty">
+                            No differences to display.
+                        </div>
+
+                        <table
```
scottyaslan, line 86: Ths styles for this table do not match the style for other listings throughout the app.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.scss
```diff
@@ -0,0 +1,94 @@
...
+ * limitations under the License.
+ */
+
+@use '@angular/material' as mat;
+
+.flow-diff {
+    @include mat.button-density(-1);
+
+    .flow-diff-message {
+        color: var(--mat-sys-on-surface-variant);
```
scottyaslan: I am curious as to why this is needed? Typography should already be properly styled.

```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.scss
```diff
@@ -0,0 +1,94 @@
...
+
+        .flow-diff-message-title {
+            font-weight: 600;
+            margin-bottom: 4px;
+        }
```
scottyaslan: These styles are set as part of the design system and should not be needed here.

```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.scss
```diff
@@ -0,0 +1,94 @@
...
+                .flow-diff-version {
+                    font-weight: 500;
+                }
+
+                .flow-diff-created {
+                    margin-left: 4px;
+                    color: var(--mat-sys-on-surface-variant);
+                }
+            }
+        }
```
scottyaslan: Again, all typography should be handled by the design system so these should not be needed. Similarly the li styles should already be set by the design system. If we truly need the margins then please update the template to use tailwind declarative styles rather than setting them here in the SASS partial. The reason is that all of the spacing is configurable and handled by tailwiind. This is really powerful as it allows us to update the tailwind config in a single place and the new spacing will be applied application wide.

```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.scss
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    .flow-diff-controls {
+        .mat-mdc-form-field {
+            min-width: 200px;
+        }
```
scottyaslan: This does not seem correct. Form fields should all be correctly styles by the design system and should not need width set.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.scss
```diff
@@ -0,0 +1,94 @@
...
+
+        .flow-diff-filter {
+            flex: 1 1 240px;
+        }
```
scottyaslan: Please use tailwind declaritive styles to apply these rules in the template.

```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.scss
```diff
@@ -0,0 +1,94 @@
...
+        .flow-diff-empty,
+        .flow-diff-error {
+            padding: 12px 0;
+            color: var(--mat-sys-on-surface-variant);
+        }
+
+        .flow-diff-error {
+            color: var(--mat-sys-error);
+        }
+    }
```
scottyaslan: Please use tailwind declarative styles to apply these rules in the template if they are actually needed but I don't think they are.
```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+    templateUrl: './flow-diff-dialog.html',
+    styleUrl: './flow-diff-dialog.scss'
+})
+export class FlowDiffDialog extends CloseOnEscapeDialog implements AfterViewInit {
+    private data = inject<FlowDiffDialogData>(MAT_DIALOG_DATA);
+    private registryService = inject(RegistryService);
+    private destroyRef = inject(DestroyRef);
+    private formBuilder = inject(FormBuilder);
+    private nifiCommon = inject(NiFiCommon);
+    private store = inject<Store<CanvasState>>(Store);
```
scottyaslan: We should not be injecting the store in a dialog. Data should be passed into the dialog when opened via MAT_DIALOG_DATA.

```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+    formatVersionOption(version: string): string {
+        const metadata = this.versionMetadataByVersion.get(version);
+        const formattedVersion =
+            version.length > 5 ? `${version.substring(0, 5)}...` : version;
+        const created = metadata ? this.formatTimestamp(metadata) : undefined;
+        return created ? `${formattedVersion} (${created})` : formattedVersion;
+    }
+
+    ngAfterViewInit(): void {
+        // Sorting handled automatically when MatSort is available via the ViewChild setter
```
scottyaslan: I haven't really dug in here yet but it seems sorting is handled differently for this listing then other listings throughout nifi. I am not saying this is incorrect but it is a different pattern so at the very least it would need some testing and verification. It would be nice it we could just update this to follow the pattern used in other listings. We should not need to use a ViewChild to sort an angular material table.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.html
```diff
@@ -0,0 +1,123 @@
...
+                            mat-table
+                            *ngIf="dataSource.data.length > 0"
```
scottyaslan, line 88:
```suggestion
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.html
```diff
@@ -0,0 +1,123 @@
...
+</h2>
+<div class="flow-diff">
+    <mat-dialog-content>
+        <div class="dialog-content flex flex-col gap-y-4">
+            <div
+                class="flow-diff-message"
+                data-qa="flow-diff-message"
+                *ngIf="comparisonSummary.length > 0">
+                <div class="flow-diff-message-title">Comparing versions:</div>
+                <ul class="flow-diff-message-list">
```
scottyaslan: An unordered list is not the correct way to display this information. Please follow the label/value UX used throughout the application to display values. I believe we have an example in the change version dialog as well.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+export class FlowDiffDialog extends CloseOnEscapeDialog implements AfterViewInit {
+    private data = inject<FlowDiffDialogData>(MAT_DIALOG_DATA);
+    private registryService = inject(RegistryService);
+    private destroyRef = inject(DestroyRef);
+    private formBuilder = inject(FormBuilder);
+    private nifiCommon = inject(NiFiCommon);
+    private store = inject<Store<CanvasState>>(Store);
+    private timeOffset = this.store.selectSignal(selectTimeOffset);
+
+    @ViewChild(MatSort)
```
scottyaslan: You can follow how the change version dialog implemented sorting instead of using a ViewChild here.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+
+        const now = new Date();
+        const userOffset = now.getTimezoneOffset() * 60 * 1000;
+        const date = new Date(metadata.timestamp + userOffset + this.getTimezoneOffset());
+        return this.nifiCommon.formatDateTime(date);
+    }
+
+    private getTimezoneOffset(): number {
+        return this.timeOffset() || 0;
+    }
```
scottyaslan: I don't think we typically format timestamps like this. Is there a reason to here? I now see that we do have two implementations of:
```
formatTimestamp(flowVersion: VersionedFlowSnapshotMetadata) {
        // get the current user time to properly convert the server time
        const now: Date = new Date();

        // convert the user offset to millis
        const userTimeOffset: number = now.getTimezoneOffset() * 60 * 1000;

        // create the proper date by adjusting by the offsets
        const date: Date = new Date(flowVersion.timestamp + userTimeOffset + this.timeOffset);
        return this.nifiCommon.formatDateTime(date);
    }
```

One in the select-flow-version.componet.ts and one in the import-from-registry.component.ts. If we really have a need for this formatting we should have a single implementation in a shared area. A new formatTimestamp Angular pipe in https://github.com/apache/nifi/tree/main/nifi-frontend/src/main/frontend/libs/shared/src/pipes would work well for this use case but I think you will find that formatting timestamps has lots or nuanced considerations. we used to have a library for this but it wasn't well maintained and was deprecated if that is any clue as to LOE.

scottyaslan (changes requested):
Thanks for this contribution. I have left some comments. I am happy to review and help iterate here and continue to improve the UX for this feature.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/flow-designer/ui/canvas/items/flow/flow-diff-dialog/flow-diff-dialog.ts
scottyaslan: IMO we can just leave the timestamp unformatted.

pvillard31:
Thanks for all the feedback @scottyaslan - highly appreciated! I have (hopefully) made the changes you had in mind. The only thing that is not ok is the timestamp in the dropdown lists when selecting versions. Will figure it out tomorrow. It looks so much better now, thanks for all of the review comments!

Screenshot 2025-10-30 at 22 56 03
Screenshot 2025-10-30 at 22 56 33
Screenshot 2025-10-30 at 22 57 05

pvillard31:
Addressed the timestamp issue.

Screenshot 2025-10-31 at 12 20 18

fingoltin:
Would love to see this feature!

pvillard31:
Closing. See #11474.

PR #10538, opened 2025-11-17
NIFI-15226 Allow ConsumeKafka to use static partition mapping

tpalfy:
# Summary

[NIFI-15226](https://issues.apache.org/jira/browse/NIFI-15226)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/pom.xml
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/pom.xml
@@ -63,12 +63,6 @@
             <artifactId>nifi-mock-record-utils</artifactId>
             <scope>test</scope>
         </dependency>
-<!--        <dependency>-->
-<!--            <groupId>org.apache.nifi</groupId>-->
-<!--            <artifactId>nifi-framework-nar-utils</artifactId>-->
-<!--            <version>2.7.0-SNAPSHOT</version>-->
-<!--            <scope>test</scope>-->
-<!--        </dependency>-->
         <dependency>
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-record-serialization-services</artifactId>
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/pom.xml
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/pom.xml
@@ -63,12 +63,12 @@
             <artifactId>nifi-mock-record-utils</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.apache.nifi</groupId>
-            <artifactId>nifi-framework-nar-utils</artifactId>
-            <version>2.7.0-SNAPSHOT</version>
-            <scope>test</scope>
-        </dependency>
+<!--        <dependency>-->
+<!--            <groupId>org.apache.nifi</groupId>-->
+<!--            <artifactId>nifi-framework-nar-utils</artifactId>-->
+<!--            <version>2.7.0-SNAPSHOT</version>-->
+<!--            <scope>test</scope>-->
+<!--        </dependency>-->
         <dependency>
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-record-serialization-services</artifactId>
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/pom.xml
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/pom.xml
@@ -65,6 +65,12 @@
         </dependency>
         <dependency>
             <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-nar-utils</artifactId>
+            <version>2.7.0-SNAPSHOT</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-record-serialization-services</artifactId>
             <version>2.8.0-SNAPSHOT</version>
             <scope>test</scope>
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
@@ -446,7 +446,8 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
 
                 throw new ProcessException(
                         String.format(
-                                "Illegal Partition Assignment: There are %d partition%s statically assigned using the %s.* property names, but the Kafka topic(s) have %d partition%s",
+                                "Illegal Partition Assignment: There %s %d partition%s statically assigned using the %s.* property names, but the Kafka topic(s) have %d partition%s",
+                                numAssignedPartitions == 1 ? "is" : "are",
                                 numAssignedPartitions,
                                 numAssignedPartitions == 1 ? "" : "s",
                                 PARTITIONS_PROPERTY_PREFIX,
@@ -686,14 +687,14 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
                 final String thisTopic = topicAndPartitionStates.getKey();
                 final int partitionsThisTopic = topicAndPartitionStates.getValue().size();
                 if (partitionsEachTopic != 0 && partitionsThisTopic != partitionsEachTopic) {
-                    String.format(
+                    throw new IllegalStateException(String.format(
                             "The topics do not have the same number of partitions. Topic '%s' has %d partition%s, while others have %d partition%s",
                             thisTopic,
                             partitionsThisTopic,
                             partitionsThisTopic == 1 ? "" : "s",
                             partitionsEachTopic,
                             partitionsEachTopic == 1 ? "" : "s"
-                    );
+                    ));
                 }
 
                 partitionsEachTopic = partitionsThisTopic;
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
@@ -499,9 +499,9 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
 
         final PollingContext partitionedPollingContext = consumerServiceToPartitionedPollingContext.get(consumerService);
 
-        final PollingContext pollingContext = partitionedPollingContext == null ?
-                this.pollingContext :
-                partitionedPollingContext;
+        final PollingContext pollingContext = partitionedPollingContext == null
+                ? this.pollingContext
+                : partitionedPollingContext;
 
         final long maxUncommittedMillis = context.getProperty(MAX_UNCOMMITTED_TIME).asTimePeriod(TimeUnit.MILLISECONDS);
         final long stopTime = System.currentTimeMillis() + maxUncommittedMillis;
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
@@ -457,10 +457,11 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
                 throw new ProcessException("Could not determine localhost's hostname", uhe);
             }
 
+            availablePartitionedPollingContexts.clear();
             for (int partition : assignedPartitions) {
-                final PollingContext partitionedPollingContext = createPollingContext(context, partition);
-                final KafkaConsumerService partitionedConsumerService = obtainPartitionedConsumerService(partitionedPollingContext);
-                consumerServices.add(partitionedConsumerService);
+                PollingContext partitionedPollingContext = createPollingContext(context, partition);
+
+                availablePartitionedPollingContexts.add(partitionedPollingContext);
             }
         }
     }
@@ -636,6 +637,8 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
     }
 
     private KafkaConsumerService getConsumerService(final ProcessContext context) {
+        recreatePartitionedConsumerServices();
+
         final KafkaConsumerService consumerService = consumerServices.poll();
         if (consumerService != null) {
             return consumerService;
@@ -644,18 +647,8 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
         final boolean isExplicitPartitionMapping = ConsumerPartitionsUtil.isPartitionAssignmentExplicit(context.getAllProperties());
 
         if (isExplicitPartitionMapping) {
-            final PollingContext partitionedPollingContext = availablePartitionedPollingContexts.poll();
-
-            if (partitionedPollingContext == null) {
-                getLogger().trace("No Partitioned Kafka Consumer Service available, all specified partitions are being consumed from.");
-                return null;
-            }
-
-            getLogger().info("No Partitioned Kafka Consumer Service available; creating a new one.");
-
-            final KafkaConsumerService partitionedConsumerService = obtainPartitionedConsumerService(partitionedPollingContext);
-
-            return partitionedConsumerService;
+            getLogger().trace("No Partitioned Kafka Consumer Service available, all specified partitions are being consumed from.");
+            return null;
         } else {
             final int activeCount = activeConsumerCount.incrementAndGet();
             if (activeCount > getMaxConsumerCount()) {
@@ -694,12 +687,17 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
         return partitionsEachTopic;
     }
 
-    private KafkaConsumerService obtainPartitionedConsumerService(PollingContext partitionedPollingContext) {
-        final KafkaConsumerService partitionedConsumerService = connectionService.getConsumerService(partitionedPollingContext);
+    private void recreatePartitionedConsumerServices() {
+        PollingContext partitionedPollingContext;
+        while ((partitionedPollingContext = availablePartitionedPollingContexts.poll()) != null) {
+            getLogger().info("Creating new Partitioned Kafka Consumer Service.");
 
-        consumerServiceToPartitionedPollingContext.put(partitionedConsumerService, partitionedPollingContext);
+            final KafkaConsumerService partitionedConsumerService = connectionService.getConsumerService(partitionedPollingContext);
 
-        return partitionedConsumerService;
+            consumerServiceToPartitionedPollingContext.put(partitionedConsumerService, partitionedPollingContext);
+
+            consumerServices.offer(partitionedConsumerService);
+        }
     }
 
     private int getMaxConsumerCount() {
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
@@ -324,6 +324,7 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
     private static final Set<Relationship> SUCCESS_RELATIONSHIP = Set.of(SUCCESS);
     private static final Set<Relationship> SUCCESS_FAILURE_RELATIONSHIPS = Set.of(SUCCESS, PARSE_FAILURE);
 
+    private volatile KafkaConnectionService connectionService;
     private volatile Charset headerEncoding;
     private volatile Pattern headerNamePattern;
     private volatile ProcessingStrategy processingStrategy;
@@ -341,6 +342,7 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
     private final Queue<KafkaConsumerService> consumerServices = new LinkedBlockingQueue<>();
     private final AtomicInteger activeConsumerCount = new AtomicInteger();
 
+    private final Queue<PollingContext> availablePartitionedPollingContexts = new LinkedBlockingQueue<>();
     private final Map<KafkaConsumerService, PollingContext> consumerServiceToPartitionedPollingContext = new HashMap<>();
 
     @Override
@@ -407,6 +409,8 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
 
     @OnScheduled
     public void onScheduled(final ProcessContext context) {
+        connectionService = context.getProperty(CONNECTION_SERVICE).asControllerService(KafkaConnectionService.class);
+
         pollingContext = createPollingContext(context);
         headerEncoding = Charset.forName(context.getProperty(HEADER_ENCODING).getValue());
 
@@ -424,7 +428,7 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
         keyFormat = (outputStrategy == OutputStrategy.USE_WRAPPER || outputStrategy == OutputStrategy.INJECT_METADATA)
                 ? context.getProperty(KEY_FORMAT).asAllowableValue(KeyFormat.class)
                 : KeyFormat.BYTE_ARRAY;
-        brokerUri = context.getProperty(CONNECTION_SERVICE).asControllerService(KafkaConnectionService.class).getBrokerUri();
+        brokerUri = connectionService.getBrokerUri();
         maxConsumerCount = context.getMaxConcurrentTasks();
         activeConsumerCount.set(0);
 
@@ -432,6 +436,37 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
         maxUncommittedSizeConfigured = maxUncommittedSizeProperty.isSet();
         if (maxUncommittedSizeConfigured) {
             maxUncommittedSize = maxUncommittedSizeProperty.asDataSize(DataUnit.B).longValue();
+        }
+
+        if (ConsumerPartitionsUtil.isPartitionAssignmentExplicit(context.getAllProperties())) {
+            final int numAssignedPartitions = ConsumerPartitionsUtil.getPartitionAssignmentCount(context.getAllProperties());
+            final int partitionCount = getPartitionCount(connectionService);
+
+            if (partitionCount != numAssignedPartitions) {
+                context.yield();
+
+                KafkaConsumerService service;
+                while ((service = consumerServices.poll()) != null) {
+                    close(service, "Not all partitions are assigned");
+                }
+
+                throw new ProcessException("Illegal Partition Assignment: There are "
+                        + numAssignedPartitions + " partitions statically assigned using the " + PARTITIONS_PROPERTY_PREFIX + ".* property names,"
+                        + " but the Kafka topic(s) have " + partitionCount + " partitions");
+            }
+
+            final int[] assignedPartitions;
+            try {
+                assignedPartitions = ConsumerPartitionsUtil.getPartitionsForHost(context.getAllProperties(), getLogger());
+            } catch (final UnknownHostException uhe) {
+                throw new ProcessException("Could not determine localhost's hostname", uhe);
+            }
+
+            for (int partition : assignedPartitions) {
+                PollingContext partitionedPollingContext = createPollingContext(context, partition);
+
+                availablePartitionedPollingContexts.add(partitionedPollingContext);
+            }
         }
     }
 
@@ -443,6 +478,9 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
         while ((service = consumerServices.poll()) != null) {
             close(service, "Processor stopped");
         }
+
+        availablePartitionedPollingContexts.clear();
+        consumerServiceToPartitionedPollingContext.clear();
     }
 
 
@@ -559,6 +597,11 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
         try {
             consumerService.close();
             activeConsumerCount.decrementAndGet();
+
+            final PollingContext partitionedPollingContext = consumerServiceToPartitionedPollingContext.remove(consumerService);
+            if (partitionedPollingContext != null) {
+                availablePartitionedPollingContexts.add(partitionedPollingContext);
+            }
         } catch (final IOException ioe) {
             getLogger().warn("Failed to close Kafka Consumer Service", ioe);
         }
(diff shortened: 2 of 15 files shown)


nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/pom.xml
```diff
@@ -63,6 +63,12 @@
             <artifactId>nifi-mock-record-utils</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-nar-utils</artifactId>
```
exceptionfactory: This dependency appears to be unused, can it be removed?

tpalfy: `org.apache.nifi.mock.MockComponentLogger` is referenced in `TestConsumerPartitionsUtil`

exceptionfactory: Thanks for clarifying. In that case, this dependency should be removed and `MockComponentLogger` should be replaced with `MockComponentLog` from `nifi-mock` to avoid referencing framework modules in extension modules.

exceptionfactory (changes requested):
The initial build fails on the following tests:

```
Error:    ConsumeKafkaTest.testVerifyFailed:116 » NullPointer Cannot invoke "org.apache.nifi.kafka.service.api.consumer.PollingContext.getTopics()" because "this.pollingContext" is null
Error:    ConsumeKafkaTest.testVerifySuccessful:99 » NullPointer Cannot invoke "org.apache.nifi.kafka.service.api.consumer.PollingContext.getTopics()" because "this.pollingContext" is null
```

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/pom.xml
tpalfy: I preserved this class from the older change. But actually we don't need either. We can just mock the logger with Mockito.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -540,6 +603,47 @@ private KafkaConsumerService getConsumerService(final ProcessContext context) {
             return consumerService;
         }
 
+        final KafkaConnectionService connectionService = context.getProperty(CONNECTION_SERVICE).asControllerService(KafkaConnectionService.class);
+
+        consumerServiceToPartitionedPollingContext.clear();
+
+        final boolean isExplicitPartitionMapping = ConsumerPartitionsUtil.isPartitionAssignmentExplicit(context.getAllProperties());
+
+        if (isExplicitPartitionMapping) {
```
turcsanyip: The assignment of the partitions must happen exactly once after the processor is started.
This code is called from `onTrigger()` which can cause concurrency issues when Concurrent Tasks > 1:

1. Multiple threads are trying to initialize the consumers in parallel
2. If there are configured more threads than partitions, then a thread can reinitalize the assignment if there is no available consumer in the pool

Ideally, the assigment should happen in `onScheduled()`. Isn't it possible?
Also, `consumerServiceToPartitionedPollingContext` should be cleared in `onStopped()`.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -548,10 +652,34 @@ private KafkaConsumerService getConsumerService(final ProcessContext context) {
         }
 
         getLogger().info("No Kafka Consumer Service available; creating a new one. Active count: {}", activeCount);
-        final KafkaConnectionService connectionService = context.getProperty(CONNECTION_SERVICE).asControllerService(KafkaConnectionService.class);
         return connectionService.getConsumerService(pollingContext);
     }
 
+    public int getPartitionCount(final KafkaConnectionService connectionService) {
```
turcsanyip: Minor:
```suggestion
    private int getPartitionCount(final KafkaConnectionService connectionService) {
```

turcsanyip (changes requested):

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
turcsanyip, line 1: In case of static partition mapping, the consumers should be created eagerly for each partition in `availablePartitionedPollingContexts` (similar to [recreateAssignedConsumers()](https://github.com/apache/nifi/blob/support/nifi-1.x/nifi-nar-bundles/nifi-kafka-bundle/nifi-kafka-2-6-processors/src/main/java/org/apache/nifi/processors/kafka/pubsub/ConsumerPool.java#L502) in the old 2.6 implementation).

Without that, a consumer is created for the first partition, and additional consumers are created for the other partitions only when parallel task execution triggers it. This is because a consumer is always available in the pool until multiple threads request consumers in parallel (which is not guaranteed to happen, e.g. Concurrent Tasks = 1).

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -334,6 +342,9 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
     private final Queue<KafkaConsumerService> consumerServices = new LinkedBlockingQueue<>();
     private final AtomicInteger activeConsumerCount = new AtomicInteger();
 
+    private final Queue<PollingContext> availablePartitionedPollingContexts = new LinkedBlockingQueue<>();
+    private final Map<KafkaConsumerService, PollingContext> consumerServiceToPartitionedPollingContext = new HashMap<>();
```
turcsanyip: The map can be accessed by parallel threads so it should be `ConcurrentHashMap` or similar.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -379,6 +437,37 @@ public void onScheduled(final ProcessContext context) {
...
+            final int numAssignedPartitions = ConsumerPartitionsUtil.getPartitionAssignmentCount(context.getAllProperties());
+            final int partitionCount = getPartitionCount(connectionService);
+
+            if (partitionCount != numAssignedPartitions) {
+                context.yield();
+
+                KafkaConsumerService service;
+                while ((service = consumerServices.poll()) != null) {
+                    close(service, "Not all partitions are assigned");
+                }
```
turcsanyip: I think it can never happen because no consumers have been created so far.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-service-shared/src/main/java/org/apache/nifi/kafka/service/Kafka3ConnectionService.java
```diff
@@ -238,6 +240,14 @@ public KafkaConsumerService getConsumerService(final PollingContext pollingConte
...
         final Consumer<byte[], byte[]> consumer = new KafkaConsumer<>(properties, deserializer, deserializer);
 
+        Integer partition = pollingContext.getPartition();
+        if (partition != null) {
+            List<TopicPartition> topicPartitions = pollingContext.getTopics().stream()
+                    .map(topic -> new TopicPartition(topic, partition))
+                    .collect(Collectors.toList());
+            consumer.assign(topicPartitions);
+        }
+
```
turcsanyip: I suggest moving it to `Kafka3ConsumerService`'s constructor. That way, the assign/subscribe logic would be more straightforward (a single `if-else` instead of the separate `if` statements at different points in the code).
The partition id can be passed in the `Subscription` parameter.

turcsanyip (changes requested):

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-service-shared/src/main/java/org/apache/nifi/kafka/service/consumer/Subscription.java
```diff
@@ -31,19 +31,22 @@ public class Subscription {
 
     private final String groupId;
     private final Collection<String> topics;
+    private final Integer partition;
     private final Pattern topicPattern;
     private final AutoOffsetReset autoOffsetReset;
 
-    public Subscription(final String groupId, final Collection<String> topics, final AutoOffsetReset autoOffsetReset) {
+    public Subscription(final String groupId, final Integer partition, final Collection<String> topics, final AutoOffsetReset autoOffsetReset) {
```
exceptionfactory: Very minor, but I recommend placing `partition` after `topics` to align with general hierarchy:

```suggestion
    public Subscription(final String groupId, final Collection<String> topics, final Integer partition, final AutoOffsetReset autoOffsetReset) {
```

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/test/java/org/apache/nifi/kafka/processors/consumer/TestConsumerPartitionsUtil.java
```diff
@@ -0,0 +1,120 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+
+public class TestConsumerPartitionsUtil {
```
exceptionfactory: Minor note, the `public` modifiers on the class and method level are not necessary for JUnit 5. Since this is a new test class, recommend removing them.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/test/java/org/apache/nifi/kafka/processors/consumer/TestConsumerPartitionsUtil.java
```diff
@@ -0,0 +1,120 @@
...
+    private final ComponentLog logger = mock();
+    private String hostname;
+
+    @BeforeEach
+    public void setup() throws UnknownHostException {
+        hostname = InetAddress.getLocalHost().getHostName();
+    }
+
+    @Test
+    public void testNoPartitionAssignments() throws UnknownHostException {
```
exceptionfactory: `UnknownHostException` does not appear to be thrown in this and other methods.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/test/java/org/apache/nifi/kafka/processors/consumer/TestConsumerPartitionsUtil.java
```diff
@@ -0,0 +1,120 @@
...
+        final Map<String, String> properties = Collections.singletonMap("key", "value");
```
exceptionfactory:
```suggestion
        final Map<String, String> properties = Map.of("key", "value");
```

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -617,6 +765,10 @@ private void processInputFlowFile(final ProcessSession session, final OffsetTrac
     }
 
     private PollingContext createPollingContext(final ProcessContext context) {
+        return createPollingContext(context, null);
+    }
+
+    private PollingContext createPollingContext(final ProcessContext context, Integer partition) {
```
exceptionfactory:
```suggestion
    private PollingContext createPollingContext(final ProcessContext context, final Integer partition) {
```

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -513,43 +609,95 @@ public List<ConfigVerificationResult> verify(final ProcessContext context, final
...
+            getLogger().warn("Couldn't close KafkaConsumerService after partition assignment check.", e);
+        }
+
+        return partitionsEachTopic;
+    }
+
+    private void recreatePartitionedConsumerServices() {
+        PollingContext partitionedPollingContext;
+        while ((partitionedPollingContext = availablePartitionedPollingContexts.poll()) != null) {
+            getLogger().info("Creating new Partitioned Kafka Consumer Service.");
```
exceptionfactory: Some identifying information, such as the Partition, should be included in the log message.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -513,43 +609,95 @@ public List<ConfigVerificationResult> verify(final ProcessContext context, final
...
+            for (List<PartitionState> partitionStatesForTopic : topicToPartitionStates.values()) {
+                final int partitionsThisTopic = partitionStatesForTopic.size();
+                if (partitionsEachTopic != 0 && partitionsThisTopic != partitionsEachTopic) {
+                    throw new IllegalStateException("The specific topic names do not have the same number of partitions");
+                }
+
+                partitionsEachTopic = partitionsThisTopic;
+            }
+        } catch (IOException e) {
+            getLogger().warn("Couldn't close KafkaConsumerService after partition assignment check.", e);
```
exceptionfactory: It is best to avoid conjunctions and avoid ending with `.` characters in messages:
```suggestion
            getLogger().warn("Failed to close KafkaConsumerService after partition assignment check", e);
```

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -513,43 +609,95 @@ public List<ConfigVerificationResult> verify(final ProcessContext context, final
...
+            return -1;
+        }
+
+        int partitionsEachTopic = 0;
+        try (KafkaConsumerService kafkaConsumerService = connectionService.getConsumerService(this.pollingContext)) {
+            Map<String, List<PartitionState>> topicToPartitionStates = kafkaConsumerService.getPartitionStatesByTopic();
+            for (List<PartitionState> partitionStatesForTopic : topicToPartitionStates.values()) {
+                final int partitionsThisTopic = partitionStatesForTopic.size();
+                if (partitionsEachTopic != 0 && partitionsThisTopic != partitionsEachTopic) {
+                    throw new IllegalStateException("The specific topic names do not have the same number of partitions");
```
exceptionfactory: The topic names should be included in the message for troubleshooting.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -513,43 +609,95 @@ public List<ConfigVerificationResult> verify(final ProcessContext context, final
...
+                            .explanation(String.format("Partitions [%d] found for Topics %s", partitionStates.size(), pollingContext.getTopics()));
+                } catch (final Exception e) {
+                    getLogger().error("Topics {} Partition verification failed", pollingContext.getTopics(), e);
+                    verificationPartitions
+                            .outcome(ConfigVerificationResult.Outcome.FAILED)
+                            .explanation(String.format("Topics %s Partition access failed: %s", pollingContext.getTopics(), e));
+                }
+                verificationResults.add(verificationPartitions.build());
+            } catch (IOException e) {
+                getLogger().warn("Couldn't close KafkaConsumerService after verification.", e);
```
exceptionfactory:
```suggestion
                getLogger().warn("Failed to close KafkaConsumerService after verification", e);
```

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -513,43 +609,95 @@ public List<ConfigVerificationResult> verify(final ProcessContext context, final
...
+                final ConfigVerificationResult.Builder verificationPartitions = new ConfigVerificationResult.Builder()
+                        .verificationStepName("Verify Topic Partitions");
+
+                try {
+                    final List<PartitionState> partitionStates = consumerService.getPartitionStatesByTopic().values().stream().findFirst().orElse(Collections.emptyList());
+                    verificationPartitions
+                            .outcome(ConfigVerificationResult.Outcome.SUCCESSFUL)
+                            .explanation(String.format("Partitions [%d] found for Topics %s", partitionStates.size(), pollingContext.getTopics()));
+                } catch (final Exception e) {
+                    getLogger().error("Topics {} Partition verification failed", pollingContext.getTopics(), e);
```
exceptionfactory: This is carried over from the previous implementation, but on review, the Verification Logger instance should be used instead of `getLogger()`.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -389,12 +474,18 @@ public void onStopped() {
...
+
+        availablePartitionedPollingContexts.clear();
+        consumerServiceToPartitionedPollingContext.clear();
     }
 
 
     @Override
     public void onTrigger(final ProcessContext context, final ProcessSession session) {
         final KafkaConsumerService consumerService = getConsumerService(context);
+        final PollingContext pollingContext = Optional.ofNullable(consumerServiceToPartitionedPollingContext.get(consumerService))
```
exceptionfactory: Creating an `Optional` wrapper is not necessary, and does not add sufficient value for running on every invocation of `onTrigger`, so I recommend replacing with a standard conditional.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -379,6 +437,33 @@ public void onScheduled(final ProcessContext context) {
...
+            final int numAssignedPartitions = ConsumerPartitionsUtil.getPartitionAssignmentCount(context.getAllProperties());
+            final int partitionCount = getPartitionCount(connectionService);
+
+            if (partitionCount != numAssignedPartitions) {
+                context.yield();
+
+                throw new ProcessException("Illegal Partition Assignment: There are "
+                        + numAssignedPartitions + " partitions statically assigned using the " + PARTITIONS_PROPERTY_PREFIX + ".* property names,"
+                        + " but the Kafka topic(s) have " + partitionCount + " partitions");
+            }
```
exceptionfactory: Recommend using formatted strings instead of concatenation for this message.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -379,6 +437,33 @@ public void onScheduled(final ProcessContext context) {
...
+
+            final int[] assignedPartitions;
+            try {
+                assignedPartitions = ConsumerPartitionsUtil.getPartitionsForHost(context.getAllProperties(), getLogger());
+            } catch (final UnknownHostException uhe) {
+                throw new ProcessException("Could not determine localhost's hostname", uhe);
```
exceptionfactory:
```suggestion
                throw new ProcessException("Failed to resolve local host address", uhe);
```

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/ConsumerPartitionsUtil.java
```diff
@@ -0,0 +1,209 @@
...
+
+        return new ValidationResult.Builder().valid(true).build();
+    }
+
+    public static boolean isPartitionAssignmentExplicit(final Map<String, String> properties) {
+        final Map<String, String> hostnameToPartitionMapping = mapHostnamesToPartitionStrings(properties);
+        return !hostnameToPartitionMapping.isEmpty();
+    }
+
+    public static int getPartitionAssignmentCount(final Map<String, String> properties) {
```
exceptionfactory, line 139: Recommend moving all public methods before all private methods in this class.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/ConsumerPartitionsUtil.java
```diff
@@ -0,0 +1,209 @@
...
+        final Map<String, String> hostnameToPartitionString = mapHostnamesToPartitionStrings(properties);
+        final Map<String, int[]> partitionsByHost = mapPartitionValueToIntArrays(hostnameToPartitionString);
+
+        if (partitionsByHost.isEmpty()) {
+            // Explicit partitioning is not enabled.
+            logger.debug("No explicit Consumer Partitions have been declared.");
+            return null;
+        }
+
+        logger.info("Found the following mapping of hosts to partitions: {}", hostnameToPartitionString);
```
exceptionfactory: This seems more appropriate as a debug level message.

exceptionfactory (changes requested):
Thanks for the work on this @tpalfy. I noted a handful of minor recommendations, and plan to take a closer look at some of the implementation details.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/test/java/org/apache/nifi/kafka/processors/consumer/TestConsumerPartitionsUtil.java
tpalfy: `ConsumerPartitionsUtil.getPartitionsForHost` throws it.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -502,6 +604,11 @@ private void close(final KafkaConsumerService consumerService, final String reas
         try {
             consumerService.close();
             activeConsumerCount.decrementAndGet();
+
+            final PollingContext partitionedPollingContext = consumerServiceToPartitionedPollingContext.remove(consumerService);
+            if (partitionedPollingContext != null) {
+                availablePartitionedPollingContexts.add(partitionedPollingContext);
+            }
```
turcsanyip, line 613: When closing the consumer, the unowned `PollingContext` should be returned to the pool in a `finally` block to ensure it will not be lost if an exception occurs during close.

turcsanyip (changes requested):
@tpalfy Tested with different scenarios, including fewer or more concurrent tasks than partitions, and it works as expected.

Added one minor comment inline and one more thing here: There is some legacy code referencing partition properties in [DynamicPropertyValidator](https://github.com/apache/nifi/blob/2e09ce8a30c7622aaf50053bbd27be7f4b975f5b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-shared/src/main/java/org/apache/nifi/kafka/shared/validation/DynamicPropertyValidator.java#L51-L52). Please clean this up, as the partition dynamic properties are not applied to the Kafka controller service.

turcsanyip (approved):
Thanks for the latest changes @tpalfy.

+1 from my side.

exceptionfactory:
Thanks for the updates @tpalfy and thanks for the review @turcsanyip, I will follow up and also review the latest changes soon.

exceptionfactory (changes requested):
Thanks for making the adjustments @tpalfy, and thanks for the review @turcsanyip.

On a more detailed review, I am concerned about the level of complexity introduced with these changes. The `ConsumerPartitionsUtil` is very complex with implementation details based on current host and dynamic properties. The addition of partition-based dynamic property names, plus host names, makes the configuration non-portable between NiFi installations. This complexity is also evident in the handling of `KafkaConsumerService` instances.

For these reasons, I am not supportive of the current proposed set of changes.

The historical implementation grew very complex and very difficult to maintain, so considering other options would be helpful.

One option that comes to mind is a separate Processor, named something like `ConsumeKafkaPartition`. That would clearly communicate the purpose, allowing for more focused logic. I would not attempt to extend the current `ConsumeKafka` Processor, so copying some the existing implementation would be reasonable. As much of the handling logic is in other classes, there should be some code reuse opportunities, without a subclass.

I'm open to other options that would avoid introducing this level of complexity to the current Processor.

turcsanyip:
Thanks for sharing your thoughts and concerns @exceptionfactory.

I agree that `ConsumerPartitionsUtil` is complex. However, it is worth noting that it was directly reused from the NiFi 1.x implementation where it was a single commit that worked properly. So I would not consider it legacy code that was getting more and more complicated over time. I also reviewed it in this round and I did not find how it could be simplified / changed. This brings us to your second concern about portability. The partitions need to be bound to a NiFi node and hostnames seem the only way to do it. So even if a separate processor were added, it would still have the portability issue.

Considering that the change does not affect the existing users of the processor (no config change, portability issue does not arise if the feature is not used), I do not think it would be worth adding a new processor that requires duplicated code leading to additional maintenance effort and issues going forward.

As far as I can recall, the original plan of migrating to Kafka 3 processors was to implement the core functionality in the first round and add the extra features later. Based on this, and considering that a separate implementation would also introduce complexity and non-portability issues via `ConsumerPartitionsUtil`, I believe we could add this feature to the current `ConsumeKafka` processor.

exceptionfactory:
Thanks for the thoughtful reply @turcsanyip.

Although this capability and general approach was implemented in NiFi 1, part of the purpose of refactoring Kafka support was to provide better decoupling of features. The majority of that separation comes with the Kafka Connection Service, and it is good to see that these proposed changes have minimal impact on that Controller Service interface surface.

With that being said, the proposed changes do introduce significant logic into `ConsumeKafka` that is specific to this particular capability. Although it does not impact existing configuration, it does increase the maintenance of the `ConsumeKafka` implementation in general. For that reason, I would much rather see this particular use case decoupled into its own Processor.

A separate Processor also has benefits from a flow design perspective, as it provides a clearer distinction on the intended use case of static partition assignment to individual NiFi nodes. Given that such a use case is less common, keeping it distinct from the regular `ConsumeKafka` Processor is helpful from both a flow design and component implementation perspective. From an implement perspective, it also provides the opportunity to define more of the implementation in the Processor itself, versus in a separate Util class. Of course these are tradeoffs, but I think the benefits outweigh the costs.

tpalfy:
@exceptionfactory Can you please provide some objective reasoning for considering the ConsumerPartitionsUtil too complex? Here are my objective reasoning to argue against it:
1. 210 rows. Not a big class by any measure.
2. 4 public methods and 4 private methods. This is not out of the ordinary for any util class.
3. Only java.net and java.util imports, apart from org.apache.nifi.components.ValidationResult and org.apache.nifi.logging.ComponentLog. The logic in it is exceptionally well contained.
4. In NiFi 1.x the same class had 1 change since it's inception in 2020, and that was about logging. The class is exceptionally stable and isolated.
5. I found only 2 changes related to static partition mapping in the old code. The logic implemented in this PR _intentionally_ follows that implementation for the very reason to minimize risk.

Also there's an obvious contradiction in adding a separate processor. Code duplication would mean that basically all the changes done to ConsumeKafka would need to be carried over to the new class. Now _that_ would be a maintenance nightmare.

According to the original Jira https://issues.apache.org/jira/browse/NIFI-11259, it was agreed upon that full compatibility will be achieved by adding this feature later. There is a Jira (still open) that proves this:https://issues.apache.org/jira/browse/NIFI-13654
That is also an objective fact.

Here's the implementation as agreed upon, following the original as closely as possible, which was proven to be working and stable by the previous change history. Can't see how we can get better than that.

Unless more adequate reasons are provided to deny this change, this should move forward.

joewitt:
@tpalfy I think we can all agree @exceptionfactory has done amazing work increasing quality and decreasing technical debt.  We need more of what he does - not less.  He is also far and away the most prolific reviewer which we also need more people doing.

Declaring his or any reviewers response as inadequate doesnt help your case. Stating his feedback isnt objective doesnt help either.

tpalfy:
@joewitt Thank you for the input. I would be interested in your opinion as well about the PR itself.

As I see, the original implementation being too complex and hard to maintain (which was the original argument) is not supported by the change history. It did need 2 fairly serious adjustments, but after that the end result remained stable and correct for many years without the need for change.
Maybe I'm missing something here, but I was trying to be very careful to make sure this change works properly.

Also, how would we prevent the maintenance burden, coming with the suggested approach of code duplication? Kafka is a hot spot in general, having to duplicate most changes to 2 implementations does feel like a huge tradeoff to me.

exceptionfactory (changes requested):
Without reiterating previous reasons, the concern about complexity and maintainability is not primarily related to the number of lines or number of methods. If that were the only measure of maintainability, then this would be a different conversation.

The fundamental issue is the alternative consumption strategy that static partition assignment introduces. This is described well in the additional details copied from the initial implementation. The complexity shows itself not in the number of lines and methods, but in the different code paths and configuration requirements. This surfaces in elements such as the addition of the queue and map as member variables, but are only relevant to static partition configuration.

In previous implementations, one Processor supported a large number of different use cases. The question now is whether one Processor should continue to support a significantly different use case, or whether a separate Processor provides an understandable way forward.

tpalfy:
"The question now is whether one Processor should continue to support a significantly different use case, or whether a separate Processor provides an understandable way forward."

I think that's a fair and good question. I also think it warrants a cost-benefit analysis. I would consider 2 major concerns: UX and DevX.

As for UX, there is value in encompassing the functionality of "consume from kafka" in a single entity and having the "how exactly" covered by configuration. Otherwise, the line between "similar" and "significantly different" use-case can get blurry.

As for DevX, while I see the value of extension vs modification, the new processor would basically have the same SDLC as the current one. Whatever change is needed, it would be needed for both, most, if not all, of the time. That screams bad design.

This is not a new feature. It has it's risks, but it's not something that we have nothing but prediction to rely on. We have a history of this in the previous implementation. Again, maybe I missed something, but as far as I can tell - specifically from maintenance perspective - it was fine because the logic is well isolated enough.

exceptionfactory:
Thanks for the reply @tpalfy.

I agree UX and DevX are the two major areas of concern, but the response does not address the fundamental issues I have raised with each. This is admittedly a grey area, which is the reason for this extended conversation. The question is not about SDLC, but about cohesion of functionality.

Coming at this from one additional angle, this feature was introduced in NiFi 1.13.0, which was prior to the implementation of Stateless execution. Now that the framework supports configuring any Process Group with Stateless execution, it is easier to design a coherent pipeline where the NiFi consumer does not commit offsets until after successful processing. It still requires careful pipeline design, but it means that records for a specific partition can be processed to completion, and if rebalancing occurs, the new consumer receives a coherent set of records.

In a different direction, Kafka 4.0 introduced [Queues](cwiki.apache.org/confluence/display/KAFKA/Queues+for+Kafka+(KIP-932)+-+Early+Access+Release+Notes) which provide another more coherent alternative to attempted static partition assignments on the consumer side.

Although this static partition assignment may have been a useful feature at the time, now over five years ago, these are some of the reasons I do not support bringing it back in the same way. I remain open to considering an approach with a separate Processor, and I am willing to work through a review to build that in an optimal way, although I now question whether it is even needed. With different capabilities available, and a different module design, I do not support bringing back this feature as it was previously implemented.

PR #10698, opened 2025-12-29
NIFI-15393 Fix login redirect to preserve original URL

naman-shukla-ksolves:
# Summary
[NIFI-15393](https://issues.apache.org/jira/browse/NIFI-15393)

Fixed login redirect issue where users were redirected to the home screen instead of their originally requested URL after authentication. The authentication flow now preserves the original destination URL and redirects users appropriately after login.

# Tracking

### Issue Tracking
- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-15393) issue created

### Pull Request Tracking
- [x] Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-15393`
- [x] Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-15393`

### Pull Request Formatting

# Verification

### Build
  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/service/guard/authentication.guard.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/service/guard/authentication.guard.ts
@@ -29,6 +29,7 @@ import { fullScreenError } from '../../state/error/error.actions';
 import { ErrorHelper } from '../error-helper.service';
 import { selectLoginConfiguration } from '../../state/login-configuration/login-configuration.selectors';
 import { loadLoginConfigurationSuccess } from '../../state/login-configuration/login-configuration.actions';
+import { StorageService } from '../storage.service';
 
 export const authenticationGuard: CanMatchFn = () => {
     const authService: AuthService = inject(AuthService);
@@ -36,6 +37,7 @@ export const authenticationGuard: CanMatchFn = () => {
     const errorHelper: ErrorHelper = inject(ErrorHelper);
     const store: Store<CurrentUserState> = inject(Store<CurrentUserState>);
     const router: Router = inject(Router);
+    const storageService: StorageService = inject(StorageService);
 
     const getAuthenticationConfig = store.select(selectLoginConfiguration).pipe(
         take(1),
@@ -81,7 +83,7 @@ export const authenticationGuard: CanMatchFn = () => {
                                 if (errorResponse.status === 401) {
                                     const currentUrl = router.url;
                                     if (currentUrl && currentUrl !== '/login') {
-                                        sessionStorage.setItem('returnUrl', currentUrl);
+                                        storageService.setReturnUrl(currentUrl);
                                     }
                                 }
 
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/service/guard/authentication.guard.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/service/guard/authentication.guard.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { CanMatchFn } from '@angular/router';
+import { CanMatchFn, Router } from '@angular/router';
 import { inject } from '@angular/core';
 import { AuthService } from '../auth.service';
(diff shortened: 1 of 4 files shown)


exceptionfactory (changes requested):
Thanks for proposing this improvement @naman-shukla-ksolves.

From a procedural perspective, please [sign all Git commits](https://cwiki.apache.org/confluence/display/NIFI/Contributor+Guide#ContributorGuide-EnableCommitSigning) and force-push to the branch for verification.

Regarding the substantive change, I will defer to others with more perspective on the frontend, but direct reference to `sessionStorage` seems questionable. At minimum, having a shared service abstraction seems safer than relying on decoupled references to a common item key.

naman-shukla-ksolves:
@pvillard31  Thank you for the feedback. I've addressed both points:

1. All Git commits are now signed and the branch has been force-pushed for verification.
2. Replaced direct sessionStorage references with a shared service abstraction to ensure safer, decoupled state management with a single source of truth for the storage key.

Please review when you have a chance.

naman-shukla-ksolves:
Hello @pvillard31, I have verified all the commits and pushed the code, but there is a conflict in 1 file. Please help me resolve this.

exceptionfactory:
Tagging @mcgilman or @rfellows for UI review as time permits.

mcgilman:
Thanks for the mention @exceptionfactory! I'll have a look.

nifi-frontend/src/main/frontend/apps/nifi/src/app/service/guard/authentication.guard.ts
```diff
@@ -77,6 +80,13 @@ export const authenticationGuard: CanMatchFn = () => {
                             }),
                             map(() => true),
                             catchError((errorResponse: HttpErrorResponse) => {
+                                if (errorResponse.status === 401) {
+                                    const currentUrl = router.url;
+                                    if (currentUrl && currentUrl !== '/login') {
+                                        storageService.setReturnUrl(currentUrl);
+                                    }
+                                }
+
```
mcgilman, line 89: I don't think this change is needed. The auth interceptor will handle the 401 response when the user is not logged in. That dispatches the navigateToLogIn action where the current route is captured.

nifi-frontend/src/main/frontend/apps/nifi/src/app/state/current-user/current-user.effects.ts
```diff
@@ -75,6 +77,10 @@ export class CurrentUserEffects {
             this.actions$.pipe(
                 ofType(UserActions.navigateToLogIn),
                 tap(() => {
+                    const currentUrl = window.location.hash.substring(1) || this.router.url;
+                    if (currentUrl && currentUrl !== '/login') {
+                        this.storageService.setReturnUrl(currentUrl);
+                    }
```
mcgilman, line 83: When the user logs out they are sent to the logout complete page. From there if they click `home` to log back in the user will now be returned to the logout complete page following successful authentication. Thinking that maybe that could be avoided by updating the condition to ensure it doesn't equal `/logout-complete`. I haven't verified this though.

mcgilman (changes requested):
Thanks for the PR @naman-shukla-ksolves! I've noted a few things below.

nifi-frontend/src/main/frontend/apps/nifi/src/app/state/current-user/current-user.effects.ts
```diff
@@ -75,6 +77,10 @@ export class CurrentUserEffects {
             this.actions$.pipe(
                 ofType(UserActions.navigateToLogIn),
                 tap(() => {
+                    const currentUrl = window.location.hash.substring(1) || this.router.url;
```
mcgilman, line 80: Also I don't _think_ the `router.url` fallback here is correct. The application is configured with `NavigationActionTiming.PostActivation` which causes the router to not reflect until the route is activated after the guards complete. When dispatched from a guard this value will reflect the previous URL or undefined if it's a new window/tab.

PR #10716, opened 2025-12-31
NIFI-15386 - Upgrade Box SDK to Version 10

pvillard31:
# Summary

NIFI-15386 - Upgrade Box SDK to Version 10

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxEventJsonArrayWriter.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxEventJsonArrayWriter.java
@@ -16,86 +16,215 @@
  */
 package org.apache.nifi.processors.box;
 
-import com.box.sdk.BoxEvent;
-import com.eclipsesource.json.Json;
-import com.eclipsesource.json.JsonObject;
+import com.box.sdkgen.schemas.event.Event;
+import com.box.sdkgen.schemas.eventsource.EventSource;
+import com.box.sdkgen.schemas.eventsourceresource.EventSourceResource;
+import com.box.sdkgen.schemas.file.File;
+import com.box.sdkgen.schemas.folder.Folder;
+import com.box.sdkgen.schemas.foldermini.FolderMini;
+import com.box.sdkgen.schemas.user.User;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.io.OutputStreamWriter;
-import java.io.Writer;
-import java.util.Objects;
-
-import static java.nio.charset.StandardCharsets.UTF_8;
+import java.util.Map;
 
 /**
- * A class responsible for writing {@link BoxEvent} objects into a JSON array.
+ * A class responsible for writing {@link Event} objects into a JSON array.
  * Not thread-safe.
  */
 final class BoxEventJsonArrayWriter implements Closeable {
 
-    private final Writer writer;
+    private static final JsonFactory JSON_FACTORY = new JsonFactory();
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private final JsonGenerator generator;
     private boolean hasBegun;
-    private boolean hasEntries;
     private boolean closed;
 
-    private BoxEventJsonArrayWriter(final Writer writer) {
-        this.writer = writer;
+    private BoxEventJsonArrayWriter(final JsonGenerator generator) {
+        this.generator = generator;
         this.hasBegun = false;
-        this.hasEntries = false;
         this.closed = false;
     }
 
     static BoxEventJsonArrayWriter create(final OutputStream outputStream) throws IOException {
-        final Writer writer = new OutputStreamWriter(outputStream, UTF_8);
-        return new BoxEventJsonArrayWriter(writer);
+        final JsonGenerator generator = JSON_FACTORY.createGenerator(outputStream);
+        return new BoxEventJsonArrayWriter(generator);
     }
 
-    void write(final BoxEvent event) throws IOException {
+    void write(final Event event) throws IOException {
         if (closed) {
             throw new IOException("The Writer is closed");
         }
 
         if (!hasBegun) {
-            beginArray();
+            generator.writeStartArray();
             hasBegun = true;
         }
 
-        if (hasEntries) {
-            writer.write(',');
+        writeEvent(event);
+    }
+
+    private void writeEvent(final Event event) throws IOException {
+        generator.writeStartObject();
+
+        // Map Event fields to JSON using camelCase to match the original NiFi Box processor format
+        writeStringField("createdAt", event.getCreatedAt() != null ? event.getCreatedAt().toString() : null);
+        writeStringField("recordedAt", event.getRecordedAt() != null ? event.getRecordedAt().toString() : null);
+        writeStringField("eventType", event.getEventType() != null ? event.getEventType().getValue() : null);
+        writeStringField("id", event.getEventId());
+        writeStringField("sessionID", event.getSessionId());
+        writeStringField("type", event.getType());
+
+        // Handle createdBy if present (camelCase for field name, but inner fields match Box API)
+        if (event.getCreatedBy() != null) {
+            generator.writeObjectFieldStart("createdBy");
+            writeStringField("id", event.getCreatedBy().getId());
+            writeStringField("type", event.getCreatedBy().getType() != null ? event.getCreatedBy().getType().getValue() : null);
+            writeStringField("name", event.getCreatedBy().getName());
+            writeStringField("login", event.getCreatedBy().getLogin());
+            generator.writeEndObject();
+        } else {
+            generator.writeNullField("createdBy");
         }
 
-        final JsonObject json = toRecord(event);
-        json.writeTo(writer);
+        // Handle source if present - use snake_case for inner fields to match Box API format
+        writeSource(event.getSource());
 
-        hasEntries = true;
+        // Handle additionalDetails if present - serialize as proper JSON object
+        writeAdditionalDetails(event.getAdditionalDetails());
+
+        generator.writeEndObject();
     }
 
-    private JsonObject toRecord(final BoxEvent event) {
-        final JsonObject json = Json.object();
+    private void writeSource(final EventSourceResource source) throws IOException {
+        if (source == null) {
+            generator.writeNullField("source");
+            return;
+        }
 
-        json.add("accessibleBy", event.getAccessibleBy() == null ? Json.NULL : Json.parse(event.getAccessibleBy().getJson()));
-        json.add("actionBy", event.getActionBy() == null ? Json.NULL : Json.parse(event.getActionBy().getJson()));
-        json.add("additionalDetails", Objects.requireNonNullElse(event.getAdditionalDetails(), Json.NULL));
-        json.add("createdAt", event.getCreatedAt() == null ? Json.NULL : Json.value(event.getCreatedAt().toString()));
-        json.add("createdBy", event.getCreatedBy() == null ? Json.NULL : Json.parse(event.getCreatedBy().getJson()));
-        json.add("eventType", event.getEventType() == null ? Json.NULL : Json.value(event.getEventType().name()));
-        json.add("id", Objects.requireNonNullElse(Json.value(event.getID()), Json.NULL));
-        json.add("ipAddress", Objects.requireNonNullElse(Json.value(event.getIPAddress()), Json.NULL));
-        json.add("sessionID", Objects.requireNonNullElse(Json.value(event.getSessionID()), Json.NULL));
-        json.add("source", Objects.requireNonNullElse(event.getSourceJSON(), Json.NULL));
-        json.add("typeName", Objects.requireNonNullElse(Json.value(event.getTypeName()), Json.NULL));
-
-        return json;
+        generator.writeObjectFieldStart("source");
+        try {
+            // EventSourceResource is a union type (OneOfSix) - check what kind of source we have
+            if (source.isFile()) {
+                // File source - contains file_id, file_name, and parent folder info
+                File file = source.getFile();
+                writeStringField("item_type", "file");
+                writeStringField("item_id", file.getId());
+                writeStringField("item_name", file.getName());
+                // Add file-specific fields for collaboration events
+                writeStringField("file_id", file.getId());
+                writeStringField("file_name", file.getName());
+                // Add parent folder info if available
+                FolderMini parent = file.getParent();
+                if (parent != null) {
+                    writeStringField("folder_id", parent.getId());
+                    writeStringField("folder_name", parent.getName());
+                }
+            } else if (source.isFolder()) {
+                // Folder source - contains folder_id, folder_name
+                Folder folder = source.getFolder();
+                writeStringField("item_type", "folder");
+                writeStringField("item_id", folder.getId());
+                writeStringField("item_name", folder.getName());
+                // Add folder-specific fields for collaboration events
+                writeStringField("folder_id", folder.getId());
+                writeStringField("folder_name", folder.getName());
+            } else if (source.isEventSource()) {
+                // Generic EventSource - has item_type, item_id, item_name
+                EventSource eventSource = source.getEventSource();
+                String itemType = eventSource.getItemType() != null ? eventSource.getItemType().getValue() : null;
+                writeStringField("item_type", itemType);
+                writeStringField("item_id", eventSource.getItemId());
+                writeStringField("item_name", eventSource.getItemName());
+                // For EventSource, also populate file/folder specific fields based on item_type
+                if ("file".equals(itemType)) {
+                    writeStringField("file_id", eventSource.getItemId());
+                    writeStringField("file_name", eventSource.getItemName());
+                } else if ("folder".equals(itemType)) {
+                    writeStringField("folder_id", eventSource.getItemId());
+                    writeStringField("folder_name", eventSource.getItemName());
+                }
+                // Add parent folder info if available
+                FolderMini parent = eventSource.getParent();
+                if (parent != null) {
+                    writeStringField("parent_id", parent.getId());
+                    writeStringField("parent_name", parent.getName());
+                }
+            } else if (source.isUser()) {
+                // User source
+                User user = source.getUser();
+                writeStringField("item_type", "user");
+                writeStringField("id", user.getId());
+                writeStringField("name", user.getName());
+                writeStringField("login", user.getLogin());
+            } else if (source.isMap()) {
+                // Generic map - write all entries
+                Map<String, Object> map = source.getMap();
+                for (Map.Entry<String, Object> entry : map.entrySet()) {
+                    Object value = entry.getValue();
+                    if (value != null) {
+                        generator.writeFieldName(entry.getKey());
+                        generator.writeObject(value);
+                    }
+                }
+            } else if (source.isAppItemEventSource()) {
+                // AppItemEventSource
+                writeStringField("item_type", "app_item");
+            } else {
+                writeStringField("item_type", "unknown");
+            }
+        } catch (Exception e) {
+            writeStringField("error", "Could not serialize source: " + e.getMessage());
+        }
+        generator.writeEndObject();
     }
 
-    private void beginArray() throws IOException {
-        writer.write('[');
+    private void writeAdditionalDetails(final Map<String, Object> additionalDetails) throws IOException {
+        if (additionalDetails == null) {
+            generator.writeNullField("additionalDetails");
+            return;
+        }
+
+        try {
+            // Write additionalDetails as a proper JSON object, not a string
+            generator.writeFieldName("additionalDetails");
+            generator.writeStartObject();
+            for (Map.Entry<String, Object> entry : additionalDetails.entrySet()) {
+                String key = entry.getKey();
+                Object value = entry.getValue();
+                if (value == null) {
+                    generator.writeNullField(key);
+                } else if (value instanceof String) {
+                    generator.writeStringField(key, (String) value);
+                } else if (value instanceof Number) {
+                    generator.writeNumberField(key, ((Number) value).doubleValue());
+                } else if (value instanceof Boolean) {
+                    generator.writeBooleanField(key, (Boolean) value);
+                } else if (value instanceof Map) {
+                    // Nested map - use ObjectMapper to serialize
+                    generator.writeFieldName(key);
+                    generator.writeRawValue(OBJECT_MAPPER.writeValueAsString(value));
+                } else {
+                    // For other types, convert to string
+                    generator.writeStringField(key, value.toString());
+                }
+            }
+            generator.writeEndObject();
+        } catch (Exception e) {
+            generator.writeNullField("additionalDetails");
+        }
     }
 
-    private void endArray() throws IOException {
-        writer.write(']');
+    private void writeStringField(final String fieldName, final String value) throws IOException {
+        if (value != null) {
+            generator.writeStringField(fieldName, value);
+        } else {
+            generator.writeNullField(fieldName);
+        }
     }
 
     @Override
@@ -107,10 +236,10 @@ final class BoxEventJsonArrayWriter implements Closeable {
         closed = true;
 
         if (!hasBegun) {
-            beginArray();
+            generator.writeStartArray();
         }
-        endArray();
+        generator.writeEndArray();
 
-        writer.close();
+        generator.close();
     }
 }
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxEventJsonArrayWriter.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxEventJsonArrayWriter.java
@@ -198,16 +198,16 @@ final class BoxEventJsonArrayWriter implements Closeable {
                 Object value = entry.getValue();
                 if (value == null) {
                     generator.writeNullField(key);
-                } else if (value instanceof String) {
-                    generator.writeStringField(key, (String) value);
-                } else if (value instanceof Number) {
-                    generator.writeNumberField(key, ((Number) value).doubleValue());
-                } else if (value instanceof Boolean) {
-                    generator.writeBooleanField(key, (Boolean) value);
-                } else if (value instanceof Map) {
+                } else if (value instanceof String strValue) {
+                    generator.writeStringField(key, strValue);
+                } else if (value instanceof Number numValue) {
+                    generator.writeNumberField(key, numValue.doubleValue());
+                } else if (value instanceof Boolean boolValue) {
+                    generator.writeBooleanField(key, boolValue);
+                } else if (value instanceof Map<?, ?> mapValue) {
                     // Nested map - use ObjectMapper to serialize
                     generator.writeFieldName(key);
-                    generator.writeRawValue(OBJECT_MAPPER.writeValueAsString(value));
+                    generator.writeRawValue(OBJECT_MAPPER.writeValueAsString(mapValue));
                 } else {
                     // For other types, convert to string
                     generator.writeStringField(key, value.toString());
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxFileUtils.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxFileUtils.java
@@ -16,46 +16,75 @@
  */
 package org.apache.nifi.processors.box;
 
-import com.box.sdk.BoxFile;
-import com.box.sdk.BoxFolder;
-import com.box.sdk.BoxItem;
+import com.box.sdkgen.schemas.file.File;
+import com.box.sdkgen.schemas.folder.Folder;
+import com.box.sdkgen.schemas.foldermini.FolderMini;
 import org.apache.nifi.flowfile.attributes.CoreAttributes;
 
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import static java.lang.String.valueOf;
-import static java.util.stream.Collectors.joining;
 
 public final class BoxFileUtils {
 
     public static final String BOX_URL = "https://app.box.com/file/";
 
-    public static String getParentIds(final BoxItem.Info info) {
-        return info.getPathCollection().stream()
-                .map(BoxItem.Info::getID)
-                .collect(joining(","));
-    }
-    public static String getParentPath(BoxItem.Info info) {
-        return "/" + info.getPathCollection().stream()
-                .filter(pathItemInfo -> !pathItemInfo.getID().equals("0"))
-                .map(BoxItem.Info::getName)
-                .collect(joining("/"));
+    public static String getParentIds(final File fileInfo) {
+        if (fileInfo.getPathCollection() == null || fileInfo.getPathCollection().getEntries() == null) {
+            return "";
+        }
+        return fileInfo.getPathCollection().getEntries().stream()
+                .map(FolderMini::getId)
+                .collect(Collectors.joining(","));
     }
 
-    public static String getFolderPath(BoxFolder.Info folderInfo) {
-        final String parentFolderPath = getParentPath(folderInfo);
+    public static String getParentPath(final File fileInfo) {
+        if (fileInfo.getPathCollection() == null || fileInfo.getPathCollection().getEntries() == null) {
+            return "/";
+        }
+        return "/" + fileInfo.getPathCollection().getEntries().stream()
+                .filter(pathItem -> !pathItem.getId().equals("0"))
+                .map(FolderMini::getName)
+                .collect(Collectors.joining("/"));
+    }
+
+    public static String getParentPath(final List<FolderMini> pathCollection) {
+        if (pathCollection == null) {
+            return "/";
+        }
+        return "/" + pathCollection.stream()
+                .filter(pathItem -> !pathItem.getId().equals("0"))
+                .map(FolderMini::getName)
+                .collect(Collectors.joining("/"));
+    }
+
+    public static String getFolderPath(final Folder folderInfo) {
+        final String parentFolderPath = getParentPathForFolder(folderInfo);
         return "/".equals(parentFolderPath) ? parentFolderPath + folderInfo.getName() : parentFolderPath + "/" + folderInfo.getName();
     }
 
-    public static Map<String, String> createAttributeMap(BoxFile.Info fileInfo) {
+    private static String getParentPathForFolder(final Folder folderInfo) {
+        if (folderInfo.getPathCollection() == null || folderInfo.getPathCollection().getEntries() == null) {
+            return "/";
+        }
+        return "/" + folderInfo.getPathCollection().getEntries().stream()
+                .filter(pathItem -> !pathItem.getId().equals("0"))
+                .map(FolderMini::getName)
+                .collect(Collectors.joining("/"));
+    }
+
+    public static Map<String, String> createAttributeMap(final File fileInfo) {
         final Map<String, String> attributes = new LinkedHashMap<>();
-        attributes.put(BoxFileAttributes.ID, fileInfo.getID());
+        attributes.put(BoxFileAttributes.ID, fileInfo.getId());
         attributes.put(CoreAttributes.FILENAME.key(), fileInfo.getName());
         attributes.put(CoreAttributes.PATH.key(), getParentPath(fileInfo));
-        attributes.put(BoxFileAttributes.TIMESTAMP, valueOf(fileInfo.getModifiedAt()));
+        if (fileInfo.getModifiedAt() != null) {
+            attributes.put(BoxFileAttributes.TIMESTAMP, valueOf(fileInfo.getModifiedAt()));
+        }
         attributes.put(BoxFileAttributes.SIZE, valueOf(fileInfo.getSize()));
         return attributes;
     }
-
 }
--- a/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxFileUtils.java
+++ b/nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxFileUtils.java
@@ -45,10 +45,7 @@ public final class BoxFileUtils {
         if (fileInfo.getPathCollection() == null || fileInfo.getPathCollection().getEntries() == null) {
             return "/";
         }
-        return "/" + fileInfo.getPathCollection().getEntries().stream()
-                .filter(pathItem -> !pathItem.getId().equals("0"))
-                .map(FolderMini::getName)
-                .collect(Collectors.joining("/"));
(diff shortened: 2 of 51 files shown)


pvillard31:
@awelless @ncover21 - I would highly appreciate your eyes on this!

nifi-extension-bundles/nifi-box-bundle/nifi-box-services/src/main/java/org/apache/nifi/box/controllerservices/JsonConfigBasedBoxClientService.java
```diff
@@ -257,5 +233,8 @@ public void migrateProperties(PropertyConfiguration config) {
         config.renameProperty("app-config-file", APP_CONFIG_FILE.getName());
         config.renameProperty("app-config-json", APP_CONFIG_JSON.getName());
         ProxyServiceMigration.renameProxyConfigurationServiceProperty(config);
+        // Remove timeout properties that are no longer supported in Box SDK 10.x
+        config.removeProperty("Connect Timeout");
+        config.removeProperty("Read Timeout");
```
awelless: It should be possible to still configure these fields.
```java
        final OkHttpClient okHttpClient = BoxNetworkClient.getDefaultOkHttpClientBuilder()
                .connectTimeout(context.getProperty(CONNECT_TIMEOUT).asDuration())
                .readTimeout(context.getProperty(READ_TIMEOUT).asDuration())
                .build();
        final NetworkClient networkClient = new BoxNetworkClient(okHttpClient);
        final NetworkSession networkSession = new NetworkSession.Builder()
                .networkClient(networkClient)
                .build();
        return new BoxClient.Builder(new BoxJWTAuth(...))
                .networkSession(networkSession)
                .build();
```

awelless: Another option to consider, although a more radical, is to use NiFi web-client instead of okhttp.
`NetworkClient` provides a concise interface which we can use as an adapter to a web-client.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEnterpriseEvents.java
```diff
@@ -174,55 +176,80 @@ private String initializeStartEventPosition(final ProcessContext context) {
...
+    private String extractStreamPosition(final EventsNextStreamPositionField positionField) {
+        if (positionField == null) {
+            return null;
+        }
+        if (positionField.isString()) {
+            return positionField.getString();
+        } else if (positionField.isLongNumber()) {
+            return String.valueOf(positionField.getLongNumber());
+        }
+        return null;
```
awelless: Since `EventsNextStreamPositionField ` is `OneOfTwo`, if nothing is matched an `IllegalStateException` (or another exception) should be thrown instead of returning `null`.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEnterpriseEvents.java
```diff
@@ -174,55 +176,80 @@ private String initializeStartEventPosition(final ProcessContext context) {
...
+        final GetEventsQueryParams.Builder queryParamsBuilder = new GetEventsQueryParams.Builder()
                 .limit(LIMIT)
-                .position(position)
-                .typeNames(eventTypes);
+                .streamPosition(position)
+                .streamType(GetEventsQueryParamsStreamTypeField.ADMIN_LOGS_STREAMING);
+
+        // Note: Event type filtering has been removed in SDK v10 migration
+        // The eventType filter now requires GetEventsQueryParamsEventTypeField enum values
+        // TODO: Implement event type filtering with proper enum conversion if needed
```
awelless: Does it mean we're going to bypass the filter and return all events for now?
If so, this can break the existing flows, which expect only particular event types.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/test/java/org/apache/nifi/processors/box/ConsumeBoxEnterpriseEventsTest.java
```diff
@@ -16,196 +16,115 @@
...
+    private Event createMockEvent(String eventId, EventEventTypeField eventType) {
+        Event event = mock(Event.class);
+        when(event.getEventId()).thenReturn(eventId);
+        when(event.getEventType()).thenReturn(new EnumWrapper<>(eventType));
+        when(event.getCreatedAt()).thenReturn(null);
+        when(event.getSessionId()).thenReturn(null);
+        when(event.getCreatedBy()).thenReturn(null);
+        when(event.getSource()).thenReturn(null);
+        when(event.getAdditionalDetails()).thenReturn(null);
+        return event;
```
awelless: Nitpicking. What do you think about using real objects provided by SDK instead of mocks?
Same applies to other usages of mocks for SDK data classes.
```java
        return new Event.Builder()
                .eventId(eventId)
                .eventType(eventType)
                .build();

```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEvents.java
```diff
@@ -98,10 +104,10 @@ public class ConsumeBoxEvents extends AbstractBoxProcessor implements Verifiable
 
     private static final Set<Relationship> RELATIONSHIPS = Set.of(REL_SUCCESS);
 
-    private volatile BoxAPIConnection boxAPIConnection;
-    private volatile EventStream eventStream;
-    protected volatile BlockingQueue<BoxEvent> events;
-    private volatile AtomicLong position = new AtomicLong(0);
+    private volatile BoxClient boxClient;
+    protected volatile BlockingQueue<Event> events;
+    private volatile AtomicReference<String> position = new AtomicReference<>("0");
```
awelless: This filed can be final.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEvents.java
```diff
@@ -135,47 +138,71 @@ public void onScheduled(final ProcessContext context) {
...
             newQueue.addAll(events);
             events = newQueue;
         }
 
-        eventStream.addListener(new EventListener() {
+        // Start polling for events in a background thread
+        pollingExecutor = Executors.newSingleThreadScheduledExecutor();
+        pollingExecutor.scheduleWithFixedDelay(() -> {
+            try {
+                pollEvents(context);
```
awelless: If I understand correctly, now we aren't pushed with events and we poll them instead.
In that case it seems we can remove the buffering entirely, and perform polling straight in `onTrigger`. When done this way, we will lose prefetching, but it seems to be the pattern in other processors where polling is supported.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEvents.java
```diff
@@ -135,47 +138,71 @@ public void onScheduled(final ProcessContext context) {
...
+    private String extractStreamPosition(final EventsNextStreamPositionField positionField) {
+        if (positionField == null) {
+            return null;
+        }
+        if (positionField.isString()) {
+            return positionField.getString();
+        } else if (positionField.isLongNumber()) {
+            return String.valueOf(positionField.getLongNumber());
+        }
+        return null;
```
awelless: Same here. It should be an exception instead of returning a `null`.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/test/java/org/apache/nifi/processors/box/ConsumeBoxEventsTest.java
```diff
@@ -77,9 +78,7 @@ void testCaptureEvents() {
 
         final String content = ff0.getContent();
         assertTrue(content.contains("\"id\":\"1\""));
-        assertTrue(content.contains("\"eventType\":\"ITEM_CREATE\""));
         assertTrue(content.contains("\"id\":\"2\""));
-        assertTrue(content.contains("\"eventType\":\"ITEM_TRASH\""));
```
awelless, line 82: Why did we stop writing the event types?

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxEventJsonArrayWriter.java
```diff
@@ -16,86 +16,215 @@
...
+        try {
+            // Write additionalDetails as a proper JSON object, not a string
+            generator.writeFieldName("additionalDetails");
+            generator.writeStartObject();
+            for (Map.Entry<String, Object> entry : additionalDetails.entrySet()) {
+                String key = entry.getKey();
+                Object value = entry.getValue();
+                if (value == null) {
+                    generator.writeNullField(key);
+                } else if (value instanceof String) {
```
awelless: Nitpicking. We can use [pattern matching in instanceof](https://docs.oracle.com/en/java/javase/17/language/pattern-matching-instanceof.html).
```java
if (value instanceof String strValue) {
    generator.writeStringField(key, strValue);
}
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxFileUtils.java
```diff
@@ -16,46 +16,75 @@
...
-                .collect(joining("/"));
+
+    public static String getParentPath(final File fileInfo) {
+        if (fileInfo.getPathCollection() == null || fileInfo.getPathCollection().getEntries() == null) {
+            return "/";
+        }
+        return "/" + fileInfo.getPathCollection().getEntries().stream()
+                .filter(pathItem -> !pathItem.getId().equals("0"))
+                .map(FolderMini::getName)
+                .collect(Collectors.joining("/"));
```
awelless: Nitpicking.
```suggestion
        return getParentPath(fileInfo.getPathCollection().getEntries());
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CreateBoxFileMetadataInstance.java
```diff
@@ -281,12 +280,13 @@ private boolean isArray(final RecordFieldType fieldType) {
...
-     * @param fileId The ID of the file.
-     * @return A BoxFile object for the given file ID.
+     * @param fileId         The ID of the file.
+     * @param templateKey    The template key of the metadata.
+     * @param metadataValues The metadata key-value pairs.
      */
-    BoxFile getBoxFile(final String fileId) {
-        return new BoxFile(boxAPIConnection, fileId);
+    void createFileMetadata(final String fileId, final String templateKey, final Map<String, Object> metadataValues) {
+        boxClient.getFileMetadata().createFileMetadataById(fileId, CreateFileMetadataByIdScope.ENTERPRISE, templateKey, metadataValues);
```
awelless: We shouldn't always use `ENTERPRISE`.
The older SDK had the logic where it determined a scope based on the template key. It seems we have to do the same in the processor now.
From old SDK `Metadata` class:
```java
  public static final String DEFAULT_METADATA_TYPE = "properties";


  static String scopeBasedOnType(String typeName) {
    String scope;
    if (typeName.equals(DEFAULT_METADATA_TYPE)) {
      scope = GLOBAL_METADATA_SCOPE;
    } else {
      scope = ENTERPRISE_METADATA_SCOPE;
    }
    return scope;
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/DeleteBoxFileMetadataInstance.java
```diff
@@ -177,12 +177,12 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
      *
-     * @param fileId The ID of the file.
-     * @return A BoxFile object for the given file ID.
+     * @param fileId      The ID of the file.
+     * @param templateKey The template key of the metadata to delete.
      */
-    BoxFile getBoxFile(final String fileId) {
-        return new BoxFile(boxAPIConnection, fileId);
+    void deleteFileMetadata(final String fileId, final String templateKey) {
+        boxClient.getFileMetadata().deleteFileMetadataById(fileId, DeleteFileMetadataByIdScope.ENTERPRISE, templateKey);
```
awelless: We shouldn't always use ENTERPRISE.
The older SDK had the logic where it determined a scope based on the template key. It seems we have to do the same in the processor now.
From old SDK Metadata class:
```java
  public static final String DEFAULT_METADATA_TYPE = "properties";


  static String scopeBasedOnType(String typeName) {
    String scope;
    if (typeName.equals(DEFAULT_METADATA_TYPE)) {
      scope = GLOBAL_METADATA_SCOPE;
    } else {
      scope = ENTERPRISE_METADATA_SCOPE;
    }
    return scope;
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/UpdateBoxFileMetadataInstance.java
```diff
@@ -236,106 +242,116 @@ private Map<String, Object> readDesiredState(final ProcessSession session,
...
-    /**
-     * Returns a BoxFile object for the given file ID.
-     *
-     * @param fileId The ID of the file.
-     * @return A BoxFile object for the given file ID.
-     */
-    BoxFile getBoxFile(final String fileId) {
-        return new BoxFile(boxAPIConnection, fileId);
+    MetadataFull getMetadata(final String fileId, final String templateKey) {
+        return boxClient.getFileMetadata().getFileMetadataById(fileId, GetFileMetadataByIdScope.ENTERPRISE, templateKey);
```
awelless: We shouldn't always use ENTERPRISE.
The older SDK had the logic where it determined a scope based on the template key. It seems we have to do the same in the processor now.
From old SDK Metadata class:
```java
  public static final String DEFAULT_METADATA_TYPE = "properties";


  static String scopeBasedOnType(String typeName) {
    String scope;
    if (typeName.equals(DEFAULT_METADATA_TYPE)) {
      scope = GLOBAL_METADATA_SCOPE;
    } else {
      scope = ENTERPRISE_METADATA_SCOPE;
    }
    return scope;
```

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/UpdateBoxFileMetadataInstance.java
```diff
@@ -236,106 +242,116 @@ private Map<String, Object> readDesiredState(final ProcessSession session,
...
-     * @param boxFile  The Box file to update.
-     * @param metadata The metadata to update.
+     * @param fileId      The ID of the file.
+     * @param templateKey The key of the metadata template.
+     * @param operations  The list of update operations.
      */
-    void updateBoxFileMetadata(final BoxFile boxFile, final Metadata metadata) {
-        boxFile.updateMetadata(metadata);
+    void updateBoxFileMetadata(final String fileId, final String templateKey, final List<UpdateFileMetadataByIdRequestBody> operations) {
+        boxClient.getFileMetadata().updateFileMetadataById(fileId, UpdateFileMetadataByIdScope.ENTERPRISE, templateKey, operations);
```
awelless: Same here, shouldn't be enterprise

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/UpdateBoxFileMetadataInstance.java
```diff
@@ -236,106 +242,116 @@ private Map<String, Object> readDesiredState(final ProcessSession session,
...
-            final Object currentValue = metadata.getValue(propertyPath);
-
-            // Only update if values are different
+        // If exists, check if values are different
+        if (exists && extraData != null) {
+            final String fieldName = path.substring(1);
+            final Object currentValue = extraData.get(fieldName);
             if (Objects.equals(currentValue, value)) {
-                return;
+                return null; // No change needed
```
awelless: Nitpicking. I suggest using `Optional` instead of returning `null` values.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/UpdateBoxFileMetadataInstance.java
```diff
@@ -236,106 +242,116 @@ private Map<String, Object> readDesiredState(final ProcessSession session,
...
+                .build();
+    }
 
-            // Update
-            switch (value) {
-                case Number n -> metadata.replace(propertyPath, n.doubleValue());
-                case List<?> l -> metadata.replace(propertyPath, convertListToStringList(l, propertyPath));
-                case LocalDate d -> metadata.replace(propertyPath, BoxDate.of(d).format());
-                default -> metadata.replace(propertyPath, value.toString());
+    private MetadataInstanceValue convertToMetadataInstanceValue(final Object value, final String path) {
```
awelless, line 310: Nitpicking. What do you think about using pattern matching with switch?
```java
    private MetadataInstanceValue convertToMetadataInstanceValue(final Object value, final String path) {
        return switch (value) {
            case Float f -> new MetadataInstanceValue(f.doubleValue());
            case Double d -> new MetadataInstanceValue(d);
            case Number n -> new MetadataInstanceValue(n.longValue());
            case List<?> l -> {
                final List<String> stringList = l.stream()
                        .map(obj -> {
                            if (obj == null) {
                                throw new IllegalArgumentException("Null value found in list for field: " + path);
                            }
                            return obj.toString();
                        })
                        .toList();
                yield new MetadataInstanceValue(stringList);
            }
            case LocalDate d -> new MetadataInstanceValue(BoxDate.of(d).format());
            default -> new MetadataInstanceValue(value.toString());
        };
    }
```

awelless (commented):
First round of review. 
The changes marked as "nitpicking" are mostly stylistic and subjective. Feel free to ignore them. 
The rest of the comments are worth of attention, as they may affect the processor behavior.

I reviewed the controller services classes. Also I reviewed Events, Metadata Instance, and Extract Metadata processors and their tests.
I'll review the rest during the next review round.

nifi-extension-bundles/nifi-box-bundle/nifi-box-services/src/main/java/org/apache/nifi/box/controllerservices/JsonConfigBasedBoxClientService.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEnterpriseEvents.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEnterpriseEvents.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/test/java/org/apache/nifi/processors/box/ConsumeBoxEnterpriseEventsTest.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEvents.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEvents.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ConsumeBoxEvents.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/test/java/org/apache/nifi/processors/box/ConsumeBoxEventsTest.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxEventJsonArrayWriter.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/BoxFileUtils.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/CreateBoxFileMetadataInstance.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/DeleteBoxFileMetadataInstance.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/UpdateBoxFileMetadataInstance.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/UpdateBoxFileMetadataInstance.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/UpdateBoxFileMetadataInstance.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/UpdateBoxFileMetadataInstance.java
sfc-gh-pvillard: done

nifi-extension-bundles/nifi-box-bundle/nifi-box-services/src/main/java/org/apache/nifi/box/controllerservices/JsonConfigBasedBoxClientService.java
pvillard31: Regarding the use of the NiFi web client, I'd consider this as a follow up improvement. I think this does make sense but I don't think this should be part of this initial effort.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/FetchBoxFile.java
```diff
@@ -150,21 +151,25 @@ public void migrateProperties(PropertyConfiguration config) {
...
-
     private FlowFile fetchFile(String fileId, ProcessSession session, FlowFile flowFile) {
-        final BoxFile boxFile = getBoxFile(fileId);
-        flowFile = session.write(flowFile, outputStream -> boxFile.download(outputStream));
-        flowFile = session.putAllAttributes(flowFile, BoxFileUtils.createAttributeMap(boxFile.getInfo()));
+        try (InputStream inputStream = boxClient.getDownloads().downloadFile(fileId)) {
+            flowFile = session.importFrom(inputStream, flowFile);
+        } catch (Exception e) {
+            throw new ProcessException("Failed to download file from Box", e);
+        }
```
awelless: Previously, the SDK errors thrown by `fetchFile` would've been propagated to the caller and handled in `onTrigger`. 
Since everything is wrapped into a `ProcessException`, so the handling of `BoxAPIError` in `onTrigger` won't take place. Let's catch only `IOException` here, passing the rest up the call chain

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/FetchBoxFile.java
```diff
@@ -150,21 +151,25 @@ public void migrateProperties(PropertyConfiguration config) {
...
+
+        final GetFileByIdQueryParams queryParams = new GetFileByIdQueryParams.Builder()
+                .fields(List.of("id", "name", "size", "modified_at", "path_collection"))
```
awelless: Nitpicking. These fields can be moved to a constant.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/FetchBoxFileInfo.java
```diff
@@ -148,45 +149,45 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
-
     private FlowFile fetchFileMetadata(final String fileId,
                                        final ProcessSession session,
                                        final FlowFile flowFile) {
-        final BoxFile boxFile = getBoxFile(fileId);
-        final BoxFile.Info fileInfo = boxFile.getInfo("name", "description", "size", "created_at", "modified_at",
-                "owned_by", "parent", "etag", "sha1", "item_status", "sequence_id", "path_collection",
-                "content_created_at", "content_modified_at", "trashed_at", "purged_at", "shared_link");
+        final GetFileByIdQueryParams queryParams = new GetFileByIdQueryParams.Builder()
+                .fields(List.of("name", "description", "size", "created_at", "modified_at",
```
awelless: Nitpicking. A constant for the parameter list.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/FetchBoxFileRepresentation.java
```diff
@@ -157,60 +156,86 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
 
-            flowFile = session.write(flowFile, outputStream ->
-                    // Download the file representation, box sdk handles a request to create representation if it doesn't exist
-                    boxFile.getRepresentationContent("[" + representationType + "]", "", outputStream, MAX_RETRIES)
-            );
+            // Download the representation content
+            flowFile = session.write(flowFile, outputStream -> {
+                try {
+                    HttpClient client = HttpClient.newHttpClient();
+                    HttpRequest request = HttpRequest.newBuilder()
```
awelless: I didn't find a method in the SDK to download the representation, either.
I have some concerns about using java's `HttpClient` here. Any configuration of `BoxClientService` is ignored.

I recommend considering the usage of `NetworkSession` and `NetworkClient` from `boxClient` and see how it's done in `boxClient.getDownloads().downloadFile()` method.  This is going to be more verbose, but will use an already configured Box connection.
If we decide to proceed with another client, will a dedicated `WebClient` be a better option?
If we decide to keep using java `HttpClient`, shall we create it only once in `@OnScheduled`?

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/GetBoxFileCollaborators.java
```diff
@@ -256,17 +242,21 @@ private Set<String> parseFilter(final String filter) {
      * @param attributeValues the map to populate with collaboration attributes
      * @return the total count of collaborations processed (before filtering)
      */
-    private int processCollaborations(final Iterable<BoxCollaboration.Info> collaborations,
+    private int processCollaborations(final Collaborations collaborations,
                                       final Set<String> allowedRoles,
                                       final Set<String> allowedStatuses,
                                       final Map<String, List<String>> attributeValues) {
         int count = 0;
```
awelless: Nitpicking. No need to keep `count`, as `collaborations.getEntries()` returns a list, so `size()` can be used.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ListBoxFileMetadataInstances.java
```diff
@@ -134,28 +132,32 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
+                if (templateName != null) {
+                    templateNames.add(templateName);
+                }
 
                 // Add standard metadata fields
-                processBoxMetadataInstance(fileId, metadata, instanceFields);
+                instanceFields.put("$parent", metadata.getParent());
+                instanceFields.put("$template", templateName);
+                instanceFields.put("$scope", metadata.getScope());
+                instanceFields.put("$version", metadata.getVersion());
```
awelless: Since we don't get `MetadataFull` here, we can't populate all the fields in `processBoxMetadataInstance`. Theoretically, this can break some flows, although might not be necessary for everyone.

Shall we add fetch `MetadataFull` as an optional behavior, which is enabled by default?

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/ListBoxFileMetadataTemplates.java
```diff
@@ -133,41 +132,31 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
                 final Map<String, Object> templateFields = new HashMap<>();
+                final String templateName = metadata.getTemplate();
 
-                templateNames.add(metadata.getTemplateName());
+                if (templateName != null) {
+                    templateNames.add(templateName);
+                }
 
                 // Add standard metadata fields
-                templateFields.put("$id", metadata.getID());
```
awelless, line 157: Since we don't get `MetadataFull` here, we can't get the id Theoretically, this can break some flows, although might not be necessary for everyone.

Shall we add fetch `MetadataFull` as an optional behavior, which is enabled by default?

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/PutBoxFile.java
```diff
@@ -271,145 +281,178 @@ public void migrateProperties(PropertyConfiguration config) {
...
+                .fields(List.of("name", "type"))
+                .build();
 
-    private Optional<BoxFile> getFileByName(final String filename, final BoxFolder parentFolder) {
-        return getItemByName(filename, parentFolder, BoxFile.Info.class)
-                .map(BoxFile.Info::getResource);
+        final Items items = boxClient.getFolders().getFolderItems(parentFolderId, queryParams);
+
+        if (items.getEntries() != null) {
+            for (Object itemObj : items.getEntries()) {
```
awelless: `items.getEntries()` returns a list of `Item`. `Item` is `OneOfThree<FileFull, FolderFull, WebLink>`. 
So:
1. We should use `OneOfThree` api instead of instanceof checks.
2. We can operate on `FolderFull` instead of folder ids. This will allow us to get `FolderFull` in `onTrigger` (line 232), so fetching a folder info on line 233 would be redundant.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/PutBoxFile.java
```diff
@@ -221,21 +229,22 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
 
         try {
             final long size = flowFile.getSize();
-            final BoxFolder parentFolder = getOrCreateDirectParentFolder(context, flowFile);
-            fullPath = BoxFileUtils.getFolderPath(parentFolder.getInfo());
-            BoxFile.Info uploadedFileInfo = null;
+            final String parentFolderId = getOrCreateDirectParentFolder(context, flowFile);
+            final FolderFull parentFolderInfo = getFolderInfo(parentFolderId);
```
awelless: Fetching a full folder representation seems redundant. See my comment in the `getFolderIdByName` method.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/PutBoxFile.java
```diff
@@ -271,145 +281,178 @@ public void migrateProperties(PropertyConfiguration config) {
...
-
-            if (size > chunkUploadThreshold) {
-                return existingBoxFile.uploadLargeFile(inputStream, size);
-            } else {
-                return existingBoxFile.uploadNewVersion(inputStream);
+    private FileFull replaceBoxFileIfExists(String parentFolderId, String filename, final InputStream inputStream, final long size, final long chunkUploadThreshold)
+            throws Exception {
+        final Optional<String> existingFileId = getFileIdByName(filename, parentFolderId);
+        if (existingFileId.isPresent()) {
+            final String fileId = existingFileId.get();
```
awelless: Nitpicking. `Optional.isPresent() - get()` can be replaced with a more type-safe and idiomatic handling. Otherwise it doesn't differ from a plain null check.
```java
        return getFileIdByName(filename, parentFolderId)
                .flatMap(fileId -> {
                    final UploadFileVersionRequestBodyAttributesField attributes = new UploadFileVersionRequestBodyAttributesField(filename);
                    final UploadFileVersionRequestBody requestBody = new UploadFileVersionRequestBody(attributes, inputStream);
                    final Files files = boxClient.getUploads().uploadFileVersion(fileId, requestBody);

                    return files.getEntries() != null && !files.getEntries().isEmpty()
                            ? Optional.of(files.getEntries().getFirst())
                            : Optional.empty();
                });

```

awelless: Plain null check is also fine, but then there is little value in wrapping the result of `getFileIdByName` in `Optional`.

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/PutBoxFile.java
```diff
@@ -271,145 +281,178 @@ public void migrateProperties(PropertyConfiguration config) {
...
+
+            // Upload new version
+            final UploadFileVersionRequestBodyAttributesField attributes = new UploadFileVersionRequestBodyAttributesField(filename);
+            final UploadFileVersionRequestBody requestBody = new UploadFileVersionRequestBody(attributes, inputStream);
+            final Files files = boxClient.getUploads().uploadFileVersion(fileId, requestBody);
+            if (files.getEntries() != null && !files.getEntries().isEmpty()) {
+                return files.getEntries().get(0);
             }
```
awelless: Shall we throw an exception if file upload succeeds, but no file metadata is returned?
How shall the caller handle absent `FileFull`?

nifi-extension-bundles/nifi-box-bundle/nifi-box-processors/src/main/java/org/apache/nifi/processors/box/PutBoxFile.java
```diff
@@ -271,145 +281,178 @@ public void migrateProperties(PropertyConfiguration config) {
...
+    private FileFull createBoxFile(String parentFolderId, String filename, InputStream inputStream, long size, final long chunkUploadThreshold)
+            throws Exception {
+        final UploadFileRequestBodyAttributesParentField parent = new UploadFileRequestBodyAttributesParentField(parentFolderId);
+        final UploadFileRequestBodyAttributesField attributes = new UploadFileRequestBodyAttributesField(filename, parent);
+        final UploadFileRequestBody requestBody = new UploadFileRequestBody(attributes, inputStream);
+        final Files files = boxClient.getUploads().uploadFile(requestBody);
+        if (files.getEntries() != null && !files.getEntries().isEmpty()) {
+            return files.getEntries().get(0);
         }
+        return null;
```
awelless: Shall we throw an exception if file upload succeeds, but no file metadata is returned?
How shall the caller handle absent `FileFull`?

awelless (commented):
Round 2 of review.
I reviewed all source files, but not the tests.

pvillard31:
Thanks @awelless - I pushed another commit to address your comments

PR #10890, opened 2026-02-12
NIFI-15587 - Add extensible FieldValidator and RecordValidator support to the Record API

pvillard31:
# Summary

NIFI-15587 - Add extensible FieldValidator and RecordValidator support to the Record API

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/validation/DefaultValidationError.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/validation/DefaultValidationError.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


exceptionfactory (changes requested):
Thanks for decomposing the Record and Field Validation into this specific set of changes @pvillard31.

On initial review, I'm concerned about the tight coupling of Validation to the `RecordSchema` and `RecordField` objects.

What do you think about an approach that keeps Validators independent of the Schema and Field information?

pvillard31:
Thanks for looking into this @exceptionfactory - I agree that the coupling might not the best approach. Here's the alternative I'm considering:

- Do not touch `RecordField` and `RecordSchema`.
- `FieldValidator` and `RecordValidator` interfaces are simplified:
  - FieldValidator.validate(String fieldPath, Object value). The validator captures any field metadata it needs at construction time.
  - RecordValidator.validate(Record record, String fieldPath). Same idea.
- A new SchemaValidators container (or similar name) holds the association between fields and their validators:
  - Map<String, List<FieldValidator>> mapping field paths to field-level validators
  - List<RecordValidator> for record-level validators
- SchemaValidationContext is extended to optionally carry a SchemaValidators instance.
- StandardSchemaValidator reads validators from its context, not from the schema or fields.

Thoughts?

exceptionfactory:

The alternative you described sounds like a better path forward. Decoupling the Validators from the Field and Schema objects seems like a good pattern.

pvillard31:
I pushed a commit with the new approach.

exceptionfactory (changes requested):
Thanks for the revised approach on this @pvillard31. This looks close to completion if you and rebase from the latest main branch.

exceptionfactory (approved):
Thanks @pvillard31!

PR #10891, opened 2026-02-12
NIFI-15588 - Improve date, time, and timestamp type compatibility with ISO 8601 default formatters

pvillard31:
# Summary

NIFI-15588 - Improve date, time, and timestamp type compatibility with ISO 8601 default formatters

## Summary

- Add default ISO 8601 formatter fallback to temporal compatibility checks (`isDateTypeCompatible`, `isTimeTypeCompatible`, `isTimestampTypeCompatible`) and to the field converters (`ObjectLocalDateFieldConverter`, `ObjectLocalTimeFieldConverter`, `ObjectLocalDateTimeFieldConverter`), allowing ISO-formatted strings to be recognized and converted without requiring an explicit format pattern.
- Introduce a three-argument overload for each compatibility method with an `allowDefaultFormats` parameter, enabling callers to opt out of the new ISO fallback and retain the legacy epoch-only behavior.
- Apply `ResolverStyle.STRICT` to both the default ISO formatters and the explicit format path in compatibility checks, improving date validation strictness.

## Motivation

Prior to this change, the temporal compatibility methods (`isDateTypeCompatible`, `isTimeTypeCompatible`, `isTimestampTypeCompatible`) only accepted numeric (epoch) strings when no explicit format pattern was provided. ISO 8601 strings such as `"2024-01-15"` or `"13:45:00"` were rejected, even though they are the most common standard date/time representations. Similarly, the field converters would throw a `FieldConversionException` for ISO strings when no pattern was supplied, falling through directly to the epoch-millis numeric parse.

Additionally, `isTimeTypeCompatible` and `isTimestampTypeCompatible` simply delegated to `isDateTypeCompatible`, meaning all three types shared the same set of accepted formats. This made it impossible to use type-specific default formatters.

### `AbstractCSVRecordReader`

Updated `convertSimpleIfPossible` to use the three-argument overloads with `allowDefaultFormats` set to `true` only when the corresponding format (`dateFormat`, `timeFormat`, `timestampFormat`) is non-null and non-blank. This preserves existing CSV reader behavior: without an explicit format configured, only epoch-like numeric strings are coerced.

### `ExcelRecordReader`

Applied the same guarding pattern as `AbstractCSVRecordReader` to `convertSimpleIfPossible`, preventing the new ISO fallback from changing behavior for the Excel reader when no explicit temporal format is configured.

## Behavior Changes

| Scenario | Before | After |
|---|---|---|
| `isDateTypeCompatible("2024-01-15", null)` | `false` (epoch-only) | `true` (ISO fallback) |
| `isTimeTypeCompatible("13:45:00", null)` | `false` (delegated to date check) | `true` (ISO fallback) |
| `isTimestampTypeCompatible("2024-01-15T13:45:00Z", null)` | `false` (delegated to date check) | `true` (ISO fallback) |
| `isDateTypeCompatible("2024-02-30", null)` | `false` | `false` (ISO STRICT rejects) |
| `isDateTypeCompatible("2024-01-15", null, false)` | N/A (new overload) | `false` (epoch-only) |
| CSV/Excel reader with no explicit format | epoch-only coercion | epoch-only coercion (unchanged) |
| Converter with no pattern, ISO string input | `FieldConversionException` | Parsed via ISO formatters |
| Blank format string `" "` passed to compatibility check | `IllegalArgumentException` | Treated as absent format |

**Callers unaffected by the two-argument overload change:**
- `isCompatibleDataType(value, dataType)` — always passes non-null default formats (`"yyyy-MM-dd"`, `"HH:mm:ss"`, `"yyyy-MM-dd HH:mm:ss"`), so the null-format ISO fallback path is never reached.
- `CEFRecordReader` — always passes non-null format constants.
- `AbstractCSVRecordReader` and `ExcelRecordReader` — explicitly use the three-argument overload to opt out.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateFieldConverter.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/field/ObjectLocalDateFieldConverter.java
@@ -24,12 +24,19 @@ import java.time.LocalDate;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.time.format.ResolverStyle;
 import java.util.Optional;
 
 /**
  * Convert Object to java.time.LocalDate using instanceof evaluation and optional format pattern for DateTimeFormatter
  */
 class ObjectLocalDateFieldConverter implements FieldConverter<Object, LocalDate> {
+    private static final DateTimeFormatter[] DEFAULT_DATE_FORMATTERS = new DateTimeFormatter[]{
+            DateTimeFormatter.ISO_LOCAL_DATE.withResolverStyle(ResolverStyle.STRICT),
+            DateTimeFormatter.ISO_OFFSET_DATE.withResolverStyle(ResolverStyle.STRICT),
+            DateTimeFormatter.ISO_DATE.withResolverStyle(ResolverStyle.STRICT)
+    };
+
     /**
      * Convert Object field to java.sql.Timestamp using optional format supported in DateTimeFormatter
      *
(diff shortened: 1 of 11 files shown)


exceptionfactory (commented):
Thanks for putting this together @pvillard31. On general review, I'm concerned about the potential impact of looping through multiple date and time formats when attempting conversion. I think an optimal approach would minimize the passes over the input string. I can take a closer look soon.

PR #10910, opened 2026-02-17
NIFI-15182: Fixed adding/removing user to multiple groups not  saving…

Freedom9339:
… to all groups.

# Summary

[NIFI-15182](https://issues.apache.org/jira/browse/NIFI-15182): There was an issue where if multiple groups were selected to add a user to, the update would sometimes fail on all groups, and the operation would have to be repeated to add the user to all desired groups. This also happened when deleting  a user from multiple groups. This change fixes this issue.

# Tracking

### Issue Tracking

### Pull Request Tracking

- [X] Pull request contains [commits signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) with a registered key indicating `Verified` status

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
@@ -263,17 +263,23 @@ export class UserListingEffects {
             ofType(UserListingActions.createUserSuccess),
             map((action) => action.response),
             filter((response) => response.userGroupUpdate != null),
-            mergeMap((createUserResponse) =>
-                this.actions$.pipe(
+            mergeMap((createUserResponse) => {
+                // @ts-ignore
+                const expectedCount = createUserResponse.userGroupUpdate.userGroups?.length || 0;
+                if (expectedCount === 0) {
+                    return of(UserListingActions.createUserComplete({ response: createUserResponse }));
+                }
+                return this.actions$.pipe(
                     ofType(UserListingActions.updateUserGroupSuccess),
                     filter(
                         (updateSuccess) =>
                             // @ts-ignore
                             createUserResponse.userGroupUpdate.requestId === updateSuccess.response.requestId
                     ),
+                    take(expectedCount),
                     map(() => UserListingActions.createUserComplete({ response: createUserResponse }))
-                )
-            )
+                );
+            })
         )
     );
 
@@ -525,7 +531,7 @@ export class UserListingEffects {
                     if (userGroupUpdates.length === 0) {
                         return of(UserListingActions.updateUserComplete());
                     } else {
-                        return userGroupUpdates;
+                        return from(userGroupUpdates);
                     }
                 } else {
                     return of(UserListingActions.updateUserComplete());
@@ -539,17 +545,28 @@ export class UserListingEffects {
             ofType(UserListingActions.updateUserSuccess),
             map((action) => action.response),
             filter((response) => response.userGroupUpdate != null),
-            mergeMap((updateUserResponse) =>
-                this.actions$.pipe(
+            mergeMap((updateUserResponse) => {
+                // @ts-ignore
+                const addedCount = updateUserResponse.userGroupUpdate.userGroupsAdded?.length || 0;
+                // @ts-ignore
+                const removedCount = updateUserResponse.userGroupUpdate.userGroupsRemoved?.length || 0;
+                const expectedCount = addedCount + removedCount;
+
+                if (expectedCount === 0) {
+                    return of(UserListingActions.updateUserComplete());
+                }
+
+                return this.actions$.pipe(
                     ofType(UserListingActions.updateUserGroupSuccess),
                     filter(
                         (updateSuccess) =>
                             // @ts-ignore
                             updateUserResponse.userGroupUpdate.requestId === updateSuccess.response.requestId
                     ),
+                    take(expectedCount),
                     map(() => UserListingActions.updateUserComplete())
-                )
-            )
+                );
+            })
         )
     );
 
@@ -622,7 +639,7 @@ export class UserListingEffects {
         this.actions$.pipe(
             ofType(UserListingActions.updateUserGroup),
             map((action) => action.request),
-            switchMap((request) =>
+            mergeMap((request) =>
                 from(this.usersService.updateUserGroup(request)).pipe(
                     map((response) =>
                         UserListingActions.updateUserGroupSuccess({
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
@@ -231,7 +231,7 @@ export class UserListingEffects {
                     if (userGroupUpdates.length === 0) {
                         return of(UserListingActions.createUserComplete({ response }));
                     } else {
-                        return userGroupUpdates;
+                        return from(userGroupUpdates);
                     }
                 } else {
                     return of(UserListingActions.createUserComplete({ response }));
@@ -264,8 +264,7 @@ export class UserListingEffects {
             map((action) => action.response),
             filter((response) => response.userGroupUpdate != null),
             mergeMap((createUserResponse) => {
-                // @ts-ignore
-                const expectedCount = createUserResponse.userGroupUpdate.userGroups?.length || 0;
+                const expectedCount = createUserResponse.userGroupUpdate?.userGroups?.length ?? 0;
                 if (expectedCount === 0) {
                     return of(UserListingActions.createUserComplete({ response: createUserResponse }));
                 }
@@ -546,10 +545,8 @@ export class UserListingEffects {
             map((action) => action.response),
             filter((response) => response.userGroupUpdate != null),
             mergeMap((updateUserResponse) => {
-                // @ts-ignore
-                const addedCount = updateUserResponse.userGroupUpdate.userGroupsAdded?.length || 0;
-                // @ts-ignore
-                const removedCount = updateUserResponse.userGroupUpdate.userGroupsRemoved?.length || 0;
+                const addedCount = updateUserResponse.userGroupUpdate?.userGroupsAdded?.length ?? 0;
+                const removedCount = updateUserResponse.userGroupUpdate?.userGroupsRemoved?.length ?? 0;
                 const expectedCount = addedCount + removedCount;
 
                 if (expectedCount === 0) {
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
@@ -23,7 +23,20 @@ import { Store } from '@ngrx/store';
 import { Router } from '@angular/router';
 import * as UserListingActions from './user-listing.actions';
 import { selectTenant } from './user-listing.actions';
-import { catchError, combineLatest, filter, from, map, mergeMap, of, switchMap, take, takeUntil, tap } from 'rxjs';
+import {
+    catchError,
+    combineLatest,
+    filter,
+    from,
+    last,
+    map,
+    mergeMap,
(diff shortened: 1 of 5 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -622,7 +639,7 @@ export class UserListingEffects {
         this.actions$.pipe(
             ofType(UserListingActions.updateUserGroup),
             map((action) => action.request),
-            switchMap((request) =>
+            mergeMap((request) =>
```
mcgilman, line 660: The switchMap → mergeMap change in updateUserGroup$ correctly addresses the root cause. Thanks for catching this.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -263,17 +263,23 @@ export class UserListingEffects {
             ofType(UserListingActions.createUserSuccess),
             map((action) => action.response),
             filter((response) => response.userGroupUpdate != null),
-            mergeMap((createUserResponse) =>
-                this.actions$.pipe(
+            mergeMap((createUserResponse) => {
+                // @ts-ignore
+                const expectedCount = createUserResponse.userGroupUpdate.userGroups?.length || 0;
```
mcgilman: You can avoid the @ts-ignore here by extending the optional chaining one level up — this lets the compiler handle the nullability without suppressing other potential errors:

`const expectedCount = createUserResponse.userGroupUpdate?.userGroups?.length ?? 0;`

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -539,17 +545,28 @@ export class UserListingEffects {
             ofType(UserListingActions.updateUserSuccess),
             map((action) => action.response),
             filter((response) => response.userGroupUpdate != null),
-            mergeMap((updateUserResponse) =>
-                this.actions$.pipe(
+            mergeMap((updateUserResponse) => {
+                // @ts-ignore
+                const addedCount = updateUserResponse.userGroupUpdate.userGroupsAdded?.length || 0;
+                // @ts-ignore
+                const removedCount = updateUserResponse.userGroupUpdate.userGroupsRemoved?.length || 0;
```
mcgilman: Same pattern here:

```
const addedCount = updateUserResponse.userGroupUpdate?.userGroupsAdded?.length ?? 0;
const removedCount = updateUserResponse.userGroupUpdate?.userGroupsRemoved?.length ?? 0;
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -525,7 +531,7 @@ export class UserListingEffects {
                     if (userGroupUpdates.length === 0) {
                         return of(UserListingActions.updateUserComplete());
                     } else {
-                        return userGroupUpdates;
+                        return from(userGroupUpdates);
```
mcgilman, line 553: Good change wrapping in from() here. It makes the intent clearer. Would you mind doing the same in createUserSuccess$ (line 234)? It still returns the raw array there. Both work, but it'd be nice to keep them consistent.

`return userGroupUpdates;`

mcgilman (changes requested):
Thanks for the PR @Freedom9339! Noted a few things below.

Also, could you add some test coverage for the changed effects? The existing spec file covers loadTenants scenarios, but the effects modified in this PR (updateUserGroup$, awaitUpdateUserGroupsForCreateUser$, awaitUpdateUserGroupsForUpdateUser$, updateUserSuccess$) don't have tests yet.

In particular, a test for updateUserGroup$ verifying that multiple concurrent actions all complete their HTTP requests would help guard against regression — the original switchMap → mergeMap fix is subtle enough that it could be accidentally reverted without a test catching it.

Freedom9339:
Thank you for the review @mcgilman. I've made the suggested changes and added unit tests.

Freedom9339:
Fixed Merge issues. Any updates on this?

Freedom9339:
Looks like there was a MR that went into main 5 minutes before the workflows started. Rebased.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -263,17 +263,22 @@ export class UserListingEffects {
...
+                    return of(UserListingActions.createUserComplete({ response: createUserResponse }));
+                }
+                return this.actions$.pipe(
                     ofType(UserListingActions.updateUserGroupSuccess),
                     filter(
                         (updateSuccess) =>
                             // @ts-ignore
                             createUserResponse.userGroupUpdate.requestId === updateSuccess.response.requestId
                     ),
+                    take(expectedCount),
```
mcgilman, line 292: This only considers successfully emissions. If any of the requests fail `take` won't complete.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -539,17 +544,26 @@ export class UserListingEffects {
...
+                }
+
+                return this.actions$.pipe(
                     ofType(UserListingActions.updateUserGroupSuccess),
                     filter(
                         (updateSuccess) =>
                             // @ts-ignore
                             updateUserResponse.userGroupUpdate.requestId === updateSuccess.response.requestId
                     ),
+                    take(expectedCount),
```
mcgilman, line 583: Same comment from above about `take` not completing if any of the requests fail.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -263,17 +263,22 @@ export class UserListingEffects {
...
+                }
+                return this.actions$.pipe(
                     ofType(UserListingActions.updateUserGroupSuccess),
                     filter(
                         (updateSuccess) =>
                             // @ts-ignore
                             createUserResponse.userGroupUpdate.requestId === updateSuccess.response.requestId
                     ),
+                    take(expectedCount),
                     map(() => UserListingActions.createUserComplete({ response: createUserResponse }))
```
mcgilman, line 294: The existing code here suffers from emitting `createUserComplete` for each successful user group update. I think this can be easily improved with a `last()` between the `take` and `map`.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -539,17 +544,26 @@ export class UserListingEffects {
...
+
+                return this.actions$.pipe(
                     ofType(UserListingActions.updateUserGroupSuccess),
                     filter(
                         (updateSuccess) =>
                             // @ts-ignore
                             updateUserResponse.userGroupUpdate.requestId === updateSuccess.response.requestId
                     ),
+                    take(expectedCount),
                     map(() => UserListingActions.updateUserComplete())
```
mcgilman, line 585: Same comment from above about `last()` in between `take` and `map` to limit `updateUserComplete` to a single emission.

mcgilman (changes requested):
Thanks for the updates. Left a couple more notes.

Freedom9339:
@mcgilman Sorry I lost track of this. I've added error handling so that take() completes even if an error happens. And I've added last() to only emit once.

mcgilman (commented):
Thanks for the updates. A couple of correctness notes on the count-based completion, plus one minor cleanup. Left the details inline.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -263,17 +276,24 @@ export class UserListingEffects {
...
             map((action) => action.response),
             filter((response) => response.userGroupUpdate != null),
-            mergeMap((createUserResponse) =>
-                this.actions$.pipe(
-                    ofType(UserListingActions.updateUserGroupSuccess),
+            mergeMap((createUserResponse) => {
+                const expectedCount = createUserResponse.userGroupUpdate?.userGroups?.length ?? 0;
+                if (expectedCount === 0) {
+                    return of(UserListingActions.createUserComplete({ response: createUserResponse }));
+                }
```
mcgilman, line 284: This `expectedCount === 0` early return causes a duplicate completion. `createUserSuccess$` already emits `createUserComplete` for the empty case (`userGroupUpdates.length === 0`), and this effect fires on the same `createUserSuccess` action, so creating a user with no groups completes twice — double `closeAll()`, `selectTenant()`, and `loadTenants()`. This is a common path, since `createUser` always carries a non-null `userGroupUpdate` (`userGroups` is `[]` when nothing is selected). At base this effect emitted nothing when no group updates were pending, so `createUserSuccess$` was the sole completer.

Return `EMPTY` here and let `createUserSuccess$` own the zero-group completion:

```suggestion
                if (expectedCount === 0) {
                    return EMPTY;
                }
```

`EMPTY` needs adding to the `rxjs` import. A test that runs a no-group `createUserSuccess` through both effects and asserts a single `createUserComplete` would guard this — the current isolated specs don't run the two together.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -539,17 +559,28 @@ export class UserListingEffects {
...
-                this.actions$.pipe(
-                    ofType(UserListingActions.updateUserGroupSuccess),
+            mergeMap((updateUserResponse) => {
+                const addedCount = updateUserResponse.userGroupUpdate?.userGroupsAdded?.length ?? 0;
+                const removedCount = updateUserResponse.userGroupUpdate?.userGroupsRemoved?.length ?? 0;
+                const expectedCount = addedCount + removedCount;
+
+                if (expectedCount === 0) {
+                    return of(UserListingActions.updateUserComplete());
+                }
```
mcgilman, line 574: Same duplicate-completion issue as the create path: `updateUserSuccess$` already returns `updateUserComplete` for the empty case. This branch isn't reachable from the UI today (`updateUser` only carries a `userGroupUpdate` when there's at least one add/remove), but returning `EMPTY` keeps the two await effects symmetric and avoids a duplicate if that invariant changes:

```suggestion
                if (expectedCount === 0) {
                    return EMPTY;
                }
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/users/state/user-listing/user-listing.effects.ts
```diff
@@ -263,17 +276,24 @@ export class UserListingEffects {
...
+                const expectedCount = createUserResponse.userGroupUpdate?.userGroups?.length ?? 0;
+                if (expectedCount === 0) {
+                    return of(UserListingActions.createUserComplete({ response: createUserResponse }));
+                }
+                return this.actions$.pipe(
+                    ofType(UserListingActions.updateUserGroupSuccess, UserListingActions.updateUserGroupError),
                     filter(
-                        (updateSuccess) =>
+                        (action) =>
                             // @ts-ignore
```
mcgilman: The `@ts-ignore` on the count lines is resolved. This `filter` predicate (and the equivalent one at line 575) still carries one for the `'response' in action ? action.response.requestId : action.requestId` discriminant. Since `updateUserGroupSuccess`/`updateUserGroupError` are a clean discriminated pair, this can be typed directly and the suppression dropped. Minor.

Freedom9339:
@mcgilman I made the requested changes and updated unit test.

PR #10972, opened 2026-03-05
NIFI-15673 - Add support for Author+Committer with git-based Flow Registry Clients

pvillard31:
# Summary

NIFI-15673 - Add support for Author+Committer with git-based Flow Registry Clients

Right now, in NiFi, when using a git-based Flow Registry Client, all commits are made with the identity that is configured for the authentication of the registry client. It means that if many users are using the same NiFi UI, all commits made by different users will be reported as being made with the same user (likely the service account used in the registry client configuration).

This is not ideal when working with multiple users and trying to distinguish who made the actual commits.

With the git-based implementations, it is possible to make a commit and specify both an author and a committer.

This change is to use the identity of the NiFi User connected in the UI as the author and use the identity configured in the registry client as the committer.

Important notes:
- the proposed implementation in GitHub is using reflection as a temporary workaround until a submitted improvement is accepted/merged/released (see https://github.com/hub4j/github-api/pull/2200)
- if the NiFi User does not resolve to an email address, the identity won't be resolved on the git provider side and it will just show the author as an unkown user with just the NiFi identity
- I didn't see an option to do that with Bitbucket Data Center

This has been tested with GitHub, GitLab, Azure DevOps and Bitbucket cloud. See some screenshots below.

A potential follow-up improvement would be to show this information in the UI in the views where we list flow versions.

# Screenshots

#### GitHub

Screenshot 2026-03-05 at 12 02 32

#### GitLab

Screenshot 2026-03-05 at 12 02 54

#### Bitbucket (only the author is shown in the UI - see https://jira.atlassian.com/browse/BCLOUD-13836)

Screenshot 2026-03-05 at 12 03 31

### When my NiFi identity does resolve to a user in the git provider:

Screenshot 2026-03-05 at 12 12 38

#### GitHub

Screenshot 2026-03-05 at 12 12 19

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-atlassian-bundle/nifi-atlassian-extensions/src/main/java/org/apache/nifi/atlassian/bitbucket/BitbucketRepositoryClient.java
+++ b/nifi-extension-bundles/nifi-atlassian-bundle/nifi-atlassian-extensions/src/main/java/org/apache/nifi/atlassian/bitbucket/BitbucketRepositoryClient.java
@@ -448,6 +448,13 @@ public class BitbucketRepositoryClient implements GitRepositoryClient {
             multipartBuilder.addPart(FIELD_BRANCH, StandardHttpContentType.TEXT_PLAIN, branch.getBytes(StandardCharsets.UTF_8));
             multipartBuilder.addPart(FIELD_PARENTS, StandardHttpContentType.TEXT_PLAIN, branchHead.getBytes(StandardCharsets.UTF_8));
 
+            final String authorName = request.getAuthorName();
+            final String authorEmail = request.getAuthorEmail();
+            if (authorName != null && authorEmail != null) {
+                final String authorValue = "%s <%s>".formatted(authorName, authorEmail);
+                multipartBuilder.addPart(FIELD_AUTHOR, StandardHttpContentType.TEXT_PLAIN, authorValue.getBytes(StandardCharsets.UTF_8));
+            }
+
             final HttpResponseEntity response = this.webClient.getWebClientService()
                     .post()
                     .uri(uri)
--- a/nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
+++ b/nifi-extension-bundles/nifi-azure-bundle/nifi-azure-registry-clients/src/main/java/org/apache/nifi/azure/devops/AzureDevOpsRepositoryClient.java
@@ -376,6 +376,10 @@ public class AzureDevOpsRepositoryClient implements GitRepositoryClient {
                 .addQueryParameter(API, API_VERSION)
                 .build();
 
+        final String authorName = request.getAuthorName();
+        final String authorEmail = request.getAuthorEmail();
+        final Author author = (authorName != null && authorEmail != null) ? new Author(authorName, authorEmail) : null;
+
         for (int attempt = 1; attempt <= MAX_PUSH_ATTEMPTS; attempt++) {
             if (expectedFileCommit != null) {
                 final Optional<String> currentFileCommit = getContentSha(request.getPath(), branch);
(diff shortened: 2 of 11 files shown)


exceptionfactory (changes requested):
Thanks for proposing this improvement @pvillard31.

There is certainly value is tracing the commit author, but as described and implemented, it requires the NiFi user identity to be an email address, which may not be the case in many situations. One possibility to attempting to determine if it is an email address.

Considering this from another perspective, what about adding some trailer to the commit message footer? One possibility is the common `Co-authored-by`. This still requires a name and email address for exact matching, but as a footer, it is more open to interpretation. The email could also be derived according to the Git service, such as `username@users.noreply.github.com` and `username@noreply.gitlab.com`. The approach only requires a NiFi username matching the Git service username. If it doesn't match, it is less of an issue.

I'm open to considering possible alternatives.

exceptionfactory:
Another benefit of an approach based on the message footer is that it avoids the current reflection-based workaround.

pvillard31:
> There is certainly value is tracing the commit author, but as described and implemented, it requires the NiFi user identity to be an email address, which may not be the case in many situations. One possibility to attempting to determine if it is an email address.

That is not true. Any string will work. Git providers will only try to do account matching if the provided string is an email. If not, it'll just use the string as-is. One of the screenshots is exactly that scenario.

> Considering this from another perspective, what about adding some trailer to the commit message footer? One possibility is the common Co-authored-by. This still requires a name and email address for exact matching, but as a footer, it is more open to interpretation. 

What I don't really like about this approach is that this is just some text added to the commit message. Firstly, users could add their own Co-authored-by in the comment of the flow version and secondly, the next step of this effort would be to display author/committer information in the UI when listing flow versions and relying on parsing comments would not be great IMO.

> The email could also be derived according to the Git service, such as username@users.noreply.github.com and username@noreply.gitlab.com. The approach only requires a NiFi username matching the Git service username. If it doesn't match, it is less of an issue.

I didn't know about this but not sure to really see the benefit to be honest. In a world where SSO is used in most places, I feel like just taking the identity would be fine.

> Another benefit of an approach based on the message footer is that it avoids the current reflection-based workaround.

Yeah, I'm not a big fan of that but that would be temporary. I'm trying to get the PR merged and have another rc release for the github client.

exceptionfactory:

For clarification on the NiFi user identity being an email address, I meant that it was a requirement in order to match against an actual Git user. I understand that any string is technically accepted in the author field.

It is true that user's could add their own `Co-authored-by` lines to the footer, however, GitHub and other services specifically parse the "trailer" of a commit message to determine co-authorship.

Taking a step back for a moment, I initially approached this change from the perspective of viewing commit authorship in GitHub or other Git service, as opposed to within NiFi. If the primary purpose is associating authorship in NiFi, that does change the analysis to some degree. My primary concern related to authorship linkage in the Git service, and the mismatch if the user identity is not an email address.

pvillard31:
Thanks for the feedback @exceptionfactory - I do believe that being able to surface the author in the UI would be the next logical step. This first step is to have authorship in the git provider so that things could be reflected on a pull request / git history / etc. But once the write path is done, the read path could be added to add this information in the NiFi when listing versions of a flow for example.

Good news, 2.0-rc6 with my changes was released so I pushed a commit to switch to that for Github and remove the reflection workaround.

Let me know what you think.

exceptionfactory (changes requested):
Thanks for the updates @pvillard31, good to see the latest version of the GitHub library does not need the reflection-based workaround.

With that resolved, my concern remains about the lack of linkage between the NiFi user identity and the corresponding Git user. In common cases where the NiFi user is not an email address, the commit author will have no possibility of linking to a real Git user.

Another case is where personal tokens are used, instead of service-based authentication. Although less common, that might be a case where setting the author is not preferable.

What do you think about introducing a property on the various Git Registry Clients? Perhaps something like `Commit Author Source`. The default value could be `Service User` and an optional value could be `Application User`, where `Application User` would enable the proposed behavior, setting the Author Name and Author Email to the NiFi user identity. As mentioned previously, GitHub and GitLab support "no-reply" addresses, so a potential future option could enable deriving that email address based on the Application User identity.

I'm open to other approaches on property-based configuration, but it seems like it should configurable, since it will be a somewhat visible change in behavior.

pvillard31:
I think having this behind a dedicated property is a fair request. I made the change and squashed the commits.

exceptionfactory (approved):
Thanks for adding the `Commit Author Source` property @pvillard31. With that change, the implementation looks good. +1 merging

PR #10986, opened 2026-03-09
NIFI-15671 - Flow import/export with stateful components state

pvillard31:
# Summary

NIFI-15671 - Flow import/export with stateful components state

This extends the JSON-based "Download flow definition" / "Upload flow definition" workflow to optionally include component state (processors and controller services annotated with `@Stateful`). A new `includeComponentState` query parameter on the download endpoint controls this. The feature does not apply to registry-based flow versioning.

**Key behavioral decisions:**

- **Opt-in via query parameter.** `GET /process-groups/{id}/download?includeComponentState=true` includes state; the default is `false`, preserving backward compatibility. The existing `ObjectMapper` configuration (`FAIL_ON_UNKNOWN_PROPERTIES = false`) means older NiFi versions silently ignore the new field.
- **Never registry versioning.** `FlowMappingOptions.DEFAULT_OPTIONS` (used by the registry path) has `mapComponentState=false`. This is architecturally enforced, not just a convention.
- **Components must be stopped.** Exporting with state requires all processors in the group to be stopped and all controller services disabled. If any are running/enabled, the endpoint returns `409 Conflict`. This guarantees state consistency.
- **Authorization escalation.** Exporting with state requires `WRITE` permission on the process group (accessing state requires write access). Without state, only `READ` is needed.
- **Upload only, not replace.** State is restored only for newly created components during upload. The replace endpoints (`PUT` and `POST .../replace-requests`) reject any flow definition containing component state with `400 Bad Request`, because replacing an existing group produces a mix of new/updated/unchanged components where selective state restoration would be confusing.
- **Cluster LOCAL state handling.** In a cluster, the export request is replicated to all nodes; each contributes its own LOCAL state keyed by ordinal position. The coordinator merges the responses. On import, each node picks up the state entry at its ordinal index. Ordinals are determined by sorting nodes by `address:port`.
- **Cluster topology validation on import.** If the flow definition contains local state from `S` source nodes and the destination cluster has `D` connected nodes: `S <= D` is allowed (extra nodes get no state), `S > D` is rejected with `409 Conflict`.
- **External services have state stripped.** When `includeReferencedServices=true`, controller services from parent groups are included for wiring but their state is set to null — state export is scoped to the target group only.
- **Only `@Stateful` components get state.** Non-stateful components always have null `componentState`, even when `includeComponentState=true`.

**Note:** UI changes would be as part of a follow-up effort.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -4145,19 +4145,20 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
 
         final StateManagerProvider stateManagerProvider = context.getStateManagerProvider();
         if (stateManagerProvider == null) {
-            LOG.debug("StateManagerProvider is not available; skipping state restoration for component {}", componentId);
+            LOG.warn("StateManagerProvider is not available; component state from the flow snapshot will not be restored for component {}", componentId);
             return;
         }
 
         final ConfigurableComponent component = componentNode.getComponent();
         if (component == null) {
-            LOG.debug("Component {} is not available; skipping state restoration", componentId);
+            LOG.warn("Component {} is not available; component state from the flow snapshot will not be restored", componentId);
             return;
         }
 
         final Stateful stateful = component.getClass().getAnnotation(Stateful.class);
         if (stateful == null) {
-            LOG.debug("Component {} ({}) is not annotated with @Stateful; skipping state restoration", componentId, component.getClass().getSimpleName());
+            LOG.warn("Component {} ({}) is not annotated with @Stateful; component state from the flow snapshot will not be restored",
+                    componentId, component.getClass().getSimpleName());
             return;
         }
 
@@ -4173,7 +4174,7 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
             if (supportedScopes.contains(Scope.LOCAL) && componentState.getLocalNodeStates() != null && !componentState.getLocalNodeStates().isEmpty()) {
                 final int localNodeOrdinal = context.getLocalNodeOrdinal();
                 if (localNodeOrdinal < 0) {
-                    LOG.debug("Local node ordinal is not set; skipping local state restoration for component {}", componentId);
+                    LOG.warn("Local node ordinal is not set; local component state from the flow snapshot will not be restored for component {}", componentId);
                     return;
                 }
 
@@ -4184,7 +4185,8 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
                     stateManager.setState(localState, Scope.LOCAL);
                     LOG.debug("Restored local state for component {} from node ordinal {}", componentId, localNodeOrdinal);
                 } else {
-                    LOG.debug("No local state found for component {} at node ordinal {}", componentId, localNodeOrdinal);
+                    LOG.info("No local component state was captured for node ordinal {} in the flow snapshot; local state for component {} will not be restored on this node",
+                            localNodeOrdinal, componentId);
                 }
             }
         } catch (final IOException e) {
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -17,10 +17,15 @@
 
 package org.apache.nifi.flow.synchronization;
 
+import org.apache.nifi.annotation.behavior.Stateful;
 import org.apache.nifi.asset.Asset;
 import org.apache.nifi.asset.AssetManager;
 import org.apache.nifi.bundle.BundleCoordinate;
+import org.apache.nifi.components.ConfigurableComponent;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.state.Scope;
+import org.apache.nifi.components.state.StateManager;
+import org.apache.nifi.components.state.StateManagerProvider;
 import org.apache.nifi.connectable.Connectable;
 import org.apache.nifi.connectable.ConnectableType;
 import org.apache.nifi.connectable.Connection;
@@ -57,6 +62,8 @@ import org.apache.nifi.flow.ExecutionEngine;
 import org.apache.nifi.flow.ParameterProviderReference;
 import org.apache.nifi.flow.VersionedAsset;
 import org.apache.nifi.flow.VersionedComponent;
+import org.apache.nifi.flow.VersionedComponentState;
+import org.apache.nifi.flow.VersionedConfigurableExtension;
 import org.apache.nifi.flow.VersionedConnection;
 import org.apache.nifi.flow.VersionedControllerService;
 import org.apache.nifi.flow.VersionedExternalFlow;
@@ -64,6 +71,7 @@ import org.apache.nifi.flow.VersionedFlowAnalysisRule;
 import org.apache.nifi.flow.VersionedFlowCoordinates;
 import org.apache.nifi.flow.VersionedFunnel;
 import org.apache.nifi.flow.VersionedLabel;
+import org.apache.nifi.flow.VersionedNodeState;
 import org.apache.nifi.flow.VersionedParameter;
 import org.apache.nifi.flow.VersionedParameterContext;
 import org.apache.nifi.flow.VersionedPort;
@@ -125,6 +133,8 @@ import org.apache.nifi.util.FlowDifferenceFilters;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.URL;
 import java.time.Duration;
 import java.util.ArrayList;
@@ -326,6 +336,8 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
         updatedVersionedComponentIds.clear();
         createdAndModifiedExtensions.clear();
         setSynchronizationOptions(options);
+
+        validateLocalStateTopology(versionedExternalFlow.getFlowContents());
 
         for (final FlowDifference diff : flowComparison.getDifferences()) {
             if (!FlowDifferenceFilters.isComponentUpdateRequired(diff, versionedExternalFlow.getFlowContents(), context.getFlowManager())) {
@@ -1438,6 +1450,8 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
         createdAndModifiedExtensions.add(new CreatedOrModifiedExtension(newService, decryptedProperties));
 
         updateControllerService(newService, proposed, topLevelGroup);
+
+        restoreComponentState(newService.getIdentifier(), proposed.getComponentState(), newService);
 
         return newService;
     }
@@ -2748,6 +2762,8 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
         final ProcessContext processContext = context.getProcessContextFactory().apply(procNode);
         procNode.onConfigurationRestored(processContext);
         connectableAdditionTracker.addComponent(destination.getIdentifier(), proposed.getIdentifier(), procNode);
+
+        restoreComponentState(procNode.getIdentifier(), proposed.getComponentState(), procNode);
 
         return procNode;
     }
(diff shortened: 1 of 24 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -4080,6 +4096,102 @@ private Map<String, String> getPropertyValues(final ComponentNode componentNode)
...
+    }
+
+    private void restoreComponentState(final String componentId, final VersionedComponentState componentState, final ComponentNode componentNode) {
+        if (componentState == null) {
+            return;
+        }
+
+        final StateManagerProvider stateManagerProvider = context.getStateManagerProvider();
+        if (stateManagerProvider == null) {
+            LOG.debug("StateManagerProvider is not available; skipping state restoration for component {}", componentId);
```
bobpaulin, line 4148: This makes sense that not having the state manager provider will prevent us from being able to import state.  Not blocking the import make sense since the user may want to proceed with out it. However I would challenge if this should be logged only at debug.  Shouldn't the user be notified if they were expecting to import state but it was not?  These seems like it would very likely impact the behavior after the flow is started.  I'd consider this may be a WARN level event.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -4080,6 +4096,102 @@ private Map<String, String> getPropertyValues(final ComponentNode componentNode)
...
+            return;
+        }
+
+        final ConfigurableComponent component = componentNode.getComponent();
+        if (component == null) {
+            LOG.debug("Component {} is not available; skipping state restoration", componentId);
```
bobpaulin, line 4154: WARN level event?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -4080,6 +4096,102 @@ private Map<String, String> getPropertyValues(final ComponentNode componentNode)
...
+            return;
+        }
+
+        final Stateful stateful = component.getClass().getAnnotation(Stateful.class);
+        if (stateful == null) {
+            LOG.debug("Component {} ({}) is not annotated with @Stateful; skipping state restoration", componentId, component.getClass().getSimpleName());
```
bobpaulin, line 4160: WARN level event?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -4080,6 +4096,102 @@ private Map<String, String> getPropertyValues(final ComponentNode componentNode)
...
+        try {
+            if (supportedScopes.contains(Scope.CLUSTER) && componentState.getClusterState() != null && !componentState.getClusterState().isEmpty()) {
+                stateManager.setState(componentState.getClusterState(), Scope.CLUSTER);
+                LOG.debug("Restored cluster state for component {}", componentId);
+            }
+
+            if (supportedScopes.contains(Scope.LOCAL) && componentState.getLocalNodeStates() != null && !componentState.getLocalNodeStates().isEmpty()) {
+                final int localNodeOrdinal = context.getLocalNodeOrdinal();
+                if (localNodeOrdinal < 0) {
+                    LOG.debug("Local node ordinal is not set; skipping local state restoration for component {}", componentId);
```
bobpaulin, line 4176: WARN level event?

bobpaulin (commented):
@pvillard31 This looks like an important change. Working on running some tests with this locally.  Provided a few code comments for now.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
```diff
@@ -2956,7 +2958,7 @@ public Set<String> getClusterMembers() {
     @Override
     public Optional<String> getCurrentNode() {
         if (isClustered() && getNodeId() != null) {
-            return Optional.of(getNodeId().getApiAddress());
+            return Optional.of("%s:%d".formatted(getNodeId().getApiAddress(), getNodeId().getApiPort()));
```
markap14: This is a risky change, as they break backward compatibility because it changes what is returned by the `NodeTypeProvider`. In particular, this affects `EmbeddedHazelcastCacheManager`.
After this change:

- `m + PORT_SEPARATOR + port` (line 141) becomes `"address:nifiApiPort:hazelcastPort"` — Hazelcast will fail to bind / connect.
- `hazelcastMembers.contains(getCurrentNode().get())` (line 152) compares user-configured hostnames against `"address:port"` and will always return `false`, sending the local node down the client path instead of becoming a member.
- The `customValidate` check at lines 224–236 compares user-configured Hazelcast hosts against getClusterMembers() and will always report them as "not part of the NiFi cluster".

GetAzureEventHub.getClientIdentifier() also folds the value into a client identifier, which is now silently changed. That is less likely to break things but is still a behavioral change.

The export/import feature only needs a stable, unique key for ordinal calculation. Mixing port into a public `NodeTypeProvider` method that other extensions already depend on is the wrong layer to fix it. One option: Introduce a new framework-internal helper that returns `Set<NodeIdentifier>`/sorted IDs and use it only from `StandardProcessGroup` / `StandardNiFiServiceFacade`. Leave `getClusterMembers()` / `getCurrentNode()` alone

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -4080,6 +4096,104 @@ private Map<String, String> getPropertyValues(final ComponentNode componentNode)
         return propertyValues;
     }
 
+    private void validateLocalStateTopology(final VersionedProcessGroup proposed) {
+        final int connectedNodeCount = context.getConnectedNodeCount();
```
markap14, line 4100: `connectedNodeCount` is a misleading name

This made says it compares "S source nodes" against "D connected nodes" (per the PR description), but the `connectedNodeCount` it reads is actually `Math.max(sortedMembers.size(), 1)` — the size of `getClusterMembers()`, which is all members regardless of connection state. So a 5-node cluster with only 2 connected nodes will accept a 4-node export and silently leave two of the four state slices unrestored. We should either rename the field/builder method to `clusterMemberCount` and document, or compute it from connected nodes only.

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/pg/ClusterFlowDefinitionExportImportStateIT.java
```diff
@@ -0,0 +1,330 @@
...
+        assertNotNull(state.getClusterState(), "Cluster state should be present");
+        assertNotNull(state.getClusterState().get("count"), "Cluster state should contain count");
+        assertNotNull(state.getLocalNodeStates(), "Local node states should be present");
+        assertEquals(2, state.getLocalNodeStates().size());
+        assertNotNull(state.getLocalNodeStates().get(0).getState().get("count"));
+        assertNotNull(state.getLocalNodeStates().get(1).getState().get("count"));
+    }
+
+    @Test
+    public void testClusterRoundTripSameTopologyBothScopes() throws NiFiClientException, IOException, InterruptedException {
```
markap14, line 141: This test only re-reads cluster state after import. It does not query each node's local state to confirm Node 0's exported state ends up on Node 0 (and not on Node 1). Add an assertion that uses getProcessorState(... Scope.LOCAL) per node.

markap14 (changes requested):
Thanks for adding the feature @pvillard31. I think there are a couple of issues that we need to address but they should be pretty quick to turnaround.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
pvillard31: Reverted the changes and added a new framework-internal ClusterTopologyProvider interface implemented by FlowController.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
pvillard31: `StandardProcessGroup` now derives both `localNodeOrdinal` and `connectedNodeCount` from `ClusterTopologyProvider`, which is backed by `NodeConnectionState.CONNECTED`, so the count reflects only connected nodes and matches the variable name.

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/pg/ClusterFlowDefinitionExportImportStateIT.java
pvillard31: Extended `testClusterRoundTripSameTopologyBothScopes` to capture per-node LOCAL state grouped by `clusterNodeId` before export and assert the same map after import, verifying each node's local state lands back on the correct node.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -5935,24 +5982,58 @@ private Set<String> getAllSubGroups(ProcessGroup processGroup) {
         return result;
     }
 
+    /**
+     * Computes the ordinal index of the local node among connected cluster nodes.
+     * Nodes are sorted deterministically by apiAddress. In standalone mode, returns 0.
+     *
+     * @return the ordinal index of the local node, or 0 if not clustered
+     */
+    private int computeLocalNodeOrdinal() {
```
markap14: This duplicates the logic added in FlowController.getLocalNodeOrdinal() in the same PR. The new ClusterTopologyProvider interface was introduced for exactly this purpose; the service facade should depend on it rather than reimplementing the ordinal calculation.

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/pg/ClusterFlowDefinitionExportImportStateIT.java
```diff
@@ -0,0 +1,350 @@
...
+        if (componentState != null && componentState.getLocalState() != null && componentState.getLocalState().getState() != null) {
+            for (final StateEntryDTO entry : componentState.getLocalState().getState()) {
+                final String nodeId = entry.getClusterNodeId();
+                byNode.computeIfAbsent(nodeId, id -> new HashMap<>()).put(entry.getKey(), entry.getValue());
+            }
+        }
+        return byNode;
+    }
+
+    private Map<String, String> getProcessorState(final String processorId, final Scope scope) throws NiFiClientException, IOException {
```
markap14, line 290: `waitForStatePopulated` and `getProcessorState` — the wait polls a flattened map and returns as soon as either node has populated count. The tests then assert two local-state slices are present at lines 104 (testClusterExportCapturesLocalStateFromBothNodes), 135 (testClusterExportCapturesBothScopes), and 155 / 172 (testClusterRoundTripSameTopologyBothScopes). If the second node has not yet executed onTrigger by the time the wait returns, those assertions will flake. The wait should explicitly verify both ordinals have state (e.g., getProcessorLocalStatesByNode(...).size() == 2) for the round-trip case.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ProcessGroupResource.java
```diff
@@ -323,6 +347,160 @@ public Response exportProcessGroup(
...
 
+    /**
+     * Replicates the flow export request to all cluster nodes and merges the LOCAL state from each node
+     * into a single response. Each node returns a RegisteredFlowSnapshot containing only its own LOCAL
+     * state (keyed by its ordinal). This method collects all node responses and combines the localNodeStates
+     * maps so the exported flow contains LOCAL state from every node.
+     *
+     * @return the merged response containing LOCAL state from all nodes
+     */
+    private Response replicateAndMergeFlowExport() {
```
markap14: Perhaps I'm missing something, but I think we should register an `EndpointResponseMerger` for `GET /process-groups/<uuid>/download`. This is the typical pattern that we follow to handle merging the responses. A merger would also let the framework drain unread response streams consistently.

markap14 (changes requested):
Hey @pvillard31 I think we're really close here; flagged a couple of other minor issues.

pvillard31:
Thanks @markap14 - I pushed a commit to address your feedback and proceeded with additional testing locally with the latest changes on standalone and clusters with local state and cluster state.

markap14:
Thanks for updating and for putting this together, @pvillard31 ! All looks good to me at this point. The system tests do a pretty good job of verifying basic functionality but I also tested it manually, using curl to export the flow with state. Visually inspected that it looked good, then used curl to import a new PG from it. Verified the state was copied to the new PG. Also then cleared the state on one, and ran both to ensure that they were properly updating independently without issue. All checks out +1 will merge!

PR #10994, opened 2026-03-11
NIFI-15696 - Add Rebase operation for versioned process groups

pvillard31:
# Summary

NIFI-15696 - Add Rebase operation for versioned process groups

When a versioned process group is at version N with local modifications (state LOCALLY_MODIFIED_AND_STALE), upgrading to a newer version requires reverting all local changes first, then changing version, then manually re-applying every customization. This is error-prone, causes unnecessary downtime, and discourages users from upgrading.

This change adds a "Rebase" operation that upgrades a versioned process group to a newer registry version while automatically preserving compatible local changes — analogous to git rebase.

**How it works:**

The rebase operation has two phases:

- Analysis (read-only): Computes a three-way diff between the local flow, the base registry version (N), and the target registry version. Each local change is classified as Compatible (will be preserved), Conflicting (same field changed both locally and upstream to different values), or Unsupported (change type not yet supported by a rebase handler). Rebase is only allowed when all local changes are compatible.
- Execution: Deep-clones the target version snapshot, overlays compatible local changes onto it to produce a merged snapshot, then synchronizes the process group to the merged snapshot using the existing FlowUpdateResource async infrastructure. The VCI is set to the clean target version so that subsequent state checks correctly report LOCALLY_MODIFIED.

Supported change types (for this first iteration, this will evolve over time):

- POSITION_CHANGED, SIZE_CHANGED, BENDPOINTS_CHANGED — cosmetic, always compatible (local wins)
- PROPERTY_CHANGED, PROPERTY_ADDED — compatible unless upstream changed the same property on the same component to a different value (convergent changes where both sides set the same value are accepted)
- COMMENTS_CHANGED — compatible unless upstream also changed comments on the same component
Unsupported change types (examples: COMPONENT_ADDED, COMPONENT_REMOVED) block the rebase with a clear message. The handler architecture is extensible — adding support for a new change type requires only implementing a single RebaseHandler class.

We follow a similar approach as what already exists for the new API endpoints:

- GET /versions/rebase-analysis/process-groups/{id}?targetVersion={version}
returns the three-way analysis
- POST /versions/rebase-requests/process-groups/{id}
initiates async rebase execution
- GET/DELETE /versions/rebase-requests/{id}
polling and cleanup (reuses FlowUpdateResource)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
@@ -217,6 +217,7 @@ import org.apache.nifi.registry.flow.diff.RebaseAnalysis;
 import org.apache.nifi.registry.flow.diff.RebaseEngine;
 import org.apache.nifi.registry.flow.diff.StandardComparableDataFlow;
 import org.apache.nifi.registry.flow.diff.StandardFlowComparator;
+import org.apache.nifi.registry.flow.diff.StandardRebaseEngine;
 import org.apache.nifi.registry.flow.diff.StaticDifferenceDescriptor;
 import org.apache.nifi.registry.flow.mapping.ComponentIdLookup;
 import org.apache.nifi.registry.flow.mapping.FlowMappingOptions;
@@ -6597,14 +6598,13 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
             throw new NiFiCoreException("Failed to retrieve flow from Flow Registry in order to perform rebase analysis due to " + e.getMessage(), e);
         }
 
-        final NiFiRegistryFlowMapper mapper = makeNiFiRegistryFlowMapper(controllerFacade.getExtensionManager());
+        final VersionedComponentFlowMapper mapper = makeNiFiRegistryFlowMapper(controllerFacade.getExtensionManager());
         final VersionedProcessGroup localGroup = mapper.mapProcessGroup(processGroup, controllerFacade.getControllerServiceProvider(), controllerFacade.getFlowManager(), true);
 
-        final Set<String> ancestorServiceIds = processGroup.getAncestorServiceIds();
 
         final ComparableDataFlow registryFlow = new StandardComparableDataFlow("Versioned Flow", currentRegistryGroup);
         final ComparableDataFlow localFlow = new StandardComparableDataFlow("Local Flow", localGroup);
-        final FlowComparator localComparator = new StandardFlowComparator(registryFlow, localFlow, ancestorServiceIds,
+        final FlowComparator localComparator = new StandardFlowComparator(registryFlow, localFlow,
                 new ConciseEvolvingDifferenceDescriptor(), Function.identity(), VersionedComponent::getIdentifier, FlowComparatorVersionedStrategy.SHALLOW);
         final FlowComparison localComparison = localComparator.compare();
 
@@ -6619,7 +6619,7 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
 
         final ComparableDataFlow currentFlow = new StandardComparableDataFlow("Current Version", currentRegistryGroup);
         final ComparableDataFlow targetFlow = new StandardComparableDataFlow("Target Version", targetRegistryGroup);
-        final FlowComparator upstreamComparator = new StandardFlowComparator(currentFlow, targetFlow, ancestorServiceIds,
+        final FlowComparator upstreamComparator = new StandardFlowComparator(currentFlow, targetFlow,
                 new ConciseEvolvingDifferenceDescriptor(), Function.identity(), VersionedComponent::getIdentifier, FlowComparatorVersionedStrategy.SHALLOW);
         final FlowComparison upstreamComparison = upstreamComparator.compare();
 
@@ -6640,7 +6640,7 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
             descendantFailureReason = e.getMessage();
         }
 
-        final RebaseEngine rebaseEngine = new RebaseEngine();
+        final RebaseEngine rebaseEngine = new StandardRebaseEngine();
         final RebaseAnalysis analysis = rebaseEngine.analyze(localDifferences, upstreamDifferences, targetRegistryGroup);
 
         if (descendantHasLocalModifications) {
@@ -6690,6 +6690,7 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
 
         final VersionedProcessGroup currentRegistryGroup;
         final FlowSnapshotContainer targetSnapshotContainer;
+        final VersionedProcessGroup cleanTargetSnapshot;
         try {
             final FlowSnapshotContainer currentSnapshotContainer = flowRegistry.getFlowContents(
                     FlowRegistryClientContextFactory.getContextForUser(NiFiUserUtils.getNiFiUser()), currentVersionLocation, true);
@@ -6697,18 +6698,24 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
 
             targetSnapshotContainer = flowRegistry.getFlowContents(
                     FlowRegistryClientContextFactory.getContextForUser(NiFiUserUtils.getNiFiUser()), targetVersionLocation, true);
+
+            // Retrieve a second, independent copy of the target version to retain as the clean snapshot for the Version
+            // Control Information. The snapshot passed to the RebaseEngine is mutated in place to build the merged
+            // snapshot, so a separate clean copy of the target version is required.
+            final FlowSnapshotContainer cleanTargetSnapshotContainer = flowRegistry.getFlowContents(
+                    FlowRegistryClientContextFactory.getContextForUser(NiFiUserUtils.getNiFiUser()), targetVersionLocation, true);
+            cleanTargetSnapshot = cleanTargetSnapshotContainer.getFlowSnapshot().getFlowContents();
         } catch (final IOException | FlowRegistryException e) {
             throw new NiFiCoreException("Failed to retrieve flow from Flow Registry for rebase due to " + e.getMessage(), e);
         }
 
         final VersionedProcessGroup targetRegistryGroup = targetSnapshotContainer.getFlowSnapshot().getFlowContents();
-        final NiFiRegistryFlowMapper mapper = makeNiFiRegistryFlowMapper(controllerFacade.getExtensionManager());
+        final VersionedComponentFlowMapper mapper = makeNiFiRegistryFlowMapper(controllerFacade.getExtensionManager());
         final VersionedProcessGroup localGroup = mapper.mapProcessGroup(processGroup, controllerFacade.getControllerServiceProvider(), controllerFacade.getFlowManager(), true);
-        final Set<String> ancestorServiceIds = processGroup.getAncestorServiceIds();
 
         final ComparableDataFlow registryFlow = new StandardComparableDataFlow("Versioned Flow", currentRegistryGroup);
         final ComparableDataFlow localFlow = new StandardComparableDataFlow("Local Flow", localGroup);
-        final FlowComparator localComparator = new StandardFlowComparator(registryFlow, localFlow, ancestorServiceIds,
+        final FlowComparator localComparator = new StandardFlowComparator(registryFlow, localFlow,
                 new ConciseEvolvingDifferenceDescriptor(), Function.identity(), VersionedComponent::getIdentifier, FlowComparatorVersionedStrategy.SHALLOW);
         final FlowComparison localComparison = localComparator.compare();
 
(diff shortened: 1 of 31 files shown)


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/VersionsResource.java
```diff
@@ -996,6 +998,186 @@ public Response deleteRevertRequest(
...
+                    @ApiResponse(responseCode = "400", description = "NiFi was unable to complete the request because it was invalid. The request should not be retried without modification."),
+                    @ApiResponse(responseCode = "401", description = "Client could not be authenticated."),
+                    @ApiResponse(responseCode = "403", description = "Client is not authorized to make this request."),
+                    @ApiResponse(responseCode = "404", description = "The specified resource could not be found."),
+                    @ApiResponse(responseCode = "409", description = "The request was valid but NiFi was not in the appropriate state to process it.")
+            },
+            description = "For a Process Group that is under Version Control, this will perform a rebase analysis by comparing "
+                    + "local modifications against upstream changes between the current version and the specified target version. "
+                    + "The analysis determines whether a rebase is allowed or if there are conflicts.",
```
exceptionfactory: Multiline strings can be used instead of concatenation

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/VersionsResource.java
```diff
@@ -996,6 +998,186 @@ public Response deleteRevertRequest(
...
+                    + "the endpoint will immediately return a VersionedFlowUpdateRequestEntity, and the process of rebasing the flow will occur "
+                    + "asynchronously in the background. The client may then periodically poll the status of the request by issuing a GET request to "
+                    + "/versions/rebase-requests/{requestId}. Once the request is completed, the client is expected to issue a DELETE request to "
+                    + "/versions/rebase-requests/{requestId}. " + NON_GUARANTEED_ENDPOINT,
+            security = {
+                    @SecurityRequirement(name = "Read - /process-groups/{uuid}"),
+                    @SecurityRequirement(name = "Write - /process-groups/{uuid}"),
+                    @SecurityRequirement(name = "Read - /{component-type}/{uuid} - For all encapsulated components"),
+                    @SecurityRequirement(name = "Write - /{component-type}/{uuid} - For all encapsulated components"),
+                    @SecurityRequirement(name = "Write - if the template contains any restricted components - /restricted-components"),
```
exceptionfactory: This reference to restricted components is no longer applicable and should be removed

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/RebaseAnalysis.java
```diff
@@ -0,0 +1,105 @@
...
+            this.classification = Objects.requireNonNull(classification, "Classification is required");
+            this.conflictCode = conflictCode;
+            this.conflictDetail = conflictDetail;
+        }
+
+        public static ClassifiedDifference compatible(final FlowDifference difference) {
+            return new ClassifiedDifference(difference, RebaseClassification.COMPATIBLE, null, null);
+        }
+
+        public static ClassifiedDifference conflicting(final FlowDifference difference, final String conflictCode, final String conflictDetail) {
```
exceptionfactory: Should the `conflictCode` be declared as an `enum` for reuse across the implementation?

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/RebaseEngine.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.HashMap;
+import java.util.HexFormat;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+public class RebaseEngine {
```
exceptionfactory: I recommend defining an interface and making this the standard implementation to clarify the public contract for engine operations

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/RebaseEngine.java
```diff
@@ -0,0 +1,159 @@
...
+    public static String computeConflictKey(final FlowDifference difference) {
+        final DifferenceType type = difference.getDifferenceType();
+        final String componentId = resolveComponentIdentifier(difference);
+        final Optional<String> fieldName = difference.getFieldName();
+
+        return switch (type) {
+            case PROPERTY_CHANGED -> type.name() + ":" + componentId + ":" + fieldName.orElse("");
+            case POSITION_CHANGED, SIZE_CHANGED, COMMENTS_CHANGED, DESCRIPTION_CHANGED -> type.name() + ":" + componentId;
+            case BENDPOINTS_CHANGED -> type.name() + ":" + componentId;
+            default -> type.name() + ":" + componentId + ":" + fieldName.orElse("");
```
exceptionfactory: The concatenation here is a bit hard to read, recommend defining one or more format strings and reusing

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/RebaseEngine.java
```diff
@@ -0,0 +1,159 @@
...
+        };
+    }
+
+    VersionedProcessGroup deepClone(final VersionedProcessGroup source) {
+        try {
+            final byte[] serialized = OBJECT_MAPPER.writeValueAsBytes(source);
+            return OBJECT_MAPPER.readValue(serialized, VersionedProcessGroup.class);
```
exceptionfactory: This seems like a bit of a hack to round trip and object through Jackson serialization. I recommend either explicit copy methods, or perhaps something like Spring BeanUtils

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/RebaseEngine.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private static String resolveComponentIdentifier(final FlowDifference difference) {
+        if (difference.getComponentB() != null) {
+            return difference.getComponentB().getIdentifier();
+        }
+        if (difference.getComponentA() != null) {
+            return difference.getComponentA().getIdentifier();
+        }
+        return "";
```
exceptionfactory: Is empty string a valid return?

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/main/java/org/apache/nifi/registry/flow/diff/RebaseEngine.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private String computeAnalysisFingerprint(final List<RebaseAnalysis.ClassifiedDifference> classifiedChanges,
+                                              final Set<FlowDifference> upstreamDifferences) {
+        final SortedSet<String> sortedKeys = new TreeSet<>();
+
+        for (final RebaseAnalysis.ClassifiedDifference classified : classifiedChanges) {
+            final String key = computeConflictKey(classified.getDifference());
+            sortedKeys.add("local:" + key + ":" + classified.getClassification().name());
```
exceptionfactory: See other notes regarding standard formatting for keys

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/test/java/org/apache/nifi/registry/flow/diff/TestPropertyChangedRebaseHandler.java
```diff
@@ -0,0 +1,250 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class TestPropertyChangedRebaseHandler {
```
exceptionfactory: Since these are net-new classes, recommend using `Test` as the suffix instead of the prefix for the class. The public modifiers can also be removed

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/src/test/java/org/apache/nifi/registry/flow/diff/TestPropertyChangedRebaseHandler.java
```diff
@@ -0,0 +1,250 @@
...
+    private PropertyChangedRebaseHandler handler;
+
+    @BeforeEach
+    public void setup() {
+        handler = new PropertyChangedRebaseHandler();
+    }
+
+    @Test
+    public void testNoUpstreamConflict() {
+        final VersionedProcessor processor = createProcessorWithProperty("proc-a", "propX", "oldVal");
```
exceptionfactory: There are a number of repeated string values in this test that should be moved to static final variables

nifi-registry/nifi-registry-core/nifi-registry-flow-diff/pom.xml
```diff
@@ -29,5 +29,9 @@
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
```
exceptionfactory: Is this dependency needed if the ObjectMapper deep copy is removed?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -468,6 +473,8 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
     private static final int VALIDATION_WAIT_MILLIS = 50;
     private static final String ROOT_PROCESS_GROUP = "RootProcessGroup";
 
+    private final Map<String, VersionedProcessGroup> rebaseCleanSnapshots = new ConcurrentHashMap<>();
```
exceptionfactory: Are there any concerns about concurrent rebase requests for the same Process Group?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/VersionsResource.java
```diff
@@ -1237,8 +1419,12 @@ protected ProcessGroupEntity performUpdateFlow(final String groupId, final Revis
         versionControlInfo.setVersion(metadata.getVersion());
         versionControlInfo.setState(flowSnapshot.isLatest() ? VersionedFlowState.UP_TO_DATE.name() : VersionedFlowState.STALE.name());
 
-        return serviceFacade.updateProcessGroupContents(revision, groupId, versionControlInfo, flowSnapshot, idGenerationSeed,
+        final ProcessGroupEntity result = serviceFacade.updateProcessGroupContents(revision, groupId, versionControlInfo, flowSnapshot, idGenerationSeed,
                 verifyNotModified, false, updateDescendantVersionedFlows);
+
+        serviceFacade.resetVersionControlSnapshotAfterRebase(groupId);
```
exceptionfactory: It seems like a failure to update the Process Group contents would fail to reset the snapshot, leaving the snapshot entry in the Map within the Standard Service Facade. Perhaps putting this in a `finally` block would be sufficient?

exceptionfactory (changes requested):
Thanks for putting together this new feature @pvillard31. The general implementation pattern looks good, and the `Handler` strategy for various types of changes appears to provide a good foundation for iteration.

I noted a handful of syntactic recommendations, and raised a couple questions about concurrency.

pvillard31:
Thanks for the review @exceptionfactory - it had been a while since I submitted the PR so I addressed your feedback + the rebase on a commit and I made some additional changes in another commit on top.

exceptionfactory (approved):
Thanks for the work on this new feature @pvillard31, the adjustments and changes look good! +1 merging

PR #11034, opened 2026-03-23
NIFI-15570: Track count of Content Claim References for claims that could be truncated to ensure no premature truncation

markap14:
# Summary

[NIFI-15570](https://issues.apache.org/jira/browse/NIFI-15570)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/repository/FileSystemRepository.java
```diff
@@ -1032,6 +1048,128 @@ public void purge() {
...
+                }
+            }
+
+            claimsSkipped.forEach(truncationClaimManager::addTruncationClaims);
+        }
+
+        private boolean isTruncationActiveForContainer(final String container, final Map<String, Boolean> activationCache) {
+            // If not archiving data, we consider truncation always active.
+            if (!archiveData) {
+                return true;
```
exceptionfactory, line 1117: This method has quite a number of `return` statements, what do you think about refactoring to a single return, or breaking it up in some other way?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/repository/FileSystemRepository.java
```diff
@@ -1032,6 +1048,128 @@ public void purge() {
...
+                LOG.debug("Truncation is not active for container {} because the archive was not cleared on the last run", container);
+                activationCache.put(container, false);
+                return false;
+            }
+
+            final long usableSpace;
+            try {
+                usableSpace = getContainerUsableSpace(container);
+            } catch (final IOException ioe) {
+                LOG.warn("Failed to determine usable space for container {}. Will not truncate claims for this container", container, ioe);
```
exceptionfactory, line 1133:
```suggestion
                LOG.warn("Failed to determine usable space for container [{}] Will not truncate claims for this container", container, ioe);
```

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/repository/FileSystemRepository.java
```diff
@@ -1882,19 +2045,23 @@ public synchronized void close() throws IOException {
                 // Mark the claim as no longer being able to be written to
                 resourceClaimManager.freeze(scc.getResourceClaim());
 
+                final boolean largeClaim = scc.getLength() > Math.min(1_000_000, maxAppendableClaimLength);
```
exceptionfactory, line 2048: It looks like the claim length value being checked can be computed once with `Math.min()` on instantiation, instead of checking on every method invocation for the lesser of the configured value and 1 MB.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/repository/FileSystemRepository.java
```diff
@@ -1882,19 +2045,23 @@ public synchronized void close() throws IOException {
...
+                final boolean nonStartClaim = scc.getOffset() > 0;
+                scc.setTruncationCandidate(truncationEnabled && largeClaim && nonStartClaim);
```
exceptionfactory: for readability and debugging, it seems useful to define `boolean truncationCandidate` instead of the inline logical and expression.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/repository/WriteAheadFlowFileRepository.java
```diff
@@ -786,6 +910,38 @@ public long loadFlowFiles(final QueueProvider queueProvider) throws IOException
...
+                for (final SerializedRepositoryRecord record : recordList) {
+                    final ContentClaim claim = record.getContentClaim();
+                    if (claim instanceof final StandardContentClaim standardContentClaim && !standardContentClaim.isTruncationCandidate() && markedCandidates.contains(claim)) {
+                        standardContentClaim.setTruncationCandidate(true);
+                    }
+                }
+            }
+        } else {
+            logger.info("Skipping truncation candidate marking because {} swap files were recovered; " +
+                "content claims referenced by swapped FlowFiles are not counted", recoveredSwapLocations.size());
```
exceptionfactory, line 942: This message seems rather verbose, given the string concatenation, can it be shortened?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/TestFileSystemRepository.java
```diff
@@ -897,14 +901,376 @@ protected boolean archive(Path curPath) {
...
+            }
+        } finally {
+            localRepository.shutdown();
+        }
+    }
+
+    @Test
+    @Timeout(60)
+    public void testTruncateNotActiveWhenDiskNotPressured() throws IOException, InterruptedException {
+        // Create repository with ample disk space
```
exceptionfactory: Is this comment intended for the repository? It doesn't fit above the `shutdown()` call.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/TestFileSystemRepository.java
```diff
@@ -897,14 +901,376 @@ protected boolean archive(Path curPath) {
...
+
+            assertTrue(largeClaim.isTruncationCandidate());
+
+            final Path filePath = getPath(localRepository, smallClaim);
+            final long originalSize = Files.size(filePath);
+
+            localClaimManager.decrementClaimantCount(largeClaim.getResourceClaim());
+            localClaimManager.markTruncatable(largeClaim);
+
+            Thread.sleep(3000L);
```
exceptionfactory, line 1085: Is there a reason for this 3 second sleep? Any way to make it shorter, or less time-dependent?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/TestFileSystemRepository.java
```diff
@@ -897,14 +901,376 @@ protected boolean archive(Path curPath) {
...
+            assertEquals(originalSize, Files.size(filePath));
+        } finally {
+            localRepository.shutdown();
+        }
+    }
+
+    @Test
+    @Timeout(90)
+    public void testTruncateClaimDeferredThenExecutedWhenPressureStarts() throws IOException, InterruptedException {
+        // Create a repository where disk pressure can be toggled
```
exceptionfactory: Similar to above comment, this seems misplaced.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/TestFileSystemRepository.java
```diff
@@ -897,14 +901,376 @@ protected boolean archive(Path curPath) {
...
+            assertTrue(forceEnqueued.isEmpty(), "markTruncatable should have rejected the claim due to positive ref count");
+
+            // Force-enqueue by decrementing to zero, marking truncatable, then re-incrementing
+            // before the background task can run. This simulates the Layer 3 scenario.
+            localClaimManager.decrementTruncationReferenceCount(largeClaim);
+            localClaimManager.markTruncatable(largeClaim);
+            localClaimManager.incrementTruncationReferenceCount(largeClaim);
+
+            // Wait long enough for the TruncateClaims background task to have executed at least once
+            Thread.sleep(3_000);
```
exceptionfactory: If 3 seconds is the desired amount across tests, it should be declared once and reused.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/TestFileSystemRepository.java
```diff
@@ -897,14 +901,376 @@ protected boolean archive(Path curPath) {
...
+
+            // File should NOT have been truncated because TruncateClaims checks ref count > 0
```
exceptionfactory: The comment and assert message are duplicative.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/TestFileSystemRepository.java
```diff
@@ -897,14 +901,376 @@ protected boolean archive(Path curPath) {
...
+
+    @Test
+    public void testTruncationDisabledPreventsSettingTruncationCandidate() throws IOException {
+        recreateRepositoryWithPropertyOverrides(Map.of(
+                NiFiProperties.REPOSITORY_CONTENT_PREFIX + "default", rootFile.toString(),
+                NiFiProperties.CONTENT_CLAIM_TRUNCATION_ENABLED, "false"));
+
+        final ContentClaim smallClaim = repository.create(false);
+        try (final OutputStream out = repository.write(smallClaim)) {
+            out.write(new byte[100]);
```
exceptionfactory, line 1265: This `100` value is used in multiple places.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/TestWriteAheadFlowFileRepository.java
```diff
@@ -810,4 +810,751 @@ public String changePartitionName(String swapLocation, String newPartitionName)
...
+    // Truncation Feature: Runtime Tests
+    // =========================================================================
+
+    @Test
+    public void testDeleteRecordRoutesTruncatableClaimToTruncationQueue() throws IOException {
+        final RuntimeRepoContext context = createRuntimeRepoContext();
+        final ResourceClaim resourceClaim = context.claimManager().newResourceClaim("container", "section", "1", false, false);
+        context.claimManager().incrementClaimantCount(resourceClaim);
+        context.claimManager().incrementClaimantCount(resourceClaim); // count = 2 so that after delete decrement it stays > 0 (not destructable)
+        final StandardContentClaim contentClaim = createClaim(resourceClaim, 1024L, 5_000_000L, true);
```
exceptionfactory, line 939: All of these new test methods have numerous repeated values that should be declared as constants and reused.

nifi-framework-bundle/nifi-framework/nifi-repository-models/src/main/java/org/apache/nifi/controller/repository/claim/StandardContentClaim.java
```diff
@@ -29,6 +29,7 @@ public final class StandardContentClaim implements ContentClaim, Comparable<Cont
     private final ResourceClaim resourceClaim;
     private final long offset;
     private volatile long length;
+    private volatile boolean truncationCandidate = false;
```
exceptionfactory, line 32: The `false` initialized value should be unnecessary

exceptionfactory (changes requested):
Thanks for revisiting this new feature, adding the configuration property, and addressing the previous issues @markap14. I plan to do some additional review, but I noted a handful of mostly minor things, primarily recommendations related to the test methods.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/repository/FileSystemRepository.java
markap14: I don't think having a small number of return statements should be a goal. I find this method very clear personally.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/repository/TestWriteAheadFlowFileRepository.java
markap14: I find named constants to be distracting and significantly reducing code readability so I avoid them for constants unless they meet 2 conditions: (a) the value is repeated at least twice; and (b) changing one of them necessitates or strongly influences the need to change the others.

Having several constants throughout the code that happen to be the same is perfectly fine if they are entirely independent of one another. So analysis of the repeated constants shows:

`Offsets` (1024L, 2048L, 100L, (i + 1) * 1024L): arbitrary positions per test; not coupled across tests. This should be left as-is

5_000_000L: used in ~9 runtime tests, always paired with true (pre-marked candidate). All represent the same concept: "a claim length large enough to be a truncation candidate (> 1 MB threshold)." If the threshold ever changed, all of these must be updated together. These should be extracted into a constant.

2_000_000L: used in ~5 recovery tests, but several of those tests also encode an adjacent offset (e.g. 2_000_100L = 100L + 2_000_000L). Extracting the length without also coupling the offsets would introduce inconsistency within those tests. These should be left as-is to avoid fragile cross-coupling.

`new byte[100]`: appears in five tests as "a small amount of data for a non-truncatable claim" but each is entirely independent

exceptionfactory (commented):
Thanks for the updates and responses @markap14, this looks close to completion. However, I noticed 2 of the 4 system tests had timeout-related failures related to the new tests, so it looks like those should be reviewed for stability improvements.

markap14:
Good call @exceptionfactory. I found that there was an issue with one of the Processors created for system tests, which could cause flakiness in the system tests. Fixed it, and all is looking much healthier now.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/repository/FileSystemRepository.java
exceptionfactory: Thanks for the reply, I think it depends on the implementation, and it being generally helpful to avoid short-circuit returns, although I acknowledge that may not be the best approach in every situation. I could go either way on this particular one, so leaving it sounds good.

exceptionfactory (approved):
Thanks for making the adjustments @markap14, the latest version looks good. +1 merging

PR #11129, opened 2026-04-11
NIFI-14572 Fix time offset issue in DateTimeAdapter#unmarshal

superdachuan:
# Summary

[NIFI-14572](https://issues.apache.org/jira/browse/NIFI-14572)

This PR fixes a time offset issue in `DateTimeAdapter#unmarshal`. The current implementation uses `ZonedDateTime`, which can lead to ambiguous parsing of time zone abbreviations such as `CST`, resulting in incorrect offsets. This change aligns the implementation with `TimestampAdapter` by replacing `ZonedDateTime` with `LocalDateTime` and converting it using the system default time zone, avoiding such ambiguity. Although this issue may appear related to NIFI-14581, they are not the same problem and have different root causes: NIFI-14581 is caused by inconsistencies between the time zone ID returned by `TimezoneAdapter` and the server's default locale, whereas this PR addresses ambiguity in parsing time zone abbreviations. Therefore, this change focuses specifically on fixing the time offset issue in `DateTimeAdapter` and does not address NIFI-14581.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-client-dto/src/test/java/org/apache/nifi/web/api/dto/util/TestDateTimeAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestDateTimeAdapter {
+
+    private static final String TEST_DATE_TIME = "2026-01-02T03:04:56Z";
+
+    @Test
+    public void testMarshal() throws Exception {
+        DateTimeAdapter adapter = new DateTimeAdapter();
```
exceptionfactory: The Adapter instance can be declared once at the class level

nifi-framework-bundle/nifi-framework/nifi-client-dto/src/test/java/org/apache/nifi/web/api/dto/util/TestDateTimeAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+        Date date = Date.from(Instant.parse(TEST_DATE_TIME));
```
exceptionfactory: It would be helpful to avoid the intermediate parsing to an `Instant`. Instead, using the `new Date(long)` constructor should provide a more narrow approach to construct a Date with a given value.

nifi-framework-bundle/nifi-framework/nifi-client-dto/src/test/java/org/apache/nifi/web/api/dto/util/TestDateTimeAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+        assertEquals(date, adapter.unmarshal(adapter.marshal(date)));
```
exceptionfactory: This is testing both `marshal` and `unmarshal` in one call. Instead, there should be an expected static final string value, matching the `long` timestamp.

exceptionfactory (changes requested):
Thanks for working on this issue @superdachuan.

The functional approach looks good, aligning with `TimestampAdapter` as described.

I noted several recommendations on the test class, which could use some refinement to ensure expected behavior.

superdachuan:
@exceptionfactory Thanks for the review and helpful suggestions. I updated the tests accordingly.

My original intent in combining `marshal` and `unmarshal` was to verify round-trip consistency, since the adapter output includes a time zone and I initially wanted to avoid relying on a single fixed string. I have kept that intent as explicit round-trip tests, and also added more direct assertions for multiple time zones.

I had originally avoided temporarily switching the default time zone because I was concerned about affecting other tests. After reviewing similar patterns in the codebase, I updated this test to verify behavior across multiple time zones directly.
One detail here is that `DateTimeAdapter` initializes its `DateTimeFormatter` during class loading with `ZoneId.systemDefault()`, so changing the default time zone later does not update the formatter. For that reason, the multi-time-zone assertions reload `DateTimeAdapter` with a dedicated `ClassLoader` instead of reusing the default adapter instance from this test class.

Thanks again.

exceptionfactory (changes requested):
Thanks for the updates @superdachuan. Unfortunately, the new testing approach goes too far in the direction of complexity, and implementing a custom ClassLoader is far more than needed for testing this class. Please revert the change and pursue a simpler approach.

superdachuan:
Thanks for the feedback @exceptionfactory . I reverted the custom ClassLoader-based test change and simplified the approach.

In the updated revision, I removed `withZone(ZoneId.systemDefault())` from the static `DateTimeFormatter` in `DateTimeAdapter`. In this adapter, that formatter-level zone was not necessary for the existing marshal/unmarshal behavior, so removing it does not change the intended time zone handling. It does, however, avoid binding the formatter to the JVM default time zone at class initialization time.

With that change in place, the tests can switch the default time zone directly without relying on class reloading or a custom ClassLoader.

superdachuan:
The current failed jobs do not appear to be failing on the same test, and local testing with JDK 25 passed on my side, so this may be flaky CI rather than a reproducible issue with this change.

I do not have permission to re-run failed jobs on this PR. Could a committer or maintainer please re-run the failed checks?

If the failures persist and indicate an issue in this change, I will investigate and update the PR.

exceptionfactory (approved):
Thanks again for the work on this @superdachuan. I pushed a change to the unit test class to use `ParameterizedTest`. I'm planning to merge on successful build completion.

superdachuan:
Thanks @exceptionfactory, I appreciate the update and the improvement to the test coverage. I'll take a closer look at the ParameterizedTest changes so I can follow this pattern better in future contributions.

PR #11130, opened 2026-04-12
NIFI-15829 - Support parameter value references to parameters from inherited parameter contexts

pvillard31:
# Summary

NIFI-15829 - Support parameter value references to parameters from inherited parameter contexts

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/parameter/ParameterContext.java
```diff
@@ -188,4 +188,22 @@ public interface ParameterContext extends ParameterLookup, ComponentAuthorizable
...
+
+    /**
+     * Extracts the referenced parameter name from a one-to-one parameter value reference.
+     * A one-to-one reference is a parameter value whose entire content is exactly <code>#{parameterName}</code>
+     * with no surrounding text.
+     *
+     * @param value the parameter value to check
+     * @return the referenced parameter name if the value is a one-to-one reference, or null otherwise
+     */
+    static String extractOneToOneParameterReference(final String value) {
```
markap14: I'm not sure that this util method should really be on the interface. It would make sense to pull this out into a separate util class.

nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/parameter/ParameterContext.java
```diff
@@ -188,4 +188,22 @@ public interface ParameterContext extends ParameterLookup, ComponentAuthorizable
...
+        if (value == null || value.length() < 4) {
+            return null;
+        }
+        if (value.startsWith("#{") && value.endsWith("}") && value.indexOf('}') == value.length() - 1) {
```
markap14: This is a bit overly simplistic. It does not account for things like quoted parameter names. So it will work with `#{MyParam}` and properly catches issues like `#{a}b}}` But it misses cases like `#{'My Parameter'}`
We should instead probably make use of `ExpressionLanguageAgnosticParameterParser`, and do something along the lines of:
```
final ParameterParser parser = new ExpressionLanguageAgnosticParameterParser();
final ParameterTokenList tokenList = parser.parseTokens(value);
final List<ParameterToken> tokens = tokenList.toList();
if (tokens.size() != 1) {
  logger.warn("Invalid Parameter Reference: " + value);
}
final ParameterToken firstToken = tokens.getFirst();
if (!firstToken.isParameterReference()) {
  logger.warn("Invalid Parameter Reference: " + value);
}

return tokenList.toReferenceList().getFirst().getParameterName();
```

markap14 (changes requested):
Thanks @pvillard31 I think the approach makes sense. I did flag a couple of concerns about how we're handling the parsing specifically but otherwise I think this is a reasonable approach. It would probably also be good to go ahead and update the User Guide as well.

nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/parameter/ParameterContext.java
pvillard31: Yeah, fair, I went for the easiest approach but you're absolutely right, especially with provided parameters, we may have names where simple quoting will be cleaner.

pvillard31:
Thanks @markap14 - I pushed a commit to address your comments

markap14 (approved):
Thanks @pvillard31 changes look good to me. +1 will merge on successful Github Actions completions.

PR #11141, opened 2026-04-14
NIFI-13591 Added support for multiline JSON in JSLTTransformJSON.

dan-s1:
# Summary

[NIFI-13591](https://issues.apache.org/jira/browse/NIFI-13591)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
+++ b/nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
@@ -22,10 +22,8 @@ import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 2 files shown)


exceptionfactory (changes requested):
On initial review, I'm concerned about the `Multiline Transformation Strategy` property, as it makes it more difficult to reason over the Processor configuration in general. I don't think the additional property is necessary, and instead, selecting JSON Lines as the format would effectively handle each object.

dan-s1:
I added the `Multiline Transformation Strategy` because I was under the impression we had to support the same options for each JSON line as we do when the JSON is in one FlowFile which was was operating on the whole JSON object or each individual object if the JSON line contains an array.  If that is not the case, I have no problem removing it.

exceptionfactory:
> I added the `Multiline Transformation Strategy` because I was under the impression we had to support the same options for each JSON line as we do when the JSON is in one FlowFile which was was operating on the whole JSON object or each individual object if the JSON line contains an array. If that is not the case, I have no problem removing it.

Thanks for clarifying, I don't think it is necessary when JSON Lines is selected.

dan-s1:
I have removed the use of the `Multiline Transformation Strategy`

exceptionfactory (approved):
Thanks @dan-s1, the latest version looks good. +1 merging

PR #11146, opened 2026-04-15
NIFI-15843 Remove unused methods from internal StringUtils

exceptionfactory:
# Summary

[NIFI-15843](https://issues.apache.org/jira/browse/NIFI-15843) Removes unused methods from the internal `StringUtils` class located in the `nifi-property-utils` module.

The class has multiple methods from the Apache Commons Lang 3 implementation, several of which are not used or have minimal references. The class was intended for use within `nifi-property-utils` but references have grown beyond the original design. Removing unused methods is a first step to correcting usage scope across the project.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-commons/nifi-property-utils/src/main/java/org/apache/nifi/util/StringUtils.java
dan-s1, line 33: Could this be simplified to

```suggestion
public static boolean isBlank(final String str) {
return str == null || str.isBlank();
}
```

nifi-commons/nifi-property-utils/src/main/java/org/apache/nifi/util/StringUtils.java
```diff
@@ -92,422 +91,6 @@ public static String join(final Collection<?> collection, String delimiter) {
...
-    public static String toTitleCase(String input) {
-        if (input == null || isBlank(input)) {
-            return "";
-        }
-        List<String> elements = Arrays.asList(input.trim().toLowerCase().split("\\s"));
-        return elements.stream()
-                .filter(word -> !isBlank(word))
-                .map(word -> Character.toTitleCase(word.charAt(0)) + word.substring(1))
-                .collect(Collectors.joining(" "));
+        return sb.substring(0, sb.lastIndexOf(delimiter));
```
dan-s1, line 86: This method is mostly used in unit tests. Could it be replaced with Java's `String.join`?

nifi-commons/nifi-property-utils/src/main/java/org/apache/nifi/util/StringUtils.java
dan-s1, line 55: Could this be simplified to

```suggestion
public static boolean startsWith(final String str, final String prefix) {
if (str == null || prefix == null) {
return (str == null && prefix == null);
}
if (prefix.length() > str.length()) {
return false;
}
return str.startsWith(prefix);
}
```

dan-s1 (changes requested):
I whole heartdly agree with this refactoring I just had some other possible places where the code could be reduced.

nifi-commons/nifi-property-utils/src/main/java/org/apache/nifi/util/StringUtils.java
exceptionfactory: Yes, I considered that, but wanted to keep this initial set of changes focused more narrowly. I'm planning on a follow-on PR to address more widespread usage of remaining methods.

exceptionfactory:
Thanks for the feedback @dan-s1. I pushed an update to simply two of the methods, and responded on the comment regarding the join method.

pvillard31 (approved):

PR #11160, opened 2026-04-17
NIFI-15820: Adding an expandable side panel with a Connector info car…

mcgilman:
…d to the Connector canvas.

## Summary

Introduces an expandable side panel on the connector canvas that surfaces a `Connector Info` control with inline connector management. The side panel uses a `mat-sidenav` in `side`/`disableClose` mode, toggled from the header bar, with its open/closed preference persisted via the `Storage` service so the layout is remembered across sessions.

The new `connector-info-control` card displays connector state and queued-data metrics, and exposes action buttons for:

- **Start** / **Stop** — transition the connector between `RUNNING` and `STOPPED`.
- **Drain** / **Cancel Drain** — initiate draining of queued FlowFiles and cancel an in-flight drain. Drain prompts for confirmation; Cancel Drain does not.

Buttons are kept rendered and toggled via `disabled` state (rather than unmounted) while actions are in-flight to avoid the info panel briefly shrinking to a single row and causing a flicker. Start/Stop are rendered as icon buttons; drain controls use `mat-stroked-button`; all row-one actions are left-aligned.

A new `connector-detail-header` shared component provides the name/type/state header and is reused inside the info control.

## Connector-scoped error banner

To keep API failures from the info card visible (and not silently dropped), a `CONNECTOR_CANVAS` `ErrorContextKey` was added and a `<context-error-banner>` is mounted as a floating overlay over the reusable canvas. A new effect clears prior `CONNECTOR_CANVAS` banner errors on each action dispatch so stale errors do not leak between operations or across pages. Banner chrome (background, outline, border-radius, shadow) is applied via a new `_connector-canvas.component-theme.scss` partial included from `styles.scss`, avoiding `::ng-deep` while participating in light/dark theming.

## Changes

- **State** — New `connector-canvas-entity` feature slice (`actions`, `reducer`, `effects`) handling load, start, stop, drain, cancel-drain, and a unified `connectorActionApiError` action routed to the new canvas banner context.
- **UI**
  - `connector-canvas.component` mounts the sidenav, graph controls, and floating error banner.
  - `connector-graph-controls.component` hosts the collapsible panel content.
  - `connector-info-control.component` renders the connector detail header, metrics, and actions.
  - `connector-detail-header.component` added under `libs/shared` for reuse.
- **Styling** — New component theme partial wired into `styles.scss` for both light and dark themes.
- **Error model** — `CONNECTOR_CANVAS` added to `ErrorContextKey`.


rfellows:
Reviewing...

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-canvas/connector-canvas.component.ts
```diff
@@ -64,6 +77,10 @@ export class ConnectorCanvasComponent implements OnInit, OnDestroy {
     private destroyRef = inject(DestroyRef);
     private router = inject(Router);
     private dialog = inject(MatDialog);
+    private storage = inject(Storage);
+
+    private static readonly CONTROL_VISIBILITY_KEY = 'graph-control-visibility';
+    private static readonly GRAPH_CONTROL_KEY = 'connector-graph-controls';
```
rfellows, line 81: This duplicates `GRAPH_CONTROLS_STORAGE_KEY` on line 55 above. We should get rid of one of them. At this point, `GRAPH_CONTROLS_STORAGE_KEY` is unused and actually spawns a lint warning.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-canvas/connector-canvas.component.scss
```diff
@@ -0,0 +1,27 @@
...
+context-error-banner.canvas-error-banner {
+    ::ng-deep .banner-container {
+        background-color: var(--mat-sys-surface-container);
+        border: 1px solid var(--mat-sys-outline-variant);
+        border-radius: 4px;
+        box-shadow: 0 1px 6px rgba(0, 0, 0, 0.25);
+        color: var(--mat-sys-on-surface);
+        margin-bottom: 0;
+    }
+}
```
rfellows: This file isn't leveraged/referenced in the `@Component` ts file and it duplicates the same styles as in `_connector-canvas.component-theme.scss`... can we just delete this file since it seems to be unsed?

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-canvas/graph-controls/connector-info-control/connector-info-control.component.ts
```diff
@@ -0,0 +1,213 @@
...
+        return canOperateConnector(entity) && isConnectorActionAllowed(entity, 'DRAIN_FLOWFILES');
+    }
+
+    canCancelDrain(): boolean {
+        const entity = this.connectorEntity();
+        if (!entity) {
+            return false;
+        }
+        return canOperateConnector(entity) && isConnectorActionAllowed(entity, 'CANCEL_DRAIN_FLOWFILES');
+    }
```
rfellows: These 2 `canXXX` methods differ in pattern to the others in this file in that these don't consider the `this.entitySaving()`. They are more similar to the `showXXX` methods. I see the `entitySaving` is instead used on the DOM elements to control disabled state. Maybe tighten up the pattern? I think we are getting what the desired effect is, but the logic isn't consistent.

rfellows (changes requested):
A few things...

rfellows (approved):
Looks good, @mcgilman 👍

PR #11166, opened 2026-04-21
NIFI-15863 - User edits to newly-introduced component properties silently classified as environmental changes

pvillard31:
# Summary

NIFI-15863 - User edits to newly-introduced component properties silently classified as environmental changes

When a newer bundle declares a static property that did not exist in the versioned flow snapshot, `FlowDifferenceFilters.isPropertyAddedFromMigration` (added in [NIFI-15776](https://issues.apache.org/jira/browse/NIFI-15776)) classifies the resulting `PROPERTY_ADDED` diff as environmental regardless of value. If the user later edits that property the diff relative to the registry is still `PROPERTY_ADDED` (the property descriptor is absent from the snapshot), so the edit is also swallowed as environmental: the flow is not marked dirty and the change cannot be committed through normal local-change workflows.

[NIFI-15830](https://issues.apache.org/jira/browse/NIFI-15830) refined related behavior but only addressed cases detected through `isNewPropertyWithDefaultValue`, which compares valueB against the descriptor default. It did not tighten `isPropertyAddedFromMigration`, which continues to return true for any statically-declared property missing from the snapshot, independent of the value. As a result, user edits away from the default on a newly-introduced property remain indistinguishable from the framework-applied default and are still filtered out.

We need to tighten `isPropertyAddedFromMigration` to only treat a `PROPERTY_ADDED` diff as environmental when valueB equals the descriptor's default value. This preserves the [NIFI-15776](https://issues.apache.org/jira/browse/NIFI-15776) / [NIFI-15830](https://issues.apache.org/jira/browse/NIFI-15830) behavior for bundle upgrades (framework-applied defaults remain filtered), while surfacing genuine user edits as local changes.

Narrow regression: if a component's `migrateProperties` explicitly assigns a new static property to a non-default value, that value will now show as a local change and the user must commit once after upgrade.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
@@ -997,7 +997,7 @@ public class FlowDifferenceFilters {
             componentNode = null;
         }
 
-        if (!isNotDynamicProperty(fieldName.get(), componentNode)) {
+        if (isDynamicProperty(fieldName.get(), componentNode)) {
             return false;
         }
 
@@ -1009,12 +1009,12 @@ public class FlowDifferenceFilters {
         return Objects.equals(difference.getValueB(), descriptor.getDefaultValue());
     }
 
-    private static boolean isNotDynamicProperty(final String propertyName, final ComponentNode componentNode) {
+    private static boolean isDynamicProperty(final String propertyName, final ComponentNode componentNode) {
         final ConfigurableComponent component = componentNode == null ? null : componentNode.getComponent();
         final List<PropertyDescriptor> descriptors = component == null ? List.of() : component.getPropertyDescriptors();
         return descriptors.stream()
                 .map(PropertyDescriptor::getName)
-                .anyMatch(propertyName::equals);
+                .noneMatch(propertyName::equals);
     }
 
     private static Optional<String> getComponentInstanceIdentifier(final FlowDifference difference) {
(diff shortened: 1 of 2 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
```diff
@@ -992,7 +997,16 @@ private static boolean isPropertyAddedFromMigration(final FlowDifference differe
             componentNode = null;
         }
 
-        return isNotDynamicProperty(fieldName.get(), componentNode);
+        if (!isNotDynamicProperty(fieldName.get(), componentNode)) {
```
markap14: It looks like this is the only place we're using this method. Can we rename to `isDynamicProperty` and invert it? I don't like negations in method names but we definitely shouldn't be negating the result of a method called `isNotDynamicProperty`

markap14 (changes requested):
Thanks @pvillard31 conceptually changes make sense. Just had one concern about the method name. Otherwise looks good.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/util/FlowDifferenceFilters.java
pvillard31: Fair, pushed a commit to remove the negation

markap14 (approved):

PR #11175, opened 2026-04-23
NIFI-15880: Allow Connectors to enter a Troublehshooting mode where the flow beco…

markap14:
…mes modifiable and user-managed (temporarily) rather than Connector managed.

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
```diff
@@ -637,12 +663,155 @@ public void verifyCanDelete() {
...
+
+        return authoritativeFlow;
+    }
+
+    private void verifyAllComponentsStoppedAndDisabled(final ProcessGroup group) {
+        for (final ProcessorNode processor : group.getProcessors()) {
+            if (!STOPPED_STATES.contains(processor.getScheduledState())) {
+                throw new IllegalStateException("Cannot end Troubleshooting mode for " + this + " because Processor " + processor.getIdentifier()
+                    + " is in state " + processor.getScheduledState() + "; it must be STOPPED or DISABLED.");
+            }
```
rfellows: The way this works currently is a bit confusing. You can enter troubleshooting mode while the processor is running, but you can't end troubleshooting with it still running. This forces the user to have to manually stop all processors before they would be allowed to end troubleshooting. For large connectors, this will be a tedious process of manually stopping everything.

It might be more palatable if STOP was a valid action on the connector as a whole while in troubleshooting mode. But it is explicitly not allowed in the current state.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
rfellows, line 1: ConnectorRepository.getConnectors() got a required ConnectorSyncMode parameter. But FlowController is still calling getConnectors() with no args. Results in a compile error.

kevdoran (changes requested):
## Bugs / functional concerns

### 1. Cluster flow sync of an already-TROUBLESHOOTING connector will fail with `IllegalStateException`

`StandardConnectorNode.setName` now throws if the connector is in TROUBLESHOOTING.

```140:155:nifi/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
@Override
public void setName(final String name) {
    if (getCurrentState() == ConnectorState.TROUBLESHOOTING) {
        throw new IllegalStateException("Cannot rename " + this + " while it is in Troubleshooting mode; exit Troubleshooting mode before renaming the Connector.");
    }
    this.name = name;
}
```

But `StandardConnectorRepository.syncConnector` calls `connector.setName(effectiveName)` unconditionally and `TROUBLESHOOTING` is neither in `isWaitableState` nor `isRejectableState`:

```209:225:nifi/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorRepository.java
        // Set name locally (no provider.save())
        connector.setName(effectiveName);

        final boolean wasRunning = currentState == ConnectorState.RUNNING;
        final boolean configChanged = isNewConnector || isConfigurationUpdated(connector, effectiveActiveConfig, effectiveWorkingConfig);
        final boolean restoringTroubleshooting = effectiveScheduledState == VersionedConnectorState.TROUBLESHOOTING;
```

On a fresh startup it's fine because the connector node is freshly created in `STOPPED` (then `restoreTroubleshootingState()` flips it at the end). However, in a clustered configuration when a node receives a sync after disconnect/reconnect (or any other path where `syncConnector` runs against an existing connector that is *already* `TROUBLESHOOTING` in memory), `setName` will throw and the exception will propagate out of `VersionedFlowSynchronizer` unchecked — flow sync fails hard instead of returning a `ConnectorSyncResult.failed`.

This is the same family of regressions that earlier rebases hit (see commit `8842121`). Suggested fixes:

- In `syncConnector`, skip the `setName` when `currentState == TROUBLESHOOTING && effectiveScheduledState == VersionedConnectorState.TROUBLESHOOTING && Objects.equals(effectiveName, connector.getName())`.
- Or have the rename-guard live in `ConnectorRepository.updateConnector` (which is the public mutation path) and not in `setName` so that internal sync paths can still set the name on the node. The repository already has the same guard; the node-level one is duplicative.

This is the most concerning functional bug I found. Worth a system test that exercises clustered sync while the connector is in TROUBLESHOOTING (the new IT covers single-node restart in `testEndTroubleshootingBlockedByQueuedData`, but not multi-node sync).

### 2. `verifyEnterTroubleshooting` is run twice; providers can observe two calls per transition

`StandardConnectorDAO.enterTroubleshooting` → `repository.enterTroubleshooting` → `verifyEnterTroubleshooting(connector)` (which calls `configurationProvider.verifyEnterTroubleshooting`). The REST resource also calls `serviceFacade.verifyEnterConnectorTroubleshooting(id)` as the verify phase of `withWriteLock`, which calls `dao.verifyEnterTroubleshooting(id)` → `repository.verifyEnterTroubleshooting(connector)` (again calling the provider).

For the default Apache no-op this is harmless. For third-party `ConnectorConfigurationProvider` implementations that back this with an external round-trip (e.g., to refuse troubleshooting while the upstream system thinks the connector is in a transient state), it will fire twice per transition. Worth either (a) caching/idempotency in the provider implementation, or (b) the framework taking care to only verify once.

The same shape exists for `endTroubleshooting`: `verifyCanEndTroubleshooting` is run by the verify phase (cheaply now), and then `connector.endTroubleshooting()` calls it a second time, including the expensive `initializationContext.verifyUpdateFlow(...)` preflight. Not strictly a bug, but it is a redundant heavy call on the critical path.

### 3. `getReasonCannotEndTroubleshooting` runs the expensive preflight inside `getAvailableActions`

```1568:1573:nifi/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
private ConnectorAction createEndTroubleshootingAction() {
    final Optional<String> reason = getReasonCannotEndTroubleshooting();
    return new StandardConnectorAction("END_TROUBLESHOOTING", "Exit Troubleshooting mode for the connector", reason.isEmpty(), reason.orElse(null));
}
```

`getReasonCannotEndTroubleshooting` does two things:
1. A scan of *every* processor/port/RPG/CS in the managed flow tree.
2. If that passes, it invokes the Connector's `getActiveFlow(FlowContext)` and runs `verifyUpdateFlow(...)` against the managed PG (a full flow-comparison preflight).

`getAvailableActions` is called by the GET connector endpoint (i.e., on every refresh of the Connector entity). If the user sits on the Connector while in TROUBLESHOOTING, the UI may refresh the entity frequently. Once everything is stopped, every refresh runs the full preflight on a non-trivial flow. Suggest computing `reason` only on demand (e.g., lazily, or return a coarse-grained "ready/not-ready" by checking component states only and reserve the preflight for the actual verify endpoint). At minimum gate the preflight behind a flag or compute it only when called from the REST verify path.

### 4. `applyUpdate` blocked by TROUBLESHOOTING surfaces an unfriendly `IllegalStateException`

```555:577:nifi/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorRepository.java
@Override
public void applyUpdate(final ConnectorNode connector, final ConnectorUpdateContext context) throws FlowUpdateException {
    logger.debug("Applying update to {}", connector);

    if (connector.getCurrentState() == ConnectorState.TROUBLESHOOTING) {
        throw new IllegalStateException("Cannot apply an update to " + connector + " while it is in Troubleshooting mode; "
            + "exit Troubleshooting mode before applying updates.");
    }

    if (configurationProvider != null && !configurationProvider.shouldApplyUpdate(connector.getIdentifier())) {
```

This is potentially the most user-visible touchpoint for any external system orchestrating connectors via the REST API: external API clients that call `apply-update` will now fail with an HTTP error (depends on the global exception mapper for `IllegalStateException`, but it's typically 400 in NiFi) any time the connector happens to be in TROUBLESHOOTING. It would be more correct to:

- Map this to HTTP 409 specifically. `transitionStateForUpdating` and `setName` should throw a typed exception (e.g., extending `IllegalStateException` but mapped at the REST layer to 409 with a stable code) so that API client retry policies can distinguish "transient — retry" from "user owns this, leave it alone".
- Add an `applyUpdate` short-circuit verification method on the facade so that callers can ask first, rather than swallowing a failing POST.

This is more of a maintainability ask, but it has clear downstream impact for any client that drives connector updates programmatically.

### 5. `StandardConnectorDAO.endTroubleshooting` wraps `FlowUpdateException` with `+ e`

```184:191:nifi/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardConnectorDAO.java
@Override
public void endTroubleshooting(final String id) {
    final ConnectorNode connector = getConnector(id);
    try {
        getConnectorRepository().endTroubleshooting(connector);
    } catch (final FlowUpdateException e) {
        throw new IllegalStateException("Failed to exit troubleshooting mode for Connector " + id + ": " + e, e);
    }
}
```

`+ e` inlines `e.toString()` (class name + message) inside the IllegalStateException message AND `e` is also chained as the cause. Use `e.getMessage()` for the user-facing message, or omit the inlined message entirely. Minor, but it makes the error popup uglier.

### 6. `restoreFlowPreservingIdentifiers` uses `mapAssetReferences(false)`

```3993:4007:nifi/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
// Sensitive property values in the proposed snapshot were encrypted using the same PropertyEncryptor when the snapshot
// was persisted (for example, when a Connector-managed flow is persisted in Troubleshooting mode). The currently loaded
// flow therefore must also be mapped with an equivalent SensitiveValueEncryptor so the comparison between "current" and
// "proposed" sensitive values operates on matching ciphertext; otherwise every sensitive property appears to differ and
// the decrypted value written back to the live component is the encrypted payload rather than the plaintext (or parameter
// reference) that was originally captured.
final FlowMappingOptions flowMappingOptions = new FlowMappingOptions.Builder()
    .mapSensitiveConfiguration(true)
    .mapPropertyDescriptors(true)
    .stateLookup(stateLookup)
    .sensitiveValueEncryptor(encryptor::encrypt)
    .componentIdLookup(ComponentIdLookup.VERSIONED_OR_GENERATE)
    .mapInstanceIdentifiers(true)
    .mapControllerServiceReferencesToVersionedId(true)
    .mapFlowRegistryClientId(false)
    .mapAssetReferences(false)
    .build();
```

Connectors backed by external configuration providers commonly rely on `ASSET_REFERENCE` translation for parameter/property bindings, and the managed PG may embed parameter or asset references (e.g., for controller services). If a user edits a controller service in Troubleshooting mode and changes an asset reference, will the restore-from-persisted-snapshot correctly preserve it? With `mapAssetReferences(false)` the comparator side won't normalize them. I am not 100% certain this is wrong, but it's worth confirming with @markap14 explicitly whether assets edited while in TROUBLESHOOTING survive a NiFi restart correctly (the new IT does not appear to cover this).

### 7. `endConnectorTroubleshooting` doesn't re-check cluster connectivity at mutation time

```3868:3879:nifi/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
@Override
public void verifyEndConnectorTroubleshooting(final String id) {
    final List<NodeIdentifier> unconnectedNodes = getUnconnectedNodes();
    if (!unconnectedNodes.isEmpty()) {
        throw new IllegalStateException("Cannot exit Troubleshooting mode because the following cluster nodes are not CONNECTED: "
                + unconnectedNodes + ". All nodes must be CONNECTED before this operation may proceed.");
    }
    connectorDAO.verifyEndTroubleshooting(id);
}
```

The cluster check is only in `verifyEnd…`. A node can disconnect between `verify` and `endConnectorTroubleshooting` — small TOCTOU. Low risk because `withWriteLock` will replicate the action while still holding the lock, but worth a comment/defense in depth.

### 8. `findOwningConnector` walks `getParent()`, but child groups carry a back-reference too

Just a heads-up that `flowManager.getGroup(id)` is now used in `findProcessGroup` (good fix), but `findOwningConnector` walks parents in pure Java rather than asking the connector repository for the connector ID directly. For deep PG hierarchies inside a managed flow this is O(depth) per locate; combined with the `for each connector: managedGroup.findFunnel(funnelId)` pattern in the new DAO methods, you have O(connectors × tree-walk) per component lookup. Fine for the typical "one or a few Connectors" install, but if a deployment runs many connectors on a single node, this is a hotspot.

---

## Maintainability concerns

1. **`ProcessGroup.findOwningConnector()`** is added on the interface (`ProcessGroup.java` +21 lines). Every other `ProcessGroup` implementation (`MockProcessGroup`, etc.) now has to implement it. The implementation in `StandardProcessGroup` is sensible. Consider providing a default implementation that uses `getConnectorIdentifier()` + `getParent()` so that other implementers don't have to repeat the loop.

2. **`ComponentDAO`/`AuthorizableLookup` API explosion.** Every `get*` method now has a `(id, boolean includeConnectorManaged)` overload across 6 DAOs and 5 `AuthorizableLookup` getters. The single boolean is fine for now but it's an API smell — most callers will pass `false`, the only callers that pass `true` are the Connector-scoped REST endpoints. An alternative would be a single `ConnectorAwareLookup.getInsideManagedFlow(...)` facade that doesn't pollute every interface. Worth a follow-up.

3. **`VersionedComponentFlowMapper.mapConnector(ConnectorNode)` (1-arg)** is preserved as a backward-compat overload that silently does NOT serialize the managed PG even when the connector is in TROUBLESHOOTING. This is a latent footgun for any future caller. Either inline it, deprecate it with `@Deprecated(forRemoval=true)`, or have it call the 2-arg overload with a `ControllerServiceProvider` resolved from the connector node itself.

4. **`endTroubleshooting` on the Repository does not notify the provider.** `enterTroubleshooting` calls `provider.verifyEnterTroubleshooting`, but there is no `provider.endTroubleshooting(connectorId)` or similar lifecycle hook. For the Apache default this is fine, but third-party `ConnectorConfigurationProvider` implementations may want a way to observe transitions back out of TROUBLESHOOTING. If you anticipate any need for that signal, it's better to add the hook now — adding a lifecycle method later is awkward because the default behavior would need to remain no-op for backward compatibility.

5. **Javadoc references to `ScheduledState` left behind.** `ConnectorRepository.syncConnector`'s Javadoc still says `ConnectorConfigurationProvider#getSyncDirective(String, org.apache.nifi.flow.ScheduledState)` — should be `VersionedConnectorState`:

```70:78:nifi/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/components/connector/ConnectorRepository.java
* <p>When a {@link ConnectorConfigurationProvider} is configured, the provider's
* {@link ConnectorConfigurationProvider#getSyncDirective(String, org.apache.nifi.flow.ScheduledState)}
* method is called to obtain a {@link ConnectorSyncDirective} that may override the
* working configuration, name, or ScheduledState from the versioned flow.</p>
```

6. **Typo in API Javadoc.** `ConnectorRepository.endTroubleshooting`:

```176:185:nifi/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/components/connector/ConnectorRepository.java
* @throws IllegalStateException if the Connector cannot be transitioned out of Troubleshooting mode because it doe snot adhere
```

"doe snot adhere" → "does not adhere".

7. **The reviewer's earlier comment from rfellows is still not addressed.** rfellows observed that you can *enter* Troubleshooting with components running, but you cannot *exit* until everything is stopped/disabled, leaving the user to stop everything manually. A bulk "stop all components and exit" action would dramatically improve UX. The PR partially addressed this by surfacing the reason in `END_TROUBLESHOOTING.reason`, but didn't add the bulk action. For a connector with many processors this UX gap will be felt; worth either adding it now or filing a follow-up.

8. **DAO performance asymmetry.** `StandardProcessorDAO`/`StandardControllerServiceDAO` use `flowManager.getProcessorNode(id)` / `flowManager.getControllerServiceNode(id)` (O(1) lookups). `StandardFunnelDAO`, `StandardInputPortDAO`, `StandardOutputPortDAO`, `StandardLabelDAO`, `StandardRemoteProcessGroupDAO` iterate `connectorRepository.getConnectors(LOCAL_ONLY)` and walk each managed group. For consistency it would be nicer to have `FlowManager` index these by id too, but I understand why it wasn't done in this PR — out of scope.

9. **`StandardConnectorRepository.endTroubleshooting` doesn't run a `verifyEndTroubleshooting`-style provider hook.** Compare to `enterTroubleshooting(connector)` which calls `verifyEnterTroubleshooting(connector)` first. The DAO already calls `connector.verifyCanEndTroubleshooting()` separately, but the repo method ought to call its own `verifyEndTroubleshooting(connector)` (which doesn't exist yet) for symmetry — even if the provider hook is no-op. Otherwise the only entry point that verifies is the DAO, and any future caller that goes straight to the repo loses the check.

10. **`.cursor/rules/code-style.mdc`** is committed as part of the PR. That's a 76-line code-style guide that probably shouldn't ship with an upstream Apache PR. (May just be a leftover.)

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardProcessGroupDAO.java
rfellows, line 1: To support enabling/disabling controller services for a process group in a connector there are a couple of changes that will need to be made in this file to properly locate the process group:

* in `scheduleComponents`, line 305 should probably be:
```java
        final ProcessGroup group = locateProcessGroup(flowController, groupId, true);
```

* in `activateControllerServices`, line 414 should probably be:
```java
        final ProcessGroup group = locateProcessGroup(flowController, groupId, true);
```

Otherwise, the lookup fails with a 400 error indicating it "Cannot activate Controller Services with IDs [] because the associated Process Group (id=[[pgid]]) could not be found"

kevdoran (approved):
Thanks for those changes @markap14! Looks good from my end, although there are some system tests that are failing

rfellows:
Heads-up on a related backend bug I've been hitting while testing this branch end-to-end against the Connector REST endpoints. Flagging it here because it's the same family as the `findProcessGroup` fix you just landed in `42fbdd9`, and the fix would likely sit in the same area — but happy to split it into its own JIRA if you'd prefer.

### Symptom

`GET /nifi-api/processors/{id}/descriptors?propertyName=<csTypedProp>&sensitive=false` returns `"allowableValues": []` for any controller-service-typed property on a processor that lives inside a Connector's managed flow — including when the processor's currently configured CS exists in that connector's group and is returned by `GET /nifi-api/flow/process-groups/{parentGroupId}/controller-services`. The same empty `allowableValues` shows up on CS-typed descriptors embedded in `GET /processors/{id}` and on the descriptor returned by the inline "create new service" round-trip from `PropertyTableHelper`. Flow-designer is unaffected; only the Connector context reproduces it (regardless of whether the connector is in `TROUBLESHOOTING` — I see it in normal `RUNNING` state too).

Downstream UI effect: the value cell renders the CS UUID instead of its name, the combo box can't enumerate other compatible CSes, and a freshly-created inline CS also comes back as a UUID with no allowable-values context.

### Trace

`StandardNiFiServiceFacade.getProcessorPropertyDescriptor` →

```java
return dtoFactory.createPropertyDescriptorDto(descriptor, processor.getProcessGroup().getIdentifier());
```

`DtoFactory#createPropertyDescriptorDto` (CS-typed branch) →

```java
final List<String> controllerServiceIdentifiers = new ArrayList<>(
    controllerServiceProvider.getControllerServiceIdentifiers(serviceDefinition, groupId));
```

`StandardControllerServiceProvider#getControllerServiceIdentifiers`:

```java
ProcessGroup group = getRootGroup();
if (!FlowManager.ROOT_GROUP_ID_ALIAS.equals(groupId) && !group.getIdentifier().equals(groupId)) {
    group = group.findProcessGroup(groupId);
}

if (group == null) {
    return Collections.emptySet();
}
```

Even with the `findProcessGroup` change from `42fbdd9` applied, the call from `mainRoot.findProcessGroup(connectorManagedPgId)` still returns null:

- `flowManager.getGroup(id)` (unconditional) finds the managed PG fine.
- `isOwner(group.getParent())` then walks parents. The connector's managed root only has `setExplicitParentAuthorizable(connectorNode)` set on it in `StandardFlowManager.createConnector(...)` — there is no `setParent(...)` linking it under the main root group. So `getParent()` returns `null`, the `isOwner` loop terminates immediately, and `findProcessGroup` returns null.
- `getControllerServiceIdentifiers` short-circuits to `Collections.emptySet()` before the class/scope filter ever runs — hence the totally empty `allowableValues` array (not "wrong CSes filtered out").

The same `getRootGroup().findProcessGroup(...)` pattern exists in a handful of other places in `StandardControllerServiceProvider` (`getControllerServiceNodes(String)` and friends) and would presumably miss connector-managed groups in the same way for any caller that resolves a group from a connector-context id starting at main root.

### Suggested shape of the fix

Smallest change that matches the direction this PR has been heading (i.e. the new `ConnectorManagedComponentLookup` facades and the `locateProcessGroup(..., includeConnectorManaged=true)` calls in `StandardProcessGroupDAO`): inside `StandardControllerServiceProvider#getControllerServiceIdentifiers(serviceType, groupId)`, replace the `getRootGroup().findProcessGroup(groupId)` traversal with a direct `flowManager.getGroup(groupId)` lookup. The existing class filter (`service.getProxiedControllerService().getClass()` / `isAssignableFrom`) and the per-group `getControllerServices(true, false)` enumeration still do the right thing once the group is reachable. Auth has already happened on the owning processor, so the `isOwner` hierarchy gate isn't load-bearing for this code path.

### Repro

1. Open a Connector whose managed flow contains a processor with a CS-typed property already configured to point at a CS that lives in the same connector group (e.g. PublishKafkaRecord → KafkaConnectionService).
2. `GET /nifi-api/processors/{procId}/descriptors?propertyName=Kafka%20Connection%20Service&sensitive=false` → response below.
3. `GET /nifi-api/flow/process-groups/{parentGroupId}/controller-services` → returns the CS with full details, confirming it exists and is enabled.

```json
{
  "propertyDescriptor": {
    "name": "Kafka Connection Service",
    "displayName": "Kafka Connection Service",
    "allowableValues": [],
    "identifiesControllerService": "org.apache.nifi.kafka.service.api.KafkaConnectionService",
    "identifiesControllerServiceBundle": {
      "group": "org.apache.nifi",
      "artifact": "nifi-kafka-service-api-nar",
      "version": "2.10.0-SNAPSHOT"
    },
    "dependencies": []
  }
}
```

Reproducible regardless of Troubleshooting state — the connector was in normal `RUNNING` state when I captured this. Happy to file a separate JIRA if that's a cleaner home; just wanted to flag it here since you're already in the area and the fix would fit the same shape.

markap14:
[claude-opus-4.6] @rfellows Thanks for the detailed trace — the analysis is spot-on. The `getRootGroup().findProcessGroup(groupId)` pattern in `StandardControllerServiceProvider#getControllerServiceIdentifiers` will indeed return null for connector-managed groups since they're not children of the main root, leading to the empty `allowableValues`.

Since this reproduces regardless of TROUBLESHOOTING state (i.e., it's a pre-existing issue with connector-managed flows, not something introduced by this PR), it would be cleanest to track it as a separate JIRA so this PR stays focused on the TROUBLESHOOTING feature. The fix shape you described (`flowManager.getGroup(groupId)` in place of the root-group walk) is consistent with the direction of the other `findProcessGroup` fixes already in this PR, so it should be a straightforward follow-up. Would you mind filing that?

markap14:
[claude-opus-4.6] @rfellows Fixed in 69fe7f2: replaced `group.findProcessGroup(groupId)` with `flowManager.getGroup(groupId)` in `StandardControllerServiceProvider#getControllerServiceIdentifiers`, consistent with the approach used elsewhere in this PR. Added a unit test that verifies the lookup succeeds for a group that is not under the main root.

markap14:
[claude-opus-4.6] **CI: `ubuntu-24.04 Java 25` system-tests failed — pre-existing flake, not related to this PR.**

**Failing check:** [ubuntu-24.04 Java 25](https://github.com/apache/nifi/actions/runs/27164357082/job/80187641690) (system-tests)

**Failing test:** `ClusteredConnectorIT.testBundleResolutionResolveNewestBundle`

**Error:**
```
org.apache.nifi.toolkit.client.NiFiClientException: Error deleting purge request for Connector
a8ed95db-019e-1000-0000-000002955768: Cluster is unable to service request to change flow:
Node localhost:5671 is currently connecting
```

**Why this is unrelated:** The failure is a cluster timing race during test cleanup — a node hadn't finished connecting when the test tried to delete a purge request. This PR's latest commit changed `StandardControllerServiceProvider#getControllerServiceIdentifiers` (a controller service lookup path), which is not involved in connector purge or cluster join operations. `ClusteredConnectorIT` also [failed on main](https://github.com/apache/nifi/actions/runs/26846547987) on June 2 with a similar cluster timing issue (`testDeleteConnectorOnConnectWithDataQueued` timed out).

**Plan:** Will rerun the failed shard once the parent workflow completes (sibling jobs are still in progress).

markap14:
[claude-opus-4.6] **CI update: Ubuntu system-test shards failed again on rerun — all pre-existing flakes, unrelated to this PR.**

Both `ubuntu-24.04 Java 21` and `ubuntu-24.04 Java 25` failed with different tests this time, all cluster timing races:

| Test | Error | Shards |
|------|-------|--------|
| `ClusteredConnectorTroubleshootingIT.testConfigurationAndAuthoritativeFlowRestoredAfterTroubleshootingRestart` | Node failed to reconnect within 60s after restart | Java 25 |
| `LoadBalanceIT.testPartitionByAttribute` | HTTP 409 — node still connecting during queue listing | Both |
| `ClusteredRegistryClientIT.testControllerServiceUpdateWhileRunning` | Cluster conflict | Java 25 |
| `OffloadContentClaimTruncationIT.testOffloadedFlowFileContentNotPrematurelyTruncated` | Timeout | Java 21 |

All failing tests are different from the initial run (which failed on `ClusteredConnectorIT.testBundleResolutionResolveNewestBundle`), confirming these are non-deterministic cluster timing flakes rather than a consistent regression. `OffloadContentClaimTruncationIT` [also fails on main](https://github.com/apache/nifi/actions/runs/26792184340). macOS shards (`macos-15 Java 21`, `macos-15 Java 25`) both pass all system tests cleanly.

All other checks pass: build (3 OS/JDK combos), code-compliance (Validate + Scan), docker-tests, integration-tests.

rfellows (approved):

markap14:
[claude-opus-4.6] **CI update: both Ubuntu system-test shards failed again — all pre-existing cluster timing flakes, unrelated to this PR. Rerun triggered.**

**Run:** https://github.com/apache/nifi/actions/runs/27238660428

**`ubuntu-24.04 Java 25`** — 1 failure:
- `ClusteredRegistryClientIT.testControllerServiceUpdateWhileRunning`: HTTP 409 Conflict during test teardown (node still connecting when queue empty was requested)

**`ubuntu-24.04 Java 21`** — 6 failures, all cluster timing races:
- `ClusteredConnectorIT.testDeleteConnectorOnConnect`: timed out after 5 minutes
- `LoadBalanceIT.testPartitionByAttribute`: HTTP 500 listing queue (node still connecting)
- `ClusteredRegistryClientIT.testChangeVersionOnParentThatCascadesToChild`: HTTP 500 emptying queue during teardown
- `ClusteredRegistryClientIT.testNewComponentsStartedDuringVersionChange`: HTTP 500 emptying queue during teardown
- `ClusteredReplayProvenanceIT.testReplayLastEvent`: HTTP 500 emptying queue during teardown
- `OffloadIT.testOffloadCompletesAfterStoppingAndTerminatingProcessorThatRetainsSessions`: HTTP 500 emptying queue during teardown

Every failure is an HTTP 409/500 during cluster operations where a node hadn't finished connecting, or a timeout waiting for cluster coordination. The failing tests are spread across unrelated test classes and differ from the previous run's failures, confirming non-deterministic flakes. All other checks pass (macOS, Windows, build, Validate, CodeQL).

Rerunned failed shards via `gh run rerun 27238660428 --failed`.

markap14:
[claude-opus-4.6] **CI update: Ubuntu system-test shards flaked again on rerun — different tests each time, confirming non-deterministic cluster timing issues. Triggered another rerun.**

**Rerun results (run 27238660428):**

**`ubuntu-24.04 Java 21`** — 3 errors in 454 tests:
- `LoadBalanceIT.testPartitionByAttribute`: HTTP 500 listing queue
- `LoadBalanceIT` teardown: HTTP 500 emptying queue
- `ClusteredGetLatestProvenanceEventsIT` teardown: HTTP 500 emptying queue

**`ubuntu-24.04 Java 25`** — 4 errors in 454 tests:
- `OffloadIT.testOffload`: HTTP 500
- `ClusteredConnectorDrainIT` teardown: purge connector request failed
- `LoadBalanceIT.testPartitionByAttribute`: HTTP 500
- `OffloadContentClaimTruncationIT.testOffloadedFlowFileContentNotPrematurelyTruncated`: HTTP 500

All failing tests are different from the previous run's failures. All connector-specific tests passed, including `ConnectorTroubleshootingIT` and `ClusteredConnectorTroubleshootingIT`.

All other checks pass: Scan, Validate, CodeQL, macOS (Java 21 + 25), Windows, build (3 OS/JDK combos).

Triggered another rerun of the failed shards.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardConnectorDAO.java
```diff
@@ -157,6 +158,34 @@ public void verifyCancelDrainFlowFile(final String id) {
...
+    @Override
+    public void endTroubleshooting(final String id) {
+        final ConnectorNode connector = getConnector(id);
+        try {
+            getConnectorRepository().endTroubleshooting(connector);
+        } catch (final FlowUpdateException e) {
+            throw new IllegalStateException("Failed to exit troubleshooting mode for Connector " + id + ": " + e, e);
+        }
+    }
+
```
kevdoran, line 194: should any of these use `requireConnector(id, ConnectorSyncMode.LOCAL_ONLY);` rather than `getConnector(id)`? Just making sure this is intentional.

markap14:
[claude-opus-4.6] **CI update: Ubuntu system-test flakes are a pre-existing main-branch issue — not caused by this PR.**

After three attempts, the Ubuntu system-test shards continue to fail with different cluster timing errors each time. The one consistent failure across all attempts, `LoadBalanceIT.testPartitionByAttribute`, also fails on **main** — the [latest main branch system-tests run](https://github.com/apache/nifi/actions/runs/27207646193) (June 9) shows `LoadBalanceIT.testPartitionByAttribute` as the only failure on both Ubuntu shards.

**Attempt 3 failures (run 27238660428):**

`ubuntu-24.04 Java 21` — 4 errors:
- `LoadBalanceIT.testPartitionByAttribute` (also fails on main)
- `OffloadIT.testOffload`
- `OffloadContentClaimTruncationIT.testOffloadedFlowFileContentNotPrematurelyTruncated`
- `ClusteredParameterContextIT.testAssetReplacement`

`ubuntu-24.04 Java 25` — 2 errors:
- `LoadBalanceIT.testPartitionByAttribute` (also fails on main)
- `ClusterFlowDefinitionExportImportStateIT.testClusterExportCapturesBothScopes`

All connector-specific tests pass on every attempt, including `ConnectorTroubleshootingIT` and `ClusteredConnectorTroubleshootingIT`. All non-Ubuntu checks pass (macOS Java 21 + 25, Windows, build, Validate, Scan, CodeQL). Triggered a fourth rerun attempt.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -3704,7 +3734,8 @@ public ConnectorEntity updateConnector(final Revision revision, final ConnectorD
 
     @Override
     public void verifyDeleteConnector(final String id) {
-        // For now, DAO will enforce state; expose hook for symmetry
+        final ConnectorNode connector = connectorDAO.getConnector(id);
+        connector.verifyCanDelete();
```
kevdoran: Not a blocker as this was pre-existing, and improving it could be a future extensibility API enhancement, but `verifyCanDelete()` should probably use the `ConnectorDAO->ConnectorRepository` path to mirror `verifyCanCreate()`, in case the `ConnectorRepository` wants to consult with the optional `ConnectorConfigurationProvider`

kevdoran (approved):
Nice work @markap14 - this will be a useful feature!

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/dao/impl/StandardConnectorDAO.java
markap14: [claude-opus-4.6] Good catch — you're right, these should use `LOCAL_ONLY` like the other mutation methods in this class. There's no reason to sync with the provider before entering or exiting troubleshooting. Fixed in 49cb599: all four troubleshooting methods now use `requireConnector(id, ConnectorSyncMode.LOCAL_ONLY)`. Also took the opportunity to fix the `+ e` → `+ e.getMessage()` in the `endTroubleshooting` error message per the earlier review feedback.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
markap14: [claude-opus-4.6] Agreed — `verifyDeleteConnector` was bypassing the DAO/Repository path and going directly to the node, which means a `ConnectorConfigurationProvider` wouldn't be consulted during the pre-check phase. Fixed in the same commit (49cb599): added `verifyDelete(ConnectorNode)` to `ConnectorRepository`, `verifyDelete(String)` to `ConnectorDAO`, and routed `StandardNiFiServiceFacade.verifyDeleteConnector` through the DAO. The repository implementation currently delegates to `connector.verifyCanDelete()`, but this provides the hook point for provider consultation in the future, consistent with the `verifyCreate` pattern.

PR #11176, opened 2026-04-23
NIFI-15874: FQ class name to ExtensionManager to get InstanceClassloader

bobpaulin:
# Summary

[NIFI-15874](https://issues.apache.org/jira/browse/NIFI-15874)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

Run Verify on processor with @RequiresInstanceClassLoading

Verify ClassLoader that loads the processor is the same as the one used in verify.

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


markap14 (approved):
Thanks @bobpaulin this looks correct to me. Could also use getCanonicalClassName() rather than getComponentType() but getComponentClass().getName() is also reasonable. +1 will merge pending successful GitHub Actions

bobpaulin:
Hey @markap14 reviewed a couple other places in the code and I believe you are correct.  I think the fact that Canonical can return null in certain cases ( Hidden, Local, or Anon Classes)may have led me down this path.  That said I don't think any of those cases would apply to our components and the getCanonicalClassName() is used throughout the framework so it would be better to be consistent.  Created https://issues.apache.org/jira/browse/NIFI-15878 to track this work.

PR #11184, opened 2026-04-28
NIFI-15885: Addressed issues that cause Offload to get stuck: when a …

markap14:
…processor extends AbstractSessionFactoryProcessor and stores a reference to a created ProcessSession but not its factory, garbage collection could cause framework to be unable to rollback sessions. When a processor is terminated, ensure we rollback retained sessions even if all threads are completed. Fixed related bug that caused background thread never to complete when waiting for processor thread count to reach 1 and terminate set it to 0.

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
@@ -1905,6 +1905,14 @@ public class StandardProcessorNode extends ProcessorNode implements Connectable 
                                     }
                                 }
                             }
+                        } else if (lifecycleState.isTerminated()) {
+                            // Termination was requested while the stop sequence was waiting for active threads to drain.
+                            // LifecycleState.terminate() reset the active thread count to zero, so the count==1
+                            // condition above will never be reached and rescheduling would loop forever. Complete the
+                            // stop action and exit. completeStopAction() is idempotent if procNode.terminate() already
+                            // invoked it.
+                            LOG.debug("Stop sequence for {} aborted because LifecycleState was terminated", this);
+                            completeStopAction();
                         } else {
                             // Not all of the active threads have finished. Try again in 100 milliseconds.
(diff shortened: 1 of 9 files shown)


markap14:
Introduced tests for each of the issues mentioned; ensured that all of them failed in the expected ways before fixing; ensured that all passed after the fixes.

bbende:
+1 Looks good to me, will merge if builds pass

bbende:
@markap14 after looking at this integration test that keeps failing, I clicked the button to have Copiliot explain the build failure and this seems like a legit issue...
```
The failing tests show a critical issue: IllegalArgumentException: Cannot migrate from a StandardProcessSession to a class org.apache.nifi.controller.repository.WeakHashMapProcessSessionFactory$FactoryRetainingProcessSession

This error occurs in the MergeRecord processor when attempting to migrate a FlowFile between different ProcessSession implementations. The root cause is that the code is trying to migrate a StandardProcessSession to a FactoryRetainingProcessSession, which are incompatible types.

Key Issues from the Logs:
Session Type Mismatch: The StandardProcessSession.migrate() method at line 1505 is rejecting the migration because it validates that the target session type is compatible.
Cascading Failures: The migration error prevents the MergeRecord processor from functioning, causing FlowFiles to be routed to failure and blocking the connector from draining.
Timeout Chain: The inability to process FlowFiles causes the connector to timeout during prepareForUpdate(), which is called during dynamic flow updates (like the schema registry test).
Root Cause:
The WeakHashMapProcessSessionFactory is creating FactoryRetainingProcessSession instances, but the MergeRecord processor (or related code) is attempting to migrate sessions created by StandardProcessSession to these incompatible session types.

Solution:
The fix is to validate session type compatibility before attempting migration in StandardProcessSession.migrate(). Modify the migration logic to handle or reject incompatible session types gracefully:
```

markap14:
Thanks @bbende that's actually a great catch by Copilot! Have addressed the issue with the latest commit

bbende:
+1 looks good

PR #11187, opened 2026-04-29
NIFI-15825: Adding support for controller services and queue operations in the connector canvas

mcgilman:
# Summary

Adds support for working with controller services and performing queue operations from the connector canvas in the NiFi UI. Promotes the empty-queue NgRx state slice to a location shared by the flow designer and the connector canvas so both surfaces drive the same drop-request flow.

# Changes

## Connector canvas — context menu

- **Canvas-background menu** gains `Controller Services`, `Empty All Queues`, `Drain`, and `Cancel Drain` (alongside the existing `Refresh` and `Leave Group`).
- **Component menu** gains `Controller Services` and `Empty All Queues` for process groups, and `List Queue` and `Empty Queue` for connections.
- Icons aligned with the flow designer (e.g. `fa fa-list` for Controller Services).
- `Drain` and `Cancel Drain` are gated on connector permissions and the runtime's `allowedActions` set, and disable while the connector is being saved.

## Controller services view (read-only)

- New `ConnectorControllerServicesComponent` rooted at `/connectors/{connectorId}/canvas/{processGroupId}/controller-services`, deep-linkable to a specific service.
- Reuses the shared `controller-service-table`. To support a read-only entry point without mutating entity permissions:
  - Added a `readOnly` input on `controller-service-table` that gates the row action menu down to `View Configuration`, `View State`, and `View Documentation`.
  - Added a `readonly?: boolean` flag on `EditControllerServiceDialogRequest`. When set, `EditControllerService` enters view-only mode regardless of the user's permissions or the service's run status.
- New NgRx slice (`connector-controller-services` — actions, reducer, effects, selectors) handles loading, selection, and dialog wiring.
- Navigation is dispatched as `navigateToControllerServices({ processGroupId })` and uses `BackNavigation` to return to the originating canvas.

## Queue operations — shared empty-queue state

- Promoted the previous `flow-designer/state/queue` slice to `state/empty-queue`. The flow designer and the connector canvas now dispatch the same actions through a single shared slice.
- A new `queueEmptied` action carries a `CanvasActionSource` so each surface filters on the source and only refreshes its own view.
- Bulk drains have their own success action (`submitEmptyQueuesRequestSuccess`); the shared success effect listens to both the singular and bulk forms and routes to either `deleteEmptyQueueRequest` or `startPollingEmptyQueueRequest` depending on whether the drop request is finished.
- Connector-canvas actions for `Empty Queue`, `Empty All Queues`, `List Queue`, `Drain`, and `Cancel Drain` are wired through this slice.

## Cleanup and hardening (second commit)

A follow-up commit addresses minor items grouped by theme:

- **Correctness**
  - `EmptyQueueState` clears the unused identifier on submit so subsequent polling/deletion can never run against a stale `connectionId` or `processGroupId`.
  - Replaced `processGroupId!` non-null assertions in `pollEmptyQueueRequest$` and `deleteEmptyQueueRequest$` with explicit guards that dispatch `emptyQueueApiError` if neither id is present.
  - Added `hasAttemptedLoad` to `ConnectorControllerServicesState` so the controller services skeleton is only shown on the very first load — retries after a failed initial load no longer re-flash the skeleton.
- **Design**
  - Tightened `navigateToControllerServices` to require `processGroupId` and dropped the unreachable route fallback in the effect.
- **Cleanup**
  - Pruned ten unused selectors from `connector-controller-services`.
  - Removed an unused `navigateBackToConnectorCanvas` action and effect.
- **Tests**
  - Added reducer and effect coverage for the new bulk-success action and for the cleared-identifier and null-id guard paths in `EmptyQueueEffects`.
  - Added a `readonly`-derivation spec for `EditControllerService` covering the explicit flag, missing write permission, and non-DISABLED run status.
  - Updated controller-services component and reducer tests for `hasAttemptedLoad`.

# Follow-ups

The connector canvas's `List Queue` action navigates to the existing `/queue/{connectionId}` page. That page's queue listing relies on the standard `nifi-api/flowfile-queues/{id}/...` endpoints; full functionality for connections inside a connector's runtime depends on those endpoints resolving in the connector context and is tracked separately.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/app.module.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/app.module.ts
@@ -57,6 +57,7 @@ import { BannerTextEffects } from './state/banner-text/banner-text.effects';
 import { MAT_TOOLTIP_DEFAULT_OPTIONS, MatTooltipDefaultOptions } from '@angular/material/tooltip';
 import { CLIPBOARD_OPTIONS, provideMarkdown } from 'ngx-markdown';
 import { CopyEffects } from './state/copy/copy.effects';
+import { EmptyQueueEffects } from './state/empty-queue/empty-queue.effects';
 
 const entry = localStorage.getItem('disable-animations');
 let disableAnimations = '';
@@ -109,7 +110,8 @@ export const customTooltipDefaults: MatTooltipDefaultOptions = {
             DocumentationEffects,
             ClusterSummaryEffects,
             PropertyVerificationEffects,
-            CopyEffects
+            CopyEffects,
+            EmptyQueueEffects
         ),
         StoreDevtoolsModule.instrument({
             maxAge: 25,
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector.service.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector.service.ts
@@ -23,7 +23,7 @@ import { ClusterConnectionService } from '../../../service/cluster-connection.se
 import { ConnectorsResponse, CreateConnectorRequest } from '../state';
 import { ConnectorEntity } from '@nifi/shared';
 import { SearchResultsEntity } from '../../../state/shared';
-import { DropRequestEntity } from '../../flow-designer/state/queue';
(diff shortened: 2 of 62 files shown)


rfellows:
Will review...

rfellows (approved):
Excellent, @mcgilman 👍

PR #11188, opened 2026-04-29
NIFI-7686: Support multiple key columns in database lookup services

lhapollo:
# Summary

[NIFI-7686](https://issues.apache.org/jira/browse/NIFI-7686)

Added composite key support to DatabaseRecordLookupService and SimpleDatabaseLookupService. When LOOKUP_KEY_COLUMN contains a comma-separated list of column names, the services build a composite WHERE clause and expect coordinates keyed by column name instead of the single 'key' coordinate.

Includes type coercion so String values from LookupRecord are converted to the correct SQL types based on ParameterMetaData.

Backward compatible: single-key behavior is unchanged.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/db/AbstractDatabaseLookupService.java
+++ b/nifi-extension-bundles/nifi-standard-services/nifi-lookup-services-bundle/nifi-lookup-services/src/main/java/org/apache/nifi/lookup/db/AbstractDatabaseLookupService.java
@@ -23,16 +23,15 @@ import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.migration.PropertyConfiguration;
 import org.apache.nifi.processor.util.StandardValidators;
 
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


dan-s1:
@lhapollo Please note the merge is blocked as the commits must have verified signatures. Please see [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) for creating commits with verified signatures

dan-s1:
@lhapollo It looks like there are Checkstyle and PMD violations in the two builds that are failing. Please correct those.

exceptionfactory (changes requested):
Thanks for proposing this new feature @lhapollo. On initial review, this change introduces notable behavior changes, and has duplicative logic for things like coercing values. Expanding the semantic behavior of the `Lookup Key Column` property is also potentially confusing, so at minimum, it would require renaming the property. For all of these reasons, I think the approach needs closer consideration. It may warrant a different service implementation, rather than changes to the existing implementations.

exceptionfactory:
Closing this pull request for now, glad to discuss further options on the referenced Jira issue to decide on the best path forward before returning to code review.

PR #11195, opened 2026-05-02
NIFI-15895 Set Apache HttpClient5 version to 5.6.1

exceptionfactory:
# Summary

[NIFI-15895](https://issues.apache.org/jira/browse/NIFI-15895) Sets the Apache HttpClient5 library version to [5.6.1](https://downloads.apache.org/httpcomponents/httpclient/RELEASE_NOTES-5.6.x.txt) mitigating CVE-2026-40542. The library is packaged as a dependency in `nifi-aws-services-api-nar` and `nifi-iceberg-shared-nar`. Applying the dependency management version in the root Maven configuration applies to all modules.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-aws-bundle/pom.xml
+++ b/nifi-extension-bundles/nifi-aws-bundle/pom.xml
@@ -53,6 +53,12 @@
                 <version>2.10.0-SNAPSHOT</version>
                 <scope>provided</scope>
             </dependency>
+            <!-- Override httpclient 5.6 from apache5-client -->
+            <dependency>
+                <groupId>org.apache.httpcomponents.client5</groupId>
+                <artifactId>httpclient5</artifactId>
+                <version>5.6.1</version>
+            </dependency>
             <dependency>
                 <groupId>software.amazon.awssdk</groupId>
                 <artifactId>sts</artifactId>
--- a/nifi-extension-bundles/nifi-iceberg-bundle/pom.xml
+++ b/nifi-extension-bundles/nifi-iceberg-bundle/pom.xml
@@ -69,6 +69,12 @@
                 <artifactId>aircompressor</artifactId>
                 <version>2.0.3</version>
             </dependency>
+            <!-- Override httpclient 5.6 from iceberg-core -->
+            <dependency>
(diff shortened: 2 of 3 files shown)


pvillard31 (dismissed):

pvillard31 (changes requested):
Not sure why specifically with this PR but it feels like we have a timing issue and I guess the transitive upgrade is just exposing a long standing race condition issue is those tests.

In `nifi-docker/dockermaven/integration-test.sh`, we do:

```sh
if docker exec "${container_name}" bash -c " echo Running < /dev/tcp/${ip}/${port}"; then
```

Jetty binds early but we don't have everything ready yet. So I think this logic should be improved with something along the lines of

```sh
echo "Waiting for NiFi REST API (expect status: 400)"
max_iterations=10
sleep_time=10
sleep ${sleep_time}
for i in $(seq 1 "${max_iterations}"); do
    echo "Waiting for NiFi REST API - iteration: ${i}"
    status="$(docker exec "${container_name}" bash -c \
        "curl -ksSo /dev/null -w %{http_code} -m 10 --retry 5 --retry-all-errors --retry-max-time 60 https://${ip}:${port}/nifi-api/authentication/configuration" || true)"
    if [ "${status}" = "400" ]; then
        echo "NiFi REST API responded with ${status}"
        break
    fi
    if [ "${i}" -eq "${max_iterations}" ]; then
        echo "NiFi REST API did not return 400 within expected time (last status: ${status})"
        exit 1
    fi
    sleep ${sleep_time}
done
```

exceptionfactory:
Thanks @pvillard31, I started noticing that as well on the timing issue. I will take a closer look and follow up.

exceptionfactory:
@pvillard31 On closer inspection, the override was also impacting the framework, which has a dependency on an earlier minor version of Apache HttpClient 5, resulting in startup issues. I changed the scope of the override to AWS and Iceberg, since the earlier minor version is not impacted.

pvillard31 (approved):
Thanks @exceptionfactory - will merge after checks complete

PR #11201, opened 2026-05-04
NIFI-15901 Disable controller services before updating

adwk67:
# Summary

NIFI-15901

This is similar to https://github.com/apache/nifi/pull/11111 (https://issues.apache.org/jira/browse/NIFI-15801) but applies to controller services.

If a controller service in a child process group shares an identifier with a CS in the root process group, and a node is disconnected, and the child-group CS is deleted on that node, then when the node reconnects a FlowSynchronizationException is thrown because the root CS is not disabled before being updated.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -745,17 +745,43 @@ private void synchronizeControllerServices(final ProcessGroup group, final Versi
             updateControllerService(addedService, proposedService, topLevelGroup);
         }
 
-        // Update all of the Controller Services to match the VersionedControllerService
+        // Update all Controller Services to match the VersionedControllerService.
+        // Services may be ENABLED here if the outer "affected components" pass did not
+        // disable them (e.g. COMPONENT_ADDED diffs are skipped by AffectedComponentSet).
+        // We must disable before calling updateControllerService, which calls setProperties
+        // which calls verifyModifiable and throws IllegalStateException on ENABLED services.
+        final long stopTimeout = System.currentTimeMillis() + syncOptions.getComponentStopTimeout().toMillis();
```
pvillard31: Should this deadline be recomputed per service inside the loop, like `processorStopDeadline` in `synchronizeProcessors`, so later iterations don't inherit an exhausted budget?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -745,17 +745,43 @@ private void synchronizeControllerServices(final ProcessGroup group, final Versi
...
+                    }
+                    updateControllerService(service, proposedService, topLevelGroup);
+                    createdAndModifiedExtensions.add(new CreatedOrModifiedExtension(service, getPropertyValues(service)));
+                    LOG.info("Updated {}", service);
+                } finally {
+                    // Re-enable services and restart components that were stopped for the update,
+                    // restoring the controller to its pre-update running state.
+                    if (proposedService.getScheduledState() != org.apache.nifi.flow.ScheduledState.DISABLED) {
+                        context.getControllerServiceProvider().enableControllerServicesAsync(servicesToRestart);
+                        context.getControllerServiceProvider().scheduleReferencingComponents(
```
pvillard31, line 783: Should we also call `notifyScheduledStateChange` for `servicesToRestart` (ENABLED) and `referencesToRestart` (RUNNING) here, to match the per-service `synchronize(ControllerServiceNode, …)` overload?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -745,17 +745,43 @@ private void synchronizeControllerServices(final ProcessGroup group, final Versi
...
+                        throw new FlowSynchronizationException("Interrupted while stopping Controller Service " + service, e);
+                    }
+                    updateControllerService(service, proposedService, topLevelGroup);
+                    createdAndModifiedExtensions.add(new CreatedOrModifiedExtension(service, getPropertyValues(service)));
+                    LOG.info("Updated {}", service);
+                } finally {
+                    // Re-enable services and restart components that were stopped for the update,
+                    // restoring the controller to its pre-update running state.
+                    if (proposedService.getScheduledState() != org.apache.nifi.flow.ScheduledState.DISABLED) {
+                        context.getControllerServiceProvider().enableControllerServicesAsync(servicesToRestart);
```
pvillard31: Does enabling via `controllerServiceProvider` here race with the subsequent `componentScheduler.enableControllerServicesAsync` loop, and should both paths use the same scheduler?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -745,17 +745,43 @@ private void synchronizeControllerServices(final ProcessGroup group, final Versi
             updateControllerService(addedService, proposedService, topLevelGroup);
         }
 
-        // Update all of the Controller Services to match the VersionedControllerService
+        // Update all Controller Services to match the VersionedControllerService.
+        // Services may be ENABLED here if the outer "affected components" pass did not
+        // disable them (e.g. COMPONENT_ADDED diffs are skipped by AffectedComponentSet).
```
pvillard31: Should this reference `FlowDifferenceFilters.isComponentUpdateRequired` instead of `AffectedComponentSet`, since that's the filter actually populating `updatedVersionedComponentIds`?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
adwk67: Done: d4666a4

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
adwk67: Done: d4666a4

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
adwk67: I've gone with using the componentScheduler for both as it has already been paused by the caller: d4666a4

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
adwk67: I've simplified the comment to keep the salient bit: d4666a4. If I am reading the code correctly then AffectedComponentSet and isComponentUpdateRequired are doing different things (the former marking what gets disabled, the latter marking things for update).

exceptionfactory (commented):
Tagging @markap14 for additional review. Although there are some similarities to the recent change for Processors, it looks like this needs to be considered in the larger context of calls to the Synchronizer.

markap14:
Thanks @exceptionfactory.
@adwk67 I have a couple of concerns here.

Firstly, the lifecycle management of components that need to be updated should be happening higher up the call stack, in the `public void synchronize(final ProcessGroup group, final VersionedExternalFlow versionedExternalFlow, final FlowSynchronizationOptions options)` method. The intent here is to ensure that anything that will be modified is stopped at the start so that everything is in a good state before we start changing any of the components.

Secondly, I'm struggling to fully understand what the issue is. The Jira states:
> When the proposed flow adds a controller service at a child group level with the same identifier as an existing ENABLED service at the root level...
and
> Steps to reproduce
> Deploy a 2-node NiFi cluster with a flow where the same CS identifier appears at root group level and child group level.

But we should never have multiple components with the same identifier. Can you explain the steps necessary to get into this scenario to start with? If we have multiple components with the same id, that sounds like the source of the problem, rather than how we handle the synchronization.

adwk67:
Thank you for your feedback @markap14. To give some context: we came across https://issues.apache.org/jira/browse/NIFI-15801 in one of our installations recently and I wanted to see if there was a possibility of the same thing happening with controller services, coming up with [this](https://github.com/stackabletech/nifi-operator/tree/main/tests/templates/kuttl/cluster-sync-cs-bug) test and a corresponding patch to "fix" it. You can see from the [flow](https://github.com/stackabletech/nifi-operator/blob/main/tests/templates/kuttl/cluster-sync-cs-bug/20-install-nifi.yaml.j2#L31-L161) that is used that there is a duplicate identifier in there (`"identifier": "cs-http-ctx-0000-0000-000000000001"`), so it may be that the test and my assumption are false, but that it _is_ possible to deploy a flow with duplicate identifiers.

exceptionfactory:
@adwk67 Thanks for the discussion on this issue. The referenced flow uses identifiers that do not align with standard UUID formatting, which appears to indicate manual creation outside of Apache NiFi itself. This non-standard format is not supported, and as @markap14 said, it is not possible for the framework to create Controller Services with duplicate identifiers.

With that background, I'm closing this pull request for now, but the discussion can continue on the Jira issue. It sounds like evaluating the source of the duplicate identifiers in the flow definition would be the next best step.

PR #11204, opened 2026-05-04
NIFI-14212 fixed issue where an invalid processor could get in a perp…

markobean:
…etual starting state disallowing changes

# Summary

See Jira issue for details on the symptoms and procedure to replicate. In short, an INVALID processor can be stuck in an STARTING state making it impossible to configure or even stop. This PR resolves the issue. It ensures a Processor is not in an INVALID state prior to attempts to start.

There was a logging issue noted while investigating as well. The DEBUG log message indicated a time interval of 5 sec between successive attempts to start the Processor. The actual value is 500ms.

[NIFI-14212](https://issues.apache.org/jira/browse/NIFI-14212)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
@@ -1500,11 +1500,14 @@ public class StandardProcessorNode extends ProcessorNode implements Connectable 
         LOG.debug("Starting {}", this);
 
         ScheduledState currentState;
+        ValidationStatus currentValidationStatus;
         boolean starting;
         synchronized (this) {
             currentState = this.scheduledState.get();
 
-            if (currentState == ScheduledState.STOPPED) {
+            // Ensure the processor is valid before attempting to start
+            currentValidationStatus = this.getValidationStatus();
+            if (currentState == ScheduledState.STOPPED && currentValidationStatus == ValidationStatus.VALID) {
                 starting = this.scheduledState.compareAndSet(ScheduledState.STOPPED, scheduledState);
                 if (starting) {
                     setDesiredState(desiredState);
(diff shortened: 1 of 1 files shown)


dan-s1:
@markobean Please see the requirement above for submitting signed commits.

markobean:
> @markobean Please see the requirement above for submitting signed commits.

Commit signed now.

dan-s1:
Hmm commit is still showing as `Unverified`.

markap14:
Hey @markobean I don't think this is the change we want to make. When you do a Bulk Start (e.g., at a Process Group level) it does intentionally start invalid processors. Without that, we had cases where you'd start a Process Group and think everything was running. But something was invalid so it didn't start. But then when you look a bit later, it's valid and just shows as STOPPED, leading users to think it just intermittently doesn't start components... and there are a lot of flows where processors are expected to become valid due to environment changes like a file becoming available.

The concern in the Jira indicates "Once in this state, the processor cannot be DISABLED or STOPPED." That, I do not believe, is accurate, at a framework level. This sounds like a UI bug to me, where the UI doesn't allow you to click Stop. But if you were to stop the whole Process Group, it does stop the processor appropriately. I think I'd push more for a fix to the UI so that it allows stopping an invalid component.

pvillard31:
Was the issue confirmed in a recent version of NiFi? I believe we can now properly stop such processors since NiFi 2.7.0 via NIFI-14999.

markobean:
> Was the issue confirmed in a recent version of NiFi? I believe we can now properly stop such processors since NiFi 2.7.0 via [NIFI-14999](https://issues.apache.org/jira/browse/NIFI-14999).

I was sure the bad behavior still existed. However, I re-tested today and things seemed to be ok. Given the scenario in the original issue, I was able to stop the invalid UpdateAttribute processor successfully - both individually and as part of a whole process group stop. NiFi 2.9.0. Bad day yesterday?

I will close this PR.

markobean:
Cannot reproduce original issue. Closing.

PR #11208, opened 2026-05-05
NIFI-12456 Enabled JsonTreeReader and JsonPathReader to parse JSON leniently.

dan-s1:
# Summary

[NIFI-12456](https://issues.apache.org/jira/browse/NIFI-12456)
The changes in this PR now allow for  `JsonTreeReader` and `JsonPathReader`  to parse incoming JSON leniently which includes allowing

-  Java comments (/**/ and //)
-  Yaml comments (start of line begins with #)
-  Leading plus sign in numbers (e.g. +123)
-  Leading zeros in numbers (e.g. 0001)
-  "missing" decimal numbers to end with a decimal point (e.g. 123.)
-  "missing value" in an array (i.e. sequence of two commas, without value in-between e.g. ["A",,"C"]),
-  Trailing comma in an array or member in an object
-  Use of single quotes for quoting strings (i.e. use of an apostrophe)
-  Use of unquoted field names
-  Use of unescaped control characters (ASCII characters with value less than 32)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/AbstractJsonRowRecordReader.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/AbstractJsonRowRecordReader.java
@@ -52,6 +52,8 @@ import java.util.Optional;
 import java.util.function.BiPredicate;
 
 public abstract class AbstractJsonRowRecordReader implements RecordReader {
+    public static final String OBSOLETE_ALLOW_COMMENTS = "Allow Comments";
+
     public static final String DEFAULT_MAX_STRING_LENGTH = "20 MB";
 
     public static final PropertyDescriptor MAX_STRING_LENGTH = new PropertyDescriptor.Builder()
@@ -69,6 +71,14 @@ public abstract class AbstractJsonRowRecordReader implements RecordReader {
             .allowableValues("true", "false")
             .defaultValue("false")
             .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+            .build();
+
+    public static final PropertyDescriptor JSON_PARSE_MODE = new PropertyDescriptor.Builder()
+            .name("JSON Parse Mode")
+            .description("Specifies whether to parse the incoming JSON per the JSON specification or allow for lenient parsing")
+            .required(true)
+            .allowableValues(JsonParseMode.class)
+            .defaultValue(JsonParseMode.STANDARD)
             .build();
 
     private final ComponentLog logger;
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/AbstractJsonRowRecordReader.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/AbstractJsonRowRecordReader.java
@@ -64,21 +64,12 @@ public abstract class AbstractJsonRowRecordReader implements RecordReader {
             .addValidator(StandardValidators.DATA_SIZE_VALIDATOR)
             .build();
 
-    public static final PropertyDescriptor ALLOW_COMMENTS = new PropertyDescriptor.Builder()
-            .name("Allow Comments")
-            .description("Whether to allow comments when parsing the JSON document")
+    public static final PropertyDescriptor PARSING_STRATEGY = new PropertyDescriptor.Builder()
+            .name("Parsing Strategy")
+            .description("Set the strategy for the level of JSON specification conformity required")
             .required(true)
-            .allowableValues("true", "false")
-            .defaultValue("false")
-            .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
-            .build();
-
-    public static final PropertyDescriptor JSON_PARSE_MODE = new PropertyDescriptor.Builder()
-            .name("JSON Parse Mode")
-            .description("Specifies whether to parse the incoming JSON per the JSON specification or allow for lenient parsing")
-            .required(true)
-            .allowableValues(JsonParseMode.class)
-            .defaultValue(JsonParseMode.STANDARD)
+            .allowableValues(ParsingStrategy.class)
+            .defaultValue(ParsingStrategy.STANDARD)
             .build();
 
     private final ComponentLog logger;
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParseMode.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParseMode.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.json;
+
+import org.apache.nifi.components.DescribedValue;
+
+enum JsonParseMode implements DescribedValue {
+    STANDARD("Standard", "Parse JSON per the JSON specification"),
+    LENIENT("Lenient", """
+            Parse JSON leniently allowing Java comments (/**/ and //), Yaml comments (start of line begins with #)
+            , leading plus sign in numbers (e.g. +123), leading zeros in numbers (e.g. 0001),
+            "missing" decimal numbers to end with a decimal point (e.g. 123.),
+            "missing value" in an array (i.e. sequence of two commas, without value in-between e.g. ["A",,"C"]),
+            trailing comma in an array or member in an object, use of single quotes for quoting strings (i.e. use of an apostrophe)
+            and use of unquoted field names.""");
+
+    private final String displayName;
+    private final String description;
+
+    JsonParseMode(String displayName, String description) {
+        this.displayName = displayName;
+        this.description = description;
+    }
+
+    @Override
+    public String getValue() {
+        return name();
+    }
+
+    @Override
+    public String getDisplayName() {
+        return displayName;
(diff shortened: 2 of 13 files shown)


nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/AbstractJsonRowRecordReader.java
```diff
@@ -71,6 +73,14 @@ public abstract class AbstractJsonRowRecordReader implements RecordReader {
             .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
             .build();
 
+    public static final PropertyDescriptor JSON_PARSE_MODE = new PropertyDescriptor.Builder()
+            .name("JSON Parse Mode")
+            .description("Specifies whether to parse the incoming JSON per the JSON specification or allow for lenient parsing")
```
exceptionfactory: I recommend naming this `Parsing Strategy` since `JSON` is implied.

```suggestion
    public static final PropertyDescriptor PARSING_STRATEGY = new PropertyDescriptor.Builder()
            .name("Parsing Strategy")
            .description("Set the strategy for handling input content according to JSON specifications")
```

nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParseMode.java
```diff
@@ -0,0 +1,53 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.json;
+
+import org.apache.nifi.components.DescribedValue;
+
+enum JsonParseMode implements DescribedValue {
```
exceptionfactory:
```suggestion
enum ParsingStrategy implements DescribedValue {
```

nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/JsonParserFactory.java
```diff
@@ -42,14 +42,22 @@ public JsonParserFactory() {
      * JSON Parser Factory constructor with configurable constraints
      *
      * @param streamReadConstraints Stream Read Constraints
-     * @param allowComments Allow Comments during parsing
+     * @param lenient Allow for parsing JSON leniently
      */
-    public JsonParserFactory(final StreamReadConstraints streamReadConstraints, final boolean allowComments) {
+    public JsonParserFactory(final StreamReadConstraints streamReadConstraints, final boolean lenient) {
```
exceptionfactory: It looks like this should be changed to pass in the `ParsingStrategy` instead of the `boolean`

exceptionfactory (changes requested):
Thanks for working on this configuration option @dan-s1. The general approach looks clear, I provided some recommendations on naming.

nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-json-record-utils/src/main/java/org/apache/nifi/json/AbstractJsonRowRecordReader.java
dan-s1: Okay. The wording I am not sure is accurate as if lenient is chosen, it will not be according to the JSON specifications. Jackson specifically says in their [Javadoc](https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.21.2/com/fasterxml/jackson/core/json/JsonReadFeature.html), that the leniencies are not per the JSON specification.

exceptionfactory: That's a good point, I meant to convey the level of conformity to the specification, so maybe something like "Set the strategy for the level of JSON specification conformity required"

dan-s1:
@exceptionfactory I have made all the changes you requested. I am concerned though that the configuration for the `YamlTreeReader`  is now awkward since it extends `JsonTreeReader`. The property `Parsing Strategy` is not relevant  to `YamlTreeReader` since Yaml parsing  does not support these features. Even the original `Allow Comment` does not make sense since SnakeYaml (the implementation we use for Yaml parsing) supports parsing comments when one plans to preserve them, but by default it ignores comments as seen in the [javadocs](https://www.javadoc.io/doc/org.yaml/snakeyaml/2.0/org/yaml/snakeyaml/LoaderOptions.html#setProcessComments(boolean)).  Hence since we are not preserving them we shouldn't need any option to keep them. 
The whole reason we had `YamlTreeReader`  extend `JsonTreeReader` was to not have to duplicate code as the JSON code could drive the Yaml parsing. But  the property `Parsing Strategy` should not be included.

Is the only way possible to remove `Parsing Strategy` in  `getSupportedPropertyDescriptors()` with code like below?
```
@Override
        
        final List<PropertyDescriptor> supportedPropertyDescriptors = new ArrayList<>(super.getSupportedPropertyDescriptors());
        supportedPropertyDescriptors.remove(AbstractJsonRowRecordReader.PARSING_STRATEGY);
        
        return  supportedPropertyDescriptors;
    }
```

I am assuming I cannot simply remove them in `migrateProperties`.

So I added that code. Please let me know if that is okay.

dan-s1:
@exceptionfactory I believe I addressed all the issues you had brought up. Is there anything else you would like me to change for this PR?

nifi-extension-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/json/JsonPathReader.java
```diff
@@ -152,6 +154,23 @@ protected Collection<ValidationResult> customValidate(final ValidationContext va
             .build());
     }
 
+    @Override
+    public void migrateProperties(PropertyConfiguration config) {
+        super.migrateProperties(config);
+
+        if (config.isPropertySet(AbstractJsonRowRecordReader.OBSOLETE_ALLOW_COMMENTS)) {
+            final String allowCommentsRawValue = config.getRawPropertyValue(AbstractJsonRowRecordReader.OBSOLETE_ALLOW_COMMENTS).orElse("");
```
exceptionfactory: It looks like this should fall back to `"false"` to reflect the default behavior.
```suggestion
            final String allowCommentsRawValue = config.getRawPropertyValue(AbstractJsonRowRecordReader.OBSOLETE_ALLOW_COMMENTS).orElse(Boolean.FALSE.toString());
```

nifi-extension-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/json/JsonTreeReader.java
```diff
@@ -142,10 +142,24 @@ public void migrateProperties(PropertyConfiguration config) {
         config.renameProperty("starting-field-name", STARTING_FIELD_NAME.getName());
         config.renameProperty("schema-application-strategy", SCHEMA_APPLICATION_STRATEGY.getName());
         config.renameProperty(OBSOLETE_SCHEMA_CACHE, SCHEMA_CACHE.getName());
+
+        if (config.isPropertySet(AbstractJsonRowRecordReader.OBSOLETE_ALLOW_COMMENTS)) {
+            final String allowCommentsRawValue = config.getRawPropertyValue(AbstractJsonRowRecordReader.OBSOLETE_ALLOW_COMMENTS).orElse("");
```
exceptionfactory:
```suggestion
            final String allowCommentsRawValue = config.getRawPropertyValue(AbstractJsonRowRecordReader.OBSOLETE_ALLOW_COMMENTS).orElse(Boolean.FALSE.toString());
```

nifi-extension-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/yaml/YamlTreeReader.java
```diff
@@ -44,26 +46,28 @@
         + "See the Usage of the Controller Service for more information and examples.")
 public class YamlTreeReader extends JsonTreeReader {
 
-    private static final boolean ALLOW_COMMENTS_DISABLED = false;
+    @Override
+    public void migrateProperties(PropertyConfiguration config) {
+        super.migrateProperties(config);
+        config.removeProperty(AbstractJsonRowRecordReader.PARSING_STRATEGY.getName());
```
exceptionfactory, line 53: Is this necessary? It is not part of the existing version, so it should not need to be removed in the migrate method.

exceptionfactory (changes requested):
Thanks for making the updates and also adjusting the YAML Reader. The general approach looks good. I noted a few remaining minor things, and then this should be ready to go.

nifi-extension-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/yaml/YamlTreeReader.java
dan-s1: Yes since the `super.migrateProperties(config);` creates it on lines 146-153 if `AbstractJsonRowRecordReader.OBSOLETE_ALLOW_COMMENTS` is set which YamlTreeReader could have set set also.

exceptionfactory: Thanks for clarifying, I see how this could get added now. Can you add a comment indicating something like, `// Remove Parsing Strategy from potential addition through the parent JsonTreeReader property migration`?

dan-s1: Sure!

dan-s1:
@exceptionfactory On the ticket in Jira, there was a request to also add `ALLOW_UNESCAPED_CONTROL_CHARS`. Is that something you think we should also add?

exceptionfactory:
> @exceptionfactory On the ticket in Jira, there was a request to also add `ALLOW_UNESCAPED_CONTROL_CHARS`. Is that something you think we should also add?

Yes, that seems worth adding as part of this initial implementation.

exceptionfactory (approved):
Thanks for working through the feedback @dan-s1, the latest version looks good. +1 merging

PR #11210, opened 2026-05-05
NIFI-15910 Improve Connector Secret verification

kevdoran:
# Summary

[NIFI-15910](https://issues.apache.org/jira/browse/NIFI-15910)

When a connector configuration contains a `SECRET_REFERENCE` whose fields (`fullyQualifiedSecretName`, `secretName`) are unpopulated — either because the property is gated by another property whose dependency isn't satisfied, or because it's a placeholder default emitted by an external `ConnectorConfigurationProvider` — `StandardConnectorNode` is eagerly resolving it against the SecretsManager and surfacing Property Validation - Secret Reference: The referenced secret [null] could not be found.

This change makes `StandardConnectorNode.resolvePropertyReferences` (and the matching validatePropertyReferences step traversal) skip secret resolution for two cases: 

1. Properties whose ConnectorPropertyDependency isn't satisfied — using the same algorithm as `AbstractConnector.isDependencySatisfied()`
2. Structurally-empty SECRET_REFERENCE values where neither FQN nor secret name is populated. Empty stubs are treated as unset, so the connector's own required-property check produces the actionable <name> is required message. 

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
@@ -78,6 +78,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 public class StandardConnectorNode implements ConnectorNode {
@@ -910,19 +911,8 @@ public class StandardConnectorNode implements ConnectorNode {
     @Override
     public List<ConfigVerificationResult> verifyConfigurationStep(final String stepName, final StepConfiguration configurationOverrides) {
         logger.debug("Verifying configuration step {} for {}", stepName, this);
-        final List<SecretReference> invalidSecretRefs = new ArrayList<>();
-        final List<AssetReference> invalidAssetRefs = new ArrayList<>();
-        final Map<String, String> resolvedPropertyOverrides = resolvePropertyReferences(configurationOverrides, invalidSecretRefs, invalidAssetRefs);
-
         final List<ConfigVerificationResult> results = new ArrayList<>();
         try (final NarCloseable ignored = NarCloseable.withComponentNarLoader(extensionManager, getConnector().getClass(), getIdentifier())) {
-
-            final DescribedValueProvider allowableValueProvider = (step, propertyName) -> fetchAllowableValues(step, propertyName, workingFlowContext);
-
-            final MutableConnectorConfigurationContext configContext = workingFlowContext.getConfigurationContext().createWithOverrides(stepName, resolvedPropertyOverrides);
-            final ConnectorConfiguration connectorConfig = configContext.toConnectorConfiguration();
-            final ParameterContextFacade paramContext = workingFlowContext.getParameterContext();
-            final ConnectorValidationContext validationContext = new StandardConnectorValidationContext(connectorConfig, allowableValueProvider, paramContext);
 
             final Optional<ConfigurationStep> optionalStep = getConfigurationStep(stepName);
             if (optionalStep.isEmpty()) {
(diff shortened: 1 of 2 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/components/connector/TestStandardConnectorNode.java
```diff
@@ -527,6 +531,83 @@ public void testVerifyConfigurationStepSurfacesInvalidValidationResults() throws
...
+    public void testVerifyConfigurationStepSkipsSecretReferenceWhenPropertyDependenciesNotMet() throws FlowUpdateException {
+        final DependentSecretVerifyConnector connector = new DependentSecretVerifyConnector();
+        final StandardConnectorNode connectorNode = createConnectorNode(connector);
+
+        connectorNode.transitionStateForUpdating();
+        connectorNode.prepareForUpdate();
+
+        final Map<String, ConnectorValueReference> propertyValues = new HashMap<>();
+        propertyValues.put("Mode", new StringLiteralValue("OFF"));
+        propertyValues.put("SecretKey", new SecretReference("pid", "My Provider", null, null));
```
mcgilman: This test doesn't actually isolate the dependency-skip path it advertises. `SecretReference` here is constructed with `secretName == null && fullyQualifiedName == null`, so it's also structurally empty — meaning the new `isEmptySecretReference` filter alone is enough to make this test pass, even if the dependency-aware filter regressed or was deleted.

Could you tighten this so the dependency filter is the only thing carrying the test? Something like:

```java
propertyValues.put("SecretKey", new SecretReference("pid", "My Provider", "secretA", "fqn-A"));
when(secretsManager.getSecrets(anySet()))
    .thenThrow(new AssertionError("Should not look up secrets when property dependency is unsatisfied"));
```

That makes the test fail loudly if the dependency-skip path ever stops short-circuiting the `SecretsManager` lookup.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
```diff
@@ -1033,6 +1057,92 @@ private Map<String, String> resolvePropertyReferences(final StepConfiguration co
...
+        final ConnectorValueReference ref = stepConfig.getPropertyValue(descriptor.getName());
+        if (ref == null) {
+            return descriptor.getDefaultValue();
+        }
+        if (ref instanceof StringLiteralValue stringLiteralValue) {
+            final String value = stringLiteralValue.getValue();
+            return value != null ? value : descriptor.getDefaultValue();
+        }
+        return null;
+    }
```
mcgilman, line 1101: The Javadoc says this is "aligned with `AbstractConnector.isDependencySatisfied(...)`", but the algorithms diverge in two cases:

1. A typed `StringLiteralValue` whose `getValue()` is `null` — this helper falls back to `descriptor.getDefaultValue()`, while `AbstractConnector` returns `null` (treating the dependency as unsatisfied).
2. An `AssetReference` / `SecretReference` controlling property — this helper returns `null` (unsatisfied), while `AbstractConnector` returns the *resolved* value via `propertyValue.getValue()`.

In practice this is unlikely to surface because dependency-controlling properties are almost always literal allowable-values, but the doc as written is misleading. Could you tighten it to describe what this method actually does — e.g. note that it's an approximation for `StepConfiguration` payloads scoped to `STRING_LITERAL` controlling properties, that `null`-valued literals fall back to the descriptor default rather than failing the dependency, and that non-literal reference types are treated as unsatisfied? That way the alignment claim is accurate and the divergence from `AbstractConnector` is documented rather than implied-away.

mcgilman (changes requested):
Two changes requested. The fix and its UX rationale look good overall — these are about test isolation and a misleading Javadoc claim. Full review notes inline.

kevdoran:
Thanks @mcgilman. Pushed an update:

Here is a summary of what was changed:

1. Test isolation (testVerifyConfigurationStepSkipsSecretReferenceWhenPropertyDependenciesNotMet)
- Built the test's SecretReference with all four fields populated (pid, My Provider, my-secret, My Provider.my-secret) so it is not an empty stub. The new isEmptySecretReference filter cannot mask a regression in the dependency filter.
- Wired in a strict SecretsManager whose getSecrets(anySet()) is stubbed to thenThrow(new AssertionError(...)). If the dependency-skip path ever stops short-circuiting, the test now fails loudly with a clear message.
- To support per-test SecretsManagers, added an overload createConnectorNode(Connector, SecretsManager). The original single-arg overload now delegates to it, preserving existing tests.
2. Javadoc accuracy (getConfiguredValueForDependency)
- Replaced the misleading "aligned with AbstractConnector.isDependencySatisfied(...)" claim with an explicit description of: "What this helper is for (controlling-property String value lookup against a pre-resolution StepConfiguration)."
- The two cases where it intentionally diverges from the connector-side implementation (typed null StringLiteralValue → descriptor default; non-literal references → null/unsatisfied).
- Why those divergences are acceptable (controlling properties are, in practice, STRING_LITERAL values from a fixed allowable-value set; null literals match the "switch back to default mode" UX).

mcgilman (approved):
Thanks for the updates @kevdoran. The changes look good to me but let's let @markap14 get some eyes on before merging.

markap14 (approved):

markap14:
Thanks @kevdoran @mcgilman I'm good with the changes. +1 will merge

PR #11211, opened 2026-05-06
NIFI-14472: Fix potential NPE in PutKinesisFirehose batch accumulation

rakesh-rsky:
## Summary

Fixes a race condition / NPE in PutKinesisFirehose.onTrigger() where a two-step computeIfAbsent() + get() on the batch-accumulation map could return null if the entry was evicted between the two calls.

## Changes

- Collapsed the two-step lookup into a single atomic computeIfAbsent(...).add(record) call in PutKinesisFirehose.java, eliminating the window for a NullPointerException.

## Testing

- Existing unit tests pass.
- The fix is a straightforward atomic-operation replacement with no behavioral change under normal conditions.

Fixes: https://issues.apache.org/jira/browse/NIFI-14472

### DIFF
--- a/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/firehose/PutKinesisFirehose.java
+++ b/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/firehose/PutKinesisFirehose.java
@@ -134,8 +134,11 @@ public class PutKinesisFirehose extends AbstractAwsSyncProcessor<FirehoseClient,
             for (final FlowFile flowFile : flowFiles) {
                 final String firehoseStreamName = context.getProperty(KINESIS_FIREHOSE_DELIVERY_STREAM_NAME).evaluateAttributeExpressions(flowFile).getValue();
 
-                recordHash.computeIfAbsent(firehoseStreamName, k -> new ArrayList<>());
-                session.read(flowFile, in -> recordHash.get(firehoseStreamName).add(Record.builder().data(SdkBytes.fromInputStream(in)).build()));
+                session.read(flowFile, in ->
+                        recordHash
+                                .computeIfAbsent(firehoseStreamName, k -> new ArrayList<>())
+                                .add(Record.builder().data(SdkBytes.fromInputStream(in)).build())
+                );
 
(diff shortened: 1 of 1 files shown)


exceptionfactory:
@rakesh-rsky It looks like the last update to this branch brought in a commit instead of rebasing, can you rebase your branch against `main` and force push to align with the current baseline?

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/kinesis/firehose/PutKinesisFirehose.java
```diff
@@ -134,8 +134,13 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
 
-                recordHash.computeIfAbsent(firehoseStreamName, k -> new ArrayList<>());
-                session.read(flowFile, in -> recordHash.get(firehoseStreamName).add(Record.builder().data(SdkBytes.fromInputStream(in)).build()));
+                // Use a single computeIfAbsent().add() call so the list lookup is atomic.
+                // The previous two-step pattern (computeIfAbsent then a separate get()) could
+                // throw NullPointerException when firehoseStreamName evaluates to null via
+                // expression language, because get(null) returned null after the absent-key
+                // entry was never actually inserted. (NIFI-14472)
+                session.read(flowFile, in -> recordHash.computeIfAbsent(firehoseStreamName, k -> new ArrayList<>())
+                        .add(Record.builder().data(SdkBytes.fromInputStream(in)).build()));
```
exceptionfactory: Commenting on the previous behavior is unnecessary and should be removed. In addition, it would be helpful to separate the callback implementation across multiple lines for readability.

exceptionfactory (changes requested):
Thanks for addressing this problem @rakesh-rsky. The core functional change appears to be correct, but the comments are unnecessary and the implementation line should be broken out into several lines for readability. If you can make that change, remove the extra commit, and rebase against the current main branch, this should be good to go.

exceptionfactory (changes requested):
Thanks for the update @rakesh-rsky, the functional change looks good, but the branch still includes the unrelated change to `KubernetesConfigMapStateProviderTest`.

exceptionfactory (approved):
Thanks again for making the changes @rakesh-rsky, the latest version looks good. +1 merging

PR #11212, opened 2026-05-06
NIFI-15906: Fix IllegalStateException during cluster reconnect in StandardConnection

rakesh-rsky:
## Summary

During cluster node reconnection, StandardConnection.setDestination() throws an IllegalStateException ("Cannot change destination while destination is running") when the destination is a RemoteGroupPort. Remote ports are managed by the cluster coordinator and can legally be running during reconnect — they should be exempt from this guard, just like Funnel and NiFiPort.

## Changes

- Added RemoteGroupPort to the instanceof-check exemption list in StandardConnection.setDestination().
- RemoteGroupPort was already imported in the file; no new imports needed.

## Testing

- Reproducer: disconnect a node from a cluster that has a Remote Process Group destination, then reconnect — the ISE no longer fires.
- Existing unit tests pass.

Fixes: https://issues.apache.org/jira/browse/NIFI-15906

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/connectable/StandardConnection.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/connectable/StandardConnection.java
@@ -292,7 +292,18 @@ public final class StandardConnection implements Connection {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


exceptionfactory (approved):
Thanks for providing this fix for S2S Remote Group Ports @rakesh-rsky. The change and comment look good. +1 merging

PR #11213, opened 2026-05-06
NIFI-15483: Route PublishAMQP to failure on undeliverable messages instead of silent success

rakesh-rsky:
## Summary

`PublishAMQP` silently routes FlowFiles to `REL_SUCCESS` even when the AMQP broker cannot deliver the message, causing silent data loss.

Two failure modes are addressed:

1. **Undeliverable message (`basic.return`)** — broker returns the message when no queue is bound to the exchange/routing-key. The fix uses AMQP Publisher Confirms + `basic.return` to detect and surface delivery failures, routing the FlowFile to `REL_FAILURE`.

2. **Exchange not found (`ShutdownSignalException`)** — when the exchange does not exist, the broker closes the channel with `404 NOT_FOUND`, causing `waitForConfirms()` to throw `ShutdownSignalException`. This is now caught and converted to `AMQPException` so the FlowFile routes to `REL_FAILURE` instead of causing an unhandled processor failure.

## Testing

- Publish to an exchange with no bound queues → FlowFile routes to `failure` with the broker's return reason
- Publish to a non-existent exchange → FlowFile routes to `failure` instead of unhandled processor error
- Normal publish (queue bound) → still routes to `success`
- Added regression tests for all failure scenarios — verified to fail against unfixed code

Fixes: https://issues.apache.org/jira/browse/NIFI-15483

### DIFF
--- a/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AMQPPublisher.java
+++ b/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AMQPPublisher.java
@@ -20,10 +20,13 @@ import com.rabbitmq.client.AMQP.BasicProperties;
 import com.rabbitmq.client.AlreadyClosedException;
 import com.rabbitmq.client.Connection;
 import com.rabbitmq.client.ReturnListener;
+import com.rabbitmq.client.ShutdownSignalException;
 import org.apache.nifi.logging.ComponentLog;
 
 import java.io.IOException;
 import java.net.SocketException;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Generic publisher of messages to AMQP-based messaging system. It is based on
@@ -34,6 +37,14 @@ final class AMQPPublisher extends AMQPWorker {
     private final String connectionString;
 
     /**
+     * Stores the broker's return reason when a message is published with mandatory=true
+     * but the broker cannot route it to any queue. Written by the AMQP I/O thread via
+     * {@link UndeliverableMessageLogger} and read by the publishing thread after
+     * {@link com.rabbitmq.client.Channel#waitForConfirms} synchronizes the two.
+     */
+    private final AtomicReference<String> undeliverableReturnReason = new AtomicReference<>(null);
+
+    /**
      * Creates an instance of this publisher
      *
      * @param connection instance of AMQP {@link Connection}
@@ -42,6 +53,17 @@ final class AMQPPublisher extends AMQPWorker {
         super(connection, processorLog);
         getChannel().addReturnListener(new UndeliverableMessageLogger());
         this.connectionString = connection.toString();
+
+        // Enable Publisher Confirms on this channel so that waitForConfirms() can be used
+        // after basicPublish() to create a synchronization point. This ensures that any
+        // basic.return frame sent by the broker (for mandatory messages it cannot route)
+        // will have been processed by the ReturnListener before waitForConfirms() returns,
+        // allowing undeliverable messages to be reliably detected and routed to REL_FAILURE.
+        try {
+            getChannel().confirmSelect();
+        } catch (final IOException e) {
+            throw new AMQPException("Failed to enable Publisher Confirms on AMQP channel", e);
+        }
 
         processorLog.info("Successfully connected AMQPPublisher to {}", this.connectionString);
     }
@@ -68,12 +90,49 @@ final class AMQPPublisher extends AMQPWorker {
             processorLog.debug("Successfully connected AMQPPublisher to {} and '{}' exchange with '{}' as a routing key.", this.connectionString, exchange, routingKey);
         }
 
+        // Reset any stale return reason from a previous publish before sending.
+        undeliverableReturnReason.set(null);
+
         try {
             getChannel().basicPublish(exchange, routingKey, true, properties, bytes);
         } catch (AlreadyClosedException | SocketException e) {
             throw new AMQPRollbackException("Failed to publish message because the AMQP connection is lost or has been closed", e);
         } catch (Exception e) {
             throw new AMQPException("Failed to publish message to Exchange '" + exchange + "' with Routing Key '" + routingKey + "'.", e);
+        }
+
+        // Wait for the broker's publish confirm (ack/nack). Because the broker sends a basic.return
+        // frame BEFORE the corresponding confirm frame for mandatory messages it cannot route,
+        // UndeliverableMessageLogger.handleReturn() is guaranteed to have run by the time
+        // waitForConfirms() returns. This makes undeliverable-message detection reliable.
+        //
+        // If the exchange does not exist, the broker closes the channel with a 404 NOT_FOUND
+        // channel.close frame, which causes waitForConfirms() to throw ShutdownSignalException.
+        // We catch it here and convert it to AMQPException so the FlowFile routes to REL_FAILURE
+        // instead of surfacing as an unhandled processor error.
+        try {
+            if (!getChannel().waitForConfirms(5_000L)) {
+                throw new AMQPException("Broker negatively acknowledged (NACK) message published to Exchange '"
+                        + exchange + "' with Routing Key '" + routingKey + "'");
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new AMQPException("Interrupted while waiting for publish confirmation from broker", e);
+        } catch (TimeoutException e) {
+            throw new AMQPException("Timed out waiting for publish confirmation from broker for Exchange '"
+                    + exchange + "' with Routing Key '" + routingKey + "'", e);
+        } catch (ShutdownSignalException e) {
+            // Broker closed the channel — most commonly because the exchange does not exist (404)
+            // or the vhost was deleted. Convert to AMQPException so PublishAMQP routes to REL_FAILURE.
+            throw new AMQPException("Broker closed channel while waiting for publish confirmation — "
+                    + "Exchange '" + exchange + "' may not exist: " + e.getMessage(), e);
+        }
+
+        // If the broker returned the message (e.g., no queue bound to the exchange/routing-key),
+        // surface it as a hard failure so the caller can route to REL_FAILURE instead of REL_SUCCESS.
+        final String returnReason = undeliverableReturnReason.get();
+        if (returnReason != null) {
+            throw new AMQPException("Message returned as undeliverable by broker — " + returnReason);
         }
     }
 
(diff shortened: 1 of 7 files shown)


turcsanyip:
@rakesh-rsky Thanks for working on this issue. Please note the unit tests are failing.

I tried the new error handling in my local environment but I'm getting the following runtime error (instead of routing the FlowFile to failure):
```
2026-05-06 14:34:55,697 ERROR [Timer-Driven Process Thread-4] o.a.nifi.amqp.processors.PublishAMQP PublishAMQP[id=4fd4ee4e-8e3a-3969-7db4-953498193e4d] Processor failure
com.rabbitmq.client.ShutdownSignalException: channel error; protocol method: #method<channel.close>(reply-code=404, reply-text=NOT_FOUND - no exchange 'dummy' in vhost '/', class-id=60, method-id=40)
	at com.rabbitmq.client.impl.ChannelN.waitForConfirms(ChannelN.java:219)
	at org.apache.nifi.amqp.processors.AMQPPublisher.publish(AMQPPublisher.java:108)
	at org.apache.nifi.amqp.processors.PublishAMQP.processResource(PublishAMQP.java:185)
	at org.apache.nifi.amqp.processors.PublishAMQP.processResource(PublishAMQP.java:52)
	at org.apache.nifi.amqp.processors.AbstractAMQPProcessor.onTrigger(AbstractAMQPProcessor.java:236)
	at org.apache.nifi.processor.AbstractProcessor.onTrigger(AbstractProcessor.java:27)
	at org.apache.nifi.controller.StandardProcessorNode.onTrigger(StandardProcessorNode.java:1292)
	at org.apache.nifi.controller.tasks.ConnectableTask.invoke(ConnectableTask.java:229)
	at org.apache.nifi.controller.scheduling.TimerDrivenSchedulingAgent$1.run(TimerDrivenSchedulingAgent.java:102)
	at org.apache.nifi.engine.FlowEngine.lambda$wrap$1(FlowEngine.java:105)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:358)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: com.rabbitmq.client.ShutdownSignalException: channel error; protocol method: #method<channel.close>(reply-code=404, reply-text=NOT_FOUND - no exchange 'dummy' in vhost '/', class-id=60, method-id=40)
	at com.rabbitmq.client.impl.ChannelN.asyncShutdown(ChannelN.java:529)
	at com.rabbitmq.client.impl.ChannelN.processAsync(ChannelN.java:350)
	at com.rabbitmq.client.impl.AMQChannel.handleCompleteInboundCommand(AMQChannel.java:193)
	at com.rabbitmq.client.impl.AMQChannel.handleFrame(AMQChannel.java:125)
	at com.rabbitmq.client.impl.AMQConnection.readFrame(AMQConnection.java:768)
	at com.rabbitmq.client.impl.AMQConnection.access$400(AMQConnection.java:49)
	at com.rabbitmq.client.impl.AMQConnection$MainLoop.run(AMQConnection.java:695)
	... 1 common frames omitted
```

rakesh-rsky:
@turcsanyip Thank you for validating this.

**Root cause:** When the exchange does not exist, the broker closes the channel with ``404 NOT_FOUND``, causing ``waitForConfirms()`` to throw  ``ShutdownSignalException`` — which was not caught, resulting in an unhandled processor failure instead of routing to ``REL_FAILURE``.

This has been fixed in the latest commits — ``ShutdownSignalException`` is now caught and converted to ``AMQPException``, along with regression tests covering this and related broker failure scenarios.

turcsanyip:
@rakesh-rsky I checked the latest changes and the error handling is correct now functionally.

However, the synchronous wait led to significant performance degradation compared to the original version. According to my measurement:
- local NiFi + local RabbitMQ in Docker: 10x slower
- local NiFi + RabbitMQ on a cloud VM: 300-1000x slower (depending on the cloud location)

So even if the error handling is improved, I don't think we can add the change as-is due to the performance effect. I suggest adding a feature flag property (e.g. `Use Message Confirmation` = true / false or `Delivery Guarantee` = At least once / At most once) with proper documentation on the pros and cons. This way, the user can optionally enable this feature if they want the extra failure handling, at the cost of performance.

rakesh-rsky:
@turcsanyip Thank you for the performance feedback and the suggestion.

I've added a **Delivery Guarantee** property to `PublishAMQP` with two options:

- **At most once** *(default)* - original fire-and-forget behaviour, no performance impact. Undeliverable messages are logged as a warning and the FlowFile still routes to `success`.
- **At least once** - enables Publisher Confirms. The processor waits for a broker ack/nack before routing. Undeliverable messages and NACKs route to `failure`, preventing silent data loss.

This keeps the default behaviour unchanged while giving users the option to trade throughput for delivery reliability.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AMQPPublisher.java
```diff
@@ -83,23 +139,24 @@ public String toString() {
...
      */
     private final class UndeliverableMessageLogger implements ReturnListener {
         @Override
         public void handleReturn(int replyCode, String replyText, String exchangeName, String routingKey, BasicProperties properties, byte[] message) throws IOException {
-            String logMessage = "Message destined for '" + exchangeName + "' exchange with '" + routingKey
-                    + "' as routing key came back with replyCode=" + replyCode + " and replyText=" + replyText + ".";
-            processorLog.warn(logMessage);
+            final String reason = "exchange='" + exchangeName + "' routingKey='" + routingKey
+                    + "' replyCode=" + replyCode + " replyText='" + replyText + "'";
+            undeliverableReturnReason.set(reason);
```
turcsanyip: Line 45 says: _Only populated when {@link #useConfirms} is true._
So an `if` should be added:

```suggestion
            if (useConfirms) {
                undeliverableReturnReason.set(reason);
            }
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AMQPPublisher.java
```diff
@@ -83,23 +139,24 @@ public String toString() {
...
+     * {@code waitForConfirms()} synchronizes the two threads and throw an {@link AMQPException}
+     * to trigger REL_FAILURE routing.
      */
     private final class UndeliverableMessageLogger implements ReturnListener {
         @Override
         public void handleReturn(int replyCode, String replyText, String exchangeName, String routingKey, BasicProperties properties, byte[] message) throws IOException {
-            String logMessage = "Message destined for '" + exchangeName + "' exchange with '" + routingKey
-                    + "' as routing key came back with replyCode=" + replyCode + " and replyText=" + replyText + ".";
-            processorLog.warn(logMessage);
+            final String reason = "exchange='" + exchangeName + "' routingKey='" + routingKey
```
turcsanyip: It seems to me that the prefix for the reason is "Message returned as undeliverable by broker: " in both cases where the text is used (logging and exception). So the prefix could be added right here.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/PublishAMQPTest.java
```diff
@@ -278,5 +278,43 @@ protected Connection createConnection(ProcessContext context, ExecutorService ex
...
+        final TestRunner testRunner = TestRunners.newTestRunner(proc);
+        setConnectionProperties(testRunner);
+        testRunner.setProperty(PublishAMQP.DELIVERY_GUARANTEE, PublishAMQP.DeliveryGuarantee.AT_LEAST_ONCE);
+        proc.getTestChannel().setSimulateShutdownOnConfirm(true);
+
+        testRunner.enqueue("Hello Joe".getBytes());
+        testRunner.run();
+
+        assertTrue(testRunner.getFlowFilesForRelationship(PublishAMQP.REL_SUCCESS).isEmpty());
+        assertNotNull(testRunner.getFlowFilesForRelationship(PublishAMQP.REL_FAILURE).getFirst());
```
turcsanyip: To assert that the FlowFile has been transferred to a given relationship we typically use:
```
        testRunner.assertTransferCount(PublishAMQP.REL_SUCCESS, 0);
        testRunner.assertTransferCount(PublishAMQP.REL_FAILURE, 1);
```

Or it could be even simpler here:
```
        testRunner.assertAllFlowFilesTransferred(PublishAMQP.REL_FAILURE);
```

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/PublishAMQPTest.java
```diff
@@ -278,5 +278,43 @@ protected Connection createConnection(ProcessContext context, ExecutorService ex
...
+            return connection.getTestChannel();
+        }
+    }
+
+    /**
+     * When the broker closes the channel with a 404 (exchange not found), the FlowFile
+     * must route to REL_FAILURE — not cause an unhandled processor exception.
+     */
+    @Test
+    public void validateFlowFileRoutedToFailureWhenBrokerClosesChannel() {
```
turcsanyip: Please move these test methods up, just after the last existing test method (and before the private method / nested class).

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/AMQPPublisherTest.java
```diff
@@ -105,4 +104,66 @@ public void validateSuccessfulPublishingAndUndeliverableRoutingKey() throws Exce
...
     }
 
+    /**
+     * Verifies that a {@link com.rabbitmq.client.ShutdownSignalException} thrown by
+     * {@code waitForConfirms()} (e.g., broker closes channel with 404 NOT_FOUND because the
+     * exchange does not exist) is converted to {@link AMQPException} so the FlowFile routes
+     * to REL_FAILURE instead of surfacing as an unhandled processor error.
+     */
+    @Test
+    public void failPublishWhenBrokerClosesChannelDuringConfirm() {
```
turcsanyip: Similar to `succeedsPublishWhenMessageUndeliverableInAtMostOnceMode()` test method name, `*InAtLeastOnce` suffix could be added in these methods.

turcsanyip (changes requested):
@rakesh-rsky Thanks for adding `Delivery Guarantee` property. Both options work as expected.

I only left a few minor code improvement comments.

turcsanyip:
@rakesh-rsky Could you please request for "Jira contributor access" as described in the [Contributor Guide](https://cwiki.apache.org/confluence/display/NIFI/Contributor+Guide)? This will allow you to administer the Jira tickets properly (assign, status change, etc.).
Thanks

rakesh-rsky:
@turcsanyip  Thanks for the detailed review. All inline comments have been addressed.

Regarding the Jira contributor access, I'll request it via the Contributor Guide. Thanks for the pointer.

turcsanyip (approved):
Thanks @rakesh-rsky!

+1 merging

PR #11240, opened 2026-05-12
NIFI-15932: Implemented ability to migrate a Versioned flow's Assets …

markap14:
…and config to a new Connector

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/endpoints/MigrationRequestEndpointMerger.java
```diff
@@ -0,0 +1,107 @@
...
+
+            if (nodeRequest.getFailureReason() != null) {
+                final String nodeReason = "Node " + nodeId.getApiAddress() + ":" + nodeId.getApiPort() + ": " + nodeRequest.getFailureReason();
+                final String existingReason = clientRequest.getFailureReason();
+                clientRequest.setFailureReason(existingReason == null ? nodeReason : existingReason + "; " + nodeReason);
+            }
+
+            final Date clientLastUpdated = clientRequest.getLastUpdated();
+            final Date nodeLastUpdated = nodeRequest.getLastUpdated();
+            if (nodeLastUpdated != null && (clientLastUpdated == null || nodeLastUpdated.before(clientLastUpdated))) {
```
rfellows: The condition uses `nodeLastUpdated.before(clientLastUpdated)`, which means the merged `lastUpdated` will reflect the **earliest** timestamp across nodes. The neighboring merges (`setComplete = AND`, `setPercentCompleted = min`) are "worst-of" for completion progress, but for an activity timestamp a polling client would typically expect the **most recent** update across the cluster — otherwise the merged value can appear to go backward as more nodes report in.

Is the intent here to surface the oldest timestamp (and if so, what does that represent for the client), or should this be `nodeLastUpdated.after(clientLastUpdated)`? `MigrationRequestEndpointMergerTest` doesn't currently assert anything about the merged `lastUpdated`, so either direction would pass today — happy to suggest a test once the intent is confirmed.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorRepository.java
```diff
@@ -664,7 +675,16 @@ private void waitForState(final ConnectorNode connector, final Set<ConnectorStat
...
         while (true) {
-            final ConnectorState clusterState = requestReplicator.getState(connector.getIdentifier());
+            final ConnectorState clusterState;
+            try {
+                clusterState = requestReplicator.getState(connector.getIdentifier());
+            } catch (final IOException e) {
+                final long elapsedSeconds = Duration.ofNanos(System.nanoTime() - startNanos).toSeconds();
+                logger.warn("Failed to retrieve cluster state for {} while waiting for update completion; elapsed time = {} secs", connector, elapsedSeconds, e);
+                Thread.sleep(Duration.ofSeconds(1));
+                continue;
```
rfellows: The new `catch (IOException)` block logs and `continue`s, but the only place `iterations` is incremented (and therefore the only termination opportunity short of a non-`IOException` state outcome) is the `allowableStates` branch below — which isn't reached if `requestReplicator.getState(...)` keeps throwing.

If the request replicator is persistently unhealthy, would this loop ever exit, or would the caller block indefinitely? Was an upper bound intended here — e.g. using `startNanos` to fail after a max duration, or bailing out after N consecutive `IOException`s? Happy to be wrong if there's an outer cancellation/timeout I'm not seeing.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorMigrationManager.java
```diff
@@ -0,0 +1,758 @@
...
+            logger.error("Failed to restore Connector {} to its initial state after migration failure; manual cleanup may be required",
+                    connector.getIdentifier(), e);
+            return;
+        }
+
+        flowController.getConnectorRepository().discardWorkingConfiguration(connector);
+    }
+
+    private void clearConnectorComponentState(final ConnectorNode connector) {
+        // Migration is only permitted on a Connector whose active flow is empty, so every Processor and Controller
```
rfellows: The comment here states "Migration is only permitted on a Connector whose active flow is empty", but the precondition check `verifyTargetIsAtInitialFlow` (~L603) only asserts `matchesInitialFlow()` — which is true whenever the live managed flow equals `getInitialFlow()`, including the case where the initial flow itself declares processors or controller services.

If a `MigratableConnector` implementation ever ships with a non-empty `getInitialFlow()` containing stateful components, would a Phase 2 failure here wipe state belonging to those originally-declared components? Is the empty-initial-flow assumption something we want to:

1. tighten in `verifyTargetIsAtInitialFlow` (also assert the managed flow is structurally empty),
2. track explicitly (snapshot pre-migration component IDs and only clear "new since snapshot"), or
3. document as a contract requirement on `MigratableConnector` / the developer guide?

If today's expectation is "(3) by convention", a one-line note in the developer guide and a code comment here would make the contract clear to future implementers.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ConnectorResource.java
```diff
@@ -2353,6 +2403,342 @@ public Response getConnectorRemoteProcessGroupStatusHistory(
...
+        return withWriteLock(
+                serviceFacade,
+                requestEntity,
+                lookup -> {
+                    final Authorizable connector = lookup.getConnector(connectorId);
+                    connector.authorize(authorizer, RequestAction.WRITE, user);
+                },
+                () -> {
+                    if (hasLocalSource) {
+                        serviceFacade.verifyCanMigrateConnector(connectorId, request.getLocalSource().getProcessGroupId());
```
rfellows, line 2577: Two related questions about concurrency on this submission path:

1. `withWriteLock` is revision-based and the verifier runs synchronously, but the actual migration is dispatched to `migrationRequestManager` and runs asynchronously. Is there a per-Connector serialization mechanism I'm missing that prevents two near-simultaneous submissions against the same target Connector from both passing `verifyTargetIsAtInitialFlow` and then racing in `applyMigratedConfiguration`? If not, would a per-Connector lock at the manager level — or introducing an explicit `MIGRATING` `ConnectorState` to gate transitions — be worth considering? The asymmetric-failure cluster IT exercises per-node divergence but doesn't appear to exercise two parallel requests against the same Connector.
2. The `verifyCanMigrateConnector` call is intentionally guarded by `hasLocalSource`, so for uploaded-payload sources the structural / `matchesInitialFlow` checks only fire inside the async task and surface several poll cycles later. Was that intentional? Would it be reasonable to run an equivalent synchronous verify (Connector state + `matchesInitialFlow`) for the uploaded-payload case too, so the user sees the failure on submission rather than after polling?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -6290,7 +6351,7 @@ public RegisteredFlowSnapshot getCurrentFlowSnapshotByGroupId(final String proce
                 .mapInstanceIdentifiers(false)
                 .mapControllerServiceReferencesToVersionedId(true)
                 .mapFlowRegistryClientId(false)
-                .mapAssetReferences(false)
+                .mapAssetReferences(true)
```
rfellows: Flipping `mapAssetReferences` from `false` → `true` on the shared `getCurrentFlowSnapshotByGroupId` mapping options changes the result for **every** caller of this path, not just the migration source-flow capture — e.g. process-group export and any other consumer of `buildFlowSnapshot` via this method.

Is the intent to start including asset references in those snapshots universally, or is the change really only meant for migration? If the latter, would it make sense to route the migration-source capture through a dedicated method (or add an opt-in parameter), so the existing export consumers don't inadvertently start emitting NiFi-internal asset identifiers that may not resolve on a different instance?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ConnectorResource.java
```diff
@@ -2353,6 +2403,342 @@ public Response getConnectorRemoteProcessGroupStatusHistory(
...
+                    .responseClass(MigrationPayloadEntity.class)
+                    .successfulResponseStatus(HttpResponseStatus.OK.getCode())
+                    .build();
+            final MigrationPayloadEntity entity = uploadRequestReplicator.upload(uploadRequest);
+            return generateOkResponse(entity).build();
+        }
+
+        final RegisteredFlowSnapshot flowSnapshot;
+        try {
+            flowSnapshot = OBJECT_MAPPER.readValue(payloadContents, RegisteredFlowSnapshot.class);
```
rfellows: The payload upload reads the JSON in full via `OBJECT_MAPPER.readValue(payloadContents, ...)` and pins the deserialized snapshot in `migrationPayloadsById` for up to ~10 minutes, but there's no `MaxLengthInputStream` wrap or other size cap on the upload. The asset upload path in this same file caps at `MAX_ASSET_SIZE_BYTES = 1 GB` for comparable resource reasons.

Is the expectation that the 10-min TTL + per-Connector WRITE authorization is enough bounding, or would a `MaxLengthInputStream` wrap here be worth adding (probably with a tighter cap than the 1 GB asset limit, since exported flows are typically a couple orders of magnitude smaller)?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorRepository.java
```diff
@@ -287,9 +290,19 @@ public ConnectorSyncResult syncConnector(final VersionedConnector versionedConne
...
+            // reference and may have temporarily removed references to the authoritative Assets. Deleting unreferenced
+            // Assets here would discard Assets the user still needs. Orphaned Assets are reclaimed by the next
+            // authoritative configuration change, which is only permitted once Troubleshooting mode has ended.
             return ConnectorSyncResult.syncedConfigUnchanged(connector, effectiveScheduledState);
         }
 
+        // Clean up assets after configuration has been inherited so that the Connector's flow contexts reflect the
+        // synchronized configuration; performing this before synchronization would treat still-referenced assets as
+        // unreferenced because the managed flow has not yet been populated.
+        cleanUpAssets(connector);
```
rfellows: `cleanUpAssets(connector)` is now invoked on every successful `syncConnector(...)` (a broader trigger than before this PR), and together with the new `collectReferencedParameterContextAssetIds` it's necessary for asset rollback to behave correctly after a failed migration. I want to confirm the lifecycle relationship to migration: is `syncConnector(...)` guaranteed never to run concurrently with an in-flight `applyMigratedConfiguration(...)` on the same Connector?

If a sync ever interleaved with the migration's mid-Phase-2 state — after the new working configuration is built but before it's committed — could `cleanUpAssets` see the newly-copied assets as unreferenced and delete them? Even if that interleaving isn't reachable today, a one-line comment locking the invariant in place (or a targeted test) would make this much harder to regress.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/audit/ConnectorAuditor.java
```diff
@@ -372,6 +372,29 @@ public void applyConnectorUpdateAdvice(final ProceedingJoinPoint proceedingJoinP
...
+
+        proceedingJoinPoint.proceed();
+
+        if (isAuditable()) {
+            final FlowChangeConfigureDetails actionDetails = new FlowChangeConfigureDetails();
+            actionDetails.setName("Connector migrated from Versioned Process Group");
+            actionDetails.setPreviousValue(null);
+            actionDetails.setValue(processGroupId == null ? "Uploaded payload" : processGroupId);
+
+            final Action action = generateAuditRecord(connector, Operation.Configure, actionDetails);
```
rfellows, line 391: Migration is recorded as `Operation.Configure`, but semantically it's distinct from an ordinary configuration change — it can copy assets, write component state, install a managed Process Group, disable and rename the source PG, and is effectively irreversible. Operators filtering the audit log for migration events would currently have to look for `Operation.Configure` rows whose details `name` starts with "Connector migrated from…".

Would it be reasonable to introduce a dedicated `Operation.Migrate` (alongside the existing operations) and use it here, so migrations surface as a first-class event type in the audit trail?

markap14: Agreed that a migration is semantically distinct from an ordinary `Operation.Configure` and would be better as a first-class audit event. `Operation` lives in `nifi-api`, so adding a value is a public-API change that also touches audit persistence, the audit UI filters, and the developer docs. To keep this PR focused, this is best done as a separate follow-up effort rather than folded in here.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/api/ConnectorResource.java
markap14: On (2): good point — fixed. The uploaded-payload path now runs the same synchronous readiness check (connector state + `matchesInitialFlow`) as the local-source path, via `verifyConnectorReadyForMigration`, so the user sees the failure on submission instead of several poll cycles later.

On (1): there's no per-Connector migration lock today. Submissions are serialized by the revision-based `withWriteLock` and re-verified both synchronously at submission and again at the start of the async task, so two requests that share a starting revision can't both proceed. The residual window is two requests that each pick up a fresh revision and both pass verification before either commits `applyMigratedConfiguration`. This matches how ordinary async connector updates behave, so I've left it as-is for now rather than introducing a manager-level lock or an explicit `MIGRATING` `ConnectorState`.

rfellows (approved):

nifi-docs/src/main/asciidoc/administration-guide.adoc
```diff
@@ -3852,6 +3852,93 @@ The Secrets Manager delegates to Parameter Providers to retrieve secret values.
...
+
+When the source is a local Process Group on the same NiFi instance, the Process Group must satisfy all of the following requirements before NiFi will allow the migration:
+
+- The source Process Group must be under version control.
+- The source Process Group must be `UP_TO_DATE` with the latest published version in Flow Registry.
+- All processors in the source Process Group must be stopped.
+- All controller services in the source Process Group must be disabled.
+- All queues in the source Process Group must be empty.
+- The source Process Group must not reference controller services outside of the Process Group.
+- The target Connector must be newly created, stopped, and not previously migrated.
```
bobpaulin:
```suggestion
- The target Connector must be not yet configured, stopped, and not previously migrated.
```
How recent the connector was created does not seem to come into play. It's whether it was configured prior to the attempt to migrate.

nifi-docs/src/main/asciidoc/administration-guide.adoc
```diff
@@ -3852,6 +3852,93 @@ The Secrets Manager delegates to Parameter Providers to retrieve secret values.
...
+
+=== Migration paths
+
+There are two ways to run the migration:
+
+- Local Process Group migration:
+  Select an eligible Process Group that already exists on the current NiFi canvas.
+  NiFi validates the prerequisites listed above before starting the migration.
+- Uploaded payload migration:
+  Export a Process Group definition and upload the resulting flow snapshot payload to the Connector migration endpoint.
```
bobpaulin: May want to mention the exports should include state (if it exists).

nifi-docs/src/main/asciidoc/administration-guide.adoc
```diff
@@ -3852,6 +3852,93 @@ The Secrets Manager delegates to Parameter Providers to retrieve secret values.
...
+The Connector's own migration support check determines whether the uploaded flow can be imported.
+
+=== Running a migration
+
+. Create a new Connector instance.
+. If the source Process Group is local and registry-backed, choose it from the Connector migration source list.
+. If the source Process Group is not eligible for the local path, export it with component state included and upload the payload to the Connector migration payload endpoint.
+. Start the Connector migration request.
+. Poll the request until it completes.
+. Open the Connector configuration and supply any missing sensitive values or secrets.
```
bobpaulin:
```suggestion
. Open the Connector configuration and supply any missing secrets.
```

Since only Connector Secrets are supported I'm not sure mentioning sensitive values make sense unless referring to the source flow.

nifi-docs/src/main/asciidoc/administration-guide.adoc
```diff
@@ -3852,6 +3852,93 @@ The Secrets Manager delegates to Parameter Providers to retrieve secret values.
 |`nifi.secrets.manager.cache.duration`|The duration for which resolved secret values are cached before being refreshed from the underlying Parameter Providers. Accepts any NiFi time duration value such as `5 mins`, `30 secs`, etc. A value of `0 sec` disables caching entirely. Defaults to `5 mins`.
 |====
 
+== Migrating a Connector from a Versioned Process Group
+
+NiFi supports a one-time path that migrates a newly created Connector from a source Versioned Process Group.
+The source Versioned Process Group is used as a reference: the Connector reads it and updates its own managed flow to mirror the source's configuration, parameters, and component state.
+The source flow itself is not installed onto the Connector.
+Only Connectors that explicitly opt in to migration support can use this capability.
```
bobpaulin:
```suggestion
Only Connectors that implement migration support can use this capability.
```

They opt in by implementing the support interfaces. While this is outlined in the developer guide an administrator may confuse opt in for some type of configuration rather than code.

nifi-docs/src/main/asciidoc/developer-guide.adoc
```diff
@@ -2697,6 +2697,126 @@ The following command is used to generate a standard binary distribution of Apac
 
 `mvn clean install -Pcontrib-check`
 
+== Supporting migration from a Versioned Process Group
+
+Connectors can opt in to a one-time migration flow that uses a source Versioned Process Group as a reference for populating a freshly created Connector.
```
bobpaulin:
```suggestion
Connectors can implement a one-time migration flow that uses a source Versioned Process Group as a reference for populating a freshly created Connector.
```

nifi-docs/src/main/asciidoc/developer-guide.adoc
```diff
@@ -2697,6 +2697,126 @@ The following command is used to generate a standard binary distribution of Apac
...
+
+- The exported `VersionedProcessGroup`
+- Parameter contexts
+- Referenced assets
+- Attached `VersionedComponentState` for `@Stateful` components
+- Export metadata such as flow name, registry identifiers, and version
+
+The same context also indicates whether the request came from a local Process Group on the same NiFi instance or from an uploaded payload using `isLocalMigration()`.
+
+The active `FlowContext` is available via `context.getActiveFlowContext()` and is read-only during migration: it forwards reads (configuration, managed Process Group) to the live active flow context but rejects `updateFlow(...)` with `IllegalStateException`. The Connector should treat this object as a window onto the current state of the Connector, not as a write target.
```
bobpaulin: It is mentioned later that the applyUpdate will be performed to update the active context but it might be worth mentioning here as well.

nifi-docs/src/main/asciidoc/developer-guide.adoc
```diff
@@ -2697,6 +2697,126 @@ The following command is used to generate a standard binary distribution of Apac
...
+=== Rewriting parameters and assets
+
+The exported source flow contains parameter contexts and any asset references that were attached to exported parameters.
+Connectors are responsible for mapping those values into their own target shape inside `migrateConfiguration`.
+
+A Connector can copy an asset from a local source Process Group into its own asset namespace using `ConnectorMigrationContext.copyAssetFromSource(String)`.
+This method is available only for local migrations.
+It throws `IllegalArgumentException` when the supplied source asset identifier is null or blank, and `IllegalStateException` when invoked for an uploaded-payload migration.
+Uploaded payloads do not have access to a live source asset namespace, so asset references from uploaded payloads must be handled outside of the migration request.
+
```
bobpaulin, line 2803: Code snippet for Asset Migration might be helpful.  As asset needs to be migrated and reference set.

```
    final ConnectorMigrationContext context = ...
    final VersionedParameter sourceParameter = ...
    final String targetStepName = ...
    final String targetPropertyName = ...
    if (sourceParameter == null || sourceParameter.getReferencedAssets() == null
            || sourceParameter.getReferencedAssets().isEmpty() || !context.isLocalMigration()) {
        return;
    }
    final Set<String> copiedAssetIds = new LinkedHashSet<>();
    for (final VersionedAsset asset : sourceParameter.getReferencedAssets()) {
        if (asset.getIdentifier() != null && !asset.getIdentifier().isBlank()) {
            copiedAssetIds.addAll(context.copyAssetFromSource(asset.getIdentifier()).getAssetIdentifiers());
        }
    }
    if (!copiedAssetIds.isEmpty()) {
        context.setValueReference(targetStepName, targetPropertyName, new AssetReference(copiedAssetIds));
    }
```

nifi-docs/src/main/asciidoc/developer-guide.adoc
```diff
@@ -2697,6 +2697,126 @@ The following command is used to generate a standard binary distribution of Apac
...
+- Migration is rejected when the Connector's currently managed flow no longer matches the flow returned by its `getInitialFlow()`.
+  The framework compares the two structurally (ignoring environmental differences such as component scheduled state) at the time `migrateConfiguration` is invoked, and again whenever the `MIGRATE` allowable action is evaluated, so the user is shown why migration is unavailable.
+  This check is restart-safe because both inputs are recomputed each time from durable sources: the Connector's persisted configuration drives `getInitialFlow()`, and the current managed flow is read from the live flow tree restored from `flow.json.gz`.
+- Migration is one-shot from the framework's point of view.
+  Once the Connector's flow has been updated by a migration, its currently managed flow no longer matches `getInitialFlow()`, and a subsequent migration request will be rejected until the Connector is discarded and a fresh one is created.
```
bobpaulin: Is the point about getInitialFlow still true with the recent changes?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorMigrationManager.java
```diff
@@ -0,0 +1,832 @@
...
+                workingConfiguration,
+                flowController.getAssetManager(),
+                flowController.getConnectorRepository(),
+                flowController.getStateManagerProvider(),
+                flowController
+        );
+
+        final Connector rawConnector = connector.getConnector();
+        if (!(rawConnector instanceof final MigratableConnector migratableConnector)) {
+            throw new FlowUpdateException("Connector " + connectorId + " does not support migration from the provided source flow.");
```
bobpaulin:
```suggestion
            throw new FlowUpdateException("Connector " + connectorId + " does not support migration.");
```

Nothing wrong with the source flow.  Connector just doesn't support migration

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorMigrationManager.java
```diff
@@ -0,0 +1,832 @@
...
+        final int externalControllerServiceCount = countExternalControllerServices(sourceFlow);
+        if (externalControllerServiceCount > 0) {
+            reasons.add("The Process Group references " + externalControllerServiceCount + " controller "
+                    + pluralizeNoun(externalControllerServiceCount, "service", "services") + " from outside the Process Group.");
+        }
+
+        return reasons;
+    }
+
+    private static String pluralizeIsAre(final long count) {
```
bobpaulin: While correct grammatically does this make log analysis harder?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorMigrationManager.java
```diff
@@ -0,0 +1,832 @@
...
+        return count == 1L ? "is" : "are";
+    }
+
+    private static String pluralizeNoun(final long count, final String singular, final String plural) {
```
bobpaulin: While correct grammatically does this make log analysis harder?

bobpaulin (approved):
Thanks for this update @markap14 .  This provides what we need to migration from versioned flows to connectors.  I don't see any of this feedback as blocking but much of it can incrementally improve documentation and logging.  Approve non-binding.

nifi-docs/src/main/asciidoc/administration-guide.adoc
markap14: [Sonnet 5] Updated as suggested.

nifi-docs/src/main/asciidoc/administration-guide.adoc
markap14: [Sonnet 5] Good call — updated the uploaded payload migration step to mention exporting component state.

nifi-docs/src/main/asciidoc/administration-guide.adoc
markap14: [Sonnet 5] Updated as suggested.

nifi-docs/src/main/asciidoc/administration-guide.adoc
markap14: [Sonnet 5] Updated as suggested.

nifi-docs/src/main/asciidoc/developer-guide.adoc
markap14: [Sonnet 5] Updated as suggested.

nifi-docs/src/main/asciidoc/developer-guide.adoc
markap14: [Sonnet 5] Added a mention of applyUpdate to this section as well.

nifi-docs/src/main/asciidoc/developer-guide.adoc
markap14: [Sonnet 5] Added a code example for asset migration to the developer guide, based on the pattern used by MigrationTargetConnector in the system test extensions.

nifi-docs/src/main/asciidoc/developer-guide.adoc
markap14: [Sonnet 5] No — this text was stale relative to the eligibility check added in the July 16 commit, which now uses Connector.isModified() (configuration deviates from defaults, or any managed component has stored state) rather than a structural comparison against getInitialFlow(). Reworded this section to describe the current check.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorMigrationManager.java
markap14: [Sonnet 5] Updated as suggested.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorMigrationManager.java
markap14: [Sonnet 5] Agreed — replaced the is/are and singular/plural noun logic with a fixed count-based phrasing (e.g. "Running or enabled processor count: 1.") so the wording is consistent regardless of count, and updated the corresponding unit tests.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorMigrationManager.java
markap14: [Sonnet 5] Same fix as the sibling comment above — replaced with fixed count-based phrasing.

markap14:
[Sonnet 5] Pushed a commit addressing @bobpaulin's documentation feedback: reworded the administration guide's migration prerequisites/steps, updated the developer guide's description of the migration eligibility check to match the isModified()-based check added in the July 16 commit, added an asset-migration code example, tweaked an exception message, and replaced the singular/plural ineligibility-reason wording with fixed count-based phrasing (updated the corresponding unit tests too). Replied inline to each comment. Will keep watching the PR for CI results and further feedback.

markap14:
[Sonnet 5] Pushed a follow-up fix: the `macos-15 Java 21` system-tests check failed because `ConnectorVersionedFlowMigrationIT` still asserted on the old singular/plural ineligibility-reason wording, which I changed to a fixed count-based format in the previous commit. Updated the test's expected substrings, rebuilt the framework NAR and system test runtime, and confirmed `ConnectorVersionedFlowMigrationIT` passes locally. Continuing to watch CI.

ncover21:
While testing the MIGRATE action I ran into a case where a freshly-created Connector is reported as modified, which blocks migration.

`createMigrateAction` gates MIGRATE on `isModified()`, and `StandardConnectorNode.configurationDiffersFromDefaults(...)` treats any `ConnectorValueReference` that isn't a `StringLiteralValue` as a modification:

```java
if (valueReference instanceof final StringLiteralValue stringLiteralValue) {
    if (!Objects.equals(stringLiteralValue.getValue(), propertyDefaults.get(propertyName))) {
        return true;
    }
} else {
    // any SecretReference / AssetReference returns true
    return true;
}
```

The issue is the `else` branch: a `SecretReference` or `AssetReference` that is structurally empty (null fully-qualified/secret name, or null/empty asset identifiers) represents an *unset* property, but it still returns `true`. So any Connector that declares a Secret- or Asset-typed property and has never been configured is reported as modified, and `createMigrateAction` disallows migration with "Connector has been modified since it was created."

This stays hidden when the configuration is assembled in-memory from `StringLiteralValue`s with unset secret/asset properties simply omitted from the map - the `else` branch is never reached. It surfaces once the working configuration is populated from a serialized/round-tripped configuration in which an unset secret/asset property is represented as a typed reference with null contents rather than being omitted; deserialized, that becomes an empty `SecretReference`/`AssetReference` and trips the `else`.

The framework already handles this exact shape in `resolvePropertyReferences`, which skips structurally-empty secret references via `isEmptySecretReference(...)` ("A typed-but-empty SECRET_REFERENCE acts like an unset property"). `configurationDiffersFromDefaults` should apply the same rule:

- treat a structurally-empty `SecretReference` (null FQN and null secret name) or `AssetReference` (null/empty asset identifiers) as unset, i.e. equal to the default and not a modification;
- only treat a *populated* Secret/Asset reference as a modification.

The existing `TestStandardConnectorNode` cases only exercise `StringLiteralValue`s (plus one populated `SecretReference`), so the empty-typed-reference path isn't covered. A regression test that seeds the working configuration with an empty `SecretReference`/`AssetReference` and asserts `isModified()` is `false` (and a populated one asserts `true`) would lock it down.

markap14:
[Sonnet 5] @ncover21 good catch, thanks. Confirmed the bug: `configurationDiffersFromDefaults` treated any non-`StringLiteralValue` reference as a modification, including a structurally-empty `SecretReference` or `AssetReference` — exactly the placeholder-for-unset-property shape the framework already special-cases elsewhere (`isEmptySecretReference`, `isReferenceAllowed`). Reproduced the failure with a unit test that seeds an empty `SecretReference`/`AssetReference` and asserts `isModified()` returns `false`, confirmed it failed as described, then fixed `configurationDiffersFromDefaults` to skip structurally-empty typed references, and added a test confirming a *populated* `AssetReference` is still correctly treated as a modification. Pushed.

ncover21:
The migration-complete path doesn't persist configuration through the `ConnectorConfigurationProvider`.

`StandardConnectorRepository.notifyMigrationComplete(...)` only invokes the `ConnectorConfigurationProvider.migrationComplete(...)` hook, which is a default no-op. Unlike `configureConnector(...)` / `updateConnector(...)`, it never calls `configurationProvider.save(...)`. So a provider that persists working configuration externally (rather than relying solely on `flow.json.gz`) never receives the migrated configuration after a commit - the externally persisted config stays absent/stale until some later configure/update call triggers a `save`.

The repository already has what's needed to close this: the `connectors` map, the `configurationProvider` reference, and the existing `buildWorkingConfiguration(ConnectorNode)` helper the configure/update paths use. So the save can live entirely in the repository:

```java
public void notifyMigrationComplete(String connectorId, String sourceProcessGroupId) {
    if (configurationProvider != null) {
        ConnectorNode connector = connectors.get(connectorId);
        if (connector != null) {
            configurationProvider.save(connectorId, buildWorkingConfiguration(connector));
        }
        configurationProvider.migrationComplete(connectorId, sourceProcessGroupId);
    }
}
```

Working-vs-active is safe here: `commitMigratedConfiguration(...)` writes the merged config onto the active context and then calls `recreateWorkingFlowContext()`, and `notifyMigrationComplete(...)` runs after the commit, so `buildWorkingConfiguration(connector)` (which reads the working context) reflects the migrated configuration at that point.

markap14:
[Sonnet 5] @ncover21 confirmed, thanks — good catch again. `notifyMigrationComplete(...)` only called the `migrationComplete(...)` hook, unlike `configureConnector(...)`/`updateConnector(...)` which always `save(...)` after a change. Reproduced with a unit test asserting `provider.save(...)` is invoked after `notifyMigrationComplete(...)`, confirmed it failed (save was never called), then added the save using the exact approach you suggested — `buildWorkingConfiguration(connector)`, which reflects the merged configuration since `commitMigratedConfiguration(...)` already rebuilt the working flow context from it before this runs. Pushed.

markap14:
[Sonnet 5] Two CI shards failed on the current head commit (7b3c4a0), both unrelated to this PR's changes (which only touch Connector-migration framework code, not Couchbase or Python/clustering test infrastructure):

**`Ubuntu Java 21`** — https://github.com/apache/nifi/actions/runs/29779864353/job/88478247612
```
[ERROR] PutCouchbaseIT.testPutDocument:61 Expected all Transferred FlowFiles to go to success but 1 were routed to failure
```
Preceding logs show the Couchbase Java client repeatedly failing to refresh its bucket config against the test Couchbase container (`NO_MORE_RETRIES (NODE_NOT_AVAILABLE)`), ending in "Failed to upsert document into Couchbase." This is a test-container connectivity issue in `nifi-couchbase-processors`, a module this PR does not touch.

**`ubuntu-24.04 Java 25`** — https://github.com/apache/nifi/actions/runs/29779864352/job/88478247652
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 376.9 s <<< FAILURE! -- in org.apache.nifi.tests.system.parameters.ClusteredProviderParamFlowSyncIT
java.lang.RuntimeException: Waited up to 60 seconds for all 2 nodes to connect but only -1 nodes connected. Node states: (failed to retrieve node details: Connection refused)
```
Preceded by `py4j.Py4JException` communication-channel failures against a Python processor subprocess. This is a clustered-node-startup timing issue in a Parameter Provider system test unrelated to this PR's changes.

Reran both failed shards (`gh run rerun --failed`) rather than the full pipeline. Will watch for the rerun results.

PR #11241, opened 2026-05-12
NIFI-15930: Re-init Secrets from a Provider with Controller Service

bobpaulin:
* Pre-Enable Parameter Provider Controller Services to ensure they are valid prior to secret resolution
* Enable Working Context to apply any changes in Connector Property Values and Resolved values to the Parameter Context of the Working Flow Context
* Warn when a parameter provider is not found

# Summary

[NIFI-15930](https://issues.apache.org/jira/browse/NIFI-15930)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/components/connector/secrets/ParameterProviderSecretsManager.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/components/connector/secrets/ParameterProviderSecretsManager.java
@@ -50,6 +50,12 @@ public class ParameterProviderSecretsManager implements SecretsManager {
     private Duration cacheDuration;
     private final Map<String, CachedSecret> secretCache = new ConcurrentHashMap<>();
 
+    // Per-ParameterProvider-id deduplication for the WARN log emitted whenever a SecretReference
+    // resolution is skipped because its backing provider is not VALID. An entry is present for any
+    // provider id we have already surfaced via WARN; the value is the status that was logged so the
+    // same WARN is not repeated until the status either changes or the provider returns to VALID.
+    private final Map<String, ValidationStatus> lastLoggedStatus = new ConcurrentHashMap<>();
+
     private record CachedSecret(Secret secret, long timestampNanos) {
     }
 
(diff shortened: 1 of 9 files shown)


markap14:
Hey @bobpaulin thanks for looking into this. Looking at the code, I think we can probably simplify things quite a bit. In particular, we're picking and choosing which Controller Services we "pre-enable" and then handling those separately from the rest. I presume this is due to this comment in the code:
```
                // Inherit Parameter Providers and Connectors first. Because Connectors are a bit different, in that updates could result in Exceptions being thrown,
                // due to the fact that they manipulate the flow, and changes can be aborted, we handle them first. This way, if there's any Exception,
                // we can fail before updating parts of the flow that are not managed by Connectors.
                // Because Connectors may depend on Parameter Providers, we need to ensure that we inherit Parameter Providers first.
```

But I don't think this is really necessary. Instead, what if we just inherit the Controller Services before we inherit the Parameter Providers & Connectors? That comment was really more of a "let's do connectors first, just in case there are issues." So we'd avoid inheriting a Controller Service locally if there was a failure inheriting Connectors. At the time, it seemed like this was largely "free." But now, this approach introduces significant complexity and confusion because Controller Services are selectively handled here or there.

IMO, I think we should look at just inheriting the root-level Controller Services before Parameter Providers.

exceptionfactory (commented):
Concurring with @markap14, I agree that inheriting controller-level Controller Services before Parameter Providers should be a more straightforward way to go.

bobpaulin:
Hey @markap14 and @exceptionfactory thanks for the early review.  Unfortunately I discovered that this PR was not sufficient to address the issue.  The reason is this PR does not address already live Root Controller Services (ones not in the data flow).  Seems like the trouble is some of the Parameter Providers are not valid because the already live controller services have not fully enabled.  This causes the secrets from the providers to initialize to null.  I've raised another PR that that allows the connector to pause to wait for these controller services to complete the enable phases.  https://github.com/apache/nifi/pull/11252

I'm going to close this one and work to incorporate this feedback into the new PR.

PR #11242, opened 2026-05-13
NIFI-15934 Handle oversized FlowFiles and empty header keys in PublishAMQP

ing-mattioni:
# Summary

[NIFI-15934](https://issues.apache.org/jira/browse/NIFI-15934)

This pull request updates `PublishAMQP` to improve handling of oversized FlowFiles and AMQP header attributes.

`PublishAMQP` reads FlowFile content into a byte array before publishing an AMQP message. Since Java byte arrays require an `int` length, FlowFiles larger than `Integer.MAX_VALUE` cannot be published using this path. This change validates the FlowFile size before attempting byte-array allocation and routes oversized FlowFiles to `failure` with penalization.

This pull request also updates AMQP header parsing to ignore empty header keys instead of adding them to the AMQP headers map.

P.S. The existing unit test for AMQP header parsing was extended to cover empty header keys. The oversized FlowFile guard is a defensive validation before byte-array allocation; additional targeted coverage may require avoiding real multi-GB content allocation.

## Changes

- Validates FlowFile size before extracting message content into a byte array
- Routes FlowFiles larger than `Integer.MAX_VALUE` to `failure` with penalization
- Uses `Math.toIntExact()` when converting FlowFile size to byte array length
- Ignores empty AMQP header keys when parsing AMQP header attributes
- Extends AMQP header parsing test coverage for empty header keys

# Tracking

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-15934) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

`.\mvnw.cmd -pl :nifi-amqp-processors -am test`

`.\mvnw.cmd -pl :nifi-amqp-processors -am -P contrib-check install -DskipTests`

`.\mvnw.cmd -pl :nifi-amqp-nar -am install -DskipTests`

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
+++ b/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
@@ -33,6 +33,7 @@ import org.apache.nifi.components.Validator;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.migration.PropertyConfiguration;
+import org.apache.nifi.processor.DataUnit;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
 import org.apache.nifi.processor.Relationship;
@@ -75,6 +76,7 @@ import java.util.stream.Stream;
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_CLUSTER_ID_ATTRIBUTE, description = "The ID of the AMQP Cluster"),
 })
 public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
+    private static final long MAXIMUM_INPUT_FLOWFILE_SIZE_LIMIT = 128 * 1024 * 1024L;
 
     public static final PropertyDescriptor EXCHANGE = new PropertyDescriptor.Builder()
             .name("Exchange Name")
@@ -107,6 +109,16 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             .required(true)
             .allowableValues(DeliveryGuarantee.class)
             .defaultValue(DeliveryGuarantee.AT_MOST_ONCE)
+            .build();
+    public static final PropertyDescriptor MAXIMUM_INPUT_FLOWFILE_SIZE = new PropertyDescriptor.Builder()
+            .name("Maximum Input FlowFile Size")
+            .description("Maximum size of an input FlowFile that will be read into memory before publishing. PublishAMQP reads FlowFile content into a byte array "
+                    + "before publishing, so FlowFiles larger than this value are routed to failure before content is read. Configure this value according to "
+                    + "broker limits and available JVM memory.")
+            .required(true)
+            .defaultValue("128 MB")
+            .expressionLanguageSupported(ExpressionLanguageScope.ENVIRONMENT)
+            .addValidator(StandardValidators.createDataSizeBoundsValidator(1, MAXIMUM_INPUT_FLOWFILE_SIZE_LIMIT))
             .build();
     public static final PropertyDescriptor HEADERS_SOURCE = new PropertyDescriptor.Builder()
             .name("Headers Source")
@@ -150,6 +162,7 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
                     EXCHANGE,
                     ROUTING_KEY,
                     DELIVERY_GUARANTEE,
+                    MAXIMUM_INPUT_FLOWFILE_SIZE,
                     HEADERS_SOURCE,
                     HEADERS_PATTERN,
                     HEADER_SEPARATOR
@@ -165,7 +178,7 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
     /**
      * Will construct AMQP message by extracting its body from the incoming {@link FlowFile}. AMQP Properties will be extracted from the
      * {@link FlowFile} and converted to {@link BasicProperties} to be sent along with the message. Upon success the incoming {@link FlowFile} is
-     * transferred to 'success' {@link Relationship} and upon failure FlowFile is penalized and transferred to the 'failure' {@link Relationship}
+     * transferred to 'success' {@link Relationship} and upon failure FlowFile is transferred to the 'failure' {@link Relationship}
      * <br>
      * <p>
      * NOTE: Attributes extracted from {@link FlowFile} are considered candidates for AMQP properties if their names are prefixed with
@@ -177,6 +190,14 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
     protected void processResource(final Connection connection, final AMQPPublisher publisher, ProcessContext context, ProcessSession session) throws ProcessException {
         FlowFile flowFile = session.get();
         if (flowFile == null) {
+            return;
+        }
+
+        final long maximumInputFlowFileSize = context.getProperty(MAXIMUM_INPUT_FLOWFILE_SIZE).evaluateAttributeExpressions().asDataSize(DataUnit.B).longValue();
+        if (flowFile.getSize() > maximumInputFlowFileSize) {
+            getLogger().warn("FlowFile {} with size {} bytes exceeds configured maximum input FlowFile size of {} bytes; routing to failure",
+                    flowFile, flowFile.getSize(), maximumInputFlowFileSize);
+            session.transfer(flowFile, REL_FAILURE);
             return;
         }
 
@@ -201,7 +222,7 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             session.rollback();
             throw e;
         } catch (AMQPException e) {
-            session.transfer(session.penalize(flowFile), REL_FAILURE);
+            session.transfer(flowFile, REL_FAILURE);
             throw e;
         }
 
@@ -235,7 +256,7 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
      * Extracts contents of the {@link FlowFile} as byte array.
      */
     private byte[] extractMessage(final FlowFile flowFile, ProcessSession session) {
-        final byte[] messageContent = new byte[(int) flowFile.getSize()];
+        final byte[] messageContent = new byte[Math.toIntExact(flowFile.getSize())];
         session.read(flowFile, in -> StreamUtils.fillBuffer(in, messageContent, true));
         return messageContent;
     }
@@ -332,14 +353,24 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
         for (String strEntry : strEntries) {
             final String[] kv = strEntry.split("=", -1); // without using limit, trailing delimiter would be ignored
             if (kv.length == 2) {
-                headers.put(kv[0].trim(), kv[1].trim());
+                addHeader(headers, amqpPropValue, strEntry, kv[0], kv[1].trim());
             } else if (kv.length == 1) {
-                headers.put(kv[0].trim(), null);
+                addHeader(headers, amqpPropValue, strEntry, kv[0], null);
             } else {
                 getLogger().warn("Malformed key value pair in AMQP header property ({}): {}", amqpPropValue, strEntry);
             }
         }
         return headers;
+    }
+
+    private void addHeader(final Map<String, Object> headers, final String amqpPropValue, final String strEntry, final String headerKey, final Object headerValue) {
+        final String trimmedHeaderKey = headerKey.trim();
+        if (trimmedHeaderKey.isEmpty()) {
+            getLogger().warn("Skipping AMQP header with empty key in property ({}): {}", amqpPropValue, strEntry);
+            return;
+        }
+
+        headers.put(trimmedHeaderKey, headerValue);
     }
 
     protected Pattern getPattern(ProcessContext context, InputHeaderSource selectedHeaderSource) {
--- a/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
+++ b/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
@@ -76,7 +76,7 @@ import java.util.stream.Stream;
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_CLUSTER_ID_ATTRIBUTE, description = "The ID of the AMQP Cluster"),
 })
 public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
-    private static final long MAXIMUM_INPUT_FLOWFILE_SIZE_LIMIT = 128 * 1024 * 1024L;
+    private static final long MAXIMUM_INPUT_FLOWFILE_SIZE_LIMIT = 512 * 1024 * 1024L;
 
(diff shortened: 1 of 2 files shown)


nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -220,7 +226,7 @@ public void migrateProperties(final PropertyConfiguration config) {
      * Extracts contents of the {@link FlowFile} as byte array.
      */
     private byte[] extractMessage(final FlowFile flowFile, ProcessSession session) {
-        final byte[] messageContent = new byte[(int) flowFile.getSize()];
+        final byte[] messageContent = new byte[Math.toIntExact(flowFile.getSize())];
```
pvillard31, line 261: is `Math.toIntExact` required since we filter out larger flowfiles before?

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -166,6 +166,12 @@ protected void processResource(final Connection connection, final AMQPPublisher
             return;
         }
 
+        if (flowFile.getSize() > Integer.MAX_VALUE) {
+            getLogger().error("FlowFile {} with size {} bytes is too large to publish as an AMQP message", flowFile, flowFile.getSize());
+            session.transfer(session.penalize(flowFile), REL_FAILURE);
```
pvillard31: we usually don't penalize when sending into the failure relationship unless we believe that the failure relationship is used as a self-loop on the processor (but this is not a good practice in terms of flow design given that we have the retry at framework level)

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -317,9 +323,18 @@ private Map<String, Object> validateAMQPHeaderProperty(final String amqpPropValu
...
+                    continue;
+                }
+                headers.put(key, kv[1].trim());
             } else if (kv.length == 1) {
-                headers.put(kv[0].trim(), null);
+                final String key = kv[0].trim();
+                if (key.isEmpty()) {
+                    continue;
+                }
+                headers.put(key, null);
```
pvillard31: Should the two empty-key branches log at the same level (both warn, or both silent), and could the trim plus empty check be extracted into a small helper so the behavior is identical for key=value and bare-key entries?

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -166,6 +166,12 @@ protected void processResource(final Connection connection, final AMQPPublisher
             return;
         }
 
+        if (flowFile.getSize() > Integer.MAX_VALUE) {
```
exceptionfactory: This does not seem like the right approach, the `Integer.MAX_VALUE` is a very large limit, but there should be other more sensible boundaries, probably lower.

exceptionfactory (changes requested):
On initial review, the proposed check on input FlowFile size looks questionable. It may need to be a configurable property.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
ing-mattioni: Addressed by adding a configurable outbound body-size property with a 64 MB default and upper bound, so oversized FlowFiles are routed to failure before reading content into memory.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
ing-mattioni: Resolved by checking against the configured size limit before allocation, so the remaining cast is bounded.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
ing-mattioni: Removed penalization when routing to failure and added a regression assertion.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
ing-mattioni: Extracted header insertion into a helper so empty keys are handled consistently for both key=value and bare-key entries.

ing-mattioni:
Thank you @pvillard31 and @exceptionfactory , I made the following changes:

  - Added configurable Max Outbound Message Body Size in nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java:100, default/bounded at 64 MB.
  - Oversized FlowFiles now route to failure before content is read into memory.
  - Removed penalization when routing publish failures to failure.
  - Centralized AMQP header insertion with consistent empty-key skip/warn behavior.
  - Added tests for oversized FlowFiles, non-penalized failures, and empty header keys in nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/PublishAMQPTest.java:177.

ing-mattioni:
Addressed the Checkstyle failure in the latest update.

The issue was caused by tab characters in `PublishAMQP.java`, which resulted in `FileTabCharacter` and indentation violations. I replaced the tabs with spaces, reran the local validation, and force-pushed the updated signed commit.

ing-mattioni:
The remaining test failure appears to be unrelated to this PR. The failed test is in `nifi-aws-processors`:

`org.apache.nifi.processors.aws.kinesis.stream.pause.TestStandardRecordProcessorBlocker.testBlockAndUnblock`

The failure is a timeout while waiting for an asynchronous condition, while this PR only changes `nifi-amqp-processors`.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -95,6 +97,14 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
             .build();
+    public static final PropertyDescriptor MAX_OUTBOUND_MESSAGE_BODY_SIZE = new PropertyDescriptor.Builder()
+            .name("Max Outbound Message Body Size")
```
exceptionfactory: I recommend naming this `Maximum Input FlowFile Size` since that it what is controls.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -95,6 +97,14 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
...
+            .description("Maximum body size of outbound messages. FlowFiles larger than this value are routed to failure without loading content into memory.")
+            .required(true)
+            .defaultValue("64 MB")
```
exceptionfactory: Is there a general published maximum size of AMQP messages, or some reference for using 64 MB is the default value?

exceptionfactory (changes requested):
Thanks for the updates @ing-mattioni. Adding the property looks like a good direction. I provided a comment on the naming, and raised a question about the default value.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
ing-mattioni: Thanks, that makes sense. I renamed the property to `Maximum Input FlowFile Size` to make it clear that the limit applies to the input FlowFile before `PublishAMQP` reads the content into memory.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
ing-mattioni: Good point. I updated the change to avoid treating 64 MB as a general AMQP maximum.

RabbitMQ documents `max_message_size` as configurable, with a current documented default of 128 MiB and a maximum allowed value of 512 MiB. I changed the property to use a 128 MB default and described it as a NiFi-side defensive memory guard rather than a protocol-level AMQP limit. The property description now recommends configuring the value according to broker limits and available JVM memory.

ing-mattioni:
Addressed in the latest update:

- I renamed the property to `Maximum Input FlowFile Size` to clarify that it controls the maximum input FlowFile size before `PublishAMQP` reads the content into memory.

- I also updated the default value and description so the property is described as a NiFi-side defensive memory guard rather than a general AMQP protocol limit. The value should be configured according to broker limits and available JVM memory.

ing-mattioni:
Rebased now with the latest main version after validation and build failed

ing-mattioni:
Conflicts in PublishAMQP class resolved

ing-mattioni:
Rebased now with the latest main version after validation and build failed

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -75,6 +76,7 @@
     @ReadsAttribute(attribute = AbstractAMQPProcessor.AMQP_CLUSTER_ID_ATTRIBUTE, description = "The ID of the AMQP Cluster"),
 })
 public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
+    private static final long MAXIMUM_INPUT_FLOWFILE_SIZE_LIMIT = 128 * 1024 * 1024L;
```
exceptionfactory: Based on the earlier conversation, it sounds like the maximum value should be 512 MB, not 128 MB.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
```diff
@@ -108,6 +110,16 @@ public class PublishAMQP extends AbstractAMQPProcessor<AMQPPublisher> {
             .allowableValues(DeliveryGuarantee.class)
             .defaultValue(DeliveryGuarantee.AT_MOST_ONCE)
             .build();
+    public static final PropertyDescriptor MAXIMUM_INPUT_FLOWFILE_SIZE = new PropertyDescriptor.Builder()
+            .name("Maximum Input FlowFile Size")
+            .description("Maximum size of an input FlowFile that will be read into memory before publishing. PublishAMQP reads FlowFile content into a byte array "
+                    + "before publishing, so FlowFiles larger than this value are routed to failure before content is read. Configure this value according to "
+                    + "broker limits and available JVM memory.")
```
exceptionfactory: A multline string can be used

exceptionfactory (changes requested):
Thanks for your patience @ing-mattioni. The latest version looks close to completion. I noted one more concern regarding the maximum allowed size, and then this looks ready to be merged

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
ing-mattioni: Thanks for pointing this out. I updated the maximum allowed value from 128 MB to 512 MB.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/PublishAMQP.java
ing-mattioni: Thanks for the suggestion. I updated the description to use a multiline text block.

exceptionfactory (changes requested):
@ing-mattioni, please sign all commits on the branch and rebase to address the commit signing requirement.

ing-mattioni:
@exceptionfactory Thanks for pointing this out. I rebased the branch onto the latest main, signed all commits, and force-pushed the updated branch. GitHub now reports both commits as verified.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/PublishAMQPTest.java
```diff
@@ -181,7 +210,42 @@ public void validateFailedPublishAndTransferToFailure() {
...
+        assertEquals("Maximum Input FlowFile Size", PublishAMQP.MAXIMUM_INPUT_FLOWFILE_SIZE.getName());
+        assertEquals("128 MB", PublishAMQP.MAXIMUM_INPUT_FLOWFILE_SIZE.getDefaultValue());
+
+        setConnectionProperties(runner);
+        runner.setProperty(PublishAMQP.MAXIMUM_INPUT_FLOWFILE_SIZE, "128 MB");
+        runner.assertValid();
+
+        runner.setProperty(PublishAMQP.MAXIMUM_INPUT_FLOWFILE_SIZE, "129 MB");
+        runner.assertNotValid();
     }
```
exceptionfactory: This test method is failing, but it should be removed since it is simply evaluating property descriptor values

exceptionfactory (changes requested):
Thanks for addressing the commit signing @ing-mattioni.

There is one failing unit test method after the maximum size change, but that test method should be removed.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/test/java/org/apache/nifi/amqp/processors/PublishAMQPTest.java
ing-mattioni: Removed the property descriptor test method as suggested and confirmed that all remaining PublishAMQPTest tests pass.

exceptionfactory (approved):
Thanks for working through the feedback @ing-mattioni, the latest version looks good.

PR #11243, opened 2026-05-13
NIFI-15935 Clean up AMQP resources after processor failures

ing-mattioni:
# Summary

[NIFI-15935](https://issues.apache.org/jira/browse/NIFI-15935)

This pull request updates `AbstractAMQPProcessor` to clean up AMQP resources more consistently after unexpected processor failures.

When `processResource()` throws an unexpected exception, the processor currently logs the failure and yields the context. This change also rolls back the `ProcessSession` and closes the AMQP resource, ensuring that the client is dropped and recreated instead of potentially remaining in an inconsistent state.

This change also ensures that the AMQP executor service is shut down when resource creation fails after the executor has been created.

## Changes

- Rolls back the `ProcessSession` after unexpected processor failures
- Closes the AMQP resource after unexpected processor failures
- Shuts down the AMQP executor service when resource creation fails
- Updates AMQP client executor thread naming to use a generic client name

# Tracking

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-15935) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

`.\mvnw.cmd -pl :nifi-amqp-processors -am test`

`.\mvnw.cmd -pl :nifi-amqp-processors -am -P contrib-check install -DskipTests`

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
+++ b/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
@@ -285,8 +285,9 @@ abstract class AbstractAMQPProcessor<T extends AMQPWorker> extends AbstractProce
 
     private AMQPResource<T> createResource(final ProcessContext context) {
         Connection connection = null;
+        ExecutorService executor = null;
         try {
-            final ExecutorService executor = Executors.newSingleThreadExecutor(BasicThreadFactory.builder()
+            executor = Executors.newSingleThreadExecutor(BasicThreadFactory.builder()
                     .namingPattern("AMQP Client: " + getIdentifier() + "-%d")
                     .build());
             connection = createConnection(context, executor);
@@ -298,6 +299,14 @@ abstract class AbstractAMQPProcessor<T extends AMQPWorker> extends AbstractProce
                     connection.close();
                 } catch (Exception closingEx) {
                     getLogger().error("Failed to close AMQP Connection", closingEx);
+                }
+            }
+            if (executor != null) {
+                try {
+                    executor.shutdown();
+                } catch (Exception closingEx) {
+                    getLogger().error("Failed to shut down AMQP Executor", closingEx);
+                    e.addSuppressed(closingEx);
                 }
             }
             throw e;
--- a/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
+++ b/nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
@@ -244,8 +244,10 @@ abstract class AbstractAMQPProcessor<T extends AMQPWorker> extends AbstractProce
             context.yield();
             closeResource(resource);
         } catch (Exception e) {
-            getLogger().error("Processor failure", e);
+            getLogger().error("Processor failure, dropping the client", e);
+            session.rollback();
             context.yield();
+            closeResource(resource);
         }
     }
 
@@ -284,8 +286,8 @@ abstract class AbstractAMQPProcessor<T extends AMQPWorker> extends AbstractProce
     private AMQPResource<T> createResource(final ProcessContext context) {
         Connection connection = null;
         try {
-            ExecutorService executor = Executors.newSingleThreadExecutor(BasicThreadFactory.builder()
-                    .namingPattern("AMQP Consumer: " + getIdentifier())
+            final ExecutorService executor = Executors.newSingleThreadExecutor(BasicThreadFactory.builder()
+                    .namingPattern("AMQP Client: " + getIdentifier() + "-%d")
                     .build());
             connection = createConnection(context, executor);
             T worker = createAMQPWorker(context, connection);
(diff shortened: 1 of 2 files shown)


exceptionfactory (changes requested):
Thanks for proposing improvements @ing-mattioni.

The try-catch around `session.rollback()` is not correct and should be removed.

Closing the ExecutorService also appears to be incorrect.

The goal of improving resource handling is good, but this needs to be reconsidered for the broader implications on exception handling.

ing-mattioni:
> Thanks for proposing improvements @ing-mattioni.
> 
> The try-catch around `session.rollback()` is not correct and should be removed.
> 
> Closing the ExecutorService also appears to be incorrect.
> 
> The goal of improving resource handling is good, but this needs to be reconsidered for the broader implications on exception handling.

Updated based on the review. The rollback call is no longer wrapped in a try/catch, and the resource creation failure path no longer closes the ExecutorService directly. The generic exception path still rolls back the session and drops the AMQP client resource.

ing-mattioni:
Thank you @exceptionfactory , I made the following changes:

  - Removed the try/catch wrapper around session.rollback().
  - Removed direct executor.shutdownNow() from the resource-creation failure path.
  - Kept the improved generic failure handling that rolls back the session, yields, and drops the AMQP client resource.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
```diff
@@ -244,8 +244,10 @@ public final void onTrigger(final ProcessContext context, final ProcessSession s
             context.yield();
             closeResource(resource);
         } catch (Exception e) {
-            getLogger().error("Processor failure", e);
+            getLogger().error("Processor failure, dropping the client", e);
+            session.rollback();
             context.yield();
+            closeResource(resource);
         }
```
turcsanyip, line 251: @ing-mattioni The original idea was that all client-related exceptions are wrapped into `AMQP*Exception` in `processResource()` method and the caller needs to close/recreate the resources only for those exception types but not for generic exceptions.

In case of a generic exception the client should be fine. So I'm afraid recreating the client will not help and the rolled back FlowFile will fail again and again causing an infinite loop.

Could you please give an example of a generic error when close + createResource() helps?

turcsanyip: OK, I think I understand what I missed here. When `processResource()` throws a generic exception, the client is still alive but it is no longer available for the processor because `resourceQueue.offer(resource)` will not be called due to the exception.

So closing the client seems to be right way. In this case, `basic.nack` in [NIFI-15892](https://issues.apache.org/jira/browse/NIFI-15892) is not really needed because it just redelivers the message to the client that cannot work anymore and must be closed. The real remedy is that the client gets closed and all locally buffered (unacked) messages will be discarded.

ing-mattioni: @turcsanyip so, if I understood correctly, you want to merge this PR to close both this issue and NIFI-15892, while not merging the PR opened for NIFI-15982 because the basic.nack is not really needed. Is it correct?

turcsanyip: @ing-mattioni Yes, correct. Given that `basic.nack` call does not really contribute to the fix, I would not merge that PR if you and @exceptionfactory agree.

The real fix is to close the client so that all pending (unacknowledged) messages become available for the broker to redeliver to the new client or other existing clients. The negative acknowledgment happens too early, and the messages may be redelivered to the same client that is about to be closed. Closing the client is simply enough and handles both the failed messages (that could be nacked optionally) and the ones in the client's local buffer (unprocessed yet but already fetched from the broker).

ing-mattioni: @turcsanyip agreed

turcsanyip (changes requested):
@ing-mattioni The PR rolls back `ProcessSession` and closes `AMQPResource` after an unexpected processor failure properly, but this point is missing which was listed under "Proposed changes" in the jira ticket:
> Shut down the AMQP executor service when resource creation fails

It makes sense to fix it too and I think it can be implemented only in the `catch` block of `createResource()` because `AMQPResource` object has not been created yet at this point of the execution.

nifi-extension-bundles/nifi-amqp-bundle/nifi-amqp-processors/src/main/java/org/apache/nifi/amqp/processors/AbstractAMQPProcessor.java
turcsanyip: Closed https://issues.apache.org/jira/browse/NIFI-15892 and its PR.

ing-mattioni:
> @ing-mattioni The PR rolls back `ProcessSession` and closes `AMQPResource` after an unexpected processor failure properly, but this point is missing which was listed under "Proposed changes" in the jira ticket:
> 
> > Shut down the AMQP executor service when resource creation fails
> 
> It makes sense to fix it too and I think it can be implemented only in the `catch` block of `createResource()` because `AMQPResource` object has not been created yet at this point of the execution.

Thanks @turcsanyip for pointing this out. I updated createResource() to retain the AMQP executor reference during resource creation and shut it down in the failure path before rethrowing the original exception. This covers the case where AMQPResource has not been constructed yet, so normal AMQPResource.close() ownership has not started.

I also added a focused unit test that simulates worker creation failure and verifies that the connection is closed and the executor is shut down.

turcsanyip (approved):
@ing-mattioni Thanks for adding the executor shutdown logic and the tests.

+1 from my side

@exceptionfactory Do you have any additional comments or can we proceed with the merge?

turcsanyip:
@exceptionfactory I merged this PR. Fell free to reopen the jira if something is missing and you suggest further changes.

exceptionfactory:
Sorry for the delay, thanks for approving and merging @turcsanyip, and thanks again for the work @ing-mattioni.

PR #11245, opened 2026-05-13
NIFI-15938 Improve TailFile startup and state recovery

ing-mattioni:
# Summary

[NIFI-15938](https://issues.apache.org/jira/browse/NIFI-15938)

This pull request updates `TailFile` to improve multi-file startup behavior and make persisted state recovery more robust.

The changes focus on startup performance, rollover state restoration, malformed persisted state handling, read-loop buffer behavior, and buffer-size validation.

## Changes

- Improves multi-file startup by replacing eager file listing with streaming file discovery
- Replaces repeated state membership checks with `HashSet` membership
- Restores persisted `tailingPostRollover` state correctly
- Avoids recovery failures when rollover state exists but no rolled files remain
- Hardens recovery against malformed persisted numeric and checksum state
- Fixes buffer reuse in the read loop
- Tightens buffer-size validation to prevent zero-sized or overflow-sized buffers
- Adds regression tests for the fixed TailFile scenarios

Key fixes, more technical description:

  - Improved multi-file startup by replacing eager FileUtils.listFiles and O(n²) state checks with streaming Files.find and HashSet membership.
  - Fixed persisted tailingPostRollover restoration and avoided crashes when rollover state exists but no rolled files remain.
  - Hardened recovery against malformed persisted numeric/checksum state.
  - Fixed buffer reuse in the read loop and tightened buffer-size validation to prevent zero/overflow-sized buffers.

# Tracking

### Issue Tracking

- [x] [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI-15938) issue created

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

`.\mvnw.cmd --% -pl :nifi-standard-processors -am -Dtest=TestTailFile,TestTailFileSimpleScenarios,TestTailFileGeneratedScenarios -Dsurefire.failIfNoSpecifiedTests=false test`

`.\mvnw.cmd --% -pl :nifi-standard-processors -Pcontrib-check validate`

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
@@ -16,7 +16,6 @@
  */
 package org.apache.nifi.processors.standard;
 
-import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.annotation.behavior.InputRequirement;
 import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
@@ -59,6 +58,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.UncheckedIOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.charset.StandardCharsets;
@@ -70,8 +70,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -81,13 +81,13 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 import java.util.zip.CRC32;
 import java.util.zip.CheckedInputStream;
 import java.util.zip.Checksum;
 
 import static org.apache.nifi.expression.ExpressionLanguageScope.ENVIRONMENT;
 import static org.apache.nifi.expression.ExpressionLanguageScope.NONE;
-import static org.apache.nifi.processor.util.StandardValidators.DATA_SIZE_VALIDATOR;
 import static org.apache.nifi.processor.util.StandardValidators.REGULAR_EXPRESSION_VALIDATOR;
 
 // note: it is important that this Processor is not marked as @SupportsBatching because the session commits must complete before persisting state locally; otherwise, data loss may occur
@@ -260,7 +260,7 @@ public class TailFile extends AbstractProcessor {
             "buffering all data from the tailed file, which can quickly exhaust the heap. To avoid this, the Processor will buffer only up to this amount of data before flushing the buffer, even if" +
             " it means ingesting partial data from the file.")
         .required(true)
-        .addValidator(DATA_SIZE_VALIDATOR)
+        .addValidator(StandardValidators.createDataSizeBoundsValidator(1, Integer.MAX_VALUE))
         .expressionLanguageSupported(NONE)
         .defaultValue("64 KB")
         .dependsOn(LINE_START_PATTERN)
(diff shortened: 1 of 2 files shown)


nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
```diff
@@ -415,6 +415,7 @@ public void recoverState(final ProcessContext context) throws IOException {
 
     private void initStates(final List<String> filesToTail, final Map<String, String> statesMap, final boolean isCleared) {
         int fileIndex = 0;
+        final Set<String> filesToTailSet = new HashSet<>(filesToTail);
```
exceptionfactory:
```suggestion
        final Set<String> uniqueFilesToTail = new HashSet<>(filesToTail);
```

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
```diff
@@ -427,9 +428,19 @@ private void initStates(final List<String> filesToTail, final Map<String, String
...
-                        int index = Integer.parseInt(key.split("\\.")[1]);
-                        states.put(value, new TailFileObject(index, statesMap, preAllocatedBufferSize));
+                    if (key.endsWith(TailFileState.StateKeys.FILENAME) && filesToTailSet.contains(value)) {
+                        final int index;
+                        try {
+                            index = Integer.parseInt(key.split("\\.")[1]);
+                        } catch (final RuntimeException e) {
+                            getLogger().warn("Ignoring malformed TailFile state key {}", key, e);
+                            continue;
+                        }
```
exceptionfactory: Is there a particular reason for introducing this try-catch? The Processor should control the state content, avoiding potential incorrect formatting.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
```diff
@@ -427,9 +428,19 @@ private void initStates(final List<String> filesToTail, final Map<String, String
...
+                        try {
+                            states.put(value, new TailFileObject(index, statesMap, preAllocatedBufferSize));
+                        } catch (final RuntimeException e) {
+                            getLogger().warn("Ignoring malformed TailFile state for {}", value, e);
+                        }
```
exceptionfactory: This catch also seems too broad and not necessary.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
exceptionfactory: What is the reason for creating this wrapping Set?

exceptionfactory (changes requested):
Thanks for proposing these improvements @ing-mattioni.

On initial review, I raised some concerns about the introduction of several try-catch blocks. These seem unnecessary without further explanation.

Some of the other changes also raise implementation questions, so overall, some clarification and refinement is needed before this can go forward.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
ing-mattioni: Applied. I renamed filesToTailSet to uniqueFilesToTail for clarity.

ing-mattioni: The Set is used to avoid repeated linear List.contains() checks while reconciling restored state and removing files that are no longer tailed. With many matching files, keeping this as a List would make startup/state reconciliation trend toward O(n²). The original filesToTail list is still used when adding new states, so ordering is preserved.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
ing-mattioni: Agreed. I removed this try/catch. TailFile controls the persisted state key format, so adding defensive handling for malformed internal state is not needed here.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java
ing-mattioni: Agreed. I removed this broad catch as well. If the processor-controlled state cannot be restored, that should not be silently ignored by skipping the file.

ing-mattioni:
Thank you @exceptionfactory ; I've updated the PR based on the review. The added try/catch blocks have been removed, and the Set variable has been renamed to "uniqueFilesToTail". I kept the Set conversion because it avoids repeated linear membership checks during state reconciliation, which is important when TailFile starts with many matching files.

PR #11246, opened 2026-05-13
NIFI-15937: Adding support for polling in the connector canvas.

mcgilman:
# Summary

[NIFI-15937](https://issues.apache.org/jira/browse/NIFI-15937)

The Connector canvas previously loaded the flow once on mount and only refreshed in response to user-initiated navigation or actions. As a result, server-side changes that happen passively — bulletins emitted by managed components, queue counts as data flows through, processor run-status transitions — did not appear on the canvas until the user navigated away and back. This change introduces a polling loop modeled on the flow designer's reload pattern so the connector canvas reflects server state without manual interaction.

### Behavior

- A new `startConnectorCanvasPolling` action is dispatched from `ConnectorCanvasComponent#ngOnInit`, and a matching `stopConnectorCanvasPolling` is dispatched from `ngOnDestroy`. The polling effect drives a 30 second interval and tears down on the stop action so polling never outlives the canvas.
- A new `reloadConnectorFlow` action sits between the polling source and `loadConnectorFlow`. Its effect throttles to one reload per second, resolves the connector id and process group id from canvas state (rather than the route, which can race a navigation), and skips work if the canvas has not yet completed an initial load. This mirrors the flow designer's `reloadFlow` indirection and keeps the polling source decoupled from which group is currently mounted.
- The polling tick is gated on `DocumentVisibility.Visible` so background tabs do not generate load, and on `loadingStatus !== 'loading'` so a polling reload cannot cancel a user-initiated process group navigation. Without the loading guard the load effect's `switchMap` would replace the in-flight navigation load with a load of the previously-rendered group, leaving the URL and canvas out of sync.
- A document-visibility subscription in the effects constructor triggers an immediate reload when the document becomes visible after having been hidden for longer than the polling cadence (30 s). This avoids displaying obviously stale bulletins / status while waiting for the next interval tick. A `lastReload` timestamp debounces rapid tab switches so they do not each force a fresh load.

### Test coverage

`connector-canvas.effects.spec.ts` and `connector-canvas.component.spec.ts` are extended to cover the new polling lifecycle (start dispatched on init, stop dispatched on destroy, throttle behavior on `reloadConnectorFlow`, document-visibility wake-up paths above and below the 30 s threshold, transitions to hidden), as well as a handful of pre-existing coverage gaps that surfaced while writing the polling tests (`refreshAfterQueueEmptied$` all-queues-emptied path and missing-`connectorId` guard, `viewComponentConfiguration$` entity-without-`operatePermissions` fallback, `navigateToProvenanceForComponent$` funnel back-navigation, `viewProcessorStateAction#canClear` Disabled / Invalid / missing-status branches, `canDrain` / `canCancelDrain` predicate matrix, and `returnToConnectorListing` router navigation).

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/state/connector-canvas/connector-canvas.actions.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/state/connector-canvas/connector-canvas.actions.ts
@@ -65,6 +65,34 @@ export const loadConnectorFlowFailure = createAction(
 
 export const loadConnectorFlowComplete = createAction('[Connector Canvas] Load Connector Flow Complete');
 
+/**
+ * Reload the currently displayed connector flow.
+ *
+ * The polling effect dispatches this action; the corresponding effect throttles
+ * and resolves the connector and process group identifiers from state before
+ * delegating to {@link loadConnectorFlow}. Keeping the indirection makes the
+ * polling source independent of which connector and process group are currently
+ * mounted, mirroring the flow-designer's reloadFlow pattern.
+ */
+export const reloadConnectorFlow = createAction('[Connector Canvas] Reload Connector Flow');
+
+/**
+ * Begin periodic refresh of the connector flow on the currently mounted canvas.
+ *
+ * The component dispatches this action in ngOnInit; the polling effect drives a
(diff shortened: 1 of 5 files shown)


rfellows (approved):
Thanks @mcgilman 👍

PR #11248, opened 2026-05-13
NIFI-15855 - Connector wizard - secret and boolean property handling

rfellows:
# Summary

[NIFI-15855](https://issues.apache.org/jira/browse/NIFI-15855)

Completes the connector wizard property-input behavior for `SECRET` and `BOOLEAN` property descriptors in `connector-property-input`. The `availableSecrets` / `secretsLoading` / `secretsError` inputs were declared but unused in the previous wizard PR; this change wires them through and switches `BOOLEAN` rendering to a slide-toggle to match the rest of the wizard's input controls.

### `SECRET`

- Render `SECRET` properties via `<searchable-select>` instead of a plain text input.
- Form value is the composite key `providerId::providerName::fullyQualifiedName`, built by `buildSecretKey()`. The parent `connector-configuration-step` already unpacks this on save via `parseSecretKey()`.
- Group options by provider name, falling back to `"Provider - Group"` when a provider owns multiple groups.
- Orphan handling: a saved value not present in the loaded secrets renders as a disabled `"…fqn… (no longer available)"` option grouped under the saved provider name. A pre-load placeholder is shown while secrets are still loading.
- Provider-rename auto-fix: when a saved value matches a current secret by `providerId + fullyQualifiedName` but the `providerName` has changed, the form value is rewritten to the new composite key after the next render (`afterNextRender` + `runInInjectionContext`).
- `secretsError` flips the `<searchable-select>` into its `loadError` state with the message `"Failed to load secrets"` and suppresses the description hint.
- Inline progress spinner with `"Loading secrets..."` while `secretsLoading()` is true; placeholder text follows the four `SECRET` states (`Loading secrets...`, `Failed to load secrets`, `No secrets available`, `Select a secret`).

Screenshot 2026-05-13 at 13 10 23

### `BOOLEAN`

- Replace `<mat-checkbox>` with `<mat-slide-toggle>`: label on the left, toggle on the right, description below. `writeValue` already coerces `'true'` / `true` / other to a boolean, so no API change for parent consumers.

Screenshot 2026-05-13 at 13 11 07

## Reviewer notes

- **`BOOLEAN` rendering change.** This is the first use of `<mat-slide-toggle>` for a property descriptor in `nifi-frontend`; the existing slide-toggles in the codebase (e.g. bulletin-board, update-attribute rule listing) are ad-hoc UI mode switches. The toggle uses Material's default dash/check thumb icon for visual consistency with those. 
- **`SECRET` form value contract.** The form value is the composite key, not `secret.id`. The existing parent already depends on this; any future consumer reading the value directly must use `parseSecretKey()` to unpack it.
- No parent or store changes. `availableSecrets` / `secretsLoading` / `secretsError` are already populated upstream and passed in via `connector-configuration-step`.

## Testing

- `connector-property-input.component.spec.ts` extended from 37 to 50 tests.
  - 12 `SECRET` tests: option shape, single-group vs multi-group label format, post-load orphan, pre-load orphan, provider-rename auto-fix, `loadError` wiring (including empty-string short-circuit), inline spinner show/hide, placeholder state transitions.
  - 5 `BOOLEAN` tests: slide-toggle rendering, parent → child form sync, toggle → parent round-trip via native input click, `writeValue('true')` truthy coercion, `writeValue('false')` falsy coercion.
- Full shared lib suite (741 tests) passes; `nx lint shared` and prettier are clean.

### DIFF
--- a/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.html
+++ b/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.html
@@ -18,12 +18,15 @@
 @if (prop) {
     @switch (prop.type) {
         @case ('BOOLEAN') {
-            <mat-checkbox [formControl]="formControl" data-qa="property-input-boolean">
-                <span class="text-sm">{{ prop.name }}</span>
+            <div class="flex flex-col">
+                <div class="flex gap-x-2 items-center">
+                    <span class="text-sm">{{ prop.name }}</span>
+                    <mat-slide-toggle [formControl]="formControl" data-qa="property-input-boolean"></mat-slide-toggle>
+                </div>
                 @if (prop.description) {
                     <span class="text-xs tertiary-color block">{{ prop.description }}</span>
                 }
-            </mat-checkbox>
+            </div>
         }
         @default {
             @if (shouldUseAssetUpload()) {
@@ -79,14 +82,18 @@
                         [searchPlaceholder]="'Search'"
                         [allowClear]="!prop.required"
                         [hint]="prop.description || ''"
-                        [showHint]="!!prop.description"
+                        [showHint]="!!prop.description && !hasSecretsError()"
                         [validationError]="getValidationErrorMessage()"
+                        [loadError]="hasSecretsError()"
+                        [loadErrorMessage]="'Failed to load secrets'"
                         data-qa="property-input-select">
                     </searchable-select>
-                    @if (isDynamicValuesLoading()) {
+                    @if (isDynamicValuesLoading() || isSecretsLoading()) {
                         <div class="flex items-center gap-2 mt-1" data-qa="property-input-loading">
                             <mat-progress-spinner diameter="14" mode="indeterminate"></mat-progress-spinner>
-                            <span class="text-xs tertiary-color">Loading values...</span>
+                            <span class="text-xs tertiary-color">
+                                {{ isSecretsLoading() ? 'Loading secrets...' : 'Loading values...' }}
+                            </span>
                         </div>
                     }
                 </div>
--- a/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.html
+++ b/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.html
@@ -19,10 +19,9 @@
     @switch (prop.type) {
         @case ('BOOLEAN') {
             <div class="flex flex-col">
-                <div class="flex gap-x-2 items-center">
+                <mat-slide-toggle [formControl]="formControl" labelPosition="before" data-qa="property-input-boolean">
                     <span class="text-sm">{{ prop.name }}</span>
-                    <mat-slide-toggle [formControl]="formControl" data-qa="property-input-boolean"></mat-slide-toggle>
-                </div>
+                </mat-slide-toggle>
                 @if (prop.description) {
                     <span class="text-xs tertiary-color block">{{ prop.description }}</span>
                 }
--- a/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.ts
+++ b/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.ts
@@ -142,6 +142,29 @@ export class ConnectorPropertyInput implements ControlValueAccessor, DoCheck, On
             }
             this.lastSeenPropertyName = currentName;
         });
+
+        // Reconcile form value with current secret identity after secrets load.
+        // If the saved composite key matches a current secret by providerId + fullyQualifiedName
+        // but the providerName has changed, rewrite the form value to the current key.
+        // afterNextRender defers the setValue past the current change-detection cycle.
+        // Multiple rapid re-runs are safe: the rewrite is idempotent.
+        effect(() => {
+            const secrets = this.availableSecrets();
+            if (this.property()?.type !== 'SECRET' || !secrets) return;
+            const currentValue = this.formControl.value as string | null;
+            if (!currentValue) return;
+            const parsed = parseSecretKey(currentValue);
+            const matching = secrets.find(
+                (s) => s.providerId === parsed.providerId && s.fullyQualifiedName === parsed.fullyQualifiedName
+            );
+            if (!matching) return;
+            const expected = buildSecretKey(matching.providerId, matching.providerName, matching.fullyQualifiedName);
+            if (currentValue !== expected) {
+                runInInjectionContext(this.injector, () => {
+                    afterNextRender(() => this.formControl.setValue(expected, { emitEvent: true }));
+                });
+            }
+        });
     }
 
     ngOnInit(): void {
@@ -155,6 +178,9 @@ export class ConnectorPropertyInput implements ControlValueAccessor, DoCheck, On
             normalized = value === true || value === 'true';
         }
         this.formControl.setValue(normalized, { emitEvent: false });
+        if (this.property()?.type === 'SECRET') {
+            this.selectOptions = this.computeSelectOptions();
+        }
     }
 
     registerOnChange(fn: (value: unknown) => void): void {
@@ -162,6 +188,9 @@ export class ConnectorPropertyInput implements ControlValueAccessor, DoCheck, On
             this.valueChangesSubscribed = true;
             this.formControl.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((value) => {
                 fn(value);
+                if (this.property()?.type === 'SECRET') {
+                    this.selectOptions = this.computeSelectOptions();
+                }
             });
         }
     }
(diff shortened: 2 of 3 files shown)


mcgilman:
Reviewing...

nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.html
```diff
@@ -18,12 +18,15 @@
...
+            <div class="flex flex-col">
+                <div class="flex gap-x-2 items-center">
+                    <span class="text-sm">{{ prop.name }}</span>
+                    <mat-slide-toggle [formControl]="formControl" data-qa="property-input-boolean"></mat-slide-toggle>
+                </div>
                 @if (prop.description) {
                     <span class="text-xs tertiary-color block">{{ prop.description }}</span>
                 }
-            </mat-checkbox>
+            </div>
```
mcgilman, line 28: **#1 — Accessibility regression: the new `<mat-slide-toggle>` is unlabeled.**

The previous `<mat-checkbox>` wrapped the property name as projected content, which Material associates with the inner control via `aria-labelledby`. The new `<mat-slide-toggle>` is empty — the property name lives in a sibling `<span>`, with no `aria-label`, `aria-labelledby`, or projected content. A screen reader will announce this toggle as unlabeled. Other slide-toggle usages in this codebase (e.g. `bulletin-board.component.html`) project the label inside the toggle, which is the established convention.

Could we project the label inside the toggle? That restores the screen-reader association the previous checkbox provided, and as a bonus the whole label becomes a click target:

```html
<div class="flex flex-col">
    <mat-slide-toggle
        [formControl]="formControl"
        labelPosition="before"
        data-qa="property-input-boolean">
        <span class="text-sm">{{ prop.name }}</span>
    </mat-slide-toggle>
    @if (prop.description) {
        <span class="text-xs tertiary-color block">{{ prop.description }}</span>
    }
</div>
```

nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.ts
```diff
@@ -339,10 +364,45 @@ export class ConnectorPropertyInput implements ControlValueAccessor, DoCheck, On
         return this.property()?.type === 'STRING_LIST' && this.shouldUseSelect();
     }
 
+    /**
+     * Whether the property should be rendered as a slide toggle (BOOLEAN).
+     */
+    shouldUseToggle(): boolean {
+        return this.property()?.type === 'BOOLEAN';
+    }
```
mcgilman: **#2 — `shouldUseToggle()` appears to be dead code.**

I can't find a caller for this method — the template uses `@switch (prop.type) { @case ('BOOLEAN') }` and no test references it. Could we either remove it, or switch the template over to use it for symmetry with the other `shouldUseX()` predicates in the file? Whichever is preferred, having both an unused predicate and a switch-case is a maintenance trap.

nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.ts
```diff
@@ -426,6 +499,71 @@ export class ConnectorPropertyInput implements ControlValueAccessor, DoCheck, On
...
+                    matching.providerName,
+                    matching.fullyQualifiedName
+                );
+                if (currentValue !== expected) {
+                    // Provider name changed since saving - update form value to match current secret.
+                    // Use afterNextRender to safely modify form value after change detection completes.
+                    runInInjectionContext(this.injector, () => {
+                        afterNextRender(() => this.formControl.setValue(expected, { emitEvent: true }));
+                    });
+                }
```
mcgilman: **#3 — Form-value mutation as a side effect inside `computeSecretSelectOptions()`.**

A method named `computeSecretSelectOptions` mutating the form control is surprising and makes the function harder to reason about. There's also a subtle risk: every effect re-run that observes `currentValue !== expected` schedules another `afterNextRender` callback. If the effect re-runs more than once before the first callback flushes (e.g. parent re-emits `availableSecrets` twice quickly), multiple identical `setValue` calls will fire one after another, each with `emitEvent: true`.

Could we lift this auto-fix into its own `effect()` (or a small `reconcileFormValueWithSecrets()` method) so option-compute stays pure, and guard with a single-shot flag that resets on property-name change? Something like:

```ts
private autoFixScheduled = false;

// In the constructor, alongside the existing effect:
effect(() => {
    const secrets = this.availableSecrets();
    if (this.property()?.type !== 'SECRET' || !secrets) return;
    if (this.autoFixScheduled) return;
    // ...derive expected, compare to formControl.value, schedule once...
});
```

nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.ts
```diff
@@ -114,6 +130,9 @@ export class ConnectorPropertyInput implements ControlValueAccessor, DoCheck, On
         effect(() => {
             const prop = this.property();
             this.dynamicAllowableValuesState();
+            this.availableSecrets();
+            this.secretsLoading();
+            this.secretsError();
```
mcgilman, line 135: **#4 — `selectOptions` does not recompute when `formControl.value` changes.**

The effect reads `property`, `dynamicAllowableValuesState`, `availableSecrets`, `secretsLoading`, and `secretsError`, but `formControl.value` isn't a signal and isn't tracked here. `computeSecretSelectOptions` reads `formControl.value` to decide whether to append a disabled orphan entry.

Consequence: if the user opens the dropdown with an orphan saved value and then selects a real secret, the disabled `"<fqn> (no longer available)"` entry lingers in `selectOptions` until the parent next re-emits one of the watched signals. The displayed selection is correct (searchable-select renders it from the value), but the stale orphan in the list is confusing.

Could we subscribe to `formControl.valueChanges` (with `takeUntilDestroyed(this.destroyRef)`) and recompute `selectOptions` on each emission? Or mirror the value into a signal at write time so the existing effect picks it up.

nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.ts
```diff
@@ -426,6 +499,71 @@ export class ConnectorPropertyInput implements ControlValueAccessor, DoCheck, On
...
+                });
+            }
+        } else if (currentValue) {
+            const parsed = parseSecretKey(currentValue);
+            options.push({
+                value: currentValue,
+                label: `${parsed.fullyQualifiedName} (no longer available)`,
+                disabled: true,
+                group: parsed.providerName
+            });
```
mcgilman, line 572: **#5 — Pre-load orphan label is misleading.**

While `availableSecrets() === null` and `secretsLoading() === true`, the saved value is rendered as `"<fqn> (no longer available)"` on a disabled select. The disable state stops the user from acting on it, but the wording is alarming during what is just a transient load — and if the load eventually returns the secret, the user briefly saw it labeled as missing for no reason.

Could we only append the `(no longer available)` suffix once the load has actually completed (`secretsLoading() === false && secretsError() === null`)? While loading, render the FQN without the suffix. The pre-load orphan test would need an updated expectation.

nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.ts
```diff
@@ -402,15 +462,28 @@ export class ConnectorPropertyInput implements ControlValueAccessor, DoCheck, On
...
+     *
+     * - SECRET: options come from availableSecrets, valued by a composite key
+     *   (providerId::providerName::fullyQualifiedName) and grouped by provider
+     *   (or "Provider - Group" when a provider owns multiple groups). Saved
+     *   values that no longer match an available secret are surfaced as disabled
+     *   "(no longer available)" options. When the saved value still matches but
+     *   the provider name has changed, the form value is rewritten to the new
+     *   composite key after the next render.
+     * - Otherwise: dynamic values when successfully fetched, falling back to
+     *   the descriptor's static allowableValues.
```
mcgilman: **#8 — The expanded doc comment could also note the side-effect channel and the untracked `formControl.value` dependency.**

The rewrite is a real improvement — much clearer than the previous two-liner. While it now explicitly mentions the form-value rewrite, I think it's worth also calling out:

- that **this is a side effect** scheduled from a method named "compute" (i.e. callers shouldn't assume option-compute is pure — see related comment on lines 539–545), and
- that `formControl.value` is read here but is **not** tracked by the `effect()` that triggers recompute (see related comment on lines 133–135), so options can briefly fall out of sync with the form value until one of the tracked signals re-emits.

Both are non-obvious from the method body and tend to trip up future readers.

mcgilman (changes requested):
Six findings on the new SECRET and BOOLEAN handling in `connector-property-input` — one accessibility regression (#1) plus a handful of correctness / clarity items in the SECRET path. Details inline.

Tests, lint, and prettier all came back clean on my end (50/50 tests pass); the structural changes look right and these are mostly tightening passes. The numbered prefixes (`#1` … `#8`) are just to make the comments easy to refer to in follow-up discussion.

mcgilman (approved):

PR #11250, opened 2026-05-13
NIFI-15940: Ensure that when a ConnectorConfigurationProviderExceptio…

mcgilman:
…n is thrown during a web request it doesn't fall through to the ThrowableMapper.

# Summary

[NIFI-15940](https://issues.apache.org/jira/browse/NIFI-15940)

Adds a dedicated JAX-RS `ExceptionMapper` for `ConnectorConfigurationProviderException` so failures originating from a `ConnectorConfigurationProvider` no longer fall through to the generic `ThrowableMapper`.

Previously, any exception raised by a provider implementation (during configuration load, save, discard, delete, asset management, etc.) was caught by `ThrowableMapper` and translated into a `500 Internal Server Error` with a canned "An unexpected error has occurred…" body. This obscured the provider's own diagnostic message and prevented clients from surfacing the underlying cause to the user.

With this change, `ConnectorConfigurationProviderException` is mapped to `500 Internal Server Error` with the provider-supplied message returned in the `text/plain` response body. The framework cannot classify the failure further without coupling to a specific provider implementation, so a generic 5xx status is appropriate while still preserving the original message for callers to display. This mirrors the precedent set by `AdministrationExceptionMapper`, which handles analogous failures from pluggable framework subsystems.

### Changes

- New `ConnectorConfigurationProviderExceptionMapper` in `nifi-web-api`, registered in `NiFiWebApiResourceConfig` alongside the other framework exception mappers.
- The mapper logs the exception at `WARN` and returns the exception message verbatim in the response body so clients can render it directly.


kevdoran (approved):
+1, thanks @mcgilman

PR #11252, opened 2026-05-14
NIFI-15930: Re-init Secrets from a Provider with Controller Service

bobpaulin:
* Wait for Parameter Provider Controller Services to enable they are valid prior to secret resolution
* Enable Working Context to apply any changes in Connector Property Values and Resolved values to the Parameter Context of the Working Flow Context
* Warn when a parameter provider is not found

# Summary

[NIFI-15930](https://issues.apache.org/jira/browse/NIFI-15930)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizer.java
@@ -1080,71 +1080,69 @@ public class VersionedFlowSynchronizer implements FlowSynchronizer {
     }
 
     private void inheritControllerServices(final FlowController controller, final VersionedDataflow dataflow, final AffectedComponentSet affectedComponentSet) {
+        final FlowManager flowManager = controller.getFlowManager();
+
+        final Set<ControllerServiceNode> toEnable = new HashSet<>();
+        final Set<ControllerServiceNode> toDisable = new HashSet<>();
+
+        // We need to add any Controller Services that are not yet part of the flow. We must then
+        // update the Controller Services to match what is proposed. Finally, we can enable the services.
+        // We have to do this in 3 parts because if we just configure the Controller Service as we add it,
+        // we will have a situation where Service A references Service B. And if Service A is added first,
+        // Service B's references won't be updated. To avoid this, we create them all first, and then configure/update
+        // them so that when AbstractComponentNode#setProperty is called, it properly establishes that reference.
         final List<VersionedControllerService> controllerServices = dataflow.getControllerServices();
-        if (controllerServices != null && !controllerServices.isEmpty()) {
-            final FlowManager flowManager = controller.getFlowManager();
-
-            final Set<ControllerServiceNode> toEnable = new HashSet<>();
-            final Set<ControllerServiceNode> toDisable = new HashSet<>();
-
-            // We need to add any Controller Services that are not yet part of the flow. We must then
-            // update the Controller Services to match what is proposed. Finally, we can enable the services.
-            // We have to do this in 3 parts because if we just configure the Controller Service as we add it,
-            // we will have a situation where Service A references Service B. And if Service A is added first,
-            // Service B's references won't be updated. To avoid this, we create them all first, and then configure/update
-            // them so that when AbstractComponentNode#setProperty is called, it properly establishes that reference.
-            final Map<ControllerServiceNode, Map<String, String>> controllerServicesAddedAndProperties = new HashMap<>();
-            for (final VersionedControllerService versionedControllerService : controllerServices) {
-                final ControllerServiceNode serviceNode = flowManager.getRootControllerService(versionedControllerService.getInstanceIdentifier());
-                if (serviceNode == null) {
-                    final ControllerServiceNode added = addRootControllerService(controller, versionedControllerService);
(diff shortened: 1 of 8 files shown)


kevdoran (commented):
Thanks for this @bobpaulin - here are a few things worth mentioning, but overall the fixes in this PR are good.

## Should fix

### 1. `lastLoggedStatus` is a slow leak

`ParameterProviderSecretsManager.lastLoggedStatus` is only cleaned up when a provider transitions back to `VALID`. If a provider is created, goes `INVALID`, and is then deleted before ever becoming `VALID`, its entry remains in the map for the lifetime of the manager. Over many flow synchronizations this is unbounded.

```52:57:nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/components/connector/secrets/ParameterProviderSecretsManager.java
// Per-ParameterProvider-id deduplication for the WARN log emitted whenever a SecretReference
// resolution is skipped because its backing provider is not VALID. An entry is present for any
// provider id we have already surfaced via WARN; the value is the status that was logged so the
// same WARN is not repeated until the status either changes or the provider returns to VALID.
private final Map<String, ValidationStatus> lastLoggedStatus = new ConcurrentHashMap<>();
```

Easiest fix: at the start of `getSecretProviders()`, retain only IDs currently known to `flowManager.getAllParameterProviders()`:

```java
final Set<String> currentProviderIds = new HashSet<>();
for (final ParameterProviderNode node : flowManager.getAllParameterProviders()) {
    currentProviderIds.add(node.getIdentifier());
}
lastLoggedStatus.keySet().retainAll(currentProviderIds);
```

## Consider

### 2. Move the `initializationContext == null` early-return above the "recreated" log

`recreateWorkingFlowContext` emits the "Working Flow Context has been recreated" info log even when there is no flow to refresh. Move the early-return ahead of the log so the log only fires when a refresh actually follows.

```871:894:nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
private void recreateWorkingFlowContext() {
    destroyWorkingContext();
    workingFlowContext = flowContextFactory.createWorkingFlowContext(...);
    getComponentLog().info("Working Flow Context has been recreated");

    if (initializationContext == null) {
        return;
    }
    final ConnectorConfiguration config = workingFlowContext.getConfigurationContext().toConnectorConfiguration();
    for (final NamedStepConfiguration stepConfig : config.getNamedStepConfigurations()) {
        ...
    }
}
```

### 3. Also log an `INFO` when a provider returns to `VALID`

Currently `lastLoggedStatus.remove(providerId)` runs silently when a provider becomes `VALID` again. Operators searching the log for the `Skipping Parameter Provider … because` warning have no nearby recovery line. Emitting an `INFO` at the recovery point would make the WARN/recovery pairs easy to correlate.

bobpaulin:
Thank you for the review @kevdoran .  I've addressed these feedback items.  I went with a different approach on the logging since the intent of the message is surface when the working context has been set which includes the startup.  This has been useful in debugging in past sessions. I added another message to detail when the logging has refreshed the config.

kevdoran (approved):
Thanks for those changes, LGTM, will get this merged

exceptionfactory (changes requested):
On a quick review, I'm concerned about the use of reflection for verifying `lastWarnedStatus` in the test class. Are there other options for changing this approach?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizer.java
```diff
@@ -1083,69 +1080,71 @@ private void inheritConnectors(final FlowController flowController, final Versio
...
-        // Service B's references won't be updated. To avoid this, we create them all first, and then configure/update
-        // them so that when AbstractComponentNode#setProperty is called, it properly establishes that reference.
         final List<VersionedControllerService> controllerServices = dataflow.getControllerServices();
-        final Map<ControllerServiceNode, Map<String, String>> controllerServicesAddedAndProperties = new HashMap<>();
-        for (final VersionedControllerService versionedControllerService : controllerServices) {
-            final ControllerServiceNode serviceNode = flowManager.getRootControllerService(versionedControllerService.getInstanceIdentifier());
-            if (serviceNode == null) {
-                final ControllerServiceNode added = addRootControllerService(controller, versionedControllerService);
-                controllerServicesAddedAndProperties.put(added, versionedControllerService.getProperties());
+        if (controllerServices != null && !controllerServices.isEmpty()) {
```
exceptionfactory: Have you observed the `controllerServices` return as `null`? If not, I recommend reverting this check to minimize and clarify the scope of changes to this particular framework class.

exceptionfactory (approved):
Thanks for streamlining the test and reverting the defensive change to `VersionedFlowSynchronizer` @bobpaulin. I noticed some of the log messages are more verbose than needed, but those can be addressed in a follow-on issue. The current version looks good, and I will plan to merge on successful build completion.

Thanks for the review @kevdoran!

bobpaulin:
Thank you for the review @kevdoran and @exceptionfactory .   Created https://issues.apache.org/jira/browse/NIFI-15943 to track logging improvements identified.

PR #11254, opened 2026-05-14
NIFI-15815 connector details page

scottyaslan:
# Summary

[NIFI-15815](https://issues.apache.org/jira/browse/NIFI-15815)

Screenshot 2026-05-14 at 10 44 03 AM

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-detail/connector-detail.component.html
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-detail/connector-detail.component.html
@@ -1,0 +1,66 @@
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<div class="flex flex-col h-full">
+    <header class="nifi-header">
+        <navigation></navigation>
+    </header>
+    @if (!backNavigation() && !loading && !errorMessage) {
+        <div class="pl-5 pt-2">
+            <a (click)="returnToConnectorListing()" data-qa="fallback-back-link">
+                <i class="fa fa-arrow-left mr-2" aria-hidden="true"></i>Installed connectors
+            </a>
+        </div>
+    }
+    <context-error-banner [context]="ErrorContextKey.CONNECTORS"></context-error-banner>
+    <div class="flex flex-1 min-h-0 overflow-y-auto">
+        @if (loading) {
+            <div class="flex-1 flex items-center justify-center">
+                <span *nifiSpinner="loading" data-qa="loading-spinner"></span>
+            </div>
+        } @else if (errorMessage) {
+            <div class="flex-1 flex items-center justify-center" data-qa="error-container">
+                <div class="flex flex-col items-center gap-y-4">
+                    <i class="fa fa-warning error-color text-4xl" data-qa="error-icon" aria-hidden="true"></i>
+                    <div data-qa="error-message">{{ errorMessage }}</div>
+                    <button mat-flat-button (click)="returnToConnectorListing()" data-qa="return-button">
+                        Return to Connectors
+                    </button>
+                </div>
+            </div>
+        } @else if (frameSource) {
+            <!--
+                Sandbox flags:
+                  allow-scripts        — run the embedded connector app
+                  allow-same-origin    — postMessage, style injection, API calls
+                  allow-popups         — open external doc links via target="_blank"
+                  allow-popups-to-escape-sandbox — opened tabs run without sandbox restrictions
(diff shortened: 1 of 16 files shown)


mcgilman:
Reviewing...

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-detail/connector-detail.component.ts
```diff
@@ -0,0 +1,133 @@
...
+            });
+    }
+
+    private getFrameSource(detailsUrl: string): SafeResourceUrl | null {
+        const connectorId = this.connector?.id;
+        const urlWithParams = connectorId ? `${detailsUrl}?connectorId=${connectorId}` : detailsUrl;
+
+        const sanitizedUrl = this.domSanitizer.sanitize(SecurityContext.URL, urlWithParams);
+
+        if (sanitizedUrl) {
```
mcgilman: `DomSanitizer.sanitize(SecurityContext.URL, …)` doesn't return `null` for unsafe schemes — it returns the original string with an `'unsafe:'` prefix. So for `'javascript:alert(1)'` it returns the truthy string `'unsafe:javascript:alert(1)'`, `if (sanitizedUrl)` passes, and we hand a `SafeResourceUrl` containing that string to `[src]`. The page is safe today only because no browser recognizes `unsafe:` as a scheme — not because this method rejected anything.

One-liner fix — either prefix-check:

```ts
if (!sanitizedUrl || sanitizedUrl.startsWith('unsafe:')) {
    return null;
}
return this.domSanitizer.bypassSecurityTrustResourceUrl(sanitizedUrl);
```

…or scheme-validate via `new URL(...).protocol` against `http:`/`https:`, which is more robust against future Angular sanitizer changes.

Same drift exists in `ConnectorConfigure.getFrameSource` — worth fixing in both files so we don't keep duplicating a misleading helper. If you take this change, the `'should mark invalid URLs as unsafe'` spec becomes `expect(result).toBeNull()`.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-detail/connector-detail.component.html
```diff
@@ -0,0 +1,66 @@
...
+    <div class="flex flex-1 min-h-0 overflow-y-auto">
+        @if (loading) {
+            <div class="flex-1 flex items-center justify-center">
+                <span *nifiSpinner="loading" data-qa="loading-spinner"></span>
+            </div>
+        } @else if (errorMessage) {
+            <div class="flex-1 flex items-center justify-center" data-qa="error-container">
+                <div class="flex flex-col items-center gap-y-4">
+                    <i class="fa fa-warning error-color text-4xl" data-qa="error-icon" aria-hidden="true"></i>
+                    <div data-qa="error-message">{{ errorMessage }}</div>
```
mcgilman, line 39: Minor consistency: `ConnectorConfigure` uses `<p class="text-base" data-qa="error-message">{{ errorMessage }}</p>` for the same role. Without the `<p>` + `text-base`, the error text picks up the surrounding default typography instead of Material body sizing. Suggest:

```html
<p class="text-base" data-qa="error-message">{{ errorMessage }}</p>
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-detail/connector-detail.component.ts
```diff
@@ -0,0 +1,133 @@
...
+    protected readonly ErrorContextKey = ErrorContextKey;
+
+    ngOnInit(): void {
+        this.store
+            .select(selectConnectorIdFromRoute)
+            .pipe(
+                filter((connectorId) => connectorId != null),
+                switchMap((connectorId) => {
+                    this.loading = true;
+                    this.frameSource = null;
```
mcgilman, line 78: Nit, not a bug: this is safe today because `startListening()` internally calls `stopListening()` and everything is anchored to `DestroyRef`. The one edge case is navigating from an iframe connector to a non-iframe connector — the previous subscription stays bound until the component is destroyed (harmless, since origin/source checks fail against a detached `contentWindow`).

A single `this.connectorMessageHost.stopListening()` here would make intent obvious. Optional.

mcgilman (commented):
Nice, clean addition — the split into `ConnectorDetail` + `ConnectorDetailsContent` is the right shape, sandbox flags are correctly tightened for a read-only view, and the spec coverage on the visibility branches is thorough.

A few comments below; the only one with real substance is the `getFrameSource` sanitize-then-bypass pattern, which is duplicated from `ConnectorConfigure` and worth fixing in both. The rest are small consistency/optional nits.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-detail/connector-detail.component.html
scottyaslan: This is exactly equivalent typogaphically. We do not typically use `<p class="text-base"></p>` elsewhere in the codebase.

mcgilman: It's used in these places:

1 - https://github.com/apache/nifi/blob/main/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-configure/connector-configure.component.html#L34
2 - https://github.com/apache/nifi/blob/main/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-wizard/connector-wizard.component.html#L32
3 - https://github.com/apache/nifi/blob/main/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-wizard/connector-wizard.component.html#L115
4 - https://github.com/apache/nifi/blob/main/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-wizard/connector-configuration-step/connector-configuration-step.component.html#L55

mcgilman (approved):

PR #11262, opened 2026-05-19
NIFI-15951 modularize customized angular material themed components

scottyaslan:
# Summary

[NIFI-15951](https://issues.apache.org/jira/browse/NIFI-15951)

## Summary

Decomposes the monolithic `material.scss` and `_app.scss` theme files into 22 per-component SCSS
partials under `themes/components/`. Each partial exposes a `generate-material-theme()` mixin that
is called from the bottom of `material.scss`.

## Intentional Scope Changes

The following styles were intentionally narrowed or removed as part of this refactoring:

- **`mat.select-overrides()`** (trigger text size, line-height, tracking): Removed from global
  scope. These overrides were tightly coupled to the searchable-overlay pattern and are no longer
  applied globally to all `mat-select` instances.

- **`.mat-mdc-select-panel`** (margin-top, border, border-radius): Removed from global scope.
  Select panel border styling is now scoped to `.searchable-overlay` in `_searchable-overlay.scss`.

- **`.mat-mdc-option`** general styling (min-height, padding, font-weight, `.fa` icon sizing):
  Narrowed to `.searchable-overlay` scope. The typography-related option styles (font-size,
  letter-spacing, line-height) remain global within `.mat-typography.text-base` in
  `_typography.scss`.

- **`.mat-mdc-form-field-icon-prefix .fa`** and **form-field padding** (`.mdc-text-field--outlined`,
  `.mat-mdc-form-field-has-icon-prefix`): Narrowed from global scope to `.searchable-overlay` and
  `searchable-select.component.scss` respectively, as these paddings were only relevant to
  searchable components.

## Bug Fix

- **`--themed-reusable-text-primary`**: The original `_app.scss` referenced this CSS custom property
  for `.mat-mdc-form-field-icon-prefix .fa` color, but this token does
  not exist in NiFi. Corrected to `var(--mat-app-text-color)` in `_searchable-overlay.scss`.

## Purple Theme Simplification

The purple theme previously duplicated component overrides (snackbar, checkbox, spinner, icon-button,
error-button) that were identical to the default theme. These have been removed. The purple theme now
only imports `dialog` as a demonstration of how secondary themes can selectively reuse NiFi's
component partials. Documentation comments explain the three available customization strategies.

## Other Notes

- `.mdc-dialog__content` font-size intentionally changed from hardcoded `14px` to `var(--mat-sys-body-medium-size)`
  to align with the typography scale. Currently resolves to the same value.
- `hover:cursor-pointer` added directly to copy-button template via Tailwind class, replacing a
  global `.copy-button:hover` rule.

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


rfellows:
reviewing...

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_status-variant.scss
```diff
@@ -0,0 +1,116 @@
...
+                }
+
+                &.info {
+                    color: var(--nf-success-contrast);
+                }
+            }
+        }
+    }
+
+    .darkMode {
```
rfellows: The `.darkMode { ... }` block on lines 69–115 is byte-for-byte identical to the `:root { ... }` block on lines 21–67 — every value uses `--mat-sys-*` or `--nf-*` custom properties, all of which are already rebound under `.darkMode` at the `material.scss` level. Duplicating the rules under a `.darkMode` selector emits the same CSS twice and doesn't change any computed values.

NiFi's mode-aware values are reassigned at the variable level inside `material.scss`, so any rule that uses one of those custom properties is automatically mode-aware.

The same redundancy exists in:
- `_drag-and-drop.scss` (lines 84–106)
- `_option.scss` (lines 28–32)
- `_paginator.scss` (lines 27–31)
- `_progress-bar.scss` (lines 47–71)
- `_searchable-overlay.scss` (lines 93–107)
- `_skeleton.scss` (lines 27–31)
- `_tree.scss` (lines 28–32)
- `_typography.scss` (lines 134–140)

Suggest dropping the `.darkMode { ... }` block from each of these partials. Files where the dark-mode block is genuinely needed (different raw palette indices per mode) and should stay: `_button.scss`, `_checkbox.scss`, `_progress-spinner.scss`, `_snackbar.scss`.

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_tree.scss
```diff
@@ -0,0 +1,33 @@
...
+
+@mixin generate-material-theme() {
+    :root {
+        .mat-tree {
+            display: inline !important;
+            background-color: unset !important;
+        }
+    }
+
+    .darkMode {
```
rfellows: Light/dark asymmetry: the `:root` block sets `display: inline !important;` (line 23) but the `.darkMode` block omits it — so a `mat-tree` rendered with a `.darkMode` ancestor loses the `display: inline` override.

Either restore symmetry, or (preferred, see the broader `.darkMode` redundancy comment on `_status-variant.scss`) drop the `.darkMode { ... }` block entirely since both `:root` rules use no color tokens that vary by mode.

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_date-picker.scss
```diff
@@ -0,0 +1,33 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+@use '@angular/material' as mat;
+
+@mixin generate-material-theme() {
+    :root {
+        mat-datepicker-toggle .mat-mdc-icon-button.mat-mdc-button-base.mdc-icon-button {
+            padding: 11px;
```
rfellows: The original `_app.scss` had an inline comment explaining this magic number: `padding: 11px; //center the mat-icon-button inside the date picker form field`. That comment was dropped during the move — please restore it so `11px` doesn't read as an unexplained magic number.

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_option.scss
```diff
@@ -0,0 +1,33 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+@use '@angular/material' as mat;
+
+@mixin generate-material-theme() {
+    :root {
+        .mat-mdc-option[disabled] {
```
rfellows: General observation about the `.mat-mdc-option` narrowing called out in the PR description.

The PR description lists narrowing `.mat-mdc-option`'s general sizing/padding (`min-height`, `padding`, font-weight, `.fa` icon sizing) from global scope to `.searchable-overlay` as an Intentional Scope Change, on the basis that the previously-global rules were "tightly coupled to the searchable-overlay pattern." That assumption appears to be incorrect empirically — testing the resulting build, the two `<mat-select>` dropdowns on the **Scheduling tab of the Processor Configuration dialog** now render with visibly different option padding/sizing than before. They aren't searchable selects and don't carry the `.searchable-overlay` class wrapper, so they no longer pick up the `min-height: 32px; padding: 8px 12px; ...` rules that were globally applied before.

Almost certainly there are other surfaces in the same situation — any non-searchable `<mat-select>` / `<mat-autocomplete>` panel in the controller-services view, parameter-contexts dialog, processor properties drawer, etc. The same concern probably applies to the `.mat-mdc-form-field-icon-prefix .fa` and outlined-form-field padding that were narrowed for the same reason; an audit of `<mat-select>` usages outside `.searchable-overlay` is worth doing before merge.

Two ways to fix without giving up the modularization win:
1. Re-add the general `.mat-mdc-option { min-height: 32px; padding: 8px 12px; font-weight: 400; ... }` rules to this partial's `:root` block (alongside the existing `[disabled]` rule), and let `.searchable-overlay` further narrow as needed — most of its current values are the same anyway, so the override would mostly just be reinstating the global baseline.
2. Keep `_searchable-overlay.scss` as the only owner of those rules and explicitly walk the app to apply the `.searchable-overlay` class wrapper to every panel that should pick them up — likely a much larger and riskier change than (1).

(1) seems lower-risk for a refactor PR and matches the spirit of "behavior-preserving modularization." Recommend doing it before merge so we don't ship a visual regression.

Before PR:

Image

After:

Image

rfellows (changes requested):
A few things...

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_tree.scss
scottyaslan: Ok so for a large portion of this effort I was not trying to change anything. If you look at the _app.scss `styles()` mixin and see where it is used the styles.scss files call it one time in the base scope. This should be scoped equivalently the same now as it were prior to this PR. However, with that being said I think your suggestion is a good one.

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_option.scss
scottyaslan: It is unfortunate but the PR that introduced the searchable-select and mult-select-option components fundamentally changed all mat-select options throughout the nifi design system. This we erroneously merged as the mult-select-options are just mat-options and should inherit from the same spec as mat-option. What we need to do is to restore the mat-options to their correct behavior (mostly default angular material) and instead update the searchable-select and mult-select-option components to accordingly inherit from those.

scottyaslan: Here's a summary of the changes implemented for searchable-select and multi-select-option component theming inheritance issues":

Root cause fixed: MultiSelectOption extends MatOption but Angular Material only injects MatOption's compiled CSS when a real <mat-option> is rendered. Since only the subclass was ever used directly, hover/focus/selected state styles were never applied.

Changes made:

multi-select-option.component.ts — Added mat-mdc-option mdc-list-item to the host class binding so Angular Material's CSS selectors correctly target the element.

multi-select-option.component.scss — Replaced a complex scoped approach (using mat.option-overrides() on multi-select-option tag and & self-selectors) with globally-scoped rules on .mat-mdc-option. This replicates the missing Angular Material state-layer CSS (hover, focus, active, selected backgrounds) and base layout properties (min-height, padding, cursor), relying on ViewEncapsulation.None to inject them globally.

_option.scss — Added mat.option-overrides() for selected state colors and the .fa icon / text truncation rules (previously buried inside _searchable-overlay.scss).

_searchable-overlay.scss — Removed the .mat-mdc-option block that duplicated option styling (moved to _option.scss) and removed the icon prefix .fa rule (moved to _form-field.scss). Also removed the dead var(--mat-app-text-color) reference.

_form-field.scss — Added the .mat-mdc-form-field-icon-prefix .fa sizing rule (moved from _searchable-overlay.scss).

_select.scss — New file. Added cursor and disabled-color rules for selected/hovered options for all mat-options and multi-select-options

nifi-frontend/src/main/frontend/libs/shared/src/components/multi-select-option/multi-select-option.component.scss
```diff
@@ -15,106 +15,59 @@
...
-    &:hover .mat-pseudo-checkbox,
-    &:focus .mat-pseudo-checkbox,
-    &.mat-mdc-option-active .mat-pseudo-checkbox {
-        --mat-pseudo-checkbox-minimal-selected-checkmark-color: var(
-            --mat-option-label-text-color,
-            var(--mat-sys-on-surface)
-        );
-    }
+.mat-mdc-option-active .mat-focus-indicator::before {
+    content: '';
```
rfellows: With `ViewEncapsulation.None`, these rules are global the moment the class is imported anywhere — but they're attached to one component's lifecycle. "applied whenever MultiSelectOption is used" is misleading, because the resulting CSS targets every `.mat-mdc-option` in the app, not just `MultiSelectOption`.

Recommend: move the entire `.mat-mdc-option { ... }`/`:hover`/`:focus`/`--selected` block from `multi-select-option.component.scss` into either `_option.scss` or `_select.scss` and let the partial own all global option styling. Leave `multi-select-option.component.scss` with only the .`multi-select-option` and `.mat-pseudo-checkbox-*` rules that are actually component-specific. That gives a consistent rule: global Material overrides live in `themes/components/*.scss;` component scss files are component-scoped.

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_option.scss
rfellows, line 1: The new file naming makes it hard to predict where a given rule lives.
`_option.scss` carries `mat.option-overrides()`, primary-text truncation, and `.fa` icon sizing; 
`_select.scss` carries cursor rules for `.mat-mdc-option.mdc-list-item--selected`. 

Both files are styling .`mat-mdc-option`. Recommend collapsing `_option.scss` into `_select.scss` so it covers the select panel, options, disabled-list-items, and active/selected state-layer rules in one place

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_status-variant.scss
rfellows: While i didn't list this file in the bulleted list, the comment was here so I assumed this one (`_status-variant.scss`) would also get rid of the duplicative `darkMode` block.

rfellows (changes requested):

nifi-frontend/src/main/frontend/libs/shared/src/assets/themes/components/_option.scss
scottyaslan: This is done in the latest commit — _select.scss has been deleted and all .mat-mdc-option rules now live in _option.scss. The cursor rules from _select.scss were superseded by the new explicit state-layer rules and removed entirely. The mdc-list-item--disabled color rule that was previously stranded in _select.scss's .darkMode block has been promoted to :root in _option.scss. One file now owns the full option surface: the Material mixin, base layout, disabled styling, and all interactive state layers.

nifi-frontend/src/main/frontend/libs/shared/src/components/multi-select-option/multi-select-option.component.scss
scottyaslan: This is a valid concern in principle, but multi-select-option.component.scss no longer contains any .mat-mdc-option state-layer rules — they have been moved to _option.scss in the latest commit. The current component SCSS contains only two things: the .multi-select-option host layout (display: flex; flex-direction: row) and the .mat-pseudo-checkbox-minimal checkmark color, both of which are genuinely specific to MultiSelectOption's rendering.

The comment in _option.scss (lines 62–64) explains why those rules live in the theme partial rather than the component file: Angular Material's lazy style injection never fires for MultiSelectOption because it extends MatOption via DI but renders a custom element — so the state-layer backgrounds would be missing entirely unless declared in a partial that's always in the bundle. Your concern about ViewEncapsulation.None making component styles globally scoped is correct, but the solution here isn't to move rules back to the component file; it's the opposite — get them out of the component file entirely and into the theme partial.

rfellows (approved):
Thanks @scottyaslan 👍 
This makes NiFi UI much more modern in its styling approach.

PR #11267, opened 2026-05-20
NIFI-15957 - Do not resolve parameter references when displaying parameter contexts

pvillard31:
# Summary

NIFI-15957 - Do not resolve parameter references when displaying parameter contexts

Following changes done in [NIFI-15829](https://issues.apache.org/jira/browse/NIFI-15829) we can now have parameters referencing inherited parameters. However, when showing the parameter contexts in the UI the parameters that are referencing parameters are now showing the resolved value instead of the raw reference. While this is not an issue, it can be confusing and we should display the raw reference.

Example:

```
MyInheritedParameter => MyInheritedValue
MyParameter => #{MyInheritedParameter}
```

Showing the parameter context in the UI, it would show:

```
MyParameter => MyInheritedValue
```

(even though the flow definition does have the reference)

We want to always show:

```
MyParameter => #{MyInheritedParameter}
```

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -1882,14 +1882,14 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
     }
 
     private void collectValueAndReferences(final ParameterContext parameterContext, final Map<String, ParameterValueAndReferences> valueAndRef) {
-        parameterContext.getEffectiveParameters()
+        parameterContext.getRawEffectiveParameters()
                 .forEach((pd, param) -> valueAndRef.put(pd.getName(), getValueAndReferences(param)));
     }
 
     protected Set<String> getUpdatedParameterNames(final ParameterContext parameterContext, final VersionedParameterContext proposed) {
         final Map<String, ParameterValueAndReferences> originalValues = new HashMap<>();
         collectValueAndReferences(parameterContext, originalValues);
-        parameterContext.getEffectiveParameters().forEach((pd, param) -> originalValues.put(pd.getName(), getValueAndReferences(param)));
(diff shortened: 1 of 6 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -1882,14 +1882,14 @@ public void synchronize(final ParameterContext parameterContext, final Versioned
...
-        parameterContext.getEffectiveParameters()
+        parameterContext.getRawEffectiveParameters()
                 .forEach((pd, param) -> valueAndRef.put(pd.getName(), getValueAndReferences(param)));
     }
 
     protected Set<String> getUpdatedParameterNames(final ParameterContext parameterContext, final VersionedParameterContext proposed) {
         final Map<String, ParameterValueAndReferences> originalValues = new HashMap<>();
         collectValueAndReferences(parameterContext, originalValues);
-        parameterContext.getEffectiveParameters().forEach((pd, param) -> originalValues.put(pd.getName(), getValueAndReferences(param)));
+        parameterContext.getRawEffectiveParameters().forEach((pd, param) -> originalValues.put(pd.getName(), getValueAndReferences(param)));
```
exceptionfactory: Is this call still needed? It appears to be doing the same thing as the `collectValueAndReferences` method?

exceptionfactory (commented):
Thanks for improving the behavior of Parameter reference value display. I raised one functional question for the Component Synchronizer, but otherwise the changes look ready.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
pvillard31: Good catch, addressed!

exceptionfactory (approved):
Thanks @pvillard31, the latest version looks good. +1 merging

PR #11273, opened 2026-05-21
NIFI-15967 Adjust Proxy Header Validation for Replicated Requests

exceptionfactory:
# Summary

[NIFI-15967](https://issues.apache.org/jira/browse/NIFI-15967) Adjusts the implementation of proxy header validation to support authenticated and replicated requests between NiFi cluster nodes.

Proxy host header validation applies only when the application is enabled for HTTPS, so the implementation checks for the presence of X509 Peer Certificates in the TLS Session. This occurs after a successful TLS handshake, and the presence of the client certificates in the TLS Session indicates that mutual TLS authentication completed as expected. To differentiate requests, the implementation looks for the presence of the custom `request-replicated` header, which the application sets when constructing forwarded requests. Without both of these qualifiers, standard proxy host header validation occurs, checking the values configured in the `nifi.web.proxy.host` property, as implemented in [NIFI-15953](https://issues.apache.org/jira/browse/NIFI-15953). 

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/connector/ProxyHeaderValidatorCustomizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/connector/ProxyHeaderValidatorCustomizer.java
@@ -21,9 +21,13 @@ import org.eclipse.jetty.http.HttpException;
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.http.HttpURI;
+import org.eclipse.jetty.io.Connection;
+import org.eclipse.jetty.io.EndPoint;
+import org.eclipse.jetty.server.ConnectionMetaData;
 import org.eclipse.jetty.server.HttpConfiguration;
 import org.eclipse.jetty.server.Request;
 
+import java.security.cert.X509Certificate;
 import java.util.Objects;
 import java.util.Set;
 
@@ -32,6 +36,8 @@ import java.util.Set;
  */
 public class ProxyHeaderValidatorCustomizer implements HttpConfiguration.Customizer {
     private static final String MISDIRECTED_REQUEST_REASON = "Invalid Proxy Host Requested";
+
+    private static final String REQUEST_REPLICATED_HEADER = "request-replicated";
 
     private static final Set<String> SUPPORTED_PROXY_HOST_HEADERS = Set.of(
             ProxyHeader.PROXY_HOST.getHeader(),
@@ -65,6 +71,24 @@ public class ProxyHeaderValidatorCustomizer implements HttpConfiguration.Customi
     }
 
     private Request customizeSecureRequest(final Request request) {
+        final X509Certificate peerCertificate = findPeerCertificate(request);
+
+        // Requests not authenticated with Client Certificates require header validation
+        if (peerCertificate == null) {
+            processProxyHostHeaders(request);
+        } else {
+            // Requests authenticated with Client Certificates but not indicated as replicated require header validation
+            final HttpFields requestHeaders = request.getHeaders();
+            final String requestReplicated = requestHeaders.get(REQUEST_REPLICATED_HEADER);
+            if (requestReplicated == null) {
+                processProxyHostHeaders(request);
+            }
+        }
+
+        return request;
+    }
+
+    private void processProxyHostHeaders(final Request request) {
         final HttpURI requestUri = request.getHttpURI();
         final String requestHost = requestUri.getHost();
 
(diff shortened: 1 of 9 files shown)


nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/connector/ProxyHeaderValidatorCustomizer.java
```diff
@@ -65,6 +71,24 @@ public Request customize(final Request request, final HttpFields.Mutable respons
     }
 
     private Request customizeSecureRequest(final Request request) {
+        final X509Certificate peerCertificate = findPeerCertificate(request);
+
+        // Requests not authenticated with Client Certificates require header validation
+        if (peerCertificate == null) {
+            processProxyHostHeaders(request);
+        } else {
```
pvillard31, line 78: Should this branch also require the peer certificate to match a known cluster node identity (or come from the cluster-only trust store), so that a non-node user holding a valid client certificate cannot bypass proxy host validation simply by adding the `request-replicated` header?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/connector/ProxyHeaderValidatorCustomizer.java
```diff
@@ -33,6 +37,8 @@
 public class ProxyHeaderValidatorCustomizer implements HttpConfiguration.Customizer {
     private static final String MISDIRECTED_REQUEST_REASON = "Invalid Proxy Host Requested";
 
+    private static final String REQUEST_REPLICATED_HEADER = "request-replicated";
```
pvillard31: Could we reference `RequestReplicationHeader.REQUEST_REPLICATED.getHeader()` (possibly after moving that enum, or just this single constant, into `nifi-web-servlet-shared` next to `ProxyHeader`) so the header value is defined in exactly one place?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/test/java/org/apache/nifi/web/server/StandardServerProviderTest.java
```diff
@@ -82,6 +82,8 @@ class StandardServerProviderTest {
 
     private static final String HOST_HEADER = "Host";
 
+    private static final String REQUEST_REPLICATED_HEADER = "request-replicated";
```
pvillard31: Same question on the test side: can `REQUEST_REPLICATED_HEADER` here share the canonical constant rather than redefining the literal?

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/connector/ProxyHeaderValidatorCustomizer.java
```diff
@@ -85,7 +109,24 @@ private Request customizeSecureRequest(final Request request) {
...
         }
+    }
 
-        return request;
+    private X509Certificate findPeerCertificate(final Request request) {
+        final X509Certificate peerCertificate;
+        final ConnectionMetaData connectionMetaData = request.getConnectionMetaData();
+        final Connection connection = connectionMetaData.getConnection();
+        final EndPoint endPoint = connection.getEndPoint();
+        final EndPoint.SslSessionData sslSessionData = endPoint.getSslSessionData();
```
pvillard31, line 118: Is there any code path under a successful `isSecure()` where `connection.getEndPoint()` or `endPoint.getSslSessionData()` could be null (for example during a protocol upgrade), and if so should we guard the chain rather than risk an `NullPointerException` inside the customizer?

exceptionfactory: Good question, but no, at this point in the Connection handling, the `SslSessionData` can never be `null`.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/connector/ProxyHeaderValidatorCustomizer.java
exceptionfactory: Thanks, I looked at options and ended up creating a new `ReplicationHeader` with just this new enum and put it in `nifi-framework-core`, as that appeared to be a better location given the cluster-focused nature of this header.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-jetty/src/main/java/org/apache/nifi/web/server/connector/ProxyHeaderValidatorCustomizer.java
exceptionfactory: I considered that additional approach, but it places additional constraints on information that is not necessarily available. The current approach places inherent security constraints with mTLS, and also fits within current configuration boundaries.

pvillard31 (changes requested):
Thanks for the updates @exceptionfactory - there are some test failures to address

exceptionfactory:
> Thanks for the updates @exceptionfactory - there are some test failures to address

Thanks @pvillard31, I pushed an update to adjust the filtering method in `ReplicationHeaderUtils` to maintain current behavior.

pvillard31 (approved):
Thanks, latest LGTM and can be merged after successful checks.

PR #11275, opened 2026-05-22
NIFI-15883: Fix Wait/Notify race condition with CAS-aware replace and complete

rakesh-rsky:
NIFI-15883: Fix Wait/Notify race condition with CAS-aware replace and complete

- WaitNotifyProtocol.replace(Signal): Changed from returning boolean to void.
  Throws ConcurrentModificationException when the underlying cache.replace()
  returns false (CAS failure due to concurrent Notify), making the API
  consistent with complete(Signal) which also throws on concurrent modification.

- WaitNotifyProtocol.notify(): Updated retry loop to catch
  ConcurrentModificationException from replace() instead of checking
  the boolean return value.

- WaitNotifyProtocol.complete(Signal): Replaces raw cache.remove() with a
  version-aware approach that re-fetches the cache entry and compares
  revisions before removing. Throws ConcurrentModificationException if
  a concurrent Notify updated the signal between Wait's initial read
  and the remove call, preventing silent data loss.

- Wait.java: Simplified call sites - protocol.replace(signal) and
  protocol.complete(signal) both now throw ConcurrentModificationException
  on concurrent modification. The catch block handles both IOException and
  ConcurrentModificationException with session rollback and ProcessException.

- TestWaitNotifyProtocol: Add three tests for complete(Signal):
  happy-path removal, concurrent-modification detection, and no-op
  when the signal is already absent.

- TestWait: Add two integration tests that verify session rollback and
  ProcessException wrapping when replace() throws on CAS failure
  (waitProgressed path) and when complete() detects a stale revision
  (waitCompleted path).

# Summary

[NIFI-15883](https://issues.apache.org/jira/browse/NIFI-15883)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/Wait.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/Wait.java
@@ -46,6 +46,7 @@ import org.apache.nifi.processors.standard.WaitNotifyProtocol.Signal;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -558,7 +559,7 @@ public class Wait extends AbstractProcessor {
         // Update signal if needed.
         try {
             if (waitCompleted) {
-                protocol.complete(signalId);
+                protocol.complete(signal);
                 if (logger.isDebugEnabled()) {
                     logger.debug("Completed wait for signalId='{}' and removed signal from cache", signalId);
                 }
@@ -570,6 +571,10 @@ public class Wait extends AbstractProcessor {
                 }
             }
 
+        } catch (final ConcurrentModificationException e) {
+            logger.warn("Concurrent modification detected for signal [{}], rolling back session to retry: {}", signalId, e.getMessage());
+            session.rollback();
+            throw new ProcessException(String.format("Concurrent modification detected while updating signal %s", signalId), e);
         } catch (final IOException e) {
             session.rollback();
             throw new ProcessException(String.format("Unable to communicate with cache while updating %s due to %s", signalId, e), e);
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/WaitNotifyProtocol.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/WaitNotifyProtocol.java
@@ -318,7 +318,10 @@ public class WaitNotifyProtocol {
                     signalId, expectedRevision, actualRevision));
         }
 
-        cache.remove(signalId, stringSerializer);
+        if (!cache.remove(signalId, stringSerializer)) {
+            throw new ConcurrentModificationException(String.format(
+                    "Failed to complete signal [%s]: signal was concurrently removed.", signalId));
+        }
     }
 
     public void replace(final Signal signal) throws IOException, ConcurrentModificationException {
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/WaitNotifyProtocol.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/WaitNotifyProtocol.java
@@ -199,8 +199,11 @@ public class WaitNotifyProtocol {
                 signal.counts.put(counterName, count);
             });
 
-            if (replace(signal)) {
+            try {
+                replace(signal);
                 return signal;
+            } catch (final ConcurrentModificationException ignored) {
+                // CAS failed; retry with fresh signal state after backoff.
             }
 
             long waitMillis = REPLACE_RETRY_WAIT_MILLIS * (i + 1);
@@ -238,7 +241,7 @@ public class WaitNotifyProtocol {
 
     /**
      * Retrieve a stored Signal in the cache engine.
-     * If a caller gets satisfied with the returned Signal state and finish waiting, it should call {@link #complete(String)}
+     * If a caller gets satisfied with the returned Signal state and finish waiting, it should call {@link #complete(Signal)}
      * to complete the Wait Notify protocol.
      * @param signalId a key in the underlying cache engine
      * @return A Signal instance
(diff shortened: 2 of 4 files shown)


nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/Wait.java
```diff
@@ -558,19 +559,22 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
         try {
             if (waitCompleted) {
-                protocol.complete(signalId);
+                protocol.complete(signal);
                 if (logger.isDebugEnabled()) {
                     logger.debug("Completed wait for signalId='{}' and removed signal from cache", signalId);
                 }
             } else if (waitProgressed) {
-                protocol.replace(signal);
+                if (!protocol.replace(signal)) {
```
sheelchand-bah: All though the fix works but the implementation of complete() and replace() are inconsistent. complete() throws the exception, but for replace the caller have to check.

It would be better to that replace() also throws exception when concurrent updates are detected.

sheelchand-bah (commented):
Suggestion: make replace() and complete() function consistent - either both return true/false or both throw the exception.

rakesh-rsky:
@sheelchand-bah Thank you for the review.

Good point on API consistency. Both replace(Signal) and complete(Signal) now throw ConcurrentModificationException on concurrent modification — callers don't need to check return values. I also updated the retry loop in notify() to catch ConcurrentModificationException from replace() and continue retrying (up to MAX_REPLACE_RETRY_COUNT = 5), since that was the previous behavior when it returned false.

sheelchand-bah (approved):

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/WaitNotifyProtocol.java
```diff
@@ -281,22 +284,54 @@ public Signal getSignal(final String signalId) throws IOException, Deserializati
...
+        final Signal current = getSignal(signalId);
+        if (current == null) {
+            return;
+        }
+
+        final Object expectedRevision = signal.cachedEntry != null ? signal.cachedEntry.getRevision().orElse(null) : null;
+        final Object actualRevision = current.cachedEntry.getRevision().orElse(null);
+        if (expectedRevision != null && !expectedRevision.equals(actualRevision)) {
+            throw new ConcurrentModificationException(String.format(
+                    "Failed to complete signal [%s]: signal was concurrently modified (expected revision %s, found %s). Will retry.",
```
turcsanyip: It depends on the caller whether it retries the transaction or gives up, so I suggest removing the "Will retry." suffix.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/WaitNotifyProtocol.java
```diff
@@ -281,22 +284,54 @@ public Signal getSignal(final String signalId) throws IOException, Deserializati
...
         final String signalJson = objectMapper.writeValueAsString(signal);
         if (signal.cachedEntry == null) {
             signal.cachedEntry = new AtomicCacheEntry<>(signal.identifier, signalJson, null);
         } else {
             signal.cachedEntry.setValue(signalJson);
         }
-        return cache.replace(signal.cachedEntry, stringSerializer, stringSerializer);
+        if (!cache.replace(signal.cachedEntry, stringSerializer, stringSerializer)) {
+            throw new ConcurrentModificationException(String.format(
+                    "Failed to update signal [%s] in cache due to concurrent modification. Will retry.", signal.identifier));
```
turcsanyip: Same as line 316.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/WaitNotifyProtocol.java
```diff
@@ -281,22 +284,54 @@ public Signal getSignal(final String signalId) throws IOException, Deserializati
...
+     *         since the caller last read it
      */
-    public void complete(final String signalId) throws IOException {
+    public void complete(final Signal signal) throws IOException, ConcurrentModificationException {
+        final String signalId = signal.identifier;
+
+        // Re-fetch to detect concurrent updates since the signal was last read.
+        final Signal current = getSignal(signalId);
+        if (current == null) {
+            return;
```
turcsanyip: I think this case should also be treated as a concurrent modification and should throw an exception.

Another process may have consumed and removed the signal after we initially read it, which means the value we based our decision on was no longer valid.

Possible scenario:

- signal counter = 1 initially
- 2 concurrent threads read the signal
- both decrease the counter to 0 and resume the waiting FlowFile
- both removes the signal (as the counter reached 0)

=> only 1 FF should have been resumed based on the counter

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/Wait.java
```diff
@@ -570,7 +571,7 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
                 }
             }
 
-        } catch (final IOException e) {
+        } catch (final IOException | ConcurrentModificationException e) {
             session.rollback();
             throw new ProcessException(String.format("Unable to communicate with cache while updating %s due to %s", signalId, e), e);
         }
```
turcsanyip: In the event of an optimistic lock failure, the `Wait` processor should retry the transaction silently rather than failing the FlowFile immediately. (similar to the `for` cycle and `MAX_REPLACE_RETRY_COUNT` in `WaitNotifyProtocol.notify()`).

I see the `Wait` flow is more complicated and a simple `for` cycle around it may not be possible.

Minimally, I would separate the exception handling:

- the `Unable to communicate with cache` error message is not appropriate for the concurrent modification case
- warning log may be enough instead of error

The preferred way would definitely be the internal retry with max retry count.

turcsanyip (changes requested):
Thanks for working on the issue @rakesh-rsky!

The general concept looks good. Please find my comments below.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/WaitNotifyProtocol.java
```diff
@@ -281,22 +284,55 @@ public Signal getSignal(final String signalId) throws IOException, Deserializati
...
+
+        final Object expectedRevision = signal.cachedEntry != null ? signal.cachedEntry.getRevision().orElse(null) : null;
+        final Object actualRevision = current.cachedEntry.getRevision().orElse(null);
+        if (expectedRevision != null && !expectedRevision.equals(actualRevision)) {
+            throw new ConcurrentModificationException(String.format(
+                    "Failed to complete signal [%s]: signal was concurrently modified (expected revision %s, found %s).",
+                    signalId, expectedRevision, actualRevision));
+        }
+
         cache.remove(signalId, stringSerializer);
```
turcsanyip: As you mentioned, the ideal solution would be the atomic compare-and-delete supported by `AtomicDistributedMapCacheClient`. However, the TOCTOU problem can be minimized further if we check the return value of `remove()`.
```suggestion
        if (!cache.remove(signalId, stringSerializer)) {
            throw new ConcurrentModificationException(String.format(
                    "Failed to complete signal [%s]: signal was concurrently removed.", signalId));
        }
```

turcsanyip (changes requested):
@rakesh-rsky Thanks for the review changes and sorry for the delay. I found one more improvement point (pls. see below).

When you add changes to the PR, could you please keep them in a separate commit instead of squash + force push? That would make the review easier (cleaner diff with the latest changes only). Thanks.

turcsanyip (approved):
Thanks @rakesh-rsky!

+1 LGTM

PR #11281, opened 2026-05-25
NIFI-15683 Fix SFTP Move strategy failure when destination file alrea…

rakesh-rsky:
NIFI-15683 Fix SFTP Move strategy failure when destination file already exists

When FetchSFTP uses completion strategy 'Move', some SFTP servers (notably OpenSSH) return SSH_FX_FAILURE on rename if the destination file already exists, rather than atomically overwriting it. This causes a warning and leaves the source file in place.

Add a delete-then-rename fallback: if the initial rename fails, attempt to delete the destination and retry the rename. If the destination does not exist (FileNotFoundException), the original rename exception is re-thrown to preserve the error signal for unrelated failures.

# Summary

[NIFI-15683](https://issues.apache.org/jira/browse/NIFI-15683)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FetchFileTransfer.java
+++ b/nifi-extension-bundles/nifi-extension-utils/nifi-file-transfer/src/main/java/org/apache/nifi/processor/util/file/transfer/FetchFileTransfer.java
@@ -21,6 +21,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.annotation.lifecycle.OnScheduled;
 import org.apache.nifi.annotation.lifecycle.OnStopped;
 import org.apache.nifi.components.AllowableValue;
+import org.apache.nifi.components.DescribedValue;
 import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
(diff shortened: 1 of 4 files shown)


exceptionfactory (changes requested):
Thanks for proposing this change @rakesh-rsky. A similar effort recently stalled, attempting to handle failures related to moving files. There are several problems with potential solutions, including existing expectations, and lack of clear failure paths for post-fetch operations. For these reasons, falling back to deleting a file may not be appropriate. In addition, catching the general `IOException` is too broad to be a safe approach. This implementation needs further consideration of potential options, before moving forward.

rakesh-rsky:
Thank you for the review feedback @exceptionfactory.

You're right on both points — the automatic delete-then-rename fallback was too aggressive, and catching the general `IOException` was too broad and could trigger a remote file deletion for unrelated failures like network errors or permission issues.

I've reworked the approach:

 - Added a Move Conflict Resolution property with two values: Fail (default, preserves existing behaviour) and Replace (explicit opt-in).
 - When Replace is selected, the processor proactively checks if the destination exists using getRemoteFileInfo() before attempting the rename. If it exists, it deletes it and renames. If not, it simply renames as normal.
 - This removes the broad IOException catch entirely — no exception-driven control flow, no silent destructive fallback.
 - The default is Fail, so existing deployments are unaffected.

This gives users explicit control over the conflict resolution behaviour while keeping the failure paths clean and predictable.

exceptionfactory (changes requested):
Thanks for the work on this @rakesh-rsky. The revised approach looks like a good way forward, the new property and opt-in behavior provide a good path for optional handling.

From an implementation perspective, it would be helpful to define the new allowable values in an `enum` that implements `DescribedValue`, following the pattern elsewhere in the project.

Any description strings that go over one line can also be defined as multiline strings, instead of using concatenation.

The new test class looks good, and although there are divergent naming conventions, I recommend that new tests be named with `Test` as the suffix, so `FetchSFTPTest` in this case.

If you can address these items, that would help move this forward.

exceptionfactory (commented):
Thanks for the updates @rakesh-rsky. I pushed a minor update to use `asAllowableValue()` for referencing the property value, and adjusted some wording along with rebasing.

This should be ready to go on successful automated build completion.

exceptionfactory (approved):
Thanks again @rakesh-rsky!

PR #11283, opened 2026-05-25
NIFI-15973 Fix issues caused by the use of deprecated GregorianCalendar

Kamilkime:
# Summary

[NIFI-15973](https://issues.apache.org/jira/browse/NIFI-15973)
  
Avoid Julian-calendar shift when converting `java.sql.Date` / `java.sql.Timestamp` values to `java.time` types (and back) in NiFi's record framework.

`java.sql.Timestamp#toLocalDateTime`, `java.sql.Timestamp#valueOf(LocalDateTime)`, `java.sql.Date#toLocalDate`, and `java.sql.Date#valueOf(LocalDate)` all route through GregorianCalendar, which applies Julian-calendar semantics for years before 1582-10-15. As a result, a value whose epoch milliseconds represent year 0001 in the proleptic Gregorian calendar (the calendar `java.time`, Avro, BigQuery, or Snowflake all use) is rendered as year 0000 — about a 2-day backward shift. Downstream sinks that validate against an inclusive year-1-to-9999 range (e.g., the Snowflake Ingest SDK) then reject the row: `Timestamp out of representable inclusive range of years between 1 and 9999, ... value:0000-12-30T10:47Z`.

The fix replaces every Julian-aware bridge between `java.sql.*` and `java.time.*` with an Instant-based one, which preserves the proleptic-Gregorian instant the source actually carried.
  
**Tests:**

  - ObjectLocalDateFieldConverterTest (new): null input, LocalDate passthrough, java.sql.Date modern-year round-trip, and a year-0001 java.sql.Date regression test that asserts the result is year 0001 (not year 0000).
  - TestObjectLocalDateTimeFieldConverter: add testConvertTimestampYearOneIsProlepticGregorian (year-0001 Timestamp round-trip) and testConvertStringYearOneIsProlepticGregorian (year-0001 string parsed via formatter).
  - ObjectTimestampFieldConverterTest: add testConvertFieldStringYearOneIsProlepticGregorian and testConvertFieldLocalDateTimeYearOneIsProlepticGregorian to lock in the year-0001 result for both string and LocalDateTime inputs.
  - ObjectStringFieldConverterTest: add testConvertFieldTimestampYearOneIsProlepticGregorian covering the formatted-output path.
  - New regression tests fail on main and pass on this branch.
  
**Compatibility:**
  
Backwards compatible. Instant-based conversion and Julian-aware Timestamp / Date conversion agree for every date from 1582-10-15 onward, so all values produced after the Gregorian cutover (i.e., everything the vast majority of users have ever ingested) are unchanged. Pre-cutover values — including the 1582-10-05 → 1582-10-14 gap that Julian had but proleptic Gregorian does not — now match the proleptic-Gregorian instant the source actually carried, which is what `java.time`, Avro, and downstream sinks already assume.

No public API signatures change, the fix is purely internal to the existing converter implementations.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectLocalDateFieldConverterTest.java
+++ b/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectLocalDateFieldConverterTest.java
@@ -27,8 +27,8 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 
 class ObjectLocalDateFieldConverterTest {
-    private static final ObjectLocalDateFieldConverter CONVERTER = new ObjectLocalDateFieldConverter();
 
+    private static final ObjectLocalDateFieldConverter CONVERTER = new ObjectLocalDateFieldConverter();
     private static final String FIELD_NAME = LocalDate.class.getSimpleName();
 
     @Test
@@ -46,22 +46,15 @@ class ObjectLocalDateFieldConverterTest {
     void testConvertFieldSqlDateModernYear() {
         final LocalDate localDate = LocalDate.of(2025, 5, 25);
         final Date sqlDate = new Date(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli());
-
         final LocalDate result = CONVERTER.convertField(sqlDate, Optional.empty(), FIELD_NAME);
-
         assertEquals(localDate, result);
     }
 
     @Test
     void testConvertFieldSqlDateYearOneIsProlepticGregorian() {
-        // A java.sql.Date built from a proleptic-Gregorian year-1 epoch must round-trip back
-        // to year 1, not be shifted to year 0 by Julian calendar semantics in
-        // java.sql.Date#toLocalDate (which uses deprecated getYear/getMonth/getDate).
         final LocalDate yearOne = LocalDate.of(1, 1, 1);
         final Date sqlDate = new Date(yearOne.atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli());
-
         final LocalDate result = CONVERTER.convertField(sqlDate, Optional.empty(), FIELD_NAME);
-
         assertEquals(yearOne, result);
     }
 }
--- a/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectLocalDateFieldConverterTest.java
+++ b/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectLocalDateFieldConverterTest.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.serialization.record.field;
+
+import org.junit.jupiter.api.Test;
+
+import java.sql.Date;
+import java.time.LocalDate;
+import java.time.ZoneId;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+class ObjectLocalDateFieldConverterTest {
+    private static final ObjectLocalDateFieldConverter CONVERTER = new ObjectLocalDateFieldConverter();
+
+    private static final String FIELD_NAME = LocalDate.class.getSimpleName();
+
+    @Test
+    void testConvertFieldNull() {
+        assertNull(CONVERTER.convertField(null, Optional.empty(), FIELD_NAME));
+    }
+
+    @Test
+    void testConvertFieldLocalDate() {
+        final LocalDate input = LocalDate.of(2025, 5, 25);
+        assertEquals(input, CONVERTER.convertField(input, Optional.empty(), FIELD_NAME));
+    }
+
+    @Test
+    void testConvertFieldSqlDateModernYear() {
+        final LocalDate localDate = LocalDate.of(2025, 5, 25);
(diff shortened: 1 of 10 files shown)


nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverterTest.java
```diff
@@ -162,6 +162,22 @@ public void testConvertFieldStringFormatCustomZoneOffsetCoordinatedUniversalTime
         assertEquals(expected, timestamp);
     }
 
+    @Test
+    public void testConvertFieldStringYearOneIsProlepticGregorian() {
+        final String yearOne = "0001-01-01 12:00:00";
+        final Timestamp timestamp = CONVERTER.convertField(yearOne, DEFAULT_PATTERN, FIELD_NAME);
+        final Instant expected = LocalDateTime.of(1, 1, 1, 12, 0, 0).atZone(ZoneId.systemDefault()).toInstant();
+        assertEquals(expected, timestamp.toInstant());
```
exceptionfactory: Unit tests should assert expected `Timestamp` values, not values converted to `Instant` objects.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/TestObjectLocalDateTimeFieldConverter.java
```diff
@@ -100,4 +101,18 @@ public void testWithDateFormatMicrosecondPrecision() {
...
+        final LocalDateTime yearOne = LocalDateTime.of(1, 1, 1, 12, 0, 0);
+        final Timestamp timestamp = Timestamp.from(yearOne.atZone(ZoneId.systemDefault()).toInstant());
+        final LocalDateTime result = converter.convertField(timestamp, Optional.empty(), FIELD_NAME);
+        assertEquals(yearOne, result);
+    }
+
+    @Test
+    public void testConvertStringYearOneIsProlepticGregorian() {
+        final LocalDateTime result = converter.convertField("0001-01-01 12:00:00", Optional.of("yyyy-MM-dd HH:mm:ss"), FIELD_NAME);
+        assertEquals(LocalDateTime.of(1, 1, 1, 12, 0, 0), result);
```
exceptionfactory: The expected value should be declared separately instead of inline instance creation for clarity.

nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-avro-record-utils/src/main/java/org/apache/nifi/avro/AvroTypeUtil.java
```diff
@@ -1153,11 +1153,9 @@ private static Object normalizeValue(final Object value, final Schema avroSchema
...
                 if (LOGICAL_TYPE_DATE.equals(logicalName)) {
                     // date logical name means that the value is number of days since Jan 1, 1970
-                    // Handle both Integer (legacy) and LocalDate (newer Avro libraries)
-                    if (value instanceof LocalDate localDate) {
-                        return java.sql.Date.valueOf(localDate);
-                    }
-                    return java.sql.Date.valueOf(LocalDate.ofEpochDay((int) value));
+                    // Handle both Integer (legacy) and LocalDate (newer Avro libraries).
+                    final LocalDate localDate = (value instanceof LocalDate ld) ? ld : LocalDate.ofEpochDay((int) value);
+                    return new java.sql.Date(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli());
```
exceptionfactory: The inline ternary and cast is difficult to read, and the multiple levels of conversion to create the new `java.sql.Date()` should be split out and declared.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectLocalDateFieldConverterTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @Test
+    void testConvertFieldLocalDate() {
+        final LocalDate input = LocalDate.of(2025, 5, 25);
+        assertEquals(input, CONVERTER.convertField(input, Optional.empty(), FIELD_NAME));
+    }
+
+    @Test
+    void testConvertFieldSqlDateModernYear() {
+        final LocalDate localDate = LocalDate.of(2025, 5, 25);
+        final Date sqlDate = new Date(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli());
```
exceptionfactory: Intermediate values should be declared for readability.

exceptionfactory (changes requested):
Thanks for describing the issues with the current  Date and Timestamp conversion @Kamilkime. This pull request highlights the subtle behavior of different time classes.

I noted some recommendations related to the unit test methods for readability and clear construction.

Regarding the functional change, it appears reasonable from one perspective, since date conversion of all modern values are unchanged. I would expect this to cover many use cases without concern.

The key question is whether the change in behavior for ancient dates is acceptable. At minimum, the nuance should be commented in the Converter classes that the described approach uses the Proleptic Gregorian Calendar System, avoid the implicit use of deprecated Date conversion methods. Declaring the intermediate values would also helpful emphasize the behavioral adjustment.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectLocalDateFieldConverterTest.java
Kamilkime: Extracted zonedDate

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/TestObjectLocalDateTimeFieldConverter.java
Kamilkime: Extracted expected variable

nifi-extension-bundles/nifi-extension-utils/nifi-record-utils/nifi-avro-record-utils/src/main/java/org/apache/nifi/avro/AvroTypeUtil.java
Kamilkime: Used an if statement for the localDate value, and extracted zonedDate

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/field/ObjectTimestampFieldConverterTest.java
Kamilkime: Changed to Timestamp assertions

Kamilkime (commented):
Thanks @exceptionfactory, I addressed your comments, please let me know if it looks good now

exceptionfactory (approved):
Thanks for working through the feedback and making adjustments @Kamilkime, the latest version looks good. +1 merging

PR #11286, opened 2026-05-27
NIFI-15978: Rendering allowable value descriptions in the Connector w…

mcgilman:
…izard.

## Summary

The Connector wizard's property dropdowns currently render only an allowable value's display name, even though the framework's `AllowableValueDTO` already serializes a `description` field (used elsewhere — e.g. processor property dialogs — to give users context about each option). This PR plumbs that description through to the wizard's `searchable-select` so users see the same secondary descriptive text per option that they see in the rest of the UI.

## Changes

- **`libs/shared/src/types/index.ts`** — Added the optional `description?: string` to the nested `AllowableValue.allowableValue` shape so it matches what the backend already returns from `AllowableValueDTO`.
- **`libs/shared/src/components/connector-property-input/connector-property-input.component.ts`** — In `computeSelectOptions`, propagate `av.allowableValue.description` onto the emitted `SearchableSelectOption`. The `searchable-select` component already renders an option's `description` as a secondary line (and already sizes virtual-scroll rows accordingly), so no template or styling changes are needed.
- **`libs/shared/src/components/connector-property-input/connector-property-input.component.spec.ts`** — Extended the `makeAllowable` helper to accept an optional description, adjusted existing assertions to expect `description: undefined` where appropriate, and added focused specs covering description propagation for both the static `property.allowableValues` path and the dynamic `PropertyAllowableValuesState` path.

## Backend alignment

No backend changes are required. Verified that:

- `AllowableValueDTO` already exposes `description` (`getDescription()` / `setDescription()`).
- `DtoFactory.createAllowableValueDto(...)` populates the description for statically declared allowable values.
- `StandardNiFiServiceFacade` populates the description for allowable values produced dynamically via a property's `AllowableValuesProvider`.

So any connector that declares descriptions on its `AllowableValue`s — static or dynamic — will surface them in the wizard with no further wiring.

### DIFF
--- a/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.spec.ts
+++ b/nifi-frontend/src/main/frontend/libs/shared/src/components/connector-property-input/connector-property-input.component.spec.ts
@@ -44,9 +44,9 @@ function makeProp(overrides: Partial<ConnectorPropertyDescriptor> = {}): Connect
     };
 }
 
-function makeAllowable(value: string, displayName: string = value): AllowableValue {
+function makeAllowable(value: string, displayName: string = value, description?: string): AllowableValue {
     return {
-        allowableValue: { value, displayName },
+        allowableValue: { value, displayName, description },
         canRead: true
     };
 }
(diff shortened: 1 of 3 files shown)


scottyaslan:
Reviewing...

PR #11290, opened 2026-05-27
NIFI-15960 Replaced current property in AvroSchemaRegistry with a new one which utilizes a validation strategy where it details what Avro validations occur when choosing strict validation of Avro schemas.

dan-s1:
# Summary

[NIFI-15960](https://issues.apache.org/jira/browse/NIFI-15960)

This PR primarily details the various validations which occur when strict validation is chosen for validating Avro schemas in the AvroSchemaRegistry. The unit tests were overhauled to use the NIFI testing framework and multiple unit tests were added especially to contrast the difference between strict and lenient validation of Avro schemas.
 
# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/main/java/org/apache/nifi/schemaregistry/services/AvroSchemaRegistry.java
+++ b/nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/main/java/org/apache/nifi/schemaregistry/services/AvroSchemaRegistry.java
@@ -34,7 +34,6 @@ import org.apache.nifi.schema.access.SchemaNotFoundException;
 import org.apache.nifi.serialization.record.RecordSchema;
 import org.apache.nifi.serialization.record.SchemaIdentifier;
 
-import java.io.IOException;
 import java.util.Collection;
 import java.util.EnumSet;
 import java.util.HashSet;
@@ -52,20 +51,20 @@ import java.util.concurrent.ConcurrentMap;
         description = "Adds a named schema using the JSON string representation of an Avro schema",
         expressionLanguageScope = ExpressionLanguageScope.NONE)
 public class AvroSchemaRegistry extends AbstractControllerService implements SchemaRegistry {
+    static final List<String> OBSOLETE_VALIDATE_FIELD_NAMES = List.of("avro-reg-validated-field-names", "Validate Field Names");
     private static final Set<SchemaField> schemaFields = EnumSet.of(SchemaField.SCHEMA_NAME, SchemaField.SCHEMA_TEXT, SchemaField.SCHEMA_TEXT_FORMAT);
     private final ConcurrentMap<String, RecordSchema> recordSchemas = new ConcurrentHashMap<>();
 
-    static final PropertyDescriptor VALIDATE_FIELD_NAMES = new PropertyDescriptor.Builder()
-            .name("Validate Field Names")
-            .description("Whether or not to validate the field names in the Avro schema based on Avro naming rules. If set to true, all field names must be valid Avro names, "
-                    + "which must begin with [A-Za-z_], and subsequently contain only [A-Za-z0-9_]. If set to false, no validation will be performed on the field names.")
-            .allowableValues("true", "false")
-            .defaultValue("true")
+    public static final PropertyDescriptor VALIDATION_STRATEGY = new PropertyDescriptor.Builder()
+            .name("Validation Strategy")
+            .description("Set the strategy for the level of Avro validation required for field names, namespaces and default values")
             .required(true)
+            .allowableValues(ValidationStrategy.class)
+            .defaultValue(ValidationStrategy.STRICT)
             .build();
 
     private static final List<PropertyDescriptor> PROPERTY_DESCRIPTORS = List.of(
-            VALIDATE_FIELD_NAMES
+            VALIDATION_STRATEGY
     );
 
     @Override
@@ -92,8 +91,8 @@ public class AvroSchemaRegistry extends AbstractControllerService implements Sch
 
     @Override
     protected Collection<ValidationResult> customValidate(ValidationContext validationContext) {
-        Set<ValidationResult> results = new HashSet<>();
-        boolean strict = validationContext.getProperty(VALIDATE_FIELD_NAMES).asBoolean();
+        final Set<ValidationResult> results = new HashSet<>();
+        final ValidationStrategy validationStrategy = validationContext.getProperty(VALIDATION_STRATEGY).asAllowableValue(ValidationStrategy.class);
 
         // Iterate over dynamic properties, validating the schemas, and adding results
         validationContext.getProperties().entrySet().stream().filter(entry -> entry.getKey().isDynamic()).forEach(entry -> {
@@ -101,7 +100,7 @@ public class AvroSchemaRegistry extends AbstractControllerService implements Sch
             String input = entry.getValue();
 
             try {
-                final Schema.Parser parser = strict
+                final Schema.Parser parser = ValidationStrategy.STRICT == validationStrategy
                         ? new Schema.Parser(NameValidator.STRICT_VALIDATOR).setValidateDefaults(true)
                         : new Schema.Parser(NameValidator.NO_VALIDATION).setValidateDefaults(false);
                 final Schema avroSchema = parser.parse(input);
@@ -127,7 +126,7 @@ public class AvroSchemaRegistry extends AbstractControllerService implements Sch
     }
 
     @Override
-    public RecordSchema retrieveSchema(final SchemaIdentifier schemaIdentifier) throws IOException, SchemaNotFoundException {
+    public RecordSchema retrieveSchema(final SchemaIdentifier schemaIdentifier) throws SchemaNotFoundException {
         final Optional<String> schemaName = schemaIdentifier.getName();
         if (schemaName.isPresent()) {
             return retrieveSchemaByName(schemaName.get());
@@ -138,7 +137,22 @@ public class AvroSchemaRegistry extends AbstractControllerService implements Sch
 
     @Override
     public void migrateProperties(PropertyConfiguration config) {
-        config.renameProperty("avro-reg-validated-field-names", VALIDATE_FIELD_NAMES.getName());
+        // NOTE: Although there are multiple names for OBSOLETE_VALIDATE_FIELD_NAMES, the code in the if statement
+        // will execute at most once as the assumption is a configuration has one of the obsolete property names but not all.
+        for (String obsoletePropertyName : OBSOLETE_VALIDATE_FIELD_NAMES) {
+            if (config.hasProperty(obsoletePropertyName) && config.isPropertySet(obsoletePropertyName)) {
+                final String validateFieldNamesRawValue = config.getRawPropertyValue(obsoletePropertyName).orElse(Boolean.FALSE.toString());
+                final boolean validateFieldNames = Boolean.parseBoolean(validateFieldNamesRawValue);
+
+                if (validateFieldNames) {
+                    config.setProperty(VALIDATION_STRATEGY, ValidationStrategy.STRICT.getValue());
+                } else {
+                    config.setProperty(VALIDATION_STRATEGY, ValidationStrategy.LENIENT.getValue());
+                }
+            }
+
+            config.removeProperty(obsoletePropertyName);
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/main/java/org/apache/nifi/schemaregistry/services/AvroSchemaRegistry.java
```diff
@@ -52,20 +51,20 @@
...
     private static final Set<SchemaField> schemaFields = EnumSet.of(SchemaField.SCHEMA_NAME, SchemaField.SCHEMA_TEXT, SchemaField.SCHEMA_TEXT_FORMAT);
     private final ConcurrentMap<String, RecordSchema> recordSchemas = new ConcurrentHashMap<>();
 
-    static final PropertyDescriptor VALIDATE_FIELD_NAMES = new PropertyDescriptor.Builder()
-            .name("Validate Field Names")
-            .description("Whether or not to validate the field names in the Avro schema based on Avro naming rules. If set to true, all field names must be valid Avro names, "
-                    + "which must begin with [A-Za-z_], and subsequently contain only [A-Za-z0-9_]. If set to false, no validation will be performed on the field names.")
-            .allowableValues("true", "false")
-            .defaultValue("true")
+    public static final PropertyDescriptor VALIDATION_STRATEGY = new PropertyDescriptor.Builder()
```
pvillard31: Is there a consumer outside this module that needs `VALIDATION_STRATEGY`, or can it stay package-private like the previous `VALIDATE_FIELD_NAMES`?

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/main/java/org/apache/nifi/schemaregistry/services/AvroSchemaRegistry.java
```diff
@@ -138,7 +137,22 @@ public RecordSchema retrieveSchema(final SchemaIdentifier schemaIdentifier) thro
 
     @Override
     public void migrateProperties(PropertyConfiguration config) {
-        config.renameProperty("avro-reg-validated-field-names", VALIDATE_FIELD_NAMES.getName());
+        // NOTE: Although there are multiple names for OBSOLETE_VALIDATE_FIELD_NAMES, the code in the if statement
+        // will execute at most once as the assumption is a configuration has one of the obsolete property names but not all.
+        for (String obsoletePropertyName : OBSOLETE_VALIDATE_FIELD_NAMES) {
+            if (config.hasProperty(obsoletePropertyName) && config.isPropertySet(obsoletePropertyName)) {
+                final String validateFieldNamesRawValue = config.getRawPropertyValue(obsoletePropertyName).orElse(Boolean.FALSE.toString());
```
pvillard31: Should the fallback in `getRawPropertyValue(...).orElse(Boolean.FALSE.toString())` be `Boolean.TRUE.toString()` instead, so that the rare empty-value case maps to `VALIDATE` (the historical default of the old `Validate Field Names` property) rather than `NONE`?

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/main/java/org/apache/nifi/schemaregistry/services/AvroSchemaRegistry.java
```diff
@@ -138,7 +137,22 @@ public RecordSchema retrieveSchema(final SchemaIdentifier schemaIdentifier) thro
...
+                final boolean validateFieldNames = Boolean.parseBoolean(validateFieldNamesRawValue);
+
+                if (validateFieldNames) {
+                    config.setProperty(VALIDATION_STRATEGY, ValidationStrategy.VALIDATE.getValue());
+                } else {
+                    config.setProperty(VALIDATION_STRATEGY, ValidationStrategy.NONE.getValue());
+                }
+            }
+
+            config.removeProperty(obsoletePropertyName);
```
pvillard31, line 154: Should `removeProperty(obsoletePropertyName)` be guarded by `hasProperty(obsoletePropertyName)` (or the loop broken after the first match), so that the migration result only reports the property that was actually present?

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/main/java/org/apache/nifi/schemaregistry/services/ValidationStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.apache.nifi.components.DescribedValue;
+
+public enum ValidationStrategy implements DescribedValue {
+    VALIDATE("Validate Schemas", """
+            Validate incoming Avro schemas. Validation includes field name,
+            namespace and default value validation. All field names must be valid Avro names
+            that match the following regular expression [A-Za-z_][A-Za-z0-9_]* (i.e. a name that starts with a letter or an underscore, followed by zero or more alphanumeric,
+            characters or underscores), all namespaces must be one or more valid Avro names each matching the aforementioned regular expression and separated by periods
+            and any default value specified must match the field type (e.g. a field with type int must specify a number such as 9 or null)"""),
```
pvillard31: Should the `VALIDATE` description drop the stray comma after `alphanumeric` and remove `or null` from the `int` default example, since `null` is not a valid default for a non-nullable `int`?

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/test/java/org/apache/nifi/schemaregistry/services/TestAvroSchemaRegistry.java
```diff
@@ -17,93 +17,199 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("invalidSchemasForStrictValidation")
+    public void testStrictValidation(String schema, List<String> keyWordsInExceptionMessage) {
+        runner.setProperty(avroSchemaRegistry, SUPPORTED_DYNAMIC_PROPERTY_DESCRIPTOR, schema);
+        runner.setProperty(avroSchemaRegistry, AvroSchemaRegistry.VALIDATION_STRATEGY, ValidationStrategy.VALIDATE.getValue());
+
+        final AssertionFailedError assertionFailedError = assertThrows(AssertionFailedError.class, () -> runner.assertValid(avroSchemaRegistry));
+        keyWordsInExceptionMessage.forEach(keyword -> assertTrue(assertionFailedError.getMessage().contains(keyword)));
```
pvillard31: Can these assertions check the `customValidate` results (for example, that `runner.assertNotValid(...)` is satisfied, or that a `ValidationResult` for the failing dynamic property is reported) instead of matching literal substrings from Avro's exception messages, so an Avro version bump that rewords its errors does not break the tests?

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/main/java/org/apache/nifi/schemaregistry/services/AvroSchemaRegistry.java
```diff
@@ -138,7 +137,22 @@ public RecordSchema retrieveSchema(final SchemaIdentifier schemaIdentifier) thro
 
     @Override
     public void migrateProperties(PropertyConfiguration config) {
```
pvillard31: Should the parameter `config` and the loop variable `obsoletePropertyName` be declared `final` to match the project's code style rule that every variable which can be `final` must be `final`?

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/test/java/org/apache/nifi/schemaregistry/services/TestAvroSchemaRegistry.java
```diff
@@ -17,93 +17,199 @@
...
+
+        final SchemaIdentifier schemaIdentifier = SchemaIdentifier.builder().name(schemaName).build();
+        final RecordSchema locatedSchema = avroSchemaRegistry.retrieveSchema(schemaIdentifier);
+        assertTrue(locatedSchema.getSchemaText().isPresent());
+        assertEquals(fooSchemaText, locatedSchema.getSchemaText().get());
+    }
+
+    @ParameterizedTest
+    @MethodSource("invalidSchemasForStrictValidation")
+    public void testStrictValidation(String schema, List<String> keyWordsInExceptionMessage) {
```
pvillard31: Now that the enum values are `VALIDATE` and `NONE`, should the test method names and helper sources (such as `testStrictValidation`, `testWithRelaxedValidation`, `invalidSchemasForStrictValidation`, `invalidSchemasForRelaxedValidation`, and the `Strict` / `Lenient` argument-set labels) be renamed to use the same vocabulary, so the test code reads consistently with the production enum?

pvillard31 (changes requested):

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/main/java/org/apache/nifi/schemaregistry/services/AvroSchemaRegistry.java
dan-s1: Javadoc on remove is
`Removes the property with the given name, if it exists. If the property does not exist, this is a no-op.`

Why would it be necessary then to be guarded by `hasProperty(obsoletePropertyName)` (or the loop broken after the first match)?

dan-s1: @pvillard31 Are you okay if this is left as is?

dan-s1:
@pvillard31 Thank you very much for approving my other PRs. Can you please take another look at this PR? My teammates are interested in having this documentation so it is clear what parts of the Avro schema are validated. Thanks!

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/test/java/org/apache/nifi/schemaregistry/services/TestAvroSchemaRegistry.java
```diff
@@ -17,93 +17,186 @@
...
+        final Set<String> expectedRemoved = new HashSet<>(AvroSchemaRegistry.OBSOLETE_VALIDATE_FIELD_NAMES);
+        final Set<String> expectedUpdated = Set.of(AvroSchemaRegistry.VALIDATION_STRATEGY.getName());
+
+        final AvroSchemaRegistry service = new AvroSchemaRegistry();
+        service.migrateProperties(configuration);
+
+        final PropertyMigrationResult result = configuration.toPropertyMigrationResult();
+        final Set<String> propertiesRemoved = result.getPropertiesRemoved();
+        assertEquals(expectedRemoved, propertiesRemoved);
+        assertEquals(expectedUpdated, result.getPropertiesUpdated());
```
pvillard31: Should this test also assert the migrated value of Validation Strategy, so the true to VALIDATE and false to NONE mapping is verified and the test fails if that mapping ever breaks?

nifi-extension-bundles/nifi-registry-bundle/nifi-registry-service/src/test/java/org/apache/nifi/schemaregistry/services/TestAvroSchemaRegistry.java
```diff
@@ -17,93 +17,186 @@
...
+    void testMigrateProperties(MockPropertyConfiguration configuration) {
+        final Set<String> expectedRemoved = new HashSet<>(AvroSchemaRegistry.OBSOLETE_VALIDATE_FIELD_NAMES);
+        final Set<String> expectedUpdated = Set.of(AvroSchemaRegistry.VALIDATION_STRATEGY.getName());
+
+        final AvroSchemaRegistry service = new AvroSchemaRegistry();
+        service.migrateProperties(configuration);
+
+        final PropertyMigrationResult result = configuration.toPropertyMigrationResult();
+        final Set<String> propertiesRemoved = result.getPropertiesRemoved();
+        assertEquals(expectedRemoved, propertiesRemoved);
```
pvillard31: This expects both obsolete names in the removed set, but production StandardPropertyConfiguration.removeProperty records nothing for an absent property while the mock records it unconditionally. Should the test assert only the obsolete name that was actually present, so it matches production behavior?

pvillard31 (changes requested):

PR #11299, opened 2026-05-31
NIFI-15985 - Add per-document Index Field and Timestamp Field extraction to PutElasticsearchJson

agturley:
# Summary

[NIFI-15985](https://issues.apache.org/jira/browse/NIFI-15985)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
@@ -66,6 +66,7 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -254,12 +255,12 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .description("""
                     Whether to keep the Identifier Field in the document body after extracting it \
                     for use as the Elasticsearch document ID. \
-                    When false (default), the field is removed from the document before indexing.\
+                    When true (default), the field is left in the document; set to false to remove it before indexing.\
                     """)
             .required(true)
             .allowableValues("true", "false")
-            .defaultValue("false")
-            .dependsOn(INPUT_FORMAT, InputFormat.NDJSON, InputFormat.JSON_ARRAY)
+            .defaultValue("true")
+            .dependsOn(IDENTIFIER_FIELD)
             .build();
 
     static final PropertyDescriptor INDEX_FIELD = new PropertyDescriptor.Builder()
@@ -279,11 +280,12 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .description("""
                     Whether to keep the Index Field in the document body after extracting it \
                     for use as the Elasticsearch index name. \
-                    When false (default), the field is removed from the document before indexing.\
+                    When true (default), the field is left in the document; set to false to remove it before indexing.\
                     """)
             .required(true)
             .allowableValues("true", "false")
-            .defaultValue("false")
+            .defaultValue("true")
+            .dependsOn(INDEX_FIELD)
             .build();
 
     static final PropertyDescriptor TIMESTAMP_FIELD = new PropertyDescriptor.Builder()
@@ -303,11 +305,12 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .description("""
                     Whether to keep the Timestamp Field in the document body after copying its \
                     value to @timestamp. \
-                    When false (default), the field is removed from the document before indexing.\
+                    When true (default), the field is left in the document; set to false to remove it before indexing.\
                     """)
             .required(true)
             .allowableValues("true", "false")
-            .defaultValue("false")
+            .defaultValue("true")
+            .dependsOn(TIMESTAMP_FIELD)
             .build();
 
     static final Relationship REL_BULK_REQUEST = new Relationship.Builder()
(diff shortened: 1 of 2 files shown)


nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -245,6 +249,67 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
...
+    static final PropertyDescriptor RETAIN_IDENTIFIER_FIELD = new PropertyDescriptor.Builder()
+            .name("Retain Identifier Field")
+            .description("""
+                    Whether to keep the Identifier Field in the document body after extracting it \
+                    for use as the Elasticsearch document ID. \
+                    When false (default), the field is removed from the document before indexing.\
+                    """)
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
```
pvillard31: Today the Identifier Field is kept in the indexed document. Should Retain Identifier Field default to true so existing flows keep the same document shape after upgrade?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -245,6 +249,67 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
...
+            .name("Retain Index Field")
+            .description("""
+                    Whether to keep the Index Field in the document body after extracting it \
+                    for use as the Elasticsearch index name. \
+                    When false (default), the field is removed from the document before indexing.\
+                    """)
+            .required(true)
+            .allowableValues("true", "false")
+            .defaultValue("false")
+            .build();
```
pvillard31, line 289: Should Retain Index Field use dependsOn(INDEX_FIELD), and Retain Timestamp Field use dependsOn(TIMESTAMP_FIELD), so the toggle is hidden when no source field is set?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -450,17 +526,36 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
...
+                                    final Map<String, Object> contentMap = mapReader.readValue(trimmedLine);
+                                    id = resolveId(contentMap, documentIdField, flowFileIdAttribute);
+                                    docIndex = resolveIndex(contentMap, documentIndexField, index);
+                                    if (stripId) {
+                                        contentMap.remove(documentIdField);
+                                    }
+                                    if (stripIdx) {
+                                        contentMap.remove(documentIndexField);
+                                    }
+                                    applyTimestamp(contentMap, documentTimestampField, retainTimestampField);
```
pvillard31, line 555: Can we add PutElasticsearchJsonTest coverage for the new Index Field, Timestamp Field, and the three Retain toggles across NDJSON, JSON Array, and Single JSON, so the extraction and field stripping behavior is locked in?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -931,6 +1131,41 @@ private String resolveId(final Map<String, Object> contentMap, final String idAt
...
+     * where the Map is already available. Falls back to {@code fallbackIndex} when the
+     * field is absent or blank.
+     */
+    private String resolveIndex(final Map<String, Object> contentMap, final String indexField, final String fallbackIndex) {
+        if (StringUtils.isBlank(indexField)) {
+            return fallbackIndex;
+        }
+        final Object indexObj = contentMap.get(indexField);
+        if (indexObj != null) {
+            final String value = indexObj.toString();
```
pvillard31: resolveIndex uses toString() while extractIndex uses asText(). Could a non-string index field value produce a different index name between the Map path and the JsonNode path?

pvillard31 (changes requested):

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: agreed, switched them all to default to true for consistency as well.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: Added unit tests covering Index Field, Timestamp Field, and all three Retain toggles across NDJSON, JSON Array, and Single JSON — extraction, retain-by-default, stripping when Retain=false, and fallback to the Index property when the field is absent.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: They match for normal strings but I found an actual bug while looking - if the field value is null, the NDJSON path was using the literal string "null" as the index, while the other two paths fell back correctly. Pulled all the id/index extraction through a couple of shared helpers so they behave the same: scalars use their text value, null/objects/arrays fall back. Added a couple tests for a numeric value and a null value to cover it.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: Added. Each retain flag now depends on its field, so they're hidden until you set the field. Turned out onTrigger was also reading those properties even when the field wasn't set, which isn't allowed once a dependsOn is in place, so I fixed those reads too.

pvillard31:
Latest LGTM but you need to rebase on top of main for Github checks to execute

pvillard31 (approved):
Thanks, LGTM @agturley

PR #11303, opened 2026-06-02
NIFI-15888 Add size limit to Standard Content Viewer

exceptionfactory:
# Summary

[NIFI-15888](https://issues.apache.org/jira/browse/NIFI-15888) Adds an internal size limit of 10 MB to the Standard Content Viewer, avoiding attempts to load, format, and render large FlowFiles.

The internal limit of 10 MB aligns with a previous comment on Apache Avro formatting, which actually limited to the output size to 2 MB.

The implementation uses the `LimitingInputStream` to avoid reading more than 10 MB from the referenced FlowFile. The limit applies to both formatted and unformatted responses. Unformatted responses with content exceeding the limit return an `HTTP 206` status indicating Partial Content returned. Formatted may throw an exception when the limit does not align with expected structural boundaries. In this case, the server returns an HTTP 500 with message indicating length exceeds maximum. As a best effort viewer, returning an error avoids both server and client memory consumption concerns. The hexadecimal viewer continues to provide a general way to view initial bytes.

Additional unit tests exercise length limiting behavior with formatted and unformatted responses.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [X] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/src/main/java/org/apache/nifi/web/controller/StandardContentViewerController.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/src/main/java/org/apache/nifi/web/controller/StandardContentViewerController.java
@@ -28,6 +28,7 @@ import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.io.DatumReader;
 import org.apache.nifi.authorization.AccessDeniedException;
+import org.apache.nifi.stream.io.LimitingInputStream;
 import org.apache.nifi.web.ContentAccess;
 import org.apache.nifi.web.ContentRequestContext;
 import org.apache.nifi.web.DownloadableContent;
@@ -45,6 +46,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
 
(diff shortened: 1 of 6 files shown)


nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/src/main/java/org/apache/nifi/web/controller/StandardContentViewerController.java
```diff
@@ -80,38 +92,47 @@ public void doGet(final HttpServletRequest request, final HttpServletResponse re
...
+            final ContentType formattedContentType = getFormattedContentType(request, downloadableContent.getType());
+            if (formattedContentType == null) {
+                response.sendError(HttpURLConnection.HTTP_NOT_ACCEPTABLE, "Unknown Content Type");
+            } else {
+                final String dataUri = requestContext.getDataUri();
+                writeContentFormatted(dataUri, formattedContentType, contentStream, response);
+                response.setStatus(HttpServletResponse.SC_OK);
+            }
+        } else {
             contentStream.transferTo(response.getOutputStream());
```
pvillard31, line 112: For content larger than the response buffer, the response commits during transferTo, so does the later setStatus(SC_PARTIAL_CONTENT) actually reach the client, or does it silently return 200 with truncated content?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/src/test/java/org/apache/nifi/web/controller/StandardContentViewerControllerTest.java
```diff
@@ -67,8 +75,39 @@ class StandardContentViewerControllerTest {
...
+        final DownloadableContent downloadableContent = new DownloadableContent(FILENAME, APPLICATION_OCTET_STREAM, contentStream);
+        setDownloadableContent(downloadableContent);
+
+        final MockServletOutputStream mockServletOutputStream = new MockServletOutputStream();
+        when(response.getOutputStream()).thenReturn(mockServletOutputStream);
+
+        final LimitedStandardContentViewerController limitedController = new LimitedStandardContentViewerController();
+        limitedController.doGet(request, response);
+
+        verify(response).setStatus(eq(HttpServletResponse.SC_PARTIAL_CONTENT));
```
pvillard31, line 106: Since MockServletOutputStream never commits the response, does this test really verify that a real container would send 206 after the body is written?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/src/main/java/org/apache/nifi/web/controller/StandardContentViewerController.java
exceptionfactory: Good catch, the test behavior masked this important detail. I adjusted the approach to get the expected Content Length from the source, and used that to get and set the response status prior to handling the stream.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-content-viewer/src/test/java/org/apache/nifi/web/controller/StandardContentViewerControllerTest.java
exceptionfactory: Thanks for pointing this out, I reworked the implementation to set the response status earlier.

exceptionfactory:
Thanks for the review @pvillard31, good catch on the order of setting the response status.

I made some adjustments to get the source content length and pass it to the `DownloadableContent` object, enabling determination of partial content status prior to stream handling.

The `LimitingInputStream` still enforces the limit, even if the content length is something unknown, like `-1`, but the response status determination provides a best-effort approach to indicate whether truncation occurred.

pvillard31 (approved):
Thanks @exceptionfactory - latest LGTM

PR #11304, opened 2026-06-02
NIFI-15987: Updating the global and connector specific policy managem…

mcgilman:
…ent pages to include policies for view/write data and viewing provenance events.

## Summary

Extends NiFi UI access policy management for **connectors** at global and per-instance scope, aligning with backend authorizable paths for connector data and provenance.

**Global policies** (`/access-policies`):
- Connectors resource supports extended actions: view/modify connectors, view/modify data (`/data/connectors`), view provenance (`/provenance-data/connectors`)
- Connectors use a dedicated UI branch (not the generic read/write-only path) with inherited-policy templates for global connector, data, and provenance policies

**Per-instance policies** (component access policies):
- Dedicated `connectors` branch: data and provenance actions enabled; site-to-site disabled
- `ComponentType.Connector` context with inherited templates for `/connectors`, `/data/connectors`, and `/provenance-data/connectors`

**API resource path fix** (`AccessPolicyService`):
- UI `{ resource: 'connectors', resourceIdentifier: 'data' }` maps to `/data/connectors` (not `/connectors/data`)
- Same for `provenance-data` → `/provenance-data/connectors`
- `buildResourcePath()` centralizes path construction for create/get and policy status checks

**Found vs Inherited** (`access-policy.effects.ts`):
- Compares API `component.resource` to `buildResourcePath()` so connector data/provenance policies resolve correctly (including legacy wrong paths like `/connectors/data` showing as Inherited)

**Users → Access Policies listing**:
- Human-readable labels for global `/data/connectors` and `/provenance-data/connectors` policies
- Per-instance connector policy labels and deep links to component policies

**Shared UI**:
- `ComponentContext` and `component-type-name` pipe support for `ComponentType.Connector` (plug icon, “Connector” label)
- `component-access-policies` selector corrected to `component-access-policies`

**Queue listing (bundled fix)**:
- On submit/poll API failure, dispatch `queueListingApiError`, close dialogs, and stop polling so the progress dialog does not stay open indefinitely

## Testing

- [x] `nx test nifi` — access policy service/effects specs, global/component access-policies specs, user-access-policies specs, component-context spec
- [ ] **Global policies**: Create/view policies for Connectors → read/write, read/write data, read provenance data; verify API paths `/connectors`, `/data/connectors`, `/provenance-data/connectors`
- [ ] **Per-instance**: Open access policies for a connector instance; create inherited/overrides for view connector, view/modify data, view provenance; confirm site-to-site is not offered
- [ ] **Users listing**: Confirm global connector data/provenance policies show correct labels (not generic “access connectors”)
- [ ] **Found/Inherited**: Policies at correct paths show **Found**; policies created earlier at wrong paths (e.g. `/connectors/data`) show **Inherited** until recreated
- [ ] **Queue listing**: Trigger a queue listing failure; confirm error surfaces and progress dialog closes

## Notes for reviewers

**Policy migration:** Policies created before this path transform used incorrect resources (e.g. `/connectors/data`). The UI does not migrate them; delete and recreate at the correct paths if needed.

**Connector canvas provenance preview:** Preview on the connector canvas is gated by **global** “query provenance” (`/provenance` / `currentUser.provenancePermissions.canRead`), not by `/provenance-data/connectors`. Granting only connector provenance policies does not enable the preview control.

**Queue listing change:** Included in this PR because it was discovered during connector policy QA; behavior is unrelated to access policies but fixes a stuck-dialog regression.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/access-policies/service/access-policy.service.spec.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/access-policies/service/access-policy.service.spec.ts
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { TestBed } from '@angular/core/testing';
+import { provideHttpClient } from '@angular/common/http';
+import { provideHttpClientTesting } from '@angular/common/http/testing';
+import { AccessPolicyService } from './access-policy.service';
+import { Action, ResourceAction } from '../state/shared';
+import { Client } from '../../../service/client.service';
+import { ClusterConnectionService } from '../../../service/cluster-connection.service';
+
+describe('AccessPolicyService', () => {
+    let service: AccessPolicyService;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({
+            providers: [
+                AccessPolicyService,
+                provideHttpClient(),
+                provideHttpClientTesting(),
+                {
+                    provide: Client,
+                    useValue: {
+                        getClientId: () => 'client-id',
+                        getRevision: (entity: { revision: unknown }) => entity.revision
+                    }
+                },
+                {
+                    provide: ClusterConnectionService,
+                    useValue: { isDisconnectionAcknowledged: () => false }
+                }
+            ]
+        });
+
+        service = TestBed.inject(AccessPolicyService);
+    });
+
+    describe('buildResourcePath', () => {
+        function buildPath(resourceAction: ResourceAction): string {
+            return service.buildResourcePath(resourceAction);
+        }
+
+        it('should transform connectors data policy to /data/connectors', () => {
+            expect(
+                buildPath({
+                    resource: 'connectors',
+                    resourceIdentifier: 'data',
+                    action: Action.Read
+                })
+            ).toBe('/data/connectors');
+        });
+
+        it('should transform connectors provenance-data policy to /provenance-data/connectors', () => {
+            expect(
+                buildPath({
+                    resource: 'connectors',
+                    resourceIdentifier: 'provenance-data',
+                    action: Action.Read
+                })
+            ).toBe('/provenance-data/connectors');
+        });
+
+        it('should build global connectors resource path', () => {
+            expect(
+                buildPath({
+                    resource: 'connectors',
+                    action: Action.Read
+                })
+            ).toBe('/connectors');
+        });
+
+        it('should build per-instance connector resource path', () => {
+            expect(
+                buildPath({
+                    resource: 'connectors',
+                    resourceIdentifier: 'connector-1',
+                    action: Action.Write
+                })
+            ).toBe('/connectors/connector-1');
+        });
+
+        it('should pass through non-connector resources unchanged', () => {
+            expect(
+                buildPath({
+                    resource: 'processors',
+                    resourceIdentifier: 'processor-1',
+                    action: Action.Read
+                })
+            ).toBe('/processors/processor-1');
+        });
+    });
+});
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/access-policies/service/access-policy.service.spec.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/access-policies/service/access-policy.service.spec.ts
@@ -17,7 +17,7 @@
 
 import { TestBed } from '@angular/core/testing';
 import { provideHttpClient } from '@angular/common/http';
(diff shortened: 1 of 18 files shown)


rfellows:
reviewing...

nifi-frontend/src/main/frontend/libs/shared/src/components/component-context/component-context.component.ts
```diff
@@ -63,8 +63,14 @@ export class ComponentContext {
...
+            case ComponentType.Connector:
+                return 'fa fa-plug primary-color';
             default:
                 return 'icon-drop';
         }
     }
+
+    usesFontAwesomeIcon(): boolean {
+        return this.componentIconClass.startsWith('fa');
+    }
```
rfellows: This feels a bit wrong. Since the `getIconClassName` is private, we can refactor it without risk of breaking anything else. I'd suggest changing this to return some object structure rather than just the classname string. Maybe something like:

```typescript
interface IconMeta {
    className: string;
    iconType: 'font-awesome' | 'icon-font';
}
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/access-policies/service/access-policy.service.spec.ts
rfellows, line 1: Maybe add a couple more tests that ensure the two public methods actually consume `buildResourcePath` correctly:

```typescript
it('createAccessPolicy POSTs with canonical /data/connectors resource', () => {
    const httpMock = TestBed.inject(HttpTestingController);

    service.createAccessPolicy({
        resource: 'connectors',
        resourceIdentifier: 'data',
        action: Action.Read
    }).subscribe();

    const req = httpMock.expectOne('../nifi-api/policies');
    expect(req.request.method).toBe('POST');
    expect(req.request.body.component.resource).toBe('/data/connectors');
    expect(req.request.body.component.resourceIdentifier).toBeUndefined();
    req.flush({});
    httpMock.verify();
});

it('getAccessPolicy GETs the canonical /data/connectors URL', () => {
    const httpMock = TestBed.inject(HttpTestingController);

    service.getAccessPolicy({
        resource: 'connectors',
        resourceIdentifier: 'data',
        action: Action.Read
    }).subscribe();

    httpMock.expectOne('../nifi-api/policies/read/data/connectors').flush({});
    httpMock.verify();
});
```

rfellows (changes requested):
A couple of things...

nifi-frontend/src/main/frontend/libs/shared/src/components/component-context/component-context.component.html
```diff
@@ -18,7 +18,11 @@
 <div class="component-context flex flex-col">
     <div class="flex gap-x-1 items-center">
         <div class="context-logo text-3xl flex flex-col">
-            <i class="icon tertiary-color" [class]="componentIconClass"></i>
+            @if (componentIcon.iconType === 'font-awesome') {
+                <i [class]="componentIcon.className"></i>
+            } @else {
+                <i class="icon tertiary-color" [class]="componentIcon.className"></i>
+            }
```
rfellows, line 25: This is a bit off still vertically and lacks spacing between the icon and the text.

**font-awesome:** 
Image

**flowfont:** 
Image

Also, the font-awesome icon gets its color from the method adding the class, but the others inline it right here in the dom. we should just follow that... and for that matter we could add the `fa` here too since we know these are font-awesome. Then, the method can just return the class name needed and not any extras...

```suggestion
            @if (componentIcon.iconType === 'font-awesome') {
                <i class="fa mr-2 tertiary-color" [class]="componentIcon.className"></i>
            } @else {
                <i class="icon tertiary-color" [class]="componentIcon.className"></i>
            }

```

nifi-frontend/src/main/frontend/libs/shared/src/components/component-context/component-context.component.ts
```diff
@@ -42,29 +49,31 @@ export class ComponentContext {
...
-                return 'icon-funnel';
+                return { className: 'icon-funnel', iconType: 'flowfont' };
             case ComponentType.Label:
-                return 'icon-label';
+                return { className: 'icon-label', iconType: 'flowfont' };
             case ComponentType.RemoteProcessGroup:
-                return 'icon-group-remote';
+                return { className: 'icon-group-remote', iconType: 'flowfont' };
+            case ComponentType.Connector:
+                return { className: 'fa fa-plug tertiary-color', iconType: 'font-awesome' };
```
rfellows: based on my comment above, we can just do this here:
```suggestion
                return { className: 'fa-plug', iconType: 'font-awesome' };
```

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/queue/state/queue-listing/queue-listing.effects.ts
rfellows, line 1: Can we add some tests here? I realize there were no tests prior, but I think now would be a good time to add some.

rfellows (changes requested):

nifi-frontend/src/main/frontend/libs/shared/src/components/component-context/component-context.component.html
```diff
@@ -17,11 +17,11 @@
...
 <div class="component-context flex flex-col">
     <div class="flex gap-x-1 items-center">
-        <div class="context-logo text-3xl flex flex-col">
+        <div class="context-logo text-3xl flex items-center">
             @if (componentIcon.iconType === 'font-awesome') {
-                <i [class]="componentIcon.className"></i>
+                <i class="fa mr-2 tertiary-color" [class]="componentIcon.className"></i>
             } @else {
-                <i class="icon tertiary-color" [class]="componentIcon.className"></i>
+                <i class="icon mr-2 tertiary-color" [class]="componentIcon.className"></i>
```
rfellows: we don't want this `mr-2` here. the styling from the icon class adds some spacing. we only needed it on the other one since it doesn't need/get the inherited spacing that the others do.

rfellows (approved):
Thanks @mcgilman 👍 
I'll merge it once it passes CI.

PR #11313, opened 2026-06-04
NIFI-15989: Improve purge threshold time formatting

ohnoitsyou:
Update the purge threshold output to be more human readable.

Existing output will show the value in milliseconds which can be
  difficult to read quickly. Convert milliseconds to a `Duration` to get the
  threshold in a format like "30 days", "7 days 4 hours", etc.

# Summary
[NIFI-15989](https://issues.apache.org/jira/browse/NIFI-15989)

Specifically in WriteAheadStorePartition, the maintenance task outputs the threshold as a measure of milliseconds which can be difficult to parse quickly.

Pass the value of the field `olderThan` that's passed to `purgeOldEvents` through a new formatting function in `nifi-utils` to convert it into a words representation. e.g.:

* 2592000000 -> "30 days 0 hours 0 minutes 0 seconds 0 ns"
* 2592010000 -> "30 days 0 hours 0 minutes 10 seconds 0 ns"
* 3600000 -> "1 hour 0 minutes 0 seconds 0 ns"

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java
+++ b/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java
@@ -336,19 +336,19 @@ public class FormatUtils {
         final List<String> parts = new ArrayList<>();
 
         if (days > 0) {
-            parts.add(days + (days == 1 ? " day" : " days"));
+            parts.add(days + "d");
         }
         if (hours > 0 || !parts.isEmpty()) {
-            parts.add(hours + (hours == 1 ? " hour" : " hours"));
+            parts.add(hours + "h");
         }
         if (minutes > 0 || !parts.isEmpty()) {
-            parts.add(minutes + (minutes == 1 ? " minute" : " minutes"));
+            parts.add(minutes + "m");
         }
         if (seconds > 0 || !parts.isEmpty()) {
-            parts.add(seconds + (seconds == 1 ? " second" : " seconds"));
+            parts.add(seconds + "s");
         }
         if (nanos > 0 || !parts.isEmpty()) {
-            parts.add(nanos + " ns");
+            parts.add(nanos + "ns");
         }
 
         return String.join(" ", parts);
--- a/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java
+++ b/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java
@@ -20,6 +20,7 @@ import org.apache.nifi.processor.DataUnit;
 import org.apache.nifi.time.DurationFormat;
 
 import java.text.NumberFormat;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
@@ -28,8 +29,12 @@ import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
+import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAccessor;
 import java.time.temporal.TemporalQueries;
+import java.time.temporal.TemporalUnit;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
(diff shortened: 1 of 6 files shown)


exceptionfactory (changes requested):
Thanks for proposing this improvement @ohnoitsyou.

Although other modules include the `commons-lang3` dependency, introducing this dependency for the sole purpose of improving the logging message does not seem worth the addition.

ohnoitsyou:
I had considered adding a function to the FormatUtils located in commons. Unfortunately it also doesn't have lang3 in it, but it would centralize the functionality.

The end user this request came from only mentioned this particular log. I'm sure there are other places it would make sense to have a log message formatted in this way.

exceptionfactory:
In relation to log formatting, additional dependencies should be avoided.

There is certainly a balance to strike between readability and consistency, but in this case, the log message already includes the number of units, which also makes it more consistent for potential parsing.

With that background, this doesn't seem like a good candidate for changing.

dan-s1:
@exceptionfactory Could the `org.apache.nifi.time.DurationFormat `found in the nifi-api be used for this? Or does this also have the limitation that the nifi-api is not currently a dependency?

ohnoitsyou:
@dan-s1 Unfortunately `DurationFormat` doesn't include a string based output.

dan-s1:
> @dan-s1 Unfortunately `DurationFormat` doesn't include a string based output.

Yeah I realize. I thought it might have been included as a dependency and you would have to still add some code to produce a string output.

ohnoitsyou:
@exceptionfactory If we still wanted to pursue something like this where the message is more expressive about the time frame, is there something you might suggest?

Obviously this in a single place doesn't make a huge impact, but I bet if I did some digging I could find lots of places where having a large number of millis formatted in words would improve readability.

mosermw:
@ohnoitsyou I think it would be acceptable to have a new method in the nifi-utils.jar FormatUtils class that does this.  Perhaps named `formatTimeDuration`, it would produce a String containing the largest "whole number with TimeUnit" that a duration fits in.  It's a concept similar to the existing FormatUtils `formatDataSize` method.

To get an idea how to implement it, check out the nifi-api DurationFormat class.  It has some private methods that do something similar.

exceptionfactory:
Just a caution on adding some new method, although it might be useful, I'm not sure it would make sense to change log messages. It might in very specific cases, but it would be a case-by-case question.

mosermw:
If this specific log message should not be changed, then we should close this PR.  However, I will say that I have also wondered if this specific log message could be improved to make it more human readable.

exceptionfactory:
Taking a fresh look at this, I could see adjusting the log message to include both the original number and units, plus something descriptive, immediately after, in parentheses. This kind of approach is similar to displaying seconds and nanoseconds on application startup. That would still require a new method in `nifi-utils`, but I could see that as a way forward.

ohnoitsyou:
@exceptionfactory @mosermw I took a stab at implementing a duration to words function in nifi-utils.

I also converted the `EventStorePartition` interface from from `TimeUnit` to `ChronoUnit` partly because `ChronoUnit` is part of the `java.util.temporal` package vs `java.util.concurrent` and it felt more appropriate for dealing with time. If that's not OK, I can go back.

ohnoitsyou:
@exceptionfactory Not to bug too much, but if this could be reviewed, I found a place where the formatting function would be useful in another related class and I'd like to take advantage of it.

nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java
```diff
@@ -298,4 +303,54 @@ public static Instant parseToInstant(final DateTimeFormatter formatter, final St
...
+     * Format a value of ChronoUnit into a word representation.
+     * Does not handle estimated ChronoUnit values.
+     * For anything greater than {@link ChronoUnit#DAYS}, use {@link #formatDurationToWords(Duration)}.
+     *
+     * @param value the number of the given {@link ChronoUnit} to measure
+     * @param unit {@link ChronoUnit} that the value represents
+     * @return String representation of the given value and unit
+     * @throws UnsupportedTemporalTypeException if the unit is not supported ({@link ChronoUnit#isDurationEstimated()} == true)
+     */
+    public static String formatDurationToWords(long value, ChronoUnit unit) throws UnsupportedTemporalTypeException {
```
exceptionfactory: The arguments should have the `final` keyword.

nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/FormatUtils.java
```diff
@@ -298,4 +303,54 @@ public static Instant parseToInstant(final DateTimeFormatter formatter, final St
...
+     * @return String representation of the given duration
+     */
+    public static String formatDurationToWords(Duration source) {
+        long days = source.toDaysPart();
+        long hours = source.toHoursPart();
+        long minutes = source.toMinutesPart();
+        int seconds = source.toSecondsPart();
+        int nanos = source.toNanosPart();
+
+        List<String> parts = new ArrayList<>();
```
exceptionfactory: The `final` keyword should be used across this method.

exceptionfactory (changes requested):
Thanks for continuing the work on this @ohnoitsyou. The new format methods look good in general, I noted a few stylistic recommendations, but this looks close to completion.

ohnoitsyou:
@exceptionfactory Got those updated. If I could get the workflows approved, hopefully we can push this over the finish line.

exceptionfactory (commented):
Thanks for the work on this @ohnoitsyou. It looks close to completion, just one general question on formatting.

The spelled out words make the description rather long (`1 day 2 hours 10 minutes`). What do you think about just using single letters like `1d 2h 10m 30s 100ns`? It is still readable, but less verbose, and doesn't need the pluralization.

exceptionfactory:
The other approach would be to implement some kind of rounding to the greatest two units (days and hours, hours and minutes, etc), but the single-letter abbreviation might be a way to keep the precision with readability.

ohnoitsyou:
Personally, I'd rather abbreviate and be more precise than truncate data. It's still readable while being concise.

Edit: that is to say, I'm good with the change.

exceptionfactory:
> Personally, I'd rather abbreviate and be more precise than truncate data. It's still readable while being concise.
> 
> Edit: that is to say, I'm good with the change.

Thanks, if you like the abbreviation approach, that sounds good to me.

ohnoitsyou:
@exceptionfactory Pushed that change, builds on my end. If you could approve the workflows, that'd be appreciated. Trying to keep it up-to-date with main

exceptionfactory:
Thanks @ohnoitsyou, the latest set of changes look good. I rebased the branch and plan to merge on successful build completion

exceptionfactory (approved):
Thanks for working through this and moving it to completion @ohnoitsyou! +1 merging

PR #11314, opened 2026-06-05
NIFI-15911 Add proxy support to StandardSnowflakeIngestManagerProviderService

rakesh-rsky:
NIFI-15911 Add proxy support to StandardSnowflakeIngestManagerProviderService

StandardSnowflakeIngestManagerProviderService uses SnowpipeIngestClient to send data to Snowflake via the Snowpipe REST API. Previously, the HTTP client had no proxy configuration, causing failures for NiFi deployments running behind a corporate or network proxy.

This change adds proxy support by wiring ProxyConfigurationService into the controller service.

Changes:
- Added PROXY_CONFIGURATION_SERVICE property to StandardSnowflakeIngestManagerProviderService
- Used a custom ProxySelector that correctly handles HTTP, SOCKS, and DIRECT proxy types
- Added scoped Authenticator for proxy username/password credentials
- Extended SnowpipeIngestClient with a constructor accepting ProxySelector and Authenticator

SnowpipeIngestClient

 - Added a constructor accepting ProxySelector and Authenticator, applied to the underlying HttpClient
 - Existing 3-arg constructor delegates to the new one with nulls (backward compatible)

# Summary

[NIFI-15911](https://issues.apache.org/jira/browse/NIFI-15911)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
+++ b/nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
@@ -22,6 +22,8 @@ import org.apache.nifi.annotation.documentation.Tags;
 import org.apache.nifi.annotation.lifecycle.OnDisabled;
 import org.apache.nifi.annotation.lifecycle.OnEnabled;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.ValidationContext;
+import org.apache.nifi.components.ValidationResult;
 import org.apache.nifi.controller.AbstractControllerService;
 import org.apache.nifi.controller.ConfigurationContext;
 import org.apache.nifi.expression.ExpressionLanguageScope;
@@ -32,14 +34,24 @@ import org.apache.nifi.processors.snowflake.SnowflakeIngestManagerProviderServic
 import org.apache.nifi.processors.snowflake.snowpipe.InsertFiles;
 import org.apache.nifi.processors.snowflake.snowpipe.InsertReport;
 import org.apache.nifi.processors.snowflake.util.SnowflakeProperties;
+import org.apache.nifi.proxy.ProxyConfiguration;
+import org.apache.nifi.proxy.ProxySpec;
 import org.apache.nifi.reporting.InitializationException;
 import org.apache.nifi.snowflake.service.util.AccountIdentifierFormat;
 import org.apache.nifi.snowflake.service.util.AccountIdentifierFormatParameters;
 import org.apache.nifi.snowflake.service.util.ConnectionUrlFormat;
 
+import java.io.IOException;
+import java.net.Authenticator;
+import java.net.PasswordAuthentication;
+import java.net.Proxy;
+import java.net.ProxySelector;
+import java.net.SocketAddress;
 import java.net.URI;
 import java.security.PrivateKey;
 import java.security.interfaces.RSAPrivateCrtKey;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 
 @Tags({"snowflake", "snowpipe", "database", "connection"})
@@ -124,6 +136,19 @@ public class StandardSnowflakeIngestManagerProviderService extends AbstractContr
             .required(true)
             .build();
 
+    // Only HTTP proxy type is supported. SOCKS is silently ignored by java.net.http.HttpClient.
+    // For authenticated proxies: the JDK disables Basic auth over CONNECT tunneling by default
+    // (jdk.http.auth.tunneling.disabledSchemes=Basic). To use proxy credentials with an HTTPS
+    // target like Snowflake, add -Djdk.http.auth.tunneling.disabledSchemes="" to bootstrap.conf.
+    static final PropertyDescriptor PROXY_CONFIGURATION = new PropertyDescriptor.Builder()
+            .fromPropertyDescriptor(ProxyConfiguration.createProxyConfigPropertyDescriptor(ProxySpec.HTTP_AUTH))
+            .description("Specifies the Proxy Configuration Controller Service to proxy network requests."
+                    + " Only HTTP proxy type is supported."
+                    + " For authenticated proxy with HTTPS targets, the JDK disables Basic authentication"
+                    + " over CONNECT tunneling by default. To enable it, add"
+                    + " -Djdk.http.auth.tunneling.disabledSchemes=\"\" to bootstrap.conf.")
+            .build();
+
     static final List<PropertyDescriptor> PROPERTY_DESCRIPTORS = List.of(
             ACCOUNT_IDENTIFIER_FORMAT,
             HOST_URL,
@@ -136,7 +161,8 @@ public class StandardSnowflakeIngestManagerProviderService extends AbstractContr
             PRIVATE_KEY_SERVICE,
             DATABASE,
             SCHEMA,
-            PIPE
+            PIPE,
+            PROXY_CONFIGURATION
     );
 
     private static final String HTTPS_URI_FORMAT = "https://%s";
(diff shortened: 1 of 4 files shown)


nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
```diff
@@ -136,7 +144,8 @@ public class StandardSnowflakeIngestManagerProviderService extends AbstractContr
             PRIVATE_KEY_SERVICE,
             DATABASE,
             SCHEMA,
-            PIPE
+            PIPE,
+            ProxyConfigurationService.PROXY_CONFIGURATION_SERVICE
```
pvillard31: Should we use ProxyConfiguration.createProxyConfigPropertyDescriptor(...) here and add a customValidate that calls ProxyConfiguration.validateProxySpec(...), like other proxy aware components, so an unsupported proxy type is rejected at configuration time instead of failing silently at runtime?

nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
```diff
@@ -213,6 +228,46 @@ public void migrateProperties(PropertyConfiguration config) {
         config.renameProperty(SnowflakeProperties.OLD_SCHEMA_PROPERTY_NAME, SnowflakeProperties.SCHEMA.getName());
     }
 
+    private ProxySelector buildProxySelector(final ProxyConfigurationService proxyConfigurationService) {
+        if (proxyConfigurationService == null) {
+            return null;
+        }
+        final Proxy proxy = proxyConfigurationService.getConfiguration().createProxy();
+        // Use a custom ProxySelector so that the Proxy.Type (HTTP or SOCKS) is preserved,
```
pvillard31: The JDK java.net.http.HttpClient only applies HTTP proxies and ignores a SOCKS Proxy, so a configured SOCKS proxy would silently fall back to a direct connection. Should we drop the SOCKS handling claim or validate that only HTTP proxy types are configured?

nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
```diff
@@ -213,6 +228,46 @@ public void migrateProperties(PropertyConfiguration config) {
         config.renameProperty(SnowflakeProperties.OLD_SCHEMA_PROPERTY_NAME, SnowflakeProperties.SCHEMA.getName());
     }
 
+    private ProxySelector buildProxySelector(final ProxyConfigurationService proxyConfigurationService) {
+        if (proxyConfigurationService == null) {
+            return null;
+        }
+        final Proxy proxy = proxyConfigurationService.getConfiguration().createProxy();
```
pvillard31: Could we use ProxyConfiguration.getConfiguration(context), which returns DIRECT when the service is unset, and ProxyConfiguration.hasCredential() to simplify this code and remove the two null returning helper methods?

nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
```diff
@@ -213,6 +228,46 @@ public void migrateProperties(PropertyConfiguration config) {
...
+                return List.of(proxy);
+            }
+
+            @Override
+            public void connectFailed(final URI uri, final SocketAddress sa, final IOException ioe) {
+            }
+        };
+    }
+
+    private Authenticator buildProxyAuthenticator(final ProxyConfigurationService proxyConfigurationService) {
```
pvillard31: The Snowpipe endpoint is HTTPS, and the JDK disables Basic proxy authentication over CONNECT tunneling by default via jdk.http.auth.tunneling.disabledSchemes. Was the username and password path tested against a real authenticated HTTPS proxy?

nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/test/java/org/apache/nifi/snowflake/service/SnowpipeIngestClientTest.java
```diff
@@ -264,6 +266,31 @@ void testGetInsertReportErrorResponse() {
         assertTrue(exception.getMessage().contains(String.valueOf(HttpURLConnection.HTTP_INTERNAL_ERROR)));
     }
 
+    @Test
+    void testInsertFilesViaProxy() throws InterruptedException, NoSuchAlgorithmException {
```
pvillard31, line 273: This test uses an HTTP base URI and a null authenticator, so it does not cover HTTPS tunneling or proxy credentials, which are the parts most likely to break. Can we add coverage for the authenticated proxy and HTTPS cases?

pvillard31 (changes requested):

nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/StandardSnowflakeIngestManagerProviderService.java
rakesh-rsky: You're right. The Authenticator is the correct Java API for proxy credentials, but there's a JDK limitation — Basic authentication over CONNECT tunneling is disabled by default via jdk.http.auth.tunneling.disabledSchemes=Basic. There's no way to override this per HttpClient instance since it's a JVM-wide setting. So authenticated proxy won't work out of the box for HTTPS targets like Snowflake. To handle this, I've added a note directly in the property description telling users to add `-Djdk.http.auth.tunneling.disabledSchemes=""` to their `bootstrap.conf` if they need proxy credentials with Snowflake.

nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/test/java/org/apache/nifi/snowflake/service/SnowpipeIngestClientTest.java
rakesh-rsky: I've updated the proxy test,  it now points the base URI to a different host (http://fake-snowflake.example.com) while the proxy selector points to MockWebServer. When Java's HttpClient routes through an HTTP proxy, it sends the request target in absolute form, so we assert the recorded request starts with http://fake-snowflake.example.com. This proves the proxy selector is actually being used. For the HTTPS + credentials case as mentioned in the comment above, this hits the JDK limitation with CONNECT tunneling and would need a full TLS proxy setup to test properly, which is more of an integration test scenario. The property description now documents what users need to configure to make it work.

exceptionfactory (changes requested):
Thanks for proposing this improvement @rakesh-rsky.

On initial review, it seems like a better implementation path would be the addition of the `WebClientServiceProvider`, which supports Proxy Server configuration and other HTTP properties. That approach may avoid the localized complexity of proxy configuration for this particular component. As this is a more involved change, I may look at the implementation in the near future.

exceptionfactory:
Thanks for the initial work on this pull request @rakesh-rsky.

After further consideration, I prepared an alternative pull request #11369 that introduces the `Web Client Service Provider`. Since the standard implementation already supports Proxy Server configuration, and also provides other configuration features, this approach aligns with similar implementation strategies for other components.

PR #11315, opened 2026-06-05
NIFI-16000: Allow spaces in FileUtils.getSanitizedFilename

mcgilman:
Remove the space character from the invalid-character set so interior spaces are preserved, and normalize the result by collapsing whitespace runs to a single space and stripping leading/trailing whitespace and trailing dots. This lets the asset-upload callers accept common valid filenames such as "driver (1).jar" while still rejecting non-canonical names. Add TestFileUtils covering the sanitization contract.

# Summary

[NIFI-16000](https://issues.apache.org/jira/browse/NIFI-16000)

`FileUtils.getSanitizedFilename(String)` treats the space character (code point `32`) as invalid and replaces it with an underscore. The space character is legal on every major file system (NTFS, ext4, APFS, etc.), so this is stricter than necessary.

This matters because of how the method is consumed. Both `ConnectorResource` and `ParameterContextResource` use it as a strict validation gate for the asset name supplied in the `Filename` request header — they sanitize the supplied name and reject the request if the sanitized value differs from the original:

```java
final String sanitizedAssetName = FileUtils.getSanitizedFilename(assetName);
if (!assetName.equals(sanitizedAssetName)) {
    throw new IllegalArgumentException(FILENAME_HEADER + " header contains an invalid file name");
}
```

Because any name containing a space is rewritten during sanitization, the equality check fails and the upload is rejected. As a result, common valid filenames cannot be uploaded as assets — e.g. a file produced by browser/OS download de-duplication such as `driver (1).jar` is sanitized to `driver_(1).jar` and rejected with *"... header contains an invalid file name."*

## Changes

- Removed the space character (`32`) from the invalid-character set so spaces are preserved rather than replaced.
- Spaces are kept exactly as supplied (leading, trailing, repeated, and interior); no other normalization is performed. All other characters continue to be sanitized as before.
- Added `TestFileUtils` covering the sanitization contract (null/empty, invalid-character replacement, spaces preserved, dots preserved).

The change is backward compatible: any filename that contained no spaces is sanitized exactly as before. The only behavioral change is that the space character is now preserved instead of replaced, so filenames whose sole issue was a space are now accepted by the asset-upload callers instead of being rejected.

# Tracking

### Issue Tracking

### Pull Request Tracking

- [ ] Pull Request commit message starts with Apache NiFi Jira issue number, as described in the issue tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/file/FileUtils.java
+++ b/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/util/file/FileUtils.java
@@ -581,8 +581,10 @@ public class FileUtils {
 
     // The invalid character list is derived from this Stackoverflow page.
     // https://stackoverflow.com/questions/1155107/is-there-a-cross-platform-java-method-to-remove-filename-special-chars
+    // The space character (32) is intentionally omitted: spaces are legal on every major file system, so they are
+    // preserved rather than replaced.
     private static final int[] INVALID_CHARS = {34, 60, 62, 124, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
-            16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 58, 42, 63, 92, 47, 32};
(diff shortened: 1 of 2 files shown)


exceptionfactory (approved):
Thanks for making this adjustment to allow spaces in filenames @mcgilman, the implementation changes look good. I'm going to removed the `public` modifiers from the test class on merge.

PR #11317, opened 2026-06-08
NIFI-16005 Adjust proxy header validation for requests forwarded to c…

bbende:
…luster coordinator

# Summary

[NIFI-16005](https://issues.apache.org/jira/browse/NIFI-16005)

Adds a new header `request-forwarded-to-coordinator` that is set when a node forwards a request to the cluster coordinator, similar to `request-replicated` which is set when the coordinator replicates requests to the other nodes. The `ProxyHeaderValidatorCustomizer` was then updated to skip validation when the request was forwarded to the coordinator, similar to how it skips when the coordinator replicated to other nodes.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


kevdoran (approved):
Given that this new header is additive to the existing bypass mechanics, I am approving it. 

I do think it is worth looking at other request replication call sites to strip this new header as an added defensive layer, but given that the original `request-replicated` header is already stripped or overwritten at all call-sites, that is not a blocker.

PR #11320, opened 2026-06-09
NIFI-16003 Re-copy checkpoints before deleting lingering migration table

awelless:
# Summary

[NIFI-16003](https://issues.apache.org/jira/browse/NIFI-16003)

When a prior legacy-checkpoint migration failed mid-flight (e.g. step 5 copy threw because DynamoDB's request router still cached the old table schema after recreation, see the ticket for more details), the migration table was left behind as the only surviving copy of the per-shard checkpoint state. On restart, `LegacyCheckpointMigrator.cleanupLingeringMigration` would unconditionally delete that table, silently losing all checkpoints and forcing every shard back to `INITIAL_STREAM_POSITION`.

Re-run the copy from the migration table into the checkpoint table before deleting it. This is safe because any failure in the rename sequence propagates out of ensureCheckpointTableExists and prevents the processor from starting, so the checkpoint table cannot have advanced past the migration table's contents — the re-copy either restores lost state or re-puts identical rows.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-kinesis/src/main/java/org/apache/nifi/processors/aws/kinesis/LegacyCheckpointMigrator.java
+++ b/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-kinesis/src/main/java/org/apache/nifi/processors/aws/kinesis/LegacyCheckpointMigrator.java
@@ -77,13 +77,20 @@ final class LegacyCheckpointMigrator {
         return null;
     }
 
-    void cleanupLingeringMigration() {
+    void completeLingeringMigration() {
         final String lingeringMigration = findMigrationTable();
         if (lingeringMigration == null) {
             return;
         }
-        logger.info("Deleting orphaned migration table [{}]; legacy checkpoint table [{}] retains original data", lingeringMigration, checkpointTableName);
-        CheckpointTableUtils.deleteTable(dynamoDbClient, logger, lingeringMigration);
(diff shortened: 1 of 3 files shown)


nifi-extension-bundles/nifi-aws-bundle/nifi-aws-kinesis/src/main/java/org/apache/nifi/processors/aws/kinesis/LegacyCheckpointMigrator.java
```diff
@@ -77,13 +77,20 @@ String findMigrationTable() {
...
 
-    void cleanupLingeringMigration() {
+    void completeLingeringMigration() {
         final String lingeringMigration = findMigrationTable();
         if (lingeringMigration == null) {
             return;
         }
-        logger.info("Deleting orphaned migration table [{}]; legacy checkpoint table [{}] retains original data", lingeringMigration, checkpointTableName);
-        CheckpointTableUtils.deleteTable(dynamoDbClient, logger, lingeringMigration);
+        logger.warn("Checkpoint table [{}] is in place but migration table [{}] still exists from a prior rename; not all checkpoints may have been migrated, re-copying before deletion",
```
markap14: No need to warn here, an info level message should be fine.

markap14 (commented):
Thanks for the fix @awelless - I don't think there's any need to emit a warning log, but otherwise all looks good to me.

markap14 (approved):

PR #11325, opened 2026-06-10
NIFI-16011: Reduce number of FlowFiles used in LoadBalanceIT from 100…

markap14:
… to 20 in order to avoid the excessive number of requests to the cluster in order to iterate over each FlowFile in the queue

# Summary

[NIFI-00000](https://issues.apache.org/jira/browse/NIFI-00000)

# Tracking

Please complete the following tracking steps prior to pull request creation.

### Issue Tracking

- [Apache NiFi Jira](https://issues.apache.org/jira/browse/NIFI) issue created

### Pull Request Tracking

- Pull Request title starts with Apache NiFi Jira issue number, such as `NIFI-00000`
- Pull Request commit message starts with Apache NiFi Jira issue number, as such `NIFI-00000`
- Pull request contains [commits signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) with a registered key indicating `Verified` status

### Pull Request Formatting

- Pull Request based on current revision of the `main` branch
- Pull Request refers to a feature branch with one commit containing changes

# Verification

Please indicate the verification steps performed prior to pull request creation.

### Build

- [ ] Build completed using `./mvnw clean install -P contrib-check`
  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

- [ ] New dependencies are compatible with the [Apache License 2.0](https://apache.org/licenses/LICENSE-2.0) according to the [License Policy](https://www.apache.org/legal/resolved.html)
- [ ] New dependencies are documented in applicable `LICENSE` and `NOTICE` files

### Documentation

- [ ] Documentation formatting appears as expected in rendered files

PR #11330, opened 2026-06-11
NIFI-15979 Wire Connector MDC and reporting-task visibility

kevdoran:
# Summary

[NIFI-15979](https://issues.apache.org/jira/browse/NIFI-15979)

Surface Connector observability that was part of nifi-api 2.9.0:
- ConnectorNode.getLoggingAttributes() plus StandardConnectorNode/StandardProcessGroup propagation of connector identity into descendant-component MDC.
- StandardEventAccess.getConnectorStatuses() returns a ConnectorStatus (id, name, root group status) per connector; StatusHistoryRepository gains a capture overload accepting Collection<ConnectorStatus>; FlowController invokes it.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [x] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
@@ -234,6 +234,7 @@ public final class StandardProcessGroup implements ProcessGroup {
     private static final String UNREGISTERED_PATH_SEGMENT = "UNREGISTERED";
 
     private final Map<String, String> loggingAttributes = new ConcurrentHashMap<>();
+    private volatile Map<String, String> connectorLoggingAttributes = Map.of();
     private volatile String logFileSuffix;
 
     public StandardProcessGroup(final String id, final ControllerServiceProvider serviceProvider, final ProcessScheduler scheduler,
@@ -308,6 +309,12 @@ public final class StandardProcessGroup implements ProcessGroup {
     @Override
     public void setParent(final ProcessGroup newParent) {
         parent.set(newParent);
+        // Inherit connector-supplied MDC attributes from the parent so descendants of a connector's managed
+        // flow carry the same connector metadata (attributing their logs and status metrics to the connector).
+        // Runs on every re-parent (including initial attach), so PGs added later inherit automatically.
+        if (newParent instanceof StandardProcessGroup standardParent) {
+            this.connectorLoggingAttributes = standardParent.connectorLoggingAttributes;
+        }
         setLoggingAttributes();
     }
 
@@ -4829,6 +4836,47 @@ public final class StandardProcessGroup implements ProcessGroup {
             final String registeredFlowVersion = currentVersionControl.getVersion();
             loggingAttributes.put(LoggingAttribute.REGISTERED_FLOW_VERSION.attribute, registeredFlowVersion);
         }
+
+        loggingAttributes.putAll(connectorLoggingAttributes);
+    }
+
+    /**
+     * Stores the connector-managed MDC attributes for this process group and cascades the same
+     * attributes to all descendant process groups so that components anywhere in the connector's
+     * managed flow log with consistent connectorId/connectorName/etc. context.
+     *
+     * <p>This method is called by {@code StandardConnectorNode} against its managed root process
+     * group whenever the connector's framework keys (e.g. {@code connectorName}) change or when the
+     * connector provides updated custom logging attributes. Newly created descendant groups will
+     * also inherit the attributes lazily via {@link #setParent(ProcessGroup)}.</p>
+     *
+     * @param attributes the merged set of connector logging attributes; an empty or {@code null}
+     *                   map clears any previously assigned attributes
+     */
+    @Override
+    public void setConnectorLoggingAttributes(final Map<String, String> attributes) {
+        final Map<String, String> snapshot = (attributes == null || attributes.isEmpty())
+            ? Map.of()
+            : Map.copyOf(attributes);
+        this.connectorLoggingAttributes = snapshot;
+        setLoggingAttributes();
+
+        for (final ProcessGroup child : getProcessGroups()) {
+            if (child instanceof StandardProcessGroup standardChild) {
+                standardChild.setConnectorLoggingAttributes(snapshot);
+            }
+        }
+    }
+
+    /**
+     * Returns the connector-managed MDC attributes currently assigned to this process group (inherited from
+     * its connector's managed flow root, or empty if this group is not part of a connector flow).
+     *
+     * @return an immutable map of connector logging attributes; never {@code null}
+     */
+    @Override
+    public Map<String, String> getConnectorLoggingAttributes() {
+        return connectorLoggingAttributes;
     }
 
     private void setGroupPath() {
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/components/connector/ConnectorNode.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/components/connector/ConnectorNode.java
@@ -35,6 +35,7 @@ import org.apache.nifi.logging.ComponentLog;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.Future;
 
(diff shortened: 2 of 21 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/ExtensionBuilder.java
```diff
@@ -538,11 +539,16 @@ public Resource getResource() {
             connectorValidationTrigger,
             false
         );
+        // Late-bind the logging context to the connector node so that MDC attributes assembled by
+        // the node (framework keys + provider-supplied custom keys) flow through the connector's
+        // own ComponentLog as well as the managed flow's nested components. Provider-supplied custom
+        // attributes are sourced and applied by the ConnectorRepository when the node is added/restored.
+        loggingContext.setComponent(connectorNode);
```
pvillard31, line 546: CI fails to compile here because `connectorNode` is typed as `ConnectorNode`, which does not extend `GroupedComponent` that `setComponent` requires. Should `ConnectorNode` extend `GroupedComponent`, or should this call use the `StandardConnectorNode` type?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
```diff
@@ -308,6 +309,14 @@ public ProcessGroup getParent() {
     @Override
     public void setParent(final ProcessGroup newParent) {
         parent.set(newParent);
+        // If the process group is part of a Connector flow, inherit connector-supplied MDC attributes.
```
pvillard31: These four new comment lines have trailing whitespace and fail the checkstyle check in CI. Can you remove the trailing spaces?

nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/components/connector/ConnectorNode.java
```diff
@@ -67,6 +68,25 @@ public interface ConnectorNode extends ComponentAuthorizable, VersionedComponent
 
     BundleCoordinate getBundleCoordinate();
 
+    /**
+     * @return an immutable map of the Connector's logging attributes: the framework-managed identity keys
+     *         (such as connector identifier, name, component type, and bundle coordinate) merged with any
+     *         provider-supplied custom attributes. These are the attributes applied to the MDC of the Connector's
+     *         managed flow and surfaced on Connector metrics. Never {@code null}.
+     */
+    Map<String, String> getLoggingAttributes();
```
pvillard31: Could `getLoggingAttributes()` and `setCustomLoggingAttributes(...)` be `default` methods so other implementations of `ConnectorNode` keep compiling?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
```diff
@@ -149,6 +157,103 @@ public String getName() {
...
+        }
+        rebuildLoggingAttributes();
+    }
+
+    /**
+     * Returns an immutable snapshot of the merged framework + custom logging attributes currently
+     * advertised by this connector. The framework keys are populated by the framework from the
+     * connector's identifier, name, component type, and bundle coordinate.
+     */
+    public Map<String, String> getLoggingAttributes() {
```
pvillard31, line 212: Should this method carry `@Override`, since it now implements the new `ConnectorNode.getLoggingAttributes()`?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
```diff
@@ -810,11 +810,12 @@ private FlowController(
         }
 
         eventAccess = new StandardEventAccess(flowManager, flowFileEventRepository, processScheduler, authorizer, provenanceRepository,
-                auditService, analyticsEngine, flowFileRepository, contentRepository);
+                auditService, analyticsEngine, flowFileRepository, contentRepository, connectorRepository);
 
         timerDrivenEngineRef.get().scheduleWithFixedDelay(() -> {
             try {
-                statusHistoryRepository.capture(getNodeStatusSnapshot(), eventAccess.getControllerStatus(), getGarbageCollectionStatus(), new Date());
+                statusHistoryRepository.capture(getNodeStatusSnapshot(), eventAccess.getControllerStatus(), eventAccess.getConnectorStatuses(),
```
pvillard31, line 817: Since no `StatusHistoryRepository` overrides the new `capture` overload yet, the `connectorStatuses` are dropped today; is the consuming implementation planned for a later milestone?

pvillard31 (changes requested):

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/ExtensionBuilder.java
kevdoran: Yeah this should have used StandardConnectorNode here. Fixed.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
kevdoran: Thanks, fixed.

nifi-framework-bundle/nifi-framework/nifi-framework-core-api/src/main/java/org/apache/nifi/components/connector/ConnectorNode.java
kevdoran: Done — both getLoggingAttributes() (defaults to Map.of()) and setCustomLoggingAttributes(...) (default no-op) are now default methods; StandardConnectorNode overrides both.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
kevdoran: Yep, added

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/FlowController.java
kevdoran: Great question. I was considering a follow up, but it looks trivial so I will add it to this PR.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
```diff
@@ -149,6 +157,109 @@ public String getName() {
...
+        }
+
+        pushLoggingAttributesToManagedFlow(merged);
+    }
+
+    private void pushLoggingAttributesToManagedFlow(final Map<String, String> attributes) {
+        final ProcessGroup managedProcessGroup = getProcessGroup();
+        if (managedProcessGroup instanceof StandardProcessGroup standardProcessGroup) {
+            standardProcessGroup.setConnectorLoggingAttributes(attributes);
+        }
```
markap14: We've got a few different checks going on in this PR to see if a given ProcessGroup is an instance of StandardProcessGroup. While I'm okay with that internal to `StandardProcessGroup` it seems like it's not really something we should be doing from ConnectorNode. Why not just add the get/set methods to `ProcessGroup`?

markap14 (commented):
Thanks @kevdoran I think this is helpful. The only concern that I had was the multiple checks for `instanceof StandardProcessGroup` rather than just adding the methods to `ProcessGroup`

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
kevdoran: I'll add the method to `ProcessGroup`

kevdoran:
Thanks for the feedback @markap14 - I agree, made that change

kevdoran:
Rebased on #11325 to fix the system test failures. Squashed at the same time

pvillard31 (approved):

PR #11331, opened 2026-06-11
NIFI-16006: Addressed issue where a node can be disconnected, then so…

markap14:
…on after re-connect but then be quickly told to disconnect due to a queued up 'Disconnect' message from the original disconnection. Now, we use a 'generation' flag so we know to ignore the message, and we also cancel the background task that is trying to deliver it.

# Summary

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/node/NodeClusterCoordinator.java
```diff
@@ -1297,6 +1311,19 @@ private ConnectionResponseMessage handleConnectionRequest(final ConnectionReques
         return createConnectionResponse(requestWithNodeIdentities, nodeIdentifier);
     }
 
+    private void cancelPendingDisconnection(final NodeIdentifier nodeIdentifier) {
+        final Future<Void> pendingDisconnect = pendingDisconnections.remove(nodeIdentifier.getId());
+        if (pendingDisconnect != null) {
+            final boolean cancelled = pendingDisconnect.cancel(true);
```
pvillard31: Since `CompletableFuture.cancel(true)` does not interrupt a running thread, a disconnect worker that is inside `Thread.sleep(retrySeconds * 1000L)` keeps sleeping until the next loop iteration. Should this log report the disconnect as cancelled when the worker thread may still be running?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/StandardFlowService.java
```diff
@@ -600,6 +603,11 @@ private NodeIdentifier getNodeId() {
         }
     }
 
+    // Visible for testing
+    long getConnectionGeneration() {
```
pvillard31, line 607: `getConnectionGeneration()` is marked Visible for testing, but no test seems to reference it. Is it still needed, or can it be removed?

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/StandardFlowService.java
```diff
@@ -719,32 +727,34 @@ private void offload(final String explanation) throws InterruptedException {
         }
     }
 
-    private void handleDisconnectionRequest(final DisconnectMessage request) {
+    private void handleDisconnectionRequest(final DisconnectMessage request, final long expectedConnectionGeneration) {
         logger.info("Received disconnection request message from cluster coordinator with explanation: {}", request.getExplanation());
-        disconnect(request.getExplanation());
+
+        writeLock.lock();
```
pvillard31, line 734: The new test covers the coordinator cancelling the pending disconnect, but not this node-side generation guard that ignores a stale disconnect after reconnect. Could we add a unit or system test for this path?

pvillard31 (approved):

PR #11332, opened 2026-06-11
NIFI-16014 - Remove dev-mode only feature guard gating the Connectors UI - make it available generally

rfellows:
**NIFI-16014: Remove dev-mode feature gate from the Connectors UI**

The Connectors UI was previously gated behind a build-time-only flag (`environment.enableConnectors`) that hid the route and navigation menu item in production builds. Now that the backend support is ready, the gate is removed entirely.

- Deleted `connectors-feature.guard.ts` (existed solely for this gate)
- Removed `enableConnectors` from both environment files
- Removed `connectorsFeatureGuard` from the `/connectors` route's `canMatch` array; the route now uses only `authenticationGuard`, consistent with every other top-level feature route
- Removed the `@if (enableConnectors)` wrapper from the Connectors navigation menu item; added `data-qa="global-menu-item-connectors"` for testability
- Cleaned up the now-dead `environment` import and class field from `NavigationComponent`
- Added a test asserting the Connectors menu item is unconditionally rendered

The existing `connectorsPermissions.canRead` RBAC `[disabled]` binding on the menu button is preserved and unchanged.


mcgilman:
Reviewing...

mcgilman (approved):
Thanks for the PR @rfellows!

PR #11333, opened 2026-06-12
NIFI-16002 Fix LookupRecord writing to wrong level when parent path is missing

rakesh-rsky:
NIFI-16002 Fix LookupRecord writing to wrong level when parent path is missing

When the configured Result RecordPath contains intermediate nodes that do not yet exist in the record (e.g. /enrichment/value where /enrichment is absent), the RecordPath evaluator returns a missing-child FieldValue with a null value. When updateValue is later called, getParentRecord() skips the null intermediate node and finds the root record instead, writing the result at the wrong nesting level.

The fix adds a helper method createMissingParentRecords that walks the FieldValue parent chain and creates an empty MapRecord for each missing intermediate node, setting it on its parent before the actual value is applied. After creating the intermediate Records, incorporateInactiveFields is called so the new fields appear in the active schema and the re-evaluated path resolves them correctly.

This fix applies to both scalar lookup values and Record-typed lookup values.

# Summary

[NIFI-16002](https://issues.apache.org/jira/browse/NIFI-16002)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/LookupRecord.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/LookupRecord.java
@@ -58,12 +58,15 @@ import org.apache.nifi.serialization.RecordReader;
 import org.apache.nifi.serialization.RecordReaderFactory;
 import org.apache.nifi.serialization.RecordSetWriter;
 import org.apache.nifi.serialization.RecordSetWriterFactory;
+import org.apache.nifi.serialization.SimpleRecordSchema;
 import org.apache.nifi.serialization.WriteResult;
 import org.apache.nifi.serialization.record.DataType;
+import org.apache.nifi.serialization.record.MapRecord;
 import org.apache.nifi.serialization.record.Record;
 import org.apache.nifi.serialization.record.RecordField;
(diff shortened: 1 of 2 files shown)


exceptionfactory (changes requested):
Thanks for proposing this approach to improve the behavior of `LookupRecord` @rakesh-rsky. On initial review, I'm not sure whether creating missing intermediate objects is the right approach. The primary concern is that it alters the functional schema of the `Record`, without reflecting those changes in the Record Schema.

The current behavior does seem problematic, ignoring the intermediate field, and it seems like throwing an error might be an option to consider. There also appears to be some potential performance implications of this approach. Another possibility is to configure this behavior through a new property, but it would be helpful to evaluate the potential issues before moving in that direction.

exceptionfactory (changes requested):
Returning to this pull request, I see the `incorporateInactiveFields()` call, which is helpful. However, I'm not sure about the increased cost of this approach. Having to make two passes through every record could be expensive for a large number of records. For this reason, it seems like the behavior should be controlled by a new property, named something like `Missing Field Strategy`. This introduces some complexity, but would provide a safer way to introduce the option.

exceptionfactory:
Given the age of the PR and new potential direction, I'm closing it for now, but if you are able to make the changes, feel free to re-open.

PR #11336, opened 2026-06-15
NIFI-16020 Consume request body before two-phase cluster replication responses to prevent HTTP/2 stream resets

pvillard31:
# Summary

NIFI-16020 Consume request body before two-phase cluster replication responses to prevent HTTP/2 stream resets

Several REST resource methods do not declare a request entity parameter (for example `POST /flowfile-queues/{id}/drop-requests`), so the JAX-RS layer never reads the request body even though the replicating node always sends one. When a node returns the two-phase commit validation, execution, or cancellation response without consuming that body, Jetty 12.1.10 resets the HTTP/2 stream with `CANCEL`, which surfaces on the replicating node as "RST_STREAM received Stream cancelled" and fails the replicated request (which cannot be retried).

`ApplicationResource.withWriteLock` now fully consumes the request input stream for two-phase requests before any early response is returned, allowing the replicating node to complete its upload cleanly. When the body has already been consumed (for example via entity deserialization) the drain is a no-op.

This server-side fix replaces the temporary HTTP/1.1 cluster workaround added in NIFI-16011, so the configurable `nifi.cluster.node.protocol.http.version` property introduced there and the HTTP_1_1 overrides in the system-test configurations are reverted.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


markap14 (approved):
Great catch @pvillard31 changes look good to me. I like this much better than the HTTP version property!  +1 will merge.

PR #11338, opened 2026-06-15
NIFI-15975 Fixed single line comment bug, multiple let statements bug and improved the invalid read JSLT transform message.

dan-s1:
# Summary

[NIFI-15975](https://issues.apache.org/jira/browse/NIFI-15975)
This PR along with the fix in the `nifi-api` for [NIFI-15982](https://issues.apache.org/jira/browse/NIFI-15982) solve three issues that were reported in NIFI-15975.
The first two issues had to do with when single line Java comments are used in the JSLT transform. Before the fix for NIFI-15982, the framework mistook the leading forward slash of the Java comment as a file path. The second issue was the code in the `JSLTTransformJSON` actually did not report the true reason why it failed which was the JSLT transform was not a valid file path. The fix for NIFI-15982 took care of NIFI mistakenly thinking the single line Java comment is a file and the fix in this PR ensures the real reason for failure is displayed.
The final issue reported what was the inability to place two `let` statements on separate lines without placing an additional space between them. That was due to the JSLT transform being read with the new lines being stripped out. The fix for this was to reintroduce the new lines when reconstituting the JSLT transform after it is read from the given source.

As an aside, the fix for NIFI-15982 also fixed `JoltTransformJSON` and  `ValidateJson`, as the Jolt spec and schema source could be text and single Java comments are allowed.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
+++ b/nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
@@ -272,8 +272,7 @@ public class JSLTTransformJSON extends AbstractProcessor {
             getJstlExpression(transform, null);
             builder.valid(true);
         } catch (final RuntimeException e) {
-            final String explanation = String.format("%s not valid: %s", property.getDisplayName(), e.getMessage());
-            builder.valid(false).explanation(explanation);
+            builder.valid(false).explanation(e.getMessage());
         }
         return builder.build();
     }
@@ -430,9 +429,9 @@ public class JSLTTransformJSON extends AbstractProcessor {
             return propertyValue.getValue();
         }
         try (final BufferedReader reader = new BufferedReader(new InputStreamReader(resourceReference.read()))) {
-            return reader.lines().collect(Collectors.joining());
+            return reader.lines().collect(Collectors.joining(System.lineSeparator()));
         } catch (final IOException e) {
-            throw new UncheckedIOException("Read JSLT Transform failed", e);
+            throw new UncheckedIOException(String.format("Read JSLT Transform failed, reason: %s", e.getMessage()), e);
         }
     }
 
--- a/nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/test/java/org/apache/nifi/processors/jslt/TestJSLTTransformJSON.java
+++ b/nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/test/java/org/apache/nifi/processors/jslt/TestJSLTTransformJSON.java
@@ -23,9 +23,11 @@ import org.apache.nifi.util.TestRunner;
 import org.apache.nifi.util.TestRunners;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.opentest4j.AssertionFailedError;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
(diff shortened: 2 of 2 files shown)


nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/test/java/org/apache/nifi/processors/jslt/TestJSLTTransformJSON.java
```diff
@@ -74,6 +82,87 @@ public void testInvalidJSLTTransform() {
...
+
+        final String json = """
+                {
+                    "userId" : "jdoe",
+                    "firstName": "John"
+                }
+                """;
+        runner.enqueue(json);
+
+        // TODO Uncomment after NIFI-15982 is accepted and merged and NIFI uses the next version of nifi-api
```
pvillard31: Since NIFI-15982 is already merged and this build uses a nifi-api that includes it, should this TODO comment be removed, given that nothing below it is actually commented out?

nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/test/java/org/apache/nifi/processors/jslt/TestJSLTTransformJSON.java
```diff
@@ -74,6 +82,87 @@ public void testInvalidJSLTTransform() {
...
+
+        final String json = """
+                {
+                    "userId" : "jdoe",
+                    "firstName": "John"
+                }
+                """;
+        runner.enqueue(json);
+
+        assertThrows(AssertionFailedError.class, () -> runner.run());
```
pvillard31: This test only checks that an error is thrown. Can it also assert that the validation explanation or exception message contains the new reason text, so it would fail if the message change were reverted?

nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/test/java/org/apache/nifi/processors/jslt/TestJSLTTransformJSON.java
```diff
@@ -74,6 +82,87 @@ public void testInvalidJSLTTransform() {
...
+                {
+                    "value" : {
+                        "id" : "jdoe",
+                        "client_name": "John"
+                    }
+                }
+                """;
+        runner.enqueue(json);
+
+        assertDoesNotThrow(() -> runner.run());
```
pvillard31: Is there a reason to wrap runner.run() in assertDoesNotThrow here and in the parameterized test, or can we call runner.run() directly since the test already fails on any thrown exception?

nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
```diff
@@ -430,9 +429,9 @@ private String readTransform(final PropertyValue propertyValue) {
             return propertyValue.getValue();
         }
         try (final BufferedReader reader = new BufferedReader(new InputStreamReader(resourceReference.read()))) {
-            return reader.lines().collect(Collectors.joining());
+            return reader.lines().collect(Collectors.joining(System.lineSeparator()));
```
pvillard31, line 432: Did you consider joining the lines with a fixed newline instead of System.lineSeparator(), so the compiled transform and the modifyContent provenance detail stay identical across Windows and Linux?

pvillard31 (changes requested):

nifi-extension-bundles/nifi-jslt-bundle/nifi-jslt-processors/src/main/java/org/apache/nifi/processors/jslt/JSLTTransformJSON.java
dan-s1: I did not consider that. I made the same change as was made in #7678 for `JoltTransformJSON` and `JoltTransformRecord` which solved the similar issue. Do still want this change? If so, should a similar change then be made for `JoltTransformJSON` and `JoltTransformRecord`?

pvillard31 (approved):
Latest LGTM, thanks @dan-s1

PR #11355, opened 2026-06-18
NIFI-16028 - Support nested field paths in PutElasticsearchJson

agturley:
# Summary

[NIFI-16028](https://issues.apache.org/jira/browse/NIFI-16028)

Allow the Index Field, Identifier Field, and Timestamp Field properties to reference a nested
field with a "/"-delimited path (e.g. "@metadata/index"). A value with no "/" behaves exactly as
before (top-level lookup), so existing flows are unaffected.

When the corresponding Retain property is false, the extracted leaf is removed and any ancestor
object left empty is pruned, so a transient envelope like "@metadata" is not indexed once its
fields are consumed; parents that still have other fields are kept.

Applies across NDJSON, JSON Array, and Single JSON, and all index operations. The NDJSON
raw-bytes fast path is bypassed only when a nested path is configured, since the streaming scan
only matches flat field names.

Builds on NIFI-15985.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
+++ b/nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
@@ -237,10 +237,25 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .dependsOn(INPUT_FORMAT, InputFormat.SINGLE_JSON)
             .build();
 
+    static final PropertyDescriptor FIELD_PATH_MODE = new PropertyDescriptor.Builder()
+            .name("Field Path Mode")
+            .description("""
+                    How the Identifier Field, Index Field, and Timestamp Field property values are interpreted \
+                    when locating a field within each document. In "Literal Field Name" mode (the default) each \
+                    value is the exact name of a top-level field. In "Nested Field Path" mode each value is a \
+                    "/"-delimited path into nested objects; see the processor's Additional Details for the full \
+                    path syntax, including how to reference a field name that contains a literal "/".\
+                    """)
+            .required(true)
+            .allowableValues(FieldReferenceMode.class)
+            .defaultValue(FieldReferenceMode.LITERAL.getValue())
+            .build();
+
     static final PropertyDescriptor IDENTIFIER_FIELD = new PropertyDescriptor.Builder()
             .name("Identifier Field")
             .description("""
-                    The name of the field within each document to use as the Elasticsearch document ID. \
+                    The name of the field within each document to use as the Elasticsearch document ID, \
+                    interpreted as a literal field name or a nested "/"-delimited path per the Field Path Mode property. \
                     If the field is not present in a document or this property is left blank, no document ID is set \
                     and Elasticsearch will auto-generate one.\
                     """)
@@ -255,7 +270,8 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .description("""
                     Whether to keep the Identifier Field in the document body after extracting it \
                     for use as the Elasticsearch document ID. \
-                    When true (default), the field is left in the document; set to false to remove it before indexing.\
+                    When true (default), the field is left in the document; set to false to remove it before indexing. \
+                    For a nested ("/"-delimited) path, any parent object left empty by the removal is also pruned.\
                     """)
             .required(true)
             .allowableValues("true", "false")
@@ -266,7 +282,8 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
     static final PropertyDescriptor INDEX_FIELD = new PropertyDescriptor.Builder()
             .name("Index Field")
             .description("""
-                    The name of the field within each document to use as the Elasticsearch index name. \
+                    The name of the field within each document to use as the Elasticsearch index name, \
+                    interpreted as a literal field name or a nested "/"-delimited path per the Field Path Mode property. \
                     If the field is not present in a document or this property is left blank, \
                     the configured Index property value is used as the fallback.\
                     """)
@@ -280,7 +297,8 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .description("""
                     Whether to keep the Index Field in the document body after extracting it \
                     for use as the Elasticsearch index name. \
-                    When true (default), the field is left in the document; set to false to remove it before indexing.\
+                    When true (default), the field is left in the document; set to false to remove it before indexing. \
+                    For a nested ("/"-delimited) path, any parent object left empty by the removal is also pruned.\
                     """)
             .required(true)
             .allowableValues("true", "false")
@@ -292,7 +310,8 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .name("Timestamp Field")
             .description("""
                     The name of a field within each document whose value will be written to \
-                    Elasticsearch as the @timestamp field. \
+                    Elasticsearch as the @timestamp field, interpreted as a literal field name or a nested \
+                    "/"-delimited path per the Field Path Mode property. \
                     If the field is absent or this property is left blank, no @timestamp is set.\
                     """)
             .required(false)
@@ -305,7 +324,8 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .description("""
                     Whether to keep the Timestamp Field in the document body after copying its \
                     value to @timestamp. \
-                    When true (default), the field is left in the document; set to false to remove it before indexing.\
+                    When true (default), the field is left in the document; set to false to remove it before indexing. \
+                    For a nested ("/"-delimited) path, any parent object left empty by the removal is also pruned.\
                     """)
             .required(true)
             .allowableValues("true", "false")
@@ -341,6 +361,7 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             MAX_BATCH_SIZE,
             SUPPRESS_NULLS,
             ID_ATTRIBUTE,
+            FIELD_PATH_MODE,
             IDENTIFIER_FIELD,
             RETAIN_IDENTIFIER_FIELD,
             INDEX_FIELD,
@@ -466,11 +487,20 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
         final String idAttribute = inputFormat == InputFormat.SINGLE_JSON
                 ? context.getProperty(ID_ATTRIBUTE).getValue()
                 : null;
+        // In Literal Field Name mode the configured values are escaped into single-segment paths so the
+        // shared path helpers resolve them to the exact (verbatim) field name; in Nested Field Path mode
+        // they are already "/"-delimited paths and used as-is.
+        final boolean nestedFieldPaths = FieldReferenceMode.NESTED_PATH.getValue().equals(context.getProperty(FIELD_PATH_MODE).getValue());
         final String documentIdField = inputFormat != InputFormat.SINGLE_JSON
-                ? context.getProperty(IDENTIFIER_FIELD).evaluateAttributeExpressions().getValue()
+                ? fieldPath(context.getProperty(IDENTIFIER_FIELD).evaluateAttributeExpressions().getValue(), nestedFieldPaths)
                 : null;
-        final String documentIndexField = context.getProperty(INDEX_FIELD).evaluateAttributeExpressions().getValue();
-        final String documentTimestampField = context.getProperty(TIMESTAMP_FIELD).evaluateAttributeExpressions().getValue();
+        final String documentIndexField = fieldPath(context.getProperty(INDEX_FIELD).evaluateAttributeExpressions().getValue(), nestedFieldPaths);
+        final String documentTimestampField = fieldPath(context.getProperty(TIMESTAMP_FIELD).evaluateAttributeExpressions().getValue(), nestedFieldPaths);
+        // The id/index field paths are loop-invariant, so classify them as nested-vs-flat and decode
+        // the flat names once here rather than per record in the NDJSON raw-bytes fast path below.
+        final boolean nestedExtractionField = isNestedPath(documentIdField) || isNestedPath(documentIndexField);
+        final String documentIdFieldName = documentIdField == null ? null : decodeSegment(documentIdField);
+        final String documentIndexFieldName = documentIndexField == null ? null : decodeSegment(documentIndexField);
         // The Retain toggles depend on their source field being set, so only read them when it is —
         // reading a property whose dependency is unsatisfied is invalid. When the source field is
         // blank no extraction or stripping occurs, so the retain value is irrelevant anyway.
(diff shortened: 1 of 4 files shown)


nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -241,6 +241,7 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .name("Identifier Field")
             .description("""
                     The name of the field within each document to use as the Elasticsearch document ID. \
+                    A nested field can be referenced with a "/"-delimited path (e.g. "@metadata/id"). \
```
pvillard31: Could you rebase this branch on the latest main? The validation gate is failing on "branch not up to date", and because of that the full build and test matrix was skipped, so we do not have a green CI signal yet.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -1143,30 +1156,168 @@ private String resolveId(final Map<String, Object> contentMap, final String idAt
...
         }
-        final String value = fieldValueToString(contentMap.get(indexField));
+        final String value = fieldValueToString(valueAtPath(contentMap, indexField));
         return StringUtils.isNotBlank(value) ? value : fallbackIndex;
     }
 
+    /**
+     * Whether {@code field} is a {@code /}-delimited nested path rather than a flat field name.
+     */
+    private static boolean isNestedPath(final String field) {
```
pvillard31, line 1230: A field key that literally contains a slash is now always read as a nested path. Should we note this limitation in the property descriptions so users with such field names are not surprised?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJsonTest.java
```diff
@@ -1110,4 +1110,170 @@ void testIndexFieldNullValueFallsBackConsistently() {
...
+        final List<IndexOperationRequest> ops = captureOperations();
+        runner.enqueue("{\"@metadata\":{\"index\":\"docs-single\"},\"msg\":\"x\"}");
+        runner.run();
+
+        assertEquals("docs-single", ops.getFirst().getIndex());
+        assertFalse(ops.getFirst().getFields().containsKey("@metadata"));
+    }
+
+    @Test
+    void testNestedIndexFieldKeepsParentWithRemainingSiblings() {
```
pvillard31, line 1252: Could we add a test with a 3 level path (for example a/b/c) to lock in the recursive pruning of multiple empty ancestor objects?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJsonTest.java
```diff
@@ -1110,4 +1110,170 @@ void testIndexFieldNullValueFallsBackConsistently() {
...
+        final List<IndexOperationRequest> ops = captureOperations();
+        runner.enqueue("{\"@metadata\":{\"index\":\"docs-2026\"},\"msg\":\"hello\"}\n");
+        runner.run();
+
+        assertEquals("docs-2026", ops.getFirst().getIndex());
+        assertTrue(docContent(ops.getFirst()).contains("@metadata"), "nested field retained by default");
+    }
+
+    @Test
+    void testNestedIdentifierFieldNdjsonPrunesEmptyParent() {
```
pvillard31, line 1287: Should we add a nested identifier test for an Update, Delete, or Upsert operation, so the resolveId nested path is covered as well?

pvillard31 (changes requested):

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: I've altered it to allow users to escape a backslash to allow a backslash in the path

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJsonTest.java
agturley: done

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: done

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/test/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJsonTest.java
agturley: Added testNestedIdentifierFieldUpdateOperationResolvesAndPrunes — it covers the resolveId nested path via an Update op (@metadata/id), verifying the nested ID resolves and the empty parent is pruned. Also added an escaped-slash test on the same path.

agturley:
Do you think the description updates I provided are overkill?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -241,6 +241,11 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .name("Identifier Field")
             .description("""
                     The name of the field within each document to use as the Elasticsearch document ID. \
+                    A nested field can be referenced with a "/"-delimited path: for a document \
```
pvillard31: These three properties shipped in released 2.10.0 with literal field name semantics, so an existing flow that has a slash in the configured value will now be read as a nested path after an upgrade. I am not an experienced Elasticsearch user and I do not know how common a slash is in these field names, and I agree the likelihood of a problematic value is low, but I think we should still handle this case safely with a property migration that escapes any existing slash or backslash, and add test coverage for that migration. Could we add that here?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -267,6 +273,11 @@ public class PutElasticsearchJson extends AbstractPutElasticsearch {
             .name("Index Field")
             .description("""
                     The name of the field within each document to use as the Elasticsearch index name. \
+                    A nested field can be referenced with a "/"-delimited path: for a document \
```
pvillard31: The escape explanation is repeated in all three property descriptions. Could we move the full explanation into an additionalDetails.md for the processor and have each property description point to the processor documentation for the details instead?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
```diff
@@ -1143,30 +1177,226 @@ private String resolveId(final Map<String, Object> contentMap, final String idAt
...
+        return current;
+    }
+
+    /**
+     * Removes the leaf at the {@code /}-delimited {@code path} within {@code root}, then prunes any
+     * ancestor objects that become empty as a result (the transient envelope often used for routing
+     * metadata, e.g. {@code @metadata}, should not be indexed once its fields are extracted).
+     * Does nothing when the path is absent. A path with no slash is a direct top-level removal.
+     */
+    private static void removeAtPath(final Map<String, Object> root, final String path) {
```
pvillard31, line 1359: The Map and ObjectNode versions of removeAtPath duplicate the same descend and prune loop. Could we factor the shared walk into one helper so the two cannot drift over time?

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: So I started on this and ran into something I want to flag before going further.

The problem is migrateProperties runs on every flow import and restart, and there's no way to tell a 2.10 literal field name like @metadata/id apart from a 2.11 nested path @metadata/id — they're the exact same string. So if I just escape slashes, it also escapes legit nested paths and breaks them. I actually hit this on my test cluster: imported a normal 2.11 flow and every nested-path case quietly stopped working because the migration had rewritten @metadata\/id to @metadata\\/id on import.

The only way I found to make it safe is to rename the three properties (Identifier Field → Identifier Field Path, etc). Then the old name only exists in pre-2.11 flows, so I can escape values found under the old name and leave the new ones alone. That works, but it means renaming three properties everyone's already using.

This is feeling like a lot for the case we're protecting against — a 2.10 flow that happens to have a literal / in the field name, and slashes aren't really common in ES field names to begin with. So I'm inclined to skip the migration and just document it: on upgrade, a field name with a literal / gets read as a nested path now, and you can escape it as \/ if you want the old behavior.

Let me know if you'd rather I do the rename though, happy to if you think the upgrade guarantee is worth it.

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: agreed, done

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
agturley: done

nifi-extension-bundles/nifi-elasticsearch-bundle/nifi-elasticsearch-restapi-processors/src/main/java/org/apache/nifi/processors/elasticsearch/PutElasticsearchJson.java
pvillard31: @ChrisSamo632 - thoughts on this? You're way more familiar with ES so your feedback would be appreciated if you have the chance

ChrisSamo632: I'm not aware of any restrictions around the use of `/` characters in field names in Elasticsearch or json in general, although I'd imagine the occurrence is relatively infrequent

Trying to think of examples, something like `house_name/number` could be the sort of thing someone might use, but I'd typically opt for an alternative in such a case as I'd see the `/` as unusual (and potentially problematic in some systems/parsers)

For this instance, would a workable option be to add a processor parameter that let's the user configure how a `/` should be treated in field names for this processor? The default could be to have the character treated as a literal, with a new mode that enables the nested object parsing?

Alternatively, would `.` be a better option as it matches the Elasticsearch nested query parsing, but then is further from the nifi record path parsing, which `/` emulates? The same concern would exist, although use of `.` in elasticsearch field names has been discouraged for a long time as it can confuse the Elasticsearch query parser - therefore, the chances of people using `.` litterals on field names seems lower

ChrisSamo632: Thinking about the `.` suggestion a little more - I'd be concerned with someone using that and *not* having the field removed from the json by the processor before it was indexed, which is more likely to run into problems on the Elasticsearch side

So maybe the best idea is to stick with `/` (to be nifi record path-like), and offer a property that uses it as a nested field delimiter (allowing for `\` escaping as needed), or as a literal (the default for backwards compatibility)

What do you think @agturley ?

agturley: yes i like that idea. i'll adjust and submit.

agturley:
Sorry about the delay in responding to this. I was out last week and was not able to do any updates.
Anyways, There's now a "Field Path Mode" property with two options:

**Literal Field Name (default):**  the Identifier/Index/Timestamp Field value is the exact name of a top-level field, exactly like before. / and \ are just regular characters.
**Nested Field Path:** the value is a /-delimited path into nested objects (e.g. @metadata/id), with \/ to escape a literal slash.
Because it defaults to Literal Field Name, existing flows keep working with no change and no migration is needed — nesting is purely opt-in, so there's no risk of an existing field name with a / being reinterpreted on upgrade.

I tested it end to end against a cluster (nested id/index/timestamp extraction, deep paths, escaped slashes, and plain pass-through) and it all behaves as expected. I also added an example for each mode to the processor's Additional Details, and tests covering both the literal default and the nested behavior.

pvillard31 (approved):
Thanks for making the changes, the latest LGTM. @ChrisSamo632 - anything on your side?

ChrisSamo632:
LGTM 👍

Thanks for the improvements @agturley , and the detailed review @pvillard31

ChrisSamo632 (approved):

PR #11369, opened 2026-06-25
NIFI-15911 Add Web Client Service for Snowflake Ingest Service

exceptionfactory:
# Summary

[NIFI-15911](https://issues.apache.org/jira/browse/NIFI-15911) Adds a `Web Client Service Provider` property to the `StandardSnowflakeIngestManagerProviderService` in order to support Proxy Servers and other configuration features for HTTP operations.

The additional property is required, and property migration methods add a default `StandardWebClientServiceProvider` implementation when the Snowflake Ingest Manager Provider Service does not have one configured. This enables upgrades to existing flows without requiring manual configuration.

The implementation changes replace direct use of the Java `HttpClient` with the equivalent `WebClientService` methods defined in the `nifi-web-client-api` module. Updated unit tests exercise the property migration behavior.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [X] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/SnowpipeIngestClient.java
+++ b/nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/SnowpipeIngestClient.java
@@ -22,23 +22,21 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.nifi.processors.snowflake.snowpipe.InsertFiles;
 import org.apache.nifi.processors.snowflake.snowpipe.InsertReport;
+import org.apache.nifi.web.client.api.HttpResponseEntity;
+import org.apache.nifi.web.client.api.WebClientService;
+import org.apache.nifi.web.client.api.WebClientServiceException;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.URI;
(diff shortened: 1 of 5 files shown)


nifi-extension-bundles/nifi-snowflake-bundle/nifi-snowflake-services/src/main/java/org/apache/nifi/snowflake/service/SnowpipeIngestClient.java
```diff
@@ -135,26 +135,34 @@ InsertReport getInsertReport() {
...
+                        .header(ACCEPT_HEADER, APPLICATION_JSON)
+                        .retrieve();
+                InputStream responseBodyStream = responseEntity.body()
+        ) {
+            final int statusCode = responseEntity.statusCode();
+            if (statusCode == HttpURLConnection.HTTP_OK) {
+                try {
+                    return objectMapper.readValue(responseBodyStream, InsertReport.class);
+                } catch (final JsonProcessingException e) {
+                    final byte[] responseBodyBytes = responseBodyStream.readAllBytes();
```
pvillard31: When readValue fails to parse, has the response stream already been consumed, so the body included in the error message would be empty?

exceptionfactory: Good catch, thanks! I pushed an update to read the response body as a string regardless of status code.

pvillard31 (approved):
Thanks @exceptionfactory - LGTM

PR #11370, opened 2026-06-25
NIFI-16049 Corrected if else logic in FlowFileUnpackagerV1

dan-s1:
# Summary

[NIFI-16049](https://issues.apache.org/jira/browse/NIFI-16049)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFileUnpackagerV1.java
+++ b/nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFileUnpackagerV1.java
@@ -82,13 +82,13 @@ public class FlowFileUnpackagerV1 implements FlowFileUnpackager {
                 throw new IOException("Flow file attributes object contains key of type "
                         + keyObject.getClass().getCanonicalName()
                         + " but expected java.lang.String");
-            } else if (!(keyObject instanceof String)) {
+            }
+            if (!(valueObject instanceof final String value)) {
                 throw new IOException("Flow file attributes object contains value of type "
                         + keyObject.getClass().getCanonicalName()
                         + " but expected java.lang.String");
             }
 
-            final String value = (String) valueObject;
             result.put(key, value);
         }
 
--- a/nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFileUnpackagerV1.java
+++ b/nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFileUnpackagerV1.java
@@ -85,7 +85,7 @@ public class FlowFileUnpackagerV1 implements FlowFileUnpackager {
             }
             if (!(valueObject instanceof final String value)) {
                 throw new IOException("Flow file attributes object contains value of type "
-                        + keyObject.getClass().getCanonicalName()
+                        + valueObject.getClass().getCanonicalName()
                         + " but expected java.lang.String");
             }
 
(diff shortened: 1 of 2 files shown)


nifi-commons/nifi-flowfile-packager/src/main/java/org/apache/nifi/util/FlowFileUnpackagerV1.java
```diff
@@ -78,18 +78,17 @@ protected Map<String, String> getAttributes(final TarArchiveInputStream stream)
...
-            if (!(keyObject instanceof String)) {
+            if (!(keyObject instanceof final String key)) {
                 throw new IOException("Flow file attributes object contains key of type "
                         + keyObject.getClass().getCanonicalName()
                         + " but expected java.lang.String");
-            } else if (!(keyObject instanceof String)) {
+            }
+            if (!(valueObject instanceof final String value)) {
                 throw new IOException("Flow file attributes object contains value of type "
                         + keyObject.getClass().getCanonicalName()
```
pvillard31: In the value type error, should this report valueObject.getClass().getCanonicalName() instead of keyObject, so the message shows the actual type of the bad value?

exceptionfactory (commented):
Thanks @dan-s1, is this change something that can be easily validated with a unit test?

dan-s1:
@exceptionfactory Let me see what I can do.

dan-s1:
@exceptionfactory When writing a unit test, I came across an issue that I cannot write an XML file using the `java.util.Properties` `storeToXML` method to write attributes whose key or value are not strings. Furthermore in the `FlowFilePackagerV1` code in method `packageFlowFile` it seems it only takes a map of attributes whose keys and values are only strings. In addition it seems the `java.util.Properties`  `loadFromXML`  used in `FlowFileUnpackagerV1` will only load a Properties object whose keys and values are strings. 
Hence I question, why there are these checks in `FlowFileUnpackagerV1` in the first place? It would seem you cannot write anything but strings. Perhaps we should remove these checks entirely. I do have a unit test to test unpackaging in a normal scenario which I can contribute if you wish. Please advise.

dan-s1:
@pvillard31 @exceptionfactory Per my earlier comments, I believe the way `FlowFileUnpackagerV1` was trying to retrieve the attributes from the `java.util.Properties` object was incorrect. From reading the `java.util.Properties` javadocs,  it seems once the `java.util.Properties` load its properties from an XML file, all its keys and values are strings and not any other object. Hence the previous logic which was checking the instance type was completely unnecessary. I have corrected this with my latest commit to use streams and the `java.util.Properties` `stringPropertyNames` and `getProperty` to build a map of the attributes. I have also added a unit test for testing the `FlowFilePackagerV1`  and the `FlowFileUnpackagerV1`.

mosermw:
> @pvillard31 @exceptionfactory Per my earlier comments, I believe the way `FlowFileUnpackagerV1` was trying to retrieve the attributes from the `java.util.Properties` object was incorrect. From reading the `java.util.Properties` javadocs, it seems once the `java.util.Properties` load its properties from an XML file, all its keys and values are strings and not any other object. Hence the previous logic which was checking the instance type was completely unnecessary. I have corrected this with my latest commit to use streams and the `java.util.Properties` `stringPropertyNames` and `getProperty` to build a map of the attributes. I have also added a unit test for testing the `FlowFilePackagerV1` and the `FlowFileUnpackagerV1`.

@dan-s1  I wouldn't worry about the FlowFileV1 format very much.  It is extremely old and there are good reasons why it was replaced.  Don't overthink it or waste time on it.

exceptionfactory (approved):
Thanks for the correction and the unit test @dan-s1, the latest version looks good.

PR #11372, opened 2026-06-26
NIFI-16051 Propagate logging attributes down the process group tree

boguszj:
# Summary

[NIFI-16051](https://issues.apache.org/jira/browse/NIFI-00000)

Propagating the logging attributes down solves 2 issues:

- missing registeredFlowIdentifierPath when flow is first imported from the registry
- lack of registeredFlowIdentifierPath propagation down to the children when VCI changes on the root PG

Technically, the 1st issue could probably be fixed in the initialization process itself, but propagation fixes both & makes sure they are always synchronized.

Additionally moving from a mutating the map, to overriding with an immutable instance would make sure there are no partial reads.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
@@ -315,7 +315,7 @@ public final class StandardProcessGroup implements ProcessGroup {
         if (newParent instanceof StandardProcessGroup standardParent) {
             this.connectorLoggingAttributes = standardParent.connectorLoggingAttributes;
         }
-        setLoggingAttributes();
+        setLoggingAttributes(true);
     }
 
     @Override
@@ -355,7 +355,7 @@ public final class StandardProcessGroup implements ProcessGroup {
         }
 
         this.name.set(name);
-        setLoggingAttributes();
+        setLoggingAttributes(true);
     }
 
     @Override
@@ -3695,7 +3695,7 @@ public final class StandardProcessGroup implements ProcessGroup {
                 parent.onComponentModified();
             }
 
-            setLoggingAttributes();
+            setLoggingAttributes(true);
             scheduler.submitFrameworkTask(() -> synchronizeWithFlowRegistry(flowManager));
         } finally {
             writeLock.unlock();
@@ -3770,7 +3770,7 @@ public final class StandardProcessGroup implements ProcessGroup {
         writeLock.lock();
         try {
             this.versionControlInfo.set(null);
-            setLoggingAttributes();
+            setLoggingAttributes(true);
         } finally {
             writeLock.unlock();
         }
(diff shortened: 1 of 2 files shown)


pvillard31 (commented):
LGTM and happy to merge as-is but leaving two comments up for discussion

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
```diff
@@ -4816,28 +4816,36 @@ public void setStatelessFlowTimeout(final String statelessFlowTimeout) {
...
         }
 
-        loggingAttributes.putAll(connectorLoggingAttributes);
+        attributes.putAll(connectorLoggingAttributes);
+
+        this.loggingAttributes = Map.copyOf(attributes);
+
+        for (final ProcessGroup childGroup : processGroups.values()) {
+            if (childGroup instanceof StandardProcessGroup standardChildGroup) {
+                standardChildGroup.setLoggingAttributes();
```
pvillard31: Since setConnectorLoggingAttributes already recurses into every child and each child now also calls setLoggingAttributes(), each descendant gets recomputed once per ancestor level. Did you consider skipping this subtree recursion when it is invoked from the connector cascade, to avoid the repeated recomputation?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
```diff
@@ -4816,28 +4816,36 @@ public void setStatelessFlowTimeout(final String statelessFlowTimeout) {
...
-            loggingAttributes.put(LoggingAttribute.REGISTERED_FLOW_VERSION.attribute, registeredFlowVersion);
+            attributes.put(LoggingAttribute.REGISTERED_FLOW_VERSION.attribute, registeredFlowVersion);
         }
 
-        loggingAttributes.putAll(connectorLoggingAttributes);
+        attributes.putAll(connectorLoggingAttributes);
+
+        this.loggingAttributes = Map.copyOf(attributes);
+
+        for (final ProcessGroup childGroup : processGroups.values()) {
```
pvillard31: The sibling method setConnectorLoggingAttributes iterates getProcessGroups() (a read-locked copy) while this loop iterates processGroups.values() directly. Both are safe with a ConcurrentHashMap, but should we use getProcessGroups() here too for consistency?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/groups/StandardProcessGroup.java
boguszj: Initially I was thinking that this happens rarely enough that added computation is not a problem, so I kept things simple, but on another thought maybe being explicit about it is a better idea. I know boolean arguments probably aren't ideal, but since similar pattern is also used for `getControllerServices` I added one here

pvillard31 (approved):
LGTM and will merge assuming checks are green

PR #11376, opened 2026-06-30
NIFI-16057 - Restrict trigger push branch for GitHub Workflow

apupier:
Feature branches rarely need their own CI runs: the code is already tested when a pull request is opened against a release branch. If the push trigger has no branch restriction and pull_request is also configured, every push to a branch with an open PR runs the workflow twice: once for the push and once for the PR synchronisation.

Always give the push trigger an explicit list of branches: this stops branches created from a release branch from inheriting its workflow runs.

see https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=430408443#GitHubActionsRecommendedPractices-Restrictthepushtriggertospecificbranches

# Summary

[NIFI-16057](https://issues.apache.org/jira/browse/NIFI-16057)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,11 @@
 # limitations under the License.
 name: build
 
-on: [push, pull_request]
+on:
+  push:
+    branches-ignore:
+      - dependabot/*
+  pull_request:
 
 env:
   MAVEN_COMMAND: ./mvnw
--- a/.github/workflows/code-compliance.yml
+++ b/.github/workflows/code-compliance.yml
@@ -21,6 +21,8 @@ on:
     - cron: "0 0 * * *"
   pull_request:
   push:
+    branches-ignore:
+      - dependabot/*
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/docker-tests.yml
+++ b/.github/workflows/docker-tests.yml
@@ -17,6 +17,8 @@ name: docker-tests
 
 on:
   push:
+    branches-ignore:
+      - dependabot/*
     paths:
       - '.github/workflows/docker-tests.yml'
       - 'pom.xml'
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -17,6 +17,8 @@ name: integration-tests
 
 on:
   push:
+    branches-ignore:
+      - dependabot/*
     paths:
       - '.github/workflows/integration-tests.yml'
       - 'pom.xml'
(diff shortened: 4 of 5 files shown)


exceptionfactory (changes requested):
Thanks for proposing these adjustments @apupier.

Although this might be helpful in some circumstances, it can also be limiting on repository forks. For example, pushing to a feature branch and running the GitHub workflows can provide useful validation prior to submitting a pull request. This avoids creating pull requests simply for the purpose of executing GitHub workflows.

Have you observed particular problems that prompted these changes?

apupier:
> Have you observed particular problems that prompted these changes?

Apache has a shared pool of GitHub Action runners. The shared pool is regularly hitting the limits, queuing jobs of all Apache repositories.

For instance, it will avoid to have 2 times the same build for the same changes for dependabot branches:
image

exceptionfactory:
> > Have you observed particular problems that prompted these changes?
> 
> Apache has a shared pool of GitHub Action runners. The shared pool is regularly hitting the limits, queuing jobs of all Apache repositories.
> 
> For instance, it will avoid to have 2 times the same build for the same changes for dependabot branches: image

Thanks, I have noticed that specifically with `dependabot` as well. Most pull requests come from forks, so they don't trigger this problem.

It would be great if the changes could be narrowed to the primary repository, and not applied to forks, which should address `dependabot` and any similar issues. Do you have any thoughts about such an approach?

apupier:
> It would be great if the changes could be narrowed to the primary repository, and not applied to forks, which should address `dependabot` and any similar issues. Do you have any thoughts about such an approach?

based on the [documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#push), i do not see a simple solution for that. i could only imagine to have the job triggered and then a specific check `if` to cancel the build but it doesn't sound really nice.

Maybe a reasonable mitigation could be to exclude the dependabot branches, as I suppose thta it is the case which occurs most often and a PR is always provided. In this case by using branch-ignores: dependabot/* for push events

exceptionfactory:
> Maybe a reasonable mitigation could be to exclude the dependabot branches, as I suppose thta it is the case which occurs most often and a PR is always provided. In this case by using branch-ignores: dependabot/* for push events

Yes, ignoring `dependabot/*` for push events sounds like a good strategy. Could you update the pull request with that approach?

exceptionfactory (approved):
Thanks for discussing the options and providing the alternative approach @apupier, the latest version looks good! +1 merging

PR #11379, opened 2026-06-30
NIFI-16059: validate connector postMessage origin against the host or…

mcgilman:
…igin instead of the entity-controlled configurationUrl (CWE-346)

# Summary

[NIFI-16059](https://issues.apache.org/jira/browse/NIFI-16059)

Connector custom UIs are embedded in the application via an `<iframe>` and
communicate with the host through `window.postMessage`. The host's
`ConnectorMessageHost` derived the origin it trusted — for both inbound message
validation and the outbound `targetOrigin` — from the connector-provided
`configurationUrl` / `detailsUrl`. Because those URLs are controlled by the
connector entity, the entity could effectively attest to its own message
origin, which is an origin-validation weakness (CWE-346).

This change decouples the iframe `src` from the trusted origin. The host now
trusts only the application's own origin (`document.location.origin`), which is
valid because connector custom UIs are served by the same web server that serves
the application (the `configurationUrl` / `detailsUrl` are built from the
incoming request and contribute only the path, so the iframe is same-origin with
the parent).

### Changes

- `ConnectorMessageHost`
  - Removed the `expectedOrigin` option and the static `extractOrigin()` helper.
  - Trust only `trustedOrigin` (`document.location.origin`, via the `DOCUMENT`
    token), used for both inbound origin validation and the outbound
    `targetOrigin`. Messages whose `event.origin` does not match are dropped
    (fail-closed). The existing `event.source` (iframe `contentWindow`) and
    message namespace/type checks are unchanged.
- `connector-configure` / `connector-detail`
  - No longer pass `expectedOrigin` to `startListening(...)`.
  - The disconnected-node-acknowledgment is posted to `trustedOrigin` rather than
    an origin derived from the entity URL.

### Verification

- Added a regression test asserting that a message whose origin matches an
  attacker-controlled `configurationUrl` is dropped, and that a message from the
  application origin is accepted.
- Added a direct test for the outbound disconnected-node-acknowledgment asserting
  it is posted with `targetOrigin === window.location.origin` and the expected
  message namespace/type.
- `nx test nifi` (full suite) and `nx lint nifi` pass.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector-message-host.service.spec.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector-message-host.service.spec.ts
@@ -22,7 +22,8 @@ import { CONNECTOR_MESSAGE_NAMESPACE } from '@nifi/shared';
 import { ConnectorMessageHost, ConnectorMessageHostOptions } from './connector-message-host.service';
 
 describe('ConnectorMessageHost', () => {
-    const TRUSTED_ORIGIN = 'http://localhost:4200';
+    // The host trusts only the application's own origin; tests dispatch from it.
+    const TRUSTED_ORIGIN = window.location.origin;
 
     function createMockDestroyRef(): DestroyRef {
         const callbacks: Array<() => void> = [];
@@ -49,7 +50,7 @@ describe('ConnectorMessageHost', () => {
         }).compileComponents();
 
         const service = TestBed.inject(ConnectorMessageHost);
-        const router = TestBed.inject(Router) as { navigate: ReturnType<typeof vi.fn> };
+        const router = TestBed.inject(Router) as unknown as { navigate: ReturnType<typeof vi.fn> };
 
         const destroyRef = createMockDestroyRef();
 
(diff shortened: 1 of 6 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector-message-host.service.spec.ts
```diff
@@ -350,23 +350,43 @@ describe('ConnectorMessageHost', () => {
         });
     });
 
-    describe('extractOrigin', () => {
-        it('should extract origin from a valid URL', () => {
-            expect(ConnectorMessageHost.extractOrigin('https://connector-ui.example.com/config?id=123')).toBe(
-                'https://connector-ui.example.com'
+    describe('origin decoupling (FLOW-11982)', () => {
```
rfellows: This isn't a valid nifi jira number, can you update this to the proper reference or just eliminate it?

rfellows (changes requested):

rfellows (approved):
Thanks @mcgilman 👍

PR #11380, opened 2026-06-30
NIFI-16060 Tolerate connector configuration load failures on read

kevdoran:
# Summary

[NIFI-16060](https://issues.apache.org/jira/browse/NIFI-16060)

When a connector's stored configuration cannot be loaded or parsed (e.g. a corrupt config left behind by a failed commit), the ConnectorConfigurationProvider throws ConnectorConfigurationProviderException. On the SYNC_WITH_PROVIDER read path this exception previously propagated uncaught and surfaced as an HTTP 500 from the connector REST endpoints. Because the delete flow reads the connector (for its revision, and to snapshot it before deletion), a corrupt connector became impossible to read or delete.

Make the two read retrieval methods (getConnector and getConnectors with SYNC_WITH_PROVIDER) tolerate a configuration-load failure: log a warning, mark the connector invalid so the failure remains visible to clients via the DTO's validation status, and return the in-memory node instead of propagating. The connector therefore stays readable and deletable.

Write paths (addConnector, applyUpdate) call syncFromProvider directly and continue to propagate the exception, so create/apply-config still fail on a bad configuration rather than proceeding silently.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorRepository.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorRepository.java
@@ -552,7 +552,7 @@ public class StandardConnectorRepository implements ConnectorRepository {
         try {
             syncFromProvider(connector);
         } catch (final ConnectorConfigurationProviderException e) {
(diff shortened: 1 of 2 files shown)


markap14:
Thanks @kevdoran seems reasonable; will merge

markap14 (approved):

PR #11383, opened 2026-07-02
NIFI-16058 - Treat external controller service reference changes in versioned flows as environment-specific

pvillard31:
# Summary

NIFI-16058 - Treat external controller service reference changes in versioned flows as environment-specific

NIFI-15697 (PR #11006) fixed a badge/dialog inconsistency for external controller service references via three changes: (1) removed the comparator suppression of external CS reference diffs, (2) added name-based reconciliation of the snapshot, (3) added cache invalidation when an ancestor CS is removed. (3) was the real fix; (1)+(2) changed prior behavior and caused NIFI-16058: external CS references (e.g. a root SSL Context Service) now show as phantom local changes when the service differs by id, and name-based reconciliation can't help when the name differs across envs.

This keeps (3) and restores environment-specific treatment of external CS references, but in the flow-difference filter (`FlowDifferenceFilters.isEnvironmentalChange`) rather than the framework-agnostic comparator: a CS-reference change is environmental when the snapshot value isn't a locally-accessible ancestor service while the local value is. The filter runs on both the state and "Show Local Changes" paths, so they agree by construction. Name-based reconciliation is removed; `StandardFlowComparator` is unchanged.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/test/java/org/apache/nifi/util/TestFlowDifferenceFilters.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/test/java/org/apache/nifi/util/TestFlowDifferenceFilters.java
@@ -429,21 +429,15 @@ public class TestFlowDifferenceFilters {
 
         final InstantiatedVersionedProcessGroup localGroup = new InstantiatedVersionedProcessGroup(pgInstanceId, parentGroupId);
 
-        // !accessibleA && accessibleB -> environmental: snapshot points at a service that does not exist locally
-        // (e.g. a dev-environment id), local points at an existing ancestor service.
         assertExternalServiceChange(localGroup, flowManager, parentGroupId, propertyName, foreignServiceId, ancestorServiceX, true,
                 "Switching from an unavailable external service to a local ancestor service must be treated as environmental");
 
-        // accessibleA && accessibleB -> NOT environmental: a genuine switch between two services that both exist locally.
         assertExternalServiceChange(localGroup, flowManager, parentGroupId, propertyName, ancestorServiceY, ancestorServiceX, false,
                 "Switching between two locally-accessible ancestor services must remain a reported change");
 
-        // accessibleA && !accessibleB -> NOT environmental: local points at a service that is not an accessible ancestor (e.g. one defined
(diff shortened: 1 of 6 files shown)


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/test/java/org/apache/nifi/util/TestFlowDifferenceFilters.java
```diff
@@ -393,6 +395,94 @@ public void testControllerServiceCreationEnvironmentalChangeWithoutComponentNode
...
+        assertExternalServiceChange(localGroup, flowManager, parentGroupId, propertyName, ancestorServiceY, ancestorServiceX, false,
+                "Switching between two locally-accessible ancestor services must remain a reported change");
+
+        // accessibleA && !accessibleB -> NOT environmental: local points at a service that is not an accessible ancestor (e.g. one defined
+        // inside the versioned Process Group), so the change must surface.
+        assertExternalServiceChange(localGroup, flowManager, parentGroupId, propertyName, ancestorServiceX, internalServiceId, false,
+                "Switching to a service that is not an accessible ancestor (e.g. internal to the group) must remain a reported change");
+
+        // !accessibleA && !accessibleB -> NOT environmental.
+        assertExternalServiceChange(localGroup, flowManager, parentGroupId, propertyName, "foreign-1", "foreign-2", false,
```
exceptionfactory: The comments above these assertions are duplicative of the messages and should be removed.

exceptionfactory (changes requested):
Thanks @pvillard31, the general approach looks good. I noted one minor recommendation regarding test assertion comments

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/test/java/org/apache/nifi/util/TestFlowDifferenceFilters.java
pvillard31: Done, thanks

exceptionfactory (approved):

PR #11390, opened 2026-07-02
NIFI-16071 - Enable newly added controller services on versioned flow upgrade

pvillard31:
# Summary

NIFI-16071 - Enable newly added controller services on versioned flow upgrade

NIFI-15749 made a versioned flow upgrade start/enable components that are newly added in the new version when the process group is active (mirroring the state of the components that were already running). It only worked for newly added processors. A newly added controller service that is referenced solely by an existing processor (via a new property) was left DISABLED after the upgrade, so the referencing processor became invalid and could not run. Controller services are always persisted in a versioned flow as DISABLED (`VersionedComponentStateLookup.ENABLED_OR_DISABLED`), so the synchronizer's "enable if proposed state is ENABLED" step never selected them, and `RetainExistingStateComponentScheduler`'s "new service + active group -> enable" logic was never handed these services.

What changed:
- Added `ComponentScheduler.enableAddedControllerServicesAsync(...)` with a default no-op implementation.
- `StandardVersionedComponentSynchronizer` now routes the services that were newly added during synchronization through this method.
- `RetainExistingStateComponentScheduler` overrides it to enable those services when the process group is active, consistent with how newly added processors are handled.

Because the new method defaults to a no-op, startup, cluster reconnect, and restore paths are unchanged; only the versioned-flow upgrade path enables newly added controller services.

Added a system test reproducing the scenario (new controller service referenced only by an existing processor across a version change) and unit tests for the scheduler.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -778,6 +778,16 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
                 context.getComponentScheduler().enableControllerServicesAsync(Collections.singleton(service));
             }
         });
+
+        // Controller Services are always persisted in a versioned flow as DISABLED, so newly added services never appear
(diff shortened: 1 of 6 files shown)


exceptionfactory (approved):
Thanks for finding and resolving this issue @pvillard31, the adjusted approach for enabling added Controller Services looks good. +1 merging

PR #11394, opened 2026-07-06
NIFI-16077 Change default Security Protocol for Kafka components from PLAINTEXT to SSL

wojtre:
# Summary

[NIFI-16077](https://issues.apache.org/jira/browse/NIFI-16077)

Changes the default value of the shared **Security Protocol** property (`KafkaClientComponent.SECURITY_PROTOCOL`) from `PLAINTEXT` to `SSL`. The descriptor is defined once in the shared `KafkaClientComponent` interface and reused by `ConsumeKafka`, `PublishKafka` and the Kafka connection service, so a single change applies the new default uniformly.

## Motivation

When `security.protocol` is set to a plaintext variant (e.g. `SASL_PLAINTEXT`) while the broker actually requires TLS, the Kafka client reads the returned TLS alert record as a message length (`0x15030300`, ~336 MB) and attempts to allocate that amount per broker connection, exhausting the heap and causing an OOM crash-loop within ~30-60s. The root cause lives in the Kafka client (KAFKA-4090, open since 2016 and unresolved), so NiFi cannot prevent it directly — but a TLS-based default meaningfully reduces the chance of falling into this trap.

`SSL` is used as the default rather than `SASL_SSL` so that transport security is enabled without also mandating SASL authentication. The property is `required`, so existing flows already persist an explicit value and are unaffected. The new default primarily impacts newly added components on the canvas.

## Verification

Built `nifi-kafka-shared` and related modules and ran the unit tests for `nifi-kafka-shared`, `nifi-kafka-service-shared`, `nifi-kafka-3-service`, `nifi-kafka-service-aws` and `nifi-kafka-processors` — all pass. Tests that exercise the protocol set it explicitly, so none depend on the previous default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/AbstractKafkaBaseIT.java
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/AbstractKafkaBaseIT.java
@@ -23,6 +23,7 @@ import org.apache.nifi.json.JsonRecordSetWriter;
 import org.apache.nifi.json.JsonTreeReader;
 import org.apache.nifi.kafka.service.Kafka3ConnectionService;
 import org.apache.nifi.kafka.service.api.KafkaConnectionService;
+import org.apache.nifi.kafka.shared.property.SecurityProtocol;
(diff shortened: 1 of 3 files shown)


exceptionfactory (approved):
Thanks for proposing this change @wojtre. Based on the issue described with the Kafka Client behavior and TLS, plus the use of TLS as a general best practice, this change looks good. +1 merging

PR #11396, opened 2026-07-06
NIFI-16078 - Fix StandardHashiCorpVaultClientService for non-token authentication

pvillard31:
# Summary

NIFI-16078 - Fix StandardHashiCorpVaultClientService for non-token authentication

In NiFi 2.10.0, `spring-vault-core` was upgraded 4.0.1 → 4.1.0. Non-token auth methods now build their client via `AbstractVaultConfiguration.vaultClient()`, which looks up a `clientHttpRequestFactoryWrapper` bean; AWS EC2 / Azure MSI also go through `restClient()` → `getRestTemplateFactory()`. Alsok, NiFi's `getRestTemplateFactory()` override was removed.

NiFi configures `HashiCorpVaultConfiguration` programmatically with an empty `StaticApplicationContext`, so these bean lookups fail. Token authentication is unaffected (it needs no Vault client).

With this change, `HashiCorpVaultConfiguration.setClientHttpRequestFactory(...)` registers the request factory as the `clientHttpRequestFactoryWrapper` bean. We also restore the `getRestTemplateFactory()` override built from that same factory. And `StandardHashiCorpVaultCommunicationService` passes its configured (SSL Context Service–aware) factory before `clientAuthentication()`, so the auth client uses the same TLS material.

Added new unit tests cover Kubernetes and AWS EC2 authentication; both reproduce the reported failure without the fix.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-hashicorp-vault/src/test/java/org/apache/nifi/vault/hashicorp/TestHashiCorpVaultConfiguration.java
+++ b/nifi-commons/nifi-hashicorp-vault/src/test/java/org/apache/nifi/vault/hashicorp/TestHashiCorpVaultConfiguration.java
@@ -25,6 +25,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.StandardEnvironment;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.vault.authentication.ClientAuthentication;
 import org.springframework.vault.client.VaultEndpoint;
 import org.springframework.vault.core.VaultKeyValueOperationsSupport;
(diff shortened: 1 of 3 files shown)


nifi-commons/nifi-hashicorp-vault/src/test/java/org/apache/nifi/vault/hashicorp/TestHashiCorpVaultConfiguration.java
```diff
@@ -148,6 +149,60 @@ public void testBasicProperties() {
         this.runTest("http");
     }
 
+    @Test
+    public void testKubernetesClientAuthentication() throws IOException {
+        File k8sAuthProps = null;
+        Path tokenFile = null;
+        try {
+            tokenFile = Files.createTempFile("vault-k8s-token", ".jwt");
```
dan-s1: Couldn't you use JUnit 5 `TempDir` and thereby not need the `finally` statement below since created files are automatically cleaned up by JUnit5. Ditto for the next test.

exceptionfactory (approved):
Thanks for resolving this issue @pvillard31, the fix looks good.

PR #11397, opened 2026-07-06
NIFI-16080 Switch to JDBC Enquote Identifier in UpdateDatabaseTable

exceptionfactory:
# Summary

[NIFI-16080](https://issues.apache.org/jira/browse/NIFI-16080) Adjusts the implementation of database identifier quoting in the `UpdateDatabaseTable` Processor to use the JDBC `Statement.enquoteIdentifier()` method instead of the string concatenation method. Existing tests exercise the quoting behavior and confirm expected results. The adjusted approach maintains the configurable property on the Processor to enable or disable identifier quoting as needed.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UpdateDatabaseTable.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UpdateDatabaseTable.java
@@ -588,16 +588,17 @@ public class UpdateDatabaseTable extends AbstractProcessor {
                 }
 
                 if (!columnsToAdd.isEmpty()) {
-                    final List<ColumnDefinition> columnDefinitions = columnsToAdd.stream().map(columnDescription ->
-                            new StandardColumnDefinition(
-                                    enquoteIdentifier(s, columnDescription.getColumnName(), quoteColumnNames),
-                                    columnDescription.getDataType(),
-                                    columnDescription.isNullable() ? ColumnDefinition.Nullable.YES : ColumnDefinition.Nullable.UNKNOWN,
-                                    columnDescription.isRequired()
-                            )
-                            )
-                            .map(ColumnDefinition.class::cast)
-                            .toList();
+                    final List<ColumnDefinition> columnDefinitions = new ArrayList<>(columnsToAdd.size());
+                    for (final ColumnDescription columnDescription : columnsToAdd) {
+                        final ColumnDefinition columnDefinition = new StandardColumnDefinition(
+                                enquoteIdentifier(s, columnDescription.getColumnName(), quoteColumnNames),
+                                columnDescription.getDataType(),
+                                columnDescription.isNullable() ? ColumnDefinition.Nullable.YES : ColumnDefinition.Nullable.UNKNOWN,
+                                columnDescription.isRequired()
+                        );
+                        columnDefinitions.add(columnDefinition);
+                    }
+
                     final String qualifiedTableName = enquoteIdentifier(s, tableName, quoteTableName);
                     final TableDefinition tableDefinition = new TableDefinition(Optional.empty(), Optional.empty(), qualifiedTableName, columnDefinitions);
                     final StatementRequest statementRequest = new StandardStatementRequest(StatementType.ALTER, tableDefinition);
@@ -708,17 +709,13 @@ public class UpdateDatabaseTable extends AbstractProcessor {
         );
     }
 
-    private String enquoteIdentifier(final Statement statement, final String identifier, final boolean quoteIdentifier) {
+    private String enquoteIdentifier(final Statement statement, final String identifier, final boolean quoteIdentifier) throws SQLException {
         final String enquoted;
 
         if (identifier == null) {
             enquoted = identifier;
         } else if (quoteIdentifier) {
-            try {
-                enquoted = statement.enquoteIdentifier(identifier, quoteIdentifier);
-            } catch (final SQLException e) {
-                throw new IllegalArgumentException("Enquote Identifier [%s] failed".formatted(identifier), e);
-            }
+            enquoted = statement.enquoteIdentifier(identifier, quoteIdentifier);
         } else {
             enquoted = identifier;
         }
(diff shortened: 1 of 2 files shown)


nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UpdateDatabaseTable.java
```diff
@@ -708,11 +708,22 @@ private TableDefinition getTableDefinition(final TableSchema tableSchema) {
...
+    private String enquoteIdentifier(final Statement statement, final String identifier, final boolean quoteIdentifier) {
+        final String enquoted;
+
+        if (identifier == null) {
+            enquoted = identifier;
+        } else if (quoteIdentifier) {
+            try {
+                enquoted = statement.enquoteIdentifier(identifier, quoteIdentifier);
+            } catch (final SQLException e) {
+                throw new IllegalArgumentException("Enquote Identifier [%s] failed".formatted(identifier), e);
```
pvillard31: Statement.enquoteIdentifier throws for identifiers longer than 128 characters or with invalid characters, and this wraps it into an IllegalArgumentException. Since checkAndUpdateTableSchema runs inside the try that only routes IOException and SQLException to failure, would such an identifier now cause a rollback instead of routing to REL_FAILURE?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UpdateDatabaseTable.java
```diff
@@ -537,14 +537,14 @@ private synchronized OutputMetadataHolder checkAndUpdateTableSchema(
...
                         getLogger().debug("Adding column {} to table {}", columnName, tableName);
                     }
 
-                    final String quotedCatalogName = enquoteIdentifier(catalogName, quoteString, quoteTableName);
-                    final String quotedSchemaName = enquoteIdentifier(schemaName, quoteString, quoteTableName);
-                    final String quotedTableName = enquoteIdentifier(tableName, quoteString, quoteTableName);
+                    final String quotedCatalogName = enquoteIdentifier(s, catalogName, quoteTableName);
+                    final String quotedSchemaName = enquoteIdentifier(s, schemaName, quoteTableName);
+                    final String quotedTableName = enquoteIdentifier(s, tableName, quoteTableName);
                     tableSchema = new TableSchema(quotedCatalogName, quotedSchemaName, quotedTableName, columns, translateFieldNames, normalizer, primaryKeyColumnNames, quoteString);
```
pvillard31, line 548: The quote character now comes from the driver's enquoteIdentifier, but quoteString from getIdentifierQuoteString is still passed to TableSchema. Could these two produce different quote characters for a driver that does not override the default enquoteIdentifier?

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestUpdateDatabaseTable.java
```diff
@@ -120,7 +120,7 @@ public void testCreateTable() throws Exception {
             assertColumnEquals(rs, "id", 1, "INTEGER");
             assertColumnEquals(rs, "fractional", 2, "DOUBLE PRECISION");
             assertColumnEquals(rs, "code", 3, "INTEGER");
-            assertColumnEquals(rs, "newField", 4, "DOUBLE PRECISION");
+            assertColumnEquals(rs, "spaced Field", 4, "DOUBLE PRECISION");
```
pvillard31, line 149: Would it be worth adding a test for an identifier that makes enquoteIdentifier throw (for example longer than 128 characters) to lock in the error routing behavior?

pvillard31 (changes requested):

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UpdateDatabaseTable.java
exceptionfactory: Other components still use the quote string from the `TableSchema`, and probably also need to be adjusted, but the initial goal was to keep the changes localized for `UpdateDatabaseTable`. I could look at expanding the scope of these changes if you think that would be better.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestUpdateDatabaseTable.java
exceptionfactory: Yes, that's a good suggestion given the implicit limitation, I will look at making adjustments.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/UpdateDatabaseTable.java
exceptionfactory: However, unless there is a known driver that does not override `enquoteIdentifier`, this seems very unlikely.

exceptionfactory: The only use of the `quotedIdentifierString` is in `PutDatabaseRecord`, and seems better to handle that separately.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestUpdateDatabaseTable.java
exceptionfactory: Pushed an update with an additional test to exercise the `SQLException` for a column exceeding 128 characters.

pvillard31 (approved):

PR #11399, opened 2026-07-06
NIFI-16081: Implement Migrate Connector Properties API

bobpaulin:
# Summary

[NIFI-16081](https://issues.apache.org/jira/browse/NIFI-16081)

Implement Migrate Connector Properties API
SEE [NIP-34](https://issues.apache.org/jira/projects/NIP/issues/NIP-34)

* Support for migrateProperties method in Connector
* Tests supporting Step and Connector Property updates

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-connector-mock-bundle/nifi-connector-mock/src/main/java/org/apache/nifi/mock/connector/migration/MockConnectorPropertyConfiguration.java
+++ b/nifi-connector-mock-bundle/nifi-connector-mock/src/main/java/org/apache/nifi/mock/connector/migration/MockConnectorPropertyConfiguration.java
@@ -1,0 +1,497 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nifi.mock.connector.migration;
+
+import org.apache.nifi.components.connector.ConnectorValueReference;
+import org.apache.nifi.components.connector.StepConfiguration;
+import org.apache.nifi.components.connector.StringLiteralValue;
+import org.apache.nifi.migration.ConnectorPropertyConfiguration;
+import org.apache.nifi.migration.ConnectorStepPropertyConfiguration;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Test-only {@link ConnectorPropertyConfiguration} implementation for exercising a Connector's
+ * {@link org.apache.nifi.components.connector.Connector#migrateProperties(ConnectorPropertyConfiguration) migrateProperties}
+ * method in unit tests. Tracks per-step property renames, removals, and additions, plus step-level renames,
+ * removals, and additions. Callers can drive the mock with either a plain string-literal shorthand map or a
+ * fully typed {@link ConnectorValueReference} map and then inspect the outcome via
+ * {@link #toMigrationResult()}, {@link #getStepNames()}, and {@link #forStep(String)}.
+ */
+public class MockConnectorPropertyConfiguration implements ConnectorPropertyConfiguration {
+
+    private final Map<String, StepConfiguration> stepProperties = new HashMap<>();
+    private final Set<String> initialStepNames;
+
+    private final Map<String, StepPropertyRenames> propertiesRenamed = new HashMap<>();
+    private final Map<String, Set<String>> propertiesRemoved = new HashMap<>();
+    private final Map<String, Set<String>> propertiesUpdated = new HashMap<>();
+
+    private final Map<String, String> renamedSteps = new HashMap<>();
+    private final Set<String> removedSteps = new HashSet<>();
+    private final Set<String> addedSteps = new HashSet<>();
+
+    private MockConnectorPropertyConfiguration(final Map<String, StepConfiguration> initialProperties) {
+        if (initialProperties != null) {
+            for (final Map.Entry<String, StepConfiguration> entry : initialProperties.entrySet()) {
+                final Map<String, ConnectorValueReference> copy = new HashMap<>();
+                final StepConfiguration stepConfig = entry.getValue();
+                if (stepConfig != null && stepConfig.getPropertyValues() != null) {
+                    copy.putAll(stepConfig.getPropertyValues());
+                }
+                stepProperties.put(entry.getKey(), new StepConfiguration(copy));
+            }
+        }
+        this.initialStepNames = Collections.unmodifiableSet(new HashSet<>(stepProperties.keySet()));
+    }
+
+    /**
+     * Creates a {@link MockConnectorPropertyConfiguration} pre-populated with typed {@link ConnectorValueReference}
+     * values for each step.
+     *
+     * @param initialProperties per-step {@link StepConfiguration} snapshots of {@link ConnectorValueReference} values
+     * @return a new configuration
+     */
+    public static MockConnectorPropertyConfiguration fromValueReferences(final Map<String, StepConfiguration> initialProperties) {
+        return new MockConnectorPropertyConfiguration(initialProperties);
+    }
+
+    /**
+     * Creates a {@link MockConnectorPropertyConfiguration} pre-populated with plain string-literal values for each
+     * step. Each supplied string is wrapped in a {@link StringLiteralValue}.
+     *
+     * @param initialProperties per-step {@link StepLiteralProperties} of raw string values
+     * @return a new configuration
+     */
+    public static MockConnectorPropertyConfiguration fromStringLiterals(final Map<String, StepLiteralProperties> initialProperties) {
+        final Map<String, StepConfiguration> converted = new HashMap<>();
+        if (initialProperties != null) {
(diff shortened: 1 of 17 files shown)


nifi-connector-mock-bundle/nifi-connector-mock/src/main/java/org/apache/nifi/mock/connector/migration/MockConnectorPropertyConfiguration.java
```diff
@@ -0,0 +1,354 @@
...
+        }
+
+        @Override
+        public String getStepName() {
+            return stepName;
+        }
+
+        @Override
+        public boolean renameProperty(final String propertyName, final String newName) {
+            trackRenamed(stepName, propertyName, newName);
```
pvillard31: Should trackRenamed only run after we confirm the property exists and the name actually changes, so a no-op rename does not show up in MigrationResult.propertiesRenamed()?

nifi-connector-mock-bundle/nifi-connector-mock/src/main/java/org/apache/nifi/mock/connector/migration/MockConnectorPropertyConfiguration.java
```diff
@@ -0,0 +1,354 @@
...
+                return false;
+            }
+            final ConnectorValueReference existing = properties.remove(propertyName);
+            properties.put(newName, existing);
+            return true;
+        }
+
+        @Override
+        public boolean removeProperty(final String propertyName) {
+            trackRemoved(stepName, propertyName);
```
pvillard31: Should trackRemoved only run when the property actually existed and was removed, to match the real StandardConnectorStepPropertyConfiguration behavior?

nifi-connector-mock-bundle/nifi-connector-mock/src/main/java/org/apache/nifi/mock/connector/migration/MockConnectorPropertyConfiguration.java
```diff
@@ -0,0 +1,354 @@
...
+
+/**
+ * Test-only {@link ConnectorPropertyConfiguration} implementation for exercising a Connector's
+ * {@link org.apache.nifi.components.connector.Connector#migrateProperties(ConnectorPropertyConfiguration) migrateProperties}
+ * method in unit tests. Tracks per-step property renames, removals, and additions, plus step-level renames,
+ * removals, and additions. Callers can drive the mock with either a plain string-literal shorthand map or a
+ * fully typed {@link ConnectorValueReference} map and then inspect the outcome via
+ * {@link #toMigrationResult()}, {@link #getStepNames()}, and {@link #forStep(String)}.
+ */
+public class MockConnectorPropertyConfiguration implements ConnectorPropertyConfiguration {
```
pvillard31, line 42: Would it help to add a small unit test for this class, given it is a new public test utility with no direct test coverage in this PR?

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/connectors/ConnectorPropertyMigrationIT.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    private ConnectorValueReferenceDTO createStringLiteralReference(final String value) {
+        final ConnectorValueReferenceDTO valueRef = new ConnectorValueReferenceDTO();
+        valueRef.setValueType(STRING_LITERAL_TYPE);
+        valueRef.setValue(value);
+        return valueRef;
+    }
+
+    private void switchOutNars() throws IOException {
```
pvillard31: Since PropertyMigrationIT already has an identical switchOutNars, moveNars, and findFile helper set, could this be extracted into a shared base class instead of duplicated here?

bobpaulin:
@pvillard31 thank you for the feedback.  I've made your suggested adjustments to the tests.

pvillard31 (approved):
Thanks @bobpaulin - the latest LGTM. I was wondering if it could be worth adding a way to have the source Bundle version and the target Bundle version for the connector (coordinates of the NAR - group/artifact/version) exposed to the migration code in case some logic to be implemented based on the source version. This may require API changes though so likely as a followup, thoughts?

@markap14 - do you also want to have a look?

bobpaulin:
Thank you @pvillard31 .  I think adding the source and target are interesting ideas but I don't see this ability in Controller Services or Processors (or maybe I'm missing something).  So I think it might be something to consider as a follow up task.  I expect this would have API implications beyond the scope of NIP-34. 

> Thanks @bobpaulin - the latest LGTM. I was wondering if it could be worth adding a way to have the source Bundle version and the target Bundle version for the connector (coordinates of the NAR - group/artifact/version) exposed to the migration code in case some logic to be implemented based on the source version. This may require API changes though so likely as a followup, thoughts?
> 
> @markap14 - do you also want to have a look?

pvillard31:
This does not exist for processors/controller services and I wish we had something like this before so just thought that it might be even more relevant in the context of connectors. But definitely agree that this is beyond the scope of this NIP.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/migration/StandardConnectorPropertyConfiguration.java
```diff
@@ -0,0 +1,172 @@
...
+public class StandardConnectorPropertyConfiguration implements ConnectorPropertyConfiguration {
+
+    private static final Logger logger = LoggerFactory.getLogger(StandardConnectorPropertyConfiguration.class);
+
+    private final Map<String, Map<String, ConnectorValueReference>> stepProperties;
+    private final Set<String> modifiedStepNames = new HashSet<>();
+    private final String componentDescription;
+    private boolean modified = false;
+
+    public StandardConnectorPropertyConfiguration(final Map<String, Map<String, ConnectorValueReference>> initialProperties, final String componentDescription) {
```
markap14: Might be worth creating an abstraction here instead of using `Map<String, Map<String, ConnectorValueReference>>` - generics whose types are generics, especially Map within Map get a little hairy to deal with.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
```diff
@@ -378,41 +380,71 @@ public void inheritConfiguration(final List<VersionedConfigurationStep> activeCo
...
+        getComponentLog().info("Working Flow Context has been rebuilt with independent configuration");
+        for (final String stepName : migratedWorkingProperties.keySet()) {
+            notifyStepConfigured(stepName);
         }
 
         logger.debug("Successfully inherited configuration for {}", this);
     }
 
-    private StepConfiguration createStepConfiguration(final VersionedConfigurationStep step) {
+    private Map<String, Map<String, ConnectorValueReference>> migrateProperties(final List<VersionedConfigurationStep> flowConfiguration) {
```
markap14: As noted above, probably worth considering an abstraction, rather than Maps of Maps.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/StandardConnectorNode.java
```diff
@@ -378,41 +380,71 @@ public void inheritConfiguration(final List<VersionedConfigurationStep> activeCo
...
-            setConfiguration(step.getName(), stepConfig, true);
+        // Tear down the working context that applyUpdate created aliased to active, and rebuild it around an
+        // independent configuration seeded from migratedWorkingProperties. Then fire onConfigurationStepConfigured
+        // for every step so renamed steps trigger the flow-builder callback under their new name and any
+        // value-derived flow state (resolved asset paths, secret values, etc.) is populated against the fresh
+        // working context.
+        destroyWorkingContext();
+        final MutableConnectorConfigurationContext workingConfigContext = createConfigurationContext(migratedWorkingProperties);
+        workingFlowContext = flowContextFactory.createWorkingFlowContext(identifier,
+            connectorDetails.getComponentLog(), workingConfigContext, flowContextBundle);
```
markap14: Not a big deal but hygiene-wise it's best to avoid line wraps for short lines like this.

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/connectors/ConnectorPropertyMigrationIT.java
```diff
@@ -0,0 +1,168 @@
...
+        assertNotNull(clientId, label + " configuration is missing " + CLIENT_ID_NAME);
+        assertEquals(STRING_LITERAL_TYPE, clientId.getValueType());
+        assertEquals(CLIENT_ID_MIGRATED_VALUE, clientId.getValue());
+
+        assertFalse(migrated.containsKey(LEGACY_OBSOLETE_NAME),
+            label + " configuration still contains obsolete legacy property");
+        assertFalse(migrated.containsKey(LEGACY_BROKER_URL_NAME),
+            label + " configuration still contains legacy broker URL name");
+        assertFalse(migrated.containsKey(LEGACY_CREDENTIALS_NAME),
+            label + " configuration still contains legacy credentials name");
```
markap14: Another example of overly aggressive line wraps - and making unnecessary assertions where the `assertFalse` already makes it clear that it shouldn't be there but is.

markap14 (commented):
Thanks @bobpaulin I made a couple comments inline. A few hygiene-related comments but the one that's most important I think is the passing around of embedded Maps. Probably makes sense to abstract that into some kind of `InitialPropertyConfiguration` object or something instead.

bobpaulin:
Thank you for the feedback @markap14 .  I was able to find an existing abstraction to fit the Map of Maps and all the formatting issues should also be addressed.

> Thanks @bobpaulin I made a couple comments inline. A few hygiene-related comments but the one that's most important I think is the passing around of embedded Maps. Probably makes sense to abstract that into some kind of `InitialPropertyConfiguration` object or something instead.

markap14 (approved):
Thanks @bobpaulin changes look good to me.

PR #11403, opened 2026-07-07
NIFI-16085 - Flow synchronization fails when a Parameter Provider-backed Parameter Context contains a parameter not flagged as provided

pvillard31:
# Summary

NIFI-16085 - Flow synchronization fails when a Parameter Provider-backed Parameter Context contains a parameter not flagged as provided

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizerTest.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizerTest.java
@@ -24,11 +24,13 @@ import org.apache.nifi.components.connector.ConnectorState;
 import org.apache.nifi.components.connector.ConnectorSyncMode;
 import org.apache.nifi.components.connector.ConnectorSyncResult;
 import org.apache.nifi.controller.FlowController;
+import org.apache.nifi.controller.ParameterProviderNode;
 import org.apache.nifi.controller.ReportingTaskNode;
 import org.apache.nifi.controller.SnippetManager;
 import org.apache.nifi.controller.UninheritableFlowException;
 import org.apache.nifi.controller.flow.FlowManager;
 import org.apache.nifi.controller.flow.VersionedDataflow;
+import org.apache.nifi.controller.parameter.ParameterProviderLookup;
 import org.apache.nifi.controller.service.ControllerServiceNode;
 import org.apache.nifi.controller.service.ControllerServiceProvider;
 import org.apache.nifi.encrypt.PropertyEncryptor;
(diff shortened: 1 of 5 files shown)


kevdoran:
reviewing...

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizerTest.java
```diff
@@ -276,6 +290,70 @@ private VersionedControllerService getVersionedControllerService() {
...
+        setFlowController();
+
+        final String providerId = "provider-1";
+        final String contextName = "openflow-rds-ingest";
+        final String paramName = "db.host";
+
+        // Parameter Provider infrastructure
+        final ParameterProvider parameterProvider = mock(ParameterProvider.class);
+        when(parameterProvider.getIdentifier()).thenReturn(providerId);
+        final ParameterProviderNode parameterProviderNode = mock(ParameterProviderNode.class);
```
kevdoran, line 307: Minor / non-blocking: this `ParameterProviderNode` mock doesn't stub a validation status, so `getParameterProviderValidationStatus(...)` returns `null`, `getProvidedParameters(...)` returns an empty map, and the assertion path ends up exercising the "provided parameter not found" branch in `createParameterMap` (the reconciled value becomes `null`) rather than the provider value-sourcing branch described in the comment above `assertDoesNotThrow`. The current assertions (no-throw and `provided == true`) still hold, but the value-sourcing path isn't actually covered. If you'd like to lock that in, stubbing the provider node to `VALID` with a fetched parameter group and asserting the reconciled value would exercise it end to end.

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/parameters/ClusteredProviderParamFlowSyncIT.java
```diff
@@ -0,0 +1,191 @@
...
+        node1.start(true);
+        node2.start(true);
+
+        // 4. The cluster forms: both nodes connect (this would time out if reconciliation still failed the guard),
+        //    and neither node logs the provider-backed-context guard failure or a flow mismatch.
+        waitForAllNodesConnected();
+
+        final List<NiFiInstance> nodes = List.of(node1, node2);
+        assertNoNodeLogContains(nodes, GUARD_MESSAGE);
+        assertNoNodeLogContains(nodes, "did not Match Cluster Flow");
```
kevdoran, line 141: Non-blocking: the IT confirms the nodes reconnect and the guard message is absent, which is the key regression guard. As optional strengthening, asserting the parameter's value resolves to the provider-supplied value (rather than null or the corrupted value) after reconnection would make the "reconciled as provider-managed" guarantee explicit.

kevdoran (approved):
Thanks for tracking this down. The fix looks correct and well-targeted — treating the context-level Parameter Provider binding as authoritative over the per-`VersionedParameter` `provided` flag is the right call, and threading `providerBacked` through the `createParameter`/`createParameterMap` paths covers the reconcile entry points I checked (with `createParameterContextWithoutReferences` intentionally keeping the serialized flag). The forced-provided path also degrades gracefully when the provider doesn't currently supply a name — `createParameterMap` logs and falls back to a null value rather than failing.

A couple of non-blocking suggestions on test coverage below; neither affects the correctness of the change.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/test/java/org/apache/nifi/controller/serialization/VersionedFlowSynchronizerTest.java
pvillard31: Pushed a commit to address your comment

nifi-system-tests/nifi-system-test-suite/src/test/java/org/apache/nifi/tests/system/parameters/ClusteredProviderParamFlowSyncIT.java
pvillard31: Pushed a commit to address your comment

PR #11407, opened 2026-07-08
NIFI-16088 Add scheduling diagnostics for cron primary-node-only processors

kevdoran:
# Summary

Relates to [NIFI-16088](https://issues.apache.org/jira/browse/NIFI-16088).

Cron triggers are scheduled as one-shot tasks that reschedule only after a successful run (`CronSchedulingAgent`). If the triggered task throws, or scheduling the next trigger fails, the scheduling chain silently terminates: the component stays `RUNNING` but never fires again on its schedule until it is stopped and started — with nothing in the logs, because the exception is discarded by the scheduled executor.

This adds **diagnostics only** (no scheduling or lifecycle behavior is changed):

- `CronSchedulingAgent`: log a WARN (before propagating) when a cron-scheduled trigger throws, or when rescheduling the next trigger fails, noting the component will not run again on its schedule until it is stopped and started.
- `ConnectableTask`: log at INFO (was DEBUG) when a cron-driven, Primary-Node-Only component's fire is skipped because the node is not the primary node; timer-driven components stay at DEBUG to avoid noise, since cron fires are comparatively rare.

Motivated by a case where Primary-Node-Only cron processors silently stopped firing after a heartbeat-induced disconnect/reconnect and had to be manually stopped/started; the existing logs gave no signal as to why.

# Verification

- New unit test in `CronSchedulingAgentTest` asserting a triggered exception still propagates (is not silently swallowed).
- `./mvnw -pl nifi-framework-bundle/nifi-framework/nifi-framework-core -am test -Dtest=CronSchedulingAgentTest` — green (4 tests).
- `checkstyle:check` clean for the module.

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/scheduling/CronSchedulingAgent.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/scheduling/CronSchedulingAgent.java
@@ -129,8 +129,17 @@ public class CronSchedulingAgent extends AbstractTimeBasedSchedulingAgent {
                     try {
                         continuallyRunTask.invoke();
                     } catch (final RuntimeException re) {
+                        // Diagnostic: an exception here skips the rescheduling below. Because each cron trigger is a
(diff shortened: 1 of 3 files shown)


pvillard31 (approved):
Changes are LGTM, curious to see what will be the explanation for the reported problem.

PR #11414, opened 2026-07-10
NIFI-16098 Prevent content-claim truncation from dropping content referenced by a live sibling FlowFile

pvillard31:
# Summary

NIFI-16098 Prevent content-claim truncation from dropping content referenced by a live sibling FlowFile

A FlowFile created and removed within a single session (e.g. ExecuteGroovyScript with Failure Strategy 'transfer to failure', which clones on get and removes the clone at commit) yields one repository record whose type transitions CREATE->DELETE. The truncation reference count is incremented only for a committed CREATE, so that increment never ran, but the DELETE still decremented it. For a content claim shared with a live sibling FlowFile this drove the truncation reference count to zero, allowing the claim to be truncated and the sibling's content to be lost.

Only decrement the truncation reference count on delete of a persisted FlowFile (one that has an original claim).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


pkelly-nifi:
Thanks for fixing this so quickly, @pvillard31.  I just built a copy of NiFi 2.10.0 incorporating your changes and it seems to be working correctly now.

markap14 (approved):
Thanks for the fix @pvillard31 and for reporting @pkelly-nifi ! Changes look good to me. +1 will merge

PR #11426, opened 2026-07-13
NIFI-16021 Improved JoltTransformJSON error message when specified JOLT spec is invalid and ensured leading single and multi-line Java comments are supported when the Jolt spec is specified as text. 

dan-s1:
# Summary

[NIFI-16021](https://issues.apache.org/jira/browse/NIFI-16021)

Also added tests to ensure ValidateJson causes Json parsing exception when any Java comments is specified before or inside a schema.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-jolt-bundle/nifi-jolt-processors/src/main/java/org/apache/nifi/processors/jolt/AbstractJoltTransform.java
+++ b/nifi-extension-bundles/nifi-jolt-bundle/nifi-jolt-processors/src/main/java/org/apache/nifi/processors/jolt/AbstractJoltTransform.java
@@ -256,7 +256,7 @@ public abstract class AbstractJoltTransform extends AbstractProcessor {
         try (final BufferedReader reader = new BufferedReader(new InputStreamReader(resourceReference.read(), StandardCharsets.UTF_8))) {
             return reader.lines().collect(Collectors.joining(System.lineSeparator()));
         } catch (final IOException e) {
-            throw new UncheckedIOException("Read JOLT Transform failed", e);
+            throw new UncheckedIOException("Read JOLT Transform failed, reason: " + e.getMessage(), e);
         }
     }
 
--- a/nifi-extension-bundles/nifi-jolt-bundle/nifi-jolt-processors/src/main/java/org/apache/nifi/processors/jolt/AbstractJoltTransform.java
+++ b/nifi-extension-bundles/nifi-jolt-bundle/nifi-jolt-processors/src/main/java/org/apache/nifi/processors/jolt/AbstractJoltTransform.java
@@ -190,8 +190,15 @@ public abstract class AbstractJoltTransform extends AbstractProcessor {
                     }
                 }
             } catch (final Exception e) {
-                String message = String.format("Specification not valid for the selected transformation: %s",
-                        (e.getCause() != null ? e.getCause().getMessage() : e.getMessage()));
+                final String reason;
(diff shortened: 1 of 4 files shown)


exceptionfactory (changes requested):
Thanks @dan-s1. Although changing the message itself is one approach, it seems better to handle the formatting in the custom validation method. This would align with the general practice of avoiding including exception cause messages. It would also avoid changing runtime behavior for the sake of the validation method.

dan-s1:
@exceptionfactory Please let me know if the latest change is what you had in mind. Thanks!

nifi-extension-bundles/nifi-jolt-bundle/nifi-jolt-processors/src/main/java/org/apache/nifi/processors/jolt/AbstractJoltTransform.java
```diff
@@ -190,7 +190,8 @@ protected Collection<ValidationResult> customValidate(ValidationContext validati
                     }
                 }
             } catch (final Exception e) {
-                String message = String.format("Specification not valid for the selected transformation: %s", e.getMessage());
+                String message = String.format("Specification not valid for the selected transformation: %s",
+                        (e.getCause() != null ? e.getCause().getMessage() : e.getMessage()));
```
exceptionfactory: For readability, I recommend breaking out the ternary check into a variable declaration along the lines of:
```
final String reason;
final Throwable cause = e.getCause();
if (cause == null) {
    reason = e.getMessage();
} else {
    reason = "%s [%s]".formatted(cause.getMessage(), e.getMessage);
}
message = "Specification not valid... %s".formatted(reason)
```

This approach should be a bit easier to follow, and also include both the direct message and the cause message when available.

exceptionfactory (changes requested):
Thanks @dan-s1, the revised approach looks closer to what I had in mind. I recommendation a small adjustment.

dan-s1:
@exceptionfactory When you have a chance can you please have another look at the changes? Thanks!

nifi-extension-bundles/nifi-jolt-bundle/nifi-jolt-processors/src/main/java/org/apache/nifi/processors/jolt/AbstractJoltTransform.java
```diff
@@ -190,7 +190,15 @@ protected Collection<ValidationResult> customValidate(ValidationContext validati
...
-                String message = String.format("Specification not valid for the selected transformation: %s", e.getMessage());
+                final String reason;
+                final Throwable cause = e.getCause();
+                if (cause == null) {
+                    reason = e.getMessage();
+                } else {
+                    reason = "%s [%s]".formatted(cause.getMessage(), e.getMessage());
+                }
+
+                final String message = "Specification not valid... %s".formatted(reason);
```
exceptionfactory:
```suggestion
                final String message = "Specification not valid: %s".formatted(reason);
```

exceptionfactory (changes requested):
Thanks @dan-s1. I noted one remaining minor wording adjustment, then this looks ready to go.

exceptionfactory (approved):
Thanks @dan-s1, the latest version looks good!

PR #11430, opened 2026-07-14
NIFI-16113 Adds new processor for changing visibility timeouts of mes…

pkelly-nifi:
…sages from SQS

# Summary

[NIFI-16113](https://issues.apache.org/jira/browse/NIFI-16113)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [X] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
+++ b/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
@@ -1,0 +1,142 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.aws.sqs;
+
+import org.apache.nifi.annotation.behavior.InputRequirement;
+import org.apache.nifi.annotation.behavior.InputRequirement.Requirement;
+import org.apache.nifi.annotation.behavior.SupportsBatching;
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.SeeAlso;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.expression.ExpressionLanguageScope;
+import org.apache.nifi.flowfile.FlowFile;
+import org.apache.nifi.flowfile.attributes.CoreAttributes;
+import org.apache.nifi.processor.ProcessContext;
+import org.apache.nifi.processor.ProcessSession;
+import org.apache.nifi.processor.exception.ProcessException;
+import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.processors.aws.AbstractAwsSyncProcessor;
+import software.amazon.awssdk.services.sqs.SqsClient;
+import software.amazon.awssdk.services.sqs.SqsClientBuilder;
+import software.amazon.awssdk.services.sqs.model.ChangeMessageVisibilityBatchRequest;
+import software.amazon.awssdk.services.sqs.model.ChangeMessageVisibilityBatchRequestEntry;
+import software.amazon.awssdk.services.sqs.model.ChangeMessageVisibilityBatchResponse;
+
+import java.util.List;
+
+import static org.apache.nifi.processors.aws.region.RegionUtil.CUSTOM_REGION;
+import static org.apache.nifi.processors.aws.region.RegionUtil.REGION;
+
+@SupportsBatching
+@SeeAlso({GetSQS.class, PutSQS.class})
+@InputRequirement(Requirement.INPUT_REQUIRED)
+@Tags({"Amazon", "AWS", "SQS", "Queue", "Update"})
+@CapabilityDescription("Updates the visibility timeout for a message from an Amazon Simple Queuing Service Queue")
+public class ChangeSQSMessageVisibilityTimeout extends AbstractAwsSyncProcessor<SqsClient, SqsClientBuilder> {
+
+    public static final PropertyDescriptor QUEUE_URL = new PropertyDescriptor.Builder()
+            .name("Queue URL")
+            .description("The URL of the queue delete from")
+            .addValidator(StandardValidators.URL_VALIDATOR)
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .required(true)
+            .build();
+
+    public static final PropertyDescriptor RECEIPT_HANDLE = new PropertyDescriptor.Builder()
+            .name("Receipt Handle")
+            .description("The identifier that specifies the receipt of the message")
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .required(true)
+            .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+            .defaultValue("${sqs.receipt.handle}")
+            .build();
+
+    public static final PropertyDescriptor VISIBILITY_TIMEOUT = new PropertyDescriptor.Builder()
+            .name("Visibility Timeout")
+            .description("The amount of time after a message is received but not deleted that the message is hidden from other consumers")
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .required(true)
+            .defaultValue("15 mins")
+            .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
+            .build();
+
+    public static final List<PropertyDescriptor> PROPERTY_DESCRIPTORS = List.of(
+        QUEUE_URL,
+        VISIBILITY_TIMEOUT,
+        REGION,
+        CUSTOM_REGION,
+        AWS_CREDENTIALS_PROVIDER_SERVICE,
+        SSL_CONTEXT_SERVICE,
+        RECEIPT_HANDLE,
+        TIMEOUT,
+        ENDPOINT_OVERRIDE,
+        PROXY_CONFIGURATION_SERVICE
+    );
+
+    @Override
+    protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+        return PROPERTY_DESCRIPTORS;
+    }
+
+    @Override
+    public void onTrigger(final ProcessContext context, final ProcessSession session) {
+        FlowFile flowFile = session.get();
+        if (flowFile == null) {
+            return;
+        }
+
+        final String queueUrl = context.getProperty(QUEUE_URL).evaluateAttributeExpressions(flowFile).getValue();
+
+        final SqsClient client = getClient(context);
+
+        final String receiptHandle = context.getProperty(RECEIPT_HANDLE).evaluateAttributeExpressions(flowFile).getValue();
+        final int visibilityTimeout = Math.toIntExact(context.getProperty(VISIBILITY_TIMEOUT).evaluateAttributeExpressions(flowFile).asDuration().toSeconds());
+        final String entryId = flowFile.getAttribute(CoreAttributes.UUID.key());
+        final ChangeMessageVisibilityBatchRequestEntry entry = ChangeMessageVisibilityBatchRequestEntry.builder()
+                .receiptHandle(receiptHandle)
+                .id(entryId)
+                .visibilityTimeout(visibilityTimeout)
+                .build();
+
+        final ChangeMessageVisibilityBatchRequest request = ChangeMessageVisibilityBatchRequest.builder()
+                .queueUrl(queueUrl)
+                .entries(entry)
+                .build();
+
+        try {
+            ChangeMessageVisibilityBatchResponse response = client.changeMessageVisibilityBatch(request);
+
+            // check for errors
+            if (!response.failed().isEmpty()) {
+                throw new ProcessException(response.failed().getFirst().toString());
+            }
+
+            getLogger().info("Successfully updated visibility timeout for SQS message for {}", flowFile);
+            session.transfer(flowFile, REL_SUCCESS);
+        } catch (final Exception e) {
+            getLogger().error("Failed to update visiibility timeout for SQS message", e);
+            flowFile = session.penalize(flowFile);
+            session.transfer(flowFile, REL_FAILURE);
+        }
+    }
+
+    @Override
+    protected SqsClientBuilder createClientBuilder(final ProcessContext context) {
+        return SqsClient.builder();
+    }
+}
--- a/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
+++ b/nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
@@ -28,7 +28,6 @@ import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.flowfile.attributes.CoreAttributes;
 import org.apache.nifi.processor.ProcessContext;
 import org.apache.nifi.processor.ProcessSession;
-import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.apache.nifi.processors.aws.AbstractAwsSyncProcessor;
 import software.amazon.awssdk.services.sqs.SqsClient;
@@ -49,15 +48,15 @@ import static org.apache.nifi.processors.aws.region.RegionUtil.REGION;
 @CapabilityDescription("Updates the visibility timeout for a message from an Amazon Simple Queuing Service Queue")
 public class ChangeSQSMessageVisibilityTimeout extends AbstractAwsSyncProcessor<SqsClient, SqsClientBuilder> {
 
-    public static final PropertyDescriptor QUEUE_URL = new PropertyDescriptor.Builder()
+    static final PropertyDescriptor QUEUE_URL = new PropertyDescriptor.Builder()
             .name("Queue URL")
-            .description("The URL of the queue delete from")
+            .description("The URL of the queue for which the message's visibility timeout should be updated")
             .addValidator(StandardValidators.URL_VALIDATOR)
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
             .required(true)
             .build();
 
-    public static final PropertyDescriptor RECEIPT_HANDLE = new PropertyDescriptor.Builder()
+    static final PropertyDescriptor RECEIPT_HANDLE = new PropertyDescriptor.Builder()
             .name("Receipt Handle")
             .description("The identifier that specifies the receipt of the message")
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
@@ -66,7 +65,7 @@ public class ChangeSQSMessageVisibilityTimeout extends AbstractAwsSyncProcessor<
             .defaultValue("${sqs.receipt.handle}")
             .build();
 
-    public static final PropertyDescriptor VISIBILITY_TIMEOUT = new PropertyDescriptor.Builder()
+    static final PropertyDescriptor VISIBILITY_TIMEOUT = new PropertyDescriptor.Builder()
             .name("Visibility Timeout")
             .description("The amount of time after a message is received but not deleted that the message is hidden from other consumers")
             .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
@@ -75,7 +74,7 @@ public class ChangeSQSMessageVisibilityTimeout extends AbstractAwsSyncProcessor<
             .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
             .build();
 
-    public static final List<PropertyDescriptor> PROPERTY_DESCRIPTORS = List.of(
+    private static final List<PropertyDescriptor> PROPERTY_DESCRIPTORS = List.of(
         QUEUE_URL,
         VISIBILITY_TIMEOUT,
         REGION,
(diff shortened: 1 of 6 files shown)


exceptionfactory (changes requested):
Thanks for proposing this new Processor @pkelly-nifi.

On initial review, it seems like a narrow use case that is not well suited for inclusion as part of the standard project distribution.

To understand the use case a bit more, is setting the visibility timeout that should be included on the PutSQS Processor as an optional property, as opposed to a separate Processor?

pkelly-nifi:
Hi @exceptionfactory, thank you for your input.  Admittedly this is a narrow use case, but it can be important.

To answer your question, setting the visibility timeout in PutSQS processor can't be done.  There is no option in the SQS API for setting the visibility timeout on the message during the SendMessage operation, so it's not something we could incorporate into PutSQS even if we wanted to.  You can set a message delay when sending a message, but not a visibility timeout.  The difference is that a delay affects the time between when you send the message and when a client can first receive that message, whereas the visibility timeout is the interval between when a client first receives the message (but does not delete it) and when the message is considered to be a failure and SQS automatically replays it to another client.  It will continue to be replayed at the visibility timeout interval until it is deleted by a client or considered to be a dead letter by the service, based on the configuration of the queue.

Visibility timeouts can be set in three ways:
1) It can be set when you receive the message.  NiFi currently does this in GetSQS during the ReceiveMessage call, before you know anything about the message, so all messages retrieved by a single GetSQS use a common visibility timeout.
2) The timeout can be updated after receiving the message in a ChangeMessageVisibilityTimeout call.  NiFi does not currently support this, but this new processor does.
3) A default visibility timeout can be set for the queue, but that doesn't really affect NiFi since the default visibility timeout is overwritten by the GetSQS processor when the message is retrieved.


I think there are two important use case for this processor.

First, being able to set a per-message visibility timeout for S3 event notifications, which something NiFi cannot currently do.  Let's say you have an S3 bucket that normally receives a mix of small files (on the order of a few MBs), and very large files (hundreds of GBs or into TBs), and you have S3 set up to send you event notifications to SQS for objects that are uploaded.  You drive the flow for processing those files using a GetSQS processor to receive these event notifications.  You want to be able to very quickly retry any small files that fail, but you need to allow a lot of time for processing the big files.  Amazon's documentation recommends calling ChangeMessageVisibilityTimeout in order to set appropriate visibility timeouts for each message, depending on processing needs:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html

This new processor would allow a flow designer to follow Amazon's recommendations, and allow for better retry performance for the mixed data sizes.  Currently, the flow designer would need to configure GetSQS to set a visibility timeout large enough to cover the slowest file, which for this case would be several hours.  Small files that fail would need to wait this long interval until their notifications are replayed.

What this new processor would allow is to set the visibility timeout in GetSQS to only a few minutes to cover the expected small files.  You would then feed the event notification messages into EvaluateJsonPath to extract the object's size from the message, then evaluate the size and optionally use this new processor to set a new visibility timeout for messages for large objects (SQS supports up to 12 hours) in order to give you enough time to process them.  Messages for small objects would replay on the shorter interval while allowing you enough time to process the large files without their messages getting replayed.

There are several other cases that are similar, but they are all derived from this same concept: you want to be able to set different replay intervals based on some characteristic of the message, which is something NiFi cannot currently do.

The second important case is about efficiently recovering from errors.  This new processor could be called to set a visibility timeout of "0 seconds", which would cause SQS to immediately replay the message to trigger the initial flow again.  This is useful if, for example, you know the file has failed in a way that requires processing it again from the beginning, but you don't want to wait for the original visibility timeout that GetSQS set to expire.  This is especially useful if the messages come from a queue to which you do not have SendMessage permission since you can't just resend the message to the queue in those cases, but updating the timeout to zero would allow the original message to be replayed.


I'm happy to discuss further, but I think this could be a handy, if not critical, feature for some flows.  I agree that the cases are pretty narrow, but they're important and currently cannot be done inside of NiFi without significant complication.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,142 @@
...
+import static org.apache.nifi.processors.aws.region.RegionUtil.REGION;
+
+@SupportsBatching
+@SeeAlso({GetSQS.class, PutSQS.class})
+@InputRequirement(Requirement.INPUT_REQUIRED)
+@Tags({"Amazon", "AWS", "SQS", "Queue", "Update"})
+@CapabilityDescription("Updates the visibility timeout for a message from an Amazon Simple Queuing Service Queue")
+public class ChangeSQSMessageVisibilityTimeout extends AbstractAwsSyncProcessor<SqsClient, SqsClientBuilder> {
+
+    public static final PropertyDescriptor QUEUE_URL = new PropertyDescriptor.Builder()
```
exceptionfactory: The `public` modifier can be removed from Property Descriptors.

```suggestion
    static final PropertyDescriptor QUEUE_URL = new PropertyDescriptor.Builder()
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,142 @@
...
+            .name("Queue URL")
+            .description("The URL of the queue delete from")
```
exceptionfactory: This description seems off

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,142 @@
...
+    public static final PropertyDescriptor VISIBILITY_TIMEOUT = new PropertyDescriptor.Builder()
+            .name("Visibility Timeout")
+            .description("The amount of time after a message is received but not deleted that the message is hidden from other consumers")
+            .expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)
+            .required(true)
+            .defaultValue("15 mins")
+            .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
+            .build();
+
+    public static final List<PropertyDescriptor> PROPERTY_DESCRIPTORS = List.of(
```
exceptionfactory:
```suggestion
    private static final List<PropertyDescriptor> PROPERTY_DESCRIPTORS = List.of(
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,142 @@
...
+
+        final ChangeMessageVisibilityBatchRequest request = ChangeMessageVisibilityBatchRequest.builder()
+                .queueUrl(queueUrl)
+                .entries(entry)
+                .build();
+
+        try {
+            ChangeMessageVisibilityBatchResponse response = client.changeMessageVisibilityBatch(request);
+
+            // check for errors
```
exceptionfactory: The comment seems unnecessary
```suggestion
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,142 @@
...
+            if (!response.failed().isEmpty()) {
+                throw new ProcessException(response.failed().getFirst().toString());
```
exceptionfactory: This exception would cause the message to remain in the input queue, is that the desired behavior? It seems like it should be routed to failure, or perhaps a new retry relationship.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+
+            getLogger().info("Successfully updated visibility timeout for SQS message for {}", flowFile);
```
exceptionfactory: Recommend including the value of the timeout in this message

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,142 @@
...
+            session.transfer(flowFile, REL_SUCCESS);
+        } catch (final Exception e) {
+            getLogger().error("Failed to update visiibility timeout for SQS message", e);
```
exceptionfactory: Recommend including the FlowFile in this message

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/ITChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,64 @@
...
+import software.amazon.awssdk.services.sqs.model.SendMessageRequest;
+import software.amazon.awssdk.services.sqs.model.SendMessageResponse;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class ITChangeSQSMessageVisibilityTimeout extends AbstractSQSIT {
```
exceptionfactory: The `public` modifier can be removed, and I recommend using `IT` as the class suffix

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,122 @@
...
+import software.amazon.awssdk.services.sqs.model.ChangeMessageVisibilityBatchRequest;
+import software.amazon.awssdk.services.sqs.model.ChangeMessageVisibilityBatchResponse;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestChangeSQSMessageVisibilityTimeout {
```
exceptionfactory:
```suggestion
class ChangeSQSMessageVisibilityTimeoutTest {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private TestRunner runner = null;
+    private SqsClient mockSQSClient = null;
+
+    @BeforeEach
+    public void setUp() {
```
exceptionfactory: `public` modifiers can be removed from all methods

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Test
+    public void testChangeVisibilityTimeoutWithCustomReceiptHandle() {
+        runner.setProperty(ChangeSQSMessageVisibilityTimeout.QUEUE_URL, "https://sqs.us-west-2.amazonaws.com/123456789012/test-queue-000000000");
+        runner.setProperty(ChangeSQSMessageVisibilityTimeout.RECEIPT_HANDLE, "${custom.receipt.handle}");
+        runner.setProperty(ChangeSQSMessageVisibilityTimeout.VISIBILITY_TIMEOUT, "${custom.timeout.value}");
+        final Map<String, String> ffAttributes = new HashMap<>();
+        ffAttributes.put("filename", "1.txt");
+        ffAttributes.put("custom.receipt.handle", "test-receipt-handle-1");
```
exceptionfactory: Recommend declaring static variables for reused strings

exceptionfactory (changes requested):
Thanks for the substantive reply and additional background @pkelly-nifi.

Based on the use case and AWS recommendations from the documentation, this seems like a focused Processor worth including.

I highlighted a handful of mostly minor adjustments on the code to move this forward.

pkelly-nifi:
Hi @exceptionfactory.  Thank you for your feedback and for considering this for inclusion into NiFi.  I just committed the updates you requested.

This new processor was based on the existing DeleteSQS processor, so many of your suggestions would be relevant to that one--and likely other SQS processors--as well.  Would it make sense for me to update them as part of a separate Jira and PR?

exceptionfactory:
> Hi @exceptionfactory. Thank you for your feedback and for considering this for inclusion into NiFi. I just committed the updates you requested.
> 
> This new processor was based on the existing DeleteSQS processor, so many of your suggestions would be relevant to that one--and likely other SQS processors--as well. Would it make sense for me to update them as part of a separate Jira and PR?

Thanks @pkelly-nifi, yes, if you could create a new Jira issue for similar syntax improvements for `DeleteSQS`, that would be a helpful improvement.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeoutIT.java
```diff
@@ -0,0 +1,64 @@
...
+import software.amazon.awssdk.services.sqs.model.SendMessageRequest;
+import software.amazon.awssdk.services.sqs.model.SendMessageResponse;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class ChangeSQSMessageVisibilityTimeoutIT extends AbstractSQSIT {
```
exceptionfactory: The `public` modifier can be removed

```suggestion
class ChangeSQSMessageVisibilityTimeoutIT extends AbstractSQSIT {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,126 @@
...
+import software.amazon.awssdk.services.sqs.model.ChangeMessageVisibilityBatchRequest;
+import software.amazon.awssdk.services.sqs.model.ChangeMessageVisibilityBatchResponse;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class TestChangeSQSMessageVisibilityTimeout {
```
exceptionfactory:
```suggestion
class ChangeSQSMessageVisibilityTimeoutTest {
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,141 @@
...
+
+        final ChangeMessageVisibilityBatchRequest request = ChangeMessageVisibilityBatchRequest.builder()
+                .queueUrl(queueUrl)
+                .entries(entry)
+                .build();
+
+        try {
+            ChangeMessageVisibilityBatchResponse response = client.changeMessageVisibilityBatch(request);
+
+            if (!response.failed().isEmpty()) {
```
exceptionfactory: I recommend reversing the logic so that this reads as `if (response.failed().isEmpty()) { // SUCCESS } else { // FAILURE }`

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/main/java/org/apache/nifi/processors/aws/sqs/ChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,141 @@
...
+                .visibilityTimeout(visibilityTimeout)
+                .build();
+
+        final ChangeMessageVisibilityBatchRequest request = ChangeMessageVisibilityBatchRequest.builder()
+                .queueUrl(queueUrl)
+                .entries(entry)
+                .build();
+
+        try {
+            ChangeMessageVisibilityBatchResponse response = client.changeMessageVisibilityBatch(request);
```
exceptionfactory:
```suggestion
            final ChangeMessageVisibilityBatchResponse response = client.changeMessageVisibilityBatch(request);
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private TestRunner runner = null;
+    private SqsClient mockSQSClient = null;
+
+    private static final String queueURL = "https://sqs.us-west-2.amazonaws.com/123456789012/test-queue-000000000";
+    private static final String filename = "1.txt";
+    private static final String receiptHandle = "test-receipt-handle-1";
```
exceptionfactory:
```suggestion
    private static final String QUEUE_URL = "https://sqs.us-west-2.amazonaws.com/123456789012/test-queue-000000000";
    private static final String FILENAME = "1.txt";
    private static final String RECEIPT_HANDLE = "test-receipt-handle-1";
```

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,126 @@
...
+        ffAttributes.put("sqs.receipt.handle", receiptHandle);
+        runner.enqueue("TestMessageBody", ffAttributes);
+
+        runner.assertValid();
+        runner.run(1);
+
+        ArgumentCaptor<ChangeMessageVisibilityBatchRequest> captureChangeMessageVisibilityTimeoutRequest = ArgumentCaptor.forClass(ChangeMessageVisibilityBatchRequest.class);
+        Mockito.verify(mockSQSClient, Mockito.times(1)).changeMessageVisibilityBatch(captureChangeMessageVisibilityTimeoutRequest.capture());
+        ChangeMessageVisibilityBatchRequest changeMessageVisibilityTimeoutRequest = captureChangeMessageVisibilityTimeoutRequest.getValue();
+        assertEquals("https://sqs.us-west-2.amazonaws.com/123456789012/test-queue-000000000", changeMessageVisibilityTimeoutRequest.queueUrl());
```
exceptionfactory: The expected value should use a static final variable.

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,126 @@
...
+        assertEquals("test-receipt-handle-1", changeMessageVisibilityTimeoutRequest.entries().getFirst().receiptHandle());
```
exceptionfactory: The expected value should use a static final variable

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,126 @@
...
+        final Map<String, String> ffAttributes = new HashMap<>();
+        ffAttributes.put("filename", filename);
+        ffAttributes.put("sqs.receipt.handle", receiptHandle);
+        runner.enqueue("TestMessageBody", ffAttributes);
+
+        runner.assertValid();
+        runner.run(1);
+
+        ArgumentCaptor<ChangeMessageVisibilityBatchRequest> captureChangeMessageVisibilityTimeoutRequest = ArgumentCaptor.forClass(ChangeMessageVisibilityBatchRequest.class);
+        Mockito.verify(mockSQSClient, Mockito.times(1)).changeMessageVisibilityBatch(captureChangeMessageVisibilityTimeoutRequest.capture());
```
exceptionfactory: The `Mockito.verify()` method should be statically imported

nifi-extension-bundles/nifi-aws-bundle/nifi-aws-processors/src/test/java/org/apache/nifi/processors/aws/sqs/TestChangeSQSMessageVisibilityTimeout.java
```diff
@@ -0,0 +1,126 @@
...
+        runner.assertAllFlowFilesTransferred(ChangeSQSMessageVisibilityTimeout.REL_SUCCESS, 1);
+    }
+
+    @Test
+    void testChangeVisibilityTimeoutWithCustomReceiptHandle() {
+        runner.setProperty(ChangeSQSMessageVisibilityTimeout.QUEUE_URL, queueURL);
+        runner.setProperty(ChangeSQSMessageVisibilityTimeout.RECEIPT_HANDLE, "${custom.receipt.handle}");
+        runner.setProperty(ChangeSQSMessageVisibilityTimeout.VISIBILITY_TIMEOUT, "${custom.timeout.value}");
+        final Map<String, String> ffAttributes = new HashMap<>();
+        ffAttributes.put("filename", filename);
```
exceptionfactory: The `filename` name should use CoreAttributeKeys in all methods

exceptionfactory (changes requested):
Thanks for the updates @pkelly-nifi, this looks close to completion. I noted a few more syntax recommendations and then this should be ready.

pkelly-nifi:
Thank you.  All of the requested changes have been committed.

exceptionfactory (commented):
Thanks @pkelly-nifi, the latest updates look good, planning to merge on successful build completion

exceptionfactory (approved):

PR #11432, opened 2026-07-14
NIFI-16114 Fix version change bug when controller service property has a parameter reference

esecules:
# Summary

[NIFI-16114](https://issues.apache.org/jira/browse/NIFI-16114) Fix version change bug that caused the controller service properties which have parameter values to have those parameters dereferenced leading to immediate "local changes" in my versioned flow.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [x] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
@@ -96,11 +96,13 @@ import org.apache.nifi.groups.VersionedComponentAdditions;
 import org.apache.nifi.logging.LogLevel;
 import org.apache.nifi.migration.ControllerServiceFactory;
 import org.apache.nifi.migration.StandardControllerServiceFactory;
+import org.apache.nifi.parameter.ExpressionLanguageAgnosticParameterParser;
 import org.apache.nifi.parameter.Parameter;
 import org.apache.nifi.parameter.ParameterContext;
 import org.apache.nifi.parameter.ParameterContextManager;
 import org.apache.nifi.parameter.ParameterContextNameUtils;
 import org.apache.nifi.parameter.ParameterDescriptor;
+import org.apache.nifi.parameter.ParameterParser;
 import org.apache.nifi.parameter.ParameterProviderConfiguration;
 import org.apache.nifi.parameter.ParameterReferenceManager;
 import org.apache.nifi.parameter.ParameterReferencedControllerServiceData;
@@ -164,6 +166,7 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
     public static final String ENC_PREFIX = "enc{";
     public static final String ENC_SUFFIX = "}";
 
+    private static final ParameterParser agnosticParameterParser = new ExpressionLanguageAgnosticParameterParser();
     private final VersionedFlowSynchronizationContext context;
     private final Set<String> updatedVersionedComponentIds = new HashSet<>();
     private final List<CreatedOrModifiedExtension> createdAndModifiedExtensions = new ArrayList<>();
@@ -1666,11 +1669,11 @@ public class StandardVersionedComponentSynchronizer implements VersionedComponen
                     || (versionedDescriptor != null && versionedDescriptor.getIdentifiesControllerService());
                 final boolean sensitive = (descriptor != null && descriptor.isSensitive())
                     || (versionedDescriptor != null && versionedDescriptor.isSensitive());
-
+                final String proposedValue = proposedProperties.get(propertyName);
                 final String value;
-                if (descriptor != null && referencesService && (proposedProperties.get(propertyName) != null)) {
+                if (descriptor != null && referencesService && proposedValue != null && !isReferencingParameter(proposedValue)) {
                     // Need to determine if the component's property descriptor for this service is already set to an id
-                    // of an existing service that is outside the current processor group, and if it is we want to leave
+                    // of an existing service that is outside the current processor group, and if it is, we want to leave
                     // the property set to that value
                     String existingExternalServiceId = null;
                     final String componentDescriptorValue = componentNode.getEffectivePropertyValue(descriptor);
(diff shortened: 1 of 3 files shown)


pvillard31 (changes requested):
Can you make sure that commit/PR follows the rules for the checks to execute properly? Also, did you consider a system test for it?

esecules:
> Can you make sure that commit/PR follows the rules for the checks to execute properly? Also, did you consider a system test for it?

@pvillard31 thanks for having a look. I've added a system test and I fixed up the commit message and PR title to match the rules.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -1653,11 +1655,13 @@ private Map<String, String> populatePropertiesMap(final ComponentNode componentN
                     || (versionedDescriptor != null && versionedDescriptor.getIdentifiesControllerService());
                 final boolean sensitive = (descriptor != null && descriptor.isSensitive())
                     || (versionedDescriptor != null && versionedDescriptor.isSensitive());
-
+                final String proposedValue = proposedProperties.get(propertyName);
+                final ParameterParser parameterParser = new ExpressionLanguageAgnosticParameterParser();
```
pvillard31: This parser is stateless and is created for every property on every synchronized component. Could it be a private static final field, or created once outside the loop?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
```diff
@@ -1653,11 +1655,13 @@ private Map<String, String> populatePropertiesMap(final ComponentNode componentN
...
+                final boolean proposedReferencingParameter = parameterParser.parseTokens(proposedValue).iterator().hasNext();
```
pvillard31: proposedReferencingParameter is only read inside the referencesService branch. Could the parseTokens call move into that branch so non controller service properties are not parsed?

pvillard31: Would toReferenceList().isEmpty() be more precise here than iterator().hasNext(), since the iterator also returns true for escape only tokens such as ##?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/test/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizerTest.java
```diff
@@ -1021,6 +1021,198 @@ public void testExternalControllerServiceReferenceRemoved() throws FlowSynchroni
         assertNull(properties.get("cs"));
     }
 
+    @Test
+    @SuppressWarnings("unchecked")
```
pvillard31: The test class already has a shared type safe propertiesCaptor field built with ArgumentCaptor.captor(). Could the new tests reuse it (or use captor()) so the @SuppressWarnings(unchecked) is not needed?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
esecules: Sorry for going MIA, I am back now. Im not opposed to either of the methods that return the list, I just wanted to clarify whether one could be empty while the other isnt. toList returns the token list while toReferencList returns the referenceList.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/test/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizerTest.java
esecules: When I remove the annotation I get a warning about `ArgumentCaptor.forClass(Map.class);`

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
esecules: Changed it so that proposedReferencingParameter is a function call and its last in the && chain so it will be short circuited if its non-service parameter. I organized it this way so that there is just one default else case where we assign `proposedValue` to `value`

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
esecules: Moved to a private static final

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizer.java
esecules: Looks like toReferenceList is the one I want anyway, It will filter out tokens that are not a parameter reference.

esecules:
@pvillard31 I responded to your comments and resolved a merge conflict with a similar test you added on the IT file. Looks like you resolved a bug in a similar area but it doesn't override this fix right?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/test/java/org/apache/nifi/flow/synchronization/StandardVersionedComponentSynchronizerTest.java
pvillard31: I believe the clean fix is to replace `ArgumentCaptor.forClass(Map.class)` with the type-safe `ArgumentCaptor.captor()` (already used in the class).

esecules: Done, thanks for the hint.

esecules:
Im seeing IT test failures that I think popped up after I rebased onto main which are failing on main itself while the new IT i added is passing. Would you expect `ExternalControllerServiceVersioningIT` to run in the GH actions? I don't think i am seeing it.

Mac OS with ARM Java 25 & Java 21:
image

However they all pass on my linux machine.

pvillard31 (approved):

PR #11450, opened 2026-07-21
NIFI-16130 - Fix ReplayLastEventEndpointMerger to count per-node replays correctly

pvillard31:
# Summary

NIFI-16130 - Fix ReplayLastEventEndpointMerger to count per-node replays correctly

Provenance event IDs are local counters per node, starting at 0 and incrementing independently. In a multi-node cluster two different nodes can legitimately have the same numeric event ID for entirely different physical provenance events.

`ReplayLastEventEndpointMerger` was using a `HashSet<Long>` to accumulate event IDs returned by each node after a "replay last event" request. Because both nodes in a 2-node cluster where each has processed the same number of events will return the same local event ID, the `HashSet` silently deduplicated them and reported `eventsReplayed.size() == 1` instead of `2`. This caused `ClusteredReplayProvenanceIT.testReplayLastEvent` (with `ReplayEventNodes.ALL`) to fail with `expected: <2> but was: <1>`.

**Change:** Replace `HashSet<Long>` with `ArrayList<Long>` for the event ID accumulator so each node's replay is counted independently regardless of matching local event IDs.

**Also adds:** `ReplayLastEventEndpointMergerTest` — there was no unit test for this merger class. The test covers the colliding-ID case (regression test for this fix), the distinct-ID case, and the partial-failure case.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/endpoints/ReplayLastEventEndpointMerger.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-cluster/src/main/java/org/apache/nifi/cluster/coordination/http/endpoints/ReplayLastEventEndpointMerger.java
@@ -28,6 +28,7 @@ import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


exceptionfactory (changes requested):
Thanks @pvillard31, the functional change to `List<Long>` for Provenance IDs looks good. For the new test, the public modifiers can be removed, and it looks like static values, as well as the `NodeIdentifier` objects themselves, could be declared once as static final members.

exceptionfactory (approved):
Thanks @pvillard31, the latest version looks good.

PR #11462, opened 2026-07-22
NIFI-16135 Sanitize URL-derived segments used to build nifi-api requests

mcgilman:
Add safeApiPath and isSameOriginTarget shared utilities and apply them to route/query-derived values in AccessPolicyService, DocumentationService, and the content viewer, so untrusted deep-link input cannot redirect authenticated requests or bypass the same-origin ref check.

# Summary

[NIFI-16135](https://issues.apache.org/jira/browse/NIFI-16135)

Hardens the UI's handling of values that originate from the browser URL (route
parameters and query parameters) before they are used to build authenticated
`nifi-api` requests. These values can be influenced by a crafted link opened by
an authenticated user but were previously trusted as-is when constructing the
request path.

This adds two small shared utilities in `libs/shared` and applies them at the
points where URL-derived values reach a request:

- `safeApiPath(...segments)` — validates each value as a single, atomic path
  segment (rejecting path separators, `..` traversal sequences including
  pre-encoded forms such as `%2e%2e`/`%2f`, control characters, empty and
  non-decodable values) and percent-encodes it. On an invalid segment it throws,
  so the caller fails closed and no request is issued. Legitimate values
  (identifiers and enumerated resource names) are unchanged, so existing URL
  shapes are preserved.
- `isSameOriginTarget(candidate, base, { requireBasePathPrefix })` —
  canonicalizes both values and compares origins with a segment-boundary-aware
  path check, replacing a bypassable `startsWith` prefix check.

Applied to:

- `DocumentationService` — component/step documentation requests built from
  route-supplied definition coordinates.
- `AccessPolicyService` — policy requests built from route-supplied
  action/resource/identifier values. Affected methods are wrapped in `defer(...)`
  so a rejection surfaces on the observable error channel that callers already
  handle.
- `ContentViewerComponent` — the query-supplied content `ref` guard now uses a
  same-origin comparison instead of a string prefix match.

Unit tests cover the new validation and same-origin logic, including traversal,
pre-encoded traversal, control-character, look-alike-origin, and path-boundary
cases.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/access-policies/service/access-policy.service.spec.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/access-policies/service/access-policy.service.spec.ts
@@ -174,5 +174,72 @@ describe('AccessPolicyService', () => {
             httpMock.expectOne('../nifi-api/policies/read/provenance-data/connectors').flush({});
             httpMock.verify();
         });
+
+        it('should encode an untrusted resourceIdentifier segment', () => {
+            const httpMock = TestBed.inject(HttpTestingController);
+
+            service
+                .getAccessPolicy({
+                    resource: 'processors',
+                    resourceIdentifier: 'a b',
+                    action: Action.Read
+                })
+                .subscribe();
+
+            httpMock.expectOne('../nifi-api/policies/read/processors/a%20b').flush({});
+            httpMock.verify();
+        });
+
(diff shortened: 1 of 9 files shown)


rfellows (approved):
Thanks for the security fixes here @mcgilman 👍

PR #11487, opened 2026-07-29
NIFI-16156: Add connector-scoped Controller Service GET so View Canva…

mcgilman:
…s Go To Service works without Troubleshooting.

- Adding response merging for the Controller Service endpoints.

### Problem

The "Go To Service" affordance on the property table needs to resolve a Controller Service by ID before it can navigate to it. For services managed by a Connector, the only existing lookup was `GET /controller-services/{id}`, which is gated so that it only succeeds while the Connector is in Troubleshooting mode. As a result, "Go To Service" was non-functional for Connector-managed services when viewing the canvas normally (outside Troubleshooting).

### Changes

**Backend (`nifi-web-api`)**
- Added `GET /connectors/{id}/controller-services/{controllerServiceId}` to `ConnectorResource`, authorized via `READ` on the Connector resource (not the underlying service), and resolved through `locateConnectorControllerService` so it bypasses the Troubleshooting gate — mirroring the pattern already established for connector-scoped Controller Service *state* (NIFI-15549).
- Supports the standard `uiOnly=true` query param, stripping non-UI-relevant fields the same way `ControllerServiceResource.getControllerService` does.
- Added `NiFiServiceFacade.getConnectorControllerService` / `StandardNiFiServiceFacade` implementation.

**Cluster response merging (`nifi-framework-cluster`)**
- Registered the new endpoint's URI pattern with the existing `ControllerServiceEndpointMerger`, so GET responses across cluster nodes are properly merged (validation status/errors, bulletins, referencing-component permissions) instead of an arbitrary node's response being returned unmerged.
- Also closed a related pre-existing gap: the connector-scoped Controller Services *list* endpoint (`GET /connectors/{id}/flow/process-groups/{processGroupId}/controller-services`) had no merger registered either; added its pattern to the existing `ControllerServicesEndpointMerger`.

**Toolkit client**
- Added `ConnectorClient.getControllerService` / `JerseyConnectorClient` implementation for the new endpoint.

**Frontend (`nifi-frontend`, upstream only)**
- `connector.service.ts`: added `getControllerService(connectorId, controllerServiceId)`.
- `connector-canvas.effects.ts` and `connector-controller-services.effects.ts`: wired `goToService` to call the new connector-scoped endpoint, then dispatch `navigateToControllerService` and close the dialog.
  - The fetch subscription tears down via `takeUntil(dialogRef.afterClosed())` rather than the service's lifetime, so it doesn't outlive the dialog.
  - Navigation is dispatched before the dialog closes, avoiding a race between the two.
  - If the Connector id can't be resolved from the route, a banner error is dispatched instead of silently doing nothing.

### Testing
- `TestConnectorResource`: new endpoint success, `uiOnly=true`, and not-authorized cases.
- `StandardNiFiServiceFacadeTest`: not-found case for the new facade method.
- New `ControllerServiceEndpointMergerTest` / `ControllerServicesEndpointMergerTest`: `canHandle` coverage for the connector-scoped patterns (including negative cases for malformed UUIDs, wrong HTTP method, and non-matching sub-paths).
- `connector-canvas.effects.spec.ts` / `connector-controller-services.effects.spec.ts`: success (fetch → navigate → close), HTTP error (banner error), and missing-connector-id (banner error, no HTTP call) cases for `goToService` in both dialogs.

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector.service.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector.service.ts
@@ -124,6 +124,14 @@ export class ConnectorService {
         );
     }
 
+    getControllerService(connectorId: string, controllerServiceId: string): Observable<any> {
+        const uiOnly: any = { uiOnly: true };
+        return this.httpClient.get(
+            `${ConnectorService.API}/connectors/${connectorId}/controller-services/${controllerServiceId}`,
+            { params: uiOnly }
+        );
+    }
+
     getConnectorParameterContext(
         connectorId: string,
         processGroupId: string
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector.service.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector.service.ts
@@ -23,7 +23,7 @@ import { Client } from '../../../service/client.service';
 import { ClusterConnectionService } from '../../../service/cluster-connection.service';
 import { ConnectorsResponse, CreateConnectorRequest } from '../state';
 import { ConnectorEntity } from '@nifi/shared';
-import { ParameterContextEntity, SearchResultsEntity } from '../../../state/shared';
+import { ControllerServiceEntity, ParameterContextEntity, SearchResultsEntity } from '../../../state/shared';
 import { DropRequestEntity } from '../../../state/empty-queue';
 
 @Injectable({ providedIn: 'root' })
@@ -124,11 +124,11 @@ export class ConnectorService {
         );
     }
 
-    getControllerService(connectorId: string, controllerServiceId: string): Observable<any> {
-        const uiOnly: any = { uiOnly: true };
-        return this.httpClient.get(
+    getControllerService(connectorId: string, controllerServiceId: string): Observable<ControllerServiceEntity> {
+        const params: Record<string, boolean> = { uiOnly: true };
+        return this.httpClient.get<ControllerServiceEntity>(
             `${ConnectorService.API}/connectors/${connectorId}/controller-services/${controllerServiceId}`,
-            { params: uiOnly }
+            { params }
         );
     }
 
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/state/connector-canvas/connector-canvas.effects.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/state/connector-canvas/connector-canvas.effects.ts
@@ -21,7 +21,6 @@ import { Actions, createEffect, ofType } from '@ngrx/effects';
 import { concatLatestFrom } from '@ngrx/operators';
 import { Store } from '@ngrx/store';
 import { Router } from '@angular/router';
-import { HttpErrorResponse } from '@angular/common/http';
 import { asyncScheduler, interval, NEVER, Observable, of } from 'rxjs';
 import {
     catchError,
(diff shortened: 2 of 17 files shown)


nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/service/connector.service.ts
```diff
@@ -124,6 +124,14 @@ export class ConnectorService {
         );
     }
 
+    getControllerService(connectorId: string, controllerServiceId: string): Observable<any> {
+        const uiOnly: any = { uiOnly: true };
+        return this.httpClient.get(
+            `${ConnectorService.API}/connectors/${connectorId}/controller-services/${controllerServiceId}`,
+            { params: uiOnly }
+        );
+    }
```
rfellows: Prefer typing this off `any`. `ControllerServiceEntity` already exists and is imported in the effect specs, so the response can be `Observable<ControllerServiceEntity>`, and the params can mirror `cancelConnectorDrain` above (`Record<string, ...>`) instead of `const uiOnly: any`:

```ts
getControllerService(connectorId: string, controllerServiceId: string): Observable<ControllerServiceEntity> {
    const params: Record<string, boolean> = { uiOnly: true };
    return this.httpClient.get<ControllerServiceEntity>(
        `${ConnectorService.API}/connectors/${connectorId}/controller-services/${controllerServiceId}`,
        { params }
    );
}
```

This satisfies the strict-typing / no-`any` guidance and, as a bonus, makes the effect's `serviceEntity` strongly typed so `component`'s optionality is visible at the call site. The sibling `getConnectorFlow`/`getConnectorControllerServices` also return `Observable<any>`, so if you'd rather keep parity within the file that's defensible — but tightening at least this new method is the better move.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/state/connector-canvas/connector-canvas.effects.ts
```diff
@@ -720,7 +721,49 @@ export class ConnectorCanvasEffects {
...
+                                        errorContext: {
+                                            errors: [this.errorHelper.getErrorString(errorResponse)],
+                                            context: ErrorContextKey.CONNECTOR_CANVAS
+                                        }
+                                    })
+                                );
+                            }
+                        });
+                });
+        };
```
rfellows: This ~40-line `goToService` handler is duplicated almost verbatim in `connector-controller-services.effects.ts` (lines 154-196); the two copies differ only in the `ErrorContextKey` (`CONNECTOR_CANVAS` vs `CONTROLLER_SERVICES`). Consider extracting a shared helper next to `bind-connector-parameter-context.ts`, e.g.:

```ts
bindGoToService(this.store, this.connectorService, this.errorHelper, dialogRef, ErrorContextKey.CONNECTOR_CANVAS);
```

That keeps the two canvases from drifting — as written, any fix to one branch has to be mirrored by hand in the other.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/state/connector-controller-services/connector-controller-services.effects.ts
```diff
@@ -148,7 +151,49 @@ export class ConnectorControllerServicesEffects {
...
+                                                    errorContext: {
+                                                        errors: [this.errorHelper.getErrorString(errorResponse)],
+                                                        context: ErrorContextKey.CONTROLLER_SERVICES
+                                                    }
+                                                })
+                                            );
+                                        }
+                                    });
+                            });
+                    };
```
rfellows: Duplicate of the `goToService` handler in `connector-canvas.effects.ts` (lines 724-766) — identical except for `ErrorContextKey.CONTROLLER_SERVICES`. See the suggestion there to extract a shared `bindGoToService(...)` helper so the two implementations can't drift.

rfellows (commented):
Two frontend follow-ups from review, both non-blocking (should-fix). Overall this is a clean, well-scoped, well-tested change that mirrors the existing controller-service GET nicely. Inline comments cover (1) tightening the new `getControllerService` off `any`, and (2) extracting the `goToService` handler that is currently duplicated across the two connector effects.

(Note: the earlier concern about `serviceEntity.component.parentGroupId` throwing when `component` is absent is effectively unreachable here — connector-managed components inherit the connector's authorizable via `managedRootGroup.setExplicitParentAuthorizable(connectorNode)`, so a caller that passes the endpoint's `Read - /connectors/{uuid}` gate always has read on the CS and `component` is populated. Left off as an optional hardening only.)

rfellows (approved):
Thanks for fixing this bug @mcgilman 👍 
I'll merge it once CI completes.

PR #11495, opened 2026-07-30
NIFI-16159 Add support for FileResourceService in PutFile

shubhluck:
# Summary

[NIFI-16159](https://issues.apache.org/jira/browse/NIFI-16159)

Add support for `FileResourceService` in `PutFile`, matching the `Resource Transfer Source` / `File Resource Service` pattern already available in `PutS3Object` (NIFI-12642), `PutGCSObject` (NIFI-12643), and `PutHDFS` (NIFI-12801).

`PutFile` currently writes only the incoming FlowFile's content to the destination directory. Users who want to stage a file already on the local filesystem (or reachable via a custom `FileResourceService` implementation) have to first pull it into a FlowFile with `FetchFile`/`GetFile`, which is wasteful for large local files.

**Changes**

- Add `Resource Transfer Source` and `File Resource Service` property descriptors to `PutFile`, reusing `ResourceTransferProperties` from `nifi-resource-transfer`.
- When `Resource Transfer Source = File Resource Service`, obtain a `FileResource` via `ResourceTransferUtils.getFileResource(...)` and stream its `InputStream` to the `.<filename>` temp file with `StandardCopyOption.REPLACE_EXISTING` (preserving the overwrite semantics of the original `session.exportTo(...)` call), then proceed with the existing rename / permissions / owner / group / lastModified logic.
- When `Resource Transfer Source = FlowFile Content` (default) the behavior is byte-identical to today.
- Add `nifi-resource-transfer` and `nifi-file-resource-service-api` (compile) and `nifi-file-resource-service` (test) to `nifi-standard-processors/pom.xml`.
- Add `TestPutFile#testPutFileFromLocalFile` covering the `FileResourceService` path using `StandardFileResourceService`.

**Backward compatibility**

`Resource Transfer Source` defaults to `FlowFile Content`, so existing flows are unaffected. `File Resource Service` only appears in the UI when the source is set to `File Resource Service` (declared via `dependsOn`).

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
@@ -39,6 +39,21 @@
             <version>2.11.0-SNAPSHOT</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-resource-transfer</artifactId>
+            <version>2.11.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-file-resource-service-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-file-resource-service</artifactId>
+            <version>2.11.0-SNAPSHOT</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-dbcp2</artifactId>
             <scope>test</scope>
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
@@ -39,6 +39,21 @@
             <version>2.12.0-SNAPSHOT</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-resource-transfer</artifactId>
+            <version>2.12.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-file-resource-service-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-file-resource-service</artifactId>
+            <version>2.12.0-SNAPSHOT</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-dbcp2</artifactId>
             <scope>test</scope>
--- a/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutFile.java
+++ b/nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutFile.java
@@ -27,6 +27,7 @@ import org.apache.nifi.components.PropertyDescriptor;
 import org.apache.nifi.components.ValidationResult;
 import org.apache.nifi.components.Validator;
 import org.apache.nifi.expression.ExpressionLanguageScope;
+import org.apache.nifi.fileresource.service.api.FileResource;
 import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.flowfile.attributes.CoreAttributes;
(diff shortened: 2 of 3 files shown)


shubhluck:
The `TestSocketLoadBalancedFlowFileQueue.testGetQueueSnapshotIncludesRebalancingPartition` failure on the JDK 21 EN leg is unrelated to this change (which only touches nifi-standard-processors/PutFile). The other three matrix legs passed. Looks like a flaky test.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutFile.java
```diff
@@ -304,7 +315,16 @@ public void onTrigger(final ProcessContext context, final ProcessSession session
                 }
             }
 
-            session.exportTo(flowFile, dotCopyFile, false);
+            final ResourceTransferSource resourceTransferSource = ResourceTransferSource.valueOf(
+                    context.getProperty(RESOURCE_TRANSFER_SOURCE).getValue());
+            final Optional<FileResource> fileResource = getFileResource(resourceTransferSource, context, flowFile.getAttributes());
```
exceptionfactory, line 319: This should use `asAllowableValue()` instead of `ResourceTransferSource.valueOf()`

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
```diff
@@ -38,6 +38,21 @@
             <artifactId>nifi-standard-record-utils</artifactId>
             <version>2.11.0-SNAPSHOT</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-resource-transfer</artifactId>
+            <version>2.11.0-SNAPSHOT</version>
```
exceptionfactory: The versions should be updated to `2.12.0-SNAPSHOT`.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestPutFile.java
```diff
@@ -244,6 +249,42 @@ public void testReplaceAndMaxFileLimitReach() throws IOException {
...
+        final FileResourceService service = new StandardFileResourceService();
+        runner.addControllerService(serviceId, service);
+        runner.setProperty(service, StandardFileResourceService.FILE_PATH, String.format("${%s}", attributeName));
+        runner.enableControllerService(service);
+
+        runner.setProperty(ResourceTransferProperties.RESOURCE_TRANSFER_SOURCE, ResourceTransferSource.FILE_RESOURCE_SERVICE.getValue());
+        runner.setProperty(ResourceTransferProperties.FILE_RESOURCE_SERVICE, serviceId);
+
+        final byte[] fileData = "0123456789".getBytes(StandardCharsets.UTF_8);
+        final Path tempFilePath = Files.createTempFile("PutFile_testPutFileFromLocalFile_", "");
```
exceptionfactory: This approach can leave files in the temporary directory. The JUnit TempDir annotation should be used to wire in a directory to the test method instead.

exceptionfactory (changes requested):
Thanks for proposing this addition @shubhluck. Although using the File Resource Service for local transfer is a bit of an edge case, it is minimal enough that it looks straightforward. I noted a few minor recommendations, then this should be ready to go.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutFile.java
shubhluck: Fixed as suggested switched to asAllowableValue

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/pom.xml
shubhluck: Bumped to 2.12.0-SNAPSHOT.

nifi-extension-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestPutFile.java
shubhluck: Switched to a @TempDir Path sourceDir method parameter and dropped the try/finally cleanup

exceptionfactory (approved):
Thanks @shubhluck, the latest version looks good! +1 merging

PR #11499, opened 2026-08-03
NIFI-14210 Add record writer for Protobuf messages

prankstrisse:
# Summary

[NIFI-14210](https://issues.apache.org/jira/browse/NIFI-14210)

  ## Changes                                                                                                                                                               

This PR adds a new Protobuf writer to NiFi, so it can now write Protobuf messages (before, it could only read them). As part of that work, it also fixes schema resolution in Schema Registry: schemas that reference other schemas now work correctly, whether they're in the same package or a different one.

                                                 
  - **`StandardProtobufWriter`** (`nifi-protobuf-services`) — new `SchemaRegistryService` + `RecordSetWriterFactory`. Resolves the target Proto message either statically via a `Message Name` property or dynamically via a `MessageNameResolver` service, compiles the schema with the existing `ProtobufSchemaCompiler`, and serializes each Record with a new `ProtobufDataSerializer`. A single message is written per FlowFile, since concatenated Protobuf messages cannot be delimited on read.
  - **`ProtobufDataSerializer`** — maps NiFi Record fields onto Wire `Schema`/`MessageType` descriptors and encodes the binary payload. `google.protobuf.Any`-typed fields are written as plain nested messages rather than re-wrapped as `Any`.
  - **`WriteProtobufResultWithExternalSchema`** — orchestrates header + message-index + payload sequencing, mirroring the existing `WriteAvroResultWithExternalSchema` pattern: invokes `SchemaReferenceWriter.writeHeader(...)` when a Schema Reference Writer is configured, then an optional `MessageIndexWriter.writeMessageIndex(...)` step, then the Protobuf payload.
  - **`MessageIndexWriter`** (`nifi-schema-registry-service-api`) — new shared interface, the write-side inverse of `MessageNameResolver`: given a resolved message name and schema definition, encodes the path to that message and writes it to the output stream so a reader can resolve the same name back from the encoded path.
  - **`ConfluentProtobufMessageIndexWriter`** + **`ProtobufMessageIndexEncoder`** (`nifi-confluent-protobuf-message-index-writer`, new module) — Confluent implementation of `MessageIndexWriter`. Walks nested message declarations to find the target message's declaration-order path and zigzag-varint encodes it (including the `[0]` single-message optimization), mirroring the decoding done by `ConfluentProtobufMessageNameResolver`. Together with the unchanged, already-generic `ConfluentEncodedSchemaReferenceWriter`, this reproduces the Confluent wire format (5-byte header + message-index array + payload) on write.
  - **`VarintUtils`** — relocated from the message-name-resolver module to `org.apache.nifi.confluent.schema` and made non-`package-private` so it can be shared between the resolver (read) and index writer (write) modules.                                                                                                                    
  - Minor visibility/refactoring changes to `ProtobufSchemaCompiler` and `ProtobufSchemaValidator` to support reuse from the writer.
                                                                                                                                                                           

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [x] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/main/java/org/apache/nifi/confluent/schemaregistry/ConfluentProtobufMessageIndexWriter.java
+++ b/nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/main/java/org/apache/nifi/confluent/schemaregistry/ConfluentProtobufMessageIndexWriter.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.confluent.schemaregistry;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.annotation.lifecycle.OnDisabled;
+import org.apache.nifi.annotation.lifecycle.OnEnabled;
+import org.apache.nifi.confluent.schema.AntlrProtobufMessageSchemaParser;
+import org.apache.nifi.confluent.schema.ProtobufMessageSchema;
+import org.apache.nifi.controller.AbstractControllerService;
+import org.apache.nifi.controller.ConfigurationContext;
+import org.apache.nifi.schemaregistry.services.MessageIndexWriter;
+import org.apache.nifi.schemaregistry.services.MessageName;
+import org.apache.nifi.schemaregistry.services.SchemaDefinition;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+
+@Tags({"confluent", "schema", "registry", "protobuf", "message", "index", "writer"})
+@CapabilityDescription("""
+    Writes Protobuf message index information in Confluent Schema Registry wire format by encoding the declaration-order path to a given message name within the schema definition.
+    For Confluent wire format reference see: https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#wire-format
+    """)
+public class ConfluentProtobufMessageIndexWriter extends AbstractControllerService implements MessageIndexWriter {
+
+    private static final int MAXIMUM_CACHE_SIZE = 1000;
+    private static final int CACHE_EXPIRE_HOURS = 1;
+
+    private Cache<EncodeMessageIndexArguments, byte[]> messageIndexCache;
+
+    @OnEnabled
+    public void onEnabled(final ConfigurationContext context) {
+        messageIndexCache = Caffeine.newBuilder().maximumSize(MAXIMUM_CACHE_SIZE).expireAfterWrite(Duration.ofHours(CACHE_EXPIRE_HOURS)).build();
+    }
+
+    @OnDisabled
+    public void onDisabled(final ConfigurationContext context) {
+        if (messageIndexCache != null) {
+            messageIndexCache.invalidateAll();
+            messageIndexCache = null;
+        }
+    }
+
+    @Override
+    public void writeMessageIndex(final Map<String, String> variables, final SchemaDefinition schemaDefinition, final MessageName messageName, final OutputStream outputStream) throws IOException {
+        final EncodeMessageIndexArguments encodeMessageIndexArguments = new EncodeMessageIndexArguments(schemaDefinition, messageName);
+        final byte[] encodedMessageIndex = messageIndexCache.get(encodeMessageIndexArguments, this::encodeMessageIndex);
+        outputStream.write(encodedMessageIndex);
+    }
+
+    private byte[] encodeMessageIndex(final EncodeMessageIndexArguments encodeMessageIndexArguments) {
+        try {
+            final String schemaText = encodeMessageIndexArguments.schemaDefinition().getText();
+            final AntlrProtobufMessageSchemaParser parser = new AntlrProtobufMessageSchemaParser();
+            final List<ProtobufMessageSchema> rootMessages = parser.parse(schemaText);
+            return ProtobufMessageIndexEncoder.encode(rootMessages, encodeMessageIndexArguments.messageName());
+        } catch (final Exception e) {
+            throw new IllegalStateException("Failed to parse protobuf schema", e);
+        }
+    }
+
+    private record EncodeMessageIndexArguments(SchemaDefinition schemaDefinition, MessageName messageName) {
+    }
+}
--- a/nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/main/java/org/apache/nifi/confluent/schemaregistry/ProtobufMessageIndexEncoder.java
+++ b/nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/main/java/org/apache/nifi/confluent/schemaregistry/ProtobufMessageIndexEncoder.java
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.confluent.schemaregistry;
+
+import org.apache.nifi.confluent.schema.ProtobufMessageSchema;
+import org.apache.nifi.confluent.schema.VarintUtils;
+import org.apache.nifi.schemaregistry.services.MessageName;
+
+import java.io.ByteArrayOutputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.lang.String.format;
+
+/**
+ * Computes the Confluent wire format message index path for a target message within a parsed
+ * Protobuf schema, and encodes it to bytes. This is the inverse of the message index decoding
+ * performed by {@code ConfluentProtobufMessageNameResolver}: given a fully qualified message
+ * name, it locates the path of declaration-order indexes leading to that message and encodes it
+ * as zigzag varints, applying the single-byte {@code 0x00} optimization for the common case of
+ * the first root message.
+ * <p>
+ * <a href="https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#wire-format">See the Confluent protobuf wire format.</a>
+ * <p>
+ * This class has no dependency on the NiFi framework and can be exercised directly with plain
+ * unit tests.
+ */
+final class ProtobufMessageIndexEncoder {
+
+    private static final byte[] FIRST_ROOT_MESSAGE_INDEX = {0x00};
+
+    private ProtobufMessageIndexEncoder() {
+    }
+
+    /**
+     * Encodes the message index path for the given message name within the given schema.
+     *
+     * @param rootMessages the root messages of the parsed Protobuf schema, in declaration order
+     * @param messageName  the target message name to locate
+     * @return the encoded message index bytes
+     * @throws IllegalStateException if the message name cannot be located within the schema
+     */
+    static byte[] encode(final List<ProtobufMessageSchema> rootMessages, final MessageName messageName) {
+        final List<Integer> messageIndexPath = findMessageIndexPath(rootMessages, messageName);
+
+        if (messageIndexPath.size() == 1 && messageIndexPath.getFirst() == 0) {
+            return FIRST_ROOT_MESSAGE_INDEX;
+        }
+
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        output.writeBytes(VarintUtils.writeZigZagVarint(messageIndexPath.size()));
+        for (final int index : messageIndexPath) {
+            output.writeBytes(VarintUtils.writeZigZagVarint(index));
+        }
+        return output.toByteArray();
+    }
+
+    private static List<Integer> findMessageIndexPath(final List<ProtobufMessageSchema> rootMessages, final MessageName messageName) {
+        // Match on the fully qualified name across the message tree rather than on the namespace/name split, so the
+        // encoder is robust to how the MessageName was constructed (a statically configured "package.Outer.Nested"
+        // and a resolver-produced name both reduce to the same fully qualified name).
+        final String targetFullyQualifiedName = messageName.getFullyQualifiedName();
+
+        for (int rootIndex = 0; rootIndex < rootMessages.size(); rootIndex++) {
+            final ProtobufMessageSchema rootMessage = rootMessages.get(rootIndex);
+            final String rootFullyQualifiedName = rootMessage.getPackageName()
+                .map(packageName -> packageName + "." + rootMessage.getName())
+                .orElseGet(rootMessage::getName);
+
+            final List<Integer> messageIndexPath = descend(rootMessage, rootFullyQualifiedName, targetFullyQualifiedName);
+            if (messageIndexPath != null) {
+                messageIndexPath.addFirst(rootIndex);
+                return messageIndexPath;
+            }
+        }
+
+        throw new IllegalStateException(format("Message not found in schema definition: %s", targetFullyQualifiedName));
+    }
+
+    private static List<Integer> descend(final ProtobufMessageSchema currentMessage, final String currentFullyQualifiedName, final String targetFullyQualifiedName) {
+        if (currentFullyQualifiedName.equals(targetFullyQualifiedName)) {
+            return new ArrayList<>();
+        }
+
+        final List<ProtobufMessageSchema> children = currentMessage.getChildMessageSchemas();
+        for (int childIndex = 0; childIndex < children.size(); childIndex++) {
+            final ProtobufMessageSchema child = children.get(childIndex);
+            final String childFullyQualifiedName = currentFullyQualifiedName + "." + child.getName();
+
+            final List<Integer> messageIndexPath = descend(child, childFullyQualifiedName, targetFullyQualifiedName);
+            if (messageIndexPath != null) {
+                messageIndexPath.addFirst(childIndex);
+                return messageIndexPath;
+            }
+        }
+
+        return null;
+    }
+}
--- a/nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/test/java/org/apache/nifi/confluent/schemaregistry/ProtobufMessageIndexEncoderTest.java
+++ b/nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/test/java/org/apache/nifi/confluent/schemaregistry/ProtobufMessageIndexEncoderTest.java
@@ -1,0 +1,195 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.confluent.schemaregistry;
+
+import org.apache.nifi.confluent.schema.AntlrProtobufMessageSchemaParser;
+import org.apache.nifi.confluent.schema.ProtobufMessageSchema;
+import org.apache.nifi.confluent.schema.VarintUtils;
+import org.apache.nifi.schemaregistry.services.MessageName;
+import org.apache.nifi.schemaregistry.services.StandardMessageName;
+import org.apache.nifi.schemaregistry.services.StandardMessageNameFactory;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * This class contains no NiFi framework dependencies: it exercises {@link ProtobufMessageIndexEncoder}
+ * directly, without a TestRunner or any controller-service context.
+ */
+class ProtobufMessageIndexEncoderTest {
+
+    // Schema without package (default package)
+    private static final String DEFAULT_PACKAGE_SCHEMA = """
+        syntax = "proto3";
+        message User {
+          int32 id = 1;
+          string name = 2;
+          Address address = 3;
+          message Profile {
+            string bio = 1;
+            Settings settings = 2;
+            message Settings {
+              bool notifications = 1;
+              string theme = 2;
+            }
+          }
+        }
+        message Company {
+          string name = 1;
+          Address address = 2;
+        }
+        message Address {
+          string street = 1;
+          string city = 2;
+        }""";
+
+    // Schema with explicit package
+    private static final String EXPLICIT_PACKAGE_SCHEMA = """
+        syntax = "proto3";
+        package com.example.proto;
+        message User {
+          int32 id = 1;
+          string name = 2;
+          Address address = 3;
+          message Profile {
+            string bio = 1;
+            Settings settings = 2;
+            message Settings {
+              bool notifications = 1;
+              string theme = 2;
+            }
+          }
+        }
+        message Company {
+          string name = 1;
+          Address address = 2;
+        }
+        message Address {
+          string street = 1;
+          string city = 2;
+        }""";
+
+    private static Stream<Arguments> provideMessageIndexTestCases() {
+        return Stream.of(
+            // the format is [input schema], [target message name], [expected message indexes]
+            Arguments.of(DEFAULT_PACKAGE_SCHEMA, new StandardMessageName(Optional.empty(), "User"), new int[] {0}),
+            Arguments.of(DEFAULT_PACKAGE_SCHEMA, new StandardMessageName(Optional.empty(), "Company"), new int[] {1}),
+            Arguments.of(DEFAULT_PACKAGE_SCHEMA, new StandardMessageName(Optional.empty(), "Address"), new int[] {2}),
+            Arguments.of(DEFAULT_PACKAGE_SCHEMA, new StandardMessageName(Optional.empty(), "User.Profile"), new int[] {0, 0}),
+            Arguments.of(DEFAULT_PACKAGE_SCHEMA, new StandardMessageName(Optional.empty(), "User.Profile.Settings"), new int[] {0, 0, 0}),
+
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "User"), new int[] {0}),
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "Company"), new int[] {1}),
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "Address"), new int[] {2}),
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "User.Profile"), new int[] {0, 0}),
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "User.Profile.Settings"), new int[] {0, 0, 0})
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideMessageIndexTestCases")
+    void testEncode(final String schemaText, final MessageName messageName, final int[] expectedIndexes) throws IOException {
+        final List<ProtobufMessageSchema> rootMessages = new AntlrProtobufMessageSchemaParser().parse(schemaText);
+
+        final byte[] encoded = ProtobufMessageIndexEncoder.encode(rootMessages, messageName);
+
+        assertEquals(IntStream.of(expectedIndexes).boxed().toList(), decodeIndexes(encoded));
+    }
+
+    @Test
+    void testEncodeFirstRootMessageUsesSingleByteOptimization() {
+        final List<ProtobufMessageSchema> rootMessages = new AntlrProtobufMessageSchemaParser().parse(DEFAULT_PACKAGE_SCHEMA);
+
+        final byte[] encoded = ProtobufMessageIndexEncoder.encode(rootMessages, new StandardMessageName(Optional.empty(), "User"));
+
+        assertArrayEquals(new byte[] {0x00}, encoded);
+    }
+
+    @Test
+    void testEncodeMessageNotFoundThrows() {
+        final List<ProtobufMessageSchema> rootMessages = new AntlrProtobufMessageSchemaParser().parse(DEFAULT_PACKAGE_SCHEMA);
+
+        assertThrows(IllegalStateException.class,
+            () -> ProtobufMessageIndexEncoder.encode(rootMessages, new StandardMessageName(Optional.empty(), "DoesNotExist")));
+    }
+
+    @Test
+    void testEncodeNamespaceMismatchThrows() {
+        final List<ProtobufMessageSchema> rootMessages = new AntlrProtobufMessageSchemaParser().parse(EXPLICIT_PACKAGE_SCHEMA);
+
+        assertThrows(IllegalStateException.class,
+            () -> ProtobufMessageIndexEncoder.encode(rootMessages, new StandardMessageName(Optional.empty(), "User")));
+    }
+
+    @Test
+    void testEncodeNestedMessageFromFactorySplitName() throws IOException {
+        // StandardMessageNameFactory.fromName is what the writer's static Message Name property uses. For a nested
+        // message it splits "User.Profile" into namespace="User"/name="Profile" - the encoder must still resolve it
+        // to the nested index [0, 0] by matching on the fully qualified name.
+        final List<ProtobufMessageSchema> rootMessages = new AntlrProtobufMessageSchemaParser().parse(DEFAULT_PACKAGE_SCHEMA);
+
+        final byte[] encoded = ProtobufMessageIndexEncoder.encode(rootMessages, StandardMessageNameFactory.fromName("User.Profile"));
+
+        assertEquals(List.of(0, 0), decodeIndexes(encoded));
+    }
+
+    @Test
+    void testEncodeNestedMessageWithPackageFromFactorySplitName() throws IOException {
+        // With a package, fromName splits "com.example.proto.User.Profile" into namespace="com.example.proto.User"/
+        // name="Profile"; the fully qualified name still resolves to the nested index [0, 0].
+        final List<ProtobufMessageSchema> rootMessages = new AntlrProtobufMessageSchemaParser().parse(EXPLICIT_PACKAGE_SCHEMA);
+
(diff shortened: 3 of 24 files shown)


nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/ProtobufSchemaValidator.java
```diff
@@ -17,47 +17,45 @@
...
+     * Validates that every schema reference, at any depth, is keyed by a path ending in the .proto extension.
+     * <p>
+     * A reference is keyed by the path used in the import statement of the referencing schema, for example
+     * {@code airlines/ph/cdm/shared.proto}, and the referenced schema is written to exactly that path so the import
+     * resolves. The extension is required because the compiler only discovers files named {@code *.proto}; without it
+     * the schema would fail to compile later with an unresolved import that does not indicate the cause.
+     * <p>
+     * The identifier of a referenced schema is deliberately not validated. It carries the subject the schema is
+     * registered under, which is unrelated to the import path and legitimately has no .proto suffix. Under the
+     * Confluent RecordNameStrategy, for instance, a subject is a fully qualified record name.
```
tpalfy: I would avoid referencing Confluent-specific details.

```suggestion
* The identifier of a referenced schema is deliberately not validated. It identifies the schema
* in its source registry and is unrelated to the Protobuf import path, so it is not required to
* have a .proto suffix.
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/ProtobufSchemaCompiler.java
```diff
@@ -182,23 +183,40 @@ private void safeDeleteDirectory(final Path directory) {
...
      */
     private void writeSchemaToTempDirectory(final Path tempDir, final SchemaDefinition schemaDefinition) throws IOException {
         logger.debug("Writing schema definition to temporary directory. Identifier: {}", schemaDefinition.getIdentifier());
+        writeSchemaFile(tempDir, generateSchemaFileName(schemaDefinition), schemaDefinition.getText());
+    }
+
+    /**
+     * Writes schema text to a path relative to the temporary directory, creating any parent directories.
+     * Import paths may contain directories, such as {@code airlines/ph/cdm/shared.proto}, so the enclosing
+     * directory structure has to exist before the file is written.
```
tpalfy:
```suggestion
* Import paths may contain directories, such as {@code airlines/ph/cdm/shared.proto}, so this method
* creates the enclosing directory structure before writing the file.
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+        final ProtobufWriteContext context = createWriteContext(variables);
+        return new WriteProtobufResultWithExternalSchema(context.schema(), context.messageName(), context.recordSchema(),
+            context.schemaDefinition(), schemaReferenceWriter, messageIndexWriter, variables, out);
+    }
+
+    @Override
+    protected List<PropertyDescriptor> getSupportedPropertyDescriptors() {
+        final List<PropertyDescriptor> properties = new ArrayList<>(super.getSupportedPropertyDescriptors());
+        // The Schema Reference Reader is a read-side concern: a writer determines its schema from the configured
+        // access strategy and writes references through the Schema Reference Writer instead.
```
tpalfy: Unnecessary (and a bit confusing) comment.

```suggestion
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+        properties.add(MESSAGE_NAME);
+        properties.add(SCHEMA_REFERENCE_WRITER);
+        properties.add(MESSAGE_INDEX_WRITER);
+        return properties;
+    }
+
+    @Override
+    protected List<AllowableValue> getSchemaAccessStrategyValues() {
+        // Only the strategies that createSchemaDefinition supports are offered; the inherited list also contains
+        // the Schema Reference Reader strategy, which cannot be used to obtain a schema for writing.
```
tpalfy: Unnecessary comments.

```suggestion
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+
+    private ProtobufWriteContext createWriteContext(final Map<String, String> variables) throws SchemaNotFoundException, IOException {
+        final SchemaDefinition schemaDefinition = createSchemaDefinition(variables);
+        final Schema schema = schemaCompiler.compileOrGetFromCache(schemaDefinition);
+        final MessageName messageName = messageNameResolver.getMessageName(variables, schemaDefinition, EMPTY_INPUT_STREAM);
+
+        final ProtoSchemaParser schemaParser = new ProtoSchemaParser(schema);
+        final RecordSchema parsedSchema = schemaParser.createSchema(messageName.getFullyQualifiedName());
+        // Preserve the schema identifier from the SchemaDefinition so the configured Schema Reference Writer can
+        // write the correct schema id in the Confluent header.
```
tpalfy: Useful comment, but better avoid referencing Confluent-specific details.

```suggestion
// Preserve the schema identifier required by the configured Schema Reference Writer.
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+            + "(such as the Confluent wire-format resolver used on the read side) are not applicable here.")
+        .required(true)
+        .identifiesControllerService(MessageNameResolver.class)
+        .dependsOn(MESSAGE_NAME_RESOLUTION_STRATEGY, MessageNameResolverStrategy.MESSAGE_NAME_RESOLVER)
+        .build();
+
+    public static final PropertyDescriptor SCHEMA_REFERENCE_WRITER = new PropertyDescriptor.Builder()
+        .name("Schema Reference Writer")
+        .description("Service used to write schema reference information, such as a Confluent wire-format header, before the serialized Protobuf content. "
+            + "When not configured, plain Protobuf content is written without any header.")
```
tpalfy: Better avoid Confluent specific details.

```suggestion
.description("The Controller Service used to write schema reference information before the serialized Protobuf message. "
    + "When not configured, no schema reference header is written.")
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+        .required(false)
+        .identifiesControllerService(SchemaReferenceWriter.class)
+        .build();
+
+    public static final PropertyDescriptor MESSAGE_INDEX_WRITER = new PropertyDescriptor.Builder()
+        .name("Message Index Writer")
+        .description("Service used to write the Confluent message index array identifying the target message within the schema, written after the Schema Reference Writer header. "
+            + "Applicable only when producing Confluent wire-format content.")
```
tpalfy: Better avoid Confluent specific details.
```suggestion
.description("The Controller Service used to write information identifying the selected message within the Protobuf schema. "
    + "The information is written after any schema reference header and before the Protobuf payload. ")
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_NAME_PROPERTY;
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_REFERENCE_READER;
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_REGISTRY;
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_TEXT;
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_TEXT_PROPERTY;
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_VERSION;
+import static org.apache.nifi.services.protobuf.StandardProtobufWriter.MessageNameResolverStrategy.MESSAGE_NAME_PROPERTY;
+
+@Tags({"protobuf", "record", "writer", "serializer", "confluent"})
+@CapabilityDescription("""
```
tpalfy: I have multiple suggestion to remove Confluent-specific details.
If we really wan't to give some hint for Confluent, we could add it under the CapabilityDescription.

Something like this:

"Schema Reference Writer and Message Index Writer services can optionally add format-specific framing before the Protobuf payload. Confluent wire format requires both a compatible schema reference header and message index."

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+
+    @Override
+    protected PropertyDescriptor buildSchemaTextProperty() {
+        return PROTOBUF_SCHEMA_TEXT;
+    }
+
+    private ProtobufWriteContext createWriteContext(final Map<String, String> variables) throws SchemaNotFoundException, IOException {
+        final SchemaDefinition schemaDefinition = createSchemaDefinition(variables);
+        final Schema schema = schemaCompiler.compileOrGetFromCache(schemaDefinition);
+        final MessageName messageName = messageNameResolver.getMessageName(variables, schemaDefinition, EMPTY_INPUT_STREAM);
```
tpalfy: While technically correct because the 0-length byte array behaves like an immutable object and can be reused safely, it is surprising and gives the code-reader a pause. That being said, I'm okay keeping it like this - but let me offer an alternative:

```suggestion
        final MessageName messageName = messageNameResolver.getMessageName(variables, schemaDefinition, InputStream.nullInputStream());
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+@CapabilityDescription("""
+    Serializes NiFi Records into Protocol Buffers binary format. \
+    Supports inline schema text and schema registry lookup for determining the Proto schema. \
+    When a Schema Reference Writer is configured, a Confluent wire-format header is written; when a \
+    Message Index Writer is also configured, the Confluent message index array is written after the header. \
+    The target Proto message name can be determined statically using the 'Message Name' property, \
+    or dynamically using a Message Name Resolver service.
+    A single record is written per FlowFile, since concatenated Protocol Buffers messages cannot be delimited. \
+    The 'google.protobuf.Any' well-known type is not expanded on write; a Record derived from an Any-typed message \
+    is serialized as an ordinary nested message rather than being re-wrapped as an Any.""")
```
tpalfy: Avoid referencing Confluent as part of the generic capability description. A reminded-type mentioning of the need of both services for Confluent is okay here though.

```suggestion
@CapabilityDescription("""
    Serializes NiFi Records into Protocol Buffers binary format. \
    Supports inline schema text and schema registry lookup for determining the Proto schema. \
    Optional Schema Reference Writer and Message Index Writer Controller Services can add framing before the \
    Protobuf payload. When both are configured, schema reference information is written first, followed by message \
    index information. Selected implementations must be compatible with each other and with the target wire format. \
    Confluent Protobuf wire format requires both compatible services. \
    The target Proto message name can be determined statically using the 'Message Name' property or dynamically \
    using a Message Name Resolver service. \
    A single record is written per FlowFile because concatenated Protocol Buffers messages cannot be delimited. \
    The 'google.protobuf.Any' well-known type is not expanded on write; a Record derived from an Any-typed message \
    is serialized as an ordinary nested message rather than being re-wrapped as an Any.""")
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+        schemaAccessStrategyValue = context.getProperty(SCHEMA_ACCESS_STRATEGY).getValue();
+        schemaRegistry = context.getProperty(SCHEMA_REGISTRY).asControllerService(SchemaRegistry.class);
+        schemaReferenceWriter = context.getProperty(SCHEMA_REFERENCE_WRITER).asControllerService(SchemaReferenceWriter.class);
+        messageIndexWriter = context.getProperty(MESSAGE_INDEX_WRITER).asControllerService(MessageIndexWriter.class);
+        schemaName = context.getProperty(SCHEMA_NAME);
+        schemaText = context.getProperty(SCHEMA_TEXT);
+        schemaBranchName = context.getProperty(SCHEMA_BRANCH_NAME);
+        schemaVersion = context.getProperty(SCHEMA_VERSION);
+    }
+
```
tpalfy: An inline schema via Schema Text doesn't have a Confluent-compatible identifier.
Not sure if this is intended, but if it is, we can add a custom validate to make sure that when the ConfluentEncodedSchemaReferenceWriter is set, the Schema Name strategy is selected as well.

```suggestion

    @Override
    protected Collection<ValidationResult> customValidate(final ValidationContext validationContext) {
        final List<ValidationResult> results = new ArrayList<>(super.customValidate(validationContext));

        if (!validationContext.getProperty(SCHEMA_REFERENCE_WRITER).isSet()) {
            return results;
        }

        final SchemaReferenceWriter referenceWriter = validationContext.getProperty(SCHEMA_REFERENCE_WRITER)
                .asControllerService(SchemaReferenceWriter.class);

        if (referenceWriter == null) {
            return results;
        }

        final Set<SchemaField> suppliedFields = getSuppliedSchemaFields(validationContext);
        final Set<SchemaField> missingFields = new HashSet<>(referenceWriter.getRequiredSchemaFields());
        missingFields.removeAll(suppliedFields);

        if (!missingFields.isEmpty()) {
            results.add(new ValidationResult.Builder()
                    .subject(SCHEMA_REFERENCE_WRITER.getDisplayName())
                    .valid(false)
                    .explanation("The configured Schema Reference Writer requires schema fields that are not provided "
                            + "by the configured Schema Access Strategy and Schema Registry: " + missingFields)
                    .build());
        }

        return results;
    }
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+    @Override
+    public RecordSchema getSchema(final Map<String, String> variables, final RecordSchema readSchema) throws SchemaNotFoundException, IOException {
+        return createWriteContext(variables).recordSchema();
+    }
+
+    @Override
+    public RecordSetWriter createWriter(final ComponentLog logger, final RecordSchema schema, final OutputStream out, final Map<String, String> variables) throws SchemaNotFoundException, IOException {
+        final ProtobufWriteContext context = createWriteContext(variables);
```
tpalfy:
```suggestion
        final ProtobufWriteContext context = createWriteContext(variables);

        if (schemaReferenceWriter != null) {
            schemaReferenceWriter.validateSchema(context.recordSchema());
        }
```

prankstrisse:
@tpalfy thanks for the interest in this PR. There were some merge conflicts after merging https://github.com/apache/nifi/pull/11710 . So the commit I pushed today is the resolution of them

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/WriteProtobufResultWithExternalSchema.java
```diff
@@ -0,0 +1,137 @@
...
+        this.serializer = new ProtobufDataSerializer(schema, messageName.getFullyQualifiedName());
+    }
+
+    @Override
+    protected void onBeginRecordSet() throws IOException {
+        writeConfluentFraming(buffered);
+    }
+
+    @Override
+    protected Map<String, String> onFinishRecordSet() throws IOException {
```
pvillard31: Should this return schemaReferenceWriter.getAttributes(recordSchema) when a Schema Reference Writer is configured, so attribute-based schema references are not discarded?

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/converter/ProtobufDataSerializer.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+        for (final Field field : messageType.getExtensionFields()) {
+            writeField(codedOutput, field, record.getValue(field.getName()));
+        }
+        for (final OneOf oneOf : messageType.getOneOfs()) {
+            for (final Field field : oneOf.getFields()) {
+                writeField(codedOutput, field, record.getValue(field.getName()));
+            }
+        }
+
```
pvillard31: Should a missing proto2 required field fail serialization, or should the writer reject proto2 schemas if only proto3 output is supported?

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+        schemaCompiler = new ProtobufSchemaCompiler(getIdentifier(), getLogger());
+    }
+
+    @Override
+    public RecordSchema getSchema(final Map<String, String> variables, final RecordSchema readSchema) throws SchemaNotFoundException, IOException {
+        return createWriteContext(variables).recordSchema();
+    }
+
+    @Override
+    public RecordSetWriter createWriter(final ComponentLog logger, final RecordSchema schema, final OutputStream out, final Map<String, String> variables) throws SchemaNotFoundException, IOException {
```
pvillard31: Should createWriter() use the supplied schema instead of resolving the registry schema again, so a moving latest version cannot change between getSchema() and writer creation?

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/converter/ProtobufDataSerializer.java
```diff
@@ -0,0 +1,300 @@
...
+        final ProtoType keyType = protoType.getKeyType();
+        final ProtoType valueType = protoType.getValueType();
+
+        for (final Map.Entry<?, ?> entry : map.entrySet()) {
+            final ByteArrayOutputStream entryBytes = new ByteArrayOutputStream();
+            final CodedOutputStream entryOutput = CodedOutputStream.newInstance(entryBytes);
+
+            writeSingleValue(entryOutput, MAP_KEY_TAG, keyType, entry.getKey());
```
pvillard31: Should null map keys be rejected with a clear IOException before scalar conversion?

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
```diff
@@ -0,0 +1,356 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.HexFormat;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.nifi.expression.ExpressionLanguageScope.FLOWFILE_ATTRIBUTES;
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_ACCESS_STRATEGY;
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_BRANCH_NAME;
+import static org.apache.nifi.schema.access.SchemaAccessUtils.SCHEMA_NAME;
```
pvillard31: Should @SeeAlso(StandardProtobufReader.class) be added so users can discover the companion reader?

nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/main/java/org/apache/nifi/confluent/schemaregistry/ProtobufMessageIndexEncoder.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param rootMessages the root messages of the parsed Protobuf schema, in declaration order
+     * @param messageName  the target message name to locate
+     * @return the encoded message index bytes
+     * @throws IllegalStateException if the message name cannot be located within the schema
+     */
+    static byte[] encode(final List<ProtobufMessageSchema> rootMessages, final MessageName messageName) {
+        final List<Integer> messageIndexPath = findMessageIndexPath(rootMessages, messageName);
+
+        if (messageIndexPath.size() == 1 && messageIndexPath.getFirst() == 0) {
+            return FIRST_ROOT_MESSAGE_INDEX;
```
pvillard31: Should this return a copy of FIRST_ROOT_MESSAGE_INDEX so callers cannot mutate shared cached data?

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/WriteProtobufResultWithExternalSchema.java
prankstrisse: yes, fixed. It now returns the attributes both in onFinishRecordSet() and when a record is written without an active record set

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/converter/ProtobufDataSerializer.java
prankstrisse: I've seen in the tests for the read part - it handles proto2, so I kept proto2 support in the writer for consistency and made a missing required field fail with an IOException.

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: Yes, totally agree. 
I can't use the supplied RecordSchema on its own, because writing also needs the compiled Protobuf schema and the original schema text (for the message index and imports). They cannot be rebuilt from a RecordSchema. So createWriter() now fetches the registry schema using the name and version of the supplied schema. If the supplied schema has no name and version, it falls back to the previous lookup.

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/converter/ProtobufDataSerializer.java
prankstrisse: Yes, fixed. I also changed unknown enum values to throw IOException instead of IllegalStateException, so that the errors are reported in the same way in all places

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: Yes, done

nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/main/java/org/apache/nifi/confluent/schemaregistry/ProtobufMessageIndexEncoder.java
prankstrisse: Yes, done. Side note: ConfluentProtobufMessageIndexWriter also caches the encoded arrays and returns the same array each time. That's safe because callers only write those bytes to the output stream and never change them, so I left the cache as it is

prankstrisse:
Thank you very much for the review @pvillard31 , I pushed a commit with the fixes

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/ProtobufSchemaValidator.java
prankstrisse: thanks, done

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/ProtobufSchemaCompiler.java
prankstrisse: thanks, done

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: thanks, done

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: thanks, done

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: thanks, done

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: thanks, done

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: thanks, done

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: yes, totally agree, changed it

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: done, thanks

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: Yes, that's intended: inline Schema Text has no registry identifier, so a Confluent header can't be written. Added your validation, with one small change: the catch is RuntimeException instead of only SchemaCompilationException, because the compiler can also throw other runtime exceptions and those should become a validation error too. Also added tests for an invalid schema text, an unknown message name, and Schema Text combined with the Confluent Schema Reference Writer

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/StandardProtobufWriter.java
prankstrisse: thanks, added

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/WriteProtobufResultWithExternalSchema.java
```diff
@@ -0,0 +1,141 @@
...
+        if (closed) {
+            return;
+        }
+        closed = true;
+
+        // Ensure buffered content is flushed to the underlying stream before it is closed, including the
+        // write-without-active-record-set path where onFinishRecordSet is never invoked.
+        flush();
+        super.close();
+    }
```
tpalfy: If `flush()` throws an exception, the `super.close()` won't get called. We can simply call close on the `BufferedOutputStream`. As that call is idempotent, we can call it as many times as we want. The `closed` flag becomes unnecessary.

```suggestion
@Override
public void close() throws IOException {
    buffered.close();
}
```

prankstrisse:
Thank you so much for the review, @tpalfy ! I pushed the fixes

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/WriteProtobufResultWithExternalSchema.java
```diff
@@ -0,0 +1,141 @@
...
+        // If we are not writing an active record set, then we need to ensure that the Confluent framing is written.
+        if (!isActiveRecordSet()) {
+            flush();
+            writeConfluentFraming(buffered);
+        }
+
+        final byte[] payload = serializer.serialize(record);
+        buffered.write(payload);
+        return getSchemaReferenceAttributes();
+    }
```
tpalfy: The framing handling is confusing, and can result in corrupted stream in edge cases (empty recordset, serialization failure).

The framing is tied to the single record, so we can simply remove `onBeginRecordSet()` and update this method like this:


```suggestion
@Override
public Map<String, String> writeRecord(final Record record) throws IOException {
    if (getRecordCount() > 0) {
        throw new IOException("Protobuf output supports only a single record");
    }
    final byte[] payload = serializer.serialize(record);
    writeFraming(buffered);
    buffered.write(payload);
    return getSchemaReferenceAttributes();
}
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/WriteProtobufResultWithExternalSchema.java
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Writes Records as Protocol Buffers binary content. When a {@link SchemaReferenceWriter} is
+ * configured, a Confluent wire-format header (magic byte and schema identifier) is written first;
+ * when a {@link MessageIndexWriter} is configured, the Confluent message index array follows the
+ * header. The serialized Protobuf payload is written last.
+ * <p>
+ * The Confluent framing (header and message index) is written once at the beginning of the record
+ * set, mirroring {@code WriteAvroResultWithExternalSchema}; the typical Confluent use case writes a
+ * single message per FlowFile.
+ */
```
tpalfy: Avoid Confluent-specific details.

```suggestion
/**
 * Writes a single Record as Protocol Buffers binary content. When configured, the
 * Schema Reference Writer and Message Index Writer write format-specific information
 * before the Protobuf payload, in that order.
 * <p>
 * Only one Record can be written because raw Protocol Buffers messages do not contain
 * boundaries that allow concatenated messages to be decoded independently.
 */
```

nifi-extension-bundles/nifi-protobuf-bundle/nifi-protobuf-services/src/main/java/org/apache/nifi/services/protobuf/WriteProtobufResultWithExternalSchema.java
```diff
@@ -0,0 +1,141 @@
...
+        final byte[] payload = serializer.serialize(record);
+        buffered.write(payload);
+        return getSchemaReferenceAttributes();
+    }
+
+    private Map<String, String> getSchemaReferenceAttributes() {
+        return schemaReferenceWriter == null ? Map.of() : schemaReferenceWriter.getAttributes(recordSchema);
+    }
+
+    private void writeConfluentFraming(final OutputStream out) throws IOException {
```
tpalfy:
```suggestion
    private void writeFraming(final OutputStream out) throws IOException {
```

nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/main/java/org/apache/nifi/confluent/schemaregistry/ConfluentProtobufMessageIndexWriter.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private byte[] encodeMessageIndex(final EncodeMessageIndexArguments encodeMessageIndexArguments) {
+        try {
+            final String schemaText = encodeMessageIndexArguments.schemaDefinition().getText();
+            final AntlrProtobufMessageSchemaParser parser = new AntlrProtobufMessageSchemaParser();
+            final List<ProtobufMessageSchema> rootMessages = parser.parse(schemaText);
+            return ProtobufMessageIndexEncoder.encode(rootMessages, encodeMessageIndexArguments.messageName());
+        } catch (final Exception e) {
+            throw new IllegalStateException("Failed to parse protobuf schema", e);
```
tpalfy: Very minor: this exception can be due to encoding failure as well.

```suggestion
throw new IllegalStateException("Failed to generate Protobuf message index", e);
```

nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/test/java/org/apache/nifi/confluent/schemaregistry/ProtobufMessageIndexEncoderTest.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        message Company {
+          string name = 1;
+          Address address = 2;
+        }
+        message Address {
+          string street = 1;
+          string city = 2;
+        }""";
+
```
tpalfy: All nested indexes are zeros. We can add another test to cover a bit more complex case.

```suggestion

    private static final String NON_ZERO_NESTED_INDEX_SCHEMA = """
        syntax = "proto3";
        package com.example.nested;
        message Unused {}
        message Root {
          message ChildZero {}
          message ChildOne {}
          message ChildTwo {
            message GrandchildZero {}
            message GrandchildOne {}
            message GrandchildTwo {}
            message GrandchildThree {}
          }
        }""";
```

nifi-extension-bundles/nifi-confluent-platform-bundle/nifi-confluent-protobuf-message-index-writer/src/test/java/org/apache/nifi/confluent/schemaregistry/ProtobufMessageIndexEncoderTest.java
```diff
@@ -0,0 +1,206 @@
...
+            Arguments.of(DEFAULT_PACKAGE_SCHEMA, new StandardMessageName(Optional.empty(), "Address"), new int[] {2}),
+            Arguments.of(DEFAULT_PACKAGE_SCHEMA, new StandardMessageName(Optional.empty(), "User.Profile"), new int[] {0, 0}),
+            Arguments.of(DEFAULT_PACKAGE_SCHEMA, new StandardMessageName(Optional.empty(), "User.Profile.Settings"), new int[] {0, 0, 0}),
+
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "User"), new int[] {0}),
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "Company"), new int[] {1}),
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "Address"), new int[] {2}),
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "User.Profile"), new int[] {0, 0}),
+            Arguments.of(EXPLICIT_PACKAGE_SCHEMA, new StandardMessageName(Optional.of("com.example.proto"), "User.Profile.Settings"), new int[] {0, 0, 0})
+        );
```
tpalfy:
```suggestion
        ,

            Arguments.of(NON_ZERO_NESTED_INDEX_SCHEMA, new StandardMessageName(Optional.of("com.example.nested"), "Root.ChildTwo.GrandchildThree"), new int[] {1, 2, 3})
        )
```

PR #11511, opened 2026-08-05
NIFI-16172 - Fix: Multiple nifiTooltip hover tooltips can remain open at once and can get stuck open indefinitely

rfellows:
# Summary

[NIFI-16172](https://issues.apache.org/jira/browse/NIFI-16172)

Fixes the shared `nifiTooltip` hover directive so at most one tooltip is visible, stuck-open tooltips self-heal when overlay mouseleave is dropped, and delayed open does not fire after a quick pass-through. Also enables delayed close on comment / validation-error tips in listing tables so users can move onto long scrollable tooltip content.

## Problem
Hovering the “more details” icon column (comments, validation errors, bulletins) in listing tables could leave multiple tooltips open at once. In some cases a tooltip stayed visible indefinitely after the pointer left, until another tip opened or the view was destroyed.

Separately, comment and validation-error tips used `[delayClose]="false"`, so long scrollable tooltips closed as soon as the pointer left the trigger and could not be scrolled. Bulletin tips already used delayed close.

Most noticeable on Controller Services, Reporting Tasks, Parameter Providers, Connectors, Controller Service references, Manage Remote Ports, and Flow Analysis Rules.

## Changes
**Directive** (`nifi-tooltip.directive.ts`):
- Enforce a single active tooltip via a static `openInstance`; attaching a new tip detaches the previous one
- Drive open/close from live `:hover` on trigger and tip instead of a sticky cached hover flag
- Use a self-rescheduling close watchdog that closes only when the pointer is over neither trigger nor tip (overlay mouseleave remains a fast path)
- Gate delayed open on the trigger still being hovered
- Attach overlay listeners to the persistent overlay pane; reset timers / clear the static instance on detach and destroy

**Templates:** set `[delayClose]="true"` on comment / validation-error (and related) tips in the listing tables above so behavior matches bulletins and scrollable content remains reachable.

**Tests:** Vitest coverage for the directive (9 tests), including single-active coordination, keep-open while hovering the tip, self-heal when tip mouseleave is dropped, cancel pending open, and no-open after delay when the trigger is no longer hovered. Spec `afterEach` resets the static `openInstance`.

## Testing
```bash
npx nx test shared --testFile=nifi-tooltip.directive.spec.ts

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-table/connector-table.component.html
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/connectors/ui/connector-table/connector-table.component.html
@@ -35,7 +35,7 @@
                                 <div
                                     class="pointer fa fa-warning has-errors caution-color"
                                     nifiTooltip
-                                    [delayClose]="false"
+                                    [delayClose]="true"
                                     [tooltipComponentType]="ValidationErrorsTip"
                                     [tooltipInputData]="getValidationErrorsTipData(item)"
                                     data-qa="connector-validation-errors"></div>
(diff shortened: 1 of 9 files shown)


mcgilman:
Reviewing...

mcgilman (approved):
Thanks for the PR @rfellows!

PR #11513, opened 2026-08-06
NIFI-16125 Parallelize initial component validation at startup

IPL:
# Summary

[NIFI-16125](https://issues.apache.org/jira/browse/NIFI-16125)

`FlowController.initializeFlow()` performs the initial validation of every component (Controller
 Services, Reporting Tasks, Flow Analysis Rules, Parameter Providers, Processors, Flow Registry
 Clients, and Connectors) sequentially on a single thread via `TriggerValidationTask`. For flows
 with a large number of components, this can make up a significant portion of startup time.

Per review feedback, this is now split into two dedicated `Runnable` implementations instead of
 branching a single class on whether an `ExecutorService` was supplied:

- `TriggerValidationTask` is restored to its original serial-only behavior and continues to serve
  the periodic re-validation task scheduled every 5 seconds.
- `ParallelTriggerValidationTask` is a new class dedicated to the one-time initial sweep performed
  during `FlowController` initialization. It submits each component's and Connector's validation to
  the existing `validationThreadPool` and waits for all of them to complete before returning.
  Validating distinct components concurrently is safe because each `ComponentNode` tracks its own
  validation state independently.

Both classes share a small package-private helper, `ValidatableComponents`, that gathers the same
 set of components (Controller Services, Reporting Tasks, Flow Analysis Rules, Parameter Providers,
 Processors, Flow Registry Clients, and Connectors) for both sweeps, so the two Task implementations
 cannot silently drift apart on which components they validate. Happy to inline this back into both
 classes instead if you'd rather they stay fully independent.

`ParallelTriggerValidationTask.isValidationComplete()` reports whether the most recent run actually
 validated every component. It returns `false` if the run was interrupted, or the executor rejected
 work, before every component could be validated — `FlowController` logs a warning in that case
 instead of unconditionally logging success, avoiding a silent partial-completion that the previous
 implementation would not have surfaced.

Unit tests were split accordingly: `TriggerValidationTaskTest` covers the serial path (unchanged
 behavior), and the new `ParallelTriggerValidationTaskTest` covers the parallel happy path, executor
 rejection, interruption while awaiting completion, and the constructor's null-argument checks.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/validation/ParallelTriggerValidationTask.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/validation/ParallelTriggerValidationTask.java
@@ -1,0 +1,155 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nifi.components.validation;
+
+import org.apache.nifi.components.connector.ConnectorNode;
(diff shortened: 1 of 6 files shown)


exceptionfactory (changes requested):
Thanks for proposing this improvement @IPL.

This is a critical part of the framework initialization, so it will require careful consideration.

On initial review, it seems better to have two different Task implementations, versus conditional behavior based on the presence of the `ExecutorService`. That would clarify the expected behavior at FlowController initialization, versus background scheduling.

IPL:
@exceptionfactory Thanks for the feedback — pushed a follow-up commit that splits this into two
dedicated Task implementations as you suggested: `TriggerValidationTask` stays serial-only for the
periodic re-validation, and the new `ParallelTriggerValidationTask` owns the one-time startup sweep.
PTAL when you get a chance.

PR #11515, opened 2026-08-06
NIFI-16174 - Treat a stateless process group as a single lifecycle unit when starting/stopping a controller service's referencing components

ncover21:
# Summary

[NIFI-16174](https://issues.apache.org/jira/browse/NIFI-16174)

`StandardControllerServiceProvider` scheduled controller-service referencing processors individually even when they belong to a stateless process group. Because a stateless group is a single scheduling unit, this left the group in a mixed running/stopped state with the group node stuck `RUNNING` and unable to recover. It surfaces during break-glass/troubleshooting on connectors whose stateless subgroup references a root-scoped controller service.

The start path already skips stateless members (`DefaultComponentScheduler.startNow` returns early when `resolveExecutionEngine() == STATELESS`); the stop path (`unscheduleReferencingComponents`) did not, so it stopped members individually. Simply ignoring stateless members on stop is not sufficient either: a running stateless member reports `getPhysicalScheduledState() == RUNNING`, so it counts as an active reference in `verifyCanDisable`, which would then block the controller service from being disabled.

This change treats the owning stateless group as a single unit on both paths:

- Resolve each referenced processor's owning stateless group (`STATELESS` -> self, `INHERITED` -> nearest explicit ancestor).
- On stop, call `ProcessGroup.stopProcessing()` once per distinct owning stateless group (its members transition to STOPPED together, so the service can then be disabled), mapping the group's single future to every affected member.
- On start, call `ComponentScheduler.startStatelessGroup()` once per distinct owning stateless group, so "enable service and start referencing components" restarts the group.

Standard (non-stateless) referencing components are unchanged, and the public `ControllerServiceProvider` method signatures are unchanged.

# Tracking

- [x] JIRA issue created: NIFI-16174
- [x] Commit message prefixed with `NIFI-16174`
- [x] `main` builds on the changed modules (checkstyle + PMD pass)


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/service/StandardControllerServiceProvider.java
```diff
@@ -264,6 +296,28 @@ private boolean isRunningOrStarting(final ReportingTaskNode node) {
         return scheduledState == ScheduledState.RUNNING || scheduledState == ScheduledState.STARTING;
     }
 
+    /**
+     * Returns the explicit stateless process group that the given process group belongs to, or {@code null} if the
+     * process group is not part of a stateless group. A group whose execution engine is {@code INHERITED} resolves to
+     * its nearest ancestor that explicitly declares an execution engine.
+     */
+    private ProcessGroup getStatelessGroup(final ProcessGroup start) {
```
markap14: I think there's a minor bug here. This returns the inner-most Process Group that is marked as STATELESS. But we need to operate on the outermost group. So if you have something like `STANDARD -> STANDARD -> INHERITED -> STATELESS -> INHERITED -> INHERITED` this works fine. But if it's `STANDARD -> STANDARD -> INHERITED -> STATELESS -> STATELESS -> STATELESS` we will have an issue because we'll attempt to operate on the inner-most group, which will fail because it is an inner Stateless group.

markap14 (changes requested):
Thanks @ncover21 I called out one minor bug but otherwise I think this makes sense.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/service/StandardControllerServiceProvider.java
ncover21: Yeah you're right, good catch. And it fails silently rather than erroring out: `startProcessing()` logs "Cannot start Process Group ... because its parent is configured to run using the Stateless Engine" and returns, and `stopProcessing()` returns an already completed future without stopping anything, so the caller thinks the stop worked.

Fixed in the latest commit. The helper now walks up while the group resolves to STATELESS and keeps the last one, so it returns the top-most group. Switched to `resolveExecutionEngine()` too so the INHERITED collapsing is handled there rather than recursing on it here, and documented the rule on `ProcessGroup.startProcessing()` / `stopProcessing()`. Tests cover your exact chain, plus one pinning the no-op itself since `stopComponents()` relies on it to terminate the recursion during stateless shutdown. The system test now has a nested subtree whose only referencing processor is in the inner group, it wouldnt catch this otherwise since a referencing processor in the outer group masks the nested no-op.

Also worth flagging, the same innermost-resolution helper is copied in `AffectedComponentSet` (line 461) and `StandardAssetComponentManager` (291), and `StandardProcessGroupDAO.scheduleComponents()` (310) calls start/stopProcessing directly on the requested group so it has the same exposure. Left those alone to keep this PR focused, happy to do a follow up if you'd prefer.

Let me know what you think, thanks!

markap14: Oh good catch. Yeah, would be a good idea to do that as a follow-on, thanks!

markap14:
@ncover21 thanks for the fix. Changes look good to me. Looks like build validation is failing because it's not up to date with `main` branch though.

markap14 (approved):

exceptionfactory (commented):
@ncover21 and @markap14, the integration tests workflow has failed several times with timeouts on the `CreateConnectorIT`. This could be unrelated, but it is not completely clear, so I recommend double-checking for any impact to that integration test.

ncover21:
> @ncover21 and @markap14, the integration tests workflow has failed several times with timeouts on the `CreateConnectorIT`. This could be unrelated, but it is not completely clear, so I recommend double-checking for any impact to that integration test.

@exceptionfactory thats odd, they are consistently passing locally for me. I have merged with main and pushed. Maybe that will help a bit

exceptionfactory:
> > @ncover21 and @markap14, the integration tests workflow has failed several times with timeouts on the `CreateConnectorIT`. This could be unrelated, but it is not completely clear, so I recommend double-checking for any impact to that integration test.
> 
> @exceptionfactory thats odd, they are consistently passing locally for me. I have merged with main and pushed. Maybe that will help a bit

Thanks @ncover21, it appears unrelated on closer review, so no concerns from my perspective now.

PR #11522, opened 2026-08-07
NIFI-16180 - Ensure Rebalancing Worker Stops Before Returning

pvillard31:
# Summary

NIFI-16180 - Ensure Rebalancing Worker Stops Before Returning

This change fixes a recurring race in `TestSocketLoadBalancedFlowFileQueue` by ensuring the rebalancing worker has fully stopped before `stop()` returns. It also prevents stop and start operations from overlapping and safely returns any FlowFiles already picked up by the worker so they remain correctly counted and are redistributed once after restart.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/queue/BlockingSwappablePriorityQueue.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/controller/queue/BlockingSwappablePriorityQueue.java
@@ -51,6 +51,15 @@ public class BlockingSwappablePriorityQueue extends SwappablePriorityQueue {
         }
     }
 
+    @Override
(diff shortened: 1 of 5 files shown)


exceptionfactory (commented):
Thanks for proposing an improvement to this behavior @pvillard31. I'm evaluating the current approach, as well as a potential alternative to consider various ways that Thread locking could occur.

exceptionfactory (commented):
@pvillard31 I submitted an alternative PR #11568 with a more conservative change scoped to the test class.

I'm concerned about potential locking conditions with changes to the `start()` and `stop()` methods, so if the alternative approach addresses the test behavior, I would prefer that direction.

I also considered some options to adjust the implementation of the rebalancing task itself, but I would like to consider that independently if you agree with the test improvement approach.

PR #11533, opened 2026-08-12
NIFI-16189 Deprecated StreamUtils#read(InputStream,byte[] destination, int)

dan-s1:
# Summary

[NIFI-16189](https://issues.apache.org/jira/browse/NIFI-16189)
The implementations of the `read` method in `org.apache.nifi.stream.io.StreamUtils`  and  `readNBytes(byte[], int, int)` in `java.io.Inputstream` method are almost identical. The big difference is when bytes cannot be read from the `InputStream` where  `read` throws an `EOFException` while `readNBytes` breaks out of the while loop and returns the number of byes read even though it may be 0 or -1. Although there is this difference, it seems where `read` was used in `SplitPCAP` and `StandardLoadBalanceProtocol` there are subsequent checks in the code, hence even though an exception is not thrown if there is something wrong it will be noted.

For ease of review, the implementations are brought below.
```
public static void read(final InputStream source, final byte[] destination, final int byteCount) throws IOException {
        if (destination.length < byteCount) {
            throw new IllegalArgumentException();
        }

        int bytesRead = 0;
        int len;
        while (bytesRead < byteCount) {
            len = source.read(destination, bytesRead, byteCount - bytesRead);
            if (len < 0) {
                throw new EOFException("Expected to consume " + byteCount + " bytes but consumed only " + bytesRead);
            }

            bytesRead += len;
        }
    } 
```

```
public int readNBytes(byte[] b, int off, int len) throws IOException {
        Objects.checkFromIndexSize(off, len, b.length);

        int n = 0;
        while (n < len) {
            int count = read(b, off + n, len - n);
            if (count < 0)
                break;
            n += count;
        }
        return n;
    }
```

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/stream/io/StreamUtils.java
+++ b/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/stream/io/StreamUtils.java
@@ -120,7 +120,9 @@ public class StreamUtils {
      * @throws IllegalArgumentException if the given byte array is smaller than <code>byteCount</code> elements.
      * @throws EOFException if the InputStream does not have <code>byteCount</code> bytes in the InputStream
      * @throws IOException if unable to read from the InputStream
+     * @deprecated Use {@link InputStream#readNBytes(byte[], int, int)} instead.
      */
+    @Deprecated(since = "2.12.0", forRemoval = true)
     public static void read(final InputStream source, final byte[] destination, final int byteCount) throws IOException {
         if (destination.length < byteCount) {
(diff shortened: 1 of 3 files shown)


exceptionfactory (approved):
Thanks @dan-s1. On review, the change looks good. The behavioral difference is notable, but the checksum comparison in `StandardLoadBalanceProtocol` ensures expected behavior.

PR #11537, opened 2026-08-13
NIFI-16187 FlowFile grouping strategy in ConsumeKafka

awelless:
# Summary

[NIFI-16187](https://issues.apache.org/jira/browse/NIFI-16187)

Adding a new write strategy to handle conflicting record schemas in `ConsumeKafka`.

The PR consists of 3 logical changes.
1. Improvement in `DataTypeUtils` to handle nullability for absent fields during metge.
1. Introduction of `RecordGroupingStrategy` and refactoring out `CreateNewFlowFileGrouping`. No new functionality is added.
1. Introduction of `MergeSchemaGrouping` and wiring it into `ConsumeKafka`.

The change was verified by running `ConsumeKafka` on a real NiFi installation with a real Kafka broker.

I considered extracting out `OutputStrategy` into a separate SMT, and then adding another converter for schema merging ([branch](https://github.com/awelless/nifi/tree/NIFI-16187-smt-bridge)). But the  `RecordGroupingStrategy` approach turned out to be cleaner.

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [x] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
@@ -59,6 +59,7 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.BitSet;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.EnumMap;
@@ -1603,14 +1604,7 @@ public class DataTypeUtils {
         }
 
         final List<RecordField> otherFields = otherSchema.getFields();
-        if (otherFields.isEmpty()) {
-            return thisSchema;
-        }
-
         final List<RecordField> thisFields = thisSchema.getFields();
-        if (thisFields.isEmpty()) {
-            return otherSchema;
-        }
 
         final Map<String, Integer> fieldIndices = new HashMap<>();
         final List<RecordField> fields = new ArrayList<>();
@@ -1627,7 +1621,7 @@ public class DataTypeUtils {
             fields.add(field);
         }
 
-        final Set<Integer> matchedFieldIndices = new HashSet<>();
+        final BitSet matchedFieldIndices = new BitSet(thisFields.size());
         for (final RecordField otherField : otherFields) {
             Integer fieldIndex = fieldIndices.get(otherField.getFieldName());
 
@@ -1650,7 +1644,7 @@ public class DataTypeUtils {
                 continue;
             }
 
-            matchedFieldIndices.add(fieldIndex);
+            matchedFieldIndices.set(fieldIndex);
 
             // Merge the two fields, if necessary
             final RecordField thisField = fields.get(fieldIndex);
@@ -1661,7 +1655,7 @@ public class DataTypeUtils {
         }
 
         for (int i = 0; i < thisFields.size(); i++) {
-            if (!matchedFieldIndices.contains(i)) {
+            if (!matchedFieldIndices.get(i)) {
                 fields.set(i, makeNullable(fields.get(i)));
             }
         }
--- a/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
@@ -1627,6 +1627,7 @@ public class DataTypeUtils {
             fields.add(field);
         }
 
+        final Set<Integer> matchedFieldIndices = new HashSet<>();
         for (final RecordField otherField : otherFields) {
             Integer fieldIndex = fieldIndices.get(otherField.getFieldName());
 
@@ -1642,16 +1643,26 @@ public class DataTypeUtils {
             }
 
             // If there is no field with the same name then just add 'otherField'.
+            // Fields present in only one schema are nullable in the merged schema,
+            // since the merged schema is a superset of both inputs.
             if (fieldIndex == null) {
-                fields.add(otherField);
+                fields.add(makeNullable(otherField));
                 continue;
             }
+
+            matchedFieldIndices.add(fieldIndex);
 
             // Merge the two fields, if necessary
             final RecordField thisField = fields.get(fieldIndex);
             if (isMergeRequired(thisField, otherField)) {
                 final RecordField mergedField = merge(thisField, otherField);
                 fields.set(fieldIndex, mergedField);
+            }
+        }
+
+        for (int i = 0; i < thisFields.size(); i++) {
+            if (!matchedFieldIndices.contains(i)) {
+                fields.set(i, makeNullable(fields.get(i)));
             }
         }
 
@@ -1667,7 +1678,19 @@ public class DataTypeUtils {
             return true;
         }
 
+        if (thisField.isNullable() != otherField.isNullable()) {
+            return true;
+        }
+
         return !Objects.equals(thisField.getDefaultValue(), otherField.getDefaultValue());
+    }
+
+    private static RecordField makeNullable(final RecordField field) {
+        if (field.isNullable()) {
+            return field;
+        }
+
+        return new RecordField(field.getFieldName(), field.getDataType(), field.getDefaultValue(), field.getAliases(), true);
     }
 
     public static RecordField merge(final RecordField thisField, final RecordField otherField) {
--- a/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
@@ -1413,6 +1413,21 @@ public class TestDataTypeUtils {
     }
 
     @Test
+    public void testMergeSchemasWhenOneSchemaIsStrictSubset() {
+        final RecordSchema schemaA = new SimpleRecordSchema(List.of(
+            new RecordField("x", RecordFieldType.STRING.getDataType(), false),
+            new RecordField("shared", RecordFieldType.STRING.getDataType(), false)));
+        final RecordSchema schemaB = new SimpleRecordSchema(List.of(
+            new RecordField("shared", RecordFieldType.STRING.getDataType(), false)));
+
+        final RecordSchema merged = DataTypeUtils.merge(schemaA, schemaB);
+
+        assertEquals(2, merged.getFieldCount());
+        assertTrue(merged.getField("x").orElseThrow().isNullable());
+        assertFalse(merged.getField("shared").orElseThrow().isNullable());
+    }
+
+    @Test
     public void testMergeSchemasKeepsSharedNonNullableFieldNonNullable() {
         final RecordSchema schemaA = new SimpleRecordSchema(List.of(
             new RecordField("id", RecordFieldType.STRING.getDataType(), false)));
@@ -1598,7 +1613,7 @@ public class TestDataTypeUtils {
     }
 
     @Test
-    public void testMergeSchemasWithEmptySchemaReturnsOtherUnchanged() {
+    public void testMergeSchemasWithEmptySchemaMakesOtherFieldsNullable() {
         final RecordSchema schemaA = new SimpleRecordSchema(List.of(
             new RecordField("id", RecordFieldType.STRING.getDataType(), false)));
         final RecordSchema empty = new SimpleRecordSchema(List.of());
@@ -1606,8 +1621,10 @@ public class TestDataTypeUtils {
         final RecordSchema mergedWithEmptyOther = DataTypeUtils.merge(schemaA, empty);
         final RecordSchema mergedWithEmptyThis = DataTypeUtils.merge(empty, schemaA);
 
-        assertFalse(mergedWithEmptyOther.getField("id").orElseThrow().isNullable());
-        assertFalse(mergedWithEmptyThis.getField("id").orElseThrow().isNullable());
+        assertEquals(1, mergedWithEmptyOther.getFieldCount());
+        assertTrue(mergedWithEmptyOther.getField("id").orElseThrow().isNullable());
+        assertEquals(1, mergedWithEmptyThis.getFieldCount());
+        assertTrue(mergedWithEmptyThis.getField("id").orElseThrow().isNullable());
     }
 
     @Test
--- a/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
+++ b/nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
@@ -47,6 +47,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.DoubleAdder;
@@ -1298,14 +1299,14 @@ public class TestDataTypeUtils {
     @Test
     public void testMergeDataTypesMergesRecordSchemasInsteadOfCreatingChoice() {
         final RecordSchema schemaA = new SimpleRecordSchema(List.of(
-            new RecordField("firstName", RecordFieldType.STRING.getDataType()),
-            new RecordField("lastName", RecordFieldType.STRING.getDataType()),
-            new RecordField("address", RecordFieldType.STRING.getDataType())));
+            new RecordField("firstName", RecordFieldType.STRING.getDataType(), false),
+            new RecordField("lastName", RecordFieldType.STRING.getDataType(), false),
+            new RecordField("address", RecordFieldType.STRING.getDataType(), false)));
 
         final RecordSchema schemaB = new SimpleRecordSchema(List.of(
-            new RecordField("firstName", RecordFieldType.STRING.getDataType()),
-            new RecordField("lastName", RecordFieldType.STRING.getDataType()),
-            new RecordField("age", RecordFieldType.INT.getDataType())));
+            new RecordField("firstName", RecordFieldType.STRING.getDataType(), false),
+            new RecordField("lastName", RecordFieldType.STRING.getDataType(), false),
+            new RecordField("age", RecordFieldType.INT.getDataType(), false)));
 
         final DataType recordTypeA = RecordFieldType.RECORD.getRecordDataType(schemaA);
         final DataType recordTypeB = RecordFieldType.RECORD.getRecordDataType(schemaB);
@@ -1319,6 +1320,10 @@ public class TestDataTypeUtils {
         assertTrue(mergedSchema.getField("lastName").isPresent());
         assertTrue(mergedSchema.getField("address").isPresent());
         assertTrue(mergedSchema.getField("age").isPresent());
+        assertFalse(mergedSchema.getField("firstName").orElseThrow().isNullable());
+        assertFalse(mergedSchema.getField("lastName").orElseThrow().isNullable());
+        assertTrue(mergedSchema.getField("address").orElseThrow().isNullable());
+        assertTrue(mergedSchema.getField("age").orElseThrow().isNullable());
     }
 
     @Test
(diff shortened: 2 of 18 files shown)


nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/convert/CreateNewFlowFileGrouping.java
```diff
@@ -0,0 +1,164 @@
...
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Create New FlowFile strategy: groups by write schema, topic, partition, and grouping attributes,
+ * streaming records into an open writer per group.
+ */
+public class CreateNewFlowFileGrouping implements RecordGroupingStrategy {
```
awelless, line 44: This class was extracted from `AbstractRecordStreamKafkaMessageConverter`

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/convert/MergeSchemaGrouping.java
```diff
@@ -0,0 +1,155 @@
...
+        long minOffset = Long.MAX_VALUE;
+        long minTimestamp = Long.MAX_VALUE;
+
+        MergeGroup(final Map<String, String> attributes) {
+            this.attributes = attributes;
+        }
+
+        void add(final Record recordToWrite, final RecordSchema writeSchema, final ByteRecord consumerRecord) {
+            records.add(recordToWrite);
+            mergedWriteSchema = DataTypeUtils.merge(mergedWriteSchema, writeSchema);
```
pvillard31, line 148: Should fields that are absent from any contributing schema be made nullable in the merged schema? DataTypeUtils.merge preserves their original nullability, so a strict writer can reject records that do not contain a required field.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/test/java/org/apache/nifi/kafka/processors/consumer/convert/MergeSchemaGroupingTest.java
```diff
@@ -0,0 +1,154 @@
...
+        runner.enableControllerService(writerFactory);
+
+        final Processor processor = runner.getProcessor();
+        session = new MockProcessSession(new SharedSessionState(processor, new AtomicLong(0)), processor);
+        logger = runner.getLogger();
+        grouping = new MergeSchemaGrouping(writerFactory, logger, BROKER_URI, true);
+    }
+
+    @Test
+    void testDifferentSchemasInSameTopicPartitionMergeIntoOneFlowFile() throws Exception {
```
pvillard31, line 88: Can we add a test with a schema-validating writer and disjoint non-nullable fields to verify that both records can be written with the merged schema?

pvillard31 (changes requested):

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/convert/MergeSchemaGrouping.java
awelless: That's a good catch.
Do you know whether the current behavior of `DataTypeUtils.merge` is intended? 
From my perspective the intention of `merge` is to provide a schema which will contain a superset of fields of the merged schemas. So that any record which fits schema A or B can be described with the merged schema too.

Offhand, it seems like the improvement should be introduced into `DataTypeUtils.merge`, but if the current behavior is intended, I can add nullability in this class, after merge happens.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/test/java/org/apache/nifi/kafka/processors/consumer/convert/MergeSchemaGroupingTest.java
awelless: If schema reader doesn't rely on schema inference, the fields are not marked as nullable, true.
I added the test, I'm going to push it together with the nullability fix.

awelless:
@pvillard31 I added the field presence check in `DataTypeUtils`. I expect that `merge` method returns a schema which is a superset of both original schemas.

Please let me know if this is not desired, and we want to keep the behavior of `DataTypeUtils.merge` as is. Then I'll move the logic into the Kafka classes.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -274,6 +278,17 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
             .dependsOn(PROCESSING_STRATEGY, ProcessingStrategy.RECORD)
             .build();
 
+    static final PropertyDescriptor SCHEMA_CONFLICT_RESOLUTION = new PropertyDescriptor.Builder()
+            .name("Schema Conflict Resolution")
+            .description("Specifies how to handle records with different schemas within the same topic and partition. "
+                    + "When set to Create New FlowFile, a new FlowFile is created for each distinct schema. "
+                    + "When set to Continue with Merged Schema, all schemas within a group are merged so that records are batched into a single FlowFile.")
```
markap14: We should avoid documenting what each individual value does in the Property description. That information should live in the description of each individual Allowable Value.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
```diff
@@ -274,6 +278,17 @@ public class ConsumeKafka extends AbstractProcessor implements VerifiableProcess
             .dependsOn(PROCESSING_STRATEGY, ProcessingStrategy.RECORD)
             .build();
 
+    static final PropertyDescriptor SCHEMA_CONFLICT_RESOLUTION = new PropertyDescriptor.Builder()
+            .name("Schema Conflict Resolution")
+            .description("Specifies how to handle records with different schemas within the same topic and partition. "
```
markap14: I think this sentence is lacking detail. Given this description, it sounds as if all FlowFiles for a given topic & partition should have the same schema, but that's not the case. Perhaps it should read `Specifies how to handle records with different schemas within the same output FlowFile`?

nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/record/util/DataTypeUtils.java
```diff
@@ -1627,6 +1627,7 @@ public static RecordSchema merge(final RecordSchema thisSchema, final RecordSche
             fields.add(field);
         }
 
+        final Set<Integer> matchedFieldIndices = new HashSet<>();
```
markap14: Not a big deal but given the frequency with which this might get called, it's best to keep this as efficient as we can. So I would suggest using a `BitSet` rather than a `Set<Integer>`

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/convert/MergeSchemaGrouping.java
```diff
@@ -0,0 +1,155 @@
...
+
+            final RecordSchema schemaToWrite;
+            try {
+                schemaToWrite = writerFactory.getSchema(group.attributes, group.mergedWriteSchema);
+            } catch (final SchemaNotFoundException | IOException e) {
+                throw new ProcessException("Failed to determine write schema for Kafka records", e);
+            }
+
+            final Map<String, String> flowFileAttributes = new HashMap<>();
+            final int[] recordCountHolder = new int[1];
```
markap14: This is a bad practice - should prefer using `final AtomicInteger` instead.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/convert/RecordGroupingStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public interface RecordGroupingStrategy {
+
+    void addRecord(
+            ProcessSession session,
+            ByteRecord consumerRecord,
+            Record recordToWrite,
+            RecordSchema writeSchema,
+            Map<String, String> attributes,
+            Map<String, String> groupingAttributes) throws Exception;
```
markap14: We should avoid throwing the general `Exception`

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
```diff
@@ -1392,4 +1397,278 @@ public void testMergeDataTypesWithManyDistinctRecordSchemasCompletesQuickly() {
...
+    public void testMergeSchemasWithEmptySchemaReturnsOtherUnchanged() {
+        final RecordSchema schemaA = new SimpleRecordSchema(List.of(
+            new RecordField("id", RecordFieldType.STRING.getDataType(), false)));
+        final RecordSchema empty = new SimpleRecordSchema(List.of());
+
+        final RecordSchema mergedWithEmptyOther = DataTypeUtils.merge(schemaA, empty);
+        final RecordSchema mergedWithEmptyThis = DataTypeUtils.merge(empty, schemaA);
+
+        assertFalse(mergedWithEmptyOther.getField("id").orElseThrow().isNullable());
+        assertFalse(mergedWithEmptyThis.getField("id").orElseThrow().isNullable());
```
markap14: Given the changes being made here, this is wrong. The intent of the changes is to ensure that the nullability of the returned schema is updated to account for both provided schemas in `DataTypeUtils.merge`, right? So the returned schema here should have `true` for the nullable flag because we're merging an empty schema with a schema having a required field.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
```diff
@@ -1392,4 +1397,278 @@ public void testMergeDataTypesWithManyDistinctRecordSchemasCompletesQuickly() {
...
+
+        final RecordSchema nestedMerged = ((RecordDataType) addressField.getDataType()).getChildSchema();
+        assertEquals(3, nestedMerged.getFieldCount());
+        assertFalse(nestedMerged.getField("street").orElseThrow().isNullable());
+        assertTrue(nestedMerged.getField("city").orElseThrow().isNullable());
+        assertTrue(nestedMerged.getField("zip").orElseThrow().isNullable());
+    }
+
+    @Test
+    public void testMergeSchemasMakesArrayOfRecordSingleSideFieldsNullable() {
```
markap14, line 1528: An important corner case that should be tested here is where one schema is a strict subset of the other. E.g., schemaA with x/shared, schemaB with only shared

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/ConsumeKafkaMergeSchemaIT.java
```diff
@@ -0,0 +1,310 @@
...
+        runner.setProperty(ConsumeKafka.SCHEMA_CONFLICT_RESOLUTION, SchemaConflictResolution.CONTINUE_WITH_MERGED_SCHEMA.getValue());
+        runner.setProperty(ConsumeKafka.AUTO_OFFSET_RESET, AutoOffsetReset.EARLIEST.getValue());
+
+        runner.run(1, false, true);
+
+        produce(topic, List.of(
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, RECORD_WITH_ID, List.<Header>of()),
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, RECORD_WITH_NAME, List.<Header>of())));
+
+        while (runner.getFlowFilesForRelationship(ConsumeKafka.SUCCESS).isEmpty()) {
```
markap14: This can potentially block forever; should fail after some time period or number of attempts

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/ConsumeKafkaMergeSchemaIT.java
```diff
@@ -0,0 +1,310 @@
...
+        runner.setProperty(ConsumeKafka.OUTPUT_STRATEGY, OutputStrategy.USE_VALUE.getValue());
+        runner.setProperty(ConsumeKafka.AUTO_OFFSET_RESET, AutoOffsetReset.EARLIEST.getValue());
+
+        runner.run(1, false, true);
+
+        produce(topic, List.of(
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, RECORD_WITH_ID, List.<Header>of()),
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, RECORD_WITH_NAME, List.<Header>of())));
+
+        while (runner.getFlowFilesForRelationship(ConsumeKafka.SUCCESS).size() < 2) {
```
markap14: This can potentially block forever; should fail after some time period or number of attempts

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/ConsumeKafkaMergeSchemaIT.java
```diff
@@ -0,0 +1,310 @@
...
+
+        runner.run(1, false, true);
+
+        // Publish as one producer batch so the records are available together for a single poll/onTrigger.
+        produce(topic, List.of(
+                new ProducerRecord<>(topic, 0, (String) null, RECORD_WITH_ID, List.<Header>of()),
+                new ProducerRecord<>(topic, 0, (String) null, RECORD_WITH_NAME, List.<Header>of()),
+                new ProducerRecord<>(topic, 1, (String) null, RECORD_WITH_NAME, List.<Header>of())));
+
+        while (totalRecordCount(runner.getFlowFilesForRelationship(ConsumeKafka.SUCCESS)) < 3) {
```
markap14: This can potentially block forever; should fail after some time period or number of attempts

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/ConsumeKafkaMergeSchemaIT.java
```diff
@@ -0,0 +1,310 @@
...
+        runner.setProperty(ConsumeKafka.AUTO_OFFSET_RESET, AutoOffsetReset.EARLIEST.getValue());
+
+        runner.run(1, false, true);
+
+        produce(topic, List.of(
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, RECORD_WITH_ID, List.<Header>of()),
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, INVALID_RECORD, List.<Header>of()),
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, RECORD_WITH_NAME, List.<Header>of())));
+
+        while (runner.getFlowFilesForRelationship(ConsumeKafka.SUCCESS).isEmpty()) {
```
markap14: This can potentially block forever; should fail after some time period or number of attempts

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/ConsumeKafkaMergeSchemaIT.java
```diff
@@ -0,0 +1,310 @@
...
+        runner.setProperty(ConsumeKafka.SCHEMA_CONFLICT_RESOLUTION, SchemaConflictResolution.CONTINUE_WITH_MERGED_SCHEMA.getValue());
+        runner.setProperty(ConsumeKafka.AUTO_OFFSET_RESET, AutoOffsetReset.EARLIEST.getValue());
+
+        runner.run(1, false, true);
+
+        produce(topic, List.of(
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, RECORD_WITH_ID, List.<Header>of()),
+                new ProducerRecord<>(topic, FIRST_PARTITION, (String) null, RECORD_WITH_NAME, List.<Header>of())));
+
+        while (runner.getFlowFilesForRelationship(ConsumeKafka.SUCCESS).isEmpty()) {
```
markap14: This can potentially block forever; should fail after some time period or number of attempts

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/consumer/convert/CreateNewFlowFileGrouping.java
```diff
@@ -0,0 +1,164 @@
...
+    @Override
+    public void finishAllGroups(final ProcessSession session) {
+        for (final Map.Entry<RecordGroupCriteria, RecordGroup> e : recordGroups.entrySet()) {
+            final RecordGroupCriteria criteria = e.getKey();
+            final RecordGroup group = e.getValue();
+
+            final Map<String, String> resultAttrs = new HashMap<>();
+            final int recordCount;
+            try (final RecordSetWriter writer = group.writer()) {
+                final WriteResult wr = writer.finishRecordSet();
```
markap14: Should avoid shorthand, `writeResult` rather than `wr`

markap14 (changes requested):
Thanks @awelless mostly looks good from my perspective, but there are a couple of things that I think need to be cleaned up. I noted several aesthetics concerns, around naming, etc. and some concerns around tests that we need to ensure that we address.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
awelless: Added `testMergeSchemasWhenOneSchemaIsStrictSubset`

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-processors/src/main/java/org/apache/nifi/kafka/processors/ConsumeKafka.java
awelless: I changed the description to `Specifies how records with different schemas are grouped into FlowFiles`. 

I'm not sure we want to put `records with different schemas within the same output FlowFile`, as it isn't necessary that records will be in the same output FlowFile. Perhaps we could've used `records with different schemas within the same batch`, but batch is something known internally only.

nifi-commons/nifi-record/src/test/java/org/apache/nifi/serialization/record/TestDataTypeUtils.java
awelless: Originally empty schemas were treated as missing schemas. Agree, with the nullability logic in-place we should treat them normally.
I removed the special handling for empty schemas, now they follow the normal path.

nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-3-integration/src/test/java/org/apache/nifi/kafka/processors/ConsumeKafkaMergeSchemaIT.java
awelless: I added `runUntil` method in [the abstract consumer test class](https://github.com/apache/nifi/pull/11537/changes#diff-47c8f4fb3eae98f6963918a9e5db7036a25116e82e2a04c0ead0df3054f25de6R83). I switched the tests I added to this method, but other tests still use infinite `while` loop. 
I didn't change them in this PR, but can do a follow up if needed.

awelless (commented):
Thanks for the review @markap14. I addressed all of the comments

pvillard31 (approved):

PR #11543, opened 2026-08-14
NIFI-16201: Instance Classloading support for connector methods

bobpaulin:
# Summary

[NIFI-16201](https://issues.apache.org/jira/browse/NIFI-16201)

Instance Classloading support for connector methods

* Load additional classes before invoking connector methods then unload
* Create a generic factory for instantiating components for these invocations

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation


nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
```diff
@@ -2029,59 +2030,97 @@ public List<ConnectorMethod> getConnectorMethods() {
...
-            for (final MethodArgument methodArgument : methodArguments) {
-                if (ProcessContext.class.equals(methodArgument.type())) {
-                    continue;
-                }
-
-                final String jsonValue = jsonArguments.get(methodArgument.name());
-                if (jsonValue == null && methodArgument.required()) {
-                    throw new IllegalArgumentException("Cannot invoke Connector Method '" + methodName + "' on " + this + " because the required argument '"
-                        + methodArgument.name() + "' was not provided");
+        final boolean classpathDifferent = isClasspathDifferent(processContext.getProperties());
```
markap14: I don't think this check can do what it's intended to do on this code path, and the one case where it does return `true` looks like a bug rather than a feature.

`isClasspathDifferent` was written for configuration verification, where the caller hands the framework a *proposed* set of property values, so asking whether those differ from what is currently live is a meaningful question. Connector method invocation has no proposed configuration: `StandaloneProcessorFacade.invokeConnectorMethod` builds the context via `componentContextProvider.createProcessContext(processorNode, parameterContext)` with no property overrides. So this is comparing the live configuration against itself, and the answer should always be "no difference."

There is one exception, and it is accidental. `AbstractComponentNode.getEffectivePropertyValues()` seeds an entry for every supported descriptor using `descriptor.getDefaultValue()`, while `isClasspathDifferent` looks the current value up through `getEffectivePropertyValue`, which returns `null` for a property that was never explicitly set (`getProperty` returns `PropertyConfiguration.EMPTY`). An unset property that carries a default therefore compares as changed: default value on one side, `null` on the other.

The practical effect is that any component with a `dynamicallyModifiesClasspath` property that has a default value and has not been explicitly configured will be judged "classpath is different" on *every* invocation, and will always be dispatched to a throwaway instance -- including while the processor is running or the service is enabled. `DefaultedDynamicallyModifyClasspath` in the system test extensions is exactly that shape.

That matters beyond wasted work, because the temporary instance never has `@OnScheduled` / `@OnEnabled` / `@OnConfigurationRestored` invoked, yet `ConnectorMethod.allowedStates` defaults to include `RUNNING`. A method explicitly declared as callable while running would silently be handed a fresh, never-started instance.

Suggestion: either plumb real property overrides through the facade so the comparison means something, or drop this half of the condition and let `isReloadAdditionalResourcesNecessary()` make the decision on its own.

The same applies verbatim to the corresponding line in `StandardControllerServiceNode.invokeConnectorMethod`.

markap14 (commented):
Two concerns on the new temporary-instance path in `invokeConnectorMethod`. Details inline; the same `isClasspathDifferent` concern applies identically to `StandardControllerServiceNode`.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/service/StandardControllerServiceNode.java
```diff
@@ -939,59 +940,95 @@ public void migrateConfiguration(final Map<String, String> originalPropertyValue
...
+            final Set<URL> classpathUrls = getAdditionalClasspathResources(configurationContext.getProperties().keySet(),
+                    descriptor -> configurationContext.getProperty(descriptor).getValue());
+            final String classLoaderIsolationKey = getClassLoaderIsolationKey(configurationContext);
+
+            final ClassLoader currentClassLoader = Thread.currentThread().getContextClassLoader();
+            final InstanceClassLoader detectedClassLoader = extensionManager.createInstanceClassLoader(getCanonicalClassName(), getIdentifier(), bundle, classpathUrls, false,
+                        classLoaderIsolationKey);
+            try {
+                Thread.currentThread().setContextClassLoader(detectedClassLoader);
+                final ControllerService tempControllerService = componentInstanceFactory.createControllerServiceInstance(this, detectedClassLoader);
```
markap14: This failure escapes past every caller's error handling.

`createControllerServiceInstance` throws `ControllerServiceInstantiationException`, which extends `RuntimeException`, and nothing here catches it. The processor path does the right thing -- it catches `ProcessorInstantiationException` and wraps it in `InvocationFailedException` -- but there is no equivalent here.

That is unfortunate because failing to instantiate the service under the temporary classloader is the single most likely failure mode of this feature: a driver JAR that is missing, wrong, or incomplete, which is the exact scenario NIFI-16201 describes. Callers are written against `InvocationFailedException` (see `ClasspathVerifyConnector.invokeLoadClassMethod` and `CalculateConnector`), so they would be bypassed entirely and the error would surface as an unhandled runtime exception rather than a reported invocation failure.

Suggestion: catch `ControllerServiceInstantiationException` here and wrap it in `InvocationFailedException` with the method name and component, mirroring the processor path.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
```diff
@@ -2029,59 +2030,97 @@ public List<ConnectorMethod> getConnectorMethods() {
...
+            final ClassLoader currentClassLoader = Thread.currentThread().getContextClassLoader();
+            final InstanceClassLoader detectedClassLoader = extensionManager.createInstanceClassLoader(getCanonicalClassName(), getIdentifier(), bundle, classpathUrls, false,
+                        classloaderIsolationKey);
+            try {
+                Thread.currentThread().setContextClassLoader(detectedClassLoader);
+                final Processor tempProcessor = componentInstanceFactory.createProcessorInstance(this, detectedClassLoader);
+                try {
+                    return invokeConnectorMethodOnComponent(tempProcessor, methodName, jsonArguments, processContext);
+                } finally {
+                    ReflectionUtils.quietlyInvokeMethodsWithAnnotation(OnRemoved.class, tempProcessor, processContext);
```
markap14: Invoking `@OnRemoved` on the temporary instance is riskier here than it is in `verifyConfiguration`, and I think it needs an explicit decision.

The temporary instance is initialized with the live component's identifier, and it is handed the live `ProcessContext`. For controller services it is worse: `StandardComponentInstanceFactory.createControllerServiceInstance` passes `flowController.getStateManagerProvider().getStateManager(identifier)`, so the temporary instance shares the live component's `StateManager`. An `@OnRemoved` method that touches shared state or an external system will therefore act on behalf of a component that has not been removed.

Two examples already in the tree:

* `GetSplunk.onRemoved` calls `context.getStateManager().clear(Scope.CLUSTER)`.
* `ConsumeKinesis.onRemoved` deregisters an EFO consumer against AWS.

The existing precedent in `verifyConfiguration` is defensible because verification is an explicit, user-initiated action and the component is generally not running. Connector methods are different on both counts: they are invoked implicitly by connector code (`DynamicAllowableValuesConnector` and `KafkaToS3.getAvailableTopics` both call them from `fetchAllowableValues`), and `ConnectorMethod.allowedStates` defaults to include `RUNNING`.

No shipped component currently has both a `@ConnectorMethod` and a `dynamicallyModifiesClasspath` property, so this is not reachable today -- but making that combination work is the point of NIFI-16201, so it becomes reachable as soon as the feature is used as intended.

Worth considering whether the temporary instance should receive `@OnRemoved` at all on this path, and if it should, documenting on `ProcessorNode.invokeConnectorMethod` and `ControllerServiceNode.invokeConnectorMethod` that a `@ConnectorMethod` may be dispatched to a throwaway instance whose lifecycle methods have not run and whose `@OnRemoved` will be invoked afterwards. Neither declaration currently has any Javadoc.

Same comment applies to the corresponding line in `StandardControllerServiceNode`.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
```diff
@@ -2029,59 +2030,97 @@ public List<ConnectorMethod> getConnectorMethods() {
...
-                } else {
-                    try {
-                        final Object argumentValue = OBJECT_MAPPER.readValue(jsonValue, methodArgument.type());
-                        argumentValues.add(argumentValue);
-                    } catch (final JsonProcessingException e) {
-                        throw new InvocationFailedException("Failed to deserialize argument '" + methodArgument.name() + "' as type " + methodArgument.type().getName() +
-                                                            " for Connector Method '" + methodName + "' on " + this, e);
-                    }
+                try {
+                    detectedClassLoader.close();
```
markap14: Closing the temporary ClassLoader can permanently disable shared-ClassLoader reuse for a ClassLoader isolation key.

In `StandardExtensionDiscoveringManager.createInstanceClassLoader`, when no `SharedInstanceClassLoader` exists yet for the `BaseClassLoaderKey`, a new one is created with a reference count of 1 and stored with `sharedBaseClassloaders.putIfAbsent(...)`. `InstanceClassLoader.close()` cascades to the parent `SharedInstanceClassLoader.close()`, which decrements the count to 0 and sets `closed = true`. Nothing ever removes the entry from `sharedBaseClassloaders`.

So if a temporary invocation happens to be the first to create the shared base ClassLoader for a key, the map is left holding a closed instance. Every later instance for that key then gets `incrementReferenceCount() == false`, clones the full NAR and ancestor-NAR URL set into its own `SharedInstanceClassLoader`, and `putIfAbsent` silently fails to replace the dead entry -- so sharing is broken for that key for the lifetime of the JVM, and memory grows with the number of instances.

This is reachable when the live component was instantiated with a null isolation key (the `allowsSharedClassloader && classloaderIsolationKey == null` branch creates a plain `InstanceClassLoader` and registers nothing) and a later connector method invocation resolves a non-null key.

The hazard exists in `verifyConfiguration` today, so I am not asking this PR to fix `StandardExtensionDiscoveringManager`. But connector methods are invoked far more often and far less deliberately than verification, so it is worth deciding whether that is acceptable here or whether the map should evict closed entries.

Minor, while this line is being touched: `verifyConfiguration` manages the same object with try-with-resources. Doing the same here would remove the hand-rolled `close()` and its `IOException` handling.

markap14 (commented):
Two more concerns about the lifecycle of the temporary instance and the temporary ClassLoader, both inline.

Separately, and more importantly than either of them: I do not believe any of the five new tests in `ClasspathVerifyIT` can fail without this change, so I am not able to confirm from the PR that the reported problem is actually resolved.

For `testConnectorMethodLoadsClassFromCommittedClasspath`, the only test that exercises `invokeConnectorMethod` on a success path: `VerifyClasspathResource.CLASSPATH_RESOURCE` has no default value, so `isClasspathDifferent` is correctly `false` and the branch is decided entirely by `isReloadAdditionalResourcesNecessary()`. That leaves two possibilities, and both pass on `main`:

1. If the connector's flow update sets the property on the existing processor, `AbstractComponentNode.setProperties` calls `reload(...)`, and `StandardProcessorNode.reload` sets `additionalResourcesFingerprint` using the same value lookup that `isReloadAdditionalResourcesNecessary()` uses. The fingerprints match, the new branch is skipped, and the test runs the pre-existing live-component path -- where the live ClassLoader has just been reloaded with the asset JAR, so `Class.forName` succeeds.
2. If instead flow synchronization creates the processor fresh, the fingerprint is null and the new branch is taken, but `StandardFlowManager` built the live component with `addClasspathUrls(additionalUrls)` in that case, so the live path would have succeeded as well.

The remaining four tests exercise `ProcessorFacade.verify`, which this PR does not change, or only assert that a bogus class name yields a `FAILED` outcome, which holds either way.

The gap looks structural rather than an oversight. Verification can be driven into the interesting state because the caller supplies property overrides; connector method invocation has no override mechanism, so I could not find a state reachable through the connector API where the live component's ClassLoader lacks resources that the `ProcessContext` can resolve. Could you add a test that fails on `main` in the expected way -- `ClassNotFoundException` or `NoClassDefFoundError` out of the connector method because the live component's ClassLoader lacks the resource -- and share the scenario that produces it?

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
bobpaulin: Yes this make sense.  The property overrides won't come into play here in the same sense they do for verify so it makes sense to remove it.  

I'm modifying my approach to instead consider isReloadAdditionalResourcesNecessary in combination with the FlowContextType being WORKING.   In all cases I can think of once the FlowContext is active all classes should be loaded with changes only coming from subsequent applies (where classes should also be loaded).  This is a problem that impacts the WORKING FlowContext exclusively so those are appropriate for the temp instance to be used.  Any ACTIVE flow context should be expected to use the actual instance ClassLoader of that component to avoid the issues you're mentioning above.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
bobpaulin: Thanks @markap14  for calling these out.  It is clear to me now that these could create problems with the current behavior and further create more problems if connector methods were to be added to these components.  

The intent on invoking remove was the define a place where things could be cleaned up from loading additional classes. For example with the DBCPConnectionPool unregisters the JDBC drivers on this place.  The consequence of not doing this is a memory leak from the DriverManager holding on to these driver classes.  

Considering these drawbacks to using OnRemove it seems like it may be helpful to introduce an additional lifecycle annotation to handle these cases specifically.  Happy to follow up this PR with a proposal for that if it makes sense.

For this PR can we consider leaving the OnRemove behavior in place for the verification calls (as it already exists) and excluding the OnRemove call from any Connector Method invocation?  I think this path provides consistency in the existing verify calls without introducing any of the additional bad behaviors you're calling out above for connector methods.

bobpaulin:
Hi @markap14 

> Two more concerns about the lifecycle of the temporary instance and the temporary ClassLoader, both inline.
> 
> Separately, and more importantly than either of them: I do not believe any of the five new tests in `ClasspathVerifyIT` can fail without this change, so I am not able to confirm from the PR that the reported problem is actually resolved.


You are correct.  I believe I have the scenario properly captured now.  A class that would be supplied from the JAR that is available on a method signature AND the JAR's path must be coming from a Parameter Value Reference on the Component.  I've removed the other tests that just provide superficial confirmation of the ability to use classes from the JAR via reflection as this is not relevant to the actual issue.

> 
> For `testConnectorMethodLoadsClassFromCommittedClasspath`, the only test that exercises `invokeConnectorMethod` on a success path: `VerifyClasspathResource.CLASSPATH_RESOURCE` has no default value, so `isClasspathDifferent` is correctly `false` and the branch is decided entirely by `isReloadAdditionalResourcesNecessary()`. That leaves two possibilities, and both pass on `main`:
> 
> 1. If the connector's flow update sets the property on the existing processor, `AbstractComponentNode.setProperties` calls `reload(...)`, and `StandardProcessorNode.reload` sets `additionalResourcesFingerprint` using the same value lookup that `isReloadAdditionalResourcesNecessary()` uses. The fingerprints match, the new branch is skipped, and the test runs the pre-existing live-component path -- where the live ClassLoader has just been reloaded with the asset JAR, so `Class.forName` succeeds.
> 2. If instead flow synchronization creates the processor fresh, the fingerprint is null and the new branch is taken, but `StandardFlowManager` built the live component with `addClasspathUrls(additionalUrls)` in that case, so the live path would have succeeded as well.

I think the introduction of the Parameter Value reference introduces a third path which is what is required for this bug to be reproduced.  Hardcoding the path does not reproduce the bug and the tests pass with or without this change.

> 
> The remaining four tests exercise `ProcessorFacade.verify`, which this PR does not change, or only assert that a bogus class name yields a `FAILED` outcome, which holds either way.

Agreed these are now removed.
> 
> The gap looks structural rather than an oversight. Verification can be driven into the interesting state because the caller supplies property overrides; connector method invocation has no override mechanism, so I could not find a state reachable through the connector API where the live component's ClassLoader lacks resources that the `ProcessContext` can resolve. Could you add a test that fails on `main` in the expected way -- `ClassNotFoundException` or `NoClassDefFoundError` out of the connector method because the live component's ClassLoader lacks the resource -- and share the scenario that produces it?

See above.  I think the scenario is reproducing properly now.  And the fix provided addresses the issue.  I've changed the fix to only involve the temp components in the WORKING context which I believe should address some of the stateful concerns about execution the the ACTIVE context.  If we're running on an ACTIVE context all the classloading concerns should have already been addressed so there's no need for temp components on that path.  

Thank you again for the feedback I think this fix better addresses the concerns you've raised let me know if there are still items that you feel are not addressed.

markap14:
Hey @bobpaulin reviewing this again I'm a bit concerned about the direction here. This adds quite a bit of complexity, it means that any components that have a connector method called cannot makes use of any member variables (or rather they can use them but as soon as they get set the object will be destroyed so it doesn't last), and it means that we have very different paths through the code when evaluating a component in the Working Context vs. the Active Context.

I'd actually recommend we tackle this issue in a different way, by ensuring that we are always reloading the components as necessary based on classpath changes. Please check this PR that I'd propose as an alternative:
https://github.com/apache/nifi/pull/11574

bobpaulin:
Closing in favor of https://github.com/apache/nifi/pull/11574.  Thanks @markap14

PR #11544, opened 2026-08-14
NIFI-16204 Add KafkaProvenanceReportingTask

thomews:
# Summary

[NIFI-16204](https://issues.apache.org/jira/browse/NIFI-16204)

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-provenance-reporting-task/pom.xml
+++ b/nifi-extension-bundles/nifi-kafka-bundle/nifi-kafka-provenance-reporting-task/pom.xml
@@ -13,34 +13,53 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-
     <parent>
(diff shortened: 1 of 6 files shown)


exceptionfactory (changes requested):
Thanks for proposing this new feature @thomews.

Before providing evaluating the code itself, please review the pull request requirements and push an update to the branch with signed commits.

exceptionfactory (changes requested):
Thanks for making the initial updates @thomews.

On a general review, I'm not sure that this is an optimal candidate for inclusion in the project. The properties provide a large number of configuration options on one hand, which can be helpful, but also highlight the complex nature of the implementation. There are also a number of opinionated selections in the implementation, such as field names and JSON formatting. These may be optimal for a number of use cases, but they also highlight the challenge of supporting a generic capability.

There are other issues, like the use of the Jakarta JSON API, versus Jackson, and other more minor implementation concerns.

thomews:
Thanks for the feedback @exceptionfactory. We have already addressed the JSON API concern and switched from Jakarta JSON to Jackson.

Regarding the configuration complexity, one possible direction would be to move the Kafka connection and Record Writer configuration into a separate KafkaRecordSink Controller Service and have the Reporting Task consume a RecordSinkService. This would simplify the Reporting Task configuration, while also separating the provenance reporting concerns from the Kafka-specific configuration.

The question would be whether the Reporting Task should become destination-agnostic, since any RecordSinkService could then be selected. Do you think this would be a better fit for the concerns you raised around configuration complexity?

joewitt:
I definitely appreciate the effort to contribute but I share the view we should not bring this into the project.  I'd go so far as to say we should revisit the reporting task mechanism entirely in the lens of a NiFi 3.0 and consider total removal.

In the meantime this capability can be maintained by the author in a different github repo and if enough demand warrants revisiting and we maintain reporting tasks we could reconisider.

thomews:
Thanks for the detailed review and for sharing your perspective on Reporting Tasks and NiFi 3.0. We’ll keep maintaining this capability in our own repository for now and will follow the discussions around future Reporting Task solutions in NiFi. If the approach is revisited down the road, we’d be happy to adapt and contribute accordingly.

PR #11559, opened 2026-08-17
NIFI-16217 - Fix: Unsaved Parameter Context changes silently lost when navigating to an inherited Parameter Context

rfellows:
# NIFI-16217: Prompt to save before navigating to an inherited Parameter Context

## Summary

Navigating from the Parameter Context edit dialog to an inherited parameter previously discarded any unsaved edits without warning. Users could lose local form changes simply by following an inherited parameter link.

This change adds a save-before-navigate flow for that action. When the edit form is dirty, NiFi prompts with **Save**, **Don't Save**, or **Cancel**. Choosing **Save** persists the update and keeps the user on the completed update review so they can explicitly continue with **Go to Parameter**—there is no automatic navigation after save. Clean forms navigate immediately; invalid dirty forms disable **Save** so users can discard and continue or cancel and keep editing.

JIRA: [NIFI-16217](https://issues.apache.org/jira/browse/NIFI-16217)

## Decision flow

```mermaid
flowchart TD
    A[User clicks Go to inherited Parameter] --> B{Edit form dirty?}
    B -->|No - clean| C[Navigate to inherited Parameter Context]
    B -->|Yes - dirty| D{Form valid / saveable?}
    D -->|Valid| E[Unsaved Changes dialog<br/>Save / Don't Save / Cancel]
    D -->|Invalid| F[Unsaved Changes dialog<br/>Save disabled / Don't Save / Cancel]
    E -->|Cancel| G[Stay on current edit dialog]
    F -->|Cancel| G
    E -->|Don't Save| C
    F -->|Don't Save| C
    E -->|Save| H[Submit Parameter Context update]
    H --> I[Show completed update review<br/>Close + Go to Parameter]
    I -->|Close| J[Close dialog - no navigation]
    I -->|Go to Parameter| C
    note1[No auto-navigate after save]
    H -.-> note1
```

## What's Changed

- Intercept navigation to an inherited Parameter from the Parameter Context edit dialog when the form has unsaved changes
- Add an **Unsaved Changes** confirmation dialog with **Save**, **Don't Save**, and **Cancel**
- Disable **Save** (and focus **Don't Save**) when the form is dirty but not in a saveable state
- On **Save**, submit the Parameter Context update and retain pending navigation until the user chooses **Go to Parameter** on the completed update review (no auto-navigate)
- On **Don't Save**, discard pending edits and navigate; on **Cancel**, remain in the current edit dialog
- Carry optional `highlightedParameterName` through open/edit navigation so the target parameter can be highlighted after arrival
- Extend Parameter Context listing NgRx state/effects/selectors for pending post-update navigation
- Add unit coverage for the confirmation dialog, edit-dialog review actions, parameter-table go-to behavior, and listing effects/reducer paths

## Out of scope

The property table's Parameter Context link still navigates to the bound context rather than to the specific referenced parameter, as noted by the existing TODO in `PropertyTable.canGoToParameter`. Wiring that link to the `highlightedParameterName` capability added here — including extracting the parameter name from the `#{param}` reference and resolving it through inherited Parameter Contexts — is tracked in [NIFI-16240](https://issues.apache.org/jira/browse/NIFI-16240).

## Test plan

- [ ] With a clean Parameter Context edit form, click an inherited parameter and confirm navigation proceeds immediately (no prompt)
- [ ] With a dirty valid form, click an inherited parameter and confirm the Unsaved Changes dialog offers Save / Don't Save / Cancel
- [ ] Choose **Cancel** and confirm the edit dialog remains open with unsaved changes retained
- [ ] Choose **Don't Save** and confirm navigation proceeds and pending edits are discarded
- [ ] Choose **Save**, confirm the update completes, and confirm the review actions show **Close** and **Go to Parameter** (no automatic navigation)
- [ ] From the completed review, click **Go to Parameter** and confirm navigation to the inherited Parameter Context (with parameter highlight when applicable)
- [ ] From the completed review, click **Close** and confirm the dialog closes without navigating
- [ ] With a dirty invalid form, click an inherited parameter and confirm **Save** is disabled, messaging explains the form is not saveable, and **Don't Save** / **Cancel** behave as above
- [ ] Confirm a normal Apply/Save without inherited navigation still shows only **Close** on the completed update review

### DIFF
--- a/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/parameter-contexts/state/parameter-context-listing/parameter-context-listing.effects.ts
+++ b/nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/parameter-contexts/state/parameter-context-listing/parameter-context-listing.effects.ts
@@ -46,9 +46,18 @@ import {
     selectSaving,
     selectUpdateRequest,
     selectUpdateRequestParameterContextId,
-    selectDeleteUpdateRequestInitiated
+    selectDeleteUpdateRequestInitiated,
+    selectHasPendingPostUpdateNavigation,
+    selectPostUpdateNavigation,
+    selectPostUpdateNavigationBoundary,
+    selectPostUpdateNavigationState
 } from './parameter-context-listing.selectors';
-import { EditParameterRequest, EditParameterResponse, ParameterContextUpdateRequest } from '../../../../state/shared';
+import {
+    EditParameterRequest,
+    EditParameterResponse,
+    ParameterContextUpdateRequest,
+    PostUpdateNavigationState
+} from '../../../../state/shared';
 import { EditParameterDialog } from '../../../../ui/common/edit-parameter-dialog/edit-parameter-dialog.component';
 import { OkDialog } from '../../../../ui/common/ok-dialog/ok-dialog.component';
 import { ErrorHelper } from '../../../../service/error-helper.service';
@@ -58,6 +67,8 @@ import { BackNavigation } from '../../../../state/navigation';
 import { isDefinedAndNotNull, MEDIUM_DIALOG, SMALL_DIALOG, XL_DIALOG, NiFiCommon, Storage } from '@nifi/shared';
 import { ErrorContextKey } from '../../../../state/error';
 import { EditParameterContext } from '../../../../ui/common/parameter-context/edit-parameter-context/edit-parameter-context.component';
+import { SaveParameterContextChangesDialog } from '../../../../ui/common/parameter-context/save-parameter-context-changes-dialog/save-parameter-context-changes-dialog.component';
+import { EditParameterContextUpdate } from '../../../../ui/common/parameter-context';
 
 @Injectable()
 export class ParameterContextListingEffects {
@@ -295,7 +306,8 @@ export class ParameterContextListingEffects {
                     map((response) =>
                         ParameterContextListingActions.openParameterContextDialog({
                             request: {
-                                parameterContext: response
+                                parameterContext: response,
+                                highlightedParameterName: request.highlightedParameterName
                             }
                         })
                     ),
@@ -326,7 +338,8 @@ export class ParameterContextListingEffects {
                     const editDialogReference = this.dialog.open(EditParameterContext, {
                         ...XL_DIALOG,
                         data: {
-                            parameterContext: request.parameterContext
+                            parameterContext: request.parameterContext,
+                            highlightedParameterName: request.highlightedParameterName
                         }
                     });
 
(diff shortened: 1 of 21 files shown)


mcgilman:
Will review...

nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/parameter-context/index.ts
```diff
@@ -15,10 +15,18 @@
  * limitations under the License.
  */
 
-import { ParameterContextEntity } from '../../../state/shared';
+import { ParameterContextEntity, PostUpdateNavigationState } from '../../../state/shared';
 
 export interface EditParameterContextRequest {
```
mcgilman: `EditParameterContextRequest` is declared here identically to the one in `apps/nifi/src/app/pages/parameter-contexts/state/parameter-context-listing/index.ts`. This copy isn't used by any production code — only `edit-parameter-context.component.spec.ts` imports it, for test-fixture typing. Since this PR already touches both files, can we drop this duplicate and have the spec import the type from the state module instead?

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/parameter-contexts/ui/parameter-context-listing/parameter-context-listing.component.ts
```diff
@@ -52,6 +53,7 @@ import { ParameterContextTable } from './parameter-context-table/parameter-conte
 })
 export class ParameterContextListing implements OnInit {
     private store = inject<Store<ParameterContextListingState>>(Store);
+    private router = inject(Router);
```
mcgilman, line 56: This introduces new `Router`-based logic to read `highlightedParameterName` off navigation state (see the `lastSuccessfulNavigation()` call below), but `parameter-context-listing.component.spec.ts` wasn't updated — there's no test asserting this reads the state and forwards it into `getEffectiveParameterContextAndOpenDialog` correctly. Could we add coverage for this?

Separate question (non-blocking): `lastSuccessfulNavigation()` is read after an async pipeline (`switchMap`/`filter`/`take(1)` gated on the parameter context appearing in the store). If that resolution is delayed and another navigation completes first, could the wrong `highlightedParameterName` get picked up? Worth a second look, even if the impact is only a wrong/missing highlight.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/parameter-contexts/state/parameter-context-listing/parameter-context-listing.effects.ts
```diff
@@ -338,6 +351,78 @@ export class ParameterContextListingEffects {
                             map((parameterContexts) => parameterContexts.filter((pc) => pc.id != parameterContextId))
                         );
                     editDialogReference.componentInstance.saving$ = this.store.select(selectSaving);
+                    editDialogReference.componentInstance.hasPendingPostUpdateNavigation$ = this.store.select(
+                        selectHasPendingPostUpdateNavigation
+                    );
+
+                    const navigateToInheritedParameter = (
```
mcgilman: `navigateToInheritedParameter` and the `goTo` closure below it are fairly large and nested inside this effect. Consider extracting them into named, independently-testable functions for readability — not blocking, just a suggestion.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/parameter-contexts/ui/parameter-context-listing/parameter-table/parameter-table.component.ts
```diff
@@ -74,11 +82,15 @@ export class ParameterTable implements AfterViewInit, ControlValueAccessor {
     private store = inject<Store<ParameterContextListingState>>(Store);
```
mcgilman, line 82: This adds highlight/scroll-to support within the parameter context editor, but I don't see corresponding changes to `property-table.component.ts` or any parameter-reference-parsing utility that would let a user click a `#{param}` reference inside a processor's/controller service's property value and jump to the highlighted parameter here. Is that intentionally out of scope for this PR, or planned separately? Would help to note it in the PR description either way.

mcgilman (changes requested):
Overall this looks correct — the postUpdateNavigation dirty-check flow, the new save-changes-prompt dialog, and the end-to-end threading of highlightedParameterName all work as expected. A few things below before this can merge, plus a couple of questions.

nifi-frontend/src/main/frontend/apps/nifi/src/app/ui/common/parameter-context/index.ts
rfellows: Dropped. One clarification on the premise: this copy was used by production code — `parameter-context-listing.actions.ts` was importing `EditParameterContextRequest` from here, while `edit-parameter-context.component.ts` imported the state-module copy. All three importers (actions, component, spec) now reference `pages/parameter-contexts/state/parameter-context-listing`.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/parameter-contexts/ui/parameter-context-listing/parameter-context-listing.component.ts
rfellows: Added coverage in `parameter-context-listing.component.spec.ts`: it forwards `highlightedParameterName` from the navigation state into `getEffectiveParameterContextAndOpenDialog`, dispatches without a highlight when there is no navigation state, and does not dispatch at all until the parameter context is present in the store.

On the timing question — you're right, and I've changed it. `lastSuccessfulNavigation` is a signal and was being read inside `subscribe()`, i.e. after the `switchMap`/`take(1)` that waits for the context to load, so a navigation completing during that wait would have been attributed to this request. It is now read in a `map` immediately after the route id emits, and the captured value is carried through the async wait.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/parameter-contexts/state/parameter-context-listing/parameter-context-listing.effects.ts
rfellows: Extracted. They are now `navigateToInheritedParameter(...)` and `goToInheritedParameter(...)` private methods on `ParameterContextListingEffects`, taking the source parameter context id and the edit dialog reference as arguments rather than closing over them. The `openParameterContextDialog$` body is about 50 lines shorter as a result.

nifi-frontend/src/main/frontend/apps/nifi/src/app/pages/parameter-contexts/ui/parameter-context-listing/parameter-table/parameter-table.component.ts
rfellows: Intentionally out of scope here. This PR adds the capability — `highlightedParameterName` on the Parameter Context edit route — and consumes it for inherited-parameter navigation within the Parameter Context editor. Wiring the property table's existing link to it (extracting the parameter name from the reference, threading it through `goToParameter`, and resolving the reference through inherited contexts so the highlight lands in the context that actually defines the parameter) is tracked in [NIFI-16240](https://issues.apache.org/jira/browse/NIFI-16240). That is also what the existing TODO in `PropertyTable.canGoToParameter` refers to. I have noted the boundary in the PR description.

mcgilman (approved):

PR #11561, opened 2026-08-18
NIFI-16222 Improved searchable select keyboard navigation for grouped options

rfellows:
## Summary

Fixes [NIFI-16222](https://issues.apache.org/jira/browse/NIFI-16222): when a searchable select renders **grouped** options, keyboard navigation still highlighted rows but failed to scroll the active option into view and failed to select on Enter. Ungrouped searchable selects were unaffected; mouse selection continued to work.

## Root cause

Grouped, non-virtual templates emit **group-scoped** option element ids (`groupId-optIdx`), while keyboard scroll and Enter selection looked up options using the **flat** active index id. The highlighted row’s DOM id never matched what scroll/`getElementById` and Enter selection expected, so:

- Arrow navigation could move the highlight off-screen without scrolling
- Enter could not resolve the highlighted option for selection

## Changes

Limited to `nifi-frontend` shared `searchable-select`:

- Tag the active option with a stable `activeOptionId`
  - Grouped: driven by `activeOptionRef` (option object identity) so duplicate values across groups mark exactly one row
  - Ungrouped: driven by `activeTemplateIndex`
- Scroll into view via `activeOptionId` after change detection so the id is present in the DOM before lookup
- Enter selects through the matching `MatOption` via `_selectViaInteraction()` (same user-input path Material uses for a click), with a non-ghost value fallback when needed
- Disarm `_allowNextValueChange` on Escape, panel close, and Enter on an already-selected option so closed-panel arrow auto-select cannot leak through as an intentional change
- Grouped `@for` tracking uses option object identity (values may not be unique)
- Document that **grouped + virtual + duplicate values** remains unsupported (value-based virtual highlight)

## Why this is safe / low regression risk

- **Blast radius**: changes target the non-virtual keyboard path (scroll + Enter + nav reset). Mouse selection and ungrouped keyboard behavior are preserved; virtual scrolling is unchanged aside from documenting the existing grouped/duplicate limitation.
- **Same Material path as click**: Enter goes through `MatOption._selectViaInteraction()`, so multi-select toggle and `(valueChange)` flow match a real user click rather than a synthetic DOM click / flat-id scheme.
- **Ghost exclusion**: Enter resolution filters ghost options before value fallback, avoiding accidental selection of placeholder/ghost rows when a value is missing from an async batch.
- **Async option batches**: navigation state (including `activeOptionRef`) resets when options are replaced, so stale highlights cannot select the wrong row after a batch update.
- **Duplicate-value safety (non-virtual grouped)**: reference identity ensures a single `activeOptionId` and a deterministic Enter target when the same value appears in more than one group.
- **Quality gates**: searchable-select suite is **126** specs, including new coverage for grouped scroll/Enter, multi-select, duplicate values, Escape/panel-close disarm, async batch replace, and ghosts. Shared frontend suite green as applicable before merge.

### Residual notes for reviewers

- **Private Material API**: `_selectViaInteraction` is private. It is the canonical “user selected this option” entry point (`select()`/`deselect()` skip user-input / multi-select toggle semantics). Material/CDK bumps should re-run this suite; the Enter specs are the tripwire if that method changes or disappears.
- **Known unsupported combo**: grouped + virtual scrolling with duplicate values across groups remains unsupported (documented on the input and covered by a documenting test). No current usage is expected to combine them.

## Test plan

- [ ] Open a connector configuration / property UI that uses a searchable select with **grouped** options (e.g. secrets / credential-style property with group headers)
- [ ] Open the dropdown, focus search, Arrow Down until the highlight would leave the visible area — list scrolls to keep the active option in view
- [ ] With a grouped option highlighted, press Enter — option is selected and panel closes (single-select)
- [ ] Confirm mouse click selection of grouped options still works
- [ ] Spot-check an **ungrouped** searchable select: arrows + Enter still behave as before
- [ ] Optional: multi-select grouped list — Enter toggles the highlighted row without closing; second Enter toggles off the same row when values duplicate across groups

### DIFF
--- a/nifi-frontend/src/main/frontend/libs/shared/src/components/searchable-select/searchable-select.component.html
+++ b/nifi-frontend/src/main/frontend/libs/shared/src/components/searchable-select/searchable-select.component.html
@@ -152,6 +152,9 @@
                                     {{ getGroupHeaderLabel(item) }}
                                 </div>
                             } @else {
+                                <!-- Virtual highlight is value-based (isOptionActiveByValue). Grouped +
+                                     virtual with duplicate values across groups is unsupported; see
+                                     enableVirtualScrolling input comment. -->
                                 <multi-select-option
                                     [value]="getItemValue(item)"
                                     [disabled]="disabled || getItemDisabled(item)"
@@ -208,12 +211,18 @@
                                             {{ group.groupLabel }}
                                         </div>
                                     }
-                                    @for (option of group.options; track option.value; let optIdx = $index) {
+                                    <!-- Track by object identity: values may not be unique within a group.
+                                         option.value as track would collide and break @for reconciliation. -->
+                                    @for (option of group.options; track option; let optIdx = $index) {
                                         <multi-select-option
                                             [value]="option.value"
                                             [disabled]="disabled || option.disabled"
-                                            [id]="getOptionId(group.groupId + '-' + optIdx)"
-                                            [class.mat-mdc-option-active]="isOptionActiveByValue(option.value)"
+                                            [id]="
+                                                activeOptionRef() === option
+                                                    ? activeOptionId
+                                                    : getOptionId(group.groupId + '-' + optIdx)
+                                            "
+                                            [class.mat-mdc-option-active]="activeOptionRef() === option"
                                             role="option">
                                             @if (option.svgIcon) {
                                                 <mat-icon class="shrink-0" [svgIcon]="option.svgIcon"></mat-icon>
@@ -242,7 +251,7 @@
                                     <multi-select-option
                                         [value]="option.value"
                                         [disabled]="disabled || option.disabled"
-                                        [id]="getOptionId(i)"
+                                        [id]="activeTemplateIndex() === i ? activeOptionId : getOptionId(i)"
                                         [class.mat-mdc-option-active]="activeTemplateIndex() === i"
                                         role="option">
                                         @if (option.svgIcon) {
--- a/nifi-frontend/src/main/frontend/libs/shared/src/components/searchable-select/searchable-select.component.html
+++ b/nifi-frontend/src/main/frontend/libs/shared/src/components/searchable-select/searchable-select.component.html
@@ -79,7 +79,7 @@
             </div>
 
             <!-- Unified ghost options for both virtual and non-virtual modes -->
-            @for (ghostOption of getGhostOptions(); track ghostOption.value) {
+            @for (ghostOption of getGhostOptions(); track $index) {
                 <multi-select-option
                     [value]="ghostOption.value"
                     [disabled]="disabled || ghostOption.disabled"
@@ -152,20 +152,13 @@
                                     {{ getGroupHeaderLabel(item) }}
                                 </div>
                             } @else {
-                                <!-- Virtual highlight is value-based (isOptionActiveByValue). Grouped +
-                                     virtual with duplicate values across groups is unsupported; see
-                                     enableVirtualScrolling input comment. -->
                                 <multi-select-option
                                     [value]="getItemValue(item)"
                                     [disabled]="disabled || getItemDisabled(item)"
                                     [virtuallySelected]="isVirtualOptionSelected(getItemValue(item))"
-                                    [id]="
-                                        isOptionActiveByValue(getItemValue(item))
-                                            ? activeOptionId
-                                            : getOptionId(virtualIndex)
-                                    "
+                                    [id]="isVirtualOptionActive(item) ? activeOptionId : getOptionId(virtualIndex)"
                                     [attr.aria-selected]="isVirtualOptionSelected(getItemValue(item))"
-                                    [class.mat-mdc-option-active]="isOptionActiveByValue(getItemValue(item))"
+                                    [class.mat-mdc-option-active]="isVirtualOptionActive(item)"
                                     role="option">
                                     @if (getItemSvgIcon(item)) {
                                         <mat-icon class="shrink-0" [svgIcon]="getItemSvgIcon(item)"></mat-icon>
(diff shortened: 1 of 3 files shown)


nifi-frontend/src/main/frontend/libs/shared/src/components/searchable-select/searchable-select.component.html
scottyaslan, line 167: **Issue**

Non-virtual grouped rendering now highlights by reference identity (`activeOptionRef() === option`), which is the right contract. The virtual template still highlights by value (`isOptionActiveByValue(getItemValue(item))`, template ~163-168), which leaves grouped plus virtual with three distinct defects rather than one cosmetic one.

1. Duplicate DOM ids. `activeOptionId` is a single fixed id (`getUniqueId('option-active')`), so value-based assignment hands that same id to every row sharing the active value. Two rows can carry it at once, which is invalid DOM and makes `aria-activedescendant` and `document.getElementById` ambiguous.

2. Arrow order does not match render order, independent of duplicates. `updateActiveTemplateIndex()` derives the active option from `getVisibleOptions()[_activeOptionIndex]`, and in virtual mode `getVisibleOptions()` returns `this._options` in source order. The rendered list comes from `getVirtualVisibleItems()` via `groupedOptions()`, which emits ungrouped first and then groups sorted with `localeCompare`. For interleaved input the nth arrow press does not land on the nth rendered row, and `getVirtualIndexForOptionIndex()` scrolls to the wrong row. Non-virtual grouped avoids this only because its `getVisibleOptions()` branch returns `groups.flatMap((g) => g.options)`, the same order the template renders.

3. Enter is still value-resolved in virtual mode. `toggleOption()` takes the `enableVirtualScrolling()` branch and updates `_virtualSelectedValues` by value, never consulting `activeOptionId` or `activeOptionRef` the way the non-virtual branch does. With duplicate values Enter cannot express which row was highlighted, even if the highlight is corrected.

**Why the current spec does not protect this**

`documents that virtual + groups highlight is value-based (duplicate values unsupported)` asserts the defective behavior: `isOptionActiveByValue('dup') === true` and `visible.every((o) => isOptionActiveByValue(o.value)) === true`. That spec fails only if someone fixes the component. It cannot fail when a caller configures grouped plus virtual plus duplicate values and reproduces the ambiguity in the UI. The comment on this input and the template comment are the only remaining guardrails, and comments do not fail a build.

**Fix**

Please bring grouped plus virtual to the same identity contract as grouped non-virtual in this PR.

1. Highlight by identity in the virtual template: replace `isOptionActiveByValue(getItemValue(item))` with a reference comparison against the item's option object for both `[id]` and `[class.mat-mdc-option-active]`. This needs no restructuring today, because `groupedOptions()` pushes the original `opt` references (`groupMap.get(opt.group)!.push(opt)`) and `getVirtualVisibleItems()` pushes those same objects (`group.options.forEach((opt) => items.push(opt))`), so `activeOptionRef() === asOptionItem(item)` is a valid comparison.
2. Make the navigation list match the rendered list: in the `enableVirtualScrolling()` branch of `getVisibleOptions()`, return grouped order when `hasGroupedOptions()` is true, using the same `groups.flatMap((g) => g.options)` as the non-virtual branch, so `_activeOptionIndex`, `activeOptionRef`, and `getVirtualIndexForOptionIndex()` all agree with what `cdkVirtualFor` renders.
3. Give `trackByVirtualItem` a group-scoped identity for option rows. It currently returns `item.value`, so two rows sharing a value across groups collide during `cdkVirtualFor` diffing. Group id plus value, or the option reference, both work.
4. Resolve Enter from the highlighted row in the virtual branch of `toggleOption()` rather than by value.
5. Replace the documenting spec with a regression asserting exactly one active row and exactly one element carrying `activeOptionId`, plus coverage for arrow order matching rendered grouped order, Enter selecting the highlighted duplicate, a second Enter toggling the same row in multi-select, and an active row scrolled outside the rendered buffer, since CDK recycles nodes and the active element is not guaranteed to be mounted.

**Preferred outcome**

Grouped plus virtual behaves like grouped non-virtual: at most one active row, one `activeOptionId` in the DOM, arrow order equal to render order, and Enter acting on the highlighted row. The `documents that ... is value-based` spec is gone, replaced by regressions that fail if the ambiguity returns.

If that is more than you want in this change, the acceptable alternative is a follow-up Jira referenced from this input comment together with a guard that makes the unsupported combination impossible to configure silently, for example a development-mode warning when `enableVirtualScrolling()` is set on a grouped list. Landing only comments plus a spec that locks the defect leaves nothing that fails when a caller trips it.

nifi-frontend/src/main/frontend/libs/shared/src/components/searchable-select/searchable-select.component.spec.ts
```diff
@@ -1835,6 +2427,29 @@ describe('SearchableSelect', () => {
 
                 expect(component.trackByVirtualItem(1, option)).toBe('aws-1');
             });
+
+            it('documents that virtual + groups highlight is value-based (duplicate values unsupported)', async () => {
```
scottyaslan: **Issue**

This spec locks the defect rather than the contract. `expect(component.isOptionActiveByValue('dup')).toBe(true)` and `expect(visible.every((o) => component.isOptionActiveByValue(o.value))).toBe(true)` pass precisely because two rows sharing a value are both active. It fails only when someone corrects the component, and never when a caller configures grouped plus virtual plus duplicate values and reproduces the ambiguity.

**Fix**

Replace it with a regression matching the grouped non-virtual contract: exactly one active row, exactly one element carrying `activeOptionId`, and Enter acting on the highlighted row. The component changes this depends on are described in the comment on `enableVirtualScrolling`.

**Preferred outcome**

No spec in this file asserts that duplicate rows are simultaneously active.

scottyaslan (changes requested):
Reviewed the grouped keyboard-navigation changes. The non-virtual grouped path looks right: `track option` over object identity, `activeOptionRef() === option` driving both `activeOptionId` and the active class, Enter resolving through `activeOptionId` then `activeOptionRef` before any value fallback, and the value-change permission disarmed on Enter, Escape, and panel close. The specs cover those paths well.

Requesting changes on one point: grouped plus virtual scrolling. The virtual template still highlights by value, and the accompanying spec asserts the ambiguous behavior rather than a contract, so nothing in the suite fails if a caller configures that combination. Details and a concrete fix outline are in the inline comment on `enableVirtualScrolling`.

scottyaslan (approved):
LGTM

PR #11562, opened 2026-08-19
NIFI-16226 - Parameter Context update fails when adding provider-backed inheritance in a cluster

pvillard31:
# Summary

NIFI-16226 - Parameter Context update fails when adding provider-backed inheritance in a cluster

# Tracking

### Issue Tracking

### Pull Request Tracking

### Pull Request Formatting

# Verification

### Build

  - [ ] JDK 21
  - [ ] JDK 25

### Licensing

### Documentation

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
@@ -1791,6 +1791,7 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
         final Map<String, Parameter> parameterUpdates = parameterContextDAO.getParameters(parameterContextDto, parameterContext);
         final List<ParameterContext> inheritedParameterContexts = parameterContextDAO.getInheritedParameterContexts(parameterContextDto);
         final Map<String, Parameter> proposedParameterUpdates = parameterContext.getEffectiveParameterUpdates(parameterUpdates, inheritedParameterContexts);
+        final Map<ParameterDescriptor, Parameter> localParameters = parameterContext.getParameters();
         final Map<String, ParameterEntity> parameterEntities = parameterContextDto.getParameters().stream()
                 .collect(Collectors.toMap(entity -> entity.getParameter().getName(), Function.identity()));
         parameterContextDto.getParameters().clear();
@@ -1798,6 +1799,7 @@ public class StandardNiFiServiceFacade implements NiFiServiceFacade {
         for (final Entry<String, Parameter> entry : proposedParameterUpdates.entrySet()) {
             final String parameterName = entry.getKey();
             final Parameter parameter = entry.getValue();
+            final boolean locallyOwned = localParameters.containsKey(new ParameterDescriptor.Builder().name(parameterName).build());
             final ParameterEntity parameterEntity;
             if (parameterEntities.containsKey(parameterName)) {
                 parameterEntity = parameterEntities.get(parameterName);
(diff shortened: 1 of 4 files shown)


nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/api/ParameterContextResourceTest.java
```diff
@@ -0,0 +1,182 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.same;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class ParameterContextResourceTest {
```
exceptionfactory: The `public` modifier can be removed

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/test/java/org/apache/nifi/web/StandardNiFiServiceFacadeTest.java
```diff
@@ -2429,6 +2436,119 @@ public void testGetConnectorParameterContextReturnsNullWhenNoBoundContext() {
...
+    public void testGetComponentsAffectedByParameterContextUpdateTwicePreservesInheritedProviderProvenance() {
+        final String targetContextId = "target-context";
+        final String inheritedContextId = "inherited-context";
+        final String inheritedParameterName = "inherited-provider-param";
+        final String processorId = "processor-id";
+
+        final ParameterDescriptor inheritedDescriptor = new ParameterDescriptor.Builder().name(inheritedParameterName).build();
+        final Parameter inheritedParameter = new Parameter.Builder()
+                .descriptor(inheritedDescriptor)
+                .value("provider-value")
```
exceptionfactory: It looks like like `provider-value` should be declared as a variable and reused instead of stated multiple times across different lines in this method

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -1810,14 +1812,31 @@ private void setEffectiveParameterUpdates(final ParameterContextDTO parameterCon
...
 
-            // Parameter is inherited if either this is the removal of a parameter not directly in this context, or it's parameter not specified directly in the DTO
-            final boolean isInherited = (parameter == null && !parameterContext.getParameters().containsKey(new ParameterDescriptor.Builder().name(parameterName).build()))
-                    || (parameter != null && !parameterEntities.containsKey(parameterName));
-            parameterEntity.getParameter().setInherited(isInherited);
+            if (parameter == null) {
+                parameterEntity.getParameter().setInherited(!locallyOwned);
+            } else {
+                final ParameterContext containingParameterContext = getContainingParameterContext(parameterContext, parameter, locallyOwned);
+                final boolean isInherited = !locallyOwned && !Objects.equals(parameterContext.getIdentifier(), containingParameterContext.getIdentifier());
```
exceptionfactory, line 1822: If a user has configured an alias using `#{Parameter Name}` does this approach handle it? It seems like the inheritance check would end up overriding that reference, using the resolved value.

exceptionfactory (commented):
Thanks for addressing this issue @pvillard31. I noted one question around inherited parameter handling.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
pvillard31: This was a real issue in the initial implementation. `getEffectiveParameterUpdates()` resolves `#{...}` references, so a locally configured alias could have been added to the DTO using its resolved value rather than preserving the configured reference.

I updated `setEffectiveParameterUpdates()` so that locally owned parameters are represented using the raw parameter from the target context, while the effective parameter is still used to determine the provided/inherited status and source context. I also extended the repeated-analysis test with a local alias and verified that its raw `#{...}` value and local provenance are preserved across both passes. Inherited provider-backed parameters continue to use their effective values.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -1807,17 +1810,35 @@ private void setEffectiveParameterUpdates(final ParameterContextDTO parameterCon
                 parameterDTO.setName(parameterName);
                 parameterEntity.setParameter(parameterDTO);
             } else {
-                parameterEntity = dtoFactory.createParameterEntity(parameterContext, parameter, revisionManager, parameterContextDAO);
+                final Parameter entityParameter = locallyOwned ? localParameters.get(parameterDescriptor) : parameter;
+                parameterEntity = dtoFactory.createParameterEntity(parameterContext, entityParameter, revisionManager, parameterContextDAO);
```
exceptionfactory: This appears to be a latent issue, but `createParameterEntity()` calls `createParameterDto()`, which substitutes a hard-coded mask string when the Parameter is marked as sensitive. This seems like it will cause issues when persisting the values.

nifi-framework-bundle/nifi-framework/nifi-web/nifi-web-api/src/main/java/org/apache/nifi/web/StandardNiFiServiceFacade.java
```diff
@@ -1807,17 +1810,35 @@ private void setEffectiveParameterUpdates(final ParameterContextDTO parameterCon
                 parameterDTO.setName(parameterName);
                 parameterEntity.setParameter(parameterDTO);
             } else {
-                parameterEntity = dtoFactory.createParameterEntity(parameterContext, parameter, revisionManager, parameterContextDAO);
+                final Parameter entityParameter = locallyOwned ? localParameters.get(parameterDescriptor) : parameter;
```
exceptionfactory: It looks like this introduces a discrepancy for how the parameter checking occurs when adding inheritance. This keeps the original value and requested value as the parameter reference `#{ParameterName}`, but that avoids triggering any changes.

exceptionfactory (commented):
Thanks for the updates @pvillard31. On further review, there still appear to be some edge cases, particularly with handling of sensitive parameter values.

exceptionfactory (approved):
Thanks @pvillard31!

PR #11574, opened 2026-08-20
NIFI-16201: Reload Parameter-driven extra classpath for Connector Methods

markap14:
## Summary

- When a Parameter or asset changes a Processor or Controller Service extra classpath, stopped Working components never restart, so their `InstanceClassLoader` can omit the new JAR. Connector Method discovery and invocation then fail with `NoClassDefFoundError` / `LinkageError`.
- This change reloads additional classpath resources on Parameter update, always invokes Connector Methods on the live component, wraps `LinkageError` as `InvocationFailedException`, and treats `Throwable` as a failed async request so verification cannot hang.
- Adds `ClasspathVerifyIT` covering Processor verify, Connector Method `loadClass`, instance state across calls, and method-signature resolution when the classpath comes from a Parameter, Processor property, or Controller Service property.

https://issues.apache.org/jira/browse/NIFI-16201

## Test plan

- [x] `ClasspathVerifyIT` (9 tests)
- [x] Unit tests for `nifi-framework-core-api`, `nifi-framework-components`, `nifi-web-api`
- [x] Connector system tests
- [x] Checkstyle and PMD on modified modules
- [ ] CI (`./mvnw clean install -P contrib-check`)

### DIFF
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
@@ -2023,7 +2023,7 @@ public class StandardProcessorNode extends ProcessorNode implements Connectable 
     }
 
     @Override
-    public List<ConnectorMethod> getConnectorMethods() {
+    public List<ConnectorMethod> getConnectorMethods() throws InvocationFailedException {
         return getConnectorMethods(getProcessor().getClass());
     }
 
@@ -2082,6 +2082,11 @@ public class StandardProcessorNode extends ProcessorNode implements Connectable 
             } catch (final Exception e) {
                 throw new InvocationFailedException(e);
             }
+        } catch (final LinkageError e) {
+            // Argument types, return types, and the invoked method body can all reference classes supplied by additional classpath
+            // resources. Those failures are LinkageError, not Exception, so they would otherwise escape as an uncaught Error.
+            throw new InvocationFailedException("Failed to invoke Connector Method '" + methodName + "' on " + this
+                + " because a class required by the component could not be loaded from the component's ClassLoader", e);
         }
     }
 
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
@@ -2082,9 +2082,10 @@ public class StandardProcessorNode extends ProcessorNode implements Connectable 
             } catch (final Exception e) {
                 throw new InvocationFailedException(e);
             }
-        } catch (final LinkageError e) {
-            // Argument types, return types, and the invoked method body can all reference classes supplied by additional classpath
-            // resources. Those failures are LinkageError, not Exception, so they would otherwise escape as an uncaught Error.
+        } catch (final TypeNotPresentException | LinkageError e) {
+            // MethodArgument.type() throws TypeNotPresentException when an annotation Class member is absent from the component ClassLoader.
+            // Resolving the implementation Method's parameter types can still throw LinkageError. Discovery already wraps LinkageError, and
+            // Errors thrown from the invoked method body are wrapped by Method.invoke as InvocationTargetException.
             throw new InvocationFailedException("Failed to invoke Connector Method '" + methodName + "' on " + this
                 + " because a class required by the component could not be loaded from the component's ClassLoader", e);
         }
--- a/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/facades/standalone/StandaloneParameterContextFacade.java
+++ b/nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/facades/standalone/StandaloneParameterContextFacade.java
@@ -20,6 +20,7 @@ package org.apache.nifi.components.connector.facades.standalone;
 import org.apache.nifi.asset.Asset;
 import org.apache.nifi.components.connector.components.ParameterContextFacade;
 import org.apache.nifi.components.connector.components.ParameterValue;
+import org.apache.nifi.controller.ComponentNode;
 import org.apache.nifi.controller.FlowController;
 import org.apache.nifi.controller.ProcessorNode;
 import org.apache.nifi.controller.serialization.AffectedComponentSet;
@@ -97,8 +98,13 @@ public class StandaloneParameterContextFacade implements ParameterContextFacade 
 
         // A component whose additional classpath resources are supplied by a Parameter must be reloaded so that its ClassLoader reflects the
         // updated Parameter values. Components that are not running are not restarted below, so reloading here is the only point at which their
-        // ClassLoader is rebuilt.
-        allReferencingServices.forEach(ControllerServiceNode::reloadAdditionalResourcesIfNecessary);
+        // ClassLoader is rebuilt. Changing a Controller Service property can also change the Classloader Isolation Key of components that
+        // reference that service, matching StandardControllerServiceNode.setProperties().
+        for (final ControllerServiceNode serviceNode : allReferencingServices) {
+            serviceNode.reloadAdditionalResourcesIfNecessary();
+            serviceNode.getReferences().findRecursiveReferences(ComponentNode.class).forEach(ComponentNode::reloadAdditionalResourcesIfNecessary);
+        }
+
         allReferencingProcessors.forEach(ProcessorNode::reloadAdditionalResourcesIfNecessary);
 
         allReferencingServices.forEach(ControllerServiceNode::resetValidationState);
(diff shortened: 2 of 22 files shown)


markap14:
[cursor-grok-4.6] Code-compliance Validate failed on two Checkstyle `LineLength` violations in `ClasspathVerifyIT` (lines 129 and 141 were over 200 characters). Wrapped those `verifyConnectorStepConfigWithReferences` calls in ae4dfc5073f.

markap14:
[cursor-grok-4.6] Validate Pull Request failed because `ae4dfc50` was not a verified signed commit. Replaced it with signed `62fc913219e`.

markap14:
[cursor-grok-4.6] Pre-existing flake on **ubuntu-24.04 Java 25** (system-tests): [job 96569764221](https://github.com/apache/nifi/actions/runs/32413629346/job/96569764221) in [run 32413629346](https://github.com/apache/nifi/actions/runs/32413629346).

`OffloadIT.testOffload` timed out while waiting for a disconnected node, then `OffloadContentClaimTruncationIT` failed to start node 2 because the port was still in use. This PR does not change clustering or offload; the same tests passed on ubuntu-24.04 Java 21 and both macOS shards.

```
[ERROR]   OffloadIT.testOffload » Timeout testOffload() timed out after 10 minutes
[ERROR]   OffloadContentClaimTruncationIT>NiFiSystemIT.setup:134->...waitForAllNodesConnected:333
          Waited up to 60 seconds for all 2 nodes to connect but only 1 nodes connected.
            Node [localhost:5671] Status=CONNECTED
```

```
org.apache.nifi.toolkit.client.NiFiClientException: Error retrieving node status: Cannot replicate request to Node localhost:5672 because the node is not connected
	at org.apache.nifi.toolkit.client.impl.AbstractJerseyClient.executeAction(AbstractJerseyClient.java:107)
	...
	at org.apache.nifi.tests.system.clustering.OffloadIT.disconnectNode(OffloadIT.java:156)
	at org.apache.nifi.tests.system.clustering.OffloadIT.testIteration(OffloadIT.java:77)
```

```
Caused by: java.net.BindException: Address already in use
```

Reran the failed shard: `gh run rerun 32413629346 --failed`.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
```diff
@@ -2082,6 +2082,11 @@ public String invokeConnectorMethod(final String methodName, final Map<String, S
             } catch (final Exception e) {
                 throw new InvocationFailedException(e);
             }
+        } catch (final LinkageError e) {
+            // Argument types, return types, and the invoked method body can all reference classes supplied by additional classpath
+            // resources. Those failures are LinkageError, not Exception, so they would otherwise escape as an uncaught Error.
+            throw new InvocationFailedException("Failed to invoke Connector Method '" + methodName + "' on " + this
+                + " because a class required by the component could not be loaded from the component's ClassLoader", e);
```
sfc-gh-bbende: This `catch (LinkageError)` is mostly unreachable, and the comment describes coverage that it doesn't actually provide:

- The `getDeclaredMethods()` path that this JIRA is about now throws from `discoverConnectorMethod`, which this PR changed to wrap `LinkageError` into `InvocationFailedException` itself. That error is converted one layer down and never reaches here.
- "the invoked method body" is already handled today: a `NoClassDefFoundError` thrown by the target method is wrapped by `Method.invoke` into an `InvocationTargetException`, which is an `Exception` and is caught by the existing inner `catch (final Exception e)` just above.
- The gap that does remain is not a `LinkageError`. Resolving `methodArgument.type()` when that class is missing from the component's ClassLoader throws `TypeNotPresentException`, a `RuntimeException` produced by annotation `Class`-member resolution, so it still escapes unwrapped. The `AsyncRequestManager` change keeps that from hanging the request, but the caller gets a generic failure rather than an `InvocationFailedException`.

Suggest either dropping this block as redundant, or broadening it to also cover `TypeNotPresentException` and tightening the comment to describe what it really catches. The same applies to the identical block added to `StandardControllerServiceNode`.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/facades/standalone/StandaloneParameterContextFacade.java
```diff
@@ -95,8 +95,14 @@ public void updateParameters(final Collection<ParameterValue> updatedValues) {
         final Map<String, Parameter> updatedParameters = createParameterMap(updatedValues);
         managedProcessGroup.getParameterContext().setParameters(updatedParameters);
 
-        allReferencingProcessors.forEach(ProcessorNode::resetValidationState);
+        // A component whose additional classpath resources are supplied by a Parameter must be reloaded so that its ClassLoader reflects the
+        // updated Parameter values. Components that are not running are not restarted below, so reloading here is the only point at which their
+        // ClassLoader is rebuilt.
+        allReferencingServices.forEach(ControllerServiceNode::reloadAdditionalResourcesIfNecessary);
+        allReferencingProcessors.forEach(ProcessorNode::reloadAdditionalResourcesIfNecessary);
```
sfc-gh-bbende: This reloads the directly referencing components, which is narrower than the equivalent property-change path. `StandardControllerServiceNode.setProperties` reloads recursively, because changing a Controller Service property can alter the classloader isolation key of components that reference it:

```java
// It's possible that changing the properties of this Controller Service could alter the Classloader Isolation Key of a referencing
// component so reload any referencing component as necessary.
getReferences().findRecursiveReferences(ComponentNode.class).forEach(ComponentNode::reloadAdditionalResourcesIfNecessary);
```

A Parameter update can change that same isolation-key-relevant service property, so a stopped Processor that references one of these services will still keep a stale `InstanceClassLoader` after this call. Active referencing components are covered, since `activeSet` includes service references and gets restarted below, but stopped ones are not. Worth applying the same recursive reload for `allReferencingServices`.

nifi-framework-bundle/nifi-framework/nifi-framework-components/src/main/java/org/apache/nifi/controller/StandardProcessorNode.java
markap14: @sfc-gh-bbende Agreed. Addressed in f6be5026c9f: the outer catch is now `TypeNotPresentException | LinkageError`, and the comment describes the remaining paths (`MethodArgument.type()`, implementation Method parameter types, vs discovery wrapping `LinkageError` and method-body Errors arriving as `InvocationTargetException`). The same change is on `StandardControllerServiceNode`.

nifi-framework-bundle/nifi-framework/nifi-framework-core/src/main/java/org/apache/nifi/components/connector/facades/standalone/StandaloneParameterContextFacade.java
markap14: @sfc-gh-bbende Also addressed in f6be5026c9f. After reloading each service in `allReferencingServices`, we now call `findRecursiveReferences(ComponentNode.class)` and reload those components, matching `StandardControllerServiceNode.setProperties()`. Covered by `StandaloneParameterContextFacadeTest`.

bbende:
+1 latest updates look good, will merge after CI passes

markap14:
Two CI failures on `f6be5026` look unrelated to this PR. Reran only the failed shards.

**integration-tests — [Ubuntu Java 21](https://github.com/apache/nifi/actions/runs/32497455285/job/96819149028)** in [run 32497455285](https://github.com/apache/nifi/actions/runs/32497455285)

`CreateConnectorIT.testStopConnectorWithTimeoutStopsRunningConnector` timed out after 2 minutes waiting for a mock Connector to reach STOPPED. That test starts `GenerateAndLog` and immediately stops it; it does not exercise Parameter classpath reloads or Connector Method discovery. The other integration/compliance shards on this head passed.

```
[ERROR] org.apache.nifi.mock.connectors.tests.CreateConnectorIT.testStopConnectorWithTimeoutStopsRunningConnector -- Time elapsed: 120.9 s <<< FAILURE!
org.opentest4j.AssertionFailedError: Unexpected exception thrown: java.util.concurrent.TimeoutException: Timed out waiting for the Connector to stop after PT2M
	at org.apache.nifi.mock.connectors.tests.CreateConnectorIT.testStopConnectorWithTimeoutStopsRunningConnector(CreateConnectorIT.java:85)
Caused by: java.util.concurrent.TimeoutException: Timed out waiting for the Connector to stop after PT2M
	at org.apache.nifi.mock.connector.server.StandardConnectorMockServer.stopConnector(StandardConnectorMockServer.java:387)
```

**system-tests — [macos-15 Java 21](https://github.com/apache/nifi/actions/runs/32497455268/job/96819148793)** in [run 32497455268](https://github.com/apache/nifi/actions/runs/32497455268)

`ClusteredConnectorBacklogIT.testClusteredReadQueueAttributesLoadBalancedConnection` failed in teardown: delete of `BacklogReportingTestConnector` got HTTP 409 because 2 FlowFiles were still queued after a ROUND_ROBIN load-balanced purge. `ClasspathVerifyIT` passed on this shard, and ubuntu-24.04 Java 21/25 plus macos-15 Java 25 all passed.

```
[ERROR] ClusteredConnectorBacklogIT.testClusteredReadQueueAttributesLoadBalancedConnection -- Time elapsed: 11.44 s <<< ERROR!
org.apache.nifi.toolkit.client.NiFiClientException: Error deleting Connector: Node localhost:5671 is unable to fulfill this request due to: Cannot delete StandardConnectorNode[..., name=BacklogReportingTestConnector, state=STOPPED] because its Process Group has 2 FlowFiles queued; all FlowFiles must be removed before it can be deleted.
```

Node logs show the purge reported success, then the cluster DELETE was rejected for leftover queue data — in-flight load-balance leftovers, not a ClassLoader reload.

Reran: `gh run rerun 32497455285 --failed` and `gh run rerun 32497455268 --failed`.

